PR #46, opened 2016-08-11
Add M3TS encoding scheme

martin-mao:
Adding M3TS encoding scheme and refactoring existing encoding package structure


encoding/encoding.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+// SignExtend sign extends the highest bit of v which has numBits (<=64)
+func SignExtend(v uint64, numBits int) int64 {
+	shift := uint(64 - numBits)
+	return (int64(v) << shift) >> shift
+}
+
+// GenerateFloatVal generates a random float val given the number of digits and decimal places
+func GenerateFloatVal(r *rand.Rand, numDig, numDec int) float64 {
```
robskillington: Can we put this in a separate module considering it's only included by tests and benchmarks?

storage/options.go
```diff
@@ -174,41 +191,68 @@ func (o *options) EncodingTszPooled() Options {
...
-	return (&opts).encodingTsz(encodingOpts)
+	return (&opts).encoding(useM3ts)
 }
 
-func (o *options) EncodingTsz() Options {
-	return o.encodingTsz(tsz.NewOptions())
-}
-
-func (o *options) encodingTsz(encodingOpts tsz.Options) Options {
+func (o *options) encoding(useM3ts bool) Options {
```
robskillington: Instead of bool probably best to make this a switch on an enum yeah?

storage/options.go
```diff
@@ -154,6 +155,22 @@ func (o *options) GetCommitLogOptions() commitlog.Options {
...
+
+func (o *options) EncodingM3tsPooled() Options {
+	return o.encodingPooled(true)
+}
+
+func (o *options) EncodingM3ts() Options {
+	return o.encoding(true)
+}
+
+func (o *options) encodingPooled(useM3ts bool) Options {
```
robskillington: Instead of bool probably best to make this a switch on an enum yeah?

encoding/m3tsz/decoder.go
```diff
@@ -27,18 +27,19 @@ import (
...
 
 type decoder struct {
-	opts Options
+	opts         encoding.Options
+	intOptimized bool
 }
 
 // NewDecoder creates a decoder.
-func NewDecoder(opts Options) encoding.Decoder {
+func NewDecoder(opts encoding.Options, intOptimized bool) encoding.Decoder {
```
robskillington: Nit: can we make `opts` be the second parameter here?  This is traditionally the pattern we have with other components that take `opts` as well as some parameters in their constructors.

encoding/m3tsz/iterator.go
```diff
@@ -35,33 +35,39 @@ import (
...
 	tu        xtime.Unit    // current time unit
 	tuChanged bool          // whether we have a new time unit
-
-	closed bool
+	closed    bool
 }
 
 // NewReaderIterator returns a new iterator for a given reader
-func NewReaderIterator(reader io.Reader, opts Options) encoding.ReaderIterator {
+func NewReaderIterator(reader io.Reader, opts encoding.Options, intOptimized bool) encoding.ReaderIterator {
```
robskillington: Nit: can we make opts be the last parameter here? This is traditionally the pattern we have with other components that take opts as well as some parameters in their constructors.

encoding/m3tsz/encoder.go
```diff
@@ -238,17 +255,84 @@ func (enc *encoder) writeDeltaOfDeltaTimeUnitUnchanged(prevDelta, curDelta time.
...
-	xor := enc.vb ^ vb
+	if !enc.intOptimized {
+		enc.writeFloatXOR(math.Float64bits(v))
+		return
+	}
+
+	// Attempt to convert float to int for int optimization
+	val, mult, floatVal := convertToIntFloat(v, enc.maxMult)
+	if floatVal {
+		enc.writeFloatVal(math.Float64bits(val), mult)
```
robskillington: Not doing `xor` for when floats?

encoding/m3tsz/encoder.go
```diff
@@ -273,6 +357,108 @@ func (enc *encoder) writeXOR(prevXOR, curXOR uint64) {
...
+	}
+
+	neg := false
+	if valDiff < 0 {
+		neg = true
+		valDiff = -1 * valDiff
+	}
+
+	valDiffBits := uint64(int64(valDiff))
+	numSig := encoding.NumSig(valDiffBits)
```
robskillington: Does `NumSig` need to be in `encoding` now that tsz is combined with m3ts?

encoding/m3tsz/encoder.go
```diff
@@ -140,7 +157,7 @@ func (enc *encoder) writeAnnotation(ant ts.Annotation) {
 		return
 	}
 	scheme := enc.opts.GetMarkerEncodingScheme()
-	writeSpecialMarker(enc.os, scheme, scheme.Annotation())
+	encoding.WriteSpecialMarker(enc.os, scheme, scheme.Annotation())
```
robskillington: Does `WriteSpecialMarker ` need to be in `encoding` now that tsz is combined with m3ts?

encoding/m3tsz/encoder.go
```diff
@@ -273,6 +357,108 @@ func (enc *encoder) writeXOR(prevXOR, curXOR uint64) {
...
+		if enc.numLowerSig >= sigRepeatThreshold {
+			newSig = enc.curLowestSig
+			enc.numLowerSig = 0
+		}
+
+	} else {
+		enc.numLowerSig = 0
+	}
+
+	return newSig
```
robskillington: `enc.numSig` not set inside this function, is that "ok"?  It seems this function does have side effects for other properties on the encoder, why not make it also set `enc.numSig` and rename the function to `setNewSig(...)` to make it consistent?

encoding/m3tsz/encoder.go
```diff
@@ -315,14 +507,13 @@ func (enc *encoder) Stream() xio.SegmentReader {
 		tail = scheme.Tail(b[blen-1], pos)
 	}
 
-	segment := ts.Segment{Head: head, Tail: tail, TailShared: true}
 	readerPool := enc.opts.GetSegmentReaderPool()
 	if readerPool != nil {
 		reader := readerPool.Get()
-		reader.Reset(segment)
+		reader.Reset(ts.Segment{Head: head, Tail: tail})
```
robskillington: Why removing `TailedShared: true`?  This will cause the tail to be returned a bytes pool.

robskillington:
Done with first round.

encoding/m3tsz/encoder.go
martin-mao: We can set in here - I was trying to avoid the extra redundant assignment/check in this function since we already have to check this in the caller function for when the numSig is different. Every NS counts when it only takes like 200 per datapoint.

encoding/m3tsz/encoder.go
martin-mao: Not sure, think this was a merge issue. Will revert.

encoding/m3tsz/encoder.go
robskillington: Hm, it's just weird that it mutates some encoder state but not others.  Perhaps we just rename this to `accountAndGetNewSig(...)` to reflect that it's doing some accounting as a side effect of calling it?

client/options.go
```diff
@@ -194,10 +197,10 @@ func (o *options) GetInstrumentOptions() instrument.Options {
 	return o.instrumentOpts
 }
 
-func (o *options) EncodingTsz() Options {
+func (o *options) EncodingM3TSZ() Options {
 	opts := *o
 	opts.readerIteratorAllocate = func(r io.Reader) encoding.ReaderIterator {
-		return tsz.NewReaderIterator(r, tsz.NewOptions())
+		return m3tsz.NewReaderIterator(r, defaultIntOptimizationEnabled, encoding.NewOptions())
```
xichen2020: `defaultIntOptimizationEnabled` is defined both here and in `storage/options.go`. Can we make it part of the encoding options so you can use it via something like `encoding.NewOptions().GetIntOptimizationEnabled()` down below?

client/types.go
```diff
@@ -126,8 +126,8 @@ type Options interface {
 	// GetInstrumentOptions returns the instrumentation options
 	GetInstrumentOptions() instrument.Options
 
-	// EncodingTsz sets tsz encoding
-	EncodingTsz() Options
+	// EncodingM3TSZ sets tsz encoding
```
xichen2020: `...sets m3tsz encoding`

encoding/encoding.go
```diff
@@ -18,9 +18,28 @@
...
+// Bit is just a byte
+type Bit byte
+
+// NumSig returns the number of significant values in a uint64
+func NumSig(v uint64) uint8 {
+	if v == 0 {
+		return 0
+	}
+
+	numLeading := uint8(0)
```
xichen2020: nit: can pull out the logic to calculate leading zeros into a separate function and use it both here and `LeadingAndTrailingZeros`

encoding/types.go
```diff
@@ -151,6 +197,28 @@ type EncoderAllocate func() Encoder
 // ReaderIteratorAllocate allocates a ReaderIterator for a pool.
 type ReaderIteratorAllocate func(reader io.Reader) ReaderIterator
 
+// Istream encapsulates a readable stream.
+type Istream interface {
```
xichen2020: Perhaps this should be `IStream`?

encoding/types.go
```diff
@@ -151,6 +197,28 @@ type EncoderAllocate func() Encoder
...
+	ReadBit() (Bit, error)
+	ReadByte() (byte, error)
+	ReadBits(numBits int) (uint64, error)
+	PeekBits(numBits int) (uint64, error)
+	Reset(r io.Reader)
+}
+
+// Ostream encapsulates a writable stream.
+type Ostream interface {
```
xichen2020: Perhaps this should be `OStream`?

encoding/m3tsz/decoder.go
```diff
@@ -27,18 +27,19 @@ import (
 )
 
 type decoder struct {
-	opts Options
+	opts         encoding.Options
+	intOptimized bool
```
xichen2020: Can we make this part of the encoding options? Seems cleaner that way

storage/options.go
```diff
@@ -174,49 +177,49 @@ func (o *options) EncodingTszPooled() Options {
...
+	encodingOpts := encoding.NewOptions().
 		BytesPool(bytesPool).
 		EncoderPool(encoderPool).
 		ReaderIteratorPool(readerIteratorPool).
 		SegmentReaderPool(segmentReaderPool)
 
 	// initialize encoder pool
 	encoderPool.Init(func() encoding.Encoder {
-		return tsz.NewEncoder(timeZero, nil, encodingOpts)
+		return m3tsz.NewEncoder(timeZero, nil, defaultIntOptimizationEnabled, encodingOpts)
```
xichen2020: yeah if we make `intOptimizationEnabled` part of the encoding options, you won't need to pass it in explicitly

storage/types.go
```diff
@@ -207,11 +207,11 @@ type Options interface {
...
 	GetCommitLogOptions() commitlog.Options
 
-	// EncodingTszPooled sets tsz encoding with pooling
-	EncodingTszPooled() Options
+	// EncodingM3TSZPooled sets m3tsz encoding with pooling
+	EncodingM3TSZPooled() Options
 
-	// EncodingTsz sets tsz encoding
-	EncodingTsz() Options
+	// EncodingM3TSZ sets tsz encoding
```
xichen2020: `...set m3tsz encoding`

xichen2020:
LGTM

PR #48, opened 2016-08-15
Merge ShardScheme and ShardSet

cw9:


cw9:

sharding/types.go
```diff
@@ -23,23 +23,14 @@ package sharding
...
-// ShardScheme is a sharding scheme
-type ShardScheme interface {
-	// Shard will return a shard for a given identifer
+// ShardSet contains a sharding function and a set of shards, this interface allows for potentially out of order shard sets
+type ShardSet interface {
+	// Shard will return a shard for a given identifier
 	Shard(identifer string) uint32
 
 	// CreateSet will return a new shard set from a set of shards
 	CreateSet(shards []uint32) (ShardSet, error)
```
cw9: CreateSet seems to be used only in test, should I remove this?

xichen2020: Yeah seems similar to `NewShardSet` in functionality, if it's not used anywhere, let's remove it.

robskillington: How do you create a subset from a top level shardset then with the same `HashFn`?  Perhaps at least add a method "HashFn() HashFn" which returns the `HashFn` so that someone can then create a subset of the shards with the same `HashFn`.

storage/database.go
```diff
@@ -100,7 +99,7 @@ type db struct {
 
 	// Contains an entry to all shards for fast shard lookup, an
 	// entry will be nil when this shard does not belong to current database
-	shards []databaseShard
+	shards map[uint32]databaseShard
```
robskillington: Specifically did not want to use a map here as map access is significantly slower than array lookup.  Can you revert to `[]databaseShard`?

cw9: but since the shards here are no longer consecutive, we will have to loop through the array to look up a particular shard.

sharding/types.go
cw9: If we do need to create subset from top level shardset then we should probably keep the createSet function, but do we?

robskillington: Not right now, let's just add a `HashFn` however at least so you can get this back should need it for construction of other objects.  Right now not being able to get the `HashFn` out could make building utilities/whatever else an upstream change that is very easy for us to just add now as a getter.

storage/database.go
robskillington: Not true, we always look it up directly based on the shard ID, e.g.:
```go
	d.RLock()
	shardID := d.shardScheme.Shard(id)
	shard := d.shards[shardID]
	d.RUnlock()

	if shard == nil {
		return fmt.Errorf("not responsible for shard %d", shardID)
	}
	return shard.Write(ctx, id, timestamp, value, unit, annotation)
```

sharding/types.go
cw9: sure thing

robskillington:
LGTM

sharding/scheme.go
```diff
@@ -22,73 +22,59 @@ package sharding
...
-	scheme ShardScheme
+func (s *shardSet) Shards() []uint32 {
+	return s.shards[:]
 }
 
-// NewShardSet creates a new shard set
-func NewShardSet(shards []uint32, scheme ShardScheme) ShardSet {
-	return &shardSet{shards, scheme}
+func (s *shardSet) Min() uint32 {
+	min := uint32(math.MaxUint32)
```
xichen2020: can probably cache this since `shards` are immutable once they are created.

sharding/scheme.go
```diff
@@ -22,73 +22,59 @@ package sharding
...
+			min = shard
+		}
+	}
+	return min
 }
 
-func (s *shardSet) Shards() []uint32 {
-	return s.shards[:]
+func (s *shardSet) Max() uint32 {
+	max := uint32(0)
```
xichen2020: same here

sharding/scheme.go
cw9: can do, but Min() is not being used now and Max() is only used when creating a DB, not sure how much value added from the caching

PR #52, opened 2016-08-18
Server side implementation for peer streaming

xichen2020:


network/server/tchannelthrift/node/service.go
```diff
@@ -106,6 +107,23 @@ func (s *service) Fetch(tctx thrift.Context, req *rpc.FetchRequest) (*rpc.FetchR
...
+	s := &rpc.Segments{}
+	if len(readers) == 1 {
+		seg := readers[0].Segment()
+		s.Merged = &rpc.Segment{Head: seg.Head, Tail: seg.Tail}
+	} else {
+		for _, reader := range readers {
+			seg := reader.Segment()
+			s.Unmerged = append(s.Unmerged, &rpc.Segment{Head: seg.Head, Tail: seg.Tail})
+		}
+	}
```
robskillington: Could reduce indentation here somewhat:
```go
if len(readers) == 1 {
  seg := readers[0].Segment()
  s.Merged = &rpc.Segment{Head: seg.Head, Tail: seg.Tail}
  return
}

// handle unmerged
```

network/server/tchannelthrift/node/service.go
```diff
@@ -106,6 +107,23 @@ func (s *service) Fetch(tctx thrift.Context, req *rpc.FetchRequest) (*rpc.FetchR
 	return result, nil
 }
 
+func toSegments(readers []xio.SegmentReader) *rpc.Segments {
```
robskillington: Maybe this should be in just the `tchannelthrift` package is it's not really related to the `node` service itself.

network/server/tchannelthrift/node/service.go
```diff
@@ -133,30 +151,91 @@ func (s *service) FetchRawBatch(tctx thrift.Context, req *rpc.FetchRawBatchReque
...
+	}
+	return res, nil
 }
 
-func (s *service) FetchBlocksMetadata(ctx thrift.Context, req *rpc.FetchBlocksMetadataRequest) (*rpc.FetchBlocksMetadataResult_, error) {
-	return nil, fmt.Errorf("not implemented")
+func (s *service) FetchBlocksMetadata(tctx thrift.Context, req *rpc.FetchBlocksMetadataRequest) (*rpc.FetchBlocksMetadataResult_, error) {
+	if req.Limit <= 0 {
+		return nil, nil
+	}
```
robskillington: Perhaps do this below the context creation just for consistency, shouldn't hurt

network/server/tchannelthrift/node/service.go
```diff
@@ -133,30 +151,91 @@ func (s *service) FetchRawBatch(tctx thrift.Context, req *rpc.FetchRawBatchReque
...
+			blockMetadata := rpc.NewBlockMetadata()
+			blockMetadata.Start = xtime.ToNanoseconds(fetchedMetadataBlock.Start())
+			blockMetadata.Size = fetchedMetadataBlock.Size()
+			blocksMetadata.Blocks[j] = blockMetadata
+		}
+		result.Elements[i] = blocksMetadata
+	}
+	if err != nil {
+		result.Err = tterrors.NewInternalError(err)
+	}
```
robskillington: This err case should be just after `FetchBlocksMetadata` call yeah?

storage/buffer.go
```diff
@@ -185,12 +186,6 @@ func (b *dbBuffer) ReadEncoded(ctx context.Context, start, end time.Time) [][]xi
 	// TODO(r): pool these results arrays
 	var results [][]xio.SegmentReader
 	b.forEachBucketAsc(now, func(bucket *dbBufferBucket, current time.Time) {
-		shouldRead, _, _ := b.bucketState(now, bucket, current)
-		if !shouldRead {
-			// Requires reset and drained already or not written to
-			return
-		}
-
```
robskillington: Why are we removing this case out of interest?

storage/buffer.go
```diff
@@ -199,26 +194,97 @@ func (b *dbBuffer) ReadEncoded(ctx context.Context, start, end time.Time) [][]xi
...
-			stream := bucket.encoders[i].encoder.Stream()
-			if stream == nil {
-				// TODO(r): log an error and emit a metric, this is pretty bad as this
-				// encoder should have values if "shouldRead" returned true
-				continue
+		b.readBucketStreams(ctx, bucket, now, current, func(stream xio.SegmentReader) {
+			unmerged = append(unmerged, stream)
+		})
+
+		results = append(results, unmerged)
```
robskillington: Is unmerged guarenteed to be of some size?  Probably need to avoid adding nil entries to results here.

network/server/tchannelthrift/node/service.go
xichen2020: Sure can do

network/server/tchannelthrift/node/service.go
xichen2020: can do

network/server/tchannelthrift/node/service.go
xichen2020: Putting it at the top avoids a context get and defer but yeah can do

storage/buffer.go
```diff
@@ -199,26 +194,97 @@ func (b *dbBuffer) ReadEncoded(ctx context.Context, start, end time.Time) [][]xi
...
+	return results
+}
+
+func (b *dbBuffer) FetchBlocks(ctx context.Context, starts []time.Time) []FetchBlockResult {
+	var res []FetchBlockResult
+
+	now := b.nowFn()
+	b.forEachBucketAsc(now, func(bucket *dbBufferBucket, current time.Time) {
+		found := false
+		// starts has ~24 items, linear search should be okay time-wise to
```
robskillington: Is starts sorted ascending?  if so we could search in reverse and once we pass the third in reverse we know no other start will match.  

i.e. We probably should not bank on this being 24.   Someone could use a bucket size of 1hr and perform 4 day retention for instance.

robskillington: Would perhaps remove this comment specifically saying 24 items.  Just mention that it's a low order of magnitude perhaps.  It's never great to attach fixed numbers and expectations to code.

storage/buffer.go
```diff
@@ -199,26 +194,97 @@ func (b *dbBuffer) ReadEncoded(ctx context.Context, start, end time.Time) [][]xi
...
+	var res []block.DatabaseBlockMetadata
+
+	now := b.nowFn()
+	b.forEachBucketAsc(now, func(bucket *dbBufferBucket, current time.Time) {
+		var pSize *int64
+		if includeSizes {
+			var size int64
+			b.readBucketStreams(ctx, bucket, now, current, func(stream xio.SegmentReader) {
+				segment := stream.Segment()
+				size += int64(len(segment.Head) + len(segment.Tail))
```
robskillington: Probably want to `stream.Close()` before the end of this method here, otherwise the segment reader won't get returned to the pool.

network/server/tchannelthrift/node/service.go
xichen2020: The reason why it's here is that we are taking a best effort approach and fetch as much as we can even if there are errors occurred while fetching the metadata. If we return after `FetchBlocksMetadata`, the partial metadata fetched won't make it back to the client.

storage/buffer.go
xichen2020: The logic has been moved to `readBucketStreams`.

storage/buffer.go
xichen2020: The nil check is done inside `readBucketStreams`

storage/buffer.go
```diff
@@ -199,26 +194,97 @@ func (b *dbBuffer) ReadEncoded(ctx context.Context, start, end time.Time) [][]xi
...
 			})
+			// If we have no data in this bucket, return early without
+			// appending it to the result.
```
robskillington: Seems like we don't shortcut with this branch if `includeSizes == false`, probably should include this shortcut for both cases or neither.  I'm fine with either approach, although probably in preference of always not including if size is zero as that's more representative of what the server has (i.e. we never flush these blocks if they never get data written to them so they're almost like "standby blocks" really at this point in time).

storage/buffer.go
xichen2020: Sure can do

storage/buffer.go
```diff
@@ -199,26 +194,97 @@ func (b *dbBuffer) ReadEncoded(ctx context.Context, start, end time.Time) [][]xi
...
+		stream := bucket.encoders[i].encoder.Stream()
+		if stream == nil {
+			// TODO(r): log an error and emit a metric, this is pretty bad as this
+			// encoder should have values if "shouldRead" returned true
+			continue
 		}
 
-		results = append(results, unmerged)
-	})
+		ctx.RegisterCloser(func() {
```
robskillington: We can get a tiny bit more clever here with:
```go
ctx.RegisterCloser(stream.Close)
```

storage/database.go
```diff
@@ -203,25 +205,67 @@ func (d *db) Write(
...
+func (d *db) FetchBlocks(
+	ctx context.Context,
+	id string,
+	starts []time.Time,
+) []FetchBlockResult {
+	shardID := d.shardSet.Shard(id)
+	shard, err := d.readableShard(shardID)
+	if err != nil {
+		// If we don't own the shard associated with the id, return an error
+		// so the client knows it needs to retry these.
```
robskillington: For this case it's actually fine to just return a top level `xerrors.NewInvalidParamsError(...)`.  I perform error checking first at the top level and secondary at the block level if there are partial success and partial errors.  For entire request failed I'd much prefer to get just a full top level error (not wrapped in the result/response object).

storage/database.go
```diff
@@ -304,7 +348,7 @@ func (d *db) splayedTick() {
 	}
 }
 
-func (d *db) nextUniqueIndex() uint64 {
+func (d *db) next() uint64 {
```
robskillington: Any reason to change this?  `next()` gives no context for what it's returning or who might use this function.

storage/buffer.go
xichen2020: Agreed with we shouldn't include blocks with size zero but not sure I understood your concern here, the logic is inside the `if includSizes {...}`, so if `includeSizes == false`, we won't even get here.

storage/buffer.go
xichen2020: Yeah can do

storage/series.go
```diff
@@ -233,6 +233,80 @@ func (s *dbSeries) ReadEncoded(
...
+		res      []block.DatabaseBlockMetadata
+		multiErr xerrors.MultiError
+	)
+
+	s.RLock()
+
+	// Iterate over the data blocks
+	blocks := s.blocks.GetAllBlocks()
+	for t, b := range blocks {
+		reader, err := b.Stream(ctx)
```
robskillington: Might be worth creating a temporary context and passing it here, otherwise for the entire request all the StreamReaders are going to be not returned to the pool until all series and all block metadatas have been sent back to the client.

storage/series.go
```diff
@@ -233,6 +233,80 @@ func (s *dbSeries) ReadEncoded(
...
+		if err != nil {
+			multiErr = multiErr.Add(fmt.Errorf("unable to retrieve block stream for series %s time %v: %v", s.seriesID, t, err))
+			continue
+		}
+		var pSize *int64
+		if includeSizes {
+			if reader == nil {
+				// If there are no datapoints in the block, continue
+				// and don't append it to the result.
+				continue
```
robskillington: Again would be good to shortcut irregardless of `includeSizes` being true or false.

storage/shard.go
```diff
@@ -107,13 +108,13 @@ func (s *dbShard) forEachSeries(continueOnError bool, seriesFn databaseSeriesFn)
 	s.RUnlock()
 
 	// TODO(xichen): pool or cache this.
-	curSeries := make([]databaseSeries, 0, batchSize)
+	curEntries := make([]*dbShardEntry, 0, batchSize)
```
robskillington: Could we call this `currEntries` for consistency? I use `currThing` instead of `curThing` very frequently.  We could also just do a grep and see which one is more common and apply that, I don't mind either way but would like to be consistent.

storage/database.go
xichen2020: Sure did it this way because i wasn't sure if you were performing top-level checks. If you are, agreed we should just return a top-level error.

storage/database.go
xichen2020: We already have `index` in the interface name `increasingIndex`, so figured `next` is self- explanatory enough. Can change it to `nextIndex` if that's more clear.

storage/series.go
xichen2020: Not sure I understood what you meant by "shortcut". Can you elaborate?

storage/shard.go
xichen2020: sure `currEntries` it is.

storage/series.go
xichen2020: Yeah sure can do

storage/buffer.go
```diff
@@ -199,28 +194,95 @@ func (b *dbBuffer) ReadEncoded(ctx context.Context, start, end time.Time) [][]xi
...
 		results = append(results, unmerged)
 	})
 
 	return results
 }
 
+func (b *dbBuffer) FetchBlocks(ctx context.Context, starts []time.Time) []FetchBlockResult {
+	var res []FetchBlockResult
+
+	now := b.nowFn()
```
martin-mao: Why not just pass b.NowFn() into forEachBucket?

storage/series.go
```diff
@@ -233,6 +233,79 @@ func (s *dbSeries) ReadEncoded(
 	return results, nil
 }
 
+func (s *dbSeries) FetchBlocks(ctx context.Context, starts []time.Time) []FetchBlockResult {
+	var res []FetchBlockResult
```
martin-mao: Nit: Make this with cap of len(starts) + 1?

storage/series.go
```diff
@@ -233,6 +233,79 @@ func (s *dbSeries) ReadEncoded(
...
+	s.RUnlock()
+
+	sortFetchBlockResultByTimeAscending(res)
+
+	return res
+}
+
+func (s *dbSeries) FetchBlocksMetadata(ctx context.Context, includeSizes bool) (block.DatabaseBlocksMetadata, error) {
+	var (
+		res      []block.DatabaseBlockMetadata
```
martin-mao: Add TODO to make bucket pools of these?

storage/database.go
robskillington: Yeah perhaps `nextIndex`?

storage/database.go
robskillington: Right I see, yeah didn't realize the fetchblocks call doesn't have a shard specified in it by default.  Wondering if we should actually add that to the call to reduce having to hash all the IDs, also means we can return one top level error if the shard isn't owned rather than have to return an error per ID per block?  It sounds like a good idea TBH.

network/server/tchannelthrift/node/service.go
robskillington: I see, what are the potential errors that occur here and there's no reason we can add them to the block level error?

storage/buffer.go
robskillington: Sounds good.

storage/buffer.go
robskillington: Makes sense.

storage/buffer.go
robskillington: Yeah, probably is ok.  Also we'll need the sizes the majority of the time (this is primarily used for peer streaming and peer streaming requires the size).  Perhaps we can cache the size somewhere in the future.

network/server/tchannelthrift/node/service.go
xichen2020: When trying to read a shard that's not owned by this node, or when reading a block that's already closed, the error will bubble up here. The former is a top-level error, so we can return a bad request error when that happens. The latter however is a block-level error, so we'll need to add a block-level error field to the result struct, i.e.,
```
struct BlockMetadata {
	1: required i64 start
	2: optional i64 size
        3: optional Error err
}
```
so we don't wrap everything up all in one error object.

storage/database.go
xichen2020: Sure yeah, that means the client needs to make sure to always send IDs in the same shard to the server for the `FetchBlocks` call. If that doesn't require a lot of client-side changes, then  yeah let's do that.

storage/database.go
xichen2020: Done

storage/buffer.go
xichen2020: Just felt more readable that way, but yeah sure can do.

storage/series.go
xichen2020: Yeah can do

storage/series.go
xichen2020: For now this is only called at the beginning of peer streaming, which only happens during bootstrapping, so I'd rather let them GC'ed than keeping them around in a pool. I guess in the future if we need background repairs we might need to call this function semi-frequently, yeah can add the TODO.

storage/buffer.go
xichen2020: Actually `now` is also passed to `readBucketStreams`, so going to leave it as is.

storage/shard.go
```diff
@@ -227,46 +236,103 @@ func (s *dbShard) ReadEncoded(
...
+
+		return nil
+	}, func(entry *dbShardEntry) bool {
+		// Break out of the iteration loop once we've accumulated enough entries.
+		if int64(len(res)) < limit {
+			return false
+		}
+		nextPageToken := int64(entry.index)
+		pNextPageToken = &nextPageToken
+		return true
```
robskillington: Hey so what if the stop function doesn't get called because we run out of results, will `nextPageToken` get correctly set?  If so how?  We need to return a next token that's going to not cause another read, i.e. a non-existent index I would think.

xichen2020: If we run out of results, we'll return nil for the `nextPageToken`, which the client side can use to determine we've exhausted all entries.

robskillington:
LGTM, I'll update client side to respect this nil behavior on return

PR #54, opened 2016-08-23
Add peers bootstrapper

robskillington:
cc @xichen2020 @martin-mao


client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+			default:
+			}
+		}
+		waitDone = func() error {
+			return <-doneCh
+		}
+	)
+
+	s.RLock()
+	var peers []hostQueue
```
martin-mao: Nit: make peers with cap of len(s.queues)?

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+			}
+		}()
+	}
+
+	wg.Wait()
+
+	errLock.RLock()
+	defer errLock.RUnlock()
+
+	if errLen == len(peers) {
```
martin-mao: Should we log a warning if errLen > 0?

robskillington: We already do on line 679 at the completion of each failure to stream metadata from each peer.

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+
+func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
+	perPeerBlocksMetadata []*blocksMetadata,
+	peerQueues peerBlocksQueues,
+	pooledCurrStart, pooledCurrEligible []*blocksMetadata,
+	pooledBlocksMetadataQueues []blocksMetadataQueue,
+) error {
+	// Use pooled arrays
+	var (
+		currStart            = pooledCurrStart[:0]
```
martin-mao: Hmm, is the assumption that all users of these pooled arrays have to reset themselves? Why not have the pool reset on fetch/put?

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+			received.submitted = true
+		}
+	}
+
+	// Enqueue all unsubmitted received metadata
+	for _, received := range metadata {
+		if received.submitted {
+			continue
+		}
+		enqueueCh.enqueue(received.results)
```
martin-mao: No need to check if len(received.results) == peersLen? We're ok if it's not right?

client/session.go
robskillington: There is no pool per say, it's literally just a single array and we pass that into this method each time.  That avoids any pool get/put contention as the caller of this method calls in sequentially (no concurrency).

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+	}
+}
+
+func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
```
martin-mao: This is one loooooong function.

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+	blocksResult *blocksResult,
+	enqueueCh *enqueueChannel,
+) {
+	// Prepare request
+	var (
+		req    = rpc.NewFetchBlocksRequest()
+		result *rpc.FetchBlocksResult_
+	)
+	for i := range batch {
+		var starts []int64
```
martin-mao: Nit: make this with len(batch[i].blocks) and then just assign to position j in following loop?

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+			if block.Start != batch[i].blocks[j].start.UnixNano() {
+				failed := []blockMetadata{batch[i].blocks[j]}
+				s.streamBlocksReattemptFromPeers(failed, enqueueCh)
+				s.log.WithFields(
+					xlog.NewLogField("id", id),
+					xlog.NewLogField("expectedStart", batch[i].blocks[j].start.UnixNano()),
+					xlog.NewLogField("actualStart", block.Start),
+					xlog.NewLogField("indexID", i),
+					xlog.NewLogField("indexBlock", j),
+				).Errorf("stream blocks response from peer %v returned mismatched ID", peer.Host().String())
```
martin-mao: Nit: Log msg should be returned mismatched Start Time

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+				continue
```
martin-mao: Is it worth continuing if one of the starts are off? Does that screw up the whole block or do we want to return a partial block?

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+			if block.Err != nil {
+				failed := []blockMetadata{batch[i].blocks[j]}
+				s.streamBlocksReattemptFromPeers(failed, enqueueCh)
+				s.log.WithFields(
+					xlog.NewLogField("id", id),
+					xlog.NewLogField("expectedStart", batch[i].blocks[j].start.UnixNano()),
+					xlog.NewLogField("actualStart", block.Start),
+					xlog.NewLogField("indexID", i),
+					xlog.NewLogField("indexBlock", j),
+				).Errorf("stream blocks response from peer %v returned mismatched block start", peer.Host().String())
```
martin-mao: Error message here incorrect

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+) {
+	// Must do this asynchronously or else could get into a deadlock scenario
+	// where cannot enqueue into the reattempt channel because no more work is
+	// getting done because new attempts are blocked on existing attempts completing
+	// and existing attempts are trying to enqueue into a full reattempt channel
+	enqueue := enqueueCh.enqueueDelayed()
+	s.streamBlocksReattemptWorkers.Go(func() {
+		for i := range blocks {
+			// Reconstruct peers metadata for reattempt
+			peersLen := len(blocks[i].reattempt.peersMetadata)
```
martin-mao: Nit: Why not just combine these lines since peersLen is only used on the next line?

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+		iter := encoding.NewMultiReaderIterator(alloc, nil)
+		iter.Reset(readers)
+
+		encoder := r.encoderPool.Get()
+		for iter.Next() {
+			dp, unit, annotation := iter.Current()
+			encoder.Encode(dp, unit, annotation)
+		}
+		if err := iter.Err(); err != nil {
+			return err
```
martin-mao: Return encoder back to the pool?

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+func (q *peerBlocksQueue) enqueue(bm *blocksMetadata, doneFn func()) {
+	q.Lock()
+
+	if len(q.queue) == 0 && cap(q.queue) < q.maxQueueSize {
+		// Lazy initialize queue
+		q.queue = make([]*blocksMetadata, 0, q.maxQueueSize)
+	}
+	q.queue = append(q.queue, bm)
+	if doneFn != nil {
+		q.doneFns = append(q.doneFns, doneFn)
```
martin-mao: Why not lazily initialize the doneFns slice as well above?

client/session.go
robskillington: Yeah so this is after the metadata channel has closed, i.e. no more results.  The case where `len(received.results) != peersLen` is if one of the peers stopped receiving results, in that case we still want to try streaming data from the remaining peer, hence this is ok and is a case we want to support.

client/session.go
robskillington: This was apart of another function already, I'm kind of inclined not too tbh just based on what it might look like was kinda confusing when I tried to do this.

client/session.go

client/session.go
robskillington: Good catch, ta

client/session.go
robskillington: Yeah should do, will add.

client/session.go
robskillington: Yeah, I can just do `encoder.Close()` which will return to pool.  Will do this.

client/session.go
robskillington: Kind of a long line, will combine though and add a new line and indent the value assignment rhs.

client/session.go
robskillington: Yup fixed that.

client/session.go
robskillington: Since I re-enqueue bad blocks for retry one by one here to ensure the other ones get re-enqueued too I need to continue.  This is fine.

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			err := s.streamBlocksMetadataFromPeer(shard, peer, ch)
+			if err != nil {
+				errLock.Lock()
+				defer errLock.Unlock()
+				errLen++
+				multiErr.Add(err)
+				s.log.Warnf("failed to stream blocks metadata from peer %v for shard %d: %v", peer.Host().String(), shard, err)
```
xichen2020: nit: use `%s` for strings

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+			}
+		}()
+	}
+
+	wg.Wait()
+
+	errLock.RLock()
```
xichen2020: Why do we need to `RLock` here? seems all goroutines should have finished at this point so we can just read `errLen` without locks, which means we can probably change `errLock` to a simple mutex.

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+	defer errLock.RUnlock()
+
+	if errLen == len(peers) {
+		s.log.Errorf("failed to complete streaming blocks from all peers for shard %d", shard)
+		return multiErr
```
xichen2020: Perhaps do `multiErr.FinalError()` as we do elsewhere?

client/session.go
robskillington: Done.

client/session.go
robskillington: Yeah, can do, wanted to avoid data race complaining but probably is smart enough to realize this although not 100% sure.

client/session.go
```diff
@@ -563,3 +595,936 @@ func (s *session) Close() error {
...
+					start = time.Unix(0, b.Start)
+					size  = int64(0)
+				)
+				if b.Size != nil {
+					size = *b.Size
+				} else {
+					s.log.
+						WithFields(
+						xlog.NewLogField("id", elem.ID),
+						xlog.NewLogField("start", start),
```
xichen2020: Can probably log `b.Err` as well

client/session.go
```diff
@@ -563,3 +595,945 @@ func (s *session) Close() error {
...
+		enqueueCh.enqueue(received.results)
+	}
+}
+
+func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
+	perPeerBlocksMetadata []*blocksMetadata,
+	peerQueues peerBlocksQueues,
+	pooledCurrStart, pooledCurrEligible []*blocksMetadata,
+	pooledBlocksMetadataQueues []blocksMetadataQueue,
+) error {
```
xichen2020: Seems we always return nil in this method, so maybe just not return anything?

client/session.go
```diff
@@ -563,3 +595,945 @@ func (s *session) Close() error {
...
+	for {
+		// Find the earliest start time
+		var earliestStart time.Time
+		for _, blocksMetadata := range perPeerBlocksMetadata {
+			if len(blocksMetadata.unselectedBlocks()) == 0 {
+				// No unselected blocks
+				continue
+			}
+			unselected := blocksMetadata.unselectedBlocks()
+			if earliestStart.Equal(time.Time{}) ||
```
xichen2020: nit: `if earliestStart.IsZero() || ...`

xichen2020: Take it or leave it, but this can be merged with the next block so it's more terse and we don't have to iterate over the metadata twice:
```
currStart = currStart[:0]
for _, blocksMetadata := range perPeerBlocksMetadata {
  if len(blocksMetadata.unselectedBlocks()) == 0 {
    // No unselected blocks
    continue
  }
  unselected := blocksMetadata.unselectedBlocks()
  if earliestStart.IsZero() || unselected[0].start.Before(earliestStart) {
    earliestStart = unselected[0].start
    currStart = currStart[:0]
    currStart = append(currStart, blocksMetadata)
  } else if unselected[0].start == earliestStart {
    currStart = append(currStart, blocksMetadata)
  }
}
```

client/session.go
```diff
@@ -563,3 +595,945 @@ func (s *session) Close() error {
...
+func (arr peerBlocksMetadata) Less(i, j int) bool {
+	return strings.Compare(arr[i].peer.Host().ID(), arr[j].peer.Host().ID()) < 0
+}
+
+type blocksMetadataQueue struct {
+	blocksMetadata *blocksMetadata
+	queue          *peerBlocksQueue
+}
+
+type blocksMetadatasQueuesByOutstanding []blocksMetadataQueue
```
xichen2020: `blocksMetadatasQueuesByOutstandingAscending` perhaps?

client/session.go
```diff
@@ -563,3 +595,945 @@ func (s *session) Close() error {
...
+		// Order by least outstanding blocks being fetched
+		blocksMetadataQueues = blocksMetadataQueues[:0]
+		for i := range currEligible {
+			insert := blocksMetadataQueue{
+				blocksMetadata: currEligible[i],
+				queue:          peerQueues.findQueue(currEligible[i].peer),
+			}
+			blocksMetadataQueues = append(blocksMetadataQueues, insert)
+		}
+		sort.Stable(blocksMetadatasQueuesByOutstanding(blocksMetadataQueues))
```
xichen2020: Just curious, why stable sort here?

client/session.go
```diff
@@ -563,3 +595,945 @@ func (s *session) Close() error {
...
+
+		// Select the best peer
+		bestPeerBlocksQueue := blocksMetadataQueues[0].queue
+		blocksMetadataQueues = blocksMetadataQueues[:0]
```
xichen2020: nit: don't think this is needed since we already cleared it in line 1020

client/session.go

client/session.go
robskillington: There is one error case, see `len(currStart) == 0`

client/session.go
robskillington: Sure, might do `blocksMetadatasQueuesByOutstandingAsc` instead

client/session.go
robskillington: For predictable test assertions, also the first ordered queue if they are all equal will get it every time which makes for more predictable outcomes when observing behavior.

client/session.go
xichen2020: hm..I don't get it, in that case it breaks out of the loop and returns `nil` as well. If we always return nil, why not make it not return anything.

client/session.go
```diff
@@ -563,3 +595,943 @@ func (s *session) Close() error {
...
+		// Remove the block from all other peers and increment index for selected peer
+		for i := range currStart {
+			peer := currStart[i].peer
+			if peer == bestPeerBlocksQueue.peer {
+				// Select this block
+				idx := currStart[i].idx
+				currStart[i].idx = idx + 1
+
+				// Set the reattempt metadata
+				time.Sleep(time.Millisecond)
```
xichen2020: Why do we need to sleep here?

client/session.go
```diff
@@ -563,3 +595,943 @@ func (s *session) Close() error {
...
+		received, ok := metadata[m.id]
+		if !ok {
+			received = &receivedBlocks{
+				results: make([]*blocksMetadata, 0, peersLen),
+			}
+			metadata[m.id] = received
+		}
+		if received.submitted {
+			// Already submitted to enqueue channel
+			continue
```
xichen2020: When will we get here? Do we want to log it?

client/session.go
robskillington: Oh ok, yeah there used to be an error condition, this is now just the break condition.  Will remove the error return type.

client/session.go
robskillington: Yeah, can remove.

client/session.go
robskillington: Weird, think this might've been for debugging.  Will remove.

client/session.go
robskillington: This will happen say if the pagination has a bug in it and your read the same ID twice for some reason or potentially if you start reading metadata off a new host during the streaming process (currently not supported or possible).

client/session.go
```diff
@@ -563,3 +595,943 @@ func (s *session) Close() error {
...
+	batch []*blocksMetadata,
+	blocksResult *blocksResult,
+	enqueueCh *enqueueChannel,
+) {
+	// Prepare request
+	var (
+		req    = rpc.NewFetchBlocksRequest()
+		result *rpc.FetchBlocksResult_
+	)
+	for i := range batch {
```
xichen2020: `req.Elements := make([]*rpc.FetchBlocksParam, len(batch))` perhaps?

client/session.go
```diff
@@ -563,3 +595,943 @@ func (s *session) Close() error {
...
+func (s *session) streamBlocksBatchFromPeer(
+	retrier xretry.Retrier,
+	peer hostQueue,
+	batch []*blocksMetadata,
+	blocksResult *blocksResult,
+	enqueueCh *enqueueChannel,
+) {
+	// Prepare request
+	var (
+		req    = rpc.NewFetchBlocksRequest()
```
xichen2020: The `FetchBlocks` API now assumes the batch request is for a single shard as we discussed, so we'll need to set the shard ID here.

client/session.go
```diff
@@ -563,3 +595,943 @@ func (s *session) Close() error {
...
+		iter.Reset(readers)
+
+		encoder := r.encoderPool.Get()
+		defer encoder.Close()
+
+		for iter.Next() {
+			dp, unit, annotation := iter.Current()
+			encoder.Encode(dp, unit, annotation)
+		}
+		if err := iter.Err(); err != nil {
```
xichen2020: Do we need to close the iterator?

client/session.go
```diff
@@ -563,3 +595,943 @@ func (s *session) Close() error {
...
+	}
+}
+
+func (c *enqueueChannel) enqueue(peersMetadata []*blocksMetadata) {
+	atomic.AddUint64(&c.enqueued, 1)
+	c.peersMetadataCh <- peersMetadata
+}
+
+func (c *enqueueChannel) enqueueDelayed() func([]*blocksMetadata) {
+	atomic.AddUint64(&c.enqueued, 1)
```
xichen2020: Is there a reason why this is not done inside the function?

client/session.go
xichen2020: Yeah we should log it so we know something weird is happening

client/session.go

client/session.go
robskillington: Good catch, added.

client/session.go
robskillington: Probably should yeah, will do.

client/session.go
robskillington: This is a tricky one to explain.  Essentially before the call to `streamBlocksBatchFromPeer` finishes (which will trigger a `enqueueCh.trackProcessed(1)`) we need to ensure we have tracked another outstanding enqueued piece of work for the retry or else the loop that's checking for no more work could potentially think there's no more work left and close the enqueue channel.

client/session.go
xichen2020: Sounds good

client/session.go
robskillington: Done.

xichen2020:
LGTM

client/session.go
```diff
@@ -1146,7 +1143,7 @@ func (s *session) streamBlocksBatchFromPeer(
 					xlog.NewLogField("actualStart", block.Start),
 					xlog.NewLogField("indexID", i),
 					xlog.NewLogField("indexBlock", j),
-				).Errorf("stream blocks response from peer %v returned mismatched ID", peer.Host().String())
+				).Errorf("stream blocks response from peer %s returned mismatched block start", peer.Host().String())
```
martin-mao: For this error msg and the ones above, would it be useful for debugging to add more details? Ie, here insert the expected block start and actual block start so that we know it's always off by X mins etc?

martin-mao:
Minor comment on some error logging, but LGTM

PR #55, opened 2016-08-23
Cleanup logic to clean up expired data files and commit logs

xichen2020:

### DIFF
--- a/integration/disk_cleanup_test.go
+++ b/integration/disk_cleanup_test.go
@@ -1,0 +1,131 @@
+// +build integration
+
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package integration
+
+import (
+	"errors"
+	"os"
+	"testing"
+	"time"
+
+	"github.com/m3db/m3db/persist/fs"
+	"github.com/m3db/m3db/storage"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	errDataCleanupTimedOut = errors.New("cleaning up data files took too long")
+)
+
+func createWriter(storageOpts storage.Options) fs.FileSetWriter {
+	fsOpts := storageOpts.GetCommitLogOptions().GetFilesystemOptions()
+	blockSize := storageOpts.GetRetentionOptions().GetBlockSize()
+	filePathPrefix := fsOpts.GetFilePathPrefix()
+	writerBufferSize := fsOpts.GetWriterBufferSize()
+	newFileMode := fsOpts.GetNewFileMode()
+	newDirectoryMode := fsOpts.GetNewDirectoryMode()
+	return fs.NewWriter(blockSize, filePathPrefix, writerBufferSize, newFileMode, newDirectoryMode)
+}
+
+func createFilesetFiles(t *testing.T, storageOpts storage.Options, shard uint32, fileTimes []time.Time) {
+	writer := createWriter(storageOpts)
+	for _, start := range fileTimes {
+		require.NoError(t, writer.Open(shard, start))
+		require.NoError(t, writer.Close())
+	}
+}
+
+func createCommitLogs(t *testing.T, filePathPrefix string, fileTimes []time.Time) {
+	for _, start := range fileTimes {
+		commitLogFile, _ := fs.NextCommitLogsFile(filePathPrefix, start)
+		_, err := os.Create(commitLogFile)
+		require.NoError(t, err)
+	}
+}
+
+func waitUntilDataCleanedUp(filePathPrefix string, shard uint32, toDelete time.Time, timeout time.Duration) error {
+	dataCleanedUp := func() bool {
+		if fs.FilesetFileExistsAt(filePathPrefix, shard, toDelete) {
+			return false
+		}
+		_, index := fs.NextCommitLogsFile(filePathPrefix, toDelete)
+		if index != 0 {
+			return false
+		}
+		return true
+	}
+	if waitUntil(dataCleanedUp, timeout) {
+		return nil
+	}
+	return errDataCleanupTimedOut
+}
+
+func TestDiskCleanup(t *testing.T) {
+	// Test setup
+	testSetup, err := newTestSetup(newTestOptions())
+	require.NoError(t, err)
+	defer testSetup.close()
+
+	testSetup.storageOpts =
+		testSetup.storageOpts.
+			RetentionOptions(testSetup.storageOpts.GetRetentionOptions().
+				BufferDrain(3 * time.Second).
+				RetentionPeriod(6 * time.Hour))
+
+	blockSize := testSetup.storageOpts.GetRetentionOptions().GetBlockSize()
+	filePathPrefix := testSetup.storageOpts.GetCommitLogOptions().GetFilesystemOptions().GetFilePathPrefix()
+	retentionPeriod := testSetup.storageOpts.GetRetentionOptions().GetRetentionPeriod()
+
+	// Start the server
+	log := testSetup.storageOpts.GetInstrumentOptions().GetLogger()
+	log.Debug("disk cleanup test")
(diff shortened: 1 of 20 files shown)


integration/disk_cleanup_test.go
```diff
@@ -0,0 +1,131 @@
...
+	// Now create some fileset files and commit logs
+	shard := uint32(0)
+	numTimes := 10
+	fileTimes := make([]time.Time, numTimes)
+	now := testSetup.getNowFn()
+	for i := 0; i < numTimes; i++ {
+		fileTimes[i] = now.Add(time.Duration(i) * blockSize)
+	}
+	createFilesetFiles(t, testSetup.storageOpts, shard, fileTimes)
+	createCommitLogs(t, filePathPrefix, fileTimes)
```
robskillington: Are these going to collide with the actual server's commit log files?

persist/fs/files.go
```diff
@@ -252,8 +334,8 @@ func CommitLogsDirPath(prefix string) string {
 	return path.Join(prefix, commitLogsDirName)
 }
 
-// FileExistsAt determines whether a data file exists for the given shard and block start time.
-func FileExistsAt(prefix string, shard uint32, blockStart time.Time) bool {
+// FilesetFileExistsAt determines whether a data file exists for the given shard and block start time.
+func FilesetFileExistsAt(prefix string, shard uint32, blockStart time.Time) bool {
```
robskillington: Perhaps just `FilesetExistsAt(...)` is a better name?

storage/cleanup.go
```diff
@@ -0,0 +1,159 @@
...
+	shards := mgr.database.getOwnedShards()
+	for _, shard := range shards {
+		shardID := shard.ShardNum()
+		expired, err := mgr.filesetFilesBeforeFn(mgr.filePathPrefix, shardID, earliestToRetain)
+		if err != nil {
+			detailedErr := fmt.Errorf("encountered errors when getting fileset files for prefix %s shard %d: %v", mgr.filePathPrefix, shardID, err)
+			multiErr = multiErr.Add(detailedErr)
+		}
+		if err := mgr.deleteFilesFn(expired); err != nil {
+			multiErr = multiErr.Add(err)
```
martin-mao: Wrap error with more info like on line 94 or log?

storage/cleanup.go
```diff
@@ -0,0 +1,159 @@
...
+	earliest, latest := mgr.commitLogTimeRange(t)
+
+	// TODO(xichen): preallocate the slice here
+	var commitLogTimes []time.Time
+	for commitLogTime := latest; !commitLogTime.Before(earliest); commitLogTime = commitLogTime.Add(-mgr.blockSize) {
+		hasFlushedAll := true
+		leftBlockStart := commitLogTime.Add(-mgr.blockSize)
+		rightBlockStart := commitLogTime.Add(mgr.blockSize)
+		for blockStart := leftBlockStart; !blockStart.After(rightBlockStart); blockStart = blockStart.Add(mgr.blockSize) {
+			if !mgr.fm.HasFlushed(blockStart) {
```
martin-mao: Is it possible to have a corrupt commit log where HasFlushed always returns false in which case we never clean it up?

integration/disk_cleanup_test.go
xichen2020: it's using NextCommitLog API to find the next commit log file to write so it'll just use next index if the server has written some commit logs.

persist/fs/files.go
xichen2020: Sounds good

storage/cleanup.go
xichen2020: It'll eventually get cleaned up once the data expires, i.e., when its timestamp is earlier than now - retentionPeriod.

storage/flush.go
```diff
@@ -21,157 +21,132 @@
...
-			fm.flushAttempted[flushTime] = flushState
-			fm.Unlock()
+	ctx := mgr.opts.GetContextPool().Get()
+	defer ctx.Close()
+
+	multiErr := xerrors.NewMultiError()
+	for _, flushTime := range timesToFlush {
+		mgr.Lock()
+		if !mgr.needsFlush(flushTime) {
+			continue
```
martin-mao: Need to call mgt.Unlock() before continue?

storage/flush.go
```diff
@@ -21,157 +21,132 @@
...
-
-		fm.Lock()
-		fm.fs = flushNotStarted
-		fm.Unlock()
+		flushState := mgr.flushStates[flushTime]
+		flushState.Status = fileOpInProgress
+		mgr.flushStates[flushTime] = flushState
+		mgr.Unlock()
+
+		flushErr := mgr.flushWithTime(ctx, flushTime)
```
martin-mao: Why do we unlock here to just flush? Can we make flushWithTime not lock and ensure all callers have the lock before calling?

storage/flush.go
xichen2020: Yeah good catch

storage/flush.go
xichen2020: It's because `flushWithTime` could take a long time to finish while writing the in-memory data out to disk so we don't want to hold the lock while data is being flushed.

storage/flush.go
```diff
@@ -21,157 +21,132 @@
...
-	defer fm.Unlock()
-	fm.flushTimes = fm.flushTimes[:0]
-	for flushTime := firstBlockStart; !flushTime.Before(earliestTime); flushTime = flushTime.Add(-blockSize) {
-		if flushState, exists := fm.flushAttempted[flushTime]; !exists || (flushState.Status == flushFailed && flushState.NumFailures < maxFlushRetries) {
-			flushState.Status = flushInProgress
-			fm.flushTimes = append(fm.flushTimes, flushTime)
-			fm.flushAttempted[flushTime] = flushState
-		}
+// needsFlush returns true if we need to flush data for a given time.
+func (mgr *flushManager) needsFlush(t time.Time) bool {
```
martin-mao: Maybe leave in the comment that this needs to have the mgr read lock before calling needsFlush?

integration/disk_cleanup_test.go
robskillington: Sounds good.

storage/flush.go
xichen2020: Yeah I'll rename it to needsFlushWithLock

storage/bootstrap.go
```diff
@@ -89,18 +91,6 @@ func (bsm *bootstrapManager) cutoverTime(writeStart time.Time) time.Time {
...
-	// NB(xichen): when bootstrap process is finished, we should have bootstrapped
-	// everything until writeStart + bufferFuture, which means the current time is
-	// at least writeStart + bufferFuture + bufferPast.
-	rops := bsm.opts.GetRetentionOptions()
-	bufferPast := rops.GetBufferPast()
-	bufferFuture := rops.GetBufferFuture()
-	return writeStart.Add(bufferFuture).Add(bufferPast)
-}
-
 // NB(xichen): Bootstrap must be called after the server starts accepting writes.
```
robskillington: Maybe we should enforce this btw?  Thoughts?  We can check if the db is open and if it is already return an error (I know that's not part of this change but just reading this comment makes me sad.

storage/cleanup.go
xichen2020: The error returned from `fs.DeleteFiles` already has all the information including the files that caused the error and the actual error.

storage/bootstrap.go
xichen2020: Yeah so the issue here is that it needs to be called after server starts accepting writes (not after the database is open) so we don't miss writes. 

Maybe we can have the server push a notification to a channel after it starts up and have `Bootstrap` listen to that channel before it starts bootstrapping?

storage/cleanup.go
```diff
@@ -0,0 +1,159 @@
...
+		multiErr = multiErr.Add(detailedErr)
+	}
+	return multiErr.FinalError()
+}
+
+func (mgr *cleanupManager) cleanupFilesetFiles(earliestToRetain time.Time) error {
+	multiErr := xerrors.NewMultiError()
+	shards := mgr.database.getOwnedShards()
+	for _, shard := range shards {
+		shardID := shard.ShardNum()
```
robskillington: Let's use this opportunity to rename `ShardNum()` to `ID()` to match `m3cluster`?

storage/cleanup.go
```diff
@@ -0,0 +1,159 @@
...
+		commitLogsDir:           commitLogsDir,
+		fm:                      fm,
+		filesetFilesBeforeFn:    fs.FilesetFilesBefore,
+		commitLogFilesBeforeFn:  fs.CommitLogFilesBefore,
+		commitLogFilesForTimeFn: fs.CommitLogFilesForTime,
+		deleteFilesFn:           fs.DeleteFiles,
+	}
+}
+
+func (mgr *cleanupManager) Cleanup(t time.Time) error {
```
robskillington: Perhaps name `mgr` just `m` to try to keep to single letter prefixes?

storage/cleanup.go
```diff
@@ -0,0 +1,159 @@
...
+	return multiErr.FinalError()
+}
+
+// NB(xichen): since each commit log contains data needed for bootstrapping not only
+// its own block size period but also its left and right block neighbors due to past
+// writes and future writes, we need to shift flush time range by block size as the
+// time range for commit log files we need to check.
+func (mgr *cleanupManager) commitLogTimeRange(t time.Time) (time.Time, time.Time) {
+	flushStart, flushEnd := mgr.fm.FlushTimeStart(t), mgr.fm.FlushTimeEnd(t)
+	return flushStart.Add(-mgr.blockSize), flushEnd.Add(-mgr.blockSize)
```
robskillington: If we need to do right neighbour shouldn't the second return arg be `flushEnd.Add(m.blockSize)`?

storage/bootstrap.go
robskillington: Yeah let's punt on this as discussed.

storage/cleanup.go
xichen2020: yeah so this is because `flushEnd` is the latest time that can be flushed, so the latest commit log that can be deleted is `flushEnd - blockSize`.

storage/cleanup.go
xichen2020: yeah can do

storage/cleanup.go
xichen2020: sure

martin-mao:
LGTM

robskillington:
LGTM

PR #57, opened 2016-08-24
Add dynamic topology

cw9:


client/session.go
```diff
@@ -153,19 +153,28 @@ func (s *session) Open() error {
...
-			if err := s.setTopologyMap(value); err != nil {
-				log.Errorf("could not update topology map: %v", err)
+		for _ = range s.topoWatch.C() {
+			m, ok := s.topoWatch.Get().(topology.Map)
+			if !ok {
+				log.Errorf("Received invalid update.")
+				continue
+			}
+			if err = s.setTopologyMap(m); err != nil {
+				log.Errorf("Could not update topology map: %v", err)
```
robskillington: Hey for consistency could you make both these log statements start with lower case letters and remove periods at the end?

client/session.go
```diff
@@ -153,19 +153,28 @@ func (s *session) Open() error {
...
 
 	go func() {
 		log := s.opts.GetInstrumentOptions().GetLogger()
-		for value := range s.topoMapCh {
-			if err := s.setTopologyMap(value); err != nil {
-				log.Errorf("could not update topology map: %v", err)
+		for _ = range s.topoWatch.C() {
+			m, ok := s.topoWatch.Get().(topology.Map)
+			if !ok {
+				log.Errorf("Received invalid update.")
```
robskillington: Perhaps more descriptive would be "received invalid update for topology map"?

topology/dynamic.go
```diff
@@ -0,0 +1,208 @@
...
+type dynamicTopology struct {
+	sync.Mutex
+
+	s      xwatch.Source
+	w      xwatch.Watch
+	closed bool
+	logger xlog.Logger
+}
+
+func newDynamicTopology(options Options) (Topology, error) {
```
robskillington: Might be worth moving dynamic topology and static topology into their own packages `static` and `dynamic` under `topology` package as it's kind of weird they share the same options because depending which one you're using you may or may not have to set some properties.  Thoughts?

topology/dynamic.go
```diff
@@ -0,0 +1,208 @@
...
+	w       xwatch.Watch
+	retrier xretry.Retrier
+	closed  bool
+	// NB(chaowang) the watch may be initiated and not get any updates on its channel any time soon,
+	// so we should try whatever value the watch has as first poll to avoid source init time out.
+	// If the value the watch has is invalid, the source is still not initiated
+	initPoll bool
+}
+
+func newServiceTopologyInput(w xwatch.Watch, opt xretry.Options) xwatch.SourceInput {
```
robskillington: Perhaps `opts` instead of `opt` is better here?

topology/dynamic.go
```diff
@@ -0,0 +1,208 @@
...
+				s.initPoll = false
+			}
+
+			if m, err = getMap(s.w.Get()); err != nil {
+				return err
+			}
+			return nil
+		}); attemptErr != nil {
+		return nil, attemptErr
+	}
```
robskillington: This is kind of stylistic, but to stay consistent we usually in prefer inlining the function definition, i.e.:

```go
if finalErr := s.retrier.Attempt(func() error {
  // ...
}); finalErr != nil {
  return nil, finalErr
}
```

topology/dynamic.go
```diff
@@ -0,0 +1,208 @@
...
+					return xwatch.ErrSourceClosed
+				}
+			} else {
+				s.initPoll = false
+			}
+
+			if m, err = getMap(s.w.Get()); err != nil {
+				return err
+			}
+			return nil
```
robskillington: You can avoid the `if` here and just:
```go
m, err = newStaticMapFromServiceInstances(s.w.Get())
return err
```

topology/dynamic.go
```diff
@@ -0,0 +1,208 @@
...
+		}); attemptErr != nil {
+		return nil, attemptErr
+	}
+
+	return m, nil
+}
+
+func getMap(data interface{}) (Map, error) {
```
robskillington: Better name perhaps would be `newStaticMapFromServiceInstances`

topology/dynamic.go
```diff
@@ -0,0 +1,208 @@
...
+			var err error
+			if !s.initPoll {
+				// for normal attempts (from 2nd attempt on), block on the notification channel
+				if _, ok := <-s.w.C(); !ok {
+					s.closed = true
+					return xwatch.ErrSourceClosed
+				}
+			} else {
+				s.initPoll = false
+			}
```
robskillington: Hmm to make it slightly nicer you could:
```go
type serviceTopologyInput struct {
  // ....
  poll int
  // ....
}

func (s *serviceTopologyInput) Poll() (interface{}, error) {
  var m Map
  if finalErr := s.retrier.Attempt(func() error {
    if s.poll > 0 {
      if _, ok := <-s.w.C(); !ok {
        s.closed = true
        return xwatch.ErrSourceClosed
      }
    }
    var err error
    s.poll++
    m, err = newStaticMapFromServiceInstances(s.w.Get())
    return err
  }); finalErr != nil {
    return nil, err
  }
  return m, nil
}

```

topology/dynamic.go
```diff
@@ -0,0 +1,208 @@
...
+		hostShardSets[i] = hs
+	}
+
+	return NewMapOptions().
+		Replicas(r).
+		ShardSet(allShardSet).
+		HostShardSets(hostShardSets), nil
+}
+
+func getUniqueShardsAndReplicas(instances []services.ServiceInstance) ([]uint32, int) {
```
robskillington: Perhaps just call `uniqueShardsAndReplicas(...)`

topology/dynamic.go
```diff
@@ -0,0 +1,208 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if err := to.Validate(); err != nil {
+		return nil, err
+	}
+	return newStaticMap(to), nil
+}
+
+func buildTopologyOptionsFromServiceInstances(instances []services.ServiceInstance) (MapOptions, error) {
```
robskillington: Perhaps just `mapOptionsFromServiceInstances`?

topology/dynamic.go
```diff
@@ -0,0 +1,208 @@
...
+	allShards, r := getUniqueShardsAndReplicas(instances)
+
+	fn := sharding.DefaultHashFn(uint32(len(allShards)))
```
robskillington: I don't think we should specify the hashing function in concrete here.  We should probably pass this all the way in the top to new dynamic topology constructor.

topology/host.go
```diff
@@ -63,6 +64,23 @@ func NewHostShardSet(host Host, shardSet sharding.ShardSet) HostShardSet {
...
+	shardSet, err := sharding.NewShardSet(shards, fn)
+	if err != nil {
+		return nil, err
+	}
+	return &hostShardSet{host: NewHost(si.ID(), getAddress(si)), shardSet: shardSet}, nil
+}
+
+// TODO: verify this
+func getAddress(si services.ServiceInstance) string {
+	return si.ID() + ":" + si.Endpoint()
```
robskillington: So `Endpoint()` should be the Address.  `ID` in the future will just be a UUID.

topology/host.go
```diff
@@ -63,6 +64,23 @@ func NewHostShardSet(host Host, shardSet sharding.ShardSet) HostShardSet {
...
+func newHostShardSetFromServiceInstance(si services.ServiceInstance, fn sharding.HashFn) (HostShardSet, error) {
+	shards := make([]uint32, 0, si.Shards().ShardsLen())
+	for _, s := range si.Shards().Shards() {
+		shards = append(shards, s.ID())
+	}
+	shardSet, err := sharding.NewShardSet(shards, fn)
+	if err != nil {
+		return nil, err
+	}
+	return &hostShardSet{host: NewHost(si.ID(), getAddress(si)), shardSet: shardSet}, nil
```
robskillington: Just make this `NewHost(si.ID(), si.Endpoint())`

topology/init.go
```diff
@@ -0,0 +1,53 @@
...
+func NewInitializer(opts Options) Initializer {
+	return topologyInitializer{opts}
+}
+
+func (c topologyInitializer) Init() (Topology, error) {
+	t, err := newDynamicTopology(c.opts)
+	if err == nil {
+		return t, nil
+	}
+	// Failed to create dynamic topology, try to fail back to static topology
```
robskillington: Hm doesn't seem like a very explicit strategy.  I'd be in prefer of creating both a static initializer and a dynamic initializer and each resides in its own package.

robskillington:
Done with first round.

topology/dynamic.go
cw9: yeah, make sense, will do

topology/dynamic.go
cw9: yeah, sounds good, will change

client/session_test.go
```diff
@@ -26,11 +26,10 @@ import (
 	"sync/atomic"
 	"testing"
 
+	"github.com/golang/mock/gomock"
```
xichen2020: Reorder this?

client/session.go
```diff
@@ -153,18 +153,27 @@ func (s *session) Open() error {
 	s.seriesIteratorsPool.Init()
 	s.Unlock()
 
-	currTopologyMap := s.topo.GetAndSubscribe(s.topoMapCh)
+	currTopologyMap, topologyWatch, err := s.topo.GetAndSubscribe()
+	if err != nil {
+		return err
```
robskillington: Need to ensure we return to not open state unfortunately here:
```go
s.Lock()
s.state = stateNotOpen
s.Unlock()
return err
```

sharding/shardset.go
```diff
@@ -90,3 +92,10 @@ func validateShards(shards []uint32) error {
 	}
 	return nil
 }
+
+// DefaultHashGen generates a HashFn based on murmur32
+func DefaultHashGen(l uint32) HashFn {
```
robskillington: How about just calling this `DefaultHashShardRangeFn`?

sharding/types.go
```diff
@@ -20,6 +20,10 @@
 
 package sharding
 
+
+// HashGen generates HashFn
+type HashGen func(l uint32) HashFn
```
robskillington: HashGen sounds a little ambiguous.  Perhaps call this `NewHashShardRangeFn` e.g.:
```go
type NewHashShardRangeFn(length uint32) HashFn
```

topology/dynamic.go
```diff
@@ -0,0 +1,222 @@
...
+	return NewStaticOptions().
+		Replicas(r).
+		ShardSet(allShardSet).
+		HostShardSets(hostShardSets), nil
+}
+
+func uniqueShardsAndReplicas(instances []services.ServiceInstance) ([]uint32, int) {
+	m := make(map[uint32]int)
+	for _, i := range instances {
+		for _, s := range i.Shards().Shards() {
```
robskillington: Should probably check `i.Shards() != nil`?  This should never happen in production but it's possible I would assume.  If so would prefer to bubble an error all the way up rather than cause a `nil` panic.

topology/dynamic.go
```diff
@@ -0,0 +1,222 @@
...
+		}
+	}
+	r := math.MaxUint32
+	s := make([]uint32, 0, len(m))
+
+	// NB(chaowang): We allow different replicas for a shard in case cluster is changing capacity
+	for k, v := range m {
+		s = append(s, k)
+		if r > v {
+			r = v
```
robskillington: Yeah so this might actually be problematic in the future.  We really want "replicas" to actually come from config ideally.  

Otherwise take this scenario:
1) Placement algorithm introduces a bug.
2) Run again and one shard ends up with only two hosts assigned.
3) Now we disseminate changes and suddenly entire cluster sees "replicas 2" which will change the condition of a write succeeding, etc.

robskillington: Ok so yeah I believe to `Services` interface in `m3cluster` we're going to have to add some kind of queryable endpoint to query for this type of metadata:

```go
type Services interface {
  Advertise(ad Advertisement) error
  Unadvertise(service, id string) error
  QueryInstances(service string, opts QueryOptions) ([]ServiceInstance, error)
  WatchInstances(service string, opts QueryOptions) (xwatch.Watch, error)

  // QueryShardingInfo queries information regarding a service's sharding
  QueryShardingInfo(service string) (shard.ShardingInfo, error)
}
```

Then in the `shard` package:

```go
type ShardingInfo interface {
  Replicas() int
  ShardRange() uint32
}
```

That way you can verify the actual data returned here when setting the assigned shards that all shards meet the baseline replicas that's supposed to be set.  

Also it means when you're creating the `HashFn` you don't have to just get the length of the shards and can do it more explicitly:

```go
func newMapOptionsFromServiceInstances(..., shardingInfo m3clustershard.ShardingInfo, newHashFn sharding.NewHashFnFromShardRange, ...) (StaticOptions, error) {
  // ....
  fn := newHashFn(shardingInfo.ShardRange())
  // ...
}
```

topology/host.go
```diff
@@ -63,6 +64,18 @@ func NewHostShardSet(host Host, shardSet sharding.ShardSet) HostShardSet {
 	return &hostShardSet{host, shardSet}
 }
 
+func newHostShardSetFromServiceInstance(si services.ServiceInstance, fn sharding.HashFn) (HostShardSet, error) {
+	shards := make([]uint32, 0, si.Shards().ShardsLen())
```
robskillington: Should nil check for `if si.Shards() == nil { return errInstanceHasNoShards }`

topology/dynamic.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if err := to.Validate(); err != nil {
+		return nil, err
+	}
+	return newStaticMap(to), nil
+}
+
+func newMapOptionsFromServiceInstances(instances []services.ServiceInstance, hashGen sharding.HashGen) (StaticOptions, error) {
```
robskillington: Perhaps rename this `newStaticOptionsFromServiceInstances`?

topology/host.go
```diff
@@ -63,6 +64,18 @@ func NewHostShardSet(host Host, shardSet sharding.ShardSet) HostShardSet {
...
+	for _, s := range si.Shards().Shards() {
+		shards = append(shards, s.ID())
+	}
+	shardSet, err := sharding.NewShardSet(shards, fn)
+	if err != nil {
+		return nil, err
+	}
+	return &hostShardSet{host: NewHost(si.ID(), si.Endpoint()), shardSet: shardSet}, nil
```
robskillington: Any reason not to use the constructor here?  `NewHostShardSet(NewHost(si.ID(), si.Endpoint()), shardSet), nil`?

sharding/shardset.go
robskillington: I take it back, HashGen is probably fine.

robskillington:
Done with second round.

client/session_test.go
cw9: will do

client/session.go
cw9: I see, maybe I can do the s.topo.GetAndSubscribe() and s.setTopologyMap(currTopologyMap) in the beginning of Open()
before the line s.state = stateOpen?

topology/dynamic.go
cw9: sure, will do

client/session.go
robskillington: That would mean two consumers could concurrently call open which we're trying to prevent here unfortunately :(

topology/dynamic.go
cw9: I don't know about this actually, I feel the validation of a placement should be done when a placement is being generated, if a newly generated placement is wrong, you don't want to save it in the first place, the output from the Services interface should already be well validated and just simply be applied in usage.

topology/host.go
cw9: will do

topology/host.go
cw9: will do

topology/dynamic.go
robskillington: Agreed it should be validated and every human should avoid saving a placement that looks wrong.  

Fact is this _can_ happen, then it's just a matter of if/when it does happen, do you want one of the following two outcomes:

1. Replicas adjust to "2" and potentially also not all shards are distributed properly
2. You see a `log.Errorf("received invalid update for topology map")` because it failed to meet the hardcoded constraints

topology/dynamic.go
```diff
@@ -0,0 +1,222 @@
...
+	poll int
+}
+
+func newServiceTopologyInput(w xwatch.Watch, h sharding.HashGen, opts xretry.Options) xwatch.SourceInput {
+	return &serviceTopologyInput{w: w, hashGen: h, retrier: xretry.NewRetrier(opts)}
+}
+
+func (s *serviceTopologyInput) Poll() (interface{}, error) {
+	var m Map
+	if attemptErr := s.retrier.Attempt(func() error {
```
robskillington: This should be called `finalErr` btw, each attempt is inside the function.

cw9: will rename

client/session.go
cw9: I moved the GetAndSubscribe() and error check up before setting s.state to open, the s was locked so no two consumers can call it at the same time, no?

client/session.go
```diff
@@ -153,18 +159,23 @@ func (s *session) Open() error {
...
 	s.Unlock()
 
-	currTopologyMap := s.topo.GetAndSubscribe(s.topoMapCh)
 	if err := s.setTopologyMap(currTopologyMap); err != nil {
 		s.Lock()
 		s.state = stateNotOpen
 		s.Unlock()
 		return err
 	}
+	s.topoWatch = topologyWatch
```
robskillington: The `s.topoWatch = topologyWatch` assignment should be in a lock, also if we're moving the `GetAndSubscribe()` into the lock above all the state setting, I would move the the `if err := setTopologyMap(currTopologyMap); err != nil { return err }` directly below it.

topology/dynamic.go
```diff
@@ -0,0 +1,227 @@
...
+	return dt, nil
+}
+
+func (t *dynamicTopology) Get() Map {
+	return t.s.Get().(Map)
+}
+
+func (t *dynamicTopology) GetAndSubscribe() (Map, xwatch.Watch, error) {
+	m, w, err := t.s.Watch()
+	return m.(Map), w, err
```
robskillington: if `err != nil` here will `m.(Map)` cause a panic?

topology/dynamic.go
cw9: Right now we have validation() code in the PlacementService before a new placement is saved/returned rather than human eyeballing and we can improve the validation there maybe. But since we need a manual step in the saving process, I agree this manual part might be error prone if that's what you are worrying about here. 

I feel catching the error here in the client is already kinda too late, the running clients can just discard the update, but any new clients trying to start without a good previous state can only fail with this error. We should avoid bad placement as early as possible, but I guess it's safer to do the error check here, but that would involve a bigger cluster interface change, maybe I should do that in another diff?

topology/dynamic.go
robskillington: seems so: https://play.golang.org/p/psvINjzfqi

cw9: good catch, will fix

client/session.go
cw9: kk, I'll do the original way you proposed, seems cleaner that way

client/session.go
```diff
@@ -151,29 +165,33 @@ func (s *session) Open() error {
...
+			s.setTopologyWithLock(m, queues, replicas, majority)
+			s.Unlock()
 		}
 	}()
 
 	return nil
 }
 
-func (s *session) setTopologyMap(topologyMap topology.Map) error {
+func (s *session) validateMap(topologyMap topology.Map) ([]hostQueue, int, int, error) {
```
robskillington: Better name would be "validateAndConnectUsingMap(...)"

robskillington:
Other than naming of that one function LGTM.  Make sure to pull m3cluster and that Travis is passing tests before we merge?

PR #58, opened 2016-08-24
Fix the predictability of TestFetchBootstrapBlocksAllPeersSucceed

robskillington:
cc @xichen2020 @martin-mao


robskillington:
Essentially the per blocks queues draining themselves in the background caused once in a blue moon for the "foo" and the "baz" blocks fetch to be fetched in separate batches.  This caused the expected mock client call to not match the request arguments that were set as they appeared as two calls instead of one batch call.  This in turn would hang the test from time to time.

xichen2020:
LGTM

PR #59, opened 2016-08-24
Add Unseal() to enable unsealing sealed encoders

xichen2020:


encoding/m3tsz/encoder.go
```diff
@@ -553,6 +553,32 @@ func (enc *encoder) Seal() {
 	enc.os.WriteBytes(tail)
 }
 
+func (enc *encoder) Unseal() error {
+	if enc.closed || enc.writable {
+		// If the encoder is already closed, or the encoder is already writable,
+		// no action is necessary.
+		return nil
```
robskillington: Maybe this should be an error?

xichen2020: sure can do

robskillington:
LGTM

PR #60, opened 2016-08-25
Strip GOPATH from source path in auto-generated mocks

xichen2020:


robskillington:
LGTM

robskillington:
PS we probably need to do this for all our other repos too... or create an `m3build` repo and use as submodule.

xichen2020:
Yeah although I don't think those even have build targets set up to automatically generate mocks so we'll need to add that first =]

robskillington:
m3cluster I think has it?

xichen2020:
Indeed, here is the PR to fix that as well: https://github.com/m3db/m3cluster/pull/13

PR #61, opened 2016-08-26
Unseal blocks streamed from peers and track processed channel items without spawning a goroutine

robskillington:
cc @xichen2020 @martin-mao

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -1264,7 +1264,6 @@ func (r *blocksResult) addBlockFromPeer(id string, block *rpc.Block) error {
 
 		encoder := r.encoderPool.Get()
 		encoder.Reset(start, r.blockAllocSize)
-		defer encoder.Close()
 
 		for iter.Next() {
 			dp, unit, annotation := iter.Current()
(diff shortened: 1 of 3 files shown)


client/session.go
```diff
@@ -1252,6 +1263,7 @@ func (r *blocksResult) addBlockFromPeer(id string, block *rpc.Block) error {
 		defer iter.Close()
 
 		encoder := r.encoderPool.Get()
+		encoder.Reset(start, r.blockAllocSize)
 		defer encoder.Close()
```
xichen2020: Are we sure we want to close the encoder here?

client/session.go
```diff
@@ -1492,6 +1502,15 @@ func (b blocksMetadata) unselectedBlocks() []blockMetadata {
 type blocksMetadatas []*blocksMetadata
 
 func (arr blocksMetadatas) swap(i, j int) { arr[i], arr[j] = arr[j], arr[i] }
+func (arr blocksMetadatas) hasBlocksLen() int {
+	count := 0
+	for i := range arr {
+		if len(arr[i].blocks) > 0 {
```
xichen2020: Can `arr[i]` ever be nil?

client/session.go
robskillington: Yeah probably not lol, will update and also assert not closed by pulling the test encoder out and checking it's state.

client/session.go
robskillington: Should not be, but I'll add this check anyway.

client/session.go
xichen2020: Sounds good

xichen2020:
LGTM

PR #65, opened 2016-08-30
xmetrics -> tally

jra3:
Convert m3db over to use tally Scope instead of m3x

### DIFF
--- a/persist/fs/commitlog/commit_log.go
+++ b/persist/fs/commitlog/commit_log.go
@@ -29,8 +29,8 @@ import (
 	"github.com/m3db/m3db/clock"
 	"github.com/m3db/m3db/ts"
 	"github.com/m3db/m3x/log"
-	"github.com/m3db/m3x/metrics"
 	"github.com/m3db/m3x/time"
+	"github.com/uber-go/tally"
 )
 
 var (
@@ -50,8 +50,10 @@ type commitLog struct {
 	sync.RWMutex
 	opts    Options
 	nowFn   clock.NowFn
-	metrics xmetrics.Scope
-	log     xlog.Logger
+	scope   tally.Scope
+	metrics commitLogMetrics
+
+	log xlog.Logger
 
 	newCommitLogWriterFn newCommitLogWriterFn
 	commitLogFailFn      commitLogFailFn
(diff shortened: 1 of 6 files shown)


persist/fs/commitlog/commit_log.go
```diff
@@ -71,6 +73,14 @@ type commitLog struct {
 	closeErr       chan error
 }
 
+type commitLogMetrics struct {
+	success      tally.Counter
+	errors       tally.Counter
+	open_errors  tally.Counter
+	close_errors tally.Counter
+	flush_errors tally.Counter
```
robskillington: Can we make these camel cased?

martin-mao:
LGTM

persist/fs/commitlog/commit_log.go
```diff
@@ -50,8 +50,10 @@ type commitLog struct {
 	sync.RWMutex
 	opts    Options
 	nowFn   clock.NowFn
-	metrics xmetrics.Scope
-	log     xlog.Logger
+	scope   tally.Scope
+	metrics *commitLogMetrics
```
robskillington: Perhaps just make this a struct value rather than a ref?  Slightly better memory locality.

persist/fs/commitlog/iterator.go
```diff
@@ -37,9 +37,14 @@ var (
 	errIndexDoesNotMatch    = errors.New("commit log file index does not match filename")
 )
 
+type iteratorMetrics struct {
+	reads_errors tally.Counter
```
robskillington: Make this camel cased?

persist/fs/commitlog/iterator.go
```diff
@@ -37,9 +37,14 @@ var (
...
+}
+
 type iterator struct {
 	opts    Options
-	metrics xmetrics.Scope
+	scope   tally.Scope
+	metrics *iteratorMetrics
```
robskillington: Use struct as value instead of ref for slightly better memory locality?

robskillington:
Hey @jra3 do you mind rebasing from master?  There's one more metric emission in `commit_log.go` that needs converting over.

persist/fs/commitlog/iterator.go
xichen2020: camel case?

persist/fs/commitlog/iterator.go
xichen2020: Can we make this a struct instead of a pointer?

persist/fs/commitlog/iterator.go
jra3: old habits die hard

robskillington:
LGTM

xichen2020:
LGTM

PR #66, opened 2016-09-01
Namespacing logic

xichen2020:


client/host_queue.go
```diff
@@ -318,6 +320,29 @@ func (q *queue) asyncFetch(wg *sync.WaitGroup, op *fetchBatchOp) {
...
+		}
+
+		ctx, _ := thrift.NewContext(q.opts.GetTruncateRequestTimeout())
+		if err := client.TruncateNamespace(ctx, &op.request); err != nil {
+			op.completionFn(nil, err)
+		} else {
+			op.completionFn(nil, nil)
+		}
+
+		wg.Done()
```
robskillington: This should probably be `defer wg.Done()` at the top so it is run during the error case too yeah?

client/session.go
```diff
@@ -404,7 +406,7 @@ func (s *session) Fetch(id string, startInclusive, endExclusive time.Time) (enco
 	return iter, nil
 }
 
-func (s *session) FetchAll(ids []string, startInclusive, endExclusive time.Time) (encoding.SeriesIterators, error) {
+func (s *session) FetchAll(idns []idWithNamespace, startInclusive, endExclusive time.Time) (encoding.SeriesIterators, error) {
```
robskillington: This is a public facing API, we can't make one of the types here be a non-exported type =[

Perhaps just a string for namespace as well and then you can only fetch from a specific namespace.

generated/thrift/rpc.thrift
```diff
@@ -77,10 +83,11 @@ struct Block {
 }
 
 struct FetchBlocksMetadataRequest {
-	1: required i32 shard
-	2: required i64 limit
-	3: optional i64 pageToken
-	4: optional bool includeSizes
+    1: required string ns
+	2: required i32 shard
```
robskillington: Whitespace formatting seems out of whack here

generated/thrift/rpc.thrift
```diff
@@ -49,11 +49,17 @@ service Node {
...
+
+struct TruncateNamespaceRequest {
+    1: required string ns
 }
 
 struct FetchBlocksRequest {
-	1: required i32 shard
-	2: required list<FetchBlocksParam> elements
+    1: required string ns
+	2: required i32 shard
```
robskillington: Whitespace formatting seems out of whack here

generated/thrift/rpc.thrift
```diff
@@ -49,11 +49,17 @@ service Node {
 	FetchRawBatchResult fetchRawBatch(1: FetchRawBatchRequest req) throws (1: Error err)
 	FetchBlocksResult fetchBlocks(1: FetchBlocksRequest req) throws (1: Error err)
 	FetchBlocksMetadataResult fetchBlocksMetadata(1: FetchBlocksMetadataRequest req) throws (1: Error err)
+	void truncateNamespace(1: TruncateNamespaceRequest req) throws (1: Error err)
+}
+
+struct TruncateNamespaceRequest {
+    1: required string ns
```
robskillington: Any reason to use `ns` instead of more descriptive `namespace` here?  This will also be our HTTP/JSON API too so might be best to make it `namespace` here and in the other structs.

generated/thrift/rpc.thrift
```diff
@@ -104,8 +111,13 @@ struct HealthResult {
...
 }
 
+struct IDWithNamespace {
+    1: required string ns
+    2: required string id
+}
+
 struct WriteRequest {
-	1: required string id
+    1: required IDWithNamespace idWithNamespace
```
robskillington: Any way we can avoid sending the full namespace string along with each write?  This might up our total traffic out of the machine by up to 20% or so considering the ID string makes up the majority of a write's payload and the namespace string could be as long as 20% of the full ID.

generated/thrift/rpc.thrift
```diff
@@ -140,7 +152,7 @@ struct Datapoint {
 struct FetchRawBatchRequest {
 	1: required i64 rangeStart
 	2: required i64 rangeEnd
-	3: required list<string> ids
+	3: required list<IDWithNamespace> idsWithNamespace
```
robskillington: For both fetch requests I think making them top level makes sense, we really don't need to fetch across namespaces during fetch requests.  It will also reduce the payload by up to 20% too if we do this.

generated/thrift/rpc.thrift
robskillington: We could move this up to the batch level?

client/truncate.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func (t *truncateOp) Size() int {
+	// Truncate is always a single op
+	return 1
+}
+
+func (t *truncateOp) GetCompletionFn() completionFn {
+	return t.completionFn
+}
```
robskillington: nit: add end of line to this file?

client/session.go
xichen2020: Right yeah, I think that's fair given our usecase

generated/thrift/rpc.thrift
xichen2020: `namespace` is a thrift keyword and thrift compiler doesn't like it unfortunatley :(

generated/thrift/rpc.thrift
xichen2020: Sure can do given our use case

client/session.go
```diff
@@ -617,7 +622,54 @@ func (s *session) Close() error {
 	return nil
 }
 
+func (s *session) TruncateNamespace(namespace string, shard uint32) error {
```
robskillington: Can we make this not take ` shard` and to just simply for over all host queues?

generated/proto/persist_fs_commitlog.proto
```diff
@@ -21,5 +21,6 @@ message CommitLog {
 
 message CommitLogMetadata {
 	string id = 1;
-	uint32 shard = 2; 
+	string namespace = 2;
+	uint32 shard = 3;
```
robskillington: This is going to be compatibility breaking, should just remember that as we upgrade things =]

network/server/httpjson/handlers.go
```diff
@@ -62,6 +62,7 @@ func RegisterHandlers(mux *http.ServeMux, service interface{}, opts ServerOption
 		// - methodName(RequestObject) error
 		// - methodName(RequestObject) (ResultObject, error)
 
+		// TODO(xichen): make HTTP server understand truncate method.
```
robskillington: If you add a `TruncateResult` as the return type for the RPC it'll get automatically registered.  Perhaps do that?

network/server/tchannelthrift/cluster/service.go
```diff
@@ -158,7 +158,7 @@ func (s *service) Write(tctx thrift.Context, req *rpc.WriteRequest) error {
 		return tterrors.NewBadRequestError(err)
 	}
 	ts := xtime.FromNormalizedTime(req.Datapoint.Timestamp, d)
-	err = session.Write(req.ID, ts, req.Datapoint.Value, unit, req.Datapoint.Annotation)
```
robskillington: Should probably also add a truncate method to the cluster service that calls truncate from the session.

network/server/tchannelthrift/node/service.go
```diff
@@ -277,3 +278,8 @@ func (s *service) WriteBatch(tctx thrift.Context, req *rpc.WriteBatchRequest) er
 	}
 	return nil
 }
+
+func (s *service) TruncateNamespace(tctx thrift.Context, req *rpc.TruncateNamespaceRequest) error {
+	s.db.TruncateNamespace(req.Ns)
+	return nil
```
robskillington: Yeah we should return something from this method describing something about the result of the operation so we can tell if it did things when we run it.

robskillington: i.e. maybe how many series were ejected from memory, or something.

storage/database.go
```diff
@@ -264,20 +257,38 @@ func (d *db) IsBootstrapped() bool {
 	return d.bsm.IsBootstrapped()
 }
 
-func (d *db) getOwnedShards() []databaseShard {
+func (d *db) TruncateNamespace(namespace string) error {
+	n, err := d.readableNamespace(namespace)
+	if err != nil {
+		return err
+	}
+	n.Truncate()
```
robskillington: Truncate should probably return an error yeah?

storage/database.go
```diff
@@ -170,90 +175,78 @@ func (d *db) Close() error {
...
 	d.RLock()
-	shardID := d.shardSet.Shard(id)
-	shard := d.shards[shardID]
+	n, exists := d.namespaces[namespace]
 	d.RUnlock()
 
-	if shard == nil {
-		return fmt.Errorf("not responsible for shard %d", shardID)
+	if !exists {
+		return fmt.Errorf("not responsible for namespace %s", namespace)
```
robskillington: Perhaps better namespace error would be `fmt.Errorf("no such namespace: %s", namespace)`

storage/database.go
```diff
@@ -264,20 +257,38 @@ func (d *db) IsBootstrapped() bool {
...
 	}
-	shards := d.shardSet.Shards()
-	databaseShards := make([]databaseShard, len(shards))
-	for i, shard := range shards {
-		databaseShards[i] = d.shards[shard]
+	n, exists := d.namespaces[namespace]
+	d.RUnlock()
+
+	if !exists {
+		return nil, fmt.Errorf("not responsible for namespace %s", namespace)
```
robskillington: Perhaps better namespace error would be `fmt.Errorf("no such namespace: %s", namespace)`

storage/database.go
```diff
@@ -264,20 +257,38 @@ func (d *db) IsBootstrapped() bool {
...
+	}
+	return n, nil
+}
+
+func (d *db) getOwnedNamespaces() []databaseNamespace {
+	d.RLock()
+	namespaces := make([]databaseNamespace, 0, len(d.namespaces))
+	for _, n := range d.namespaces {
+		namespaces = append(namespaces, n)
```
robskillington: You could copy here to save lines/simplify:
```go
namespaces := make([]databaseNamespace, len(d.namespaces))
copy(namespaces, d.namespaces)
```

storage/database.go
```diff
@@ -292,32 +303,17 @@ func (d *db) ongoingTick() {
...
-			time.Sleep(splayApart)
-		}
-		wg.Add(1)
-		go func() {
-			// TODO(r): instrument timing of this tick
-			shard.Tick()
-			wg.Done()
-		}()
+	for _, n := range namespaces {
+		// TODO(xichen): sleep during two consecutive ticks
```
robskillington: Hmm how do we fairly splay here?  Wonder if we can keep a count of the IDs in each namespace then fairly divide the splay between them at the top level and tell each namespace what their deadline is.  We could potentially do this in a later diff.

storage/namespace.go
```diff
@@ -0,0 +1,270 @@
...
+	increasingIndex  increasingIndex
+	writeCommitLogFn writeCommitLogFn
+}
+
+func newDatabaseNamespace(
+	metadata namespace.Metadata,
+	shardSet sharding.ShardSet,
+	increasingIndex increasingIndex,
+	writeCommitLogFn writeCommitLogFn,
+	dopts Options,
```
robskillington: Maybe `sopts` for storage options?

storage/namespace.go
```diff
@@ -0,0 +1,270 @@
...
+	return n.name
+}
+
+func (n *dbNamespace) Tick() {
+	shards := n.getOwnedShards()
+	if len(shards) == 0 {
+		return
+	}
+
+	// Tick through the shards sequentially to avoid parallel data flushes
```
robskillington: We should definitely do some splaying here, or else it'll be even worse than it is in production right now lol.

robskillington: Alternative is to actually add some sleeping during batch shard traversal, that actually might be more fair.

robskillington: I'm also fine if we want to test some of this stuff before adding the splay.  We should profile the spikes more carefully and ensure it's the splay that's the extra hot part of the CPU spikes.

storage/namespace.go
```diff
@@ -0,0 +1,270 @@
...
+	pm persist.Manager,
+) error {
+	n.RLock()
+	if n.bs != bootstrapped {
+		n.RUnlock()
+		return errNamespaceNotBootstrapped
+	}
+	n.RUnlock()
+
+	if !n.nopts.GetNeedsFlush() {
```
robskillington: Nice, this is a great knob to have.

storage/namespace.go
```diff
@@ -0,0 +1,270 @@
...
+		return nil, fmt.Errorf("not responsible for shard %d", shardID)
+	}
+	return shard, nil
+}
+
+func (n *dbNamespace) initShards() {
+	shards := n.shardSet.Shards()
+	dbShards := make([]databaseShard, n.shardSet.Max()+1)
+	for _, shard := range shards {
+		dbShards[shard] = newDatabaseShard(shard, n.increasingIndex, n.writeCommitLogFn, n.dopts)
```
robskillington: We should probably pass a no-op function here if `!n.nopts.GetNeedsFlush()` to avoid writes to the commit log for this namespace too.

storage/database.go
xichen2020: Sounds good

storage/namespace.go
xichen2020: There is a separate knob called `WritesToCommitLog` for exactly that purpose

storage/shard.go
```diff
@@ -425,3 +442,17 @@ func (s *dbShard) Flush(ctx context.Context, blockStart time.Time, pm persist.Ma
...
+	multiErr := xerrors.NewMultiError()
+	expired, err := s.filesetBeforeFn(filePathPrefix, namespace, s.ID(), earliestToRetain)
+	if err != nil {
+		detailedErr := fmt.Errorf("encountered errors when getting fileset files for prefix %s namespace %s shard %d: %v", filePathPrefix, namespace, s.ID(), err)
+		multiErr = multiErr.Add(detailedErr)
+	}
+	if err := s.deleteFilesFn(expired); err != nil {
+		multiErr = multiErr.Add(err)
+	}
+	return multiErr.FinalError()
```
robskillington: Is there a reason we moved CleanupFileset to the shard level?  Seems more clean to have it in the namespace?  Also in the future if we ever do consolidate all shard data into single files we'd have to move this back again.

xichen2020: Yeah if we ever do consolidate data, we'll need to move it up, but for now feels like it's more consistent that way since that's also how flush and bootstrap are structured.

storage/shard.go
```diff
@@ -425,3 +442,17 @@ func (s *dbShard) Flush(ctx context.Context, blockStart time.Time, pm persist.Ma
...
 	return multiErr.FinalError()
 }
+
+func (s *dbShard) CleanupFileset(namespace string, earliestToRetain time.Time) error {
+	filePathPrefix := s.opts.GetCommitLogOptions().GetFilesystemOptions().GetFilePathPrefix()
+	multiErr := xerrors.NewMultiError()
+	expired, err := s.filesetBeforeFn(filePathPrefix, namespace, s.ID(), earliestToRetain)
+	if err != nil {
+		detailedErr := fmt.Errorf("encountered errors when getting fileset files for prefix %s namespace %s shard %d: %v", filePathPrefix, namespace, s.ID(), err)
+		multiErr = multiErr.Add(detailedErr)
```
cw9: should probably return here?

xichen2020: Nah we want to clean the ones we got even if we encountered some errors

generated/thrift/rpc.thrift
robskillington: Damn, perhaps `target`?

network/server/tchannelthrift/node/service.go
robskillington: Should just be able to grab a RLock and take a snap of the `len(s.lookup)` and add them across the current owned shards perhaps before calling `initShards()` in namespace?

storage/namespace.go
robskillington: Ah good stuff 👍

storage/shard.go
robskillington: Sure thing.

generated/thrift/rpc.thrift
xichen2020: hm don't like `target`, maybe `nameSpace`

network/server/tchannelthrift/node/service.go
xichen2020: Yeah can do, just saying we don't need to deal with existing shards with the current implementation, but sure

generated/thrift/rpc.thrift
xichen2020: We could, but just an FYI that means `WriteBatchRequest` is no longer simply a list of `WriteRequest` because we end up with something like
```
struct NamedDatapoint {
    1: required string id
    2: required Datapoint datapoint
}

struct WriteRequest {
    1: required string nameSpace
    2: required NamedDatapoint namedDatapoint
}

struct WriteBatchRequest {
    1: required string nameSpace
	2: required list<NamedDatapoint> elements
}
```
so the pooling logic for `WriteRequest` and `WriteBatchRequest` in host_queue will need to change too.

client/host_queue.go
xichen2020: It's only in two places so it's a small cost to pay for a little perf gain

generated/thrift/rpc.thrift
xichen2020: hmm seems fine in my editor

client/truncate.go
xichen2020: sure

client/session.go
xichen2020: sure

storage/database.go
xichen2020: Will leave it till after we've done some amount of perf testing

storage/namespace.go
xichen2020: sure

storage/namespace.go
xichen2020: Will leave it till after we've done some amount of perf testing

storage/database.go
xichen2020: welp, `d.namespaces` is a map, so can't use `copy`

robskillington: Ah I see.  Ignore my ignorance then =]

client/session.go
```diff
@@ -824,9 +821,9 @@ func (s *session) streamBlocksMetadataFromPeer(
 				if b.Size == nil {
 					s.log.
 						WithFields(
-							xlog.NewLogField("id", elem.ID),
-							xlog.NewLogField("start", blocks[i].start),
-						).Warnf("stream blocks metadata requested block size and not returned")
+						xlog.NewLogField("id", elem.ID),
```
robskillington: nit: This indentation looks weird, maybe just make it:
```go
s.log.WithFields(
  xlog.NewLogField("id", elem.ID),
  xlog.NewLogField("start", blocks[i].start),
).Warnf("stream blocks metadata requested block size and not returned")
```

client/host_queue.go
```diff
@@ -169,46 +169,62 @@ func (q *queue) rotateOpsWithLock() []op {
 }
 
 func (q *queue) drain() {
-	wgAll := &sync.WaitGroup{}
+	var (
+		wgAll                       = &sync.WaitGroup{}
+		currWriteOpsByNamespace     = make(map[string][]op)
```
robskillington: Is there any new test for checking that writing two different namespaces will generate two fetch batch requests?  Thinking this might be a good idea to make sure going forward people never break this logic.

xichen2020: There is a new integration test that covers scenarios including fetching from different keyspaces etc. I can add another unit test for this.

storage/database.go
```diff
@@ -134,6 +133,17 @@ func NewDatabase(shardSet sharding.ShardSet, opts Options) (Database, error) {
 		return nil, errCommitLogStrategyUnknown
 	}
 
+	ns := make(map[string]databaseNamespace, len(namespaces))
+	for _, n := range namespaces {
+		if _, exists := ns[n.Name()]; exists {
+			return nil, errDuplicateNamespaces
+		}
+		// NB(xichen): shardSet is used only for reads but not writes once created
+		// so can be shared by different namespaces
```
robskillington: shardSet is immutable, there should be no issue sharing a shardSet amongst two things no?  Not sure what this comment is getting at.

storage/namespace.go
```diff
@@ -0,0 +1,280 @@
...
+	var totalNumSeries int64
+
+	n.RLock()
+	shards := n.shardSet.Shards()
+	for _, shard := range shards {
+		totalNumSeries += n.shards[shard].NumSeries()
+	}
+	n.RUnlock()
+
+	n.initShards()
```
robskillington: I believe it actually might not be as easy as calling `initShards`.  If we do it this way then nothing will be returned to all the pools.  I think we need to go and manually cleanup each series and call close on all the blocks, etc.  We should probably audit the "expired" series case too and make sure it does this. 

We could potentially do this in a follow up diff if we want to?

storage/database.go
xichen2020: It's basically saying the same thing lol

storage/namespace.go
xichen2020: I was hoping GC will kick in and reclaim the memory after the shards are recreated. After all nothing is holding onto them anyways. I'd imagine if we ever truncate a namespace, it's likely for the purpose of dropping memory footprint, in which case it'll be done when GC kicks in.

client/host_queue.go
xichen2020: Done

robskillington:
LGTM

PR #67, opened 2016-09-01
Refactor bootstrap process to allow for sequential and parallel bootstrap sources

robskillington:


client/session.go
```diff
@@ -758,30 +763,41 @@ func (s *session) streamBlocksMetadataFromPeer(
...
 
 		for _, elem := range result.Elements {
-			blocks := make([]blockMetadata, len(elem.Blocks))
-			for i, b := range elem.Blocks {
-				blocks[i].start = time.Unix(0, b.Start)
+			blockMetas := make([]blockMetadata, 0, len(elem.Blocks))
+			for _, b := range elem.Blocks {
+				var (
+					blockStart = time.Unix(0, b.Start)
+					blockEnd   = start.Add(blockSize)
```
xichen2020: This probably should be `blockEnd = blockStart.Add(blockSize)`?

client/session.go
```diff
@@ -758,30 +763,41 @@ func (s *session) streamBlocksMetadataFromPeer(
...
+				)
+				if start.After(blockEnd) || !blockStart.Before(end) {
```
xichen2020: nit: I think `blockEnd` is exclusive, so could do `!start.Before(blockEnd) || !blockStart.Before(end)`

persist/fs/commitlog/types.go
```diff
@@ -49,8 +48,6 @@ const (
 
 // CommitLog provides a synchronized commit log
 type CommitLog interface {
-	io.Closer
```
xichen2020: Out of curiosity, this is because we needed to use file mode for mockgen and we don't want to supply aux files?

storage/bootstrap/bootstrapper/base.go
```diff
@@ -34,77 +35,118 @@ const (
...
 
-// Bootstrap performs bootstrapping for the given shards and the associated time ranges.
-func (bsb *baseBootstrapper) Bootstrap(shard uint32, targetRanges xtime.Ranges) (bootstrap.ShardResult, xtime.Ranges) {
-	if xtime.IsEmpty(targetRanges) {
+func (b *baseBootstrapper) Can(strategy bootstrap.Strategy) bool {
+	return b.src.Can(strategy)
+}
+
+func (b *baseBootstrapper) Bootstrap(shardsTimeRanges bootstrap.ShardTimeRanges) (bootstrap.Result, error) {
+	if shardsTimeRanges.IsEmpty() {
```
xichen2020: Can `shardsTimeRanges` ever be nil?

storage/bootstrap/bootstrapper/base.go
```diff
@@ -34,77 +35,118 @@ const (
...
 		return nil, nil
 	}
 
-	availableRanges := bsb.s.GetAvailability(shard, targetRanges)
-	remainingRanges := targetRanges.RemoveRanges(availableRanges)
+	available := b.src.Available(shardsTimeRanges)
+
+	remaining := make(map[uint32]xtime.Ranges)
+	for shard, ranges := range shardsTimeRanges {
```
xichen2020: Can we move this logic to a separate method (e.g., `Subtract`) on the `ShardTimeRanges` interface?

storage/bootstrap/bootstrapper/base.go
```diff
@@ -34,77 +35,118 @@ const (
...
+		firstNextUnfulfilled bootstrap.ShardTimeRanges
+	)
+	if nextResult != nil {
+		// Union the results
+		mergedResult.ShardResults().AddResults(nextResult.ShardResults())
+		// Save the first next unfulfilled time ranges
+		firstNextUnfulfilled = nextResult.Unfulfilled()
+	} else {
+		// Union just the unfulfilled ranges from current and the remaining ranges
+		currUnfulfilled.AddRanges(remaining)
```
xichen2020: This only happens if both `nextErr` and `nextResult` are `nil`, which means `remaining` is empty (?)

xichen2020: I see so this can also happen when the source is not parallizable

storage/bootstrap/bootstrapper/base.go
```diff
@@ -34,77 +35,118 @@ const (
...
 	opts bootstrap.Options
-	s    bootstrap.Source
+	src  bootstrap.Source
 	next bootstrap.Bootstrapper
 }
 
 // NewBaseBootstrapper creates a new base bootstrapper.
 func NewBaseBootstrapper(
-	s bootstrap.Source,
+	src bootstrap.Source,
```
cw9: need gofmt?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -21,70 +21,211 @@
...
-		if series.Shard != shard {
+
+		blockStart := dp.Timestamp.Truncate(blockSize)
+		blockEnd := blockStart.Add(blockSize)
+		blockRange := xtime.Range{
+			Start: blockStart,
+			End:   blockEnd,
+		}
+
+		ranges, ok := shardsTimeRanges[series.Shard]
```
xichen2020: nit: can be done a bit earlier at line 94

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -21,70 +21,211 @@
...
+					for iter.Next() {
+						dp, unit, annotation := iter.Current()
+						encodeErr := enc.Encode(dp, unit, annotation)
+						if encodeErr != nil {
+							err = encodeErr
+							errs++
+							break
+						}
+					}
+					if err != nil {
```
xichen2020: Probably should be `if err == nil`

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -21,70 +21,211 @@
...
+						err = iter.Err()
+					}
+
+					iter.Close()
+					for i := range unmergedBlock {
+						unmergedBlock[i].enc.Close()
+					}
+
+					if err != nil {
+						continue
```
xichen2020: Log the error perhaps?

client/session.go
robskillington: Indeed, cheers.

storage/bootstrap/bootstrapper/commitlog/source.go
xichen2020: I see, so we log how many errors happened during the process, which is probably fine

persist/fs/commitlog/types.go
robskillington: Indeed it is (facepalm)

storage/bootstrap/bootstrapper/base.go
robskillington: It should never be no, it's considered just a callable interface type essentially.

storage/bootstrap/bootstrapper/base.go
robskillington: Sounds good.

storage/bootstrap/bootstrapper/base.go
robskillington: Yeah, this is primarily for the source is not parallelizable concern.

storage/bootstrap/bootstrapper/commitlog/source.go
robskillington: Sure thing.

storage/bootstrap/bootstrapper/commitlog/source.go
robskillington: Good catch, cheers.

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -25,36 +25,54 @@ import (
...
 		readerBufferSize: opts.GetFilesystemOptions().GetReaderBufferSize(),
 		newReaderFn:      fs.NewReader,
 	}
 }
 
-// GetAvailability returns what time ranges are available for a given shard
-func (fss *fileSystemSource) GetAvailability(shard uint32, targetRangesForShard xtime.Ranges) xtime.Ranges {
+func (s *fileSystemSource) Can(strategy bootstrap.Strategy) bool {
+	switch strategy {
+	case bootstrap.BootstrapParallel:
```
xichen2020: nit: `case bootstrap.BootstrapParallel, bootstrap.BootstrapSequential`

client/session.go
robskillington: Sounds good.

storage/bootstrap/process.go
```diff
@@ -68,26 +71,43 @@ func (b *bootstrapProcess) initTimeRanges(writeStart time.Time) xtime.Ranges {
...
-		time.Sleep(b.opts.GetThrottlePeriod())
+
+		b.log.WithFields(
+			xlog.NewLogField("shardsLen", len(shards)),
+			xlog.NewLogField("from", window.Start.String()),
+			xlog.NewLogField("to", window.End.String()),
+			xlog.NewLogField("length", window.End.Sub(window.Start).String()),
+		).Infof("bootstrapping shards for range")
+
+		res, err := b.bootstrapper.Bootstrap(shardsTimeRanges)
```
xichen2020: hm..not sure if we want to bail if there is an error, i'd add whatever we've bootstrapped to `result` and rely on background repair to patch the gaps

storage/bootstrap/result.go
```diff
@@ -22,14 +22,57 @@ package bootstrap
...
+	return r.results
+}
+
+func (r *result) Unfulfilled() ShardTimeRanges {
+	return r.unfulfilled
+}
+
+func (r *result) AddShardResult(shard uint32, result ShardResult, unfulfilled xtime.Ranges) {
+	if result != nil && len(result.AllSeries()) > 0 {
+		r.results[shard] = result
```
xichen2020: What if we already have some result for `shard`? Is there a usecase for that?

storage/bootstrap/types.go
```diff
@@ -50,34 +72,169 @@ type ShardResult interface {
...
-
 	// Close closes a shard result.
 	Close()
 }
 
+// ShardResults is a map of shards to shard results.
+type ShardResults map[uint32]ShardResult
+
+// AddResults adds other shard results to the current shard results.
+func (r ShardResults) AddResults(other ShardResults) {
```
xichen2020: Move the implementation to a separate file (e.g., `result.go`)? `types.go` should only contain types

storage/bootstrap/types.go
```diff
@@ -50,34 +72,169 @@ type ShardResult interface {
...
+		if it.Next() && otherIt.Next() {
+			value := it.Value()
+			otherValue := otherIt.Value()
+			if !value.Start.Equal(otherValue.Start) ||
+				!value.End.Equal(otherValue.End) {
+				return false
+			}
+			matched++
+		}
+		if matched != ranges.Len() {
```
xichen2020: If we compare the length upfront, seems we don't need `matched`?

storage/shard.go
```diff
@@ -349,24 +351,13 @@ func (s *dbShard) Bootstrap(bs bootstrap.Bootstrap, writeStart time.Time, cutove
 	s.Unlock()
 
 	multiErr := xerrors.NewMultiError()
-	sr, err := bs.Run(writeStart, s.shard)
-	if err != nil {
-		renamedErr := fmt.Errorf("error occurred bootstrapping shard %d from external sources: %v", s.shard, err)
-		err = xerrors.NewRenamedError(err, renamedErr)
+	for id, dbBlocks := range bootstrappedSeries {
+		series, _, completionFn := s.writableSeries(id)
```
xichen2020: Indentation is out of place

storage/bootstrap/bootstrapper/fs/source.go
robskillington: Sounds good.

storage/bootstrap/process.go
robskillington: I wanted to introduce a way to bail inconsolably.  The bootstrappers are all written to never return unless there is a very severe error.  The only case that it currently would actually exhibit this behavior is if all bootstrappers fail and the commit log bootstrapper cannot open the files needed to bootstrap. 

i.e. only dies in the case where FS cannot be accessed.

storage/bootstrap/result.go
robskillington: Yeah I thought this was going to be handled elsewhere but on review seems it's not.  Might be best to merge with ShardResult AddResult(other ShardResult) method

storage/bootstrap/types.go
robskillington: Yeah, sounds good.

storage/bootstrap/types.go
robskillington: This is in case it.Next() returns false even though it said it had a certain length.  Thoughts on keeping this check or discarding?  It does serve some purpose, but then I don't know how likely it is that it could ever happen.

storage/shard.go
robskillington: ? Doesn't look like it:
screen shot 2016-09-01 at 5 12 38 pm

xichen2020: Welp..it's out of place on the "Files Changed" page, if it's not, all good

storage/bootstrap/types.go
xichen2020: Yeah probably not necessary, the implementation is such that if there are more items, it's guaranteed to return true if you call `Next` on the iterator

storage/bootstrap/result.go
xichen2020: Sounds good

storage/bootstrap/process.go
xichen2020: In that case I suppose the results returned from all the bootstrappers will just be `nil` so it works either way.

I'm fine with returning `nil` on error but would prefer putting a comment on the bootstrapper interface to explicitly call it out that bootstrappers should only return severe errors so future bootstrapper implementers are aware.

storage/bootstrap/types.go
robskillington: Ok sure thing, I'll remove this.

storage/bootstrap/process.go
robskillington: Will add a comment.

storage/bootstrap.go
```diff
@@ -121,41 +124,17 @@ func (bsm *bootstrapManager) Bootstrap() error {
...
-		results := result.ShardResults()
-		for _, shard := range shards {
-			var bootstrappedSeries map[string]block.DatabaseSeriesBlocks
-			if result, ok := results[shard.ID()]; ok {
-				bootstrappedSeries = result.AllSeries()
-			}
-
-			err := shard.Bootstrap(bootstrappedSeries, writeStart, cutover)
+	for _, namespace := range bsm.database.getOwnedNamespaces() {
+		bs := bsm.newBootstrapFn()
```
xichen2020: This can be moved out of the for loop

storage/shard.go
```diff
@@ -22,6 +22,7 @@ package storage
 
 import (
 	"container/list"
+	"fmt"
```
xichen2020: Why do we need this now?

xichen2020:
LGTM

PR #68, opened 2016-09-02
Convert dependency management to glide

jra3:
Summary:
It is tedious for (glide based) projects depending on m3db to
also keep in sync the dependencies of m3x, m3cluster, and any other
deps of m3db. Using glide allows other projects to use this code much
more easily.

Also it eliminates the usage of git submodules which is almost always
the right choice...

Test Plan:
```
rm -rf ./vendor
make install-ci
make
```


jra3:
a 0.01% coverage regression counts as a failure for coveralls. Boooooooo.

Makefile
```diff
@@ -69,9 +69,9 @@ tools-linux-amd64:
 $(foreach SERVICE,$(SERVICES),$(eval $(SERVICE_RULES)))
 $(foreach TOOL,$(TOOLS),$(eval $(TOOL_RULES)))
 
-install-vendor: .gitmodules
-	@echo Updating submodules
-	git submodule update --init --recursive
+install-vendor: install-glide
```
xichen2020: Probably also need to fix up `license_dir` in the Makefile to point to the new path under .ci

robskillington: Yeah think it's as simple as setting `license_dir` on line 14 to `.ci/uber-licence`

jra3:
Tested this by running `make install-license-bin`

Note that this command uses the proper american spelling

xichen2020:
LGTM

robskillington:
LGTM

PR #69, opened 2016-09-06
Move context creation and closing to tchannel server hooks

robskillington:
cc @xichen2020 @martin-mao @mmihic

### DIFF
--- a/network/server/httpjson/handlers.go
+++ b/network/server/httpjson/handlers.go
@@ -138,7 +138,6 @@ func RegisterHandlers(mux *http.ServeMux, service interface{}, opts ServerOption
 			var in interface{}
 			if reqIn != nil {
 				in = reflect.New(reqIn.Elem()).Interface()
-				defer r.Body.Close()
 				if err := json.NewDecoder(r.Body).Decode(in); err != nil {
 					writeError(w, errInvalidRequestBody)
 					return
--- a/network/server/httpjson/handlers.go
+++ b/network/server/httpjson/handlers.go
@@ -31,11 +31,13 @@ import (
 
 	"github.com/m3db/m3x/errors"
 
+	apachethrift "github.com/apache/thrift/lib/go/thrift"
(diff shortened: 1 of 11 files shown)


network/server/httpjson/handlers.go
```diff
@@ -108,23 +117,62 @@ func RegisterHandlers(mux *http.ServeMux, service interface{}, opts ServerOption
...
+				if len(values) > 0 {
+					headers[key] = values[0]
+				}
 			}
 
-			svc := reflect.ValueOf(service)
+			var in interface{}
+			if reqIn != nil {
+				in = reflect.New(reqIn.Elem()).Interface()
+				defer r.Body.Close()
```
xichen2020: We should close the body regardless whether it's a POST or a GET

robskillington: I'm pretty sure if you never open it you don't need to close it.

xichen2020: hm..just realized this is a request body, in which case I don't think we need to close it in either case.

http://grokbase.com/t/gg/golang-nuts/138p15ng41/go-nuts-http-request-body-close

robskillington: Good to know documentation still doesn't exist about this.  Cool I'll remove close.

xichen2020:
LGTM

network/server/httpjson/handlers.go
xichen2020: Well yeah kind of..it's hidden in the comments about the `Body` field inside the `Request` definition:
```
        // Body is the request's body.
        //
        // For client requests a nil body means the request has no
        // body, such as a GET request. The HTTP Client's Transport
        // is responsible for calling the Close method.
        //
        // For server requests the Request Body is always non-nil
        // but will return EOF immediately when no body is present.
        // The Server will close the request body. The ServeHTTP
        // Handler does not need to.
        Body io.ReadCloser
```

PR #71, opened 2016-09-07
Refactor dynamic topology

cw9:
Originally DynamicTopology was waiting on Source.WaitOnInit() channel, changed to simply wait on watch.C()

### DIFF
--- a/topology/dynamic.go
+++ b/topology/dynamic.go
@@ -28,17 +28,16 @@ import (
 	"github.com/m3db/m3cluster/services"
 	"github.com/m3db/m3db/sharding"
 	"github.com/m3db/m3x/log"
-	"github.com/m3db/m3x/retry"
 	"github.com/m3db/m3x/watch"
 )
 
 var (
-	errInitTimeOut               = errors.New("timed out initializing source")
+	errInitTimeOut               = errors.New("timed out waiting for initial value")
 	errInvalidService            = errors.New("service topology is invalid")
 	errUnexpectedShard           = errors.New("shard is unexpected")
 	errMissingShard              = errors.New("shard is missing")
 	errNotEnoughReplicasForShard = errors.New("replicas of shard is less than expected")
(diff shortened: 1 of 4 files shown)


cw9:
@xichen2020 @robskillington

topology/dynamic.go
```diff
@@ -113,66 +143,28 @@ func (t *dynamicTopology) Close() {
 
 	t.closed = true
 
-	t.w.Close()
-	t.s.Close()
+	t.watch.Close()
+	t.watchable.Close()
```
xichen2020: Closing the watchable should also close all the watches associated with the watchable, so don't think you need to call `watch.Close()` here.

topology/dynamic.go
```diff
@@ -113,66 +143,28 @@ func (t *dynamicTopology) Close() {
...
-	}); finalErr != nil {
-		return nil, finalErr
-	}
-
-	return m, nil
-}
-
-func newMapFromServiceInstances(data interface{}, hashGen sharding.HashGen) (Map, error) {
+func getMapFromUpdate(data interface{}, hashGen sharding.HashGen) (Map, error) {
 	service, ok := data.(services.Service)
```
xichen2020: Do we need to check if data is nil here and return an error if so?

cw9: if data is nil, ok will be false here, unless you want a specific err for nil interface{}?

topology/dynamic.go
cw9: the watch is watching the updates from config service, and then use this watchable to broadcast its values to so they are not associated.

xichen2020: Right yeah I see, cool cool

topology/dynamic.go
xichen2020: Nah it's fine

xichen2020:
LGTM

robskillington:
Should we perhaps remove the WaitInit function then completely from m3x/watch if this is a more generalized/better approach?

PR #78, opened 2016-09-09
Add peers bootstrapping integration tests for selecting blocks and node down

robskillington:

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -947,17 +947,22 @@ func (s *session) streamBlocksFromPeers(
 			currStart, currEligible, blocksMetadataQueues)
 
 		// Insert work into peer queues
-		var (
-			completed = uint32(0)
-			queues    = uint32(blocksMetadatas(perPeerBlocksMetadata).hasBlocksLen())
-			onDone    = func() {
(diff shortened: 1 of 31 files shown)


client/session.go
```diff
@@ -947,17 +947,22 @@ func (s *session) streamBlocksFromPeers(
...
+		if queues == 0 {
+			// No blocks at all available from any peers, series may have just expired
+			enqueueCh.trackProcessed(1)
+			continue
+		}
+
+		completed := uint32(0)
+		onDone := func() {
+			// Mark completion of work from the enqueue channel when all queues drained
+			if atomic.AddUint32(&completed, 1) != queues {
```
martin-mao: It's not possible for there to be more things on the channel than queues right? Otherwise this will go ahead before they're all complete.

robskillington: Correct, queues is the exact amount we are going to enqueue because in the for loop below the only queues we do not enqueue into are the ones that have no blocks.  We calculate `queues` by doing the same for loop calculation and the data structures are not mutated between then and here.

martin-mao:
LGTM

PR #79, opened 2016-09-11
Update TestEncoderUnsealSealed to uncover bug with Unseal

robskillington:
cc @xichen2020 

Can you have a look at this bug?


robskillington:
We are abandoning unsealing encoders and blocks.

PR #82, opened 2016-09-12
Add peers bootstrap merge local writes integration test

robskillington:


integration/data.go
```diff
@@ -97,36 +99,88 @@ func verifySeriesMapForRange(
...
+	)
+	if debugFilePathPrefix != "" {
+		for _, entry := range []struct {
+			suffix   string
+			filePath *string
+		}{
+			{"expected.log", &expectedDebugFilePath},
+			{"actual.log", &actualDebugFilePath},
+		} {
+			*entry.filePath = debugFilePathPrefix + "_" + entry.suffix
```
cw9: nit: put the file operations in a function and call it on "expected.log" and "actual.log" separately? not sure if it worth to loop through 2 structs

robskillington: This is just a test harness however?

robskillington: Will call another function though, guess it's not too much of a pollution.

integration/data.go
```diff
@@ -97,37 +99,84 @@ func verifySeriesMapForRange(
...
+	}
+	if len(actualDebugFilePath) > 0 {
+		writeVerifyDebugOutput(t, actualDebugFilePath, start, end, actual)
+	}
+
 	require.Equal(t, expected, actual)
 }
 
+func writeVerifyDebugOutput(t *testing.T, filePath string, start, end time.Time, series seriesList) {
+	w, err := os.OpenFile(filePath, os.O_APPEND|os.O_WRONLY, os.ModeAppend)
```
martin-mao: Does this get cleaned up later since it's append?

storage/namespace.go
```diff
@@ -216,18 +219,35 @@ func (n *dbNamespace) Bootstrap(
 	}
 	n.metrics.bootstrap.Success.Inc(1)
 
+	// Bootstrap shards using at least half the CPUs available
+	workers := pool.NewWorkerPool(int(math.Ceil(float64(runtime.NumCPU()) / 2)))
```
martin-mao: Why not just do runtime.NumCPU() / 2 and leave it to the int rounding.

integration/data.go
robskillington: This is an integration test component that is only enabled "when you want it".  It's for viewing the difference between expected and actual datapoints that arrived at the destination node.  It needs to live on after the test run is finished so you can do a diff between the two files.

storage/namespace.go
```diff
@@ -216,18 +219,35 @@ func (n *dbNamespace) Bootstrap(
...
+			var bootstrappedSeries map[string]block.DatabaseSeriesBlocks
+			if result, ok := results[shard.ID()]; ok {
+				bootstrappedSeries = result.AllSeries()
+			}
 
-		err := shard.Bootstrap(bootstrappedSeries, writeStart, cutover)
-		multiErr = multiErr.Add(err)
+			err := shard.Bootstrap(bootstrappedSeries, writeStart, cutover)
+
+			mutex.Lock()
```
martin-mao: Nit: move unlock to line below multiErr to a void defer overhead.

storage/namespace.go
robskillington: I don't want it to round down to 0 if NumCPU() == 1.  e.g. https://play.golang.org/p/J8MuG1wGNF

storage/namespace.go
robskillington: Sure thing.

storage/series.go
```diff
@@ -333,33 +338,54 @@ func (s *dbSeries) bufferDrained(start time.Time, encoder encoding.Encoder) {
...
-		}
-		blockStart := dp.Timestamp.Truncate(s.opts.RetentionOptions().BlockSize())
-		block := blocks.BlockOrAdd(blockStart)
+	multiIter := s.opts.MultiReaderIteratorPool().Get()
+	multiIter.Reset(readers)
+	defer multiIter.Close()
 
-		if err := block.Write(dp.Timestamp, dp.Value, unit, annotation); err != nil {
+	blopts := s.opts.DatabaseBlockOptions()
+	encoder := s.opts.EncoderPool().Get()
```
martin-mao: Need to defer close the encoder for return on line 380?

robskillington: Encoders get left open and it is up to the block to close it when the block is finished with.

storage/namespace.go
```diff
@@ -216,18 +219,35 @@ func (n *dbNamespace) Bootstrap(
 	}
 	n.metrics.bootstrap.Success.Inc(1)
 
+	// Bootstrap shards using at least half the CPUs available
+	workers := pool.NewWorkerPool(int(math.Ceil(float64(runtime.NumCPU()) / 2)))
+	workers.Init()
+
 	multiErr := xerrors.NewMultiError()
+	mutex := sync.Mutex{}
```
xichen2020: nit: using `var mutex sync.Mutex` is probably more idiomatic

xichen2020:
other than the nit LGTM

storage/series.go
martin-mao: But it won't get to the block until line 382. What happens in the case where we get an error on line 380, the encoder would be fetched from the pool and never returned no?

PR #83, opened 2016-09-13
Compare blocks metadata periodically in the background

xichen2020:
cc @robskillington @martin-mao @cw9

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -658,6 +658,17 @@ func (s *session) Close() error {
 	return nil
 }
 
+func (s *session) Origin() topology.Host {
+	return s.origin
+}
+
+func (s *session) Replicas() int {
+	s.RLock()
+	replicas := s.replicas
+	s.RUnlock()
+	return replicas
+}
+
 func (s *session) Truncate(namespace string) (int64, error) {
 	var (
 		wg            sync.WaitGroup
@@ -702,6 +713,94 @@ func (s *session) Truncate(namespace string) (int64, error) {
 	return truncated, resultErr.FinalError()
 }
 
+func (s *session) getPeersForShard(shard uint32) ([]hostQueue, error) {
+	s.RLock()
+	peers := make([]hostQueue, 0, s.topoMap.Replicas())
+	err := s.topoMap.RouteShardForEach(shard, func(idx int, host topology.Host) {
+		if s.origin != nil && s.origin.ID() == host.ID() {
+			// Don't include the origin host
+			return
+		}
+		peers = append(peers, s.queues[idx])
+	})
+	s.RUnlock()
+	if err != nil {
+		return nil, err
+	}
+	return peers, nil
+}
+
+type metadataIter struct {
+	inputCh  <-chan blocksMetadata
+	errCh    <-chan error
+	host     topology.Host
+	blocks   []block.Metadata
+	metadata block.BlocksMetadata
+	done     bool
+	err      error
+}
+
+func newMetadataIter(inputCh <-chan blocksMetadata, errCh <-chan error) PeerBlocksMetadataIter {
+	return &metadataIter{
+		inputCh: inputCh,
+		errCh:   errCh,
+		blocks:  make([]block.Metadata, 0, 64),
+	}
+}
+
+func (it *metadataIter) Next() bool {
+	if it.done || it.err != nil {
+		return false
+	}
+	m, more := <-it.inputCh
+	if !more {
+		it.err = <-it.errCh
+		it.done = true
+		return false
+	}
+	it.host = m.peer.Host()
+	it.blocks = it.blocks[:0]
+	for _, b := range m.blocks {
+		bm := block.NewMetadata(b.start, b.size, b.checksum)
+		it.blocks = append(it.blocks, bm)
+	}
+	it.metadata = block.NewBlocksMetadata(m.id, it.blocks)
+	return true
+}
+
+func (it *metadataIter) Current() (topology.Host, block.BlocksMetadata) {
+	return it.host, it.metadata
+}
+
+func (it *metadataIter) Err() error {
+	return it.err
+}
+
+func (s *session) FetchBlocksMetadataFromPeers(
+	namespace string,
+	shard uint32,
+	start, end time.Time,
+	blockSize time.Duration,
+) (PeerBlocksMetadataIter, error) {
+	peers, err := s.getPeersForShard(shard)
+	if err != nil {
+		return nil, err
+	}
+
+	var (
+		metadataCh = make(chan blocksMetadata, 4096)
+		errCh      = make(chan error, 1)
+	)
+
+	go func() {
+		errCh <- s.streamBlocksMetadataFromPeers(namespace, shard, peers, start, end, blockSize, metadataCh)
+		close(metadataCh)
+		close(errCh)
+	}()
+
+	return newMetadataIter(metadataCh, errCh), nil
+}
+
 func (s *session) FetchBootstrapBlocksFromPeers(
 	namespace string,
 	shard uint32,
@@ -722,16 +821,7 @@ func (s *session) FetchBootstrapBlocksFromPeers(
 		}
 	)
 
-	s.RLock()
-	peers := make([]hostQueue, 0, s.topoMap.Replicas())
-	err := s.topoMap.RouteShardForEach(shard, func(idx int, host topology.Host) {
-		if s.origin != nil && s.origin.ID() == host.ID() {
-			// Don't include the origin host
-			return
-		}
-		peers = append(peers, s.queues[idx])
-	})
-	s.RUnlock()
+	peers, err := s.getPeersForShard(shard)
 	if err != nil {
 		return nil, err
 	}
@@ -815,8 +905,9 @@ func (s *session) streamBlocksMetadataFromPeer(
 				SetMax(3).
 				SetInitialBackoff(time.Second).
 				SetJitter(true))
-		optionIncludeSizes = true
-		moreResults        = true
+		optionIncludeSizes     = true
+		optionIncludeChecksums = true
+		moreResults            = true
 	)
 	// Declare before loop to avoid redeclaring each iteration
 	attemptFn := func() error {
@@ -832,6 +923,7 @@ func (s *session) streamBlocksMetadataFromPeer(
 		req.Limit = int64(s.streamBlocksBatchSize)
 		req.PageToken = pageToken
 		req.IncludeSizes = &optionIncludeSizes
+		req.IncludeChecksums = &optionIncludeChecksums
 
 		result, err := client.FetchBlocksMetadata(tctx, req)
 		if err != nil {
@@ -876,9 +968,16 @@ func (s *session) streamBlocksMetadataFromPeer(
 					continue
 				}
 
+				var pChecksum *uint32
+				if b.Checksum != nil {
+					cksum := uint32(*b.Checksum)
+					pChecksum = &cksum
+				}
+
 				blockMetas = append(blockMetas, blockMetadata{
-					start: blockStart,
-					size:  *b.Size,
+					start:    blockStart,
+					size:     *b.Size,
+					checksum: pChecksum,
 				})
 			}
 			ch <- blocksMetadata{
@@ -1164,9 +1263,10 @@ func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
 		for i := range currStart {
 			unselected := currStart[i].unselectedBlocks()
 			metadata := blockMetadataReattemptPeerMetadata{
-				peer:  currStart[i].peer,
-				start: unselected[0].start,
-				size:  unselected[0].size,
+				peer:     currStart[i].peer,
+				start:    unselected[0].start,
+				size:     unselected[0].size,
+				checksum: unselected[0].checksum,
 			}
 			peersMetadata = append(peersMetadata, metadata)
 		}
@@ -1333,6 +1433,7 @@ func (s *session) streamBlocksReattemptFromPeers(
 					blocks: []blockMetadata{blockMetadata{
 						start:     blocks[i].reattempt.peersMetadata[j].start,
 						size:      blocks[i].reattempt.peersMetadata[j].size,
+						checksum:  blocks[i].reattempt.peersMetadata[j].checksum,
 						reattempt: blocks[i].reattempt,
 					}},
 				}
@@ -1684,6 +1785,7 @@ func (arr blocksMetadatasQueuesByOutstandingAsc) Less(i, j int) bool {
 type blockMetadata struct {
 	start     time.Time
 	size      int64
+	checksum  *uint32
 	reattempt blockMetadataReattempt
 }
 
@@ -1695,9 +1797,10 @@ type blockMetadataReattempt struct {
 }
 
 type blockMetadataReattemptPeerMetadata struct {
-	peer  hostQueue
-	start time.Time
-	size  int64
+	peer     hostQueue
+	start    time.Time
+	size     int64
+	checksum *uint32
 }
 
 func (b blockMetadataReattempt) hasAttemptedPeer(peer hostQueue) bool {
--- a/client/session.go
+++ b/client/session.go
@@ -50,8 +50,9 @@ import (
 )
 
 const (
-	clusterConnectWaitInterval    = 10 * time.Millisecond
-	blocksMetadataInitialCapacity = 64
+	clusterConnectWaitInterval           = 10 * time.Millisecond
+	blocksMetadataInitialCapacity        = 64
+	blocksMetadataChannelInitialCapacity = 4096
 )
 
 var (
@@ -716,7 +717,7 @@ func (s *session) Truncate(namespace string) (int64, error) {
 
 func (s *session) peersForShard(shard uint32) ([]hostQueue, error) {
 	s.RLock()
-	peers := make([]hostQueue, 0, s.topoMap.Replicas())
+	peers := make([]hostQueue, 0, s.topoMap.Replicas()-1)
 	err := s.topoMap.RouteShardForEach(shard, func(idx int, host topology.Host) {
 		if s.origin != nil && s.origin.ID() == host.ID() {
 			// Don't include the origin host
@@ -743,7 +744,7 @@ func (s *session) FetchBlocksMetadataFromPeers(
 	}
 
 	var (
-		metadataCh = make(chan blocksMetadata, 4096)
+		metadataCh = make(chan blocksMetadata, blocksMetadataChannelInitialCapacity)
 		errCh      = make(chan error, 1)
 	)
 
--- a/client/session.go
+++ b/client/session.go
@@ -50,7 +50,8 @@ import (
 )
 
 const (
-	clusterConnectWaitInterval = 10 * time.Millisecond
+	clusterConnectWaitInterval    = 10 * time.Millisecond
+	blocksMetadataInitialCapacity = 64
 )
 
 var (
@@ -713,7 +714,7 @@ func (s *session) Truncate(namespace string) (int64, error) {
 	return truncated, resultErr.FinalError()
 }
 
-func (s *session) getPeersForShard(shard uint32) ([]hostQueue, error) {
+func (s *session) peersForShard(shard uint32) ([]hostQueue, error) {
 	s.RLock()
 	peers := make([]hostQueue, 0, s.topoMap.Replicas())
 	err := s.topoMap.RouteShardForEach(shard, func(idx int, host topology.Host) {
@@ -730,59 +731,13 @@ func (s *session) getPeersForShard(shard uint32) ([]hostQueue, error) {
 	return peers, nil
 }
 
-type metadataIter struct {
-	inputCh  <-chan blocksMetadata
-	errCh    <-chan error
-	host     topology.Host
-	blocks   []block.Metadata
-	metadata block.BlocksMetadata
-	done     bool
-	err      error
-}
-
-func newMetadataIter(inputCh <-chan blocksMetadata, errCh <-chan error) PeerBlocksMetadataIter {
-	return &metadataIter{
-		inputCh: inputCh,
-		errCh:   errCh,
-		blocks:  make([]block.Metadata, 0, 64),
-	}
-}
-
-func (it *metadataIter) Next() bool {
-	if it.done || it.err != nil {
-		return false
-	}
-	m, more := <-it.inputCh
-	if !more {
-		it.err = <-it.errCh
-		it.done = true
-		return false
-	}
-	it.host = m.peer.Host()
-	it.blocks = it.blocks[:0]
-	for _, b := range m.blocks {
-		bm := block.NewMetadata(b.start, b.size, b.checksum)
-		it.blocks = append(it.blocks, bm)
-	}
-	it.metadata = block.NewBlocksMetadata(m.id, it.blocks)
-	return true
-}
-
-func (it *metadataIter) Current() (topology.Host, block.BlocksMetadata) {
-	return it.host, it.metadata
-}
-
-func (it *metadataIter) Err() error {
-	return it.err
-}
-
 func (s *session) FetchBlocksMetadataFromPeers(
 	namespace string,
 	shard uint32,
 	start, end time.Time,
 	blockSize time.Duration,
 ) (PeerBlocksMetadataIter, error) {
-	peers, err := s.getPeersForShard(shard)
+	peers, err := s.peersForShard(shard)
 	if err != nil {
 		return nil, err
 	}
@@ -821,7 +776,7 @@ func (s *session) FetchBootstrapBlocksFromPeers(
 		}
 	)
 
-	peers, err := s.getPeersForShard(shard)
+	peers, err := s.peersForShard(shard)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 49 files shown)


client/session.go
```diff
@@ -702,6 +713,94 @@ func (s *session) Truncate(namespace string) (int64, error) {
...
+		peers = append(peers, s.queues[idx])
+	})
+	s.RUnlock()
+	if err != nil {
+		return nil, err
+	}
+	return peers, nil
+}
+
+type metadataIter struct {
```
robskillington: Perhaps move the `metadataIter` struct to its own file, it's a little strange sitting sandwiched between `*session` methods?

client/session.go
```diff
@@ -702,6 +713,94 @@ func (s *session) Truncate(namespace string) (int64, error) {
...
+	metadata block.BlocksMetadata
+	done     bool
+	err      error
+}
+
+func newMetadataIter(inputCh <-chan blocksMetadata, errCh <-chan error) PeerBlocksMetadataIter {
+	return &metadataIter{
+		inputCh: inputCh,
+		errCh:   errCh,
+		blocks:  make([]block.Metadata, 0, 64),
```
robskillington: Perhaps make the capacity of this a const?

client/session.go
```diff
@@ -702,6 +713,94 @@ func (s *session) Truncate(namespace string) (int64, error) {
 	return truncated, resultErr.FinalError()
 }
 
+func (s *session) getPeersForShard(shard uint32) ([]hostQueue, error) {
```
robskillington: Maybe name this just `peersForShard` to be inline with the Effective Go advice?

client/session.go
```diff
@@ -702,6 +713,94 @@ func (s *session) Truncate(namespace string) (int64, error) {
...
+	it.blocks = it.blocks[:0]
+	for _, b := range m.blocks {
+		bm := block.NewMetadata(b.start, b.size, b.checksum)
+		it.blocks = append(it.blocks, bm)
+	}
+	it.metadata = block.NewBlocksMetadata(m.id, it.blocks)
+	return true
+}
+
+func (it *metadataIter) Current() (topology.Host, block.BlocksMetadata) {
```
robskillington: Hm, how do you know which peer it came from?

client/session.go
```diff
@@ -876,9 +968,16 @@ func (s *session) streamBlocksMetadataFromPeer(
 					continue
 				}
 
+				var pChecksum *uint32
+				if b.Checksum != nil {
+					cksum := uint32(*b.Checksum)
```
robskillington: Maybe call `cksum` just `value`?

client/session.go
```diff
@@ -1695,9 +1797,10 @@ type blockMetadataReattempt struct {
 }
 
 type blockMetadataReattemptPeerMetadata struct {
-	peer  hostQueue
-	start time.Time
-	size  int64
+	peer     hostQueue
+	start    time.Time
+	size     int64
+	checksum *uint32
```
robskillington: Probably don't need to add checksum to the reattempt peer metadata.  Checksums aren't used for selecting the best block on reattempting fetching a block.

storage/repair/types.go
```diff
@@ -0,0 +1,120 @@
...
+	ChecksumDifferences ReplicaSeriesMetadata
+}
+
+// Options are the repair options
+type Options interface {
+	// SetNewAdminSessionFn sets the new admin session function
+	SetNewAdminSessionFn(value client.NewAdminSessionFn) Options
+
+	// NewAdminSessionFn returns the new admin session function
+	NewAdminSessionFn() client.NewAdminSessionFn
```
robskillington: I think perhaps just making this take an actual `AdminSession` is better.  That way the same admin session that the peers bootstrapper takes and the repairer can be the same shared one.  Session's are pretty heavy and use a reasonable amount of memory so would a much better idea to share.

storage/repair.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func (r *dbRepairer) Start() {
+	if r.repairInterval <= 0 {
+		return
+	}
+
+	start := r.nowFn()
+	for {
+		r.RLock()
```
cw9: seems the lock is only used around r.closed, a Mutex is enough?

network/server/tchannelthrift/node/service.go
```diff
@@ -207,6 +212,10 @@ func (s *service) FetchBlocksMetadata(tctx thrift.Context, req *rpc.FetchBlocksM
 			blockMetadata := rpc.NewBlockMetadata()
 			blockMetadata.Start = xtime.ToNanoseconds(fetchedMetadataBlock.Start())
 			blockMetadata.Size = fetchedMetadataBlock.Size()
+			if checksum := fetchedMetadataBlock.Checksum(); checksum != nil {
+				cksum := int64(*checksum)
```
robskillington: `value` instead of `cksum` perhaps?

storage/repair/types.go
xichen2020: That was the initial approach I went with, but I later realized that means I needed to create a session before creating/open the database, which means the client would try to connect to the servers before the database is open (which is naturally before the actual server startup).

The peers bootstrapper and the repairer can still share the same session, we just need to capture the session in a closure that's passed to the repair options.

client/session.go
xichen2020: Correct right now we don't use checksums to select the best peer, but I was thinking we might need to use that info later, e.g., when the peer checksums disagree, we might want to pick the peers whose checksums are the majority.

client/session.go
xichen2020: `blocksMetadata` has a `peer` field, which is the host queue where the request was served from

client/session.go
xichen2020: It was only used in `session.go` but sure

client/session.go
xichen2020: sure

client/session.go
xichen2020: ok

client/session.go
xichen2020: Sure

network/server/tchannelthrift/node/service.go
xichen2020: Sure

storage/repair.go
xichen2020: Sure

storage/shard.go
```diff
@@ -317,9 +317,10 @@ func (s *dbShard) FetchBlocksMetadata(
 	limit int64,
 	pageToken int64,
 	includeSizes bool,
-) ([]FetchBlocksMetadataResult, *int64) {
+	includeChecksums bool,
+) ([]block.FetchBlocksMetadataResult, *int64) {
 	var (
-		res            = make([]FetchBlocksMetadataResult, 0, limit)
+		res            = make([]block.FetchBlocksMetadataResult, 0, int(math.Min(float64(1024), float64(limit))))
```
xichen2020: Sure

client/session.go
robskillington: I see, yup makes sense!

client/session.go
robskillington: Fair enough, makes sense.

storage/repair/types.go
robskillington: That's fine, opening a session before opening the database is ok.  That's why I implemented the `ConnectConsistencyLevelAny` so that you can attempt to connect to all, then to majority, then to one and then to none.

generated/thrift/rpc.thrift
```diff
@@ -107,7 +108,8 @@ struct BlocksMetadata {
 struct BlockMetadata {
 	1: required i64 start
 	2: optional i64 size
-	3: optional Error err
+	3: optional i64 checksum
+	4: optional Error err
```
robskillington: Hm, maybe we should make err as field 1 so we don't have to make breaking changes in the future like this?

storage/block/result.go
```diff
@@ -88,10 +98,78 @@ type fetchBlocksMetadataResult struct {
...
+
+type filteredBlocksMetadataIter struct {
+	start     time.Time
+	end       time.Time
+	blockSize time.Duration
+	res       []FetchBlocksMetadataResult
+
+	id       string
+	metadata Metadata
+	ri       int
```
robskillington: What's `ri` and `bi`?  Perhaps a more descriptive name?

storage/block/result.go
```diff
@@ -88,10 +98,78 @@ type fetchBlocksMetadataResult struct {
...
+	bi       int
+}
+
+// NewFilteredBlocksMetadataIter creates a new filtered blocks metadata iterator
+func NewFilteredBlocksMetadataIter(
+	start time.Time,
+	end time.Time,
```
robskillington: Can make this `start, end time.Time,`

storage/block/result.go
```diff
@@ -88,10 +98,78 @@ type fetchBlocksMetadataResult struct {
...
+func (it *filteredBlocksMetadataIter) Next() bool {
+	if it.ri >= len(it.res) {
+		return false
+	}
+	blocks := it.res[it.ri].Blocks()
+	for it.bi < len(blocks) {
+		block := blocks[it.bi]
+		if block.Err() != nil {
+			it.bi++
+			continue
```
robskillington: Hm, wonder if this is something we want to know about, if errors are being read back for some blocks.  i.e. we might have 100% perfect checksum match rate with a peer but then there might be errors for say 50% of the blocks we try to retrieve checksums for, that would be an issue that goes unnoticed if we just skip over blocks that returned error from the peer no?

storage/database.go
```diff
@@ -201,6 +211,8 @@ func (d *db) Open() error {
 	// All goroutines must be accounted for with dbOngoingTasks to receive done signal
 	go d.reportLoop()
 	go d.ongoingTick()
+	go d.repairer.Start()
```
robskillington: Have to change `dbOngoingTasks` to be 3 now.

robskillington: Maybe add an endline before calling this and explicitly mention that `Stop()` is explicitly called on close so that's why `dbOngoingTasks` is not accounting for this process?

storage/repair.go
```diff
@@ -0,0 +1,278 @@
...
+		closed := r.closed
+		r.RUnlock()
+
+		if closed {
+			break
+		}
+
+		end := r.nowFn()
+		if end.Sub(start) < r.repairInterval {
+			time.Sleep(r.repairCheckInterval)
```
robskillington: I don't get this loop, won't this always just wait `r.repairCheckInterval`? Start gets set _after_ running a repair, then end gets set directly after checking if closed or not (which will be very fast, i.e. 1ms).  Why not just wait `r.repairCheckInterval`?

For a method that waits the difference of the deadline perhaps use the same method as the splayed tick?

storage/repair.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+func (r *dbRepairer) Repair() error {
+	// Don't attempt a repair if the database is not bootstrapped yet
+	if !r.database.IsBootstrapped() {
+		return nil
+	}
+
+	// Lazily create the shard repairer
+	if r.shardRepairer == nil {
```
robskillington: Any reason to create this lazily?

storage/repair.go
```diff
@@ -0,0 +1,278 @@
...
+
+func newShardRepairer(opts Options, rpOpts repair.Options) (databaseShardRepairer, error) {
+	newSessionFn := rpOpts.NewAdminSessionFn()
+	session, err := newSessionFn()
+	if err != nil {
+		return nil, err
+	}
+
+	iOpts := opts.InstrumentOptions()
+	rtOpts := opts.RetentionOptions()
```
robskillington: For these short names, I've actually not been camel casing them. i.e. `iopts := opts.InstrumentOptions()` and `ropts := opts.RetentionOptions()`

storage/repair.go
```diff
@@ -0,0 +1,278 @@
...
+
+	r := shardRepairer{
+		opts:    opts,
+		rtOpts:  rtOpts,
+		session: session,
+
+		logger:    iOpts.Logger(),
+		scope:     iOpts.MetricsScope().SubScope("database"),
```
robskillington: Perhaps better scope is "repair"?

storage/repair/types.go
xichen2020: Yeah just that it seems a bit weird to see a bunch of connection errors even if the server starts up normally.

generated/thrift/rpc.thrift
xichen2020: I'm fine w/ that

storage/block/result.go
xichen2020: `resultIdx` and `blockIdx` as you mentioned. Not as terse as `ri` and `bi` though but yeah can change

storage/block/result.go
xichen2020: sure

storage/block/result.go
xichen2020: In that case, there will be no checksum for this peer, which means the number of hosts with checksums for this block is going to be smaller than the replica count, in which case we are going to log this block in the result.

storage/database.go
xichen2020: sure thing

storage/repair.go
```diff
@@ -0,0 +1,278 @@
...
+			"shard":     strconv.Itoa(int(shard.ID())),
+		})
+		totalScope        = shardScope.Tagged(map[string]string{"resultType": "total"})
+		sizeDiffScope     = shardScope.Tagged(map[string]string{"resultType": "sizeDiff"})
+		checksumDiffScope = shardScope.Tagged(map[string]string{"resultType": "checksumDiff"})
+	)
+
+	// Record total number of series and total number of blocks
+	totalScope.Counter("repair.series").Inc(diffRes.NumSeries)
+	totalScope.Counter("repair.blocks").Inc(diffRes.NumBlocks)
```
robskillington: Create these counters on the shard repairer on instantiation?

storage/repair.go
```diff
@@ -0,0 +1,278 @@
...
+	r.logger.WithFields(
+		xlog.NewLogField("namespace", namespace),
+		xlog.NewLogField("shard", shard.ID()),
+		xlog.NewLogField("resultType", "total"),
+	).Infof("numSeries=%d, numBlocks=%d", diffRes.NumSeries, diffRes.NumBlocks)
```
robskillington: Should make NumSeries and NumBlocks log fields as well I would think?

storage/repair.go
xichen2020: Yes this always waits `r.repairCheckInterval` until `r.repairInterval` has elapsed. Are you suggesting setting `end`  after instead of before the sleep? I can do that but it should work either way.

storage/repair.go
xichen2020: The same reason I'm not creating session upfront, to avoid the spurious connection errors. But if we are fine with those errors, there's no need for lazy creation.

client/session.go
robskillington: Can just move to bottom of file then perhaps?  Main issue is just being sandwiched between `*session` methods.

xichen2020: Yeah can do

storage/repair/types.go
robskillington: Yeah I mean we just need to limit our logging to after it finishes connecting, etc.  I can put up a diff for that if you'd like?

storage/block/result.go
robskillington: Ah ok, no problems.  Ignore my comment!

storage/repair.go
robskillington: Repair interval will probably be like, 5 minutes.  Perhaps just sleep before doing the closed check at the top of the loop then?

storage/repair/types.go
xichen2020: Sure yeah, I guess I can also add that to this diff too, either way works

storage/repair.go
xichen2020: guess it'll have to be SubScope("database").SubScope("repair")

storage/repair.go
xichen2020: Unfortunately these depend on the `shardScope`, which are namespace+shard specific. We could create all of them upfront and store them in a map keyed off of namespace, shard combination but seems overkill if we only run repair once every 2 hours

storage/repair.go
xichen2020: sure

storage/database.go
```diff
@@ -198,9 +208,14 @@ func (d *db) Open() error {
...
 	d.state = databaseOpen
 
-	// All goroutines must be accounted for with dbOngoingTasks to receive done signal
+	// Goroutines must be accounted for with dbOngoingTasks in order to receive done signal
 	go d.reportLoop()
 	go d.ongoingTick()
+
+	// Explicitly start the repairer in Open and stop it in Close so there is no need to
+	// register this goroutine with dbOngoingTasks
+	go d.repairer.Start()
```
robskillington: Actually maybe the repairer should launch the goroutine itself, wouldn't this make sense?  Most components we have Start/Stop methods for they start the routines themselves and are non-blocking.

xichen2020: I wanted it to be explicit w.r.t. whether a goroutine is spawned. I tried to follow the same pattern as other components but couldn't find any with Start/Stop. That said, I don't feel too strongly either way.

storage/repair.go
```diff
@@ -0,0 +1,283 @@
...
+		numBlocks += len(series.Blocks())
+	}
+	checksumDiffScope.Counter("series").Inc(int64(len(checksumDiffSeries)))
+	checksumDiffScope.Counter("blocks").Inc(int64(numBlocks))
+	r.logger.WithFields(
+		xlog.NewLogField("namespace", namespace),
+		xlog.NewLogField("shard", shard.ID()),
+		xlog.NewLogField("numSeries", len(checksumDiffSeries)),
+		xlog.NewLogField("numBlocks", numBlocks),
+	).Infof("repair checksum difference count")
```
robskillington: Hm wonder if we should actually sum across all shards and log once, that's going to be likely best no?  For metrics by shard is fine but for logging it's going to be super noisy and hard to read =[

storage/database.go
robskillington: Can you change this to `Run()` if you want to keep it?  Otherwise we should make it non-blocking and call it `Start` and have it spawn its own goroutine.

xichen2020: I'll make it nonblocking then

client/session.go
```diff
@@ -702,6 +714,48 @@ func (s *session) Truncate(namespace string) (int64, error) {
 	return truncated, resultErr.FinalError()
 }
 
+func (s *session) peersForShard(shard uint32) ([]hostQueue, error) {
+	s.RLock()
+	peers := make([]hostQueue, 0, s.topoMap.Replicas())
```
martin-mao: Does replicas include the origin host? If so, should this be peers := make([]hostQueue, 0, s.topoMap.Replicas()-1)?

client/session.go
```diff
@@ -702,6 +714,48 @@ func (s *session) Truncate(namespace string) (int64, error) {
...
+	start, end time.Time,
+	blockSize time.Duration,
+) (PeerBlocksMetadataIter, error) {
+	peers, err := s.peersForShard(shard)
+	if err != nil {
+		return nil, err
+	}
+
+	var (
+		metadataCh = make(chan blocksMetadata, 4096)
```
martin-mao: Should this 4096 be in some config/const somewhere?

client/session.go
```diff
@@ -876,9 +923,16 @@ func (s *session) streamBlocksMetadataFromPeer(
 					continue
 				}
 
+				var pChecksum *uint32
+				if b.Checksum != nil {
```
martin-mao: What are the use cases where b.Checksum == nil? Are we ok with creating a blockMetadata with checksum of nil pointer?

storage/repair.go
xichen2020: I don't think its too bad considering this only runs every two hours and each node owns 64 shards, or 192 lines / 2 hours. Don't feel too strongly either though.

Might also need to include hostname in the tags though, otherwise the metrics are going to be aggregated across hosts and not useful.

network/server/tchannelthrift/node/channel/channel.go
```diff
@@ -0,0 +1,26 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package channel
+
+const (
+	// ChannelName is the TChannel channel name the node service is exposed on
+	ChannelName = "Node"
```
martin-mao: Lol awesome file ;)

network/server/tchannelthrift/node/service.go
```diff
@@ -207,6 +212,10 @@ func (s *service) FetchBlocksMetadata(tctx thrift.Context, req *rpc.FetchBlocksM
 			blockMetadata := rpc.NewBlockMetadata()
 			blockMetadata.Start = xtime.ToNanoseconds(fetchedMetadataBlock.Start())
 			blockMetadata.Size = fetchedMetadataBlock.Size()
+			if checksum := fetchedMetadataBlock.Checksum(); checksum != nil {
```
martin-mao: We're ok with just not having a checksum on the blockMetadata? I'm guessing this is for blocks that are not yet sealed. Is there a better way we can indicate that the block is not sealed, ie with a sealed field instead of empty checksum?

storage/block/block.go
```diff
@@ -169,6 +180,7 @@ func (b *dbBlock) Seal() {
 	b.writable = false
 	if stream := b.encoder.Stream(); stream != nil {
 		b.segment = stream.Segment()
+		b.checksum = digest.SegmentChecksum(b.segment)
```
martin-mao: Do you need to set b.checksum to 0 on Unseal method? We can depend on writable I guess, but seems like an implicit dependency that is hard to remember to keep in future code changes.

storage/repair.go
```diff
@@ -0,0 +1,283 @@
...
+
+func (r *dbRepairer) Start() {
+	if r.repairInterval <= 0 {
+		return
+	}
+
+	var curIntervalStart time.Time
+
+	for {
+		r.Lock()
```
martin-mao: Nit: Use RWMutex and r.RLock()?

storage/repair/metadata.go
```diff
@@ -0,0 +1,196 @@
...
+	block, exists := m[start]
+	if exists {
+		return block
+	}
+	block = NewReplicaBlockMetadata(start)
+	m[start] = block
+	return block
+}
+
+type replicaSeriesMetadata map[string]ReplicaBlocksMetadata
```
martin-mao: Worth calling out that this doesn't support multiple threads?

storage/repair.go
robskillington: I more mean to read it ourselves.  The metrics are fine we can sum them together, but the log message will be unreadable.

storage/shard.go
```diff
@@ -317,9 +318,13 @@ func (s *dbShard) FetchBlocksMetadata(
...
 	pageToken int64,
 	includeSizes bool,
-) ([]FetchBlocksMetadataResult, *int64) {
+	includeChecksums bool,
+) ([]block.FetchBlocksMetadataResult, *int64) {
 	var (
-		res            = make([]FetchBlocksMetadataResult, 0, limit)
+		// Restrict the maximum capacity so we don't over allocate or panic if
+		// someone passes in a very large limit
+		resCapacity    = int(math.Min(float64(blocksMetadataResultMaxInitialCapacity), float64(limit)))
```
martin-mao: Is it cheaper to do if statement?

client/session.go
xichen2020: It doesn't include itself, sure can make the change

client/session.go
xichen2020: sure it was refactored from previous code, but can make the change

network/server/tchannelthrift/node/channel/channel.go
xichen2020: Yeah this is mainly to break up the circular dependency between the client package and the storage package, so now storage depends on client but not the other way around

network/server/tchannelthrift/node/service.go
xichen2020: There can be a variety of reasons, e.g., blocks not sealed, or the `includeChecksum` flag is set to false in the RPC call (which we don't do now but might do if the sizes are good enough and we want to save some bandwidth and processing power), or some error happened when reading/decoding the block's data.

storage/block/block.go
xichen2020: There is no `Unseal` method for the block, once it's sealed, it'll remain sealed. We do reset the checksum when the block is reset.

storage/repair.go
xichen2020: I started with `RWMutex` but the only time we need a write lock is when the database is closed (which only happens during server shutdown), so a `Mutex` should be more performant in this case.

storage/repair/metadata.go
xichen2020: Sure I'll add a comment. It's only used by the repairer though.

storage/shard.go
xichen2020: Sure can switch to if

client/session.go
xichen2020: See blow

PR #89, opened 2016-09-18
Make host queue drain channel buffered to match size of op arrays pool

robskillington:
cc @xichen2020 @martin-mao


xichen2020:
LGTM

PR #90, opened 2016-09-18
Sample timers and add service-level metrics to accurately measure latencies

xichen2020:


network/server/tchannelthrift/node/service.go
```diff
@@ -240,34 +298,48 @@ func (s *service) FetchBlocksMetadata(tctx thrift.Context, req *rpc.FetchBlocksM
...
 	return result, nil
 }
 
 func (s *service) Write(tctx thrift.Context, req *rpc.WriteRequest) error {
+	sw := s.metrics.write.Latency.Start()
 	ctx := tchannelthrift.Context(tctx)
 
 	if req.IdDatapoint == nil || req.IdDatapoint.Datapoint == nil {
+		s.metrics.write.Error.Inc(1)
+		s.metrics.write.Latency.Stop(sw)
```
robskillington: Maybe we should just log the latency if it's a successful operation?  I separated out success/error latency in `m3x/retry` package, however here maybe we just emit latency on success?

This could help reduce some of the duplicate code statements in this file.

instrument/options.go
```diff
@@ -27,9 +27,15 @@ import (
 	"github.com/uber-go/tally"
 )
 
+const (
+	defaultSamplingRate   = 0.01
```
robskillington: Perhaps we should make the default 1.0?

network/server/tchannelthrift/node/service.go
xichen2020: Sure, doing it this way because that's how it was done inside database.go and namespace.go.

I'll add an ErrorLatency and fix the existing ones in db and namespace as well.

instrument/options.go
xichen2020: sure

PR #94, opened 2016-09-22
Add jitter and throttle when comparing block metadata in the background

xichen2020:
cc @robskillington @martin-mao @cw9

### DIFF
--- a/storage/namespace.go
+++ b/storage/namespace.go
@@ -359,6 +359,7 @@ func (n *dbNamespace) Repair(repairer databaseShardRepairer) error {
 
 	multiErr := xerrors.NewMultiError()
 	shards := n.getOwnedShards()
+	throttle := repairer.Options().RepairShardThrottle()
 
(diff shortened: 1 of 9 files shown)


storage/namespace.go
```diff
@@ -373,6 +374,9 @@ func (n *dbNamespace) Repair(repairer databaseShardRepairer) error {
 			numChecksumDiffSeries += metadataRes.ChecksumDifferences.NumSeries()
 			numChecksumDiffBlocks += metadataRes.ChecksumDifferences.NumBlocks()
 		}
+		if throttle > 0 {
+			time.Sleep(throttle)
```
robskillington: Maybe we should have a throttle value that is fixed and we divide it by how many shards we own to get the value for this sleep here?  That way we'll get a known amount of sleeping no matter how many shards a node might own.

xichen2020: sure can do

PR #95, opened 2016-09-23
Add workerpool for writing batch

xichen2020:
cc @robskillington

### DIFF
--- a/client/host_queue.go
+++ b/client/host_queue.go
@@ -30,6 +30,7 @@ import (
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	"github.com/m3db/m3db/topology"
 
+	"github.com/uber/tchannel-go"
 	"github.com/uber/tchannel-go/thrift"
 )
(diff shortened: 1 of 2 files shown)


network/server/tchannelthrift/node/service.go
```diff
@@ -327,38 +334,64 @@ func (s *service) WriteBatch(tctx thrift.Context, req *rpc.WriteBatchRequest) er
...
+			if err != nil {
+				atomic.AddUint32(&numErrs, 1)
+				if xerrors.IsInvalidParams(err) {
+					errCh <- tterrors.NewBadRequestWriteBatchError(i, err)
+				} else {
+					errCh <- tterrors.NewWriteBatchError(i, err)
+				}
 			} else {
-				errs = append(errs, tterrors.NewWriteBatchError(i, err))
+				errCh <- nil
```
cw9: why do we need to write a nil to errCh?

PR #96, opened 2016-09-26
Instrument pools, add refill low watermark to object pool, don't use contexts for large operations

robskillington:
cc @xichen2020 @martin-mao

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -45,6 +45,7 @@ import (
 	"github.com/m3db/m3x/log"
 	"github.com/m3db/m3x/retry"
 	xtime "github.com/m3db/m3x/time"
+	"github.com/uber-go/tally"
 
 	"github.com/uber/tchannel-go/thrift"
 )
@@ -77,6 +78,7 @@ type session struct {
 	sync.RWMutex
 
 	opts                             Options
+	scope                            tally.Scope
 	nowFn                            clock.NowFn
 	log                              xlog.Logger
 	level                            topology.ConsistencyLevel
@@ -121,6 +123,7 @@ func newSession(opts Options) (clientSession, error) {
 
 	s := &session{
 		opts:                 opts,
+		scope:                opts.InstrumentOptions().MetricsScope(),
 		nowFn:                opts.ClockOptions().NowFn(),
 		log:                  opts.InstrumentOptions().Logger(),
 		level:                opts.ConsistencyLevel(),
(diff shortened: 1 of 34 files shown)


client/session.go
```diff
@@ -172,11 +175,20 @@ func (s *session) Open() error {
...
+		SetMetricsScope(s.scope.SubScope("fetch-batch-op-pool"))
+	s.fetchBatchOpPool = newFetchBatchOpPool(fetchBatchOpPoolOpts, s.fetchBatchSize)
 	s.fetchBatchOpPool.Init()
-	s.seriesIteratorPool = encoding.NewSeriesIteratorPool(s.opts.SeriesIteratorPoolSize())
+	seriesIteratorPoolOpts := pool.NewObjectPoolOptions().
+		SetSize(s.opts.SeriesIteratorPoolSize()).
+		SetMetricsScope(s.scope.SubScope("series-iterator-pool"))
+	s.seriesIteratorPool = encoding.NewSeriesIteratorPool(seriesIteratorPoolOpts)
 	s.seriesIteratorPool.Init()
 	s.seriesIteratorsPool = encoding.NewMutableSeriesIteratorsPool(s.opts.SeriesIteratorArrayPoolBuckets())
```
xichen2020: Do we need to instrument `seriesIteratorsPool` as well?

pool/object.go
```diff
@@ -42,17 +81,57 @@ func (p *objectPool) Init(alloc Allocator) {
 func (p *objectPool) Get() interface{} {
 	select {
 	case v := <-p.values:
+		p.trySetGauges()
```
xichen2020: Move `p.trySetGauges` out of select perhaps?

pool/object.go
```diff
@@ -42,17 +81,57 @@ func (p *objectPool) Init(alloc Allocator) {
...
+		}
+		p.metrics.getOnEmpty.Inc(1)
 		return p.alloc()
 	}
 }
 
 func (p *objectPool) Put(obj interface{}) {
 	select {
 	case p.values <- obj:
+		p.trySetGauges()
```
xichen2020: Move `p.trySetGauges` out of select perhaps?

pool/object.go
```diff
@@ -42,17 +81,57 @@ func (p *objectPool) Init(alloc Allocator) {
...
+
+func (p *objectPool) tryFill() {
+	if !atomic.CompareAndSwapInt64(&p.filling, 0, 1) {
 		return
 	}
+	go func() {
+		full := false
+		for !full {
+			select {
+			case p.values <- p.alloc():
```
xichen2020: I wonder if we always want to replenish the entire pool as opposed to say 25% of the pool size, because if the demand for more objects is transient, we might allocate more than necessary

pool/object.go
robskillington: I tried that, however that would mean the gauge would never appear as zero because it would be before the select.  I could try using a local var perhaps and call this after assignment, let me try that.

client/session.go
robskillington: You can always pass noop scope as the metric scope if you wish.  I figure all or nothing.  We can find a way to opt out of certain pools being instrumented if possible.

PR #98, opened 2016-09-28
Optimize filesystem bootstrapper performance

xichen2020:

This reduces the time to bootstrap 48 hours of data from disk on a single node from 30 mins to 3-4 mins.

### DIFF
--- a/storage/bootstrap/bootstrapper/fs/source.go
+++ b/storage/bootstrap/bootstrapper/fs/source.go
@@ -21,9 +21,11 @@
 package fs
 
 import (
+	"sync"
 	"time"
 
 	"github.com/m3db/m3db/persist/fs"
+	"github.com/m3db/m3db/pool"
 	"github.com/m3db/m3db/storage/bootstrap"
 	"github.com/m3db/m3x/log"
 	"github.com/m3db/m3x/time"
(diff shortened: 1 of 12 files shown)


storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -95,76 +107,196 @@ func (s *fileSystemSource) shardAvailability(
...
+	go s.enqueueReaders(namespace, shardsTimeRanges, readersCh)
+	result := s.bootstrapFromReaders(readersCh)
+
+	end := time.Now()
+
+	s.log.WithFields(
+		xlog.NewLogField("namespace", namespace),
+		xlog.NewLogField("numShards", len(shardsTimeRanges)),
+		xlog.NewLogField("duration", end.Sub(start).String()),
+	).Info("finished reading data from filesystem for bootstrapping")
```
robskillington: Should we keep these more detailed logs?  If so perhaps we should make them more generic and do it one layer high in the bootstrap process?

storage/bootstrap/process.go
```diff
@@ -67,18 +67,22 @@ func (b *bootstrapProcess) Run(
 			shardsTimeRanges[s] = r
 		}
 
-		b.log.WithFields(
+		logFields := []xlog.LogField{
 			xlog.NewLogField("namespace", namespace),
 			xlog.NewLogField("numShards", len(shards)),
 			xlog.NewLogField("from", window.Start.String()),
 			xlog.NewLogField("to", window.End.String()),
 			xlog.NewLogField("length", window.End.Sub(window.Start).String()),
```
robskillington: Should we add the name of the bootstrapper we're calling somehow?

storage/namespace.go
```diff
@@ -225,6 +229,8 @@ func (n *dbNamespace) Bootstrap(
 	}
 	n.metrics.bootstrap.Success.Inc(1)
 
+	n.log.WithFields(namespaceLogField).Info("finished reading external data")
```
robskillington: Probably don't need the log statements in this file yeah?

storage/bootstrap/bootstrapper/fs/source.go
xichen2020: meh, they are somewhat useful but for now i'll move logging into the bootstrap manager

storage/namespace.go
xichen2020: It does provide timing breakdown of the different stages during bootstrap, will move it to bootstrap manager for now

storage/bootstrap/process.go
xichen2020: Sure thing

PR #101, opened 2016-09-30
Optimize block draining when we don't need to merge

xichen2020:

### DIFF
--- a/storage/series.go
+++ b/storage/series.go
@@ -347,34 +347,49 @@ func (s *dbSeries) bufferDrained(start time.Time, encoder encoding.Encoder) {
 
 	// NB(r): this will occur if after bootstrap we have a partial
 	// block and now the buffer is passing the rest of that block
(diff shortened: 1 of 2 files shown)


storage/series.go
```diff
@@ -347,34 +347,49 @@ func (s *dbSeries) bufferDrained(start time.Time, encoder encoding.Encoder) {
 
 	// NB(r): this will occur if after bootstrap we have a partial
 	// block and now the buffer is passing the rest of that block
-	stream := encoder.Stream()
-	s.drainStreamWithLock(s.blocks, start, stream)
-	stream.Close()
+	s.drainStreamWithLock(s.blocks, start, encoder)
 }
 
 func (s *dbSeries) drainStreamWithLock(
```
robskillington: Maybe rename to `drainBufferedEncoderWithLock`?

xichen2020: sure

PR #103, opened 2016-09-30
Add inner splay to the tick and make default tick duration 2mins

robskillington:
cc @xichen2020 @martin-mao

### DIFF
--- a/storage/database.go
+++ b/storage/database.go
@@ -108,6 +108,8 @@ type db struct {
 
 	created      uint64
 	tickDeadline time.Duration
+	ticking      int64
+	flushing     int64
 
 	scope   tally.Scope
 	metrics databaseMetrics
@@ -118,7 +120,9 @@ type db struct {
 type databaseMetrics struct {
 	bootstrapStatus tally.Gauge
 	tickStatus      tally.Gauge
+	flushStatus     tally.Gauge
 
+	tickDuration       tally.Timer
 	tickDeadlineMissed tally.Counter
 	tickDeadlineMet    tally.Counter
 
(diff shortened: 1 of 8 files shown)


storage/database.go
```diff
@@ -408,24 +415,39 @@ func (d *db) splayedTick() {
 		return
 	}
 
+	// Cleanup and/or flush if required to cleanup files and/or flush blocks to disk
+	if now := d.nowFn(); d.fsm.ShouldRun(now) {
```
xichen2020: We need to move this after ticking because the existing logic is such that the blocks are flushed as soon as possible but no earlier than when they become available. If we run the file ops before ticking, there is a possibility where the blocks just become available but haven't been drained yet so we'll miss those.

storage/shard.go
```diff
@@ -164,16 +172,25 @@ func (s *dbShard) forBatchWithLock(
...
+func (s *dbShard) Tick(softDeadline time.Duration) {
+	s.tickAndExpire(softDeadline)
 }
 
-func (s *dbShard) tickAndExpire() int {
+func (s *dbShard) tickAndExpire(softDeadline time.Duration) int {
 	var (
-		expired []databaseSeries
-		total   int
+		perEntrySoftDeadline = softDeadline / time.Duration(s.NumSeries())
```
xichen2020: Perhaps check `Numseries` isn't zero here?

storage/database.go
robskillington: Sure thing.

storage/shard.go
robskillington: Yeah just added that.

PR #104, opened 2016-10-02
Add jittering when flushing data to disk

xichen2020:

### DIFF
--- a/storage/fs.go
+++ b/storage/fs.go
@@ -21,6 +21,7 @@
 package storage
 
 import (
+	"math/rand"
 	"sync"
 	"time"
 )
@@ -46,6 +47,7 @@ type fileSystemManager struct {
 
 	database  database               // storage database
 	opts      Options                // database options
+	jitter    time.Duration          // file operation jitter
 	status    fileOpStatus           // current file operation status
 	processed map[time.Time]struct{} // times we have already processed
 }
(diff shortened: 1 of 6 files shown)


robskillington:
We might want to think about the implications of this, for background repair it's probably ok that some take a little longer than others to repair.  However for flushing to disk, delaying the beginning of that process means we extend the time where data is held purely in memory (with the exception of the commit log which is very expensive to bootstrap from).  I think we can get away with a very small jitter, perhaps say 2-5 minutes, but 30 minutes definitely seems to aggressive.

storage/fs.go
```diff
@@ -55,11 +57,19 @@ func newFileSystemManager(database database) databaseFileSystemManager {
 	fm := newFlushManager(database)
 	cm := newCleanupManager(database, fm)
 
+	var jitter time.Duration
+	if maxJitter := opts.FileOpOptions().Jitter(); maxJitter > 0 {
+		nowFn := opts.ClockOptions().NowFn()
+		rand.Seed(nowFn().UnixNano())
```
robskillington: We should avoid setting the global rand seed here, we probably need to create a new rand source and then seed it.

xichen2020: sure can do

xichen2020:
Agreed we want to flush as early as possible, although 2-5 minutes might not be not long enough to spread out disk flushing across all nodes given a reasonably large cluster. Can start with 5 minutes and go from there.

robskillington:
Yeah, perhaps let's go with 5 minutes and I'll try to fix any underlying issues that cause degradation of perf during the actual flush.  

Can you perhaps fix the `dbRepairer` also setting a global `rand.Seed` to use it's own source?

PR #110, opened 2016-10-05
[Perf] Convert namespace and IDs to []byte

kobolog:
This change is meant to remove strings and string casting from the M3DB core so that we spend less time collecting the garbage left after these casts. The majority of string casts happen in Thrift, Protobuf and Sharding code (specifically, when we cast strings to []byte to calculate the hash value).


ts/types.go
```diff
@@ -21,14 +21,69 @@
...
+
+// Data returns the binary representation of an ID
+func (id *id) Data() []byte {
+	return id.data
+}
+
+// Hash returns (and calculates, if needed) the hash for an ID
+func (id *id) Hash() Hash {
+	if bytes.Equal(id.hash[:], null[:]) {
+		id.hash = md5.Sum(id.data)
```
robskillington: We going to try sip hash perhaps?

sharding/shardset.go
```diff
@@ -93,9 +93,9 @@ func validateShards(shards []uint32) error {
 	return nil
 }
 
-// DefaultHashGen generates a HashFn based on murmur32
+// DefaultHashGen generates a HashFn based on ID's internal hash.
 func DefaultHashGen(length int) HashFn {
-	return func(id string) uint32 {
-		return murmur3.Sum32([]byte(id)) % uint32(length)
+	return func(id ts.ID) uint32 {
+		return uint32(id.Hash().Fold() % uint64(length))
```
robskillington: Hm not sure this is a good idea, if we use the same hash function for map keys as we do for distributing our sharded data that's not a great idea.  The two hash functions have different goals.  I'm worried that we won't get as good distribution for our data shards potentially.

Can't we just do `murmur3.Sum32(id.Data())` here?

client/session.go
```diff
@@ -1334,8 +1339,8 @@ func (s *session) streamBlocksBatchFromPeer(
 			break
 		}
 
-		id := result.Elements[i].ID
-		if id != batch[i].id {
+		id := ts.BinaryID(result.Elements[i].ID)
+		if id.Hash() != batch[i].id.Hash() {
```
robskillington: Hmm don't we need a `bytes.Equal` alternative here?  Isn't `type Hash [...]byte`?  I thought you couldn't compare two `[]byte` slices?

client/session_bootstrap_blocks_test.go
```diff
@@ -1104,7 +1110,8 @@ func (m *fetchBlocksReqMatcher) Matches(x interface{}) bool {
 	}
 
 	for i := range params {
-		if params[i].id != req.Elements[i].ID {
+		id := ts.BinaryID(req.Elements[i].ID)
+		if params[i].id.Hash() != id.Hash() {
```
robskillington: Again, not sure about hash comparison instead of deep equal (which is what default string is)

generated/thrift/rpc.thrift
```diff
@@ -196,3 +194,8 @@ service Cluster {
 	FetchResult fetch(1: FetchRequest req) throws (1: Error err)
 	TruncateResult truncate(1: TruncateRequest req) throws (1: Error err)
 }
+
+struct HealthResult {
+	1: required bool ok
+	2: required string status
+}
```
robskillington: nit: add endline at end of file?

services/m3dbnode/main/main.go
```diff
@@ -119,7 +120,7 @@ func main() {
 		// Currently bootstrapping, cannot send close clean signal
 		log.Warnf("abort bootstrap")
 		go func() {
-			db.Close()
+			// db.Close()
```
robskillington: Why comment this out?

robskillington:
Probably need to rebase soon too.

client/host_queue_test.go
```diff
@@ -526,13 +531,19 @@ func testHostQueueFetchBatches(
 		wg.Done()
 	}
 
+	rawIDs := make([][]byte, len(ids))
+
+	for i, id := range ids {
+		rawIDs[i] = []byte(id)
+	}
```
kobolog: We only do this here.

ts/types.go
kobolog: I'm not so sure about it. I can't really find anything about it being specifically resistant to collisions. I'd rather go with a conservative MD5 here for now and see how it behaves in production.

sharding/shardset.go
kobolog: Don't be worried, one of the points of hash functions is to provide a uniform distribution of the output regardless of the input, which is both applicable to map keys and shard assignment.

kobolog:
Rebased.

PR #112, opened 2016-10-06
Return early on client with majority or one consistency on success criteria

robskillington:
cc @martin-mao

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -86,8 +86,8 @@ type session struct {
 	topo                             topology.Topology
 	topoMap                          topology.Map
 	topoWatch                        topology.MapWatch
-	replicas                         int
-	majority                         int
+	replicas                         int32
+	majority                         int32
 	queues                           []hostQueue
 	state                            state
 	writeOpPool                      writeOpPool
@@ -310,9 +310,11 @@ func (s *session) setTopologyWithLock(topologyMap topology.Map, queues []hostQue
 	prev := s.queues
 	s.queues = queues
 	s.topoMap = topologyMap
-	prevReplicas := s.replicas
-	s.replicas = replicas
-	s.majority = majority
+
+	prevReplicas := atomic.LoadInt32(&s.replicas)
+	atomic.StoreInt32(&s.replicas, int32(replicas))
+	atomic.StoreInt32(&s.majority, int32(majority))
+
 	if s.fetchBatchOpArrayArrayPool == nil ||
 		s.fetchBatchOpArrayArrayPool.Entries() != len(queues) {
 		poolOpts := pool.NewObjectPoolOptions().
@@ -328,7 +330,7 @@ func (s *session) setTopologyWithLock(topologyMap topology.Map, queues []hostQue
 		prevReplicas != s.replicas {
 		s.iteratorArrayPool = encoding.NewIteratorArrayPool([]pool.Bucket{
 			pool.Bucket{
-				Capacity: s.replicas,
+				Capacity: replicas,
 				Count:    s.opts.SeriesIteratorPoolSize(),
 			},
 		})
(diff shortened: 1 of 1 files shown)


client/session.go
```diff
@@ -417,29 +420,66 @@ func (s *session) Write(namespace string, id string, t time.Time, value float64,
 	w.datapoint.Timestamp = ts
 	w.datapoint.TimestampType = timeType
 	w.datapoint.Annotation = annotation
+
+	wg.Add(1)
+
+	s.RLock()
+	majority = int32(s.majority)
```
martin-mao: Nit: Why not use atomic for majority and move the RLock below this.

client/session.go
```diff
@@ -417,29 +420,66 @@ func (s *session) Write(namespace string, id string, t time.Time, value float64,
...
+			}
+		case topology.ConsistencyLevelAll:
+			if doneAll && atomic.CompareAndSwapInt32(&wgIsDone, 0, 1) {
+				wg.Done()
+			}
+		}
+		// Finalize resources that can be returned
+		if doneAll {
+			// Return write to pool
+			s.writeOpPool.Put(w)
```
martin-mao: Does this depend on success of write? ie if a write doesn't succeed, we grab a new write from pool for retries?

client/session.go
```diff
@@ -501,80 +542,113 @@ func (s *session) FetchAll(namespace string, ids []string, startInclusive, endEx
...
-			}
-		}
-
-		// Return fetch ops array array to pool
-		s.fetchBatchOpArrayArrayPool.Put(fetchBatchOpsByHostIdx)
-	}()
-
 	s.RLock()
-	majority = s.majority
+	majority = int32(s.majority)
```
martin-mao: Again, why not use atomic for majority?

client/session.go
```diff
@@ -501,80 +542,113 @@ func (s *session) FetchAll(namespace string, ids []string, startInclusive, endEx
...
 			errs     int32
 		)
 
 		wg.Add(1)
+		allCompletionFn := func() {
+			var reportErr error
+			errsLen := atomic.LoadInt32(&errs)
+			if errsLen > 0 {
+				resultErrLock.RLock()
+				reportErr = firstErr
```
martin-mao: Do we even need firstErr? Why not just assign err to reportErr if it's not nil?

client/session.go
```diff
@@ -501,80 +542,113 @@ func (s *session) FetchAll(namespace string, ids []string, startInclusive, endEx
...
+				resultErrLock.Lock()
+				if resultErr == nil {
+					resultErr = err
+				}
+				resultErrs++
+				resultErrLock.Unlock()
+			} else {
+				// This is the final result, its safe to load "success" now without atomic operation
+				iter := s.seriesIteratorPool.Get()
+				iter.Reset(ids[idx], startInclusive, endExclusive, results[:atomic.LoadInt32(&success)])
```
martin-mao: Do we even return this iter to the seriesIteratorPool? Can't see that in this diff.

client/session.go
robskillington: Yeah can do that, sounds good.

client/session.go
robskillington: So `doneAll` signifies all host's have been contacted that's part of the write which means no one will read `writeOp` anymore.  The retries occur at a layer above this so users of M3DB can specify their own retry policies.

client/session.go
robskillington: `firstErr` is the error that occurred for this read, this is needed as otherwise we wouldn't have a reference to any of the errors that occurred for this ID.  The main reason `reportErr` is introduced is so we can default it to nil and only acquire the read lock if we truly need to read the contents of `firstErr` - it's simply an optimization in reducing the need to call the read lock to read `firstErr`.

client/session.go
robskillington: That happens when you call `iters.Close()` which is returned to the caller.  i.e. the iterators are only done once you've read all the data, as a caller you must call `iters.Close()` which will close each individual iterator including itself to return both itself and all the inner iterators to their respective pools.

martin-mao:
LGTM

PR #115, opened 2016-10-07
Return all errors instead of the first error from nodes for client returned errors

robskillington:
cc @martin-mao @xichen2020 @kobolog

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -399,8 +399,8 @@ func (s *session) Write(namespace string, id string, t time.Time, value float64,
 		enqueued      int32
 		pending       int32
 		resultErrLock sync.RWMutex
-		resultErr     error
 		resultErrs    int32
+		errors        []error
 		majority      int32
 		success       int32
 	)
@@ -431,9 +431,7 @@ func (s *session) Write(namespace string, id string, t time.Time, value float64,
 		if err != nil {
 			atomic.AddInt32(&resultErrs, 1)
 			resultErrLock.Lock()
-			if resultErr == nil {
-				resultErr = err
-			}
+			errors = append(errors, err)
 			resultErrLock.Unlock()
 		} else {
 			snapshotSuccess = atomic.AddInt32(&success, 1)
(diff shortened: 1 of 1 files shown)


client/session.go
```diff
@@ -431,9 +431,7 @@ func (s *session) Write(namespace string, id string, t time.Time, value float64,
 		if err != nil {
 			atomic.AddInt32(&resultErrs, 1)
```
kobolog: What's the point of this atomic integral (`resultErrs`, `pending`, etc) if you immediately grab a lock and extend a slice after that? A simple addition under the lock won't change anything perf-wise but looks cleaner. Disclaimer: I'm not super-familiar with the codebase yet so I might as well just be missing something obvious.

client/session.go
```diff
@@ -717,17 +721,20 @@ func (s *session) FetchAll(namespace string, ids []string, startInclusive, endEx
...
 		return nil
 	}
 
 	// Check consistency level satisfied
 	success := enqueued - resultErrs
 	reportErr := func() error {
-		return xerrors.NewRenamedError(resultErr, fmt.Errorf(
-			"failed to meet %s with %d/%d success, first error: %v",
-			s.level.String(), success, enqueued, resultErr))
+		return xerrors.NewRenamedError(errors[0], fmt.Errorf(
```
kobolog: Does this mean that only the first error will be returned as-is, but it will have all errors formatted into the error string?

client/session.go
robskillington: The main reason the atomic ints are used extensively is to avoid having to use the write/read lock unless there absolutely is an error that must be read and returned.  In the 99% case where there is no error it avoids the use of any lock use whatsoever which is ideal, however it's definitely not extremely clean.

client/session.go
robskillington: Correct, this is a little bit of a hack to give somewhat usefulness to the `IsBadRequestError` or `IsInternalServerError` functions that act on the underlying error underneath the renamed error.

robskillington: For further reading see: https://github.com/m3db/m3db/blob/master/client/errors.go

robskillington:
I'm going to go ahead and merge this in the interest of rolling forward.

PR #117, opened 2016-10-09
Add ability to perform a number of attempts on connection before discarding from connection pool

robskillington:
cc @martin-mao @kobolog @xichen2020

### DIFF
--- a/client/connection_pool.go
+++ b/client/connection_pool.go
@@ -240,10 +240,10 @@ func (p *connPool) healthCheckEvery(interval time.Duration, stutter time.Duratio
 					if err := p.healthCheck(client, p.opts); err != nil {
 						checkErr = err
 						failed++
-					} else {
-						// Healthy
-						break
+						continue
 					}
+					// Healthy
+					break
 				}
 
 				healthy := failed < attempts
--- a/client/connection_pool.go
+++ b/client/connection_pool.go
@@ -42,9 +42,15 @@ import (
 const (
 	channelName = "Client"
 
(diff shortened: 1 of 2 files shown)


client/connection_pool.go
```diff
@@ -41,6 +41,10 @@ import (
 
 const (
 	channelName = "Client"
+
+	// healthCheckFall is the count of health checks to fail to discard a
+	// connection from the connection pool.
+	healthCheckFall = 2
```
xichen2020: `healthCheckFail`?

client/connection_pool.go
```diff
@@ -210,54 +214,69 @@ func (p *connPool) healthCheckEvery(interval time.Duration, stutter time.Duratio
...
+				var (
+					attempts = healthCheckFall
+					failed   = 0
+					checkErr error
+				)
+				for j := 0; j < attempts; j++ {
+					if err := p.healthCheck(client, p.opts); err != nil {
+						checkErr = err
+						failed++
+						continue
```
xichen2020: Should we add some throttling here perhaps? If the health check failures were due to cpu usage spikes, we'll probably need to wait for the cpu usage to subside before the next health check suceeds

client/connection_pool.go
robskillington: Sounds good

client/connection_pool.go
robskillington: Sounds good, I'll add a throttle here that is a factor of the host connect timeout.

PR #118, opened 2016-10-10
Add fixes for block metadata and block exchanges

robskillington:
cc @xichen2020 @martin-mao

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -1354,6 +1354,7 @@ func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
 				currStart[i].idx = idx + 1
 
 				// Set the reattempt metadata
+				currStart[i].blocks[idx].reattempt.id = currID
 				currStart[i].blocks[idx].reattempt.attempt++
 				currStart[i].blocks[idx].reattempt.attempted =
 					append(currStart[i].blocks[idx].reattempt.attempted, peer)
@@ -1392,6 +1393,7 @@ func (s *session) streamBlocksBatchFromPeer(
 	req.Elements = make([]*rpc.FetchBlocksRawRequestElement, len(batch))
 	for i := range batch {
 		starts := make([]int64, len(batch[i].blocks))
+		sort.Sort(blockMetadatasByTime(batch[i].blocks))
 		for j := range batch[i].blocks {
 			starts[j] = batch[i].blocks[j].start.UnixNano()
 		}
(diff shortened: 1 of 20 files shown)


xichen2020:
Apart from the failing tests and the minor comment, LGTM

client/session.go
```diff
@@ -1174,8 +1174,6 @@ func (s *session) streamCollectedBlocksMetadata(
 		}
 		if received.submitted {
 			// Already submitted to enqueue channel
```
martin-mao: Is this meant to be a common case? Maybe replace with metric instead?

client/session.go
```diff
@@ -1437,20 +1437,27 @@ func (s *session) streamBlocksBatchFromPeer(
...
 				s.log.Errorf("stream blocks response from peer %s returned more blocks than expected", peer.Host().String())
 				break
 			}
 
-			block := result.Elements[i].Blocks[j]
+			// Index of the received block could be offset by missed blocks
+			block := result.Elements[i].Blocks[j-missed]
+
 			if block.Start != batch[i].blocks[j].start.UnixNano() {
+				missed++
```
martin-mao: Does this potentially introduce a cascading effect of errors? Ie, one start time mismatched, but all the others matches, then n would be compared with n-1 for the rest of the blocks, causing them all to be wrong. Why not use set overlap and disjoint sets to determine what to reattempt?

client/session.go
```diff
@@ -1895,6 +1902,22 @@ func (b blockMetadatasByTime) Less(i, j int) bool {
 	return b[i].start.Before(b[j].start)
 }
 
+func newTimesByUnixNanos(values []int64) []time.Time {
+	result := make([]time.Time, len(values))
```
martin-mao: Nit: Should we use nano uint64 as return type here and below to reduce mem usage?

storage/namespace.go
```diff
@@ -373,24 +379,39 @@ func (n *dbNamespace) Repair(repairer databaseShardRepairer) error {
...
+		workers.Go(func() {
+			defer wg.Done()
+
+			metadataRes, err := shard.Repair(n.id, repairer)
+
+			mutex.Lock()
+			if err != nil {
+				multiErr = multiErr.Add(err)
+			} else {
+				numShardsRepaired++
```
martin-mao: Nit: Can we just use atomics here and avoid the lock altogether? (Assuming multiErr.Add is thread safe)

context/context.go
```diff
@@ -133,6 +148,10 @@ func (c *ctx) Reset() {
 			// Free any large arrays that are created
 			c.dep.closers = nil
 		} else {
+			for i := range c.dep.closers {
```
robskillington: Unfortunately the array that is backing the slice still holds onto this memory:
http://stackoverflow.com/a/16972044

client/session.go
robskillington: This can definitely happen when there is a node being added to the cluster.  i.e. you expect two results back but since a new node is bootstrapping a third shard you will get an answer back from them too.  This is expected behavior.

client/session.go
robskillington: To avoid this we always order the timestamps ascending, both on the callers and the callees end and the callee only ever returns a subset of timestamps so the only case to cover here is missing blocks.

These blocks can be pretty big considering this is a batch call of 128 IDs so I'd prefer not to allocate a map here if at all possible.

client/session.go
robskillington: This is only for a log output, this is an error log too so needs to be avoided in production anyhow.

storage/namespace.go
robskillington: `multiErr.Add` is not thread safe.

robskillington: Also this happens once per shard, there's only ever a few shards being repaired at any one time.  It's super low contention.

martin-mao: Can we just lock around multiErr = multiErr.Add() then and convert rest to atomics?

xichen2020:
LGTM

PR #119, opened 2016-10-10
Add jitter for database file operations

xichen2020:

### DIFF
--- a/storage/fs.go
+++ b/storage/fs.go
@@ -53,13 +53,17 @@ type fileSystemManager struct {
 	processed map[time.Time]struct{} // times we have already processed
 }
 
-func newFileSystemManager(database database) databaseFileSystemManager {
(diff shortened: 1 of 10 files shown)


robskillington:
Perhaps we should add checking that this is less than block size to storage options as we discussed

storage/fs.go
```diff
@@ -55,11 +58,19 @@ func newFileSystemManager(database database) databaseFileSystemManager {
 	fm := newFlushManager(database)
 	cm := newCleanupManager(database, fm)
 
+	var jitter time.Duration
+	if maxJitter := opts.FileOpOptions().Jitter(); maxJitter > 0 {
```
martin-mao: Nit: Might be worth pulling in the epoch based rand - would be much cheaper than creating a new rand object each time.

PR #120, opened 2016-10-10
Pool IDs & drop some unneeded allocations

kobolog:
This PR introduces pooling of IDs in the RPC layer and removes some extraneous allocations done in the storage layer.


network/server/tchannelthrift/node/service.go
```diff
@@ -74,19 +75,27 @@ type service struct {
...
-		metrics: newServiceMetrics(scope, iops.MetricsSamplingRate()),
-		health:  &rpc.NodeHealthResult_{Ok: true, Status: "up", Bootstrapped: false},
+		nowFn:   db.Options().ClockOptions().NowFn(),
+		metrics: newServiceMetrics(scope, iopts.MetricsSamplingRate()),
+		health: &rpc.NodeHealthResult_{
+			Ok:           true,
+			Status:       "up",
+			Bootstrapped: false,
+		},
+		idpool: ts.NewPool(pool.NewObjectPoolOptions()),
```
robskillington: We should this pool become a part of the `storage.Options` so others can define the pool size, metrics, etc and it is accessible to the rest of storage.

robskillington: i.e. SetIDPool(...) and `IDPool() ts.IDPool` on `storage.Options`

ts/types.go
```diff
@@ -45,27 +50,70 @@ type id struct {
...
+	return &id{data: v.data, hash: v.hash}
+}
+
+// Pool represents an object pool of IDs
+type Pool struct {
+	pool pool.ObjectPool
+}
+
+// NewPool constructs a new Pool
+func NewPool(options pool.ObjectPoolOptions) *Pool {
```
robskillington: `NewIDPool(...)` would be better in case we add a pool for a different type of object in the `ts` package?

Also: we've been returning interface types to ease mocking, for consistency would be good do the same for this pool and unexport the struct.

ts/types.go
```diff
@@ -45,27 +50,70 @@ type id struct {
...
+// StringID returns a new ID based on a string
+func (p *Pool) StringID(v string) ID {
+	id := p.pool.Get().(*id)
+	id.data = []byte(v)
+
+	return id
+}
+
+// Release returns a set of IDs back into the pool
+func (p *Pool) Release(ids ...ID) {
```
robskillington: In each of our other pools we've used `Get(...)` and `Put(...)` to make it similar to the `sync.Pool` API.  Perhaps `GetBinaryID(...)` `GetStringID(...)` and `Put(...)` would be more consistent with our other pools?

storage/shard.go
```diff
@@ -424,9 +427,9 @@ func (s *dbShard) Bootstrap(
 
 	multiErr := xerrors.NewMultiError()
 	for _, dbBlocks := range bootstrappedSeries {
-		series, _, completionFn := s.writableSeries(dbBlocks.ID)
-		err := series.Bootstrap(dbBlocks.Blocks)
-		completionFn()
+		entry := s.writableSeries(dbBlocks.ID)
+		err := entry.series.Bootstrap(dbBlocks.Blocks)
+		entry.decrementWriterCount()
```
robskillington: Perhaps on `entry` we should add a method called `beginWriteOp(...)` that returns the entry itself after incrementing the writer count and then a wrapper for decrement writer count too:
```go
func (e *dbShardEntry) beginWriteOp() *dbShardEntry {
  e.incrementWriterCount()
  return e
}

func (e *dbShardEntry) endWriteOp() *dbShardEntry {
  e.decrementWriterCount()
  return e
}
```

That way we can write this as:
```go
entry := s.writableSeries(dbBlocks.ID).beginWriteOp()
err := entry.series.Bootstrap(dbBlocks.Blocks)
entry.endWriteOp()
```

Which is much more sane.  I'm sure the functions for increment/decrement would get inlined into the two syntactic methods.

persist/fs/commitlog/commit_log.go
```diff
@@ -378,6 +378,9 @@ func (l *commitLog) WriteBehind(
 		return errCommitLogClosed
 	}
 
+	// Clone since it can outlive the enclosing request
+	series.ID = series.ID.Clone()
```
robskillington: Hm, I'd argue we should do this where we create the `series.ID`, not here.

network/server/tchannelthrift/node/service.go
```diff
@@ -307,57 +357,75 @@ func (s *service) Write(tctx thrift.Context, req *rpc.WriteRequest) error {
...
-			}
+			ctx, nsID, tsID, timestamp, elem.Datapoint.Value, unit, elem.Datapoint.Annotation)
+		s.idpool.Release(tsID)
+
+		switch {
+		case err == nil:
+		case xerrors.IsInvalidParams(err):
+			errs = append(errs, tterrors.NewBadRequestWriteBatchRawError(i, err))
+		default:
+			errs = append(errs, tterrors.NewWriteBatchRawError(i, err))
```
robskillington: This switch is neat, however not super idiomatic.  I'd  be in preference of leaving it as the original writing or to avoid one more level of indentation just:
```go
if err != nil && xerrors.IsInvalidParams(err) { 
  errs = append(errs, tterrors.NewBadRequestWriteBatchRawError(i, err))
} else if err != nil {
  errs = append(errs, tterrors.NewWriteBatchRawError(i, err))
}
```

network/server/tchannelthrift/node/service.go
```diff
@@ -206,37 +233,50 @@ func (s *service) FetchBlocksRaw(tctx thrift.Context, req *rpc.FetchBlocksRawReq
...
 		for _, start := range request.Starts {
 			blockStarts = append(blockStarts, xtime.FromNanoseconds(start))
 		}
-		fetched, err := s.db.FetchBlocks(ctx, ns, shardID, id, blockStarts)
+
+		tsID := s.idpool.BinaryID(request.ID)
+		fetched, err := s.db.FetchBlocks(ctx, nsID, shardID, tsID, blockStarts)
+
 		if err != nil {
 			s.metrics.fetchBlocks.ReportError(s.nowFn().Sub(callStart))
```
xichen2020: Probably should return `tsID` to pool on error too

network/server/tchannelthrift/node/service.go
xichen2020: +1

ts/types.go
xichen2020: Will this allocate a slice of IDs under the hood? I'd be in favor of changing it to take one id at a time to minimize heap allocation, since it's only called on two ids at most.

network/server/tchannelthrift/node/service.go
```diff
@@ -74,19 +75,27 @@ type service struct {
 	nowFn   clock.NowFn
 	metrics serviceMetrics
 	health  *rpc.NodeHealthResult_
+	idpool  *ts.Pool
```
xichen2020: Might make sense to make `idpool` an interface to be consistent with other pools in the codebase

storage/shard.go
kobolog: This one makes little sense to me. It's a one-line method that calls another one-line method one the same object?

ts/types.go
kobolog: @xichen2020 I don't think it actually allocates it, but rather does some compiler magic, but  I guess worth benchmarking later – I've removed this part altogether in this PR.

persist/fs/commitlog/commit_log.go
kobolog: Dropped since `Context.DependsOn()` covers this case perfectly.

network/server/tchannelthrift/node/service.go
kobolog: Not applicable anymore.

storage/shard.go
robskillington: Agreed, maybe we should just rename the the existing methods to make it clear what they're doing.

kobolog:
Bump. Let's get this merged!

persist/fs/commitlog/commit_log.go
```diff
@@ -315,6 +316,7 @@ func (l *commitLog) openWriter(now time.Time) error {
 }
 
 func (l *commitLog) Write(
+	ctx context.Context,
```
xichen2020: remove `ctx` since not used here

persist/fs/commitlog/commit_log.go
```diff
@@ -366,6 +370,7 @@ func (l *commitLog) Write(
 }
 
 func (l *commitLog) WriteBehind(
+	ctx context.Context,
```
xichen2020: remove `ctx` since it's not used here

PR #121, opened 2016-10-10
Lazily create contexts and add pooling

xichen2020:


context/pool.go
```diff
@@ -24,23 +24,81 @@ import (
...
+	return p.closersPool.Get()
+}
+
+func (p *poolOfContexts) PutClosers(closers []Closer) {
+	p.closersPool.Put(closers)
 }
 
-func (p *contextPool) Put(context Context) {
-	p.pool.Put(context)
+func createClosers() []Closer {
```
robskillington: Perhaps `allocClosers` is more consistent with our other naming?

PR #123, opened 2016-10-11
Do not prematurely return results to pool and synchronize incrementing success and setting results

xichen2020:


client/session.go
```diff
@@ -639,10 +647,17 @@ func (s *session) FetchAll(namespace string, ids []string, startInclusive, endEx
 					allCompletionFn()
 				}
 			}
+
 			if doneAll {
+				// NB(xichen): wait for allCompletionFn to run so we can safely return results to
+				// pool, otherwise we might end up in a situation where results are returned to
+				// pool even though allCompletionFn still accesses results.
+				allCompletionWg.Wait()
```
robskillington: We call allCompletionFn inline just above, can we keep a bool seeing if we've executed it and if so call this instead?

PR #124, opened 2016-10-11
Add log for topology change

cw9:


client/session.go
```diff
@@ -361,6 +362,8 @@ func (s *session) setTopologyWithLock(topologyMap topology.Map, queues []hostQue
 			prev[i].Close()
 		}
 	}()
+
+	s.log.Infof("successully updated topology to %v hosts", topologyMap.HostsLen())
```
xichen2020: `successully` -> `successfully`, also while you are there, might change `%v` to `%d`?

cw9: Sure, will do

PR #127, opened 2016-10-13
Add instrumentation for fetch bootstrap blocks procedure

robskillington:
cc @xichen2020 @martin-mao @kobolog

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -112,7 +112,7 @@ type session struct {
 
 type sessionMetrics struct {
 	sync.RWMutex
-	streamFromPeersMetrics map[uint32]*streamFromPeersMetrics
+	streamFromPeersMetrics map[uint32]streamFromPeersMetrics
 }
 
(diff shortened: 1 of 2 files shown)


client/session.go
```diff
@@ -106,6 +107,22 @@ type session struct {
 	streamBlocksBatchSize            int
 	streamBlocksMetadataBatchTimeout time.Duration
 	streamBlocksBatchTimeout         time.Duration
+	metrics                          sessionMetrics
+}
+
+type sessionMetrics struct {
+	sync.RWMutex
+	streamFromPeersMetrics map[uint32]*streamFromPeersMetrics
```
xichen2020: Might be more memory efficient to use a value instead of a pointer for the metrics

xichen2020:
Other than the minor comment, LGTM

client/session.go
robskillington: Perhaps I'll store them by value but return a reference to them.

PR #128, opened 2016-10-13
Check if we need to seal blocks more efficiently

xichen2020:
cc @robskillington @martin-mao @kobolog

### DIFF
--- a/storage/block/types.go
+++ b/storage/block/types.go
@@ -138,8 +138,8 @@ type DatabaseSeriesBlocks interface {
 	// IsSealed returns whether all the blocks are sealed
 	IsSealed() bool
 
-	// MarkSealed marks the series as sealed
-	MarkSealed()
+	// Seal seals all the blocks
+	Seal()
 
 	// AddBlock adds a data block.
 	AddBlock(block DatabaseBlock)
--- a/storage/block/types.go
+++ b/storage/block/types.go
@@ -135,6 +135,12 @@ type DatabaseSeriesBlocks interface {
 	// Len returns the number of blocks contained in the collection.
 	Len() int
 
+	// IsSealed returns whether all the blocks are sealed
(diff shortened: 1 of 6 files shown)


storage/block/types.go
```diff
@@ -135,6 +135,12 @@ type DatabaseSeriesBlocks interface {
 	// Len returns the number of blocks contained in the collection.
 	Len() int
 
+	// IsSealed returns whether all the blocks are sealed
+	IsSealed() bool
+
+	// MarkSealed marks the series as sealed
+	MarkSealed()
```
robskillington: I'd prefer if we had a method that actually sealed the unsealed rather than someone on the outside sealing the unsealed then telling the blocks collection that they are now sealed.  Seems like better encapsulation.

Perhaps a method `SealAllUnsealed()`?

xichen2020: yeah sounds good

storage/series/series.go
```diff
@@ -161,22 +152,29 @@ func (s *dbSeries) shouldExpire(now, blockStart time.Time) bool {
...
+			continue
 		} else if s.shouldSeal(now, blockStart, block) {
 			block.Seal()
+		} else {
+			allBlocksSealed = false
 		}
 	}
+	if allBlocksSealed {
+		s.blocks.Seal()
+	}
```
robskillington: For this method, what about:

```go 
now := s.opts.ClockOptions().NowFn()()
for blockStart, block := range s.blocks.AllBlocks() {
  if s.shouldExpire(now, blockStart) {
    s.blocks.RemoveBlockAt(blockStart)
    block.Close()
  }
}
s.blocks.Seal()
```

robskillington: Can't do this as discussed due to the `shouldSeal` method needing to be taken into account.

PR #130, opened 2016-10-14
Allocate IDs for CommitLog writes from a pool.

kobolog:
This PR introduces pooling to the ID cloning code that happens when we enqueue writes to the CommitLog. Due to the asynchronous nature of the CommitLog, we cannot reuse the IDs passed from the RPC code and using nested contexts and `DependsOn() ` machinery proved to be prohibitively expensive.

### DIFF
--- a/persist/fs/commitlog/commit_log.go
+++ b/persist/fs/commitlog/commit_log.go
@@ -104,9 +104,10 @@ type commitLogWrite struct {
 
 // NewCommitLog creates a new commit log
 func NewCommitLog(opts Options) CommitLog {
-	iops := opts.InstrumentOptions()
-	iops = iops.SetMetricsScope(iops.MetricsScope().SubScope("commitlog"))
-	scope := iops.MetricsScope()
+	iopts := opts.InstrumentOptions().SetMetricsScope(
+		opts.InstrumentOptions().MetricsScope().SubScope("commitlog"))
+	scope := iopts.MetricsScope()
+
 	commitLog := &commitLog{
 		opts:  opts,
 		nowFn: opts.ClockOptions().NowFn(),
@@ -120,7 +121,7 @@ func NewCommitLog(opts Options) CommitLog {
 			flushErrors: scope.Counter("writes.flush-errors"),
 			flushDone:   scope.Counter("writes.flush-done"),
 		},
-		log:                  iops.Logger(),
+		log:                  iopts.Logger(),
 		newCommitLogWriterFn: newCommitLogWriter,
 		writes:               make(chan commitLogWrite, opts.BacklogQueueSize()),
 		closeErr:             make(chan error),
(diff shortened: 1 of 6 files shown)


persist/fs/commitlog/commit_log.go
```diff
@@ -99,14 +99,16 @@ type commitLogWrite struct {
 	datapoint    ts.Datapoint
 	unit         xtime.Unit
 	annotation   ts.Annotation
+	ctx          context.Context
```
robskillington: Nit: maybe have this as the first field in the struct?  There's a linter rule that got recently added that complains about ctx not being passed as the first arg in methods, may as well put it at the top of structs too in case that is another rule that gets added...

persist/fs/commitlog/commit_log.go
```diff
@@ -187,19 +197,18 @@ func (l *commitLog) flushEvery(interval time.Duration) {
...
 			// Flushed already recently, sleep until we would next consider flushing
-			sleepForOverride = interval - sinceLastFlush
+			sleepForOverride = interval - sinceFlush
 			continue
 		}
 
 		// Request a flush
-		l.RLock()
-		if l.closed {
+		if l.RLock(); l.closed {
```
robskillington: Any reason to combine these?  Seems like `l.RLock()` is not really a part of the condition.

persist/fs/commitlog/commit_log.go
```diff
@@ -227,26 +238,32 @@ func (l *commitLog) write() {
 			continue
 		}
 
-		now := l.nowFn()
-		if !now.Before(l.writerExpireAt) {
+		if now = l.nowFn(); !now.Before(l.writerExpireAt) {
```
robskillington: Could this not stay declared inline with `if now := l.nowFn(); !now.Before(l.writerExpireAt) { ..`?  It seems now is not used outside of the if clause.

persist/fs/commitlog/commit_log.go
```diff
@@ -376,9 +396,7 @@ func (l *commitLog) WriteBehind(
 	unit xtime.Unit,
 	annotation ts.Annotation,
 ) error {
-	l.RLock()
-
-	if l.closed {
+	if l.RLock(); l.closed {
```
robskillington: Here too, seems `l.RLock()` is not pertinent to the condition, seems it would belong better to the lexical scope above.

persist/fs/commitlog/commit_log.go
```diff
@@ -398,12 +417,12 @@ func (l *commitLog) WriteBehind(
 	default:
 	}
 
+	l.RUnlock()
```
robskillington: Ah, good placement 👍

persist/fs/commitlog/commit_log.go
```diff
@@ -412,14 +431,15 @@ func (l *commitLog) Iter() (Iterator, error) {
 }
 
 func (l *commitLog) Close() error {
-	l.Lock()
-	if l.closed {
+	if l.Lock(); l.closed {
```
robskillington: Yeah, think `l.Lock()` belongs in the upper lexical scope here.

persist/fs/commitlog/commit_log_test.go
```diff
@@ -267,7 +266,8 @@ func writeCommitLogs(
 			defer wg.Done()
 			datapoint := ts.Datapoint{Timestamp: write.t, Value: write.v}
 
-			err := writeFn(ctx, write.series, datapoint, write.u, write.a)
+			err := writeFn(context.NewContext(), write.series, datapoint,
```
robskillington: Is this created context ever closed?

persist/fs/commitlog/commit_log.go
kobolog: I've put it here since it sort of belongs with `completionFn`, but if there's a new lint rule, then yeah makes sense to move it.

PR #134, opened 2016-10-18
Merge blocks from peers with different or nil checksums during peer bootstrapping

xichen2020:

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -32,6 +32,7 @@ import (
 	"time"
 
 	"github.com/m3db/m3db/clock"
+	"github.com/m3db/m3db/context"
 	"github.com/m3db/m3db/encoding"
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	"github.com/m3db/m3db/network/server/tchannelthrift/convert"
@@ -1759,14 +1760,18 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
 	// No longer need the encoder, seal the block
 	result.Seal()
 
-	resultReader, err := result.Stream(nil)
+	resultCtx := context.NewContext()
+	defer resultCtx.Close()
+
+	resultReader, err := result.Stream(resultCtx)
 	if err != nil {
 		return err
 	}
 	if resultReader == nil {
 		return nil
 	}
-	defer resultReader.Close()
+
+	var tmpCtx context.Context
 
 	for {
 		r.Lock()
@@ -1784,7 +1789,13 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
 
 		// If we've already received data for this block, merge them
 		// with the new block if possible
-		currReader, err := currBlock.Stream(nil)
+		if tmpCtx == nil {
+			tmpCtx = context.NewContext()
+		} else {
+			tmpCtx.Reset()
+		}
+
+		currReader, err := currBlock.Stream(tmpCtx)
 		if err != nil {
 			return err
 		}
@@ -1797,7 +1808,8 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
 
 		readers := []io.Reader{currReader, resultReader}
 		encoder, err := r.mergeReaders(start, readers)
-		currReader.Close()
+		tmpCtx.BlockingClose()
+
 		if err != nil {
 			return err
 		}
--- a/client/session.go
+++ b/client/session.go
@@ -1403,79 +1403,113 @@ func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
 			continue
 		}
 
-		// Determine maximum size
-		currMaxSize := int64(math.MinInt64)
+		var (
+			sameNonNilChecksum = true
+			curChecksum        *uint32
+		)
+
 		for i := 0; i < len(currEligible); i++ {
 			unselected := currEligible[i].unselectedBlocks()
-			// Track maximum size
-			if unselected[0].size > currMaxSize {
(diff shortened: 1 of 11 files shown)


client/session.go
```diff
@@ -1716,13 +1738,88 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
 	// No longer need the encoder, seal the block
 	result.Seal()
 
-	r.Lock()
-	r.result.AddBlock(id, result)
-	r.Unlock()
+	resultCtx := context.NewContext()
```
robskillington: Get this from a pool too?

client/session.go
```diff
@@ -1716,13 +1738,88 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
...
+		result.Reset(start, encoder)
+		result.Seal()
+	}
 
 	return nil
 }
 
+func (r *blocksResult) mergeReaders(start time.Time, readers []io.Reader) (encoding.Encoder, error) {
+	alloc := r.opts.ReaderIteratorAllocate()
+	iter := encoding.NewMultiReaderIterator(alloc, nil)
```
robskillington: We should use the multi reader iterator pool to get this here I think?

client/session.go
```diff
@@ -1716,13 +1738,88 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
...
+	iter.Reset(readers)
+	defer iter.Close()
+
+	encoder := r.encoderPool.Get()
+	encoder.Reset(start, r.blockAllocSize)
+
+	for iter.Next() {
+		dp, unit, annotation := iter.Current()
+		encoder.Encode(dp, unit, annotation)
```
robskillington: We should probably check the return of `Encode(...)` and return if err here?

integration/peers_bootstrap_merge_peer_blocks_test.go
```diff
@@ -0,0 +1,121 @@
...
+	if testing.Short() {
+		t.SkipNow() // Just skip if we're doing a short run
+	}
+
+	// Test setups
+	log := xlog.NewLogger(os.Stdout)
+	namesp := namespace.NewMetadata(testNamespaces[0], namespace.NewOptions())
+	opts := newTestOptions().
+		SetNamespaces([]namespace.Metadata{namesp})
+	testSleepFn, setSleepFn := newTestSleepFn()
```
xichen2020: sounds good

client/session.go
```diff
@@ -1716,13 +1738,88 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
...
+
+		// Remove the existing block from the result so it doesn't get
+		// merged again
+		r.result.RemoveBlockAt(id, start)
+		r.Unlock()
+
+		// If we've already received data for this block, merge them
+		// with the new block if possible
+		if tmpCtx == nil {
+			tmpCtx = context.NewContext()
```
xichen2020: will do

client/session.go
xichen2020: yeah

client/session.go
xichen2020: yeah

client/session.go
xichen2020: will do, took this from existing code so didn't pay too much attention there

client/session.go
```diff
@@ -1403,79 +1404,113 @@ func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
...
-					append(currStart[i].blocks[idx].reattempt.attempted, peer)
 				currStart[i].blocks[idx].reattempt.peersMetadata = peersMetadata
 
-				// Leave the block in the current peers blocks list
-				continue
+				// Each block now has attempted multiple peers
+				for j := range currStart {
+					currStart[i].blocks[idx].reattempt.attempt++
+					currStart[i].blocks[idx].reattempt.attempted =
+						append(currStart[i].blocks[idx].reattempt.attempted, currStart[j].peer)
```
robskillington: At some point we should rethink our retry semantics here.  If any of these fail they won't be retried against the peer we want them from. 

Let me open an issue about it.  For now this is fine, but we probably want real retry semantics in the future for trying to retrieve from the same peer on failure.

client/session.go
```diff
@@ -1403,79 +1404,113 @@ func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
...
-				currStart[i].blocks[idx].reattempt.attempt++
-				currStart[i].blocks[idx].reattempt.attempted =
-					append(currStart[i].blocks[idx].reattempt.attempted, peer)
 				currStart[i].blocks[idx].reattempt.peersMetadata = peersMetadata
 
-				// Leave the block in the current peers blocks list
-				continue
+				// Each block now has attempted multiple peers
+				for j := range currStart {
+					currStart[i].blocks[idx].reattempt.attempt++
```
robskillington: Need to set `reattempt.id` here too

client/session.go
```diff
@@ -1403,79 +1404,113 @@ func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
...
+			peersMetadata := make([]blockMetadataReattemptPeerMetadata, 0, len(currStart))
+			for i := range currStart {
+				unselected := currStart[i].unselectedBlocks()
+				metadata := blockMetadataReattemptPeerMetadata{
+					peer:     currStart[i].peer,
+					start:    unselected[0].start,
+					size:     unselected[0].size,
+					checksum: unselected[0].checksum,
+				}
+				peersMetadata = append(peersMetadata, metadata)
```
xichen2020: Sure thing

PR #140, opened 2016-10-21
Bound the number of processors when adding results from fetching blocks

robskillington:


client/session.go
```diff
@@ -160,6 +161,11 @@ func newSession(opts Options) (clientSession, error) {
 		s.streamBlocksWorkers.Init()
 		s.streamBlocksReattemptWorkers = pool.NewWorkerPool(opts.FetchSeriesBlocksBatchConcurrency())
 		s.streamBlocksReattemptWorkers.Init()
+		processors := opts.FetchSeriesBlocksResultsProcessors()
+		s.streamBlocksResultsProcessors = make(chan struct{}, processors)
+		for i := 0; i < processors; i++ {
```
xichen2020: As discussed, would be good to add some comments here explaining why the channel instead of a separate worker pool

PR #142, opened 2016-10-22
Add tick metrics to measure series and blocks

robskillington:


storage/series/series.go
```diff
@@ -100,33 +105,37 @@ func (s *dbSeries) ID() ts.ID {
...
 	}
 
 	s.Lock()
 	if needsDrain {
 		s.buffer.DrainAndReset()
 	}
-
 	if needsBlockUpdate {
-		s.updateBlocksWithLock()
+		update := s.updateBlocksWithLock()
```
xichen2020: `updateResult` might be a better name

PR #147, opened 2016-10-23
Add read level consistency of unstrict majority and error metadata

robskillington:
cc @xichen2020 @martin-mao @cw9 @kobolog 

This introduces the concept of "unstrict majority" for reads such that applications can use to always attempt to achieve majority but in the case when all nodes have answered and less than the majority of nodes have responded successfully to go ahead nonetheless and return with a successful result.  

This means that if two nodes are healthy the application will always read from the majority of nodes to ensure data is merged from both healthy nodes to read what you put in at consistency level majority.  It does mean in an extreme failure case of two nodes being unhealthy that the application can continue to serve reads knowingly with no guarentees of serving partial reads.

### DIFF
--- a/client/errors.go
+++ b/client/errors.go
@@ -21,11 +21,13 @@
 package client
 
 import (
+	"fmt"
+
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	"github.com/m3db/m3x/errors"
 )
 
-// IsInternalServerError determines if the error is an internal server error.
+// IsInternalServerError determines if the error is an internal server error
 func IsInternalServerError(err error) bool {
 	errType := rpc.ErrorType_INTERNAL_ERROR
 	for err != nil {
@@ -37,7 +39,7 @@ func IsInternalServerError(err error) bool {
 	return false
 }
 
-// IsBadRequestError determines if the error is a bad request error.
+// IsBadRequestError determines if the error is a bad request error
 func IsBadRequestError(err error) bool {
 	errType := rpc.ErrorType_BAD_REQUEST
 	for err != nil {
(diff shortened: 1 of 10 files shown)


client/options.go
```diff
@@ -178,6 +182,7 @@ func newOptions() *options {
 		clockOpts:                             clock.NewOptions(),
 		instrumentOpts:                        instrument.NewOptions(),
 		consistencyLevel:                      defaultConsistencyLevel,
```
xichen2020: Should we call this `writeConsistencyLevel` assuming it's used for write consistency purposes?

client/types.go
```diff
@@ -59,6 +59,57 @@ const (
...
+	case ConnectConsistencyLevelNone:
+		return "ConnectConsistencyLevelNone"
+	case ConnectConsistencyLevelOne:
+		return "ConnectConsistencyLevelOne"
+	case ConnectConsistencyLevelMajority:
+		return "ConnectConsistencyLevelMajority"
+	case ConnectConsistencyLevelAll:
+		return "ConnectConsistencyLevelAll"
+	}
+	return ""
```
xichen2020: Might be better to return "ConnectConsistencyLevelUnknown" so it's easier to debug?

client/types.go
```diff
@@ -59,6 +59,57 @@ const (
...
+	case ReadConsistencyLevelOne:
+		return "ReadConsistencyLevelOne"
+	case ReadConsistencyLevelUnstrictMajority:
+		return "ReadConsistencyLevelUnstrictMajority"
+	case ReadConsistencyLevelMajority:
+		return "ReadConsistencyLevelMajority"
+	case ReadConsistencyLevelAll:
+		return "ReadConsistencyLevelAll"
+	}
+	return ""
```
xichen2020: Might be better to return "ReadConsistencyLevelUnknown" so it's easier to debug?

kobolog:
As a sidenote, I suggest that we start filling in PR description & details apart from CC'ing people to review it. While we can just talk to each other to figure out what's this about, I'd expect new team members and potential external users to be at least confused, most likely end up ignoring PRs and even feel excluded.

client/errors.go
```diff
@@ -48,3 +50,96 @@ func IsBadRequestError(err error) bool {
...
+}
+
+func (e consistencyResultErr) Error() string {
+	return fmt.Sprintf(
+		"failed to meet %s with %d/%d success, %d nodes responded, errors: %v",
+		e.level.String(), e.success, e.enqueued, e.responded, e.errs)
+}
+
+func (e consistencyResultErr) numResponded() int {
+	return int(e.responded)
```
robskillington: Ah yeah, these used to be int32's.  Ta

client/types.go
robskillington: Yeah, sounds good.

robskillington:
Added a PR description.  Agreed @kobolog.

PR #148, opened 2016-10-26
Add metrics for how many nodes are failing for write/fetch requests

robskillington:
cc @xichen2020 @martin-mao @cw9 @kobolog 

This change adds metrics for how many nodes are responding with errors.  It also allows for callers of client methods to determine which shard an ID belongs to so they can drill down into which shards are likely affected during times when write/read operations begin failing frequently (likely due to nodes responding with success not meeting consistency requirements).

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -115,7 +115,23 @@ type session struct {
 
 type sessionMetrics struct {
 	sync.RWMutex
-	streamFromPeersMetrics map[uint32]streamFromPeersMetrics
+	writeSuccess               tally.Counter
+	writeErrors                tally.Counter
+	writeNodesRespondingErrors []tally.Counter
+	fetchSuccess               tally.Counter
+	fetchErrors                tally.Counter
+	fetchNodesRespondingErrors []tally.Counter
+	streamFromPeersMetrics     map[uint32]streamFromPeersMetrics
+}
+
+func newSessionMetrics(scope tally.Scope) sessionMetrics {
+	return sessionMetrics{
+		writeSuccess:           scope.Counter("write.success"),
+		writeErrors:            scope.Counter("write.error"),
+		fetchSuccess:           scope.Counter("fetch.success"),
+		fetchErrors:            scope.Counter("fetch.error"),
+		streamFromPeersMetrics: make(map[uint32]streamFromPeersMetrics),
+	}
 }
 
 type streamFromPeersMetrics struct {
@@ -141,9 +157,11 @@ func newSession(opts Options) (clientSession, error) {
 		return nil, err
 	}
 
+	scope := opts.InstrumentOptions().MetricsScope()
+
 	s := &session{
 		opts:                 opts,
-		scope:                opts.InstrumentOptions().MetricsScope(),
+		scope:                scope,
 		nowFn:                opts.ClockOptions().NowFn(),
 		log:                  opts.InstrumentOptions().Logger(),
 		writeLevel:           opts.WriteConsistencyLevel(),
@@ -152,9 +170,7 @@ func newSession(opts Options) (clientSession, error) {
 		topo:                 topo,
 		fetchBatchSize:       opts.FetchBatchSize(),
 		newPeerBlocksQueueFn: newPeerBlocksQueue,
-		metrics: sessionMetrics{
-			streamFromPeersMetrics: make(map[uint32]streamFromPeersMetrics),
-		},
+		metrics:              newSessionMetrics(scope),
 	}
 
 	if opts, ok := opts.(AdminOptions); ok {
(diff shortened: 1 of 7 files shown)


x/metrics/test_reporter.go
```diff
@@ -111,23 +151,85 @@ func (r *testStatsReporter) Timers() map[string][]time.Duration {
...
+	name       string
+	tags       map[string]string
+	value      int64
+	timerValue time.Duration
+}
+
+type eventType int
+
+const (
+	eventTypeCount eventType = iota
```
xichen2020: `eventTypeCounter` might be more consistent?

x/metrics/test_reporter.go
```diff
@@ -111,23 +151,85 @@ func (r *testStatsReporter) Timers() map[string][]time.Duration {
...
+
+func (e *event) Name() string {
+	return e.name
+}
+
+func (e *event) Tags() map[string]string {
+	return e.tags
+}
+
+func (e *event) IsCount() bool {
```
xichen2020: `IsCounter` might be more consistent?

x/metrics/test_reporter.go
```diff
@@ -34,6 +34,18 @@ type TestStatsReporter interface {
...
 	Gauges() map[string]int64
 	Timers() map[string][]time.Duration
+	Events() []TestStatsReporterEvent
+}
+
+// TestStatsReporterEvent is an event that was reported to the test reporter
+type TestStatsReporterEvent interface {
+	Name() string
+	Tags() map[string]string
+	IsCount() bool
```
xichen2020: `IsCounter` might be more consistent

client/session.go
```diff
@@ -115,7 +115,23 @@ type session struct {
...
+	fetchSuccess               tally.Counter
+	fetchErrors                tally.Counter
+	fetchNodesRespondingErrors []tally.Counter
+	streamFromPeersMetrics     map[uint32]streamFromPeersMetrics
+}
+
+func newSessionMetrics(scope tally.Scope) sessionMetrics {
+	return sessionMetrics{
+		writeSuccess:           scope.Counter("write.success"),
+		writeErrors:            scope.Counter("write.error"),
```
xichen2020: `write.errors` is probably more consistent with metrics naming in other places

client/session.go
```diff
@@ -115,7 +115,23 @@ type session struct {
...
+		fetchSuccess:           scope.Counter("fetch.success"),
+		fetchErrors:            scope.Counter("fetch.error"),
```
xichen2020: write.errors is probably more consistent with metrics naming in other places

client/session.go
```diff
@@ -429,6 +489,24 @@ func (s *session) setTopologyWithLock(topologyMap topology.Map, queues []hostQue
 		s.multiReaderIteratorPool = encoding.NewMultiReaderIteratorPool(poolOpts)
 		s.multiReaderIteratorPool.Init(s.opts.ReaderIteratorAllocate())
 	}
+	if replicas > len(s.metrics.writeNodesRespondingErrors) {
+		curr := len(s.metrics.writeNodesRespondingErrors)
+		for i := curr; i < replicas; i++ {
+			tags := map[string]string{"nodes": fmt.Sprintf("%d", curr+1)}
```
xichen2020: Should it be `i+1` instead of `curr+1`?

client/session.go
```diff
@@ -429,6 +489,24 @@ func (s *session) setTopologyWithLock(topologyMap topology.Map, queues []hostQue
...
+			counter := s.scope.Tagged(tags).Counter("write.nodes-responding-error")
+			s.metrics.writeNodesRespondingErrors =
+				append(s.metrics.writeNodesRespondingErrors, counter)
+		}
+	}
+	if replicas > len(s.metrics.fetchNodesRespondingErrors) {
+		curr := len(s.metrics.fetchNodesRespondingErrors)
+		for i := curr; i < replicas; i++ {
+			tags := map[string]string{"nodes": fmt.Sprintf("%d", curr+1)}
```
xichen2020: Should it be `i+1` instead of `curr+1`?

client/session.go
```diff
@@ -218,6 +245,39 @@ func (s *session) streamFromPeersMetricsForShard(shard uint32) *streamFromPeersM
...
+	if consistencyResultErr == nil {
+		s.metrics.fetchSuccess.Inc(1)
+	} else {
+		s.metrics.fetchErrors.Inc(1)
+	}
+}
+
+func (s *session) nodesRespondingErrorsMetricIndex(respErrs int32) int32 {
+	idx := respErrs - 1
+	if respErrs > s.replicas {
```
xichen2020: Do we need a RLock here for reading `s.replicas`?

x/metrics/test_reporter.go
robskillington: Sure thing

client/session.go
robskillington: Yeah seems so, will rename

client/session.go
robskillington: Needs a loadint32, will do, thanks.

client/session.go
robskillington: Good catch, ta.

PR #149, opened 2016-10-26
Finer-grained error reporting

xichen2020:

Separately report number of retryable errors and number of non-retryable errors so we don't alert on invalid requests

### DIFF
--- a/instrument/methods.go
+++ b/instrument/methods.go
@@ -115,3 +115,45 @@ func NewMethodMetrics(scope tally.Scope, methodName string, samplingRate float64
 		SuccessLatency: newSampledTimer(scope.Timer(methodName+".success-latency"), samplingRate),
 	}
 }
+
+// BatchMethodMetrics is a bundle of common metrics for methods with batch semantics
+type BatchMethodMetrics struct {
+	RetryableErrors    tally.Counter
+	NonRetryableErrors tally.Counter
+	Success            tally.Counter
+	Latency            tally.Timer
+}
+
+// NewBatchMethodMetrics creates new batch method metrics
+func NewBatchMethodMetrics(
+	scope tally.Scope,
+	methodName string,
+	samplingRate float64,
+) BatchMethodMetrics {
+	return BatchMethodMetrics{
+		RetryableErrors:    scope.Counter(methodName + ".retryable-errors"),
+		NonRetryableErrors: scope.Counter(methodName + ".non-retryable-errors"),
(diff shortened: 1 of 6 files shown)


network/server/tchannelthrift/errors/errors.go
```diff
@@ -33,6 +33,16 @@ func newError(errType rpc.ErrorType, err error) *rpc.Error {
...
 
+// IsInternalError returns whether the error is an internal error
+func IsInternalError(err *rpc.Error) bool {
+	return err != nil && err.Type == rpc.ErrorType_INTERNAL_ERROR
+}
+
+// IsBadRequestError returns whether the error is a bad request error
+func IsBadRequestError(err *rpc.Error) bool {
+	return err != nil && err.Type == rpc.ErrorType_BAD_REQUEST
+}
```
robskillington: Perhaps update the `client/errors.go` methods to call these once they unpack an error that is a `*rpc.Error`?

instrument/methods.go
```diff
@@ -115,3 +115,45 @@ func NewMethodMetrics(scope tally.Scope, methodName string, samplingRate float64
...
+}
+
+// NewBatchMethodMetrics creates new batch method metrics
+func NewBatchMethodMetrics(
+	scope tally.Scope,
+	methodName string,
+	samplingRate float64,
+) BatchMethodMetrics {
+	return BatchMethodMetrics{
+		RetryableErrors:    scope.Counter(methodName + ".retryable-errors"),
```
robskillington: Perhaps also add just an ".errors" for backwards compatibility (and also just to get an indication of total errors too)?

xichen2020: sure thing

network/server/tchannelthrift/errors/errors.go
xichen2020: sure thing

instrument/methods.go
```diff
@@ -115,3 +115,45 @@ func NewMethodMetrics(scope tally.Scope, methodName string, samplingRate float64
...
+		NonRetryableErrors: scope.Counter(methodName + ".non-retryable-errors"),
+		Success:            scope.Counter(methodName + ".success"),
+		Latency:            newSampledTimer(scope.Timer(methodName+".latency"), samplingRate),
```
robskillington: As discussed, on second thoughts considering its a batch and can partially succeed or fail it's a little ambiguous.  One metric is fine here.

PR #151, opened 2016-10-27
Allow namespace-level retention period override

xichen2020:
cc @robskillington

### DIFF
--- a/storage/namespace/options.go
+++ b/storage/namespace/options.go
@@ -38,8 +38,9 @@ const (
 	// Namespace requires repair by default
 	defaultNeedsRepair = true
 
-	// Namespace retains data for 48 hours by default
-	defaultRetentionPeriod = 48 * time.Hour
+	// Namespace does not have retention period overrides and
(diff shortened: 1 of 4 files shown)


storage/namespace/options.go
```diff
@@ -35,6 +37,9 @@ const (
 
 	// Namespace requires repair by default
 	defaultNeedsRepair = true
+
+	// Namespace retains data for 48 hours by default
+	defaultRetentionPeriod = 48 * time.Hour
```
xichen2020: hmm can do, not sure which is more dangerous though..

kobolog:
I'd like to point out that a short simple PR description would help other M3 core engineers as well as potential external contributors to better understand the contribution process, the architecture of M3DB and improve the overall maintainability of the codebase.

xichen2020:
@kobolog 100% agreed, this was put together in a hurry and was intended as a stop-gap measure for Halloween but yeah for sure will add better PR descriptions going forward

Issue #153, opened 2016-10-31
Address data loss during block rotation when there are out of order writes 


xichen2020:
It appears when ingesters are retrying writes, sometimes the data in the same block during which the retries were happening are lost.

robskillington:
Fixed by #155

PR #155, opened 2016-11-01
Do not prematurely move center bucket until a drain and reset happens

xichen2020:

Currently we use timestamps to determine which bucket among all the buffer buckets a read or write belongs to, which in turn determines the order data in different buckets get read. However if we just crossed the block boundary and attempt to read data, the center bucket would have been shifted even though the old buckets have not been reset or drained, leading to out-of-order reads. This is what's causing sudden shadow query match rate drops at the beginning of a new block.

This PR uses `pastMostBucketIdx` to keep track of the oldest bucket so we always read data in the correct order. `pastMostBucketIdx` is updated when a drain or reset happens.

### DIFF
--- a/storage/series/buffer.go
+++ b/storage/series/buffer.go
@@ -52,13 +52,14 @@ const (
 )
 
 type dbBuffer struct {
-	opts         Options
-	nowFn        clock.NowFn
-	drainFn      databaseBufferDrainFn
-	buckets      [bucketsLen]dbBufferBucket
-	blockSize    time.Duration
-	bufferPast   time.Duration
-	bufferFuture time.Duration
+	opts              Options
+	nowFn             clock.NowFn
+	drainFn           databaseBufferDrainFn
+	pastMostBucketIdx int
+	buckets           [bucketsLen]dbBufferBucket
+	blockSize         time.Duration
+	bufferPast        time.Duration
+	bufferFuture      time.Duration
 }
 
 type databaseBufferDrainFn func(start time.Time, encoder encoding.Encoder)
@@ -77,7 +78,7 @@ func newDatabaseBuffer(drainFn databaseBufferDrainFn, opts Options) databaseBuff
 }
 
 func (b *dbBuffer) Reset() {
-	b.forEachBucketAsc(b.nowFn(), func(bucket *dbBufferBucket, start time.Time) {
+	b.pastMostBucketIdx = b.forEachBucketAscToDrainOrReset(b.nowFn(), func(bucket *dbBufferBucket, start time.Time) {
 		bucket.opts = b.opts
 		bucket.resetTo(start)
 	})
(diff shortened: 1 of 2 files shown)


xichen2020:
Fixes #153

storage/series/buffer.go
```diff
@@ -77,20 +78,20 @@ func newDatabaseBuffer(drainFn databaseBufferDrainFn, opts Options) databaseBuff
 }
 
 func (b *dbBuffer) Reset() {
-	b.forEachBucketAsc(b.nowFn(), func(bucket *dbBufferBucket, start time.Time) {
+	b.pastMostBucketIdx = b.forEachBucketAscToDrainOrReset(b.nowFn(), func(bucket *dbBufferBucket, start time.Time) {
```
robskillington: Perhaps perform the assignment of `pastMostBucketIdx` inside `forEachBucketAsc` based on a flag?  That way we just need one function not two.

Could even enum-ify this to make it more clean:

```go
type computeReadWriteBucketIdx int

const (
  readBucketIdx computeReadWriteBucketIdx = iota
  computeBucketIdx
  computeAndWriteBucketIdx
)

// ...
now := b.nowFn()
flags := computeAndWriteBucketIdx
b.forEachBucketAsc(now, flags, func(bucket *dbBufferBucket, start time.Time) {

})
```

storage/series/buffer.go
```diff
@@ -337,6 +321,22 @@ func (b *dbBufferBucket) resetTo(start time.Time) {
...
+}
+
+func (b *dbBufferBucket) needsReset(targetStart time.Time) bool {
+	return !b.start.Equal(targetStart)
+}
+
+func (b *dbBufferBucket) needsDrain(now time.Time, targetStart time.Time) bool {
+	retentionOpts := b.opts.RetentionOptions()
+	blockSize := retentionOpts.BlockSize()
+	bufferPast := retentionOpts.BufferPast()
```
robskillington: Hmm this might slow down the tick a little doing a little more indirection in this hot path.  Can we keep a reference to the `dbBuffer` then directly reference `b.buffer.blockSize` and `b.buffer.bufferPast`?

xichen2020: That adds 8 bytes per bucket, which can sum up to a few hundred MBs. These should be three fast memory accesses per bucket and likely cached, which should be on the order of tens of nanoseconds, which sums up to tens of milliseconds across all series. Since the tick always takes a minute and sleeps in between to make up the slack, this likely means we'll just sleep a bit less and have no real impact on the tick time.

robskillington: Do we need these in the `dbBuffer` then at all?

storage/series/buffer.go
robskillington: I agree with the lack of overloading functions to make it two functions, but can we make it more readable and clear by at least passing it as a typed enum? i.e. the 

```go
type computeReadWriteBucketIdx int

const (
  readBucketIdx computeReadWriteBucketIdx = iota
  computeBucketIdx
  computeAndWriteBucketIdx
)
```

Also can we call it `computedForEachBucketAsc` so it's more generic?

xichen2020: Sounds good

storage/series/buffer.go
robskillington: Looks like we do, no worries.  Let's leave them.

xichen2020: Unfortunately it's still used in `Write`, so probably need to leave them there

PR #156, opened 2016-11-01
Reduce heap escapes in session.Write().

kobolog:
Client-side runtime and heap profiles demonstrated that almost 50% of all allocations in m3dbingester happen in `session.Write()` due to inferior quality of Go's escape analysis algorithms. Essentially, the algorithm marks any variable that is captured in a lambda as heap-allocated (even if the lambda does not outlive the function scope). In our case, some lambdas in `session.Write()` might outlive the function scope, so almost all local variables are marked as escaping.

This PR explicitly recaptures all escaping variables in a pooled heap-allocated object, effectively reducing the number of heap allocations, live objects and the amount of time required to do a GC scan pass.

Here's a list of TODOs:

- [ ] ~Pooling of `ts.ID`.~
- [ ] ~Get rid of the last `RouteForEach()` lambda.~
- [x] Get rid of the `queues` slice operations.

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -101,6 +101,7 @@ type session struct {
 	multiReaderIteratorPool          encoding.MultiReaderIteratorPool
 	seriesIteratorPool               encoding.SeriesIteratorPool
 	seriesIteratorsPool              encoding.MutableSeriesIteratorsPool
+	writeMonitorPool                 sync.Pool
 	fetchBatchSize                   int
 	newPeerBlocksQueueFn             newPeerBlocksQueueFn
 	origin                           topology.Host
@@ -172,6 +173,12 @@ func newSession(opts Options) (clientSession, error) {
 		newPeerBlocksQueueFn: newPeerBlocksQueue,
 		metrics:              newSessionMetrics(scope),
 	}
+
+	s.writeMonitorPool = sync.Pool{New: func() interface{} {
+		c := &writeMonitor{session: s}
+		c.condition = &sync.Cond{L: c}
+		return c
+	}}
 
 	if opts, ok := opts.(AdminOptions); ok {
 		s.origin = opts.Origin()
@@ -548,21 +555,73 @@ func (s *session) newHostQueue(host topology.Host, topologyMap topology.Map) hos
 	return hostQueue
 }
 
+type writeMonitor struct {
+	sync.Mutex
+
+	session             *session
+	condition           *sync.Cond
+	op                  *writeOp
+	majority            int32
+	successful, pending int32
+	errors              []error
+}
+
+func (w *writeMonitor) reset(op *writeOp, majority int32) {
+	w.op, w.majority, w.successful, w.pending = op, majority, 0, 0
+	w.errors = w.errors[:]
+}
+
+func (w *writeMonitor) completionFn(result interface{}, err error) {
+	var (
+		successful int32
+		remaining  = atomic.AddInt32(&w.pending, -1)
+	)
+
+	if err != nil {
+		w.Lock()
+		w.errors = append(w.errors, err)
+		w.Unlock()
+	} else {
+		successful = atomic.AddInt32(&w.successful, 1)
+	}
+
+	switch w.session.writeLevel {
+	case topology.ConsistencyLevelOne:
+		if successful == 0 && remaining != 0 {
+			return
+		}
+	case topology.ConsistencyLevelMajority:
+		if successful < w.majority && remaining != 0 {
+			return
+		}
+	case topology.ConsistencyLevelAll:
+		if remaining != 0 {
+			return
+		}
+	}
+
+	w.Lock()
+	w.condition.Signal()
+	w.Unlock()
+
+	// Finalize resources that can be returned
+	if remaining == 0 {
+		w.release()
+	}
+}
+
+func (w *writeMonitor) release() {
+	w.session.writeOpPool.Put(w.op)
+	w.op = nil
+	w.session.writeMonitorPool.Put(w)
+}
+
 func (s *session) Write(namespace, id string, t time.Time, value float64, unit xtime.Unit, annotation []byte) error {
 	var (
-		wg            sync.WaitGroup
-		wgIsDone      int32
-		enqueueErr    error
-		enqueued      int32
-		pending       int32
-		resultErrLock sync.RWMutex
-		resultErrs    int32
-		errors        []error
-		majority      int32
-		success       int32
-		tsID          = ts.StringID(id)
+		enqueued int32
+		majority = atomic.LoadInt32(&s.majority)
+		tsID     = ts.StringID(id)
 	)
-	wg.Add(1)
 
 	timeType, timeTypeErr := convert.ToTimeType(unit)
 	if timeTypeErr != nil {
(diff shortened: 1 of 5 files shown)


kobolog:
Here's a list of TODOs:

- [ ] Pooling of `ts.ID`.
- [ ] Get rid of the last `RouteForEach()` lambda.
- [ ] Get rid of the `queues` slice operations.

client/session.go
```diff
@@ -574,99 +651,67 @@ func (s *session) Write(namespace, id string, t time.Time, value float64, unit x
...
-	if routeErr != nil {
-		return routeErr
-	}
+	c.Lock()
 
-	if enqueueErr != nil {
-		s.opts.InstrumentOptions().Logger().Errorf("failed to enqueue write: %v", enqueueErr)
-		return enqueueErr
+	for i := range c.queues {
+		if err := c.queues[i].Enqueue(c.op); err != nil {
```
xichen2020: Do we want to continue enqueuing to other queues if we encountered an error like we did before?

client/session.go
```diff
@@ -574,99 +651,67 @@ func (s *session) Write(namespace, id string, t time.Time, value float64, unit x
...
+		}
+
+		c.incref()
+		enqueued++
 	}
 
 	// Wait for writes to complete
-	wg.Wait()
+	s.RUnlock()
+	c.condition.Wait()
```
xichen2020: Conventionally we should surround the `.Wait` call with while/for loops that checks if a condition we are waiting for has been satisfied in order to deal with spurious wakeups: https://en.wikipedia.org/wiki/Spurious_wakeup. This is also recommended in the comments for the `Wait` function: https://golang.org/src/sync/cond.go?s=1407:1428#L44.

Without doing so, I worry that the thread gets waken even though the condition is not satisfied, causing the wait to terminate prematurely.

Any particular reason why we replace a waitgroup with a condition variable here?

client/session.go
kobolog: That looked like a bug to me, so I dropped it. We were enqueueing to other queues and then returning an error anyway. What's the rationale?

xichen2020: hm yeah, that's there probably to catch code bugs but in reality it should never happen, so probably okay to drop that.

client/session.go
xichen2020: Interesting, I wonder how Go managed to distinguish spurious wakeups from real notifications and make the right decision as to whether to return or continue waiting. Would probably require decent amount of load testing since I don't believe we've used condition variables in prod settings with high load so if for some reason spurious wakeups does happen, it would be really difficult to find.

Alternatively we could replace the `wgIsDone` atomic guard with `sync.Once` so it's cleaner and doesn't require keeping track of how many times `Done` is called. Seems less intrusive that way.

kobolog: Spurious wake ups don't happen because Golang's condvars has nothing to do with pthreads' condvars. According to Golang authors, `It's built on Semacquire, which does not permit spurious interrupts`.

client/session.go
kobolog: I'm fine with putting it back, it just doesn't make much sense to me. What kind of bugs is it supposed to catch?

xichen2020: e.g., the queue is closed for some reason when trying to enqueue. It should never happen unless there is a code bug around how/when we lock the queues.

kobolog: Yeah, I understand that part – but why do we want to continue enqueueing if we already know that one of the queue invariants doesn't hold anymore?

xichen2020: Yeah I think we had it there with the intention that if we had a bug somewhere that caused the queues to be closed before/during enqueuing, we'd be able to see it in the logs, which would make it easier to debug issues. Don't think that's necessary now that we've done extensive testing on the write path so I'm fine with taking that part out.

client/session.go
```diff
@@ -574,99 +651,68 @@ func (s *session) Write(namespace, id string, t time.Time, value float64, unit x
...
 
-	s.RLock()
-	if s.state != stateOpen {
+	if s.RLock(); s.state != stateOpen {
 		s.RUnlock()
 		return errSessionStateNotOpen
 	}
 
-	majority = atomic.LoadInt32(&s.majority)
+	c := s.writeStatePool.Get().(*writeState)
```
robskillington: Can we call this something like `write`?  Reading code below was hard to know what `c` was.

client/session.go
```diff
@@ -548,21 +556,90 @@ func (s *session) newHostQueue(host topology.Host, topologyMap topology.Map) hos
...
+
+func (r *rc) decref() {
+	if v := atomic.AddInt32(&r.n, -1); v == 0 {
+		r.d()
+	} else if v < 0 {
+		panic("invalid rc state")
+	}
+}
+
+type writeState struct {
```
robskillington: Can we move `writeState` and `rc` subtypes into another file?  I'll put up a PR soon to start pulling other subtypes from `session.go ` into their own files.  This file is getting too monolithic.

client/session.go
```diff
@@ -548,21 +556,90 @@ func (s *session) newHostQueue(host topology.Host, topologyMap topology.Map) hos
...
+		if successful >= w.majority || remaining == 0 {
+			w.condition.Signal()
+		}
+	case topology.ConsistencyLevelAll:
+		if remaining == 0 {
+			w.condition.Signal()
+		}
+	}
+
+	w.Unlock()
```
kobolog: This locks the WriteState object, which is not contended in 99% case.

client/session.go
kobolog: Oh, you mean the actual variable. Sure!

client/session.go
```diff
@@ -574,99 +651,68 @@ func (s *session) Write(namespace, id string, t time.Time, value float64, unit x
...
-	s.incWriteMetrics(err, errsLen)
+
+	// Wait for writes to complete. We don't need to loop over Wait() since there
+	// are no spurious wakeups in Golang and the condition is one-way and binary.
+	s.RUnlock()
+	c.condition.Wait()
+
+	err := s.writeConsistencyResult(majority,
+		enqueued, enqueued-atomic.LoadInt32(&c.pending), int32(len(c.errors)), c.errors)
+	s.incWriteMetrics(err, int32(len(c.errors)))
```
robskillington: Maybe we should add two methods to `writeState` - namely `Errors() []error` and `Pending() int32` that encapsulate perform atomic.Load, etc better separation of concerns?  i.e. I don't need to know how `writeState` is implemented in this `Write(...)` call if these are encapsulated, so less cognitive load when reading this method.

client/session.go
robskillington: I see I see, no worries, I withdraw my recommendation =]

client/session.go
```diff
@@ -173,6 +174,13 @@ func newSession(opts Options) (clientSession, error) {
 		metrics:              newSessionMetrics(scope),
 	}
 
+	s.writeStatePool = sync.Pool{New: func() interface{} {
+		c := &writeState{session: s}
+		c.d, c.condition.L = c.reset, c
+
+		return c
+	}}
```
kobolog: I'd rather not since we're struggling through object count issues now.

PR #157, opened 2016-11-02
Disable throughput limiting during bootstrap

xichen2020:

Disable rate limiting during bootstrap to speed up the bootstrap process since we flush data to disk at the end of the bootstrap process, which could take quite a while if this is a new node that just joined the cluster and bootstraps all its data from peers and we rate limit the flush process.


persist/composite/manager.go
```diff
@@ -1,183 +0,0 @@
-// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: We no longer need the composite manager?

persist/fs/manager.go
```diff
@@ -37,16 +37,15 @@ type sleepFn func(time.Duration)
...
-	throughputLimitMbps     float64
-	nowFn                   clock.NowFn
-	sleepFn                 sleepFn
-	writer                  FileSetWriter
-	start                   time.Time
-	lastCheck               time.Time
-	bytesWritten            int64
+	opts                Options
+	filePathPrefix      string
+	throughputLimitOpts persist.ThroughputLimitOptions
```
robskillington: I wonder if we should put `ThroughputLimitOptions` into `x/io` so that when we need to use something very similar for network throughput limiting we can reuse it?

persist/composite/manager.go
xichen2020: No it's not, and probably won't ever be. It's there because at one point we thought we were gonna dual write to lcoal fs and C*, but that turned out unnecessary, so now it just becomes a burden to maintain every time we need to change the `Manger` interface

PR #159, opened 2016-11-04
GC-opaque heap & object pool.

kobolog:
This PR introduces a new object pool type which is opaque to the GC. The rationale behind it is that we accumulate an enourmous number of active objects on the GC heap and eventually GC starts to struggle when scanning all of them to determine if any of these objects should be deallocated. In reality, none of these objects would ever be deallocated since they are all pooled, so GC is essentially wasting CPU. The general observation is that background GC pause time is linear with the size of the heap, so reducing its size also reduces the width of CPU consumption peaks when background GC kicks in.

The `NativePool` works around this issue by hiding pooled objects from the GC – from GC's PoV there's only one giant slice of `uint8`s. Each time a new object is requested, a new `interface{}` is allocated on the heap which points to some area inside that slice. When an object is returned, the only thing that is remembered about it is its offset into that slice.

Unfortunately, objects which involve indirection – slices, maps, interfaces, pointer fields and so on – won't work with this pool **in general case** since GC will see these pointers as unreachable because their owners are hidden from it.

- Slices and maps won't work since we don't have control over their internal pointers. For example, even though it's possible to manually construct a slice with `reflect.SliceHeader` and point its storage to an area from `NativePool`, it will work only as long as no internal storage reallocation happens.
- It is only possible to use pointer and interface members if the pointee is also allocated from an `NativePool`.

This PR also introduces a `NativeHeap` which is an elastic GC-opaque []byte allocator. The `NativeHeap` is based on `NativePool` to provide underlying storage primitives.

### DIFF
--- a/pool/heap.go
+++ b/pool/heap.go
@@ -28,6 +28,7 @@ import (
 	"time"
 	"unsafe"
 
+	"github.com/m3db/m3x/log"
 	"github.com/uber-go/tally"
 )
 
@@ -38,13 +39,13 @@ func NewNativeHeap(b []Bucket, po ObjectPoolOptions) BytesPool {
 	}
 
 	var (
-		m        = po.MetricsScope()
+		m        = po.InstrumentOptions().MetricsScope()
 		ByteType = reflect.TypeOf((byte)(0))
 	)
 
 	sort.Sort(BucketByCapacity(b))
 
-	h := heap{m: heapMetrics{
+	h := heap{l: po.InstrumentOptions().Logger(), m: heapMetrics{
 		overflows: m.Counter("overflows"),
 		misplaces: m.Counter("misplaces"),
 	}}
@@ -71,6 +72,7 @@ func NewNativeHeap(b []Bucket, po ObjectPoolOptions) BytesPool {
 type heap struct {
 	slots []*slot
 
+	l xlog.Logger
 	m heapMetrics
 }
 
@@ -205,11 +207,17 @@ func (p heap) Get(n int) []byte {
 	p.m.overflows.Inc(1)
 
 	// Allocate a segment directly from the system heap.
-	return mmap(n)[:0:n]
+	if r, err := mmap(n); err != nil {
+		panic("mmap() error: " + err.Error())
+	} else {
+		return r[0:0:n]
+	}
 }
 
 func (p heap) Put(head []byte) {
-	if p.pick(cap(head), func(s *slot) {
+	n := cap(head)
+
+	if p.pick(n, func(s *slot) {
 		s.put(unsafe.Pointer(
 			(*reflect.SliceHeader)(unsafe.Pointer(&head)).Data))
 	}) {
@@ -217,7 +225,8 @@ func (p heap) Put(head []byte) {
 	}
 
 	// Nothing fits so it must be a system heap segment.
-	if err := munmap(head[:cap(head)]); err != nil {
+	if err := munmap(head[0:n]); err != nil {
 		p.m.misplaces.Inc(1)
+		p.l.Warnf("misplaced segment: %p size=%d", &head[0], n)
 	}
 }
--- a/pool/heap.go
+++ b/pool/heap.go
@@ -1,0 +1,150 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package pool
+
+import (
+	"reflect"
+	"sort"
+	"sync"
+	"unsafe"
+)
+
+// NewNativeHeap constructs a new BytesPool based on NativePool.
+func NewNativeHeap(sizes []Bucket) BytesPool {
+	var (
+		slots = make([]*slot, len(sizes))
+		T     = reflect.TypeOf((byte)(0))
+	)
+
+	sort.Sort(BucketByCapacity(sizes))
+
+	for i, cfg := range sizes {
+		ns := &slot{class: cfg.Capacity, cfg: NativePoolOptions{
+			Size: uint(cfg.Count),
+			Type: reflect.ArrayOf(cfg.Capacity, T)}}
+		slots[i] = ns
+	}
+
+	return heap(slots)
+}
+
+type heap []*slot
+
+type slot struct {
+	sync.RWMutex
+
+	class int
+	cfg   NativePoolOptions
+	pools []NativePool
+}
+
+func (s *slot) get() interface{} {
+	if segment := s.getOr(s.RLocker(), func() interface{} {
+		return nil
+	}); segment != nil {
+		return segment
+	}
+
+	// Slow path - double-check that there are no segments left,
+	// then grow while holding an exclusive lock.
+	return s.getOr(s, func() interface{} {
+		s.pools = append(
+			[]NativePool{NewNativePool(s.cfg)}, s.pools...)
+		return s.pools[0].Get()
+	})
+}
+
+func (s *slot) getOr(l sync.Locker, fn OverflowFn) interface{} {
+	var segment interface{}
+
+	l.Lock()
+
+	for _, pool := range s.pools {
+		if segment = pool.GetOr(func() interface{} {
+			return nil
+		}); segment != nil {
+			l.Unlock()
+			return segment
+		}
+	}
+
+	segment = fn()
+
+	l.Unlock()
+
+	return segment
+}
+
+func (s *slot) put(segment interface{}) {
+	s.RLock()
+
+	for i := range s.pools {
+		if s.pools[i].Owns(segment) {
+			s.pools[i].Put(segment)
+			break
+		}
+	}
+
+	s.RUnlock()
+}
+
+func (p heap) Init() {
+	for _, slot := range p {
+		slot.pools = append(slot.pools, NewNativePool(slot.cfg))
+	}
+}
+
+func (p heap) pick(class int, action func(*slot)) bool {
+	for _, slot := range p {
+		if class <= slot.class {
+			action(slot)
+			return true
+		}
+	}
+
+	return false
+}
+
+func (p heap) Get(n int) []byte {
+	var head []byte
+
+	if !p.pick(n, func(slot *slot) {
+		head = *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
+			Data: reflect.ValueOf(slot.get()).Pointer(),
+			Len:  0,
+			Cap:  slot.class}))
+	}) {
+		// Allocate a segment directly from the system heap.
+		head = mmap(n)[0:0:n]
+	}
+
+	return head
+}
+
+func (p heap) Put(head []byte) {
+	if !p.pick(cap(head), func(slot *slot) {
+		slot.put(unsafe.Pointer(
+			(*reflect.SliceHeader)(unsafe.Pointer(&head)).Data))
+	}) {
+		// Nothing fits so it must be a system heap segment.
+		munmap(head[:cap(head)])
+	}
+}
--- a/pool/heap.go
+++ b/pool/heap.go
@@ -44,7 +44,7 @@ func NewNativeHeap(b []Bucket, po ObjectPoolOptions) BytesPool {
 
 	sort.Sort(BucketByCapacity(b))
 
-	r := heap{m: heapMetrics{
+	h := heap{m: heapMetrics{
 		overflows: m.Counter("overflows"),
 		misplaces: m.Counter("misplaces"),
 	}}
@@ -54,7 +54,7 @@ func NewNativeHeap(b []Bucket, po ObjectPoolOptions) BytesPool {
 			"class": strconv.Itoa(cfg.Capacity),
 		})
 
-		v := &slot{class: cfg.Capacity, opts: NativePoolOptions{
+		s := &slot{class: cfg.Capacity, opts: NativePoolOptions{
 			Size: uint(cfg.Count),
 			Type: reflect.ArrayOf(cfg.Capacity, ByteType),
 		}, m: slotMetrics{
@@ -62,10 +62,10 @@ func NewNativeHeap(b []Bucket, po ObjectPoolOptions) BytesPool {
 			size: m.Gauge("size"),
 		}}
 
-		r.slots = append(r.slots, v)
+		h.slots = append(h.slots, s)
 	}
 
-	return r
+	return h
 }
 
 type heap struct {
@@ -145,6 +145,11 @@ func (s *slot) put(segment interface{}) {
 	s.RUnlock()
 }
 
+func (s *slot) init() {
+	// Preallocate a pool for each sizeclass to save time later.
+	s.pools = []NativePool{NewNativePool(s.opts)}
+}
+
 func (s *slot) updateMetrics() {
 	// TODO(@kobolog): Use Dice.
 	if time.Now().UnixNano()%sampleObjectPoolLengthEvery != 0 {
@@ -169,8 +174,8 @@ func (s *slot) updateMetrics() {
 }
 
 func (p heap) Init() {
-	for _, slot := range p.slots {
-		slot.pools = append(slot.pools, NewNativePool(slot.opts))
+	for _, s := range p.slots {
+		s.init()
 	}
 }
 
@@ -188,11 +193,11 @@ func (p heap) pick(class int, action func(*slot)) bool {
 func (p heap) Get(n int) []byte {
 	var head []byte
 
-	if p.pick(n, func(slot *slot) {
+	if p.pick(n, func(s *slot) {
 		head = *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
-			Data: reflect.ValueOf(slot.get()).Pointer(),
+			Data: reflect.ValueOf(s.get()).Pointer(),
 			Len:  0,
-			Cap:  slot.class}))
+			Cap:  s.class}))
 	}) {
 		return head
 	}
@@ -204,8 +209,8 @@ func (p heap) Get(n int) []byte {
 }
 
 func (p heap) Put(head []byte) {
-	if p.pick(cap(head), func(slot *slot) {
-		slot.put(unsafe.Pointer(
+	if p.pick(cap(head), func(s *slot) {
+		s.put(unsafe.Pointer(
 			(*reflect.SliceHeader)(unsafe.Pointer(&head)).Data))
 	}) {
 		return
--- a/pool/heap.go
+++ b/pool/heap.go
@@ -59,8 +59,8 @@ func NewNativeHeap(b []Bucket, po ObjectPoolOptions) BytesPool {
 			Size: uint(cfg.Count),
 			Type: reflect.ArrayOf(cfg.Capacity, ByteType),
 		}, m: slotMetrics{
-			used: m.Gauge("used"),
-			size: m.Gauge("size"),
+			free: m.Gauge("free"),
+			size: m.Gauge("total"),
 		}}
 
 		h.slots = append(h.slots, s)
@@ -92,7 +92,7 @@ type slot struct {
 }
 
 type slotMetrics struct {
-	used tally.Gauge
+	free tally.Gauge
 	size tally.Gauge
 }
 
@@ -158,20 +158,20 @@ func (s *slot) updateMetrics() {
 		return
 	}
 
-	var used, size int64
+	var free, size int64
 
 	s.RLock()
 
 	for i := range s.pools {
 		a, b := s.pools[i].Size()
 
-		used += int64(a)
+		free += int64(a)
 		size += int64(b)
 	}
 
 	s.RUnlock()
 
-	s.m.used.Update(used)
+	s.m.free.Update(free)
 	s.m.size.Update(size)
 }
 
--- a/pool/heap.go
+++ b/pool/heap.go
@@ -51,7 +51,7 @@ func NewNativeHeap(b []Bucket, po ObjectPoolOptions) BytesPool {
 
 	for _, cfg := range b {
 		m := m.Tagged(map[string]string{
-			"class": strconv.Itoa(cfg.Capacity),
+			"bucket-capacity": strconv.Itoa(cfg.Capacity),
 		})
 
 		s := &slot{class: cfg.Capacity, opts: NativePoolOptions{
--- a/pool/heap.go
+++ b/pool/heap.go
@@ -21,39 +21,75 @@
 package pool
 
 import (
+	"fmt"
 	"reflect"
 	"sort"
 	"sync"
+	"time"
 	"unsafe"
+
+	"github.com/uber-go/tally"
 )
 
 // NewNativeHeap constructs a new BytesPool based on NativePool.
-func NewNativeHeap(sizes []Bucket) BytesPool {
-	var (
-		slots = make([]*slot, len(sizes))
-		T     = reflect.TypeOf((byte)(0))
-	)
-
-	sort.Sort(BucketByCapacity(sizes))
-
-	for i, cfg := range sizes {
-		ns := &slot{class: cfg.Capacity, cfg: NativePoolOptions{
-			Size: uint(cfg.Count),
-			Type: reflect.ArrayOf(cfg.Capacity, T)}}
-		slots[i] = ns
+func NewNativeHeap(b []Bucket, po ObjectPoolOptions) BytesPool {
+	if po == nil {
+		po = NewObjectPoolOptions()
 	}
 
-	return heap(slots)
+	var (
+		m        = po.MetricsScope()
+		ByteType = reflect.TypeOf((byte)(0))
+	)
+
+	sort.Sort(BucketByCapacity(b))
+
+	r := heap{m: heapMetrics{
+		overflows: m.Counter("overflows"),
+		misplaces: m.Counter("misplaces"),
+	}}
+
+	for _, cfg := range b {
+		m := m.SubScope(fmt.Sprintf("slot-%d", cfg.Capacity))
+
+		v := &slot{class: cfg.Capacity, opts: NativePoolOptions{
+			Size: uint(cfg.Count),
+			Type: reflect.ArrayOf(cfg.Capacity, ByteType),
+		}, m: slotMetrics{
+			used: m.Gauge("used"),
+			size: m.Gauge("size"),
+		}}
+
+		r.slots = append(r.slots, v)
+	}
+
+	return r
 }
 
-type heap []*slot
+type heap struct {
+	slots []*slot
+
+	m heapMetrics
+}
+
+type heapMetrics struct {
+	overflows tally.Counter
+	misplaces tally.Counter
+}
 
 type slot struct {
 	sync.RWMutex
 
 	class int
-	cfg   NativePoolOptions
+	opts  NativePoolOptions
 	pools []NativePool
+
+	m slotMetrics
+}
+
+type slotMetrics struct {
+	used tally.Gauge
+	size tally.Gauge
 }
 
 func (s *slot) get() interface{} {
@@ -67,7 +103,7 @@ func (s *slot) get() interface{} {
 	// then grow while holding an exclusive lock.
 	return s.getOr(s, func() interface{} {
 		s.pools = append(
-			[]NativePool{NewNativePool(s.cfg)}, s.pools...)
+			[]NativePool{NewNativePool(s.opts)}, s.pools...)
 		return s.pools[0].Get()
 	})
 }
@@ -82,6 +118,7 @@ func (s *slot) getOr(l sync.Locker, fn OverflowFn) interface{} {
 			return nil
 		}); segment != nil {
 			l.Unlock()
+			s.updateMetrics()
 			return segment
 		}
 	}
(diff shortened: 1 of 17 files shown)


kobolog:
GC traces for a test allocating 1,000,000 objects demonstrate that OpaquePool reduces GC scan phase pressure to almost nothing as well as end up having a smaller heap size in general:

```
=== RUN   TestOpaquePoolMemoryUsage
gc 11 @2.202s 1%: 0.008+2.5+2.9 ms clock, 0.034+0/2.3/0.47+11 ms cpu, 4143->4143->3944 MB, 4144 MB goal, 4 P
gc 12 @2.359s 2%: 0.002+0+29 ms clock, 0.008+0/2.3/0.47+119 ms cpu, 3944->3944->3937 MB, 3944 MB goal, 4 P (forced)
--- PASS: TestOpaquePoolMemoryUsage (2.09s)
=== RUN   TestObjectPoolMemoryUsage
gc 13 @2.391s 3%: 0.008+5.3+7.3 ms clock, 0.035+0.17/0.22/2.5+29 ms cpu, 3944->3960->23 MB, 5118 MB goal, 4 P
gc 14 @2.404s 3%: 0.006+6.8+8.1 ms clock, 0.025+0.10/0.25/7.1+32 ms cpu, 26->27->27 MB, 30 MB goal, 4 P
gc 15 @2.421s 3%: 0.013+9.1+5.2 ms clock, 0.055+0/0.61/11+21 ms cpu, 32->33->32 MB, 35 MB goal, 4 P
gc 16 @2.437s 3%: 0.008+11+7.5 ms clock, 0.033+0.039/0.90/11+30 ms cpu, 40->41->40 MB, 42 MB goal, 4 P
gc 17 @2.459s 4%: 0.012+8.1+11 ms clock, 0.048+0.17/0.55/8.7+45 ms cpu, 51->52->51 MB, 52 MB goal, 4 P
gc 18 @2.481s 4%: 0.009+7.5+12 ms clock, 0.039+0.16/7.4/1.4+51 ms cpu, 65->66->65 MB, 66 MB goal, 4 P
gc 19 @2.505s 5%: 0.008+7.7+9.2 ms clock, 0.035+0.83/7.5/0.88+37 ms cpu, 83->84->83 MB, 85 MB goal, 4 P
gc 20 @2.529s 5%: 0.013+8.2+13 ms clock, 0.055+0.47/8.2/2.2+53 ms cpu, 106->107->105 MB, 108 MB goal, 4 P
gc 21 @2.557s 6%: 0.013+13+13 ms clock, 0.054+1.4/13/1.5+52 ms cpu, 136->137->135 MB, 137 MB goal, 4 P
gc 22 @2.594s 6%: 0.034+10+10 ms clock, 0.13+0/2.7/11+41 ms cpu, 173->174->172 MB, 175 MB goal, 4 P
gc 23 @2.629s 7%: 0.016+11+13 ms clock, 0.066+2.3/11/5.8+55 ms cpu, 221->222->219 MB, 223 MB goal, 4 P
gc 24 @2.669s 7%: 0.020+13+16 ms clock, 0.080+7.3/11/3.0+66 ms cpu, 282->283->279 MB, 285 MB goal, 4 P
gc 25 @2.716s 8%: 0.046+16+18 ms clock, 0.18+0/9.0/16+73 ms cpu, 359->360->355 MB, 363 MB goal, 4 P
gc 26 @2.774s 9%: 0.032+15+18 ms clock, 0.12+0.006/7.2/16+75 ms cpu, 456->457->451 MB, 461 MB goal, 4 P
gc 27 @2.840s 9%: 0.024+17+16 ms clock, 0.097+0.11/10/18+64 ms cpu, 579->581->573 MB, 586 MB goal, 4 P
gc 28 @2.906s 10%: 0.028+24+24 ms clock, 0.11+0.16/20/23+97 ms cpu, 737->742->732 MB, 745 MB goal, 4 P
gc 29 @3.009s 10%: 0.012+17+22 ms clock, 0.048+0.087/14/18+89 ms cpu, 940->943->930 MB, 951 MB goal, 4 P
gc 30 @3.112s 11%: 0.14+21+27 ms clock, 0.57+17/21/1.8+108 ms cpu, 1195->1199->1183 MB, 1209 MB goal, 4 P
gc 31 @3.236s 12%: 0.041+38+30 ms clock, 0.16+9.4/38/28+120 ms cpu, 1520->1530->1508 MB, 1537 MB goal, 4 P
gc 32 @3.402s 13%: 0.010+31+37 ms clock, 0.043+0.12/29/30+150 ms cpu, 1938->1950->1923 MB, 1961 MB goal, 4 P
gc 33 @3.590s 14%: 0.046+40+49 ms clock, 0.18+32/40/4.8+197 ms cpu, 2471->2488->2452 MB, 2500 MB goal, 4 P
gc 34 @3.834s 15%: 0.054+49+57 ms clock, 0.21+39/49/4.9+228 ms cpu, 3150->3172->3127 MB, 3188 MB goal, 4 P
gc 35 @4.156s 16%: 0.008+66+72 ms clock, 0.035+43/66/6.7+289 ms cpu, 4016->4045->3988 MB, 4066 MB goal, 4 P
gc 36 @4.641s 18%: 0.002+0+149 ms clock, 0.008+43/66/6.7+599 ms cpu, 4446->4446->4402 MB, 4446 MB goal, 4 P (forced)
```

In this case, total background GC time for `OpaquePool` is around 5ms while `ObjectPool` is around 300ms, which is a 60x gain. STW mark termination times are through the roof as well.

pool/opaque.go
```diff
@@ -0,0 +1,97 @@
...
+// Get provides an object from the pool.
+func (p *OpaquePool) Get() interface{} {
+	var val interface{}
+
+	select {
+	case offset := <-p.free:
+		val = reflect.NewAt(
+			p.opts.Type, unsafe.Pointer(&p.heap[offset])).Interface()
+	default:
+		val = reflect.New(p.opts.Type).Interface()
```
robskillington: Will this have a GC root?  If not I assume we might have to somehow considering this is from outside the pool.

pool/opaque.go
```diff
@@ -0,0 +1,97 @@
...
+	ptr := unsafe.Pointer(reflect.ValueOf(object).Pointer())
+
+	if uintptr(ptr) < uintptr(unsafe.Pointer(&p.heap[0])) ||
+		uintptr(ptr) > uintptr(unsafe.Pointer(&p.heap[len(p.heap)-1])) {
+		return
+	}
+
+	// We know that this object is on our heap, so it's okay
+	// to read memory directly in front of it.
+	p.free <- (*hdr)(unsafe.Pointer(uintptr(ptr) - hsz)).offset
```
robskillington: Maybe we should check the address and see if it was from the pool, and if so directly enqueue it like we are here, but otherwise we just need to let the GC collect it.  i.e. we can't put normal heap allocated object address back in the pool.

pool/opaque.go
kobolog: Yeah, `reflect.New()` ends up calling `mallocgc` which will properly set up the heap bitmap for it so that GC scan could see it.

pool/opaque.go
kobolog: Yeah that's what I'm doing on the previous line. If the object address is outside of our heap, we just ignore it, letting GC collect it later.

pool/opaque_test.go
```diff
@@ -0,0 +1,108 @@
...
+	return n.y
+}
+
+func TestOpaquePool(t *testing.T) {
+	i := NewOpaquePool(OpaquePoolOptions{
+		Size: 50000,
+		Type: reflect.TypeOf((*N)(nil)).Elem()})
+
+	p := NewOpaquePool(OpaquePoolOptions{
+		Construct: func(ptr interface{}) { ptr.(*T).x = i.Get().(I) },
```
kobolog: Yes. As we discussed, this works for pointers and interfaces, but unfortunately not for maps, slices or channels – these types allocate memory internally and we don't have control over it. For example, a slice would allocate its storage via `mallocgc()`, which will properly flag it on the heap, but GC will see it as unreachable and reclaim it since its ownership cannot be traced through our opaque heap.

I might make it work for "immutable" slices, e.g. by explicitly allocating slice storage from the opaque heap, populating the `reflect.SliceHeader` structure and casting it to a slice.

Maps and channels are complicated structures and are not exposed via `reflect`, so no luck here.

pool/opaque.go
robskillington: Oh great 👍

pool/opaque.go
```diff
@@ -0,0 +1,106 @@
...
+
+		// Alternatively, we can use a channel of interfaces, but it
+		// would keep opts.Size objects rooted in the channel.
+		// We aim to avoid allocating unnecessary GC-visible objects.
+		p.free <- hdr.offset
+	}
+}
+
+// Get provides an object from the pool.
+func (p *OpaquePool) Get() interface{} {
```
xichen2020: Seems we don't reset the object when getting it from the pool or returning it to the pool. Would it be better to add a comment saying the callers should reset the state of the object returned from pool?

pool/opaque.go
```diff
@@ -0,0 +1,106 @@
...
+
+	return val
+}
+
+// Put returns an object to the pool.
+func (p *OpaquePool) Put(object interface{}) {
+	ptr := unsafe.Pointer(reflect.ValueOf(object).Pointer())
+
+	if (uintptr(ptr) < uintptr(unsafe.Pointer(&p.heap[0]))) ||
+		uintptr(ptr) > uintptr(unsafe.Pointer(&p.heap[p.size-1])) {
```
xichen2020: Might also make sense to check if `ptr` - `hsz` is a multiple of `p.step`?

pool/opaque.go
kobolog: We don't reset objects in `ObjectPool` either, so I'm just following the API here.

pool/opaque.go
kobolog: If it's not a multiple of `p.step`, then we got some sort of heap corruption – since offsets are immutable. I sort of assume that nothing will change the heap outside object boundaries. How would that happen (apart of deliberate action)?

kobolog:
Updating this thread with the most recent results from M3DB testing – in general, the version with IDs being allocated from an `OpaquePool` has ~1.5x less time spent in background GC with is **surprising** to say the least. I do not know if there are some other factors in play here yet. It looks like repairs are happening from time to time on both nodes, based on their logs, which I guess adds some uncertainty. Comments @r & @xichen2020?

- Without hidden IDs:
```
gc 62 @1636.897s 8%: 1.4+17949+5.8 ms clock, 17+60052/53289/3364+69 ms cpu, 38334->39982->31776 MB, 40704 MB goal, 12 P
gc 63 @1711.550s 8%: 2.5+21556+3.7 ms clock, 30+5929/64643/53115+45 ms cpu, 38339->40397->32694 MB, 41309 MB goal, 12 P
gc 64 @1795.742s 9%: 1.8+17143+7.5 ms clock, 22+58984/50885/5663+90 ms cpu, 39817->41613->32402 MB, 42502 MB goal, 12 P
gc 65 @1863.426s 9%: 4.7+21064+4.3 ms clock, 57+28487/62624/21636+52 ms cpu, 38882->41093->33218 MB, 42122 MB goal, 12 P
gc 66 @1936.206s 9%: 1.3+18280+4.4 ms clock, 16+8069/54668/49076+53 ms cpu, 39878->41529->32903 MB, 43183 MB goal, 12 P
gc 67 @2017.157s 8%: 1.3+15909+3.5 ms clock, 9.2+6840/47614/53832+24 ms cpu, 40200->41653->32746 MB, 42774 MB goal, 12 P
gc 68 @2113.823s 8%: 1.4+16361+4.2 ms clock, 17+36463/48581/34116+50 ms cpu, 40462->41596->32379 MB, 42569 MB goal, 12 P
gc 69 @2250.202s 8%: 27+16683+3.9 ms clock, 223+1433/49757/63019+31 ms cpu, 39572->40527->32266 MB, 42093 MB goal, 12 P
```

- With hidden IDs:
```
gc 62 @1402.601s 8%: 1.3+14543+4.6 ms clock, 15+325/41129/62375+55 ms cpu, 38093->39184->30675 MB, 40114 MB goal, 12 P
gc 63 @1477.154s 9%: 1.5+13125+3.7 ms clock, 18+67984/39005/2568+44 ms cpu, 38327->39379->30816 MB, 39877 MB goal, 12 P
gc 64 @1586.261s 8%: 1.3+14167+3.3 ms clock, 16+2202/42099/67841+40 ms cpu, 37841->38716->30694 MB, 40061 MB goal, 12 P
gc 65 @1720.435s 8%: 1.5+11393+3.2 ms clock, 7.5+2710/33315/64665+16 ms cpu, 38062->38707->30538 MB, 39902 MB goal, 12 P
gc 66 @1851.845s 7%: 1.3+11827+3.3 ms clock, 16+5189/35476/61434+39 ms cpu, 37767->38455->30654 MB, 39700 MB goal, 12 P
gc 67 @1983.689s 7%: 1.3+11957+3.6 ms clock, 16+4913/35200/60748+43 ms cpu, 37877->38614->30772 MB, 39851 MB goal, 12 P
gc 68 @2115.661s 7%: 1.3+11691+3.0 ms clock, 15+395/34181/63322+36 ms cpu, 38078->38800->30816 MB, 40003 MB goal, 12 P
gc 69 @2247.360s 6%: 1.3+11417+3.7 ms clock, 16+380/33441/65109+44 ms cpu, 38017->38694->30797 MB, 40061 MB goal, 12 P
```

robskillington:
The big background GC spikes we've seen happen more frequently than repairs, so not sure we can pin it on that.  Repairs will probably exarcebate the large background GC spikes but it definitely occurs much more frequently than repairs run (roughly once every 2h).  You can see these large spikes roughly every 10mins or so, sometimes more frequently too.

The results look promising, although we'll likely never be truly rid of this issue unless we move somethings into C/C++ and call across a CGO boundary.

kobolog:
I got rid of spikes almost entirely in testing with this PR & certain reconfiguration of byte pool sizes. Currently, the GC heap size I see there is ~28G vs. ~60G and 75% spikes happen maybe every 30 minutes as opposed to 100% peaks every ~2 minutes normally (which is when `runtime.sysmon` forces GC runs).

pool/heap.go
```diff
@@ -0,0 +1,216 @@
...
+
+	sort.Sort(BucketByCapacity(b))
+
+	r := heap{m: heapMetrics{
+		overflows: m.Counter("overflows"),
+		misplaces: m.Counter("misplaces"),
+	}}
+
+	for _, cfg := range b {
+		m := m.SubScope(fmt.Sprintf("slot-%d", cfg.Capacity))
```
xichen2020: Would it be slightly better to make this a tag (e.g., `slot: capacity`)?

pool/heap.go
```diff
@@ -0,0 +1,216 @@
...
+	m slotMetrics
+}
+
+type slotMetrics struct {
+	used tally.Gauge
+	size tally.Gauge
+}
+
+func (s *slot) get() interface{} {
+	if segment := s.getOr(s.RLocker(), func() interface{} {
```
xichen2020: `object` might be a better name than segment?

pool/heap.go
```diff
@@ -0,0 +1,216 @@
...
+		return nil
+	}); segment != nil {
+		return segment
+	}
+
+	// Slow path - double-check that there are no segments left,
+	// then grow while holding an exclusive lock.
+	return s.getOr(s, func() interface{} {
+		s.pools = append(
```
xichen2020: How expensive is it to create a new native pool? If it is expensive, I wonder if we should create the pool outside the write lock and only acquire the lock when the pool is created

pool/heap.go
```diff
@@ -0,0 +1,216 @@
...
+
+type heapMetrics struct {
+	overflows tally.Counter
+	misplaces tally.Counter
+}
+
+type slot struct {
+	sync.RWMutex
+
+	class int
```
xichen2020: Should this be called capacity?

pool/mmap.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+const (
+	mmapReadWriteAccess = syscall.PROT_READ | syscall.PROT_WRITE
+	mmapMemory          = syscall.MAP_ANON | syscall.MAP_PRIVATE
+)
+
+func munmap(head []byte) error {
+	if err := syscall.Munmap(head); err != nil {
+		fmt.Printf("munmap(%p, %d) error: %v\n", &head[0], cap(head), err)
```
xichen2020: Probably better to use `fmt.Errorf`?

pool/native.go
```diff
@@ -0,0 +1,157 @@
...
+		// TODO(@kobolog): alignment & padding.
+		step: uint(hsz + opts.Type.Size())}
+
+	p.size = p.opts.Size * p.step
+
+	if p.size == 0 {
+		panic("native-pool: pool size is zero")
+	}
+
+	if p.step%uint(p.opts.Type.Align()) != 0 {
```
xichen2020: hmm how can the user guarantee that this doesn't happen?

pool/native.go
```diff
@@ -0,0 +1,157 @@
...
+
+type nativePool struct {
+	pool       []byte
+	free       chan uint
+	opts       NativePoolOptions
+	step, size uint
+}
+
+func (p *nativePool) Size() (uint, uint) {
+	return p.size - uint(len(p.free))*p.step, uint(cap(p.free)) * p.step
```
xichen2020: nit: assign the latter to a var and reuse it

ts/identifier_pool.go
```diff
@@ -21,26 +21,50 @@
...
+func create() interface{} {
+	return &id{}
 }
 
 // GetBinaryID returns a new ID based on a binary value
 func (p *identifierPool) GetBinaryID(ctx context.Context, v []byte) ID {
-	id := p.pool.Get().(*id)
-	id.data = v
+	id := p.pool.GetOr(create).(*id)
+	id.pool, id.data = p, append(p.heap.Get(len(v)), v...)
```
xichen2020: We are changing the semantics here from sharing the byte slice to owning the slice, is that intentional?

pool/heap.go
kobolog: It's not really an object, but a pointer to the beginning of the memory segment (or block, or something, but definitely not an object).

pool/heap.go
kobolog: It's relatively expensive (milliseconds), but in general case shouldn't happen if pool sizes are properly set.

pool/mmap.go
kobolog: `fmt.Errorf` won't print anything, right? Or you mean something like `fmt.Println(fmt.Errorf(...))`?

pool/native.go
kobolog: This is just a sanity check. User can't guarantee it, as far as I understand golang compiler it shouldn't even happen, but in case something is wrong (like, they change the compiler, we run it on weird hardware, etc.) it's here.

pool/heap.go
xichen2020: fair enough

pool/native.go
kobolog: It's 'cap' vs. 'len', nothing to reuse here.

pool/mmap.go
xichen2020: oh duh, meant `log.Errorf(...)`. You can set instrumentation options on the pool options, which contains a logger you can use.

pool/heap.go
xichen2020: fair enough

pool/native.go
xichen2020: If I understand it correctly, this means `p.step`, which is `hsz + opts.Type.Size()`, must be a multiple of `opts.Type.Align()`, which means `hsz` must be a multiple of `opts.Type.Align()`. Is that guaranteed by the compiler?

kobolog: So the compiler pads types to be properly aligned, which means both `hdr` and whatever type was passed in `opts` both are supposed to be fine automatically. So I guess this invariant could be only broken if those alignments are somehow incompatible, e.g. `alignof(hdr)` < `alignof(T)`. I don't know whether this could happen or not, so I added this check just in case.

Actually, I'm gonna change `hdr.idx` to uint64 so that it has the highest alignment, which probably should always work.

pool/heap.go
```diff
@@ -0,0 +1,218 @@
...
+	sort.Sort(BucketByCapacity(b))
+
+	r := heap{m: heapMetrics{
+		overflows: m.Counter("overflows"),
+		misplaces: m.Counter("misplaces"),
+	}}
+
+	for _, cfg := range b {
+		m := m.Tagged(map[string]string{
+			"class": strconv.Itoa(cfg.Capacity),
```
robskillington: Can we change this tag name to "bucket-capacity" to make it compatible with the other bytes pool?  Then we can use the same graphs we already have setup to pivot by the pool name.

pool/heap.go
```diff
@@ -0,0 +1,218 @@
...
+	}
+
+	var (
+		m        = po.MetricsScope()
+		ByteType = reflect.TypeOf((byte)(0))
+	)
+
+	sort.Sort(BucketByCapacity(b))
+
+	r := heap{m: heapMetrics{
```
robskillington: Perhaps `h` is a better local var name here

pool/heap.go
```diff
@@ -0,0 +1,218 @@
...
+		overflows: m.Counter("overflows"),
+		misplaces: m.Counter("misplaces"),
+	}}
+
+	for _, cfg := range b {
+		m := m.Tagged(map[string]string{
+			"class": strconv.Itoa(cfg.Capacity),
+		})
+
+		v := &slot{class: cfg.Capacity, opts: NativePoolOptions{
```
robskillington: `s` for slot instead of `v`?

pool/heap.go
```diff
@@ -0,0 +1,218 @@
...
+
+	s.RUnlock()
+
+	s.m.used.Update(used)
+	s.m.size.Update(size)
+}
+
+func (p heap) Init() {
+	for _, slot := range p.slots {
+		slot.pools = append(slot.pools, NewNativePool(slot.opts))
```
robskillington: I think it would be cleaner if `slot` had it's own `init` method that sets the `pools` property.  That way any other initialization that may need to be added in the future can be encapsulated well and the responsibility to initialize fields is delegated to the `slot` itself.

pool/heap.go
```diff
@@ -0,0 +1,223 @@
...
+		return head
+	}
+
+	p.m.overflows.Inc(1)
+
+	// Allocate a segment directly from the system heap.
+	return mmap(n)[:0:n]
+}
+
+func (p heap) Put(head []byte) {
```
robskillington: How do we guard against this not being a Go allocated `[]byte` being returned to the heap here?  Or is this at a higher layer?

pool/mmap.go
robskillington: yeah might be worth just returning the error directly and then in the pool which has access to the `ObjectPoolOptions` peel off the logger with `opts.InstrumentOptions().Logger()` and then attach to the pool struct and `p.log.Errorf(....)`.

robskillington: Now that I'm remembering this, might be worth making it a `p.log.Debugf(...)` and incrementing a metric instead.  Either or I suppose.

Leaving it as a log will make us fix it quicker, but also leaves us open to log spam taking out our processes.

pool/heap.go
kobolog: If it matches one of the slot sizes, no pools there will pass the `pool.Owns()` guard in `slot.put()`, and if it's big enough to end up being `munmap()`-ed, then it will simply fail because `munmap()` keeps track of mapped regions and it won't be one of them.

pool/mmap_linux.go
```diff
@@ -0,0 +1,49 @@
...
+	if err != nil {
+		panic(fmt.Errorf("mmap(%d) error: %v", n, err))
+	}
+
+	if n < mmapHugePageSize {
+		return r
+	}
+
+	if err := syscall.Madvise(r, syscall.MADV_HUGEPAGE); err != nil {
+		fmt.Printf("madvise(%p, HUGEPAGE) error: %v\n", &r[0], err)
```
robskillington: We should probably panic here too.

pool/native.go
```diff
@@ -0,0 +1,157 @@
...
+	p := &nativePool{
+		free: make(chan uint64, opts.Size),
+		opts: opts,
+		// TODO(@kobolog): alignment & padding.
+		step: uint64(hsz + opts.Type.Size())}
+
+	p.size = uint64(p.opts.Size) * p.step
+
+	if p.size == 0 {
+		panic("native-pool: pool size is zero")
```
robskillington: Would be more consistent and lets the caller decide on what to do in a case like this if we returned this as an error and then by default we can `log.Fatalf("could not instantiate pool: %v", err)` .

pool/native.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	if p.step%uint64(p.opts.Type.Align()) != 0 {
+		panic("native-pool: invalid alignment")
```
robskillington: Here too.

pool/mmap_linux.go
kobolog: No, it's fine if we failed to madvise() - some kernels simply might not have THP built-in, and in this case we don't care that much, since it's just a performance penalty.

pool/native.go
```diff
@@ -0,0 +1,157 @@
...
+	return p.size - uint64(len(p.free))*p.step, uint64(cap(p.free)) * p.step
+}
+
+type hdr struct {
+	// Offset from the beginning of the arena to the object.
+	idx uint64
+}
+
+const (
+	hsz = unsafe.Sizeof(*(*hdr)(nil))
```
robskillington: Can we move this and `type hdr` to the top of the file so its comes first when you're reading this file?  I can to scan around to find it.

pool/native.go
kobolog: This is a sanity check (an assert(), basically), if it fails there's absolutely nothing you can do.

pool/native.go
kobolog: Moved all unexported types together. Won't move it to the top of the file since it's for **exported** stuff, not implementation detail.

pool/mmap.go
kobolog: `ObjectPoolOptions` don't have `InstrumentOptions`.

robskillington: That's fine, we probably should move it to `m3x` anyhow.  Sounds good.

pool/native.go
robskillington: Ah yeah, this one makes sense as a panic 👍

pool/native.go
robskillington: Ah right, yeah so the pattern here is usually to have the options, e.g. for this case `NativePoolOptions`, implement a `Validate() error` method and then call it and pass back if `Size` option is `< 1` .

Also `NativePoolOptions` is a value type which is not consistent with any of our other option types.  If it's not too much effort it would be better to make it look the same as all the other `Option` types in the project.

Regarding if `p.step` is zero, totally agree `panic` there would make sense.

pool/native.go
```diff
@@ -0,0 +1,159 @@
...
+	return p
+}
+
+type hdr struct {
+	// Offset from the beginning of the arena to the object.
+	idx uint64
+}
+
+// Header size with padding, as determined by the compiler.
+const hsz = unsafe.Sizeof(*(*hdr)(nil))
```
robskillington: Can we move this to the top of the file along with `hdr` so we can read it first before seeing it reference in the `NewNativePool` method?

pool/native.go
```diff
@@ -0,0 +1,159 @@
...
+		// Alternatively, we can use a channel of interfaces, but it
+		// would keep opts.Size objects rooted in the channel.
+		// We aim to avoid allocating unnecessary GC-visible objects.
+		p.free <- hdr.idx
+	}
+}
+
+func (p *nativePool) Size() (uint64, uint64) {
+	free := uint64(cap(p.free)) * p.step
+	used := p.size - uint64(len(p.free))*p.step
```
robskillington: Can't we reuse `free` here? i.e. `used := p.size - free`

pool/native.go
```diff
@@ -0,0 +1,159 @@
...
+
+	return used, free
+}
+
+// Get provides an object from the pool.
+func (p *nativePool) Get() interface{} {
+	return reflect.NewAt(
+		p.opts.Type, unsafe.Pointer(&p.pool[<-p.free])).Interface()
```
robskillington: Won't this block forever if there is nothing left in the free pool?

pool/native.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	p.init()
+
+	return p
+}
+
+type hdr struct {
+	// Offset from the beginning of the arena to the object.
+	idx uint64
```
robskillington: So this is a memory address yeah?  Any reason not to use `uintptr`?

ts/identifier_pool.go
robskillington: It might be more efficient to use `copy(...)` than `append(x, ...)` as it's more likely to compile to a direct memcpy, ie.:
```go
id.pool, id.data = p, p.heap.Get(len(v))[:len(v)]
copy(id.data, v)
```

ts/identifier_pool.go
```diff
@@ -52,9 +76,9 @@ func (p *identifierPool) GetStringID(ctx context.Context, v string) ID {
 }
 
 // Clone replicates given ID into a new ID from the pool
-func (p *identifierPool) Clone(other ID) ID {
-	id := p.pool.Get().(*id)
-	id.data = other.Data()
+func (p *identifierPool) Clone(v ID) ID {
+	id := p.pool.GetOr(create).(*id)
+	id.pool, id.data = p, append(p.heap.Get(len(v.Data())), v.Data()...)
```
robskillington: Here too, `copy` likely more efficient than `append`

pool/heap.go
robskillington: Sounds good, I should've followed to the pool.owns check.  Thanks.

ts/identifier_pool.go
kobolog: Actually, `copy()` has exactly the same performance.

pool/native.go
kobolog: It's an index into a byte slice, so not technically a memory address. Also, `uintptr` is not guaranteed to be exactly 8 bytes according to the spec (although it probably should always be).

ts/identifier_pool.go
kobolog: Same performance.

pool/native.go
robskillington: Right yeah, hadn't read your other comment about the alignment.

Can you perhaps just add some more text to the comment here mentioning "we use uint64 for the largest padding to try and always be aligned with the contents"?  Or something along those lines?

pool/native.go
kobolog: Yes, it will. Yes, we use it in NativeHeap to avoid the `select` overhead when we know that there are free objects in the pool.

ts/identifier_pool.go
robskillington: Seems faster for some reason, haha:
```go
package main

import "testing"

const size = 4096

func BenchmarkCopy(b *testing.B) {
	src := make([]byte, size)
	dest := make([]byte, size)
	for i := 0; i < b.N; i++ {
		copy(dest, src)
	}
}

func BenchmarkAppend(b *testing.B) {
	src := make([]byte, size)
	dest := make([]byte, size)
	for i := 0; i < b.N; i++ {
		dest = append(dest[:0], src...)
	}
}
```

```
$ go test -v -bench .
testing: warning: no tests to run
BenchmarkCopy-4         20000000               103 ns/op
BenchmarkAppend-4       20000000                67.7 ns/op
PASS
```

Anyway yeah, fine to use append then.

pool/native.go
kobolog: I actually don't know why you made all options interfaces - it's not like you have different implementations for e.g. `ObjectPoolOptions`, right? It leads to code where you have things like `NewWhateverThing(nil, nil)` instead of `NewWhateverThing(WhateverOptions{}, ThingOptions{})`, which is obviously harder to comprehend. Also, it explodes the LOC because of all the getters and setters and brings near to no benefit. I call it overengineering unless you prove otherwise =)

pool/native.go
kobolog: Well, first of all we use uint64 so that we can make pools larger than 4gb, which should be obvious without comments. I'll add the alignment comment.

pool/mmap_linux.go
kobolog: I moved munmap logging and panics to the calling code, but I'm gonna leave this warning here since I see no point in creating a new entity just to propagate a warning, which will also complicate the calling side, i.e. we'll have to see if the returned error is actually not an error, in which case we can use the result, etc.

pool/native.go
robskillington: There are reasons as discussed, mainly to save memory by using a single reference across high cardinality objects to avoid memory waste.  Making the SetThing(...) returning a new set of mutated options was just to give the ability for immutable options.

Let's keep this as opts but we will need to consolidate and setup a style guide around the advice for this type of options usage on non-high-cardinality objects.

pool/native.go
robskillington: I see, makes sense 👍

pool/heap.go
kobolog: I guess if we later on add other heap types, like structs or whatnot, then it'd make sense to call it something else. I don't want to call it slice since it's ambiguous and might make one think that we're talking golang slices here, while we are not. Object is not really working too since it's an array of objects (in this case, bytes). So let's revisit it when adding other heap types.

PR #161, opened 2016-11-15
Fix array pooling leaks where element references are not removed

robskillington:
cc @kobolog @xichen2020 @martin-mao

### DIFF
--- a/storage/block/result.go
+++ b/storage/block/result.go
@@ -107,15 +107,14 @@ func (s *fetchBlockMetadataResults) Sort() {
 }
 
 func (s *fetchBlockMetadataResults) Reset() {
-	s.results = s.results[:0]
-}
-
-func (s *fetchBlockMetadataResults) Close() {
 	var zeroed FetchBlockMetadataResult
 	for i := range s.results {
 		s.results[i] = zeroed
 	}
 	s.results = s.results[:0]
+}
+
+func (s *fetchBlockMetadataResults) Close() {
 	if s.pool != nil {
 		s.pool.Put(s)
 	}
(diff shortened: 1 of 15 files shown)


storage/block/result.go
```diff
@@ -111,6 +111,11 @@ func (s *fetchBlockMetadataResults) Reset() {
 }
 
 func (s *fetchBlockMetadataResults) Close() {
+	var zeroed FetchBlockMetadataResult
```
xichen2020: We should do this inside `fetchBlockMetadataResultsPool.Put()` since it already calls `Reset` above and is consistent with the other changes where we zero out the underlying slice when putting back to the pool

storage/block/result.go
```diff
@@ -151,9 +156,12 @@ func (s *fetchBlocksMetadataResults) Reset() {
...
 
 func (s *fetchBlocksMetadataResults) Close() {
-	for _, res := range s.results {
-		res.Blocks.Close()
+	var zeroed FetchBlocksMetadataResult
+	for i := range s.results {
+		s.results[i].Blocks.Close()
+		s.results[i] = zeroed
 	}
+	s.results = s.results[:0]
```
xichen2020: Same here, should do it inside `fetchBlocksMetadataResultsPool.Put()` for reasons mentioned above

storage/repair/metadata.go
```diff
@@ -60,6 +60,11 @@ func (s *hostBlockMetadataSlice) Reset() {
 }
 
 func (s *hostBlockMetadataSlice) Close() {
+	var zeroed HostBlockMetadata
+	for i := range s.metadata {
+		s.metadata[i] = zeroed
+	}
+	s.metadata = s.metadata[:0]
```
xichen2020: Same here, should do it inside `hostBlockMetadataSlicePool.Put()`

PR #163, opened 2016-11-17
Add mixed disk mode retention option and file seeker

martin-mao:
cc @xichen2020 @robskillington


robskillington:
Can we reuse some more code between the reader and seeker?

PR #166, opened 2016-11-23
Pull m3cluster with updated service discovery interfaces

cw9:
pick up m3cluster interface change introduced in the shard state pr, feel free to close this and do glide update in your own diff @robskillington


robskillington:
I've introduced these changes with https://github.com/m3db/m3db/pull/168 - thanks!

xichen2020:
@cw9 I assume this is no longer needed?

cw9:
I actually need this for the statsdex change that I plan to land before Rob's resharding change, just updated with the master m3cluster SHA

PR #167, opened 2016-11-23
Move out pool and instrument packages to M3X

kobolog:
This PR moves out `pool` and `instrument` common packages to M3X and fixes imports in M3DB.


client/session_bootstrap_blocks_test.go
```diff
@@ -1392,11 +1392,18 @@ func (s *synchronousWorkerPool) Init() {
...
 	work()
 }
 
-func (s *synchronousWorkerPool) GoIfAvailable(work pool.Work) bool {
+func (s *synchronousWorkerPool) GoIfAvailable(work xsync.Work) bool {
+	work()
+	return true
+}
+
+func (s *synchronousWorkerPool) GoWithTimeout(
```
xichen2020: Is this function used?

kobolog: I'm not sure, but I'm not cleaning up anything here either, just moving stuff.

xichen2020: hm..this is a new function no?

kobolog: Oh, sorry, I misunderstood the question – yeah it's a new function for this Mock because we had two WorkerPools: one in M3DB and one in M3X. I dropped the one that has less functionality, so this function was added to match the interface.

PR #168, opened 2016-11-29
Add dynamic resharding of nodes and introduce cluster database

robskillington:
cc @xichen2020 @martin-mao @cw9 @prateek @ben-lerner 

Need to address: We would not re-flush any newly bootstrapped shards to disk as we keep a map of processed time ranges, not time ranges by shards when we flush


services/m3dbnode/main/main.go
```diff
@@ -75,32 +79,29 @@ func main() {
 		}
 	}
 
-	clientOpts, err := server.DefaultClientOptions(id, tchannelNodeAddr, shardSet)
-	if err != nil {
-		log.Fatalf("could not create client options: %v", err)
-	}
-
-	c, err := client.NewClient(clientOpts)
+	cli, err := client.NewClient(server.DefaultClientOptions(topoInit))
```
cw9: probably should call topoInit.Init() here and reuse the Topology in both client and database options?

integration/integration.go
```diff
@@ -351,3 +359,62 @@ func writeToDisk(
 
 	return nil
 }
+
+func newShardsRange(from, to uint32) []uint32 {
+	var ids []uint32
```
xichen2020: Make it `ids := make([]int32, 0, to - from + 1) perhaps`?

integration/integration.go
```diff
@@ -351,3 +359,62 @@ func writeToDisk(
...
+		time.Sleep(time.Second)
+	}
+}
+
+func hasBootstrappedShardsExactly(
+	db storage.Database,
+	shards []uint32,
+) bool {
+	for _, namespace := range db.Namespaces() {
+		expect := make(map[uint32]struct{})
```
xichen2020: Doesn't look like `expect` changes, can we initialize it once and reuses it in the loop?

services/m3dbnode/server/serve.go
```diff
@@ -38,9 +38,9 @@ import (
 	"github.com/m3db/m3db/ts"
 )
 
-const defaultNamespaceName = "default"
+const defaultNamespaceName = "metrics"
```
xichen2020: Might be better to use something more generic than "metrics" since people can use m3db for other purposes

storage/bootstrap.go
```diff
@@ -86,43 +86,85 @@ func newBootstrapManager(
...
-		return nil
+	switch m.state {
+	case bootstrapping:
+		// NB(r): Already bootstrapping, now a consequent bootstrap
+		// request comes in - we queue this up to bootstrap again
+		// once the current bootstrap has completed.
+		// This is an edge case that can occur if during either an
+		// initial bootstrap or a resharding bootstrap if a new
+		// reshard occurs and we need to bootstrap more shards.
+		m.pending = true
```
xichen2020: Wonder if we can shortcut the current bootstrap and exit early since we need to rebootstrap anyway

storage/bootstrap.go
```diff
@@ -68,6 +67,7 @@ type bootstrapManager struct {
 	nowFn          clock.NowFn
 	newBootstrapFn NewBootstrapFn
 	state          bootstrapState
+	pending        bool
```
xichen2020: `hasPending` might be better?

integration/integration.go
```diff
@@ -351,3 +359,62 @@ func writeToDisk(
...
+func newShardsRange(from, to uint32) []uint32 {
+	var ids []uint32
+	for i := from; i <= to; i++ {
+		ids = append(ids, i)
+	}
+	return ids
+}
+
+func newClusterShardsRange(from, to uint32) shard.Shards {
+	return shard.NewShardsWithIDs(newShardsRange(from, to))
```
cw9: just FYI the shard.NewShardsWithIDs will be removed in the new m3cluster diff, you will need to construct a new Shards object with []Shard.

services/m3dbnode/server/serve.go
```diff
@@ -51,57 +51,72 @@ func DefaultShardSet() (sharding.ShardSet, error) {
...
 
-	hostShardSet := topology.NewHostShardSet(topology.NewHost(id, localNodeAddr), shardSet)
+	hostShardSet := topology.NewHostShardSet(topology.NewHost(hostID, localNodeAddr), shardSet)
 	staticOptions := topology.NewStaticOptions().
 		SetShardSet(shardSet).
 		SetReplicas(1).
 		SetHostShardSets([]topology.HostShardSet{hostShardSet})
 
-	return client.NewOptions().SetTopologyInitializer(topology.NewStaticInitializer(staticOptions)), nil
+	return topology.NewStaticInitializer(staticOptions), nil
```
cw9: Should the topology initializer be a dynamic one for dynamic resharding? it should take in the Config Service client like the M3DB client, or even have them both take in the dynamic topology directly?

storage/bootstrap.go
robskillington: So the namespace bootstrap method has been changed to only bootstrap shards that haven't been bootstrapped yet on request.  Therefore this current bootstrap still will make forward progress on the shards that it already has.  It just needs to bootstrap any remaining on the next bootstrap method.  

It's a pretty heavy operation to have to cancel all the possible in flight operations, especially when it comes to peer streaming, and start again.  Also the intermediate state for the shards that were bootstrapped wouldn't be saved and flushed to disk.

This reminds me however: it seems we would not re-flush the new shards to disk as we keep a map of processed time ranges, not time ranges by shards...

integration/integration.go
robskillington: This is just integration tests, no need to avoid allocations for this IMO.

integration/integration.go
robskillington: Could... but then again there's only one namespace we ever check against and also this is integration tests.  

I'd prefer keep test code simple and idiomatic rather than avoid allocations, considering, yknow, it's tests.

services/m3dbnode/server/serve.go
robskillington: Sure, I'll reset this to default.

storage/cluster/database.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, err
+	}
+
+	watch, err := topo.Watch()
+	if err != nil {
+		return nil, err
+	}
+
+	// Wait for the topology to be available
+	<-watch.C()
```
robskillington: I don't want to necessarily change how this works in production right now.  This is how the client works.

services/m3dbnode/main/main.go
robskillington: So to avoid having to initialize it up front, I can lazily initialize it and put a lock on the dynamic topology so consequent calls to `.Init()` return the same instance it created the first time. 

This should work well yes?

integration/integration.go
cw9: yes

storage/cluster/database.go
```diff
@@ -0,0 +1,199 @@
...
+		watch:    watch,
+	}, nil
+}
+
+func (d *clusterDB) Open() error {
+	select {
+	case <-d.watch.C():
+		shardSet, ok := hostOrEmptyShardSet(d.hostID, d.watch.Get())
+		if !ok {
+			d.log.Warnf("topology has no shard set for host ID: %s", d.hostID)
```
martin-mao: Nit: Emit a gauge value of the number of shards owned?

storage/cluster/database.go
```diff
@@ -0,0 +1,199 @@
...
+
+func (d *clusterDB) activeTopologyWatch() {
+	for {
+		select {
+		case <-d.doneCh:
+			d.closedCh <- struct{}{}
+			return
+		case <-d.watch.C():
+			shardSet, ok := hostOrEmptyShardSet(d.hostID, d.watch.Get())
+			if !ok {
```
martin-mao: ??

storage/namespace.go
```diff
@@ -143,6 +143,69 @@ func (n *dbNamespace) NumSeries() int64 {
...
+	for i, shard := range shards {
+		databaseShards[i] = n.shards[shard]
+	}
+	n.RUnlock()
+	return databaseShards
+}
+
+func (n *dbNamespace) AssignShardSet(shardSet sharding.ShardSet) {
+	var (
+		incoming = make(map[uint32]struct{})
```
martin-mao: Nit: make with len(shardSet.Shards()) to avoid realloc

storage/namespace.go
```diff
@@ -143,6 +143,69 @@ func (n *dbNamespace) NumSeries() int64 {
...
+			continue
+		}
+		if _, ok := incoming[shard.ID()]; !ok {
+			closing = append(closing, shard)
+		}
+	}
+	n.shardSet = shardSet
+	n.shards = make([]databaseShard, n.shardSet.Max()+1)
+	for _, shard := range n.shardSet.Shards() {
+		if int(shard) < len(existing) && existing[shard] != nil {
```
martin-mao: Why is the int(shard) < len(existing) required here?

storage/namespace.go
```diff
@@ -143,6 +143,69 @@ func (n *dbNamespace) NumSeries() int64 {
...
+	// point of control for determining how impactful closing shards may
+	// be to performance, we let this be the single gate and simply spin
+	// up a goroutine per shard that we need to close and rely on the self
+	// throttling of each shard as determined by the close shard deadline to
+	// gate the impact.
+	for _, shard := range closing {
+		shard := shard
+		go func() {
+			if err := shard.Close(); err != nil {
+				n.log.WithFields(
```
martin-mao: Nit: emit metric?

storage/namespace.go
```diff
@@ -143,6 +143,69 @@ func (n *dbNamespace) NumSeries() int64 {
...
+		if int(shard) < len(existing) && existing[shard] != nil {
+			n.shards[shard] = existing[shard]
+		} else {
+			n.shards[shard] = newDatabaseShard(n.id, shard, n.increasingIndex,
+				n.writeCommitLogFn, n.nopts.NeedsBootstrap(), n.sopts)
+		}
+	}
+	n.Unlock()
+
+	// NB(r): There is a shard close deadline that controls how fast each
```
xichen2020: Is it a soft or a hard deadline? What happens if we don't meet it?

storage/namespace.go
```diff
@@ -277,6 +346,16 @@ func (n *dbNamespace) Bootstrap(
 	workers := pool.NewWorkerPool(int(math.Ceil(float64(runtime.NumCPU()) / 2)))
```
xichen2020: might make sense to reuse this by making it a namespace field if we might bootstrap multiple times

storage/shard.go
```diff
@@ -41,6 +42,7 @@ import (
 	xio "github.com/m3db/m3db/x/io"
 	xerrors "github.com/m3db/m3x/errors"
 	xtime "github.com/m3db/m3x/time"
+	"github.com/uber-go/tally"
```
xichen2020: Do we need a blank line to separate m3db repos from the 3rd party ones?

storage/shard.go
```diff
@@ -345,23 +455,28 @@ func (s *dbShard) writableSeries(id ts.ID) *dbShardEntry {
 	}
 
 	s.Lock()
-	if entry, _, exists := s.getEntryWithLock(id); exists {
+	if entry, _, err := s.lookupEntryWithLock(id); err == nil {
 		entry.incrementWriterCount()
-		s.Unlock()
+		s.RUnlock()
```
xichen2020: Should be `s.Unlock`

storage/shard.go
```diff
@@ -345,23 +455,28 @@ func (s *dbShard) writableSeries(id ts.ID) *dbShardEntry {
...
 		// During Rlock -> Wlock promotion the entry was inserted
-		return entry
+		return entry, nil
+	} else if err != errShardEntryNotFound {
+		s.RUnlock()
```
xichen2020: Should be `s.Unlock`

storage/shard.go
```diff
@@ -370,9 +485,9 @@ func (s *dbShard) FetchBlocks(
 	starts []time.Time,
 ) []block.FetchBlockResult {
 	s.RLock()
-	entry, _, exists := s.getEntryWithLock(id)
+	entry, _, err := s.lookupEntryWithLock(id)
 	s.RUnlock()
-	if !exists {
+	if err != nil {
```
xichen2020: Do we not want to return the actual error here if `err` is not `errShardEntryNotFound`?

storage/bootstrap.go
```diff
@@ -86,43 +86,85 @@ func newBootstrapManager(
...
 }
 
-func (m *bootstrapManager) targetRanges(writeStart time.Time) xtime.Ranges {
+func (m *bootstrapManager) targetRanges() xtime.Ranges {
+	now := m.nowFn()
 	ropts := m.opts.RetentionOptions()
-	start := writeStart.Add(-ropts.RetentionPeriod())
-	midPoint := writeStart.
+	start := now.Add(-ropts.RetentionPeriod())
+	midPoint := now.
```
robskillington: Sure thing.

storage/bootstrap.go
```diff
@@ -86,43 +86,85 @@ func newBootstrapManager(
...
+	switch m.state {
+	case bootstrapping:
+		// NB(r): Already bootstrapping, now a consequent bootstrap
+		// request comes in - we queue this up to bootstrap again
+		// once the current bootstrap has completed.
+		// This is an edge case that can occur if during either an
+		// initial bootstrap or a resharding bootstrap if a new
+		// reshard occurs and we need to bootstrap more shards.
+		m.hasPending = true
+		enqueued = true
```
robskillington: Sure thing.

storage/cluster/database.go
robskillington: We already do that during ticking.

storage/cluster/database.go
robskillington: Yeah it's meant for when you want to deploy new nodes and make them bootstrap as soon as you do a shard reassignment, rather than having to force you to do the reassignment first and make it go into a fatal/crash loop until it has a shard assignment.

storage/namespace.go
robskillington: Sure thing.

storage/cluster/database.go
```diff
@@ -0,0 +1,199 @@
...
+		}
+		d.Database.AssignShardSet(shardSet)
+	default:
+		// No updates to the topology since cluster DB created
+	}
+
+	if err := d.Database.Open(); err != nil {
+		return err
+	}
+	if err := d.startActiveTopologyWatch(); err != nil {
```
robskillington: Sure thing.

storage/cluster/database.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	return nil
+}
+
+func (d *clusterDB) Close() error {
+	if err := d.Database.Close(); err != nil {
+		return err
+	}
+	if err := d.stopActiveTopologyWatch(); err != nil {
```
robskillington: Sure thing.

storage/cluster/database.go
```diff
@@ -0,0 +1,199 @@
...
+	close(d.closedCh)
+
+	return nil
+}
+
+func (d *clusterDB) activeTopologyWatch() {
+	for {
+		select {
+		case <-d.doneCh:
+			d.closedCh <- struct{}{}
```
robskillington: Sure thing.

storage/cluster/database.go
```diff
@@ -0,0 +1,199 @@
...
+// otherwise false.
+func hostOrEmptyShardSet(
+	hostID string,
+	m topology.Map,
+) (sharding.ShardSet, bool) {
+	if hostShardSet, ok := m.LookupHostShardSet(hostID); ok {
+		return hostShardSet.ShardSet(), true
+	}
+	allShardSet := m.ShardSet()
+	shardSet, _ := sharding.NewShardSet(nil, allShardSet.HashFn())
```
robskillington: If there is non-unique shards.  Let me just create a constructor that returns a nil shard set without an error.

storage/database.go
```diff
@@ -215,6 +218,30 @@ func (d *db) Options() Options {
 	return d.opts
 }
 
+func (d *db) AssignShardSet(shardSet sharding.ShardSet) {
+	d.RLock()
+	defer d.RUnlock()
+	for _, ns := range d.namespaces {
+		ns.AssignShardSet(shardSet)
+	}
+	if d.bootstraps > 0 {
```
robskillington: It's to determine if bootstrapping or bootstrapped.  Hence did not want to create a bool called "bootstrapped" because it's really more like "beganInitialBootstrap" which is just confusing.

The int may come in handy later anyhow.

storage/namespace.go
robskillington: Soft deadline.  If we don't meet it we'll just take a little longer than we asked it to.

storage/namespace.go
robskillington: Bootstraps are such low frequency and this will only create a chan of like 16 structs on a 32 CPU machine anyhow.  I doubt this allocation will hurt overall performance of the process in any way.  There are several magnitude larger allocations that occur during a bootstrap run.

storage/shard.go
robskillington: Yeah thanks, missed this.

storage/shard.go
robskillington: Thanks, good catch!

storage/shard.go
robskillington: Thanks, good catch!

storage/shard.go
robskillington: I'm just retaining the original semantics here.

robskillington: I'll add an error to this return type.  Strange it didn't have this before.

storage/cluster/database.go
```diff
@@ -0,0 +1,285 @@
...
+
+	return nil
+}
+
+func (d *clusterDB) activeTopologyWatch() {
+	reportClosingCh := make(chan struct{}, 1)
+	reportClosedCh := make(chan struct{}, 1)
+	go func() {
+		for {
+			ticker := time.NewTicker(time.Second)
```
xichen2020: Shoudn't this be outside the loop?

sharding/types.go
```diff
@@ -22,6 +22,8 @@ package sharding
 
 import (
 	"github.com/m3db/m3db/ts"
+
+	"github.com/m3db/m3cluster/shard"
```
robskillington: Yeah, can reorder.  Think this was an auto import.

storage/cluster/database.go
robskillington: Good catch, bad refactoring here.

services/m3dbnode/main/main.go
cw9: yeah looks good

storage/cluster/database.go
cw9: yeah my bad, waiting on the watch is safe as a new watch will get a initial push

robskillington:
Will address not flushing to disk the newly bootstrapped shards with a new PR to avoid rolling too many changes into this single PR.

PR #171, opened 2016-12-02
Add mixed disk mode retention option and file seeker

martin-mao:


persist/fs/seek.go
```diff
@@ -0,0 +1,247 @@
...
+
+func (s *seeker) readDigest() error {
+	var err error
+	if s.expectedInfoDigest, err = s.digestFdWithDigestContents.ReadDigest(); err != nil {
+		return err
+	}
+	if s.expectedIndexDigest, err = s.digestFdWithDigestContents.ReadDigest(); err != nil {
+		return err
+	}
+
```
robskillington: Remove this empty line perhaps?

persist/fs/seek.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+	s.dataReader.Reset(s.dataFd)
+
+	expectedSize := markerLen + idxLen + int(entry.Size)
+	var data []byte
+	if s.bytesPool != nil {
+		data = s.bytesPool.Get(expectedSize)[:expectedSize]
+	} else {
+		data = make([]byte, expectedSize)
+	}
```
robskillington: Need to do similar thing to the file set reader and use a separate prologue buffer.  Otherwise the pointer we return from this method won't be from the bytes pool.  e.g. returning `data[markerLen+idxLen:]` instead of returning `data`

persist/fs/types.go
```diff
@@ -70,6 +70,24 @@ type FileSetReader interface {
...
+// FileSetSeeker provides an out of order reader for a TSDB file set
+type FileSetSeeker interface {
+	io.Closer
+
+	// Open opens the files for the given shard and version for reading
+	Open(namespace ts.ID, shard uint32, start time.Time) error
+
+	// Seek returns the data for specified id provided the index was loaded upon open. An
+	// error will be returned if the index was not loaded or id cannot be found
+	Seek(id string) (data []byte, err error)
```
robskillington: We should make the `id` a `ts.ID` instead of `string` to be consistent with all our other interfaces.

storage/series/series.go
```diff
@@ -155,8 +155,16 @@ func (s *dbSeries) needsBlockUpdateWithRLock() bool {
...
+	var cutoff time.Time
 	rops := s.opts.RetentionOptions()
-	cutoff := now.Add(-rops.RetentionPeriod()).Truncate(rops.BlockSize())
+	if rops.DiskMode() {
+		// Some blocks of series is stored in disk only
+		cutoff = now.Add(-rops.DiskModeInMemPeriod()).Truncate(rops.BlockSize())
+	} else {
+		// Everything for the retention period is kept in mem
+		cutoff = now.Add(-rops.RetentionPeriod()).Truncate(rops.BlockSize())
+	}
```
robskillington: This won't be extremely useful - you can achieve the same thing by just changing the retention period shorter on the nodes you are running.  We don't need an extra option to achieve that.

To achieve what you want, which is keeping the files around longer - it would be better to simply add a file cleanup offset so that it expires and writes to disk but then doesn't cleanup until it's older than `- retentionPeriod - cleanupFilesRetentionOffset`.

PR #172, opened 2016-12-02
bl refactor

ben-lerner:
Break up and refactor client/session.go. There are no changes to the logic in this diff.


client/enqueue_channel.go
```diff
@@ -0,0 +1,89 @@
...
+
+func (c *enqueueChannel) updateGauge() {
+	for atomic.LoadInt32(&c.closed) == 0 {
+		c.qGauge.Update(int64(len(c.peersMetadataCh)))
+		time.Sleep(gaugeReportInterval)
+	}
+}
+
+func (c *enqueueChannel) enqueue(peersMetadata []*blocksMetadata) {
+	atomic.AddInt32(&c.enqueued, 1)
```
robskillington: We likely want to move to a model where we aren't tracking how many are enqueued/etc as it becomes race-y detecting when it falls to zero, etc.

client/peer_blocks_queue.go
```diff
@@ -0,0 +1,171 @@
...
+		// Call done callbacks
+		for i := range doneFns {
+			doneFns[i]()
+		}
+		// Track completed blocks
+		completed := 0
+		for i := range enqueued {
+			completed += len(enqueued[i].blocks)
+		}
+		q.trackCompleted(completed)
```
robskillington: As the previous comment mentions, we hopefully shouldn't need to do this kind of accounting with a different approach.

client/topology.go
```diff
@@ -0,0 +1,231 @@
...
+
+	"github.com/m3db/m3db/encoding"
+	"github.com/m3db/m3db/topology"
+
+	"github.com/m3db/m3x/pool"
+
+	"github.com/uber-go/tally"
+)
+
+func (s *session) initHostQueues(topologyMap topology.Map) ([]hostQueue, int, int, error) {
```
robskillington: Splitting up methods that sit directly on the `session` struct is likely not great, it'll just spread the methods around some files.  

The better approach to refactoring this beast will likely be making it composition based, i.e. have a few new types that handle entire responsibility of some workflows and have the `session` delegate to them.

PR #176, opened 2016-12-06
Pool DatabaseSeries IDs

kobolog:
This PR changes the `DatabaseSeries` construction to use the ID pool instead of allocating IDs in place out of the GC heap.

### DIFF
--- a/storage/options.go
+++ b/storage/options.go
@@ -85,7 +85,8 @@ func NewSeriesOptionsFromOptions(opts Options) series.Options {
 		SetDatabaseBlockOptions(opts.DatabaseBlockOptions()).
 		SetContextPool(opts.ContextPool()).
 		SetEncoderPool(opts.EncoderPool()).
-		SetMultiReaderIteratorPool(opts.MultiReaderIteratorPool())
+		SetMultiReaderIteratorPool(opts.MultiReaderIteratorPool()).
+		SetIdentifierPool(opts.IdentifierPool())
 }
 
 type options struct {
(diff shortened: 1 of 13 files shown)


storage/options.go
```diff
@@ -280,7 +281,8 @@ func (o *options) SetEncodingM3TSZPooled() Options {
 		SetDatabaseBlockOptions(opts.blockOpts).
 		SetContextPool(contextPool).
 		SetEncoderPool(encoderPool).
-		SetMultiReaderIteratorPool(multiReaderIteratorPool)
+		SetMultiReaderIteratorPool(multiReaderIteratorPool).
+		SetIdentifierPool(opts.IdentifierPool())
```
robskillington: Hm, maybe we should just call `NewSeriesOptionsFromOptions` here?  e.g. `seriesOpts := NewSeriesOptionsFromOptions(opts)`?

storage/series/series.go
```diff
@@ -570,7 +570,8 @@ func (s *dbSeries) Close() {
 
 func (s *dbSeries) Reset(id ts.ID) {
 	s.Lock()
-	s.id = ts.BinaryID(id.Data())
+	s.id.Close()
```
robskillington: The pattern has been to put all the closing/returning logic in the `Close(...)` methods on entities.  That way they return them to the pool before they themselves are reused.  

Would be good to move `s.id.Close()` to inside the `Close()` method to follow suit.

ts/identifier.go
```diff
@@ -25,14 +25,16 @@ import (
 	"crypto/md5"
 )
 
-// BinaryID constructs a new ID based on a binary value
+// BinaryID constructs a new ID based on a binary value.
+// WARNING: Does not copy the underlying data, do not use
+// when cloning a pooled ID object.
 func BinaryID(v []byte) ID {
-	return &id{data: append(make([]byte, 0, len(v)), v...)}
+	return &id{data: v}
```
robskillington: Might it be better to keep the copy semantics but just always clone instead of `id.BinaryID(otherID.Data())`?  That way we enforce semantics and you can use `Clone()` to hit the fast path.

PR #177, opened 2016-12-06
Add read_index_ids tool

prateek:
The `index_reader` tool we whipped up for use. 
/cc @robskillington @martin-mao @xichen2020


persist/fs/seek.go
```diff
@@ -78,6 +78,17 @@ func NewSeeker(filePathPrefix string, bufferSize int, bytesPool pool.BytesPool)
 	}
 }
 
+// GetFileIds retrieves all the identifiers present in the configured file
+func (s *seeker) GetFileIds() []string {
+	fileIds := make([]string, len(s.indexMap))
+	idx := 0
+	for _, indexEntry := range s.indexMap {
+		fileIds[idx] = string(indexEntry.Id)
+		idx++
```
robskillington: To avoid needing the `idx` variable here you can do the following:
```go
ids := make([]string, 0, len(s.indexMap))
for _, entry := range s.indexMap {
	ids = append(ids, ts.BinaryID(entry.Id))
}
return ids
```

This will still avoid allocations as we create the slice with the capacity of what we will insert into it.

tools/index_reader/main/index_reader.go
```diff
@@ -0,0 +1,84 @@
...
+	err := seeker.Open(ts.StringID(*optNamespace), uint32(shardID), time.Unix(0, int64(blockstartInt)))
+
+	if err != nil {
+		fmt.Println("Unable to open file: ", err)
+		os.Exit(1)
+	}
+
+	fileIds := seeker.GetFileIds()
+	if fileIds == nil {
+		fmt.Printf("Unable to get ids")
```
robskillington: Probably best to use logger similar to `read_ids` here:

```go
import xlog "github.com/m3db/m3x/log"

// ...

log := xlog.NewLogger(os.Stderr)
```

This way only real output goes to `stdout` and log output goes to `stderr` to make `stdout` just purely "results".

tools/index_reader/main/index_reader.go
```diff
@@ -0,0 +1,84 @@
...
+
+	if len(*optNamespace) == 0 {
+		getopt.Usage()
+		os.Exit(1)
+	}
+
+	if len(*optShard) == 0 {
+		getopt.Usage()
+		os.Exit(1)
+	}
```
robskillington: Can compact this to make it similar to `read_ids`:

```go
	flag.Parse()

	if *optPathPrefix == "" ||
		*optNamespace == "" ||
		optShard < 0  {
		flag.Usage()
		os.Exit(1)
	}
```

tools/index_reader/main/index_reader.go
```diff
@@ -0,0 +1,84 @@
...
+	defaultBufferReadSize  = 10
+	defaultBufferCapacity  = 1024 * 1024 * 1024
+	defaultBufferPoolCount = 10
+)
+
+func main() {
+	var (
+		optPathPrefix = getopt.StringLong("path-prefix", 'p', "", "Path prefix [e.g. /var/lib/m3db]")
+		optNamespace  = getopt.StringLong("namespace", 'n', "", "Namespace [e.g. metrics]")
+		optShard      = getopt.StringLong("shard-id", 's', "", "Shard ID [expected format uint32]")
```
robskillington: Better to use an int var here rather than parsing below.

tools/index_reader/main/index_reader.go
```diff
@@ -0,0 +1,84 @@
...
+		optBlockstart = getopt.StringLong("block-start", 'b', "", "Block Start Time [in nsec]")
```
robskillington: Same for blockstart.

Makefile
```diff
@@ -32,8 +32,9 @@ VENDOR_ENV := GO15VENDOREXPERIMENT=1
 SERVICES := \
 	m3dbnode
 
-TOOLS := \
-	read_ids
+TOOLS :=   \
+	read_ids \
+	index_reader
```
robskillington: For slightly more consistent naming maybe it makes sense to make this `read_index_ids`?

persist/fs/seek.go
```diff
@@ -78,6 +78,17 @@ func NewSeeker(filePathPrefix string, bufferSize int, bytesPool pool.BytesPool)
 	}
 }
 
+// GetFileIds retrieves all the identifiers present in the configured file
```
prateek: Yep, sounds good. 

I added em in because golint was unhappy, do we have a config to control the scope of what it runs on?

Makefile
prateek: Yep, sounds good.

persist/fs/types.go
```diff
@@ -86,6 +86,9 @@ type FileSetSeeker interface {
 
 	// Entries returns the count of entries in the volume
 	Entries() int
+
+	// GetFileIds retrieves all the identifiers present in the file set
+	GetFileIds() []string
```
prateek: Will do.

persist/fs/seek.go
prateek: noice!

persist/fs/seek.go
robskillington: On interesting, golint should realize that it's an interface method implementation... Hm.  Was this before it was added to the `Seeker` interface perhaps?  Or still complains about it now if this comment is removed?

prateek: Yeah I think that's it.

tools/read_index_ids/main/main.go
```diff
@@ -0,0 +1,57 @@
+package main
+
+import "fmt"
```
robskillington: Perhaps move this import into the main imports below?

tools/read_index_ids/main/main.go
```diff
@@ -0,0 +1,57 @@
...
+
+	err := seeker.Open(ts.StringID(*optNamespace), uint32(*optShard), time.Unix(0, int64(*optBlockstart)))
+
+	if err != nil {
+		log.Fatalf("Unable to open file: %v", err)
+	}
+
+	fileIds := seeker.IDs()
+	if fileIds == nil {
+		log.Fatalf("Unable to retrieve ids")
```
robskillington: Minor minor nit: all our log messages currently begin with lower case, this seems to be something of a standard common practice in other Go services/applications too, might be worth conforming these log statements too.

Also for this specific log message, perhaps "no ids in index" is more correct?  If the seeker opened correctly then the file format and the header structures are all valid, there's just no IDs in the file itself.

prateek: Yeah agreed on both points.

tools/read_index_ids/main/main.go
prateek: :| my reliance on gofmt is going down

PR #178, opened 2016-12-06
Protect ID hash from data races

xichen2020:
cc @robskillington @kobolog @martin-mao @cw9 @prateek @ben-lerner 

This PR adds logic to protect ID hashes from data races.

Previously there is a race condition where two goroutines have reference to the same ID object, now if the first goroutine calls `id.Hash()`, it'll realize that the hash hasn't been initialized and starts computing. It'll then proceed to set the hash. However if halfway through setting the hash, the second goroutine also calls `id.Hash()`, it'll see a half-filled hash array and think the hash has been initialized, and return that hash even though it's incomplete. This in turn will trigger a crash in the client code.

The reason why we don't use a RWLock is to save space (4 bytes compared to 24 bytes needed for a RWLock object). This overhead adds up given we have a large number of IDs.

### DIFF
--- a/ts/identifier.go
+++ b/ts/identifier.go
@@ -23,6 +23,8 @@ package ts
 import (
 	"bytes"
 	"crypto/md5"
+	"fmt"
+	"sync/atomic"
 )
 
 // BinaryID constructs a new ID based on a binary value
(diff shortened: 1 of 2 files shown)


ts/identifier.go
```diff
@@ -54,11 +66,26 @@ var null = Hash{}
...
+		return md5.Sum(v.data)
+	case uninitialized:
+		// If the id hash is unitialized, and this goroutine gains exclusive
+		// access to the hash field, computes the hash and sets the hash
+		if atomic.CompareAndSwapInt32(&v.state, int32(uninitialized), int32(computing)) {
+			v.hash = md5.Sum(v.data)
+			return v.hash
+		}
+		// Otherwise compute the hash without waiting
+		return md5.Sum(v.data)
```
xichen2020: sure

PR #181, opened 2016-12-07
Check IsBootstrapped per shard for data access

cw9:
@robskillington


robskillington:
Let's perhaps change the initializing shards after truncation to pass `false` to needing bootstrap to the new shards.

PR #182, opened 2016-12-08
Add a log when marking shard as available

cw9:


storage/cluster/database.go
```diff
@@ -252,6 +252,7 @@ func (d *clusterDB) analyzeAndReportShardStates() {
 			d.log.Errorf("cluster db failed marking shard %d available: %v",
 				id, err)
 		}
+		d.log.Infof("successfully marked shard %d available", id)
```
robskillington: This should be in an `else` branch yeah?

cw9: lol yeah, fixing it

PR #183, opened 2016-12-09
Move flush state into shards to allow for flushing shards new & modified

robskillington:
cc @xichen2020 @cw9 @martin-mao @kobolog @ben-lerner @prateek

### DIFF
--- a/storage/flush.go
+++ b/storage/flush.go
@@ -21,17 +21,23 @@
 package storage
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 	"time"
 
 	"github.com/m3db/m3db/clock"
-	"github.com/m3db/m3x/instrument"
 	"github.com/m3db/m3db/persist"
 	"github.com/m3db/m3db/ratelimit"
+	"github.com/m3db/m3db/retention"
 	"github.com/m3db/m3x/errors"
+	"github.com/m3db/m3x/instrument"
 
 	"github.com/uber-go/tally"
+)
+
+var (
+	errFlushAlreadyInProgress = errors.New("flush already in progress")
 )
 
 type flushManager struct {
@@ -42,7 +48,6 @@ type flushManager struct {
 	nowFn           clock.NowFn
 	blockSize       time.Duration
 	pm              persist.Manager
-	flushStates     map[time.Time]fileOpState
 	flushInProgress bool
 	metrics         instrument.MethodMetrics
 }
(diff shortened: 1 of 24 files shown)


storage/namespace.go
```diff
@@ -446,14 +446,61 @@ func (n *dbNamespace) Flush(
...
+			// If any in progress or any not started then make it appear as
+			// namespace has the same status
+			return state
+		}
+		if !anyNumFailures && state.NumFailures > 0 {
+			anyNumFailures = true
+			result.NumFailures = state.NumFailures
+		} else if state.NumFailures < result.NumFailures {
+			// Prefer reporting we have less failures, as this will determine
+			// if a retry is attempted or not - this means max retries is not
```
cw9: seems the usage of MaxRetries() no longer means "max retries" now? should probably create a new field/ rename in options? 

Also does this mean it could retry forever if it keeps failing to flush?

storage/namespace.go
```diff
@@ -446,14 +446,61 @@ func (n *dbNamespace) Flush(
...
+			result.Status = fileOpFailed
+		case fileOpNotStarted, fileOpInProgress:
+			// If any in progress or any not started then make it appear as
+			// namespace has the same status
+			return state
+		}
+		if !anyNumFailures && state.NumFailures > 0 {
+			anyNumFailures = true
+			result.NumFailures = state.NumFailures
+		} else if state.NumFailures < result.NumFailures {
```
robskillington: Sure thing, sounds good.

storage/namespace.go
cw9: Got it, thanks for the explaination

storage/flush.go
```diff
@@ -154,42 +129,46 @@ func (m *flushManager) RateLimitOptions() ratelimit.Options {
...
-	flushState, exists := m.flushStates[t]
-	if !exists {
-		return true
+// needsFlush returns true if we need to flush data for a given time.
+func (m *flushManager) needsFlush(t time.Time) bool {
+	maxRetries := m.opts.MaxFlushRetries()
+	namespaces := m.database.getOwnedNamespaces()
+	for _, n := range namespaces {
+		state := n.FlushState(t)
+		if state.Status != fileOpSuccess &&
```
xichen2020: Might be worth moving this logic to `needsFlush` in the namespace implementation for better encapsulation

storage/namespace.go
```diff
@@ -443,14 +443,53 @@ func (n *dbNamespace) Flush(
...
+		}
+		state := shard.FlushState(blockStart)
+		status := state.Status
+		if status == fileOpNotStarted || status == fileOpInProgress {
+			// Any not started or in progress overrides other states
+			return state
+		}
+		if status == fileOpFailed {
+			anyFailed = true
+			if state.NumFailures < minNumFailures {
```
xichen2020: Seems like if we make the decision whether we should flush here in namespace.go as opposed to passing back the state to the caller, should make it less confusing?

storage/namespace.go
```diff
@@ -467,29 +464,25 @@ func (n *dbNamespace) FlushState(blockStart time.Time) fileOpState {
...
+		if state.Status == fileOpInProgress {
+			return false
 		}
-		if status == fileOpFailed {
+		if state.Status == fileOpNotStarted {
+			return true
+		}
+		if state.Status == fileOpFailed {
 			anyFailed = true
 			if state.NumFailures < minNumFailures {
```
xichen2020: hm why not compare `state.NumFailures` with `n.opts.MaxFlushRetries` and if it's smaller, return true immediatley? Don't think we actually need the `anyFailed` flag

PR #185, opened 2016-12-12
[WIP] Bl/inflight quorum

ben-lerner:
[WIP] - Logic is done and probably works, preliminary feedback would be appreciated. Currently fixing old tests and adding new ones.

Add support for bootstrapping to the write client. While bootstrapping the initializing and leaving shards count as a single shard towards a quorum, and only if the client successfully writes to both of them.


xichen2020:
Macro comment: Looks like there are both refactoring changes (which are presumably no-op changes as far as functionality is concerned) and functionality changes in this PR. Would be nice if we can separate those into two diffs to make easier for reviewers.

client/metadata_iter.go
```diff
@@ -0,0 +1,76 @@
...
+		return false
+	}
+	m, more := <-it.inputCh
+	if !more {
+		it.err = <-it.errCh
+		it.done = true
+		return false
+	}
+	it.host = m.peer.Host()
+	var zeroed block.Metadata
```
martin-mao: Why not just use nil or move this to const somewhere?

client/enqueue_channel.go
```diff
@@ -0,0 +1,90 @@
...
+
+func newEnqueueChannel(m *streamFromPeersMetrics) *enqueueChannel {
+	c := &enqueueChannel{
+		peersMetadataCh: make(chan []*blocksMetadata, 4096),
+		closed:          0,
+		metrics:         m,
+	}
+	go func() {
+		for atomic.LoadInt64(&c.closed) == 0 {
+			m.blocksEnqueueChannel.Update(int64(len(c.peersMetadataCh)))
```
martin-mao: Is this not atomic.LoadUint64(&c.enqueued) - atomic.LoadUint64(&c.processed)?

client/host_queue.go
```diff
@@ -192,37 +190,36 @@ func (q *queue) drain() {
...
-				currBatchElementsByNamespace[namespace.Hash()] = currBatchElements
+				currWriteOpsByNamespace[hash] = currWriteOps
+				currBatchElementsByNamespace[hash] = currBatchElements
 
 				if len(currWriteOps) == writeBatchSize {
 					// Reached write batch limit, write async and reset
 					q.asyncWrite(wgAll, namespace, currWriteOps, currBatchElements)
-					currWriteOpsByNamespace[namespace.Hash()] = nil
-					currBatchElementsByNamespace[namespace.Hash()] = nil
+					currWriteOpsByNamespace[hash] = nil
```
kobolog: This and in other places should probably be delete(currWriteOpsByNamespace, hash), I think?

robskillington:
Hey some of the refactors are good but conflict with changes that have already been merged to master.  Can you reduce this to the minimal change to implement this?  We can do some of the refactors that you've done here as a separate independent change.  

This will help us ship faster and keep this change to a minimal set we can audit and review more carefully.

ben-lerner:
Will do.

PR #186, opened 2016-12-12
Write quorum logic during bootstrapping

ben-lerner:
Add support for bootstrapping to the write client. While bootstrapping the initializing and leaving shards count as a single shard towards a quorum, and only if the client successfully writes to both of them.

### DIFF
--- a/client/connection_pool_test.go
+++ b/client/connection_pool_test.go
@@ -35,9 +35,13 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
+const (
+	testHost     = "testhost"
+	testHostAddr = testHost + ":9000"
+)
+
 var (
-	testHost    = "testhost"
-	h           = topology.NewHost(testHost, testHost+":9000")
+	h           = topology.NewHost(testHost, testHostAddr)
 	channelNone = &nullChannel{}
 )
 
--- a/client/connection_pool_test.go
+++ b/client/connection_pool_test.go
@@ -29,14 +29,15 @@ import (
 
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	"github.com/m3db/m3db/topology"
-	"github.com/m3db/m3x/close"
+	xclose "github.com/m3db/m3x/close"
 	"github.com/uber/tchannel-go/thrift"
 
 	"github.com/stretchr/testify/assert"
 )
 
 var (
-	h           = topology.NewHost("testhost", "testhost:9000")
+	testHost    = "testhost"
+	h           = topology.NewHost(testHost, testHost+":9000")
 	channelNone = &nullChannel{}
 )
 
--- a/client/host_queue.go
+++ b/client/host_queue.go
@@ -275,10 +275,13 @@ func (q *queue) asyncWrite(
 			q.Done()
 		}
 
+		hostID := q.host.ID() // NB(bl): this is passed to writeState to
+		// determine the state of the shard on the node we're writing to
+
 		client, err := q.connPool.NextClient()
 		if err != nil {
 			// No client available
-			callAllCompletionFns(ops, nil, err)
+			callAllCompletionFns(ops, hostID, err)
 			cleanup()
 			return
 		}
@@ -287,7 +290,7 @@ func (q *queue) asyncWrite(
 		err = client.WriteBatchRaw(ctx, req)
 		if err == nil {
 			// All succeeded
-			callAllCompletionFns(ops, nil, nil)
+			callAllCompletionFns(ops, hostID, nil)
 			cleanup()
 			return
 		}
@@ -297,14 +300,14 @@ func (q *queue) asyncWrite(
 			hasErr := make(map[int]struct{})
 			for _, batchErr := range batchErrs.Errors {
 				op := ops[batchErr.Index]
-				op.CompletionFn()(nil, batchErr.Err)
+				op.CompletionFn()(hostID, batchErr.Err)
 				hasErr[int(batchErr.Index)] = struct{}{}
 			}
 			// Callback all writes with no errors
 			for i := range ops {
 				if _, ok := hasErr[i]; !ok {
 					// No error
-					ops[i].CompletionFn()(nil, nil)
+					ops[i].CompletionFn()(hostID, nil)
 				}
 			}
 			cleanup()
@@ -312,7 +315,7 @@ func (q *queue) asyncWrite(
 		}
 
 		// Entire batch failed
-		callAllCompletionFns(ops, nil, err)
+		callAllCompletionFns(ops, hostID, err)
 		cleanup()
 	}()
 }
--- a/client/host_queue.go
+++ b/client/host_queue.go
@@ -275,13 +275,13 @@ func (q *queue) asyncWrite(
 			q.Done()
 		}
 
-		hostID := q.host.ID() // NB(bl): this is passed to writeState to
-		// determine the state of the shard on the node we're writing to
+		// NB(bl): host is passed to writeState to determine the state of the
+		// shard on the node we're writing to
 
 		client, err := q.connPool.NextClient()
 		if err != nil {
 			// No client available
-			callAllCompletionFns(ops, hostID, err)
+			callAllCompletionFns(ops, q.host, err)
 			cleanup()
 			return
 		}
@@ -290,7 +290,7 @@ func (q *queue) asyncWrite(
 		err = client.WriteBatchRaw(ctx, req)
 		if err == nil {
 			// All succeeded
-			callAllCompletionFns(ops, hostID, nil)
+			callAllCompletionFns(ops, q.host, nil)
 			cleanup()
 			return
 		}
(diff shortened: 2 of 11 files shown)


client/connection_pool_test.go
```diff
@@ -29,14 +29,15 @@ import (
...
+	xclose "github.com/m3db/m3x/close"
 	"github.com/uber/tchannel-go/thrift"
 
 	"github.com/stretchr/testify/assert"
 )
 
 var (
-	h           = topology.NewHost("testhost", "testhost:9000")
+	testHost    = "testhost"
+	h           = topology.NewHost(testHost, testHost+":9000")
```
kobolog: I'd prefer either both parts to be named constants or none, also `fmt` > string concatenation.

client/write.go
```diff
@@ -117,25 +121,54 @@ func (w *writeState) reset() {
...
-		w.Unlock()
+		w.addError(err)
+	} else if hostShardSet, ok := w.topoMap.LookupHostShardSet(hostID); !ok {
+		w.addError(fmt.Errorf("Missing host shard in writeState completionFn: %s", hostID))
+	} else if shardState, err :=
+		hostShardSet.ShardSet().LookupStateByID(w.op.shardID); err != nil {
+		w.addError(fmt.Errorf("Missing shard %d in host %s", w.op.shardID, hostID))
 	} else {
-		successful = atomic.AddInt32(&w.successful, 1)
+		// NB(bl): We use one variable, halfSuccess, to avoid atomically
```
kobolog: Honestly, this logic, while being correct, isn't really straightforward and easy to comprehend. Arguably, I'd go with getting a slight performance hit of an extra atomic operation vs. writing hard to understand code. Thoughts, @robskillington?

sharding/shardset.go
```diff
@@ -56,20 +60,32 @@ func NewEmptyShardSet(fn HashFn) ShardSet {
...
 
 func (s *shardSet) Lookup(identifier ts.ID) uint32 {
 	return s.fn(identifier)
 }
 
+func (s *shardSet) LookupStateByID(shardID uint32) (shard.State, error) {
+	hostShard, ok := s.shardMap[shardID]
+	if !ok {
+		var state shard.State
+		return state, ErrInvalidShardID
```
kobolog: `return shard.State{}, ErrInvalidShardID`? Also, the whole thing can be collapsed into one compound if statement.

client/write.go
ben-lerner: imo, the confusing part of the logic is the halfSuccess thing, and that won't go away even if broken up into two ints. It'd be clearest to swap it for a success int, a leavingWrite int, and an initializingWrite int, but at that point I think it's code bloat.

kobolog: Further discussed this with @ben-lerner, maybe it actually would make sense to drop atomics here completely, with the reasoning as follows:

- The `writeState` object is not contended between many goroutines, as I understand it it will only be shared among N goroutines where N == number of requested replicas.
- It is already a Lockable and it is already being locked for enqueueing, for recording errors and for completion invocation.
- I'd argue that there'll be minor to no performance hit from converting integral result variables into plain integers and mutating them under the lock.

client/write.go
```diff
@@ -117,25 +121,54 @@ func (w *writeState) reset() {
...
+	w.errors = append(w.errors, err)
+	w.Unlock()
+}
+
+// todo@bl - modify comment on what happens to inflight writes
 func (w *writeState) completionFn(result interface{}, err error) {
 	var (
 		successful int32
 		remaining  = atomic.AddInt32(&w.pending, -1)
+		hostID     = result.(string)
```
cw9: what about passing the shard state or even score instead of passing hostID in the result argument?

client/connection_pool_test.go
ben-lerner: Changed it to two constants. Go doesn't recognize the output of fmt.Sprintf as a constant, unfortunately, so it's still using string concatenation.

sharding/shardset.go
ben-lerner: shard.State{} doesn't work for type aliases, it would have to be
`return *new(shard.State)` afaict.

client/write.go
ben-lerner: We should avoid multiple atomics to prevent races. I don't have a strong opinion about sticking with one atomic or adding more locking, though I prefer the former for being less code. 

@other-reviewers Please lmk if the code is too confusing as-is, and I'll update it.

client/write.go
ben-lerner: The host queue doesn't currently know anything about shards, which I think is the correct separation of concerns. If we wanted to get the shard state or a score earlier, the host queue would need to inspect the writeOp for the shardID and contain logic on getting and scoring shard states. 

Alternatively, we could also add a "score" function to the writeOp that takes a hostID and use that from the host queue, but that's needlessly roundabout.

sharding/shardset.go
ben-lerner: also, turns out the linter doesn't like compressing it into one statement.

client/host_queue.go
```diff
@@ -275,10 +275,13 @@ func (q *queue) asyncWrite(
 			q.Done()
 		}
 
+		hostID := q.host.ID() // NB(bl): this is passed to writeState to
+		// determine the state of the shard on the node we're writing to
```
robskillington: Might it be worth passing in the `topology.Host` so we have more context of the host that is completing a write request?

ben-lerner: Good point, done.

sharding/shardset.go
```diff
@@ -56,20 +60,32 @@ func NewEmptyShardSet(fn HashFn) ShardSet {
...
+		shardMap: shardMap,
+		fn:       fn,
 	}
 }
 
 func (s *shardSet) Lookup(identifier ts.ID) uint32 {
 	return s.fn(identifier)
 }
 
+func (s *shardSet) LookupStateByID(shardID uint32) (shard.State, error) {
```
cw9: Might be better to have the method to return the Shard instead of just the State? in case there are use cases in the future that also want to access the SourceID field

client/write.go
```diff
@@ -131,31 +136,29 @@ func (w *writeState) close() {
...
-	case topology.ConsistencyLevelMajority:
-		if successful >= w.majority || remaining == 0 {
-			w.Signal()
-		}
-	case topology.ConsistencyLevelAll:
-		if remaining == 0 {
-			w.Signal()
-		}
+	wLevel := w.session.writeLevel
+	if w.pending == 0 ||
```
ben-lerner: I don't think there's a delay, the two versions should be logically equivalent. Did I mess something up?

cw9: yeah you are right, sorry I'm blind

client/write.go
```diff
@@ -98,11 +100,12 @@ type writeState struct {
...
-	session             *session
-	op                  *writeOp
-	majority            int32
-	successful, pending int32
-	errors              []error
+	session           *session
+	topoMap           topology.Map
+	op                *writeOp
+	majority, pending int32
+	success           int32
```
robskillington: If you're using locks now, you can make these just `int`s.

client/write.go
```diff
@@ -117,42 +120,37 @@ func (w *writeState) reset() {
...
-		w.Unlock()
-	} else {
-		successful = atomic.AddInt32(&w.successful, 1)
+	} else if hostShardSet, ok := w.topoMap.LookupHostShardSet(hostID); !ok {
+		w.errors = append(w.errors, fmt.Errorf("Missing host shard in writeState completionFn: %s", hostID))
+	} else if shardState, err :=
+		hostShardSet.ShardSet().LookupStateByID(w.op.shardID); err != nil {
+		w.errors = append(w.errors, fmt.Errorf("Missing shard %d in host %s", w.op.shardID, hostID))
+	} else if shardState == shard.Available {
+		// NB(bl): only count writes to available shards towards success
```
robskillington: Correct yeah.

client/write.go
```diff
@@ -131,31 +136,29 @@ func (w *writeState) close() {
...
 	w.Lock()
+	w.pending--
 
 	if err != nil {
 		w.errors = append(w.errors, err)
+	} else if hostShardSet, ok := w.topoMap.LookupHostShardSet(hostID); !ok {
+		w.errors = append(w.errors, fmt.Errorf("Missing host shard in writeState completionFn: %s", hostID))
+	} else if shardState, err :=
+		hostShardSet.ShardSet().LookupStateByID(w.op.shardID); err != nil {
+		w.errors = append(w.errors, fmt.Errorf("Missing shard %d in host %s", w.op.shardID, hostID))
```
robskillington: To stay consistent with our other errors make these both start with lower case characters yeah?

sharding/shardset.go
```diff
@@ -56,20 +60,32 @@ func NewEmptyShardSet(fn HashFn) ShardSet {
...
+	hostShard, ok := s.shardMap[shardID]
+	if !ok {
+		var noState shard.State
+		return noState, ErrInvalidShardID
```
robskillington: A little simpler would be: `return shard.State(0), ErrInvalidShardID`

client/write.go
ben-lerner: They interact with a bunch of int32's in session.go, so this avoids casting all over the place.

client/write.go
ben-lerner: No strong preference, rolling back.

PR #187, opened 2016-12-12
Use active host queues for peer operations in client sessions

robskillington:
cc @cw9 @xichen2020 @martin-mao


client/session.go
```diff
@@ -1672,14 +1703,12 @@ func (s *session) streamBlocksBatchFromPeer(
...
 	if err := retrier.Attempt(func() error {
-		client, err := peer.ConnectionPool().NextClient()
-		if err != nil {
-			return err
-		}
-
-		tctx, _ := thrift.NewContext(s.streamBlocksBatchTimeout)
-		result, err = client.FetchBlocksRaw(tctx, req)
-		return err
+		var attempErr error
```
xichen2020: `attemptErr` perhaps?

PR #188, opened 2016-12-13
Fix buffer encoders being closed before all upstream callers done

robskillington:
cc @xichen2020 @cw9 @martin-mao

### DIFF
--- a/encoding/m3tsz/encoder.go
+++ b/encoding/m3tsz/encoder.go
@@ -571,13 +571,13 @@ func (enc *encoder) Close() {
 	enc.writable = false
 	enc.closed = true
 
+	// Reset the ostream to avoid reusing this encoder
+	// using the buffer we are returning to the pool
+	enc.os.Reset(nil)
+
 	bytesPool := enc.opts.BytesPool()
 	if bytesPool != nil {
 		buffer, _ := enc.os.Rawbytes()
-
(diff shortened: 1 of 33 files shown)


storage/series/series.go
```diff
@@ -379,96 +372,93 @@ func (s *dbSeries) FetchBlocksMetadata(
...
+	defer stream.Close()
+
+	reader, err := existingBlock.Stream(ctx)
 	if err != nil {
 		return err
 	}
+
+	var readers [2]io.Reader
+	readers[0] = stream
 	readers[1] = reader
```
xichen2020: Do we need to close `reader` too?

robskillington: That would be ideal to avoid leaking, good catch.

encoding/m3tsz/encoder.go
```diff
@@ -106,6 +107,9 @@ func initialTimeUnit(start time.Time, tu xtime.Unit) xtime.Unit {
 
 // Encode encodes the timestamp and the value of a datapoint.
 func (enc *encoder) Encode(dp ts.Datapoint, tu xtime.Unit, ant ts.Annotation) error {
+	if enc.closed {
+		return errEncoderClosed
+	}
 	if !enc.writable {
```
xichen2020: Don't think this or the `Seal` method is needed either

PR #189, opened 2016-12-14
Add host description to host queue errors

ben-lerner:
Add host IDs to host queue errors.

### DIFF
--- a/client/host_queue.go
+++ b/client/host_queue.go
@@ -456,20 +456,14 @@ func (q *queue) Close() {
 
 // errors
 
-type errQueueNotOpen string
-
-func (s errQueueNotOpen) Error() string {
-	return fmt.Sprintf("host operation queue not open for host: %s", string(s))
+func errQueueNotOpen(hostID string) error {
+	return fmt.Errorf("host operation queue not open for host: %s", hostID)
 }
 
-type errQueueUnknownOperation string
-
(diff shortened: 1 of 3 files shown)


client/host_queue.go
```diff
@@ -456,3 +450,23 @@ func (q *queue) Close() {
 	close(q.drainIn)
 	q.Unlock()
 }
+
+// errors
+
+type errQueueNotOpen string
```
robskillington: Instead of implementing the `error` interface, might it not be better to just make these methods create errors using the standard `fmt.Errorf`? 

e.g. 
```go
func newErrQueueNotOpen(hostID string) error {
   return fmt.Errorf("host operation queue not open for host: %s", hostID)
}

func newErrQueueUnknownOperation(hostID string) {
    return fmt.Errorf("host operation queue received unknown operation for host: %s", hostID)
}
```

ben-lerner: Implementing them as types is the [recommended Go approach](https://blog.golang.org/error-handling-and-go) to be able to check for specific errors, which we do in tests. Are we okay dropping that to make the change?

ben-lerner: Updated.

robskillington:
Seems like the errors returned from `asyncWrite(...)` aren't attached with the host details?  This is the main errors we want to decorate so we know which hosts are returning bad errors.

PR #193, opened 2016-12-15
Refactor context.Context

kobolog:
This PR refactors `context.Context` to avoid having a pointer to the `dependecies` object. In practice, all Contexts always have it initialized eventually – after a few minutes of runtime – and we have tens of millions of Contexts active. This effectively **reduces the number of active pointers in the GC heap by almost 20%**.

This PR also refactors Closers concept and renames it to Finalizers to avoid a naming ambiguity with regards to `Close()`/`OnClose()` calls. The proposed naming is still not perfect but in my head is less confusing. Up for a discussion. An outstanding issue with this mechanics is the `context.CloserFn` conversion that takes up a considerable portion of total allocations but in essence is a no-op, i.e. an object with `Close()` is effectively is a Closer already, but a conversion has to be made because of the `Close()`/`OnClose()` disparity.

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -24,11 +24,11 @@ import "sync"
 
 // NB(r): using golang.org/x/net/context is too GC expensive.
 type ctx struct {
-	sync.WaitGroup
 	sync.RWMutex
 
 	pool       contextPool
 	done       bool
+	wg         sync.WaitGroup
 	finalizers []Finalizer
 }
 
@@ -75,7 +75,7 @@ func (c *ctx) DependsOn(blocker Context) {
 	c.Lock()
 
 	if !c.done {
-		c.Add(1)
+		c.wg.Add(1)
 		blocker.RegisterFinalizer(c)
 	}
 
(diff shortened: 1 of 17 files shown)


context/context.go
```diff
@@ -20,140 +20,141 @@
...
 		c.Unlock()
 		return
 	}
-	c.closed = true
-	if c.dep != nil {
-		closers = c.dep.closers[:]
-	}
+
+	c.done = true
+	finalizers := c.finalizers[:]
```
kobolog: This code is a little bit suspicious to me – do we assume that somewhere at some point we call `ctx.Close(); ctx.Reset()`, i.e. we need to keep another reference to the finalizer slice alive because `c.finalizers` can somehow go away?

xichen2020: hmm looks like a race condition, i.e., if `finalizers` is empty, but then someone gets in and adds a dependency between line 113 and line 115, we'll never wait on that dependency and return the context straightly to pool without calling the finalizer

kobolog: But we shouldn't get new finalizers since we set `c.done = true` and that should stop `RegisterFinalizer` before doing any harm.

xichen2020: Ah yeah good point

kobolog: So yeah that's why I was wondering why this code is there at all.

PR #195, opened 2016-12-19
Return copied byte streams for x/io.SegmentReader to pool

robskillington:
cc @xichen2020 @kobolog @cw9

### DIFF
--- a/x/io/reader.go
+++ b/x/io/reader.go
@@ -156,6 +156,15 @@ func (sr *segmentReader) Reset(segment ts.Segment) {
 }
 
 func (sr *segmentReader) Close() {
+	// Return head and tail to pools if their pools are set on the segment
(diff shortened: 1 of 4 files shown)


x/io/reader.go
```diff
@@ -156,6 +156,15 @@ func (sr *segmentReader) Reset(segment ts.Segment) {
 }
 
 func (sr *segmentReader) Close() {
+	// Return head and tail to pools if their pools are set on the segment
+	seg := sr.segment
```
xichen2020: The ownership semantics used to be that the reader doesn't own the segment, so this is now changed. I did a quick search across the codebase, doesn't look this new ownership semantics should break anything, but might be cautious to comb through the codebase a bit more carefully.

PR #196, opened 2016-12-19
Add peer block success/error fetch metrics

robskillington:
cc @xichen2020

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -1704,7 +1704,7 @@ func (s *session) streamBlocksBatchFromPeer(
 			}
 			starts = append(starts, blockStart.UnixNano())
 		}
-		reqBlocksLen++
+		reqBlocksLen += int64(len(starts))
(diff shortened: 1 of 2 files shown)


client/session.go
```diff
@@ -1698,6 +1704,7 @@ func (s *session) streamBlocksBatchFromPeer(
 			}
 			starts = append(starts, blockStart.UnixNano())
 		}
+		reqBlocksLen++
```
xichen2020: Should this be inside the for loop above?

client/session.go
```diff
@@ -1730,6 +1738,7 @@ func (s *session) streamBlocksBatchFromPeer(
 	// Parse and act on result
 	for i := range result.Elements {
 		if i >= len(batch) {
+			m.fetchBlockError.Inc(1)
```
xichen2020: Do we want to distinguish errors for batch of blocks vs. individual block errors? Right now feels like `fetchBlockError` is used for both

PR #198, opened 2016-12-20
Avoid discarding byte slices on drain and reset

robskillington:
cc @xichen2020 @kobolog

### DIFF
--- a/encoding/m3tsz/encoder.go
+++ b/encoding/m3tsz/encoder.go
@@ -487,11 +487,18 @@ func (enc *encoder) newBuffer(capacity int) []byte {
 }
 
 func (enc *encoder) Reset(start time.Time, capacity int) {
-	enc.ResetSetData(start, enc.newBuffer(capacity))
+	enc.ResetSetData(start, ts.Segment{Head: enc.newBuffer(capacity)})
 }
 
-func (enc *encoder) ResetSetData(start time.Time, data []byte) {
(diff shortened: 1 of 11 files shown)


encoding/m3tsz/encoder.go
```diff
@@ -487,11 +487,18 @@ func (enc *encoder) newBuffer(capacity int) []byte {
...
+	enc.ResetSetData(start, ts.Segment{Head: enc.newBuffer(capacity)})
 }
 
-func (enc *encoder) ResetSetData(start time.Time, data []byte) {
-	enc.os.Reset(data)
+func (enc *encoder) ResetSetData(start time.Time, data ts.Segment) ts.Segment {
+	// Return what this encoder owned by reference
+	segment := enc.segment(byRefResultType)
+
+	bytes := data.Head
```
xichen2020: This changes the byte slices in `data`, is this ok?

storage/series/buffer.go
```diff
@@ -161,14 +161,11 @@ func (b *dbBuffer) DrainAndReset() {
 
 			// After we merge there is always only a single encoder with all data
 			encoder := bucket.encoders[0].encoder
-			if stream := encoder.Stream(); stream != nil {
+			data := encoder.ResetSetData(current, ts.Segment{})
```
kobolog: Why not simply `encoder.Discard()`? That matches the semantics of the call.

PR #200, opened 2016-12-20
Avoid keeping reference to returned byte slice on close

robskillington:
cc @kobolog @xichen2020

### DIFF
--- a/encoding/m3tsz/encoder.go
+++ b/encoding/m3tsz/encoder.go
@@ -480,9 +480,6 @@ func (enc *encoder) trackNewSig(numSig uint8) uint8 {
 }
 
 func (enc *encoder) newBuffer(capacity int) []byte {
-	if capacity < 1 {
-		return nil
-	}
 	if bytesPool := enc.opts.BytesPool(); bytesPool != nil {
 		return bytesPool.Get(capacity)
 	}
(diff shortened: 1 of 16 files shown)


encoding/m3tsz/encoder.go
```diff
@@ -480,26 +480,20 @@ func (enc *encoder) trackNewSig(numSig uint8) uint8 {
 }
 
 func (enc *encoder) newBuffer(capacity int) []byte {
+	if capacity < 1 {
```
kobolog: I think we should make it a part of BytesPool API.

encoding/m3tsz/encoder.go
```diff
@@ -480,26 +480,20 @@ func (enc *encoder) trackNewSig(numSig uint8) uint8 {
...
-	segment := enc.segment(byRefResultType)
-
-	bytes := data.Head
-	if data.Tail != nil {
-		// NB(r): If you pass a tail this will cause a copy if
-		// head cannot fit the tail
-		bytes = append(bytes, data.Tail...)
-	}
+func (enc *encoder) reset(start time.Time, bytes []byte) {
 	enc.os.Reset(bytes)
```
kobolog: Is it possible that we end up calling `Reset(someTime, someSize)` while encoder is already holding on to some buffer? If yes, then perhaps we should add something like `if rb, _ := enc.os.Rawbytes(); rb != nil { /* return to pool */ }`.

encoding/m3tsz/encoder.go
robskillington: Opened PR: https://github.com/m3db/m3x/pull/31

PR #201, opened 2016-12-20
add AdminSession endpoint for streaming gets

prateek:
- New endpoint in client for retrieving blocks from specified peers. Intent is to use with the shardRepairer
- Refactor `bootstrap.(Options|ShardResult|...)` -> new package `client/result`


robskillington:
Beginning review, this is a long-un!

client/result/result.go
```diff
@@ -31,37 +31,37 @@ import (
...
+type bootstrapResult struct {
 	results     ShardResults
 	unfulfilled ShardTimeRanges
 }
 
-// NewResult creates a new result.
-func NewResult() Result {
-	return &result{
+// NewBootstrapResult creates a new result.
+func NewBootstrapResult() BootstrapResult {
```
robskillington: Hm, I believe this is used extensively in the `storage/bootstrap/...` modules also.  Seems making them hierarchically underneath `client/...` may be strange?

client/session.go
```diff
@@ -1856,43 +1931,50 @@ func (s *session) streamBlocksReattemptFromPeers(
...
+	if err := iter.Err(); err != nil {
+		encoder.Close()
+		return nil, err
+	}
+
+	return encoder, nil
 }
 
-func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
+func (b *baseBlocksResult) getDatabaseBlock(block *rpc.Block) (block.DatabaseBlock, error) {
```
robskillington: Perhaps `newDatabaseBlock(...)` instead of `get...`?

client/session.go
```diff
@@ -1912,17 +1994,135 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
...
+}
+
+func newBulkBlocksResult(
+	opts Options,
+	resultOpts result.Options,
+	multiReaderIteratorPool encoding.MultiReaderIteratorPool,
+) *bulkBlocksResult {
+	blockOpts := resultOpts.DatabaseBlockOptions()
+	return &bulkBlocksResult{
+		baseBlocksResult: baseBlocksResult{
```
robskillington: We can extract this constructing of `baseBlocksResult` into a `newBaseBlocksResult(...)` you can call from both `streamBlocksResult` and `bulkBlocksResult` constructors perhaps?

client/session.go
```diff
@@ -1177,6 +1177,81 @@ func (s *session) FetchBootstrapBlocksFromPeers(
...
+		m.fetchBlocksFromPeers.Update(0)
+	}()
+
+	// Transform provided repairBlocks into channel
+	metadataCh := make(chan blocksMetadata, 4096)
+	go func() {
+		for _, rb := range repairBlocks {
+			peer, ok := peersByHost[rb.Peer.ID()]
+			if !ok {
+				err := errors.New("Unable to find peer: " + rb.Peer.ID())
```
prateek: Yep, will do.

client/session.go
prateek: willdo.

client/session.go
prateek: +1

prateek:
@robskillington addressed feedback, could you take a look at this when you get a sec.

prateek:
Rebased onto master, should have no merge conflicts now.

prateek:
Rebased onto master, should have no merge conflicts now.

client/result/result.go
```diff
@@ -0,0 +1,352 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
prateek: The `Result` object is not in the `client` package, it's in `client/result`.

My rationale for this package structure: originally, the `client` package depended upon `bootstrap` because of `bootstrap.Result` object. That felt like a more brittle dependency. So instead, I introduced a new package for the `client` specific parts of the `Result` object. This package didn't feel heavy enough to merit a new top-level package, so I stuck it in `client/result`. This way, `client` and `bootstrap` both depend on `client/Result`.

Am I overthinking this?

client/result/result.go
```diff
@@ -0,0 +1,352 @@
...
+
+		buf.WriteString(fmt.Sprintf("%d: %s", shard, ranges.String()))
+	}
+	buf.WriteString("}")
+
+	return buf.String()
+}
+
+// SummaryString returns a summary description of the time ranges
+func (r ShardTimeRanges) SummaryString() string {
```
prateek: Sure, will address in the next patch

client/result/types.go
```diff
@@ -0,0 +1,102 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
prateek: responded to above.

client/types.go
```diff
@@ -191,8 +204,17 @@ type AdminSession interface {
...
 		shard uint32,
 		start, end time.Time,
-		opts bootstrap.Options,
-	) (bootstrap.ShardResult, error)
+		opts result.Options,
+	) (result.ShardResult, error)
+
+	// FetchRepairBlocksFromPeers will fetch the requested blocks from the
+	// peers specified
+	FetchRepairBlocksFromPeers(
```
prateek: Good point. Will address in next patch

storage/bootstrap/bootstrapper/base.go
```diff
@@ -35,7 +36,7 @@ const (
 
 // baseBootstrapper provides a skeleton for the interface methods.
 type baseBootstrapper struct {
-	opts bootstrap.Options
+	opts result.Options
```
prateek: as above, base bootstrapper depends on `client/result`, not `client`

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -25,6 +25,7 @@ import (
 	"io"
 	"time"
 
+	"github.com/m3db/m3db/client/result"
```
prateek: same as above

client/session.go
```diff
@@ -1177,6 +1177,81 @@ func (s *session) FetchBootstrapBlocksFromPeers(
 	return result.result, nil
 }
 
+func (s *session) FetchRepairBlocksFromPeers(
+	namespace ts.ID,
+	shard uint32,
+	repairBlocks []block.ReplicaMetadata,
```
prateek: Will do. Don't have a strong opinion on this.

client/session.go
```diff
@@ -1912,17 +2010,124 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
 				Tail: segments.Unmerged[i].Tail,
 			})
 		}
-		encoder, err := r.mergeReaders(start, readers)
+		encoder, err := b.mergeReaders(start, readers)
 		if err != nil {
-			return err
+			encoder.Close() // return encoder to pool
```
prateek: Nice catch.

client/session.go
```diff
@@ -1177,6 +1177,81 @@ func (s *session) FetchBootstrapBlocksFromPeers(
...
+		outputCh = make(chan peerBlocksDatapoint, 4096)
+		result   = newStreamBlocksResult(s.opts, opts, s.multiReaderIteratorPool, outputCh)
+		onDone   = func(err error) {
+			atomic.StoreInt64(&complete, 1)
+			select {
+			case doneCh <- err:
+			default:
+			}
+		}
+		m = s.streamFromPeersMetricsForShard(shard)
```
prateek: nice catch! will address

client/session.go
prateek: nice catch!

storage/block/types.go
```diff
@@ -43,6 +44,16 @@ type BlocksMetadata struct {
 	Blocks []Metadata
 }
 
+// ReplicaMetadata captures block metadata along with corresponding peer identifier
+// for a single replica of a block
+type ReplicaMetadata struct {
```
prateek: Makes sense.

prateek:
@xichen2020 re: unit tests, there's some included in this pr [quick link](https://github.com/m3db/m3db/pull/201/files/93d354c6ea564ddd8f4d28280139fa0d02e72a07#diff-376e1661307a93f204e75960224020d9)

re: integration tests, will do

client/result/result.go
xichen2020: Right yeah that's what I meant by "in the `client` package`. 

AFAIR, `result` is always used for bootstrapping, hence it made sense for it to live inside the bootstrap package. Is it also used for repair now? If so, I can see a need to move it out of bootstrap, although probably not under `client/...`

client/session.go
```diff
@@ -1177,6 +1177,81 @@ func (s *session) FetchBootstrapBlocksFromPeers(
...
+	}()
+
+	// Transform provided repairBlocks into channel
+	metadataCh := make(chan blocksMetadata, 4096)
+	go func() {
+		for _, rb := range repairBlocks {
+			peer, ok := peersByHost[rb.Peer.ID()]
+			if !ok {
+				err := fmt.Errorf("unable to find peer: %v", rb.Peer.ID())
+				onDone(err) // bail early
```
prateek: I had it this way to sanitise input, didn't consider the topology change case at all. That said, the way I've got the repair code structured, we merge units of peer block replicas individually anyway. So yeah, logging a warning and continuing makes sense.

client/result/result.go
prateek: The only bits of `client/result` shared by `repair` and `bootstrap` is the `Options` object. So least invasive change overall would be to only extract that into `client`. 

My preference is to leave it as is currently in the PR because having `client` depend on `bootstrap` mandates no file in the top-level `bootstrap` package import client due to it introducing a circular dependency. Feels icky. But obv the bootstrapping code doesn't do that now, so not opposed to doing the least invasive change of moving `Options` out.

xichen2020: Agreed that having `client` depend on `bootstrap` could lead to circular dependency issues if not careful, so I'm not opposed to moving `result` to a separate package that both `client` and `bootstrap` depend on. My main argument is to not have that live under `client/result`, since in most cases a nested package either is closed tied to or implements interfaces in its parent package, whereas in this case `result` is neither to the `client` package, so have the options and types related to bootstrapping results live in `client/result` feels wrong to me. Instead, `result` is logically an integral part of the bootstrap process, so IMO a better option would be moving it to `bootstrap/result`.

prateek: Sold! Bits shared between `repair` and `bootstrap` -> `client`. Rest of `client/result` -> `bootstrap/result`.

xichen2020: Awesome (thumbsup)!

Is `Options` the only thing shared? If so, might make more sense to leave that in `bootstrap/result` too so bootstrap results and bootstrap result options don't end up living in separate packages

prateek: That makes sense. Essentially the same as current state with `client/result` ->`bootstrap/result`

client/session.go
prateek: lol, so turns out `blocksMetadata` is defined as a struct in this package. Changing var name to `requiredBlocksMetadata`. Lemme know if you have a better suggestion.

xichen2020: lol, `replicasMetadata` perhaps?

client/result/result.go
xichen2020: Right yeah

robskillington: This conclusion seems good 👍

client/session.go
robskillington: How about just blocks?

prateek: `metadatas` has my vote

prateek:
@xichen2020 updated per feedback. 

- in result refactor, also moved `bootstrap.Result` -> `bootstrap/result.BootstrapResult`
- for `streamFromPeersMetricsForShard`, it looks like we already had some collisions in the metrics scopes (repair metadata retrieval and bootstrap blocks retrieval could report metrics under the same name). The fix I put in adds an additional tag in the scope for the three usages (bootstrap|metadata|raw blocks). 
- added integration tests
- added moar unit tests

One specific thing I want feedback on - take a look at the diff between the two latest commits (5f42ff5 and 2fc0231). The latest commit changes `streamCollectedBlocksMetadata` in `session.go` to simplify `FetchBlocksFromPeers` (metadata no longer requires grouping), but I'm not sure if it breaks any existing code. The tests all run, but could you take a look and make sure it's not breaking any invariants you had in mind when writing it.

storage/bootstrap/result/result.go
```diff
@@ -344,46 +346,33 @@ func (r ShardTimeRanges) String() string {
...
-			curr := it.Value()
-			duration += curr.End.Sub(curr.Start)
-		}
-		buf.WriteString(fmt.Sprintf("%d: %s", shard, duration.String()))
-	}
-	buf.WriteString("}")
-
-	return buf.String()
+// SummaryString returns a summary description of the time ranges
+func (r ShardTimeRanges) SummaryString() string {
```
xichen2020: 👍

client/session.go
```diff
@@ -57,6 +57,10 @@ const (
 	blocksMetadataInitialCapacity        = 64
 	blocksMetadataChannelInitialCapacity = 4096
 	gaugeReportInterval                  = 500 * time.Millisecond
+
+	resultTypeMetadata  = "metadata"
```
xichen2020: Make this an enum perhaps?

client/session.go
```diff
@@ -1429,17 +1526,12 @@ func (s *session) streamCollectedBlocksMetadata(
...
 
 		received, ok := metadata[m.id.Hash()]
-		if !ok {
+		if !ok || received.submitted {
 			received = &receivedBlocks{
 				results: make([]*blocksMetadata, 0, peersLen),
 			}
 			metadata[m.id.Hash()] = received
 		}
-		if received.submitted {
```
xichen2020: I don't think we should move `received.submitted` up here. Honestly I don't know when this would happen, it seems like an error to me, but in any case if it's already submitted, we probably don't want to re-submit them

client/session.go
prateek: will do

client/session.go
prateek: You're right, I don't think this event can happen when going through the bootstrap codepath. It can occur when going through the `FetchBlocksFromPeers` method because the interface doesn't mandate that all the blocks per unique ID be grouped into a single `blocksMetadata` equivalent. One of the new unit tests and integration test both capture this. 

for this fix, I think there's 3 choices:
(1) perform the grouping per id in the session.go code
(2) change the `FetchBlocksFromPeers` method interface
(3) move `received.submitted` up

I tried (1) - 2fc0231, didn't like it mainly because it makes testing a lot more cumbersome. Not against doing (2) but I can't think of a good way to capture the grouping dependency in the interface - I don't like mandating it implicitly via docs or by something explicit like taking in map. Both push more onus on the user of the API. So settled on (3) for now. What do you think?

xichen2020: hm..maybe i'm missing the point, what's the issue with leaving it where it is now?

prateek: Spoke to @xichen2020 offline. tl;dr - need the change because how we submit specify BlocksMetadata changes in Bootstrap v Repair. Next patch refactors this by making separate functions for the enqueue ops

PR #202, opened 2016-12-20
Fixes for IDs

kobolog:
This PR fixes a bug where ID hashes are never cached. It also reverts IDs to be immutable again.


ts/identifier.go
```diff
@@ -78,6 +78,7 @@ func (v *id) Hash() Hash {
 		// access to the hash field, compute and cache it.
 		if atomic.CompareAndSwapInt32(&v.flag, int32(invalid), int32(pending)) {
 			v.hash = hash(v.data)
+			atomic.StoreInt32(&v.flag, int32(computed))
```
xichen2020: This looks like a bug fix?

PR #203, opened 2016-12-20
Fix some leaks here and there

kobolog:
This PR fixes some leaks around the codebase.

### DIFF
--- a/network/server/tchannelthrift/node/service.go
+++ b/network/server/tchannelthrift/node/service.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/m3db/m3db/clock"
+	"github.com/m3db/m3db/context"
 	"github.com/m3db/m3db/encoding"
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	"github.com/m3db/m3db/network/server/tchannelthrift"
@@ -338,6 +339,7 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
 		s.metrics.fetchBlocksMetadata.ReportError(s.nowFn().Sub(callStart))
 		return nil, convert.ToRPCError(err)
 	}
+	ctx.RegisterFinalizer(context.FinalizerFn(fetched.Close))
 
 	fetchedResults := fetched.Results()
 	result := rpc.NewFetchBlocksMetadataRawResult_()
@@ -350,8 +352,7 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
 
 	for _, fetchedMetadata := range fetchedResults {
 		blocksMetadata := s.blocksMetadataPool.Get()
-		// LEAK(@kobolog): `fetched` is closed before return, can ID expire?
-		blocksMetadata.ID = append([]byte{}, fetchedMetadata.ID.Data()...)
+		blocksMetadata.ID = fetchedMetadata.ID.Data()
 		fetchedMetadataBlocks := fetchedMetadata.Blocks.Results()
 		blocksMetadata.Blocks = s.blockMetadataSlicePool.Get()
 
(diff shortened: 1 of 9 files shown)


network/server/tchannelthrift/node/service.go
```diff
@@ -350,7 +350,8 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
 
 	for _, fetchedMetadata := range fetchedResults {
 		blocksMetadata := s.blocksMetadataPool.Get()
-		blocksMetadata.ID = fetchedMetadata.ID.Data()
+		// LEAK(@kobolog): `fetched` is closed before return, can ID expire?
+		blocksMetadata.ID = append([]byte{}, fetchedMetadata.ID.Data()...)
```
robskillington: Perhaps use `append([]byte(nil), bytes...)` to avoid alloc of the `[]byte{}`

robskillington: So what might be best in terms of efficiency, you can `ctx.RegisterCloser(fetched)` then that way the ID won't expire.

storage/series/buffer.go
```diff
@@ -482,6 +482,7 @@ func (b *dbBufferBucket) merge() {
 	for i := range b.encoders {
 		if stream := b.encoders[i].encoder.Stream(); stream != nil {
 			readers = append(readers, stream)
+			defer stream.Segment().Finalize()
```
robskillington: How about just `defer stream.Finalize()`?

network/server/tchannelthrift/node/service.go
robskillington: Also: we need to Clone the ID that is put in the metadata object here in `storage/block/NewFetchBlocksMetadataResult(id ts.ID, ...)`.

ts/segment.go
```diff
@@ -40,12 +40,12 @@ type Segment struct {
 }
 
 // Len returns the length of the head and tail.
-func (s *Segment) Len() int {
+func (s Segment) Len() int {
 	return len(s.Head) + len(s.Tail)
 }
 
 // Finalize will release resources kept by the segment if pooled.
-func (s *Segment) Finalize() {
```
robskillington: Only thing that's not great about making this a non-pointer type is that now it won't actually `nil` out the `Head` and `Tail` of whoever is calling these methods no?

kobolog: Yeah. But we always use it as a value type everywhere, so to actually call `Finalize()` one would have to write something like `s := stream.Segment(); (&s).Finalize()` otherwise. Thoughts?

robskillington: Well golang has a magic shorthand where you just call `s.Finalize()` and it will essentially compile it as shorthand for `(&s).Finalize()`.  Hehe.

kobolog: Well it actually will not =)

storage/series/series.go
```diff
@@ -82,8 +82,11 @@ func NewPooledDatabaseSeries(id ts.ID, pool DatabaseSeriesPool, opts Options) Da
 }
 
 func newDatabaseSeries(id ts.ID, opts Options) *dbSeries {
+	if id != nil {
+		id = opts.IdentifierPool().Clone(id)
+	}
```
kobolog: This actually is never called with anything except nil.

storage/block/result.go
```diff
@@ -161,6 +161,7 @@ func (s *fetchBlocksMetadataResults) Reset() {
 func (s *fetchBlocksMetadataResults) Close() {
 	for i := range s.results {
 		s.results[i].Blocks.Close()
+		s.results[i].ID.Finalize()
```
xichen2020: I don't think the result `ID`s are cloned, we probably should, though, otherwise we might return the series ID to the pool

kobolog: They are now!

PR #207, opened 2016-12-22
Bl/writeop fix

ben-lerner:
Actually store shardID in writeop.

### DIFF
--- a/client/write.go
+++ b/client/write.go
@@ -26,6 +26,7 @@ import (
 	"sync"
 	"sync/atomic"
 
+	"github.com/m3db/m3cluster/shard"
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	"github.com/m3db/m3db/topology"
 	"github.com/m3db/m3db/ts"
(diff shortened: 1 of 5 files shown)


client/write.go
```diff
@@ -136,11 +139,11 @@ func (w *writeState) close() {
 }
 
 func (w *writeState) completionFn(result interface{}, err error) {
-	hostID := result.(topology.Host).ID()
+	hostID := result.(string)
```
xichen2020: If `result` is assertively a string type, why not pass `result` in as a string?

client/write.go
```diff
@@ -136,11 +139,11 @@ func (w *writeState) close() {
...
 	// NB(bl) panic on invalid result, it indicates a bug in the code
 
-	w.Lock()
-	w.pending--
+	pending := atomic.AddInt32(&w.pending, -1)
+	var addToSuccess int32
 
 	if err != nil {
 		w.errors = append(w.errors, fmt.Errorf("error writing to host %s: %v", hostID, err))
```
xichen2020: Do we need to protect `w.errors` here in case multiple go routines access it at the same time?

client/write.go
```diff
@@ -151,24 +154,26 @@ func (w *writeState) completionFn(result interface{}, err error) {
 		w.errors = append(w.errors, fmt.Errorf("missing shard %d in host %s", w.op.shardID, hostID))
 	} else if shardState == shard.Available {
 		// NB(bl): only count writes to available shards towards success
-		w.success++
+		addToSuccess = 1
 	}
+	success := atomic.AddInt32(&w.success, addToSuccess)
```
xichen2020: nit: can move this to line 157, no need to do an add in error cases

client/write.go
```diff
@@ -151,24 +154,26 @@ func (w *writeState) completionFn(result interface{}, err error) {
...
+	w.Lock()
 	switch w.session.writeLevel {
 	case topology.ConsistencyLevelOne:
-		if w.success > 0 || w.pending == 0 {
+		if success > 0 || pending == 0 {
 			w.Signal()
 		}
 	case topology.ConsistencyLevelMajority:
-		if w.success > w.majority || w.pending == 0 {
+		if success >= w.majority || pending == 0 {
```
xichen2020: Can we add a test to make sure we catch this in the future? (presumably it fails with the master but passes with this change)

PR #208, opened 2016-12-23
Write quorum takes shard.State into account

ben-lerner:
Fix two bugs in the write quorum logic, add regression tests.


client/session_write_test.go
```diff
@@ -218,15 +221,32 @@ func testWriteConsistencyLevel(
...
-	writeWg.Wait()
+	// Wait for write to complete, or timeout
+	doneCh := make(chan struct{})
+	timeoutCh := make(chan struct{})
+	go func() {
+		writeWg.Wait()
+		close(doneCh)
+	}()
+	go func() {
+		time.Sleep(10 * time.Millisecond)
```
xichen2020: When will they time out?

client/session_write_test.go
```diff
@@ -242,16 +219,28 @@ func testWriteConsistencyLevel(
...
-	// Wait for write to complete
-	writeWg.Wait()
+	// Wait for write to complete or timeout
+	doneCh := make(chan struct{})
+	go func() {
+		writeWg.Wait()
+		close(doneCh)
+	}()
+
+	select { // NB(bl): Check that we're correctly signaling in write_state.completionFn.
```
robskillington: Out of interest what is this comment explaining?

client/session_write_test.go
```diff
@@ -290,3 +279,33 @@ func testWriteConsistencyLevel(
 		}
 	}
 }
+
+type writeStub struct {
```
robskillington: 👍

client/session.go
```diff
@@ -683,28 +685,19 @@ func (s *session) Write(namespace, id string, t time.Time, value float64, unit x
...
-
-			// NB(r): if this ever happens we have a code bug, once we
-			// are in the read lock the current queues we are using should
-			// never be closed
-			s.RUnlock()
-			s.opts.InstrumentOptions().Logger().Errorf("failed to enqueue write: %v", err)
-			return err
+			// NB(r): if this happens we have a bug, once we are in the read
+			// lock the current queues should never be closed
+			panic(fmt.Sprintf("failed to enqueue write: %v", err))
```
robskillington: A better error description is warranted if we want to make this a panic.  Also as an engineer that would prefer to avoid a whole set of ingesters panicing if this case is introduced in a new release, I'd prefer to see the write retried rather than the entire process panicing and taking a while to come back up while the same issue sporadically affects the rest of the fleet.

I think we should leave the non-panic behavior intact.

client/session.go
```diff
@@ -413,9 +413,10 @@ func (s *session) BorrowConnection(hostID string, fn withConnectionFn) error {
 
 func (s *session) newHostQueues(topoMap topology.Map) ([]hostQueue, int, int, error) {
 	// NB(r): we leave existing writes in the host queues to finish
-	// as they are already enroute to their destination, this is ok
-	// as part of adding a host is to add another replica for the
-	// shard set and only once bootstrapped decomission the old node
+	// as they are already enroute to their destination. This is an edge case
+	// that might result in leaving nodes counting towards quorum, but fixing it
+	// would result in additional chatter.
```
robskillington: 👍

client/session.go
```diff
@@ -683,28 +685,19 @@ func (s *session) Write(namespace, id string, t time.Time, value float64, unit x
...
+			// lock the current queues should never be closed
+			panic(fmt.Sprintf("failed to enqueue write: %v", err))
 		}
 
 		state.incRef()
 		enqueued++
 	}
 
-	// Wait for writes to complete. We don't need to loop over Wait() since there
-	// are no spurious wakeups in Golang and the condition is one-way and binary.
```
robskillington: Any reason to remove this comment?

glide.lock
```diff
@@ -74,7 +74,7 @@ imports:
 - name: github.com/willf/bitset
   version: 2e6e8094ef4745224150c88c16191c7dceaad16f
 - name: golang.org/x/net
-  version: 60c41d1de8da134c05b7b40154a9a82bf5b7edb9
```
robskillington: Can you revert changes to glide.lock file?  Or just checkout out to the copy of glide.lock in master?  This is an annoying side affect of us not pinning `golang.org/x/net` that it is non-deterministic about the version it gets.  

We'll address this in a separate PR.

integration/cluster_add_one_node_test.go
```diff
@@ -59,45 +56,26 @@ func TestClusterAddOneNode(t *testing.T) {
...
-			services.NewServiceInstance().
-				SetInstanceID("testhost0").
-				SetEndpoint("127.0.0.1:9000").
-				SetShards(newClusterShardsRange(0, 1023, shard.Available)),
-			services.NewServiceInstance().
-				SetInstanceID("testhost1").
-				SetEndpoint("127.0.0.1:9004").
-				SetShards(newClusterEmptyShardsRange()),
+			node(t, 0, newClusterShardsRange(0, 1023, shard.Available)),
+			node(t, 1, newClusterEmptyShardsRange()),
```
robskillington: Good stuff, much more terse.

integration/cluster_add_one_node_test.go
```diff
@@ -250,3 +228,13 @@ func TestClusterAddOneNode(t *testing.T) {
 		verifySeriesMaps(t, setups[i], namesp.ID(), expectedSeriesMaps[i])
 	}
 }
+
+// utils
+
+func node(t *testing.T, n int, shards shard.Shards) services.ServiceInstance {
```
robskillington: Perhaps put this into `integration.go` for more central place?

integration/write_quorum_test.go
```diff
@@ -0,0 +1,177 @@
...
+	// Writes succeed to two available nodes
+	require.NoError(t, nodes[2].startServer())
+	assert.NoError(t, testWrite(topology.ConsistencyLevelOne))
+	assert.NoError(t, testWrite(topology.ConsistencyLevelMajority))
+	assert.Error(t, testWrite(topology.ConsistencyLevelAll))
+}
+
+type testWriteFn func(topology.ConsistencyLevel) error
+
+func makeNodes(t *testing.T, instances []services.ServiceInstance) (testSetups,
```
robskillington: 👍 👍 👍  parameterized cluster setup will definitely be useful down the line.

Want to refactor it so it doesn't need to return a `testWrite`, i.e. make a `func makeTestWrite(topoInit) testWriteFn { ... }` that you can call separately.

That way `makeNodes` can just focus on creating the nodes themselves?  Then we can reuse it in other tests down the line.  Once this refactor is done you can move `makeNodes(...)` to `integration.go` for reuse by others.

Also very minor nit: we could call it `newNodes(...)` to be slightly more consistent with our other constructor naming.

robskillington:
Finished review round, probably only an iteration away from getting this merged 👍 .

client/session_write_test.go
ben-lerner: If the completionFn is incorrect, the write above will hang. Re-worded for clarity.

client/session.go
ben-lerner: Changed back.

client/session.go
ben-lerner: It seems like it's just explaining how `.Wait()` works, am I missing something subtle? I'd expect a comment if we were looping over some variable, not if we do the default Go thing.

PR #209, opened 2016-12-24
Fix use-after-free in repair logic.

kobolog:
This PR fixes an issue where pooled resources in metadata results are being used after the enclosing object was destroyed.


kobolog:
Note: this PR is an imperfect solution. The real defect lies in the fact that `Shard::FetchBlocksMetadata()` doesn't use the passed `Context` to bind pooled resource lifetimes obtained in `Series::FetchBlocksMetadata()`.

PR #212, opened 2016-12-28
Use checked bytes pool to add raw buffer read/write guarantees

robskillington:
This change pulls in the `checked.Bytes` type from m3x and begins using `pool.CheckedBytesPool` from m3x in all places where we previously used a `pool.BytesPool` to add safety to our use of pooled bytes.

This will help us catch any subtle pooling bugs as they are introduced.

cc @xichen2020 @kobolog @martin-mao @cw9


robskillington:
@xichen2020 this is ready for review when you get a chance.  Muchos gracias.

client/session_bootstrap_blocks_test.go
```diff
@@ -1159,7 +1161,7 @@ func (m *fetchBlocksReqMatcher) Matches(x interface{}) bool {
 	}
 
 	for i := range params {
-		if !params[i].id.Equal(ts.BinaryID(req.Elements[i].ID)) {
+		if params[i].id.String() != string(req.Elements[i].ID) {
```
kobolog: This allocates two strings vs. zero.

network/server/tchannelthrift/node/service.go
```diff
@@ -332,9 +341,16 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
 		includeChecksums = *req.IncludeChecksums
 	}
 
-	fetched, nextPageToken, err := s.db.FetchBlocksMetadata(
-		ctx, s.idPool.GetBinaryID(ctx, req.NameSpace), uint32(req.Shard),
-		start, end, req.Limit, pageToken, includeSizes, includeChecksums)
+	buffer := s.bytesPool.Get(len(req.NameSpace))
```
kobolog: What's the point in pooling these bytes? `req.NameSpace` is allocated by thrift outside of pools, why copy it into a pooled memory?

persist/fs/read.go
```diff
@@ -260,7 +271,7 @@ func (r *reader) Read() (ts.ID, []byte, error) {
 
 	r.entriesRead++
 
-	return ts.BinaryID(entry.Id), data, nil
+	return ts.StringID(string(entry.Id)), data, nil
```
kobolog: These string casts here and elsewhere are exactly why binary IDs were introduced in the first place – reverting them back would skyrocket p99s back to multiple seconds.

robskillington: Well this is only during bootstrap, once per ID.  Instead of on the write and read path which is once per datapoint.  

This was definitely a hack as I was unsure about the ownership of bytes created by the protobuf library so wanted to perform a copy here.  Instead of `StringID(...)` though I can introduce a `BinaryID(append(r.bytesPool.Get(len(entry.Id)), entry.Id...))`.

client/session_bootstrap_blocks_test.go
robskillington: Yeah, mainly just because this in tests and I didn't want to have to copy the bytes behind `req.Elements[i].ID` to be safe so thought it would be simpler to just stringify them.  I can copy the bytes however.

network/server/tchannelthrift/node/service.go
robskillington: The main thing is that `ts.BinaryID(...)` now requires a `checked.Bytes` ref to be passed.  The issue here is that there's no easy way to reuse the `checked.Bytes` from the bytes pool without using the actual backing bytes.

Perhaps I'll add a simple `sync.Pool` for the use of `checked.Bytes` in this file.  Let me update with that.

client/reader_slice_of_slices_iterator.go
```diff
@@ -98,8 +118,20 @@ func (it *readerSliceOfSlicesIterator) Close() {
...
 	it.closed = true
-	if it.pool != nil {
-		it.pool.Put(it)
+	// Release any refs to segments
+	it.segments = nil
+	// Release any refs to segment byte slices
+	for i := range it.segmentReaders {
+		seg := it.segmentReaders[i].Segment()
+		if seg.Head != nil {
+			seg.Head.Reset(nil)
```
xichen2020: Do we need to do a `DecRef` here since we no longer own the bytes?

client/reader_slice_of_slices_iterator.go
```diff
@@ -98,8 +118,20 @@ func (it *readerSliceOfSlicesIterator) Close() {
...
+		}
+		if seg.Tail != nil {
+			seg.Tail.Reset(nil)
```
xichen2020: Same here

client/session.go
```diff
@@ -1747,7 +1749,11 @@ func (s *session) streamBlocksBatchFromPeer(
 			continue
 		}
 
-		id := ts.BinaryID(result.Elements[i].ID)
+		buffer := bytesPool.Get(len(result.Elements[i].ID))
+		buffer.IncRef()
+		buffer.AppendAll(result.Elements[i].ID)
+		buffer.DecRef()
+		id := ts.BinaryID(buffer)
```
xichen2020: Semantically it might makes more sense to do a `DecRef` after assigning `buffer` to `id`

client/session.go
```diff
@@ -1898,21 +1924,24 @@ func (r *blocksResult) addBlockFromPeer(id ts.ID, block *rpc.Block) error {
...
-			})
+			reader := readerPool.Get()
+			reader.Reset(r.segmentForBlock(segments.Unmerged[i]))
+			readers[i] = reader
 		}
+
 		encoder, err := r.mergeReaders(start, readers)
+		for _, reader := range readers {
+			// Close each reader
+			reader.(xio.SegmentReader).Close()
```
xichen2020: hm, why the cast here?

network/server/tchannelthrift/cluster/service.go
```diff
@@ -180,7 +180,7 @@ func (s *service) Truncate(tctx thrift.Context, req *rpc.TruncateRequest) (*rpc.
 	if !ok {
 		return nil, tterrors.NewInternalError(errors.New("unable to get an admin session"))
 	}
-	truncated, err := adminSession.Truncate(ts.BinaryID(req.NameSpace))
+	truncated, err := adminSession.Truncate(ts.StringID(string(req.NameSpace)))
```
xichen2020: Why do we convert it to string id now?

network/server/tchannelthrift/node/service.go
```diff
@@ -332,9 +341,16 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
...
+	buffer.IncRef()
+	buffer.AppendAll(req.NameSpace)
+	buffer.DecRef()
+	nsID := s.idPool.GetBinaryID(ctx, buffer)
```
xichen2020: Seems you can use `newID(ctx, req.NameSpace)` here?

persist/fs/commitlog/reader.go
```diff
@@ -115,8 +115,8 @@ func (r *reader) Read() (
 		}
 		r.metadataLookup[r.log.Index] = Series{
 			UniqueIndex: r.log.Index,
-			ID:          ts.BinaryID(r.metadata.Id),
-			Namespace:   ts.BinaryID(r.metadata.Namespace),
+			ID:          ts.StringID(string(r.metadata.Id)),
```
xichen2020: Why are we converting the IDs to strings?

storage/namespace/metadata.go
```diff
@@ -29,7 +29,7 @@ type metadata struct {
 
 // NewMetadata creates a new namespace metadata
 func NewMetadata(id ts.ID, opts Options) Metadata {
-	return metadata{id: ts.BinaryID(append([]byte{}, id.Data()...)), opts: opts}
+	return metadata{id: ts.StringID(string(id.Data().Get())), opts: opts}
```
xichen2020: Same here, why string id instead of binary id?

ts/segment.go
```diff
@@ -20,38 +20,83 @@
...
+// bytes in a Segment.
+type SegmentFlags uint8
+
+const (
+	// FinalizeNone specifies to finalize neither of the bytes
+	FinalizeNone SegmentFlags = 0 << 0
+	// FinalizeHead specifies to finalize the head bytes
+	FinalizeHead SegmentFlags = 1 << 0
+	// FinalizeTail specifies to finalize the tail bytes
+	FinalizeTail SegmentFlags = 2 << 0
```
xichen2020: Why "<<0"? Did you mean `1 << 1` instead?

client/reader_slice_of_slices_iterator.go
robskillington: So when `Reset` is called we don't actually give up ref to the checked bytes struct itself and we reuse it.  Hence why I never DecRef for the `checked.Bytes` here.

client/reader_slice_of_slices_iterator.go
robskillington: Same reply as above.

client/session.go

client/session.go
robskillington: Because `readers` is an slice of `[]io.Reader` which `mergeReaders` takes, not a slice of `[]xio.SegmentReader` and the `io.Reader` interface has no `Close()` method on it.

network/server/tchannelthrift/cluster/service.go
robskillington: Don't have to, but it's either this or creating a new `checked.NewBytes(...)`.  I'll change to `checked.NewBytes(...)` and use `ts.BinaryID(...)`, but it really is quite similar - especially for truncate which barely if ever gets called.

network/server/tchannelthrift/node/service.go
robskillington: Yup, will do.

persist/fs/commitlog/reader.go
robskillington: Can use `checked.NewBytes(...)` here.  Will do that.

storage/namespace/metadata.go
robskillington: Can use checked.NewBytes(...) here. Will do that.

ts/segment.go
robskillington: Ah yeah, good catch.

client/session.go
xichen2020: I see, makes sense, might wanna do that in the for loop below too

ts/segment.go
```diff
@@ -43,11 +43,11 @@ type SegmentFlags uint8
...
 const (
 	// FinalizeNone specifies to finalize neither of the bytes
-	FinalizeNone SegmentFlags = 0 << 0
+	FinalizeNone SegmentFlags = 0 << 1
 	// FinalizeHead specifies to finalize the head bytes
-	FinalizeHead SegmentFlags = 1 << 0
+	FinalizeHead SegmentFlags = 1 << 1
 	// FinalizeTail specifies to finalize the tail bytes
-	FinalizeTail SegmentFlags = 2 << 0
+	FinalizeTail SegmentFlags = 2 << 1
```
xichen2020: nit: more conventional to do `1 << 2`

persist/fs/read.go
kobolog: I think that would be more appropriate here yeah, since right now it looks like an unnecessary string conversion and somebody might accidentally remove it in the future.

integration/client.go
```diff
@@ -181,5 +181,5 @@ func m3dbClientTruncate(c client.Client, req *rpc.TruncateRequest) (int64, error
 		return 0, errors.New("unable to get an admin session")
 	}
 
-	return adminSession.Truncate(ts.BinaryID(req.NameSpace))
+	return adminSession.Truncate(ts.StringID(string(req.NameSpace)))
```
kobolog: Same suspicious string cast here as well.

client/reader_slice_of_slices_iterator.go
robskillington: So calling `Finalize(...)` on the `checked.Bytes` means it will return itself to a pool if its pooled.  Here we aren't using pooled `checked.Bytes` hence the use of reset and wanting to avoid any `Finalize(...)` calls.

ts/segment.go
```diff
@@ -20,38 +20,106 @@
...
 
-// Finalize will release resources kept by the segment if pooled.
-func (s *Segment) Finalize() {
-	if s.HeadPool != nil && s.Head != nil {
-		s.HeadPool.Put(s.Head)
+// Equal returns if this segment is equal to another.
+// WARNIG: this   should only be used in code paths not
+// executed often as it allocates bytes to concat each
+// segment head and tail together before comparing the contents.
+func (s *Segment) Equal(other *Segment) bool {
```
xichen2020: Maybe comparing checksums on the bytes is quicker?

ts/identifier_pool.go
```diff
@@ -39,59 +38,83 @@ type simpleIdentifierPool struct {
...
+	p.pool.Init(func() interface{} {
+		return &id{pool: p}
+	})
+	return p
 }
 
 type nativeIdentifierPool struct {
-	pool pool.NativePool
-	heap pool.BytesPool
+	pool pool.ObjectPool
```
xichen2020: A comment or two here explaining why we switched to normal pooling would make it easier to comprehend the rationale behind it.

ts/segment.go
robskillington: Could be, however the checksums aren't present on the `Segment`.  This method is only used in tests right now and I don't expect to use it in any other place.  If we do need a faster version we can make it faster at that time IMO.

ts/identifier_pool.go
robskillington: Will do.

ts/segment.go
xichen2020: Fair enough

persist/fs/commitlog/types.go
```diff
@@ -105,8 +105,24 @@ type Series struct {
...
+
+	// SegmentFlags declares whether to finalize when finalizing the Series.
+	Flags SeriesFlags
 }
 
+// SeriesFlags describes the option to finalize or not finalize
+// bytes in a Series.
+type SeriesFlags uint8
+
+const (
```
kobolog: This way of assigning flag values is wrong, it should be 1 << 0, 1 << 1, 1 << 2 and so on. Otherwise you'll end up with overlapping values, e.g. 3 << 1 == 6, which is 0b00000110 – note two 1s.

ts/segment.go
kobolog: There's a typo and some extra space in the comment: WARNIG vs. WARNING.

storage/series/series.go
```diff
@@ -57,9 +57,14 @@ var (
 )
 
 var (
-	nilID = ts.BinaryID(checked.NewBytes(nil, nil))
+	nilIDBytes = checked.NewBytes(nil, nil)
+	nilID      = ts.BinaryID(nilIDBytes)
 )
 
+func init() {
+	nilIDBytes.IncRef()
```
kobolog: I'm not sure about the package initialization ordering, but potentially `IncRef()` will be called after `ts.BinaryID()`, could you verify it's not what's happening?

ts/segment.go
robskillington: Will fix the comment, ta.

persist/fs/commitlog/types.go
robskillington: Gah... yeah good call, thanks for the catch.

storage/series/series.go
robskillington: I'll put everything in the init perhaps to make sure its safe.

persist/fs/commitlog/types.go
robskillington: Fixed.

PR #214, opened 2017-01-05
Repair inconsistent blocks using peers

prateek:
This PR is now ready for review. 

~There's a few open items marked under ["TODO(prateek)"] in here. Putting this up for early feedback while working on those. This PR is branched off the pending PR for the new AdminSession endpoint ( #201), not going to merge this until that one's sorted.~


storage/repair.go
```diff
@@ -124,23 +128,266 @@ func (r shardRepairer) Repair(
...
+	}
+
+	// TODO(prateek): better handling of repaired shard writes
+	// 	- write a new version of the file for the timestamp, keep last 'n' versions, do NOT delete old version before writing a new version
+	// 	- consider not always writing, factor in a minimum number of blocks repaired, and number of blocks still requiring repair
+
+	if err := multiErr.FinalError(); err == nil {
+		return repairSummary, nil
+	}
+	return repairSummary, multiErr
```
robskillington: You can probably just `return repairSummary, multiErr.FinalError()` here and avoid the nil check as it will have the same result

storage/shard.go
```diff
@@ -726,6 +732,48 @@ func (s *dbShard) Repair(
...
+	markDirty bool,
+) error {
+
+	// ensure shard is bootstrapped
+	s.Lock()
+	if s.bs != bootstrapped {
+		s.Unlock()
+		return errShardNotBootstrapped
+	}
+	s.Unlock()
```
robskillington: Can just use an RLock/RUnlock here yeah?

storage/shard.go
```diff
@@ -726,6 +732,48 @@ func (s *dbShard) Repair(
...
+	}
+
+	// Perform update
+	err = entry.series.Update(blk)
+	entry.decrementWriterCount()
+	if err != nil {
+		return err
+	}
+
+	if markDirty {
```
robskillington: Is this ever false?  It might be worth just omitting the parameter altogether if so?

storage/series/series.go
```diff
@@ -445,6 +447,32 @@ func (s *dbSeries) mergeBlock(
...
+	if !blk.StartTime().Before(min) {
+		return errSeriesUpdateBuffered
+	}
+
+	existingBlocks := s.blocks
+	err := s.mergeBlock(existingBlocks, blk)
+	if err != nil {
+		return err
+	}
+	s.blocks = existingBlocks
```
robskillington: Don't need to reset s.blocks back to existing blocks here, mergeBlock performs update in place.

storage/repair.go
prateek: Oh true, nice!

storage/shard.go
prateek: Yep, will do

storage/shard.go
prateek: I've got a TODO in the code that is going to need this - more info (here)[https://github.com/m3db/m3db/pull/214/files?diff=unified#diff-97d06de2227f251da9f03d6e77a03927R340]

storage/series/series.go
prateek: cool, will do

prateek:
There's one big pending item still but this patch is big enough that I want to do those under another PR. 

The pending item is how we handle re-flushing a repaired shard. Current implementation doesn't add any logic here, so the code would not re-flush any repaired shard if it had previously been flushed. the changes would only be retained in memory. 

@robskillington @xichen2020 take a look please

xichen2020:
@prateek : i'd imagine some kind of versioning is needed for the on-disk files to re-flush data for a shard that has been flushed before.

prateek:
@xichen2020 yep, will do.

prateek:
Added one more change: I was observing 
Moved the repairer to run in the db ticks.

storage/namespace.go
```diff
@@ -481,6 +481,9 @@ func (n *dbNamespace) NeedsFlush(blockStart time.Time) bool {
 		if state.Status == fileOpFailed && state.NumFailures < maxRetries {
 			return true
 		}
+		if state.Status == fileOpDirty {
```
xichen2020: nit: could merge this with the condition check above into `if state.Status == fileOpNotStarted || state.Status == fileOpDirty`

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
-		return repair.MetadataComparisonResult{}, err
+		return repair.Result{}, err
 	}
 
 	metadataRes := metadata.Compare()
+	result := repair.Result{DifferenceSummary: metadataRes}
 
-	r.recordFn(namespace, shard, metadataRes)
+	res, err := r.repairFn(namespace, shard, metadataRes)
+	if err != nil {
```
xichen2020: nit: it might be cleaner to do
```
if err == nil {
  result.RepairSummary = res
}
r.recordFn(namespace, shard, result)
return result, err
```

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
-func (r shardRepairer) recordDifferences(
+type hostSet map[string]repair.HostBlockMetadata
+
+type blockID struct {
+	idHash ts.Hash
+	start  time.Time
+}
+
+func newHostSet() *hostSet {
+	hs := make(hostSet, 3)
```
xichen2020: Don't think we should use the magic number `3` here, shouldn't it be the replica count?

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+	return &hs
+}
+
+func (h *hostSet) insert(hbm repair.HostBlockMetadata) {
```
xichen2020: Think passing by value suffices here since map is a reference type, so you don't need to dereference the pointer

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+	if host := hbm.Host; host != nil {
+		(*h)[host.String()] = hbm
+	}
+}
+
+func (h *hostSet) insertOneFromSet(oh *hostSet) {
```
xichen2020: same, value type should be enough

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+	for _, hbm := range *oh {
+		h.insert(hbm)
+		return
+	}
+}
+
+func (h *hostSet) contains(host topology.Host) bool {
```
xichen2020: same, value type should be enough

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+	if host == nil {
+		return false
+	}
+	_, ok := (*h)[host.String()]
+	return ok
+}
+
+func (h *hostSet) remove(host topology.Host) {
+	delete(*h, host.String())
```
xichen2020: same, value type should be enough. Also do we need the nil check as in `contains`?

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+}
+
+func (h *hostSet) empty() bool {
```
xichen2020: same, value type should be enough

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+	return len(*h) == 0
+}
+
+func (r shardRepairer) newAdminRepairBlocksRequest(
```
xichen2020: `newRequest` might be enough?

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+	}
+	return repairBlocks
+}
+
+// newPendingReplicasMap returns a map from (series_id, block_start) -> set of host replicas,
+// using which repairs are to be performed
+//
+// NB(prateek): In the scenario when multiple peers have identical replicas for a single block, the current
+// implementation arbitrarily picks one of the peers as the source to request the replica from. In the event
+// the replica is unavailable from the selected peer, the current implementation fails. If this becomes an issue,
```
xichen2020: Yeah i think we'll need a retry mechanism just case the peer picked goes down (e.g., for deployment)

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+	idMap := make(map[ts.Hash]ts.ID, diffRes.NumBlocks)
+
+	// add size differences
+	for _, rsm := range diffRes.SizeDifferences.Series() {
+		idHash := rsm.ID.Hash()
+		idMap[idHash] = rsm.ID
+		for start, blk := range rsm.Metadata.Blocks() {
+			blkID := blockID{idHash, start}
+			// find all unique sizes seen for this block
+			uniqueValues := make(map[int64]*hostSet, 3)
```
xichen2020: We should avoid the magic number 3

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+					pendingReplicas[blkID] = newHostSet()
+				}
+				// only care about a single value from the unique set
+				pendingReplicas[blkID].insertOneFromSet(hs)
+			}
+		}
+	}
+
+	// add checksum differences
+	for _, rsm := range diffRes.ChecksumDifferences.Series() {
```
xichen2020: The logic here looks rather similar to the one above, any chance for refactoring?

storage/repair.go
```diff
@@ -124,23 +128,280 @@ func (r shardRepairer) Repair(
...
+		dirtyBlockTimes[blkID.start] = struct{}{}
+		repairSummary.NumRepaired++
+	}
+
+	// marking flush states dirty
+	dirtyBlockTimesList := make([]time.Time, 0, len(dirtyBlockTimes))
+	for t := range dirtyBlockTimes {
+		dirtyBlockTimesList = append(dirtyBlockTimesList, t)
+	}
+	shard.MarkFlushStatesDirty(dirtyBlockTimesList...)
```
xichen2020: Maybe the `MarkFlushStatesDirty` API should just take a slice then?

storage/repair.go
```diff
@@ -163,12 +427,17 @@ func (r shardRepairer) recordDifferences(
 	// Record checksum differences
 	checksumDiffScope.Counter("series").Inc(diffRes.ChecksumDifferences.NumSeries())
 	checksumDiffScope.Counter("blocks").Inc(diffRes.ChecksumDifferences.NumBlocks())
+
+	// Record ExecutionMetrics
+	executionScope.Counter("failed").Inc(execMetrics.NumFailed)
+	executionScope.Counter("pending").Inc(execMetrics.NumPending)
+	executionScope.Counter("repaired").Inc(execMetrics.NumRepaired)
+	executionScope.Counter("requested").Inc(execMetrics.NumRequested)
+	executionScope.Counter("unrequested").Inc(execMetrics.NumUnrequested)
```
xichen2020: `Unexpected` might be better?

storage/repair.go
```diff
@@ -227,57 +493,45 @@ func newDatabaseRepairer(database database) (databaseRepairer, error) {
...
-		if intervalStart == curIntervalStart {
-			continue
-		}
+func (r *dbRepairer) ShouldRun() bool {
+	if r.IsRepairing() {
+		return false
+	}
+	now := r.nowFn()
+	maxDelta := r.repairInterval + r.repairTimeJitter
+	return now.Sub(r.lastRun).Nanoseconds() > maxDelta.Nanoseconds()
```
xichen2020: We can just compare the duration without calling `Nanoseconds`

storage/series/series.go
```diff
@@ -440,6 +442,31 @@ func (s *dbSeries) mergeBlock(
...
+
+	// ensure series is bootstraped before allowing any updates
+	if s.bs != bootstrapped {
+		return errSeriesNotBootstrapped
+	}
+
+	// we only accept updates to non-buffered data
+	// ensure buffer is past provided time
+	min, _ := s.buffer.MinMax()
+	if !blk.StartTime().Before(min) {
```
xichen2020: AFAIR `MinMax` returns the min and max timestamp of datapoints in the buffer, which means if `blk.StartTime()` is the start of the current block and the earliest datapoint in the buffer is later than that, we would still repair the block, which is probably not what you want.

storage/series/series.go
```diff
@@ -440,6 +442,31 @@ func (s *dbSeries) mergeBlock(
...
+		return errSeriesUpdateBuffered
+	}
+
+	existingBlocks := s.blocks
```
xichen2020: nit: `return s.mergeBlock(s.blocks, blk)` should suffice

storage/repair/types.go
```diff
@@ -132,6 +133,34 @@ type ReplicaMetadataComparer interface {
...
+	NumRepaired int64
+
+	// NumFailed returns the number of block replicas retrieved but failed to merge into local data
+	NumFailed int64
+
+	// NumPending returns difference between the number of block replicas requested and received
+	NumPending int64
+
+	// NumUnrequested returns a count of block replicas returned but were not requested, it should be 0 in all usual executions
+	NumUnrequested int64
```
xichen2020: `NumUnexpected` might be a better name?

integration/options.go
```diff
@@ -163,6 +175,30 @@ type testOptions interface {
...
+	RepairInterval() time.Duration
+
+	// SetRepairTimeJitter sets the repair time jitter.
+	SetRepairTimeJitter(t time.Duration) testOptions
+
+	// RepairTimeJitter returns the repair time jitter.
+	RepairTimeJitter() time.Duration
+
+	// NumShards returns the number of shards.
+	NumShards() uint32
```
prateek: will do.

storage/namespace.go
prateek: will do.

storage/repair.go
prateek: will do.

integration/admin_session_fetch_blocks_high_concurrency_test.go
```diff
@@ -0,0 +1,107 @@
...
+
+	// Advance time and sleep for a long enough time so data blocks are sealed during ticking
+	testSetup.setNowFn(testSetup.getNowFn().Add(blockSize * 2))
+	later := testSetup.getNowFn()
+	time.Sleep(testSetup.storageOpts.RetentionOptions().BufferDrain() * 4)
+
+	concurrency := 16
+	// TODO(prateek): set client.AdminOptions with the two properties below
+	// SetFetchSeriesBlocksBatchSize(8).
+	// SetFetchSeriesBlocksBatchConcurrency(64).
```
robskillington: Still want to do this or keeping concurrency 16?

storage/repair.go
prateek: will fix.

storage/repair.go
prateek: good catch, i had no idea.

storage/repair.go
prateek: will do

storage/repair.go
prateek: will do

storage/repair.go
prateek: good catch, will add.

storage/repair.go
prateek: will do

storage/repair.go
prateek: sure

storage/repair.go
prateek: will do.

storage/repair.go
prateek: Makes sense

storage/repair.go
prateek: Good call

storage/repair.go
prateek: Will do

storage/series/series.go
prateek: Will do

storage/repair/types.go
prateek: Will do

storage/database.go
```diff
@@ -490,13 +485,20 @@ func (d *db) splayedTick() {
...
+		// NB(r): Cleanup and/or flush if required to cleanup files and/or
+		// flush blocks to disk. Note this has to run after the tick as
+		// blocks may only have just become available during a tick beginning
+		// from the tick begin marker.
+		if d.fsm.ShouldRun(start) {
+			d.fsm.Run(start, false)
+		}
+
+		// NB(prateek): The repairer has to be run after any pending flushes
+		if d.repairer.ShouldRun() {
```
robskillington: Does `fsm.ShouldRun` and `repairer.ShouldRun` return false if there is already an ongoing flush or repair?

storage/repair.go
prateek: Yeah makes sense. 

This isn't as quick as if we retried from another peer when we failed a block retrieval, that said, the current implementation has a top-level retry mechanism for retrying the repair on the entire range. So if a peer goes down during repair, the repairer does as best it can in the current pass; marks the repair as failed; and the blocks pending repair will be retried the next time the repairer runs. 

Does that sufficiently address the retry concern?

storage/series/series.go
prateek: The `MinMax` looks to be set independent of the datapoints - https://github.com/m3db/m3db/blob/prateek/shard-repairer/storage/series/buffer.go#L210-L220 and https://github.com/m3db/m3db/blob/prateek/shard-repairer/storage/series/buffer.go#L345

storage/database.go
robskillington: I'm thinking we should probably launch these as two separate goroutines so they don't affect each other.  Is this ok?

I really don't want to see flushing delayed 30-60mins because the last repair took extra long to run.

storage/series/series.go
xichen2020: Ah I see, seems the min timestamp is actually the start of the past 2 hour block (not the current 2 hour block), doesn't that we won't repair the past 2 hour block?

integration/admin_session_fetch_blocks_high_concurrency_test.go
prateek: good call, will address in next patch

storage/repair.go
prateek: Yep will address in the next patch.

storage/database.go
prateek: Yep, `fsm.ShouldRun` and `repairer.ShouldRun` both check for existing running tasks.

prateek: We need the repairer to only run after the flushing has finished, or we run into the case where we can attempt to repair un-flushed data. 

I think the code addresses the starvation of the `fsm`. For the situation you point out - we kick off a tick, which kicks off the single go-routine (splayedTick.Func1). It checks, and runs the `fsm` synchronously, it then checks and runs the `repairer`. While the repairer is running, we kickoff another tick, which runs another instance of the (splayedTick.Func1). It is able to `fsm.Run()` because the earlier `fsm.Run()` finished. So it shouldn't be causing a delay in flushing.

storage/series/series.go
prateek: Yep, intentionally so we can avoid repairing anything 'active' that will require an explicit buffer `Drain()`.

Is the concern that we'll wait too long before we repair the past 2 hour block?

storage/database.go
robskillington: Ah so the repairer run is asynchronous here?  Perhaps instead of passing a boolean in here we create a new enum/type so its more explicit here?

```go
type repairRunType const

const (
  repairRunSync repairRunType = iota
  repairRunAsync
)
```

Then the call becomes very explicit:
```go
d.repairer.Run(repairRunAsync)
```

integration/admin_session_fetch_blocks_high_concurrency_test.go
prateek: done

storage/repair/types.go
prateek: done

storage/series/series.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

integration/options.go
prateek: done

storage/namespace.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/repair.go
prateek: done

storage/database.go
prateek: Nah, the repairer is run synchronously within the go routine: splayedTick.Func1.

prateek: Done

integration/options.go
```diff
@@ -57,6 +57,25 @@ const (
...
+
+	// defaultNumShards sets the default number of shards.
+	defaultNumShards = 1024
+
+	// defaultFetchSeriesBlocksBatchSize sets the number of series blocks to fetch in batch, by default.
+	defaultFetchSeriesBlocksBatchSize = 0
+
+	// defaultFetchSeriesBlocksBatchConcurrency sets the number of series blocks to fetch in
+	// batch concurrently, by default.
+	defaultFetchSeriesBlocksBatchConcurrency = 0
```
robskillington: What values do we set these if they are zero?  Or we just don't set them from the default?

integration/options.go
```diff
@@ -163,6 +182,42 @@ type testOptions interface {
...
+	SetFetchSeriesBlocksBatchSize(n int) testOptions
+
+	// FetchSeriesBlocksBatchSize returns the number of Series Blocks to fetch in batch.
+	FetchSeriesBlocksBatchSize() int
+
+	// SetFetchSeriesBlocksBatchConcurrency sets the number of series blocks to fetch in batch concurrently.
+	SetFetchSeriesBlocksBatchConcurrency(n int) testOptions
+
+	// FetchSeriesBlocksBatchConcurrency returns the number of series blocks to fetch in batch concurrently.
+	FetchSeriesBlocksBatchConcurrency() int
```
robskillington: Is there a way for us to enable/disable repair altogether for integration tests?

storage/repair.go
robskillington: @prateek do you want to open an issue about this?  The peer bootstrapping has a similar issue where it will retry a fetch for a block from a node that just went down but will attempt to choose another peer (which doesn't make sense as it's only valuable because its from a peer that it wanted to stream from).

storage/repair.go
```diff
@@ -227,66 +488,52 @@ func newDatabaseRepairer(database database) (databaseRepairer, error) {
...
 		}
+		// setting lastRun regardless of success/failure
+		// to constrain the number of repairs run
+		r.lastRun = start
+	}
+
+	if mode == runTypeAsync {
+		go run()
+	} else {
+		run()
```
robskillington: Minor nit: should we standardize our usage and use the if == async in other places too?

Also this might be shorter to use everywhere:
```go
switch mode {
case runTypeAsync: go run()
default: run()
}
```

storage/series/series.go
robskillington: Yeah this looks good, it won't try to repair any block that could still receive writes.  The past 2 hour block can still be written into for the entire duration of buffer past past when its block end time is (by default up to 10mins after).

storage/shard.go
```diff
@@ -722,6 +728,50 @@ func (s *dbShard) Repair(
...
+	}
+
+	// NB(prateek): We explicitly choose to bypass the commit log during a
+	// merge for the following reasons:
+	// (1) We only bootstrap from the commit log for the last 2 hour window,
+	// this function is only used by the shardRepairer, which only operates on
+	// sealed blocks (blocks older than 2 hours). So writing to the commit log
+	// would not provide any direct value.
+	// (2) We persist the merged state by flushing 'dirty' shards
+	return nil
```
robskillington: Instead of:
```go
	if err != nil {
		return err
	}
	// lots of comments
	return nil
```

Perhaps just:
```go
	// lots of comments
	return err
```

storage/repair.go
robskillington: On second thoughts, I think the second pass thing is valid.  Let me just open one about the peer bootstrapping issue.

prateek: Sounds good.

integration/options.go
prateek: We currently control it by setting the [Repairer admin client](https://github.com/m3db/m3db/pull/214/files/4bf93c374fdd55cb46960d13334e205abc6eb500#diff-34fc2068cceaa93f8d6b312c44397541R270) only when the opts to enable repairs are [set](https://github.com/m3db/m3db/pull/214/files/4bf93c374fdd55cb46960d13334e205abc6eb500#diff-34fc2068cceaa93f8d6b312c44397541R160). This way the testSetup's repairer is only aware of peers when we enable repairs. We do run the repairer if those opts aren't set, but in that case, the admin client isn't aware of the peers so we don't actively perform repairs. 

We could make it short circuit the repair code path by disabling repairs in the [namespace opts](https://github.com/m3db/m3db/blob/dev/storage/namespace.go#L530-L532) too.

storage/repair.go
prateek: Will do.

storage/shard.go
prateek: definitely cleaner, will do.

integration/options.go
prateek: Good catch. If `0`, we fall back to the admin options default. They're used [here](https://github.com/m3db/m3db/pull/214/files/4bf93c374fdd55cb46960d13334e205abc6eb500#diff-34fc2068cceaa93f8d6b312c44397541R186)

I had the default value set to make it explicit, but it's probably a lot cleaner to just rely on 0 being the default value if the field is unset. Will address in the next patch.

storage/series/series.go
prateek: Awesome! One thing to point out, I had an off by one in the time ranges I was operating upon - this [line](https://github.com/m3db/m3db/pull/214/files/4bf93c374fdd55cb46960d13334e205abc6eb500#diff-97d06de2227f251da9f03d6e77a03927R536). 

Will change it to `end   = retention.FlushTimeEnd(r.rtopts, startTime).Add(-blockSize)` in the next version

integration/options.go
prateek: Done

storage/repair.go
prateek: done

storage/shard.go
prateek: done

integration/options.go
robskillington: That makes sense, perhaps could we make the default integration test namespaces disable repairs by default?

Just so there's less overhead when we run other integration tests and we only run repairs when we are specifically testing the repair feature?

prateek: Makes sense, will address in next patch

prateek: Done

storage/repair.go
```diff
@@ -124,23 +128,275 @@ func (r shardRepairer) Repair(
...
+
+	// track count for pending replicas
+	for blkID, hs := range pendingBlockReplicas {
+		if hs.empty() {
+			continue
+		}
+		for _, hbm := range hs {
+			missingErr := fmt.Errorf(
+				"replica requested but not received [ host = %v, id = %v, block_start = %v ]",
+				hbm.Host.String(), hashToIDMap[blkID.idHash], blkID.start)
```
robskillington: Probably best to `.String()` the ID here with `hashToIDMap[blkID.idHash].String()`

storage/repair.go
```diff
@@ -124,23 +128,275 @@ func (r shardRepairer) Repair(
...
+			multiErr = multiErr.Add(missingErr)
+			logger.Warnf("%v", missingErr)
+			repairSummary.NumPending++
+		}
+	}
+
+	// track error if blocksIter is unhappy
+	if err := blocksIter.Err(); err != nil {
+		multiErr.Add(err)
```
robskillington: This should be `multiErr = multiErr.Add(err)`

storage/repair.go
prateek: will do

storage/repair.go
prateek: do'h, will do.

PR #215, opened 2017-01-10
Add retries and checksum & size verification for peer block streaming

robskillington:
This change adds checking and verification to the checksums and sizes when peer bootstrapping.  This means if there is any interference either inside the process itself or during network transfer for the block itself the error is identified and also retried on top of to ensure it does not corrupt nor fatally stall the bootstrapping.

cc @kobolog @xichen2020 @martin-mao @prateek

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -23,6 +23,7 @@ package client
 import (
 	"errors"
 	"fmt"
+	"hash"
 	"io"
 	"math"
 	"sort"
@@ -33,6 +34,7 @@ import (
 
 	"github.com/m3db/m3db/clock"
 	"github.com/m3db/m3db/context"
+	"github.com/m3db/m3db/digest"
 	"github.com/m3db/m3db/encoding"
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	"github.com/m3db/m3db/network/server/tchannelthrift/convert"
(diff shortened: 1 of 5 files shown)


client/session.go
```diff
@@ -1824,10 +1851,79 @@ func (s *session) streamBlocksBatchFromPeer(
...
+	if block.Segments.Merged == nil && len(block.Segments.Unmerged) == 0 {
+		return fmt.Errorf("block segments is bad: merged and unmerged not set")
+	}
+
+	if metadata.checksum != nil {
+		expected := *metadata.checksum
+		digest := s.digestPool.Get().(hash.Hash32)
+		digest.Reset()
+		if block.Segments.Merged != nil {
+			if block.Segments.Merged.Head != nil {
```
xichen2020: nit: digest.Write(...) should handle `nil` safely, no need for nil checks

client/session.go
```diff
@@ -1824,10 +1851,79 @@ func (s *session) streamBlocksBatchFromPeer(
...
+				digest.Write(block.Segments.Merged.Head)
+			}
+			if block.Segments.Merged.Tail != nil {
```
xichen2020: Same here

client/session.go
```diff
@@ -1824,10 +1851,79 @@ func (s *session) streamBlocksBatchFromPeer(
...
+				digest.Write(block.Segments.Merged.Tail)
+			}
+		} else {
+			for _, segment := range block.Segments.Unmerged {
+				if segment.Head != nil {
```
xichen2020: Same here

client/session.go
```diff
@@ -1824,10 +1851,79 @@ func (s *session) streamBlocksBatchFromPeer(
...
+					digest.Write(segment.Head)
+				}
+				if segment.Tail != nil {
```
xichen2020: Same here

PR #218, opened 2017-01-12
Remove double frees in series.mergeBlock()

prateek:
The ctx registers the reader, so it's Close() call also closes the reader. This makes the explicit deferred Close() calls double frees.

/cc @robskillington @xichen2020 @cw9


xichen2020:
LGTM

PR #220, opened 2017-01-15
Fix perf issues in allocation and make bootstrappers sequential

robskillington:
Bootstrapping in parallel adds a ton of extra pressure in scheduling goroutines and using resources, this change makes existing bootstrappers sequential.  This shows a marked improvement both in write capability and also too with the duration length to bootstrap overall.  

The fs bootstrapper only uses a single goroutine to open references to file sets now which means there is less fs reader creation which helps avoid creating many large index buffers during bootstrapping and only opens the fs readers just before use by the processors.

The concurrency for peer streaming is also set to half CPUs by default instead of 1024 which just created too many concurrent streams to manage.

There is also some allocations that are heavy that don't need to be, i.e. no bytes buffering in the simple identifier pool, not using a proto buffer in the fs reader and a few other places.

cc @xichen2020 @kobolog @prateek


client/session.go
```diff
@@ -2334,7 +2315,7 @@ type enqueueChannel struct {
 
 func newEnqueueChannel(m *streamFromPeersMetrics) *enqueueChannel {
 	c := &enqueueChannel{
-		peersMetadataCh: make(chan []*blocksMetadata, 4096),
+		peersMetadataCh: make(chan []*blocksMetadata, 32768),
```
xichen2020: Better to make this a const?

persist/fs/read.go
```diff
@@ -82,6 +82,7 @@ type reader struct {
 	entriesRead int
 	indexUnread []byte
 	prologue    []byte
+	protoBuff   *proto.Buffer
```
xichen2020: `protoBuf` might be more consistent with `unreadBuf` and `digestBuf`?

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -78,33 +78,41 @@ func (s *peersSource) Read(
...
 	for shard, ranges := range shardsTimeRanges {
-		it := ranges.Iter()
-		for it.Next() {
-			currRange := it.Value()
-			// We fetch all results in parallel and the underlying session will
-			// rate limit the total throughput with a single worker pool that
-			// is defined by the fetch series blocks batch concurrency option
-			wg.Add(1)
-			go func(shard uint32, start, end time.Time) {
+		shard, ranges := shard, ranges
```
prateek: nit: unnecessary var assignment?

client/session.go

persist/fs/read.go
robskillington: Sure thing.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: Yeah, should remove this now that we don't need to capture it for the goroutine that used to encapsulate the inner logic.

robskillington: Done.

PR #221, opened 2017-01-17
Add ability to expire blocks data and avoid bootstrapping block data

robskillington:
This change introduces the ability to bootstrap just the metadata for blocks and fetch the data from disk as necessary.  When the data is fetched it is cached in memory and during the database ticking if data hasn't been accessed for a certain duration, by default 5 minutes, it will flush the data from memory and make it retrievable from disk again.

cc @xichen2020 @kobolog @martin-mao @prateek


encoding/m3tsz/encoder.go
```diff
@@ -526,6 +526,11 @@ func (enc *encoder) Stream() xio.SegmentReader {
 	return xio.NewSegmentReader(segment)
 }
 
+func (enc *encoder) StreamLen() int {
+	buff, _ := enc.os.Rawbytes()
```
xichen2020: Worth checking if `buff` is nil?

persist/fs/retriever.go
```diff
@@ -0,0 +1,523 @@
...
+	reqsByShardIdx := make([]*shardRetrieveRequests, shard+1)
+
+	for i := range reqsByShardIdx {
+		if i < len(r.reqsByShardIdx) {
+			reqsByShardIdx[i] = r.reqsByShardIdx[i]
+			continue
+		}
+		reqsByShardIdx[i] = &shardRetrieveRequests{
+			shard:  uint32(i),
+			queued: make([]*retrieveRequest, 0, 4096),
```
xichen2020: Make `4096` a const?

retention/types.go
```diff
@@ -64,10 +64,24 @@ type Options interface {
 	ShortExpiry() bool
 
 	// SetShortExpiryPeriod sets the period that blocks should be kept in memory
-	// when  short expiry  is turned on
+	// when  short expiry is turned on
```
xichen2020: Remove extra space between `when` and `short`?

storage/block/block.go
```diff
@@ -31,95 +33,260 @@ import (
...
 
+func (b *dbBlock) now() time.Time {
+	nowFn := b.opts.ClockOptions().NowFn()
+	return nowFn()
+}
+
 func (b *dbBlock) StartTime() time.Time {
-	return b.start
+	b.RLock()
+	start := b.startWithLock()
```
xichen2020: Do we need a lock here? seems we can just atomic load `b.startUnixNano` no?

storage/block/block.go
```diff
@@ -31,95 +33,260 @@ import (
...
-	// are taken to the bytes refs and to not finalize the bytes.
-	s.Reset(ts.NewSegment(b.segment.Head, b.segment.Tail, ts.FinalizeNone))
-	blocker.RegisterFinalizer(context.FinalizerFn(s.Close))
-	return s, nil
+
+	if b.mergeTarget != nil {
+		var mergeStream xio.SegmentReader
+		mergeStream, err = b.mergeTarget.Stream(blocker)
+		if err != nil {
+			stream.Finalize()
```
xichen2020: Need `b.RUnlock` here. Any reason not to use defer so we don't need to track all the places where we could return?

storage/block/block.go
```diff
@@ -31,95 +33,260 @@ import (
...
+}
+
+func (b *dbBlock) IsRetrieved() bool {
+	b.RLock()
+	retrieved := b.retriever == nil
+	b.RUnlock()
+	return retrieved
+}
+
+func (b *dbBlock) Merge(other DatabaseBlock) {
```
xichen2020: Just curious, when will we use `Merge`?

storage/block/block.go
```diff
@@ -31,95 +33,260 @@ import (
...
 	if !b.closed {
 		b.ctx.Close()
 	}
 	b.ctx = b.opts.ContextPool().Get()
-	b.start = start
+	b.startUnixNano = start.UnixNano()
 	b.closed = false
-	b.resetSegment(segment)
+	b.resetMergeTargetWithLock()
+	atomic.StoreInt64(&b.accessedUnixNano, b.now().UnixNano())
```
xichen2020: This looks very similar to `Reset`, can we move all but the last line into a separate method and reuse it from `Reset` and `ResetRetrievable`?

storage/bootstrap.go
```diff
@@ -110,7 +110,8 @@ func (m *bootstrapManager) targetRanges(at time.Time) xtime.Ranges {
 	midPoint := at.
 		Add(-ropts.BlockSize()).
 		Add(-ropts.BufferPast()).
-		Truncate(ropts.BlockSize())
+		Truncate(ropts.BlockSize()).
+		Add(ropts.BlockSize())
```
xichen2020: hm why do we need to add block size here?

storage/bootstrap/result/result.go
```diff
@@ -65,8 +65,22 @@ func (r *bootstrapResult) AddResult(other BootstrapResult) {
...
+		size += len(sr.AllSeries())
+	}
+	for _, sr := range other.ShardResults() {
+		sizeOther += len(sr.AllSeries())
+	}
+	if size >= sizeOther {
+		r.results.AddResults(other.ShardResults())
+		r.unfulfilled.AddRanges(other.Unfulfilled())
+	} else {
+		other.AddResult(r)
```
xichen2020: Are we ok to destroy `other` here?

storage/namespace.go
```diff
@@ -230,15 +265,33 @@ func (n *dbNamespace) Tick(softDeadline time.Duration) {
...
 
 	// Tick through the shards sequentially to avoid parallel data flushes
-	var r tickResult
+	var (
+		r  tickResult
+		l  sync.Mutex
+		wg sync.WaitGroup
+	)
 	perShardDeadline := softDeadline / time.Duration(len(shards))
+	perShardDeadline *= time.Duration(n.tickWorkersConcurrency)
```
xichen2020: Why do we need to multiply here? Seems counterintuitive that the deadline is longer with higher concurrency

storage/series/series.go
```diff
@@ -108,27 +117,35 @@ func (s *dbSeries) Tick() (TickResult, error) {
...
 	s.Lock()
-	// Recalculate in case changed during lock promotion
-	needsDrain = s.buffer.NeedsDrain()
-	needsBlockUpdate = s.needsBlockUpdateWithLock()
+
+	// NB(r): don't bother to check if needs drain or needs
+	// block update still holds as running both these checks
+	// are relatively expensive and running these methods
+	// will be a no-op in case conditions no longer hold.
 	if needsDrain {
```
xichen2020: That means if `needsDrain` changes from false to true in between locking, we won't drain and reset it, is that ok?

encoding/m3tsz/encoder.go
robskillington: Good call, instead I'll use `OStream Len() int` which already does the `nil` check however.

persist/fs/retriever.go
robskillington: Sure thing.

retention/types.go
robskillington: Sure thing.

storage/block/block.go
robskillington: We could but we don't ever frequently write the start time, hence its just as quick to use an `RLock` here.  Also we extremely rarely call this method, as we iterate over the block start keys of the map the `DatabaseBlock` is in anyway instead of checking the `StartTime() time.Time` for each block.

If there are no writers pending an `RLock` and `RUnlock` is two simple `AddInt32`s anyhow:
```
https://golang.org/src/sync/rwmutex.go?s=1294:1320#L26
```

storage/block/block.go
robskillington: Good catch, yeah probably should use defer.  I was worried because this is called so often about the overhead, however I believe defer got about 50% quicker with go 1.8:
https://tip.golang.org/doc/go1.8#defer

storage/block/block.go
robskillington: We use this when a bunch of bootstrapped blocks or drained blocks are added to a series (either with `bufferDrained(...)` or `Bootstrap(...)`) and there is existing blocks.

storage/block/block.go
robskillington: Yeah fair.  Will consolidate into a single `reset(...)` they call then branch off after that.

storage/bootstrap.go
robskillington: So after looking at logs, because we bootstrap using the "from" inclusive and "to" exclusive - it meant that we weren't properly bootstrapping everything up to the latest block for the FS bootstrapper.  We were bootstrapping up to just before the latest block in the first pass with the FS bootstrapper, then would do the latest flushed block from FS bootstrapper and the current block from Peers bootstrapper.  

This is not what we want, this change corrects this behavior.

storage/bootstrap/result/result.go
robskillington: Can change the naming/or something to reflect this perhaps?  This was a definite issue we saw though when merging the zero sized FS bootstrap result with a huge Peers bootstrap result.  We basically added each series one by one into the empty FS bootstrap result.  

Definitely want to avoid this case.

storage/namespace.go
robskillington: So the deadline is calculated for how much time each shard should get if we were ticking in a serial manner.  

The deadline is multiplied because we will be ticking concurrently, so essentially we are meeting the same `softDeadline` overall, but give each shard a little longer because they are being done concurrently.

storage/series/series.go
robskillington: This is fine, it's exactly the same as if this method was called a few nanoseconds earlier and needsDrain would've been false in both places.  This helped save a lot of CPU for ticking.  

Besides when checking when to "flush" or not at the end of a tick, the flush manager looks at the start time of the tick:
```go
// Begin ticking
start := d.nowFn()

// .. actually do the tick

if d.fsm.ShouldRun(start) {
  // actually flush
}
```

encoding/m3tsz/encoder.go
xichen2020: Sounds good

storage/block/block.go
xichen2020: Sure sounds good

storage/bootstrap.go
xichen2020: Ah okay, may be worth adding a test here to catch it if there isn't already one?

storage/bootstrap/result/result.go
xichen2020: I see, perhaps document this behavior more explicitly?

Alternatively we could follow c++'s convention and make this method a non-member method that takes two results as the arguments and returns a result
```
MergeResults(result1, result2 BootstrapResult) BootstrapResult
```
Because this is not a member method, users are less inclined to assume this affects the object on which the method is called.

storage/series/series.go
xichen2020: Sounds good

storage/block/block.go
robskillington: Done.

storage/block/block.go
robskillington: Done.

storage/bootstrap.go
robskillington: Sure thing.

storage/bootstrap/result/result.go
robskillington: Yeah, can make this just a module method.  Sounds good.

robskillington:
Abandoning as we are using master as trunk for now.

PR #222, opened 2017-01-17
Cleanup persist/fs.ReadInfoFile usage

prateek:
tiny changes: 
- Make fs bootstrapper read start from protobuf records within Info file rather than the filename. 
- Add a unit test for Info file read/write coverage

### DIFF
--- a/storage/bootstrap/bootstrapper/fs/source.go
+++ b/storage/bootstrap/bootstrapper/fs/source.go
@@ -116,11 +116,7 @@ func (s *fileSystemSource) enqueueReaders(
 	readersCh chan<- shardReaders,
 ) {
 	for shard, tr := range shardsTimeRanges {
-		var files []string
-		fs.ForEachInfoFile(s.filePathPrefix, namespace, shard, s.readerBufferSize, func(fname string, _ []byte) {
-			files = append(files, fname)
-		})
-
+		files := fs.ReadInfoFiles(s.filePathPrefix, namespace, shard, s.readerBufferSize)
 		if len(files) == 0 {
 			// Use default readers value to indicate no readers for this shard
 			readersCh <- shardReaders{shard: shard, tr: tr}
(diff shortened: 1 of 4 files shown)


storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -116,11 +116,7 @@ func (s *fileSystemSource) enqueueReaders(
 	readersCh chan<- shardReaders,
 ) {
 	for shard, tr := range shardsTimeRanges {
-		var files []string
-		fs.ForEachInfoFile(s.filePathPrefix, namespace, shard, s.readerBufferSize, func(fname string, _ []byte) {
-			files = append(files, fname)
-		})
-
+		files := fs.ReadInfoFiles(s.filePathPrefix, namespace, shard, s.readerBufferSize)
```
xichen2020: This doesn't check for the existence of checkpoint files or perform digest validation, is there a reason for this change?

robskillington: Perhaps we should make one place where these checks happen and make sure all our APIs flow through the one place that can read everything and you simply pass in a struct describing which things you want to read and what level of validation?

That way we can centralize our layers and ensure it's explicit about what you are parsing and validating.

prateek: @r yep that was the intent of this change. The file `source.go` is the only non-package caller of the persist/fs ReadInfo* calls. This change makes the two calls use the `ReadInfoFiles` method.

@xichen2020 `source.go` earlier called `ForEachInfoFile` and `ReadInfoFiles` both. Now it calls `ReadInfoFiles` in both places. In turn, `ReadInfoFiles` internally calls `ForEachInfoFile` - [1] (renamed to `forEachInfoFile`), so it still performs all the sanity checking you're asking about - [2]. It performs one additional check compared to earlier - the readInfo call in the internal func definition.

[1]: https://github.com/m3db/m3db/blob/9f335641e25380c73eb6b8dc61c320738f1779ad/persist/fs/files.go#L209-L214
[2]: https://github.com/m3db/m3db/blob/9f335641e25380c73eb6b8dc61c320738f1779ad/persist/fs/files.go#L177-L200

xichen2020: Ah I see, misread it somehow :\ although the new change makes it read the info data even though it doesn't need to, which adds a bit of overhead.

prateek: All good. I think standardising on where we're reading data from is worth the overhead, so earlier we used blockstart time extracted from filename and from the info file both, in separate code-paths. This way at least we avoid doing that. 

Also, wrt to the overhead, `readInfo` is only doing unmarshalling so I don't think it's it's going to be much of a change.

xichen2020: I wonder if it's possible instead to dictate what kind of validation/data reading we perform by passing in an enum flag etc so that we don't perform additional unnecessary operation if we don' thave

prateek: Sure, lemme take a pass at it.

PR #224, opened 2017-01-21
Fix travis builds by pinning glide to v0.12.3

prateek:
- Travis builds have been broken since `glide 0.13.0`, best guess is the issue has to do with: https://github.com/Masterminds/glide/issues/745
- Eg broken builds:  
  - https://travis-ci.org/m3db/m3db/builds/194081709
  - https://travis-ci.org/m3db/m3cluster/jobs/194083048
- This patch pins the glide install to `0.12.3` which works around the issue for now.

/cc @xichen2020 @robskillington @kobolog @cw9

PR #226, opened 2017-01-22
Fix travis builds by pinning glide to v0.12.3 

prateek:
details here: #224 

/cc @cw9

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ install-proto-bin: install-vendor
 	go install $(m3db_package)/$(vendor_prefix)/$(protoc_go_package)
 
 install-glide:
-	@which glide > /dev/null || go get -u github.com/Masterminds/glide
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -93,7 +93,7 @@ install-proto-bin: install-vendor
 	go install $(m3db_package)/$(vendor_prefix)/$(protoc_go_package)
 
 install-glide:
-	@which glide > /dev/null || go get -u github.com/Masterminds/glide
+	@which glide > /dev/null || mkdir -p $(GOPATH)/src/github.com/Masterminds && cd $(GOPATH)/src/github.com/Masterminds && git clone https://github.com/Masterminds/glide.git && cd glide && git checkout v0.12.3 && go install
```
cw9: any particular reason to prefer git clone than go get here?

robskillington: Yeah have a feeling we should use `go get` and then just cd, checkout and install.  `go get` will check it out as a git repository (or whatever repository it ends up being).

prateek: Sounds good. Making new version.

prateek: @cw9 @robskillington done

PR #228, opened 2017-01-23
Exclude glide tests to avoid pinning glide

robskillington:


robskillington:
Does not fix submodule checkout issues with tip of glide.  Going to keep pinning.

PR #238, opened 2017-02-01
Explicitly control bitset reallocation in commit logs

xichen2020:
cc @robskillington @kobolog @martin-mao @cw9

This PR addresses an inefficiency in how we use bitset in commit logs. The current bitset implementation keeps track of the maximum value it has seen so far and triggers a reallocation if it needs to set a value larger the current max. However it only reallocates space enough to represent the new value, which means in case each reallocation only grows the internal slice the bitset implementation by 1 slot and copies the existing over, causing significant CPU overhead and GC time. As a result, we explicitly control when the reallocation happens and grow the slice explicitly to amortize the allocation cost.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: a24ea9a4ca6dce78a27abcd75db43c68ac00a460a68e1603770a55dc94502674
-updated: 2017-01-23T16:23:13.981943905-05:00
+hash: 029ba941f724d765bde940f3b4ab40f4d61c19fafac214db369d9b32117e22f3
+updated: 2017-02-01T14:26:06.521913538-05:00
 imports:
 - name: github.com/apache/thrift
   version: 53dd39833a08ce33582e5ff31fa18bb4735d6731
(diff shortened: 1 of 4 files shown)


robskillington:
Hey @xichen2020 are we just using an old version?  Seems it already does 2x growth:
https://github.com/willf/bitset/blob/master/bitset.go#L134
```
func (b *BitSet) extendSetMaybe(i uint) {
	if i >= b.length { // if we need more bits, make 'em
		nsize := wordsNeeded(i + 1)
		if b.set == nil {
			b.set = make([]uint64, nsize)
		} else if cap(b.set) >= nsize {
			b.set = b.set[:nsize] // fast resize
		} else if len(b.set) < nsize {
			newset := make([]uint64, nsize, 2*nsize) // increase capacity 2x
			copy(newset, b.set)
			b.set = newset
		}
		b.length = i + 1
	}
}
```

xichen2020:
Yeah looks like we were using an older version :\ upgraded the version and adapted the test

PR #239, opened 2017-02-01
Add option to disable repair in database

cw9:
For the m3db in collector, there will be no peers to repair from,
should have an option to just turn repair off and save some resource

### DIFF
--- a/storage/repair.go
+++ b/storage/repair.go
@@ -614,5 +614,5 @@ func newNoopDatabaseRepairer() databaseRepairer {
 }
 
 func (r noopRepairer) ShouldRun(start time.Time) bool    { return false }
-func (r noopRepairer) Run(start time.Time, mode runType) { return }
+func (r noopRepairer) Run(start time.Time, mode runType) {}
 func (r noopRepairer) IsRepairing() bool                 { return false }
--- a/storage/repair.go
+++ b/storage/repair.go
@@ -606,3 +606,13 @@ func (r *dbRepairer) repairWithTimeRange(tr xtime.Range) error {
 func (r *dbRepairer) IsRepairing() bool {
 	return atomic.LoadInt32(&r.running) == 1
 }
+
+type noopRepairer struct{}
+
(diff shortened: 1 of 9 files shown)


cw9:
Did we use to have a toggle like this and then removed? 

@xichen2020 @robskillington @prateek

xichen2020:
Enable/Disable repair is done via namespace-level options

cw9:
Then for cases like this (don't need repair), I still need to create an admin client because the database has to create a repairer and that requires the repair options to have an admin client to be valid.

xichen2020:
In that case I suggest
- adding a new noopRepairer
- adding a `NewDatabaseRepairerFn` to the database options
- in your usecase, make the repairer ctor return the noopRepairer

cw9:
I added the no-op repairer, I'm hesitating on exposing the NewDatabaseRepairerFn() in the Options, because for the use case we have, we basically just need to toggle between 2 repairers.

xichen2020:
hm why not? This makes it very flexible in terms of what type of repairer the database uses, even though for our usecases it's either a no-op repairer or a full-fledged repairer, but it doesn't have to be and other users of the repo can pass in their custom repairer if they choose to

cw9:
because the way we create the databaseRepairer right now requires both database interface and databaseRepairer interface. It seems to be easy to expose the databaseRepairer interface, but I'm not sure about changing database to Database

storage/repair.go
```diff
@@ -606,3 +606,13 @@ func (r *dbRepairer) repairWithTimeRange(tr xtime.Range) error {
...
 }
+
+type noopRepairer struct{}
+
+func newNoopDatabaseRepairer() databaseRepairer {
+	return noopRepairer{}
+}
+
+func (r noopRepairer) ShouldRun(start time.Time) bool    { return false }
+func (r noopRepairer) Run(start time.Time, mode runType) { return }
```
xichen2020: nit: `{}` is slightly cleaner

PR #240, opened 2017-02-03
Use Msgpack to encode and decode block filesets and commit logs

xichen2020:
cc @robskillington @martin-mao @kobolog @prateek @cw9 

This PR uses MessagePack for serializing / deserializing on-disk data including block filesets and commit logs, replacing protobuf due to MessagePack's superior performance (not reflection-based) and memory usage (no allocation for byte slices by default during decoding)


glide.lock
```diff
@@ -74,7 +74,22 @@ imports:
 - name: github.com/willf/bitset
   version: 5c3c0fce48842b2c0bbaa99b4e61b0175d84b47c
 - name: golang.org/x/net
-  version: f2499483f923065a842d38eb4c7f1927e6fc6e6d
+  version: 007e530097ad7f954752df63046b4036f98ba6a6
   subpackages:
   - context
+- name: google.golang.org/appengine
```
robskillington: Msgpack relies on appengine?

persist/encoding/msgpack/decoder.go
```diff
@@ -0,0 +1,369 @@
...
+func (dec *decoder) decodeBytesLen() int {
+	if dec.err != nil {
+		return 0
+	}
+	value, err := dec.dec.DecodeBytesLen()
+	dec.err = err
+	return value
+}
+
+func (dec *decoder) readBytes(start int, n int) []byte {
```
robskillington: Can make this `start, n int` yeah?

persist/encoding/msgpack/decoder.go
```diff
@@ -0,0 +1,369 @@
...
+	numBytes := len(dec.data)
+	if n < 0 || start < 0 || start+n > numBytes {
```
robskillington: Maybe also check that `start < n`?

persist/encoding/msgpack/decoder.go
```diff
@@ -0,0 +1,369 @@
...
+		dec.err = fmt.Errorf("invalid start %d and length %d, numBytes=%d", start, n, numBytes)
+		return nil
+	}
+	_, err := dec.reader.Seek(int64(n), io.SeekCurrent)
```
robskillington: Any reason not to `Seek(int64(start+n), io.SeekStart)` just to be sure we land where we think we should?

persist/encoding/msgpack/decoder.go
robskillington: May it be worth just inlining this whole function?  It's only used in one place from decodeBytes no?

Mainly because: it's strange that the parameters are absolute numbers, but the reader is moved a relative amount based on how much is being read from an absolute position.

persist/encoding/msgpack/options.go
```diff
@@ -0,0 +1,57 @@
...
+
+// DecodingOptions provide a set of options for decoding data
+type DecodingOptions interface {
+	// SetNewAllocForBytes sets whether we allocate new space when decoding
+	// a byte slice
+	SetNewAllocForBytes(value bool) DecodingOptions
+
+	// NewAllocForBytes determines whether we allocate new space when decoding
+	// a byte slice
+	NewAllocForBytes() bool
```
robskillington: Might `AllocReturnedBytes()` be a better name?

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -38,13 +39,15 @@ type newFileSetReaderFn func(
...
 	bytesPool pool.CheckedBytesPool,
+	decodingOpts msgpack.DecodingOptions,
 ) fs.FileSetReader
 
 type fileSystemSource struct {
 	opts             Options
 	log              xlog.Logger
 	filePathPrefix   string
 	readerBufferSize int
+	decodingOpts     msgpack.DecodingOptions
```
robskillington: Seems like this never gets set from the constructor?  Should `bootstrapper/fs/Options` perhaps get a `DecodingOptions()` and `SetDecodingOptions(value) Options` methods?

glide.lock
xichen2020: Yeah it has special APIs for appengine key/value types (shrug)

persist/encoding/msgpack/decoder.go
xichen2020: Sure I'm fine with inlining the function

persist/encoding/msgpack/decoder.go
xichen2020: Okay, although both essentially do the same thing

persist/encoding/msgpack/options.go
xichen2020: Sure

storage/bootstrap/bootstrapper/fs/source.go
xichen2020: Yeah I already added the decoding options to `FileSystemOptions` but forgot to populate it here, good catch

PR #243, opened 2017-02-07
Ensure client write errors propagate retryable, non-retryable, and bad request errors

ben-lerner:
Correctly propagate error types in write.go

### DIFF
--- a/client/write_test.go
+++ b/client/write_test.go
@@ -21,6 +21,7 @@
 package client
 
 import (
+	"errors"
 	"sync"
 	"testing"
 
@@ -28,6 +29,8 @@ import (
 	"github.com/m3db/m3db/topology"
 
 	"github.com/golang/mock/gomock"
+	tterrors "github.com/m3db/m3db/network/server/tchannelthrift/errors"
+	xerrors "github.com/m3db/m3x/errors"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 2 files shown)


client/write_test.go
```diff
@@ -107,3 +110,102 @@ func setShardStates(t *testing.T, s *session, host topology.Host, state shard.St
...
+
+		s = newDefaultTestSession(t).(*session)
+		w := newWriteStub()
+
+		var completionFn completionFn
+		enqueueWg := mockHostQueues(ctrl, s, sessionTestReplicas, []testEnqueueFn{func(idx int, op op) {
+			completionFn = op.CompletionFn()
+		}})
+
+		s.Open()
```
robskillington: Should probably `require.NoError(t, s.Open())` here

client/write_test.go
```diff
@@ -107,3 +110,102 @@ func setShardStates(t *testing.T, s *session, host topology.Host, state shard.St
...
+		defer s.Close()
```
robskillington: Also should probably check close error too?

ben-lerner: Fixed both

client/write_test.go
```diff
@@ -107,3 +112,104 @@ func setShardStates(t *testing.T, s *session, host topology.Host, state shard.St
...
+		require.True(t, s.topoMap.Replicas() == sessionTestReplicas)
+		for i := 0; i < s.topoMap.Replicas(); i++ {
+			completionFn(host, nil) // maintain session state
+		}
+	}
+
+	testCleanup := func() {
+		wState.decRef() // end introspection
+		writeWg.Wait()  // wait for write to complete
+	}
```
robskillington: Instead of a testsetup and a testcleanup - perhaps instead use a test for each one of these cases for better isolating an issue if it comes up.  i.e. instead of one huge test failing, you have one of many smaller ones failing so you can track down the issue faster.

i.e.

```go
type testWriteFn func(
  	wState *writeState,
  	s *session,
  	host topology.Host,
  	writeWg sync.WaitGroup,
)

func testWrite(t *testing.T, fn testWriteFn) {
  	// all the code for test setup

  	fn(wState, s, host, writeWg)

  	// all the code for test cleanup
}

// now write tests 

func TestWriteNonRetryableError(t *testing.T) {
    testWrite(t,  func(wState *writeState, s *session, host topology.Host, writeWg sync.WaitGroup) {
 	 	wState.completionFn(host, xerrors.NewNonRetryableError(errors.New("")))
 	  	require.True(t, len(wState.errors) == 1)
 	  	assert.True(t, xerrors.IsNonRetryableError(wState.errors[0]))
    })
}

// other tests

File: client/write_test.go [hunk 4]
```diff
@@ -21,30 +21,141 @@
...
+
+type errTestFn func(error) bool
+
+func retryabilityCheck(t *testing.T, wState *writeState, writeWg *sync.WaitGroup, testFn errTestFn) {
+	require.True(t, len(wState.errors) == 1)
+	assert.True(t, testFn(wState.errors[0]))
+
+	// cleanup
+	wState.decRef() // end introspection
+	writeWg.Wait()  // wait for write to complete
```
robskillington: Hm feels strange to have common cleanup code here when `testSetup` is broken out and independent.  Also since the functions declared in this file is global to the package `*_test.go` files it might be better to call it `writeTestSetup` and `writeTestTeardown` so its clear where to use them.

xichen2020:
Can we cherry-pick this back to dev once it's merged into master?

ben-lerner:
cherry-picked

xichen2020:
Thanks!

PR #248, opened 2017-02-08
Add identifier pool to client session to avoid allocations and potential stack splits

robskillington:
Using `ts.ID` with `checked.Bytes` the allocations become quite heavy at large write volumes.  This change adds an identifier pool to the client session to avoid these ongoing heavy allocations.

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -110,6 +110,8 @@ type session struct {
 	queues                           []hostQueue
 	queuesByHostID                   map[string]hostQueue
 	state                            state
+	contextPool                      context.Pool
+	idPool                           ts.IdentifierPool
 	writeOpPool                      writeOpPool
 	fetchBatchOpPool                 fetchBatchOpPool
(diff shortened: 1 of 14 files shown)


client/session.go
```diff
@@ -653,8 +659,9 @@ func (s *session) Write(namespace, id string, t time.Time, value float64, unit x
 
 	// todo@bl: Can we combine the writeOpPool and the writeStatePool?
 	state.op, state.majority = s.writeOpPool.Get(), majority
+	state.ctx, state.nsID, state.tsID = ctx, state.nsID, state.tsID
```
xichen2020: hm why not just `state.ctx = ctx`?

ben-lerner: should probably be `state.ctx, state.nsID, state.tsID = ctx, nsID, tsID`

robskillington: Great catch, will fix this.

PR #250, opened 2017-02-09
Keep less open file descriptors for fs Seeker and Reader

robskillington:
Using a high shard count with a lot of open seekers the process can use a ton of FDs:
```
250 shards * 24 filesets * (info fd + index fd + digest fd + data fd) = 24,000
```

We can cut down the amount of idle open FDs by 75% by closing them after the initial open.

### DIFF
--- a/persist/fs/seek.go
+++ b/persist/fs/seek.go
@@ -110,7 +110,7 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
 	s.digestFdWithDigestContents.Reset(digestFd)
 
 	defer func() {
-		// NB(r): We don't need to keep these FDs open as we these up front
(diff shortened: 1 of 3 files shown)


persist/fs/seek.go
```diff
@@ -262,8 +271,11 @@ func (s *seeker) Entries() int {
...
 func (s *seeker) Close() error {
-	return closeAll(
-		s.infoFdWithDigest,
-		s.indexFdWithDigest,
-	)
+	// Prepare for reuse
+	for key := range s.indexMap {
+		delete(s.indexMap, key)
+	}
+	err := s.dataFd.Close()
```
xichen2020: need a nil check here in case `Open` failed

robskillington: Sure thing.

PR #251, opened 2017-02-09
Report shard states from cluster database

xichen2020:

This PR reports shard states for m3db nodes so we have better visibility into shard movements / progress when cluster topology changes


ben-lerner:
LGTM

PR #253, opened 2017-02-09
Add more detailed error messages for peer streaming

xichen2020:

This PR records the errors that happened when fetching blocks from peers so we can better understand why streaming sometimes fails during peer bootstrapping

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -1884,7 +1884,7 @@ func (s *session) streamBlocksBatchFromPeer(
 		if !bytes.Equal(id.Data().Get(), result.Elements[i].ID) {
 			b := batch[i].blocks
 			blocksErr := fmt.Errorf(
-				"stream blocks mismatched ID: expectedID=%s,actualID=%s,indexID=%d,peer=%s",
+				"stream blocks mismatched ID: expectedID=%s, actualID=%s, indexID=%d, peer=%s",
 				batch[i].id.String(), id.String(), i, peer.Host().String(),
 			)
 			s.reattemptStreamBlocksFromPeersFn(b, enqueueCh, blocksErr, respErrReason, m)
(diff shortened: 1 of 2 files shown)


client/session.go
```diff
@@ -1881,14 +1883,13 @@ func (s *session) streamBlocksBatchFromPeer(
 		id := batch[i].id
 		if !bytes.Equal(id.Data().Get(), result.Elements[i].ID) {
 			b := batch[i].blocks
-			s.reattemptStreamBlocksFromPeersFn(b, enqueueCh, respErrReason, m)
+			blocksErr := fmt.Errorf(
+				"stream blocks mismatched ID: expectedID=%s,actualID=%s,indexID=%d,peer=%s",
```
robskillington: Spaces perhaps between the "," to match the above err?

client/session.go
```diff
@@ -2005,28 +2010,35 @@ func (s *session) streamBlocksReattemptFromPeers(
...
 	for i := range blocks {
 		// Reconstruct peers metadata for reattempt
 		reattemptBlocksMetadata :=
 			make([]*blocksMetadata, len(blocks[i].reattempt.peersMetadata))
 		for j := range reattemptBlocksMetadata {
+			reattempt := blocks[i].reattempt
+			// Copy the errors for every peer so they don't shard the same error
+			// slice and therefore are not subject to race conditions when the
+			// error slice is modified
+			reattempt.errs = append([]error{attemptErr}, reattempt.errs...)
```
robskillington: Perhaps `append(append([]error(nil), reattempt.errs...), attemptErr)` for in order errors?

client/session.go
xichen2020: can do

client/session.go
xichen2020: sure thing

PR #255, opened 2017-02-10
Make seeker reusable

cw9:
We were setting decoder to nil after calling seeker.readIndex, which causes future read calls to panic on nil decoder, I don't think it's necessary to set the decoder to nil, let me know your thoughts, will add unit test if this approach looks good.

### DIFF
--- a/persist/fs/seek.go
+++ b/persist/fs/seek.go
@@ -181,6 +181,13 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
 		return err
 	}
 
+	if !s.keepUnreadBuf {
+		// NB(r): Free the unread buffer and reset the decoder as unless
+		// using this seeker in the seeker manager we never use this buffer again
+		s.unreadBuf = nil
+		s.decoder.Reset(nil)
+	}
+
 	s.dataFd = dataFd
 	return nil
 }
@@ -256,12 +263,6 @@ func (s *seeker) readIndex(size int) error {
 			id := ts.BinaryID(checked.NewBytes(entryID, nil))
 			s.indexIDs = append(s.indexIDs, id)
 		}
-	}
-
-	if !s.keepUnreadBuf {
-		// NB(r): Free the unread buffer and reset the decoder as unless
-		// using this seeker in the seeker manager we never use this buffer again
-		s.unreadBuf = nil
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


persist/fs/seek.go
```diff
@@ -262,7 +262,6 @@ func (s *seeker) readIndex(size int) error {
 		// NB(r): Free the unread buffer and reset the decoder as unless
 		// using this seeker in the seeker manager we never use this buffer again
 		s.unreadBuf = nil
-		s.decoder = nil
```
xichen2020: In that case can you call `s.decoder.Reset(nil)` instead? Otherwise the decoder holds on to the underlying byte slice which could be a significant chunk of memory

cw9: Sounds good, will call Reset().

keepUnreadBuf is statically set to false now in the seeker constructor

persist/fs/seek.go
```diff
@@ -181,6 +181,13 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
 		return err
 	}
 
+	if !s.keepUnreadBuf {
```
xichen2020: Why move it inside `Open`? The block retrieve manager explicitly reuses the buffer across multiple `Open` calls to reduce memory allocation and GC, and moving it inside `Open` defeats this purpose

cw9: hmm? it was called in readIndex and readIndex is only called in Open(), I moved it here because if we call readIndex before readInfo, the readInfo call would panic on nil decoder as well, so it's safer to call it here?

xichen2020: The primary use case is:
```
1. Set the unread buffer externally for reuse across multiple seekers
2. Call `Open` for each seeker
3. In readIndex, the buffer is nil'ed out after the index is read
```

If we instead nil out the buffer early on in `Open`, it means the buffer will always be reallocated instead of being reused across seekers, which incurs significant memory/GC overhead.

cw9: I'm not following, I think I'm still resetting the buffer after the index is read? although the code is moved but it is literally still in the same place in the code path?

xichen2020: Ah I see, yeah I misread it and thought somehow this was called before the `readIndex` call. It's probably okay to leave it since both `readInfo` and `readIndex` are unexported methods but I'm fine with moving it to `Open`, too

cw9: yeah I felt it's nicer to be moved to Open since it's not strongly related to readIndex and this way we don't necessarily have to do readInfo first and then readIndex.

PR #257, opened 2017-02-11
Format session errors with fmt.Sprintf(..) vs custom fmt

robskillington:
This PR uses the inbuilt `fmt.Sprintf(...)` to format errors in the session select blocks from peers method.


robskillington:
Perhaps JSON using the JSON marshaller then?

PR #258, opened 2017-02-13
Maintain block order after peer selection or exhaustion during peer streaming

xichen2020:

This PR fixes an issue when removing blocks from peers after we select a block from multiple peer candidate or the peer list has been exhausted. In particular when a block is removed, we used to swap it with the last block and shrink the block list, thereby destroying invariant that the blocks are sorted by their start times. This in turn causes issues because we use the block start times to determine which peers can provide the block and relies on the block order to align the peers. When the block ordering is disrupted, it leads to the error case where a block is only selected from one peer, even though multiple peers are eligible to provide this block.

Also added two test cases to catch the error cases. Both cases are failing against HEAD and passing against this PR.

### DIFF
--- a/client/session_fetch_bulk_blocks_test.go
+++ b/client/session_fetch_bulk_blocks_test.go
@@ -1036,13 +1036,174 @@ func TestSelectBlocksForSeriesFromPeerBlocksMetadataPerformsRetries(t *testing.T
 	assert.Equal(t, 1, perPeer[1].blocks[0].reattempt.attempt)
 	assert.Equal(t, []peer{peerB}, perPeer[1].blocks[0].reattempt.attempted)
 
-	// Assert first block second peer
+	// Assert second block second peer
 	assert.Equal(t, start.Add(time.Hour*2), perPeer[1].blocks[1].start)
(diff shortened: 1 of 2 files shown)


client/session_fetch_bulk_blocks_test.go
```diff
@@ -1045,6 +1045,167 @@ func TestSelectBlocksForSeriesFromPeerBlocksMetadataPerformsRetries(t *testing.T
...
 }
 
+func TestSelectBlocksForSeriesFromPeerBlocksMetadataMaintainsBlockOrderAfterPeerExhaustion(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	opts := newSessionTestAdminOptions().
+		SetFetchSeriesBlocksMaxBlockRetries(2)
+	s, err := newSession(opts)
+	assert.NoError(t, err)
```
ben-lerner: probably `require.NoError`

robskillington: Let me update this file and these changes to reflect this.

PR #259, opened 2017-02-13
Peer streaming only fails when unable to read data from any peer during merging reads

xichen2020:
cc @robskillington @ben-lerner @prateek 

This PR fixes an issue during merging reads where if we failed to read from even one peer, the block will be retried across all peers, tripling the load on the cluster. Additionally, it is considered as a failure if we couldn't  read from any of the peers. This is not what we want, in particular if one peer is malfunctioning, we should still consider it a success if we could read from the other peers.

This PR changes it so that if we manage to read data from at least one of the peers, we consider it a success. In addition, if we fail to read data from a peer during a merging read, the attempt will only be retried on that peer and therefore significantly reducing the number of attempts.

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -1655,6 +1655,7 @@ func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
 		// Only select from peers not already attempted
 		currEligible = currStart[:]
 		currID := currStart[0].id
+		currUnselected := currStart[0].unselectedBlocks()[0]
 		for i := len(currEligible) - 1; i >= 0; i-- {
 			unselected := currEligible[i].unselectedBlocks()
 			if unselected[0].reattempt.attempt == 0 {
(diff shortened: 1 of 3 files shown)


client/session.go
```diff
@@ -1731,25 +1727,43 @@ func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
...
 		// fewest attempts and fewest outstanding requests
-		if sameNonNilChecksum {
-			// Order by least attempts then by least outstanding blocks being fetched
-			blocksMetadataQueues = blocksMetadataQueues[:0]
-			for i := range currEligible {
-				insert := blocksMetadataQueue{
-					blocksMetadata: currEligible[i],
-					queue:          peerQueues.findQueue(currEligible[i].peer),
+		if singlePeer || sameNonNilChecksum {
+			// Prepare the reattempt peers metadata
```
robskillington: Perhaps add to this comment, i.e. something like "Prepare the reattempt peers metadata so any possible peer is a candidate to retry from"

xichen2020: Done

PR #262, opened 2017-02-13
Merge latest peer streaming fixes from dev into master

xichen2020:


xichen2020:
Merging because there's no change here, just cherry picking

PR #264, opened 2017-02-14
Flush last writes in the host queue when the queue is closed

xichen2020:

This PR fixes an issue that causes the writes to be lost when a queue is closed. In particular, the current implementation throws away the writes queued if the flush goroutine sees that the queue is closed, which is wrong. This issue was further exacerbated by the fact that during a node removal, multiple versions would be written, which means multiple iterations of tearing down and setting up host queues, which means quite a few writes could be lost without us noticing any errors whatsoever.

This PR forces the writes queued up to be flushed on queue close. Additionally, it also holds the write lock when rotating ops during a flush so we don't accidentally dequeue writes right before a queue is closed.

All in all it's a subtle bug.

### DIFF
--- a/client/host_queue_test.go
+++ b/client/host_queue_test.go
@@ -496,6 +496,85 @@ func TestHostQueueFetchBatchesErrorOnResultError(t *testing.T) {
 	})
 }
 
+func TestHostQueueDrainOnClose(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockConnPool := NewMockconnectionPool(ctrl)
+
+	opts := newHostQueueTestOptions()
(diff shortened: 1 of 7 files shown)


client/host_queue_test.go
```diff
@@ -496,6 +496,85 @@ func TestHostQueueFetchBatchesErrorOnResultError(t *testing.T) {
...
+	go func() {
+		// Wait for all writes
+		wg.Wait()
+
+		close(closeCh)
+	}()
+
+	select {
+	case <-closeCh:
+	case <-time.After(time.Second):
```
robskillington: Maybe make this say 10-60 seconds to account for CI process hanging?  You could also just leave it be and the test will timeout/panic after 10min.

xichen2020: Fair enough, updated

PR #265, opened 2017-02-14
Make blocks to be keyed on UnixNano instead of time.Time

cw9:
To void map miss caused by different locale set in time.Time

@xichen2020 @robskillington

### DIFF
--- a/storage/block/block.go
+++ b/storage/block/block.go
@@ -22,6 +22,7 @@ package block
 
 import (
 	"errors"
+	"math"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -36,7 +37,10 @@ var (
 	errReadFromClosedBlock         = errors.New("attempt to read from a closed block")
 	errRetrievableBlockNoRetriever = errors.New("attempt to read from a retrievable block with no retriever")
 
-	timeZero = time.Time{}
+	timeUnixNanoZero = int64(math.MinInt64)
+	// NB(cw): the behavior of calling UnixNano() on time.Time{} is undefined
+	// using this timeZero makes testing a bit easier.
+	timeZero = time.Unix(0, timeUnixNanoZero)
 )
 
 type dbBlock struct {
@@ -282,16 +286,18 @@ func (b *dbBlock) resetMergeTargetWithLock() {
 
 type databaseSeriesBlocks struct {
 	opts  Options
-	elems map[time.Time]DatabaseBlock
-	min   time.Time
-	max   time.Time
+	elems map[int64]DatabaseBlock
+	min   int64
+	max   int64
 }
 
 // NewDatabaseSeriesBlocks creates a databaseSeriesBlocks instance.
 func NewDatabaseSeriesBlocks(capacity int, opts Options) DatabaseSeriesBlocks {
 	return &databaseSeriesBlocks{
 		opts:  opts,
-		elems: make(map[time.Time]DatabaseBlock, capacity),
+		elems: make(map[int64]DatabaseBlock, capacity),
+		min:   timeUnixNanoZero,
+		max:   timeUnixNanoZero,
 	}
 }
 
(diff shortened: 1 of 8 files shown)


storage/block/block.go
```diff
@@ -36,7 +37,10 @@ var (
 	errReadFromClosedBlock         = errors.New("attempt to read from a closed block")
 	errRetrievableBlockNoRetriever = errors.New("attempt to read from a retrievable block with no retriever")
 
-	timeZero = time.Time{}
+	timeUnixNanoZero = int64(math.MinInt64)
+	// NB(cw): the behavior of calling UnixNano() on time.Time{} is undefined
+	// using this timeZero makes testing a bit easier.
+	timeZero = time.Unix(0, timeUnixNanoZero)
```
xichen2020: Why not just `time.Unix(0, 0)`

storage/block/block.go
```diff
@@ -326,40 +332,45 @@ func (dbb *databaseSeriesBlocks) AddSeries(other DatabaseSeriesBlocks) {
...
+	return time.Unix(0, dbb.max)
 }
 
 func (dbb *databaseSeriesBlocks) BlockAt(t time.Time) (DatabaseBlock, bool) {
-	b, ok := dbb.elems[t]
+	b, ok := dbb.elems[t.UnixNano()]
 	return b, ok
 }
 
 func (dbb *databaseSeriesBlocks) AllBlocks() map[time.Time]DatabaseBlock {
```
xichen2020: hm..where do we use this function?

storage/block/block.go
robskillington: Or even just `timeZero = int64(0)`

storage/block/block.go
```diff
@@ -326,40 +332,45 @@ func (dbb *databaseSeriesBlocks) AddSeries(other DatabaseSeriesBlocks) {
...
+	res := make(map[time.Time]DatabaseBlock, len(dbb.elems))
+	for nano, db := range dbb.elems {
+		res[time.Unix(0, nano)] = db
+	}
+	return res
 }
 
 func (dbb *databaseSeriesBlocks) RemoveBlockAt(t time.Time) {
-	if _, exists := dbb.elems[t]; !exists {
+	tNano := t.UnixNano()
```
robskillington: Maybe a better name is `unixNano := t.UnixNano()`?

storage/block/block.go
```diff
@@ -326,40 +332,45 @@ func (dbb *databaseSeriesBlocks) AddSeries(other DatabaseSeriesBlocks) {
...
 
 func (dbb *databaseSeriesBlocks) BlockAt(t time.Time) (DatabaseBlock, bool) {
-	b, ok := dbb.elems[t]
+	b, ok := dbb.elems[t.UnixNano()]
 	return b, ok
 }
 
 func (dbb *databaseSeriesBlocks) AllBlocks() map[time.Time]DatabaseBlock {
-	return dbb.elems
+	res := make(map[time.Time]DatabaseBlock, len(dbb.elems))
```
robskillington: Yeah we can't create a new map each time, this is called on some huge maps. A copy each time is extremely expensive (think millions of IDs all getting `AllBlocks(...)` called on it.

robskillington:
Need to also update the following places that `map[time.Time]...` is used:

```
persist/fs/seek_manager.go (2 definitions)
storage/bootstrap/bootstrapper/commitlog/source.go (1 definition)
storage/repair.go (1 definition)
storage/repair/metadata.go (1 definition)
storage/repair/types.go (1 definition)
storage/shard.go (1 definition)
```

storage/block/block.go
cw9: I kinda want to make timeUnixNanoZero and timeZero interchangeable, gonna do:
	timeZero         = time.Unix(0, 0)
	timeUnixNanoZero = timeZero.UnixNano()

storage/block/block.go
cw9: sure

storage/block/block.go
cw9: Sure, then I'll change the interface to return map[int64]DatabaseBlock

cw9:
Yeah, I'm thinking to keep the blast radiance down per diff, I'm gonna fix all the usage of this interface in this one and handle those other places in separate diffs.

cw9:
Talked with Rob, will postpone this pr and do a full sweep of time.Time in maps across the code base

PR #267, opened 2017-02-15
Do not bootstrap until in-flight file operations have finished

xichen2020:

This PR adds logic to pause bootstrap attempts until in-flight file operations have finished to minimize the impact of filesystem operations on node bootstrapping performance, which trades off a little bootstrap time for reduced CPU load and higher write success rate during bootstrapping due to cluster topology changes


storage/bootstrap.go
```diff
@@ -193,7 +206,7 @@ func (m *bootstrapManager) bootstrap() error {
 	// all the data we bootstrapped.
 	rateLimitOpts := m.fsManager.RateLimitOptions()
 	m.fsManager.SetRateLimitOptions(rateLimitOpts.SetLimitEnabled(false))
-	m.fsManager.Run(m.nowFn(), false)
+	m.fsManager.Run(m.nowFn(), false, true)
```
prateek: Can this be done without needing`force`? This is the only place that uses it afaict, and fsManager will be stopped by virtue of the loop above.

xichen2020: When we get here, the file system manager is disabled, but we do want to flush the data out to disk, which is why the force flag is in place.

prateek: Makes sense. Took me a second to get that we need it to indicate that we should flush despite the db not being bootstrapped.

PR #270, opened 2017-02-16
Add database mediator to mediate interactions among various database job managers

### LINKED ISSUE #269: Add back the ability to blocking flush after a bootstrap
The following PR works around an issue we encountered when naively trying to flush after a bootstrap:
https://github.com/m3db/m3db/pull/268

It does this by simply waiting for the next tick and a flush to occur to actually finish the bootstrap process.  

Ideally we do actually flush directly after a bootstrap, however we need to coordinate with the ticking procedure to ensure that all buffers have been rotated into each series by waiting for a tick to rotate all the buffers before we can start a flush cycle (otherwise me miss blocks that are just about to rotate in for that time window when we flush and we don't flush them again because the time window is marked flushed).

xichen2020:

This PR introduces the database mediator to mediate interactions among various database job managers including the bootstrap manager, filesystem manager, tick manager, and the repairer. This way the interactions are centralized in one file, the mediator interface is fairly lean, and the database can delegate all job activities to the mediator without worrying about which managers it needs to use, making the code easier to read.

By virtue of centralizing the interaction logic:
* The constraint that a tick must happen before a flush to properly drain database buffers is now easily satisfied by calling the unexported `tickWithFileOp` method in `mediator.go`
* The constraint that in-flight file ops must finish before starting a bootstrap is now easily satisified by calling the `Bootstrap` method in `mediator.go`

Because we prefer flushing/cleaning up filesets immediately following a bootstrap, it is required that we have the ability to cancel ongoing ticks. This is implemented using `Cancellable` interface in the `context` package and the main logic is in the `tick.go`.

Fixes #269

### DIFF
--- a/context/cancel.go
+++ b/context/cancel.go
@@ -23,17 +23,17 @@ package context
 import "sync/atomic"
 
 type cancellable struct {
-	cancelled *int32
+	cancelled int32
 }
 
 // NewCancellable creates a new cancellable object
 func NewCancellable() Cancellable {
-	return cancellable{cancelled: new(int32)}
+	return &cancellable{}
 }
 
-func (c cancellable) IsCancelled() bool { return atomic.LoadInt32(c.cancelled) == 1 }
-func (c cancellable) Cancel()           { atomic.StoreInt32(c.cancelled, 1) }
-func (c cancellable) Reset()            { atomic.StoreInt32(c.cancelled, 0) }
+func (c *cancellable) IsCancelled() bool { return atomic.LoadInt32(&c.cancelled) == 1 }
+func (c *cancellable) Cancel()           { atomic.StoreInt32(&c.cancelled, 1) }
+func (c *cancellable) Reset()            { atomic.StoreInt32(&c.cancelled, 0) }
 
 // NoOpCancellable is a no-op cancellable
 var noOpCancellable Cancellable = cancellableNoOp{}
--- a/context/cancel.go
+++ b/context/cancel.go
@@ -1,0 +1,45 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package context
+
+import "sync/atomic"
+
+type cancellable struct {
+	cancelled *int32
+}
+
+// NewCancellable creates a new cancellable object
+func NewCancellable() Cancellable {
+	return cancellable{cancelled: new(int32)}
+}
+
+func (c cancellable) IsCancelled() bool { return atomic.LoadInt32(c.cancelled) == 1 }
+func (c cancellable) Cancel()           { atomic.StoreInt32(c.cancelled, 1) }
+func (c cancellable) Reset()            { atomic.StoreInt32(c.cancelled, 0) }
+
+// NoOpCancellable is a no-op cancellable
+var NoOpCancellable Cancellable = noOpCancellable{}
+
+type noOpCancellable struct{}
+
+func (c noOpCancellable) IsCancelled() bool { return false }
+func (c noOpCancellable) Cancel()           {}
+func (c noOpCancellable) Reset()            {}
--- a/context/cancel.go
+++ b/context/cancel.go
@@ -36,10 +36,12 @@ func (c cancellable) Cancel()           { atomic.StoreInt32(c.cancelled, 1) }
 func (c cancellable) Reset()            { atomic.StoreInt32(c.cancelled, 0) }
 
 // NoOpCancellable is a no-op cancellable
-var NoOpCancellable Cancellable = noOpCancellable{}
+var noOpCancellable Cancellable = cancellableNoOp{}
 
-type noOpCancellable struct{}
+type cancellableNoOp struct{}
 
-func (c noOpCancellable) IsCancelled() bool { return false }
-func (c noOpCancellable) Cancel()           {}
-func (c noOpCancellable) Reset()            {}
+// NewNoOpCanncellable returns a no-op cancellable
+func NewNoOpCanncellable() Cancellable      { return noOpCancellable }
+func (c cancellableNoOp) IsCancelled() bool { return false }
+func (c cancellableNoOp) Cancel()           {}
+func (c cancellableNoOp) Reset()            {}
--- a/storage/bootstrap.go
+++ b/storage/bootstrap.go
@@ -66,6 +66,7 @@ type bootstrapManager struct {
 	sync.RWMutex
 
 	database       database
+	mediator       databaseMediator
 	opts           Options
 	log            xlog.Logger
 	nowFn          clock.NowFn
@@ -75,10 +76,14 @@ type bootstrapManager struct {
 	sleepFn        sleepFn
 }
 
-func newBootstrapManager(database database) databaseBootstrapManager {
+func newBootstrapManager(
+	database database,
+	mediator databaseMediator,
+) databaseBootstrapManager {
 	opts := database.Options()
 	return &bootstrapManager{
 		database:       database,
+		mediator:       mediator,
 		opts:           opts,
 		log:            opts.InstrumentOptions().Logger(),
 		nowFn:          opts.ClockOptions().NowFn(),
@@ -139,6 +144,11 @@ func (m *bootstrapManager) Bootstrap() error {
 	}
 	m.Unlock()
 
+	// NB(xichen): disable filesystem manager before we bootstrap to minimize
+	// the impact of file operations on bootstrapping performance
+	m.mediator.DisableFileOps()
+	defer m.mediator.EnableFileOps()
+
 	// Keep performing bootstraps until none pending
 	multiErr := xerrors.NewMultiError()
 	for {
@@ -161,6 +171,9 @@ func (m *bootstrapManager) Bootstrap() error {
 			break
 		}
 	}
+
+	// Forcing a tick to perform necessary file operations
+	m.mediator.Tick(0, false, true)
 
 	return multiErr.FinalError()
 }
--- a/storage/bootstrap.go
+++ b/storage/bootstrap.go
@@ -171,7 +171,7 @@ func (m *bootstrapManager) Bootstrap() error {
 	}
 
 	// Forcing a tick to perform necessary file operations
-	m.mediator.Tick(0, syncRun, force)
+	m.mediator.Tick(m.opts.RetentionOptions().BufferDrain(), syncRun, force)
 
 	return multiErr.FinalError()
 }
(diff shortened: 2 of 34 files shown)


context/cancel.go
```diff
@@ -0,0 +1,45 @@
...
+func NewCancellable() Cancellable {
+	return cancellable{cancelled: new(int32)}
+}
+
+func (c cancellable) IsCancelled() bool { return atomic.LoadInt32(c.cancelled) == 1 }
+func (c cancellable) Cancel()           { atomic.StoreInt32(c.cancelled, 1) }
+func (c cancellable) Reset()            { atomic.StoreInt32(c.cancelled, 0) }
+
+// NoOpCancellable is a no-op cancellable
+var NoOpCancellable Cancellable = noOpCancellable{}
```
robskillington: Is it worth making this a method `NewNoOpCancellable()` that returns a ref to a single prealloc'd instance so that no one can mutate the var.  e.g. as is one could set `context.NoOpCancellable = nil` if they so please.

storage/bootstrap.go
```diff
@@ -72,18 +72,18 @@ type bootstrapManager struct {
 	newBootstrapFn NewBootstrapFn
 	state          bootstrapState
 	hasPending     bool
+	sleepFn        sleepFn
```
robskillington: Is this used anywhere?  I can't see atm that it is?

storage/fs.go
```diff
@@ -85,27 +86,36 @@ func newFileSystemManager(
...
+		enabled:  true,
 	}, nil
 }
 
-func (m *fileSystemManager) ShouldRun(t time.Time) bool {
-	// If we haven't bootstrapped yet, no actions necessary.
-	if !m.database.IsBootstrapped() {
-		return false
-	}
+func (m *fileSystemManager) Disable() bool {
```
robskillington: Might it be worth returning just the status here instead of a boolean?  The status as it's an enum-ish seems more descriptive.  You could also add it to the return type of `Enable` too.

storage/fs.go
```diff
@@ -85,27 +86,36 @@ func newFileSystemManager(
...
-	return m.status != fileOpInProgress
+func (m *fileSystemManager) IsRunning() bool {
+	m.RLock()
+	running := m.status == fileOpInProgress
+	m.RUnlock()
+	return running
 }
 
-func (m *fileSystemManager) Run(t time.Time, async bool) {
+func (m *fileSystemManager) Run(t time.Time, async bool, force bool) bool {
```
robskillington: Can we perhaps make these booleans enums so it's clear what the params are when we use them from calling code? i.e. something like:
```go
type runType int
const (
  syncRun runType = iota
  asyncRun
)
type forceType int
const (
  noForce forceType = iota
  force 
)

// calling
fsm.Run(t, syncRun, noForce)
```

storage/mediator.go
```diff
@@ -0,0 +1,230 @@
...
+	var repairer databaseRepairer
+	if opts.RepairEnabled() {
+		var err error
+		repairer, err = newDatabaseRepairer(database)
+		if err != nil {
+			return nil, err
+		}
+	} else {
+		repairer = newNoopDatabaseRepairer()
+	}
```
robskillington: To simplify you could:

```go
repairer := newNoopDatabaseRepairer()
if opts.RepairEnabled() { 
  // override with a real repairer
}
```

storage/mediator.go
```diff
@@ -0,0 +1,230 @@
...
+	// Wait for in-flight file operations to finish before we proceed
+	// with bootstrapping
+	for fileOpInProgess {
+		m.sleepFn(fileOpCheckInterval)
+		fileOpInProgess = m.databaseFileSystemManager.IsRunning()
+	}
+
+	// Perform file operations after bootstrapping
+	err := m.databaseBootstrapManager.Bootstrap()
+	m.tickWithFileOp(0, false, true)
```
robskillington: If these were enums params (the bools) would help with the readability here.

storage/mediator.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func (m *mediator) Close() error {
+	m.Lock()
+	defer m.Unlock()
+	if m.state == mediatorNotOpen {
+		return errMediatorNotOpen
+	}
+	if m.state == mediatorClosed {
+		return errMediatorAlreadyClosed
```
robskillington: If you wanted to save some paths you need to cover for code coverage you could combine these errors if you wanted, e.g. errMediatorCannotCloseNotOpen

storage/mediator.go
robskillington: The soft deadline of zero will definitely make this a very intensive tick.  It may be worth us thinking about reducing the parallelism if we want to tick "as fast as we possibly can" in the future.  I'm happy to leave and experiment as is of course but just want to be wary of this definitely being a cause for exerting higher pressure if this happens around the same time on all the nodes during say a node removal.

storage/mediator.go
```diff
@@ -0,0 +1,230 @@
...
+			} else {
+				m.metrics.cleanupStatus.Update(0)
+			}
+			if m.databaseFileSystemManager.IsFlushing() {
+				m.metrics.flushStatus.Update(1)
+			} else {
+				m.metrics.flushStatus.Update(0)
+			}
+		case <-m.closedCh:
+			return
```
robskillington: May want to `t.Stop()` before the return too.

storage/mediator.go
```diff
@@ -0,0 +1,230 @@
...
+	"github.com/m3db/m3db/clock"
+
+	"github.com/uber-go/tally"
+)
+
+type mediatorState int
+
+const (
+	fileOpCheckInterval = time.Second
+	numOngoingTasks     = 2
```
robskillington: Don't think you use this anywhere anymore since you just directly close the channel.

storage/shard.go
```diff
@@ -330,6 +331,9 @@ func (s *dbShard) tickAndExpire(
 	}
 	start := s.nowFn()
 	s.forEachShardEntry(func(entry *dbShardEntry) bool {
+		if c.IsCancelled() {
+			return false
+		}
```
robskillington: Perhaps move this into the `tickSleepIfAheadEvery` inner loop, so we only pause to check every 128 (which is still pretty frequent)?

storage/tick.go
```diff
@@ -0,0 +1,156 @@
...
+	tickDeadlineMissed tally.Counter
+	tickDeadlineMet    tally.Counter
+}
+
+func newTickManagerMetrics(scope tally.Scope) tickManagerMetrics {
+	return tickManagerMetrics{
+		tickDuration:       scope.Timer("tick.duration"),
+		tickCancelled:      scope.Counter("tick.cancelled"),
+		tickDeadlineMissed: scope.Counter("tick.deadline.missed"),
+		tickDeadlineMet:    scope.Counter("tick.deadline.met"),
```
robskillington: Maybe we should just make the scope have a `tick` subscope instead of prefixing now?

storage/tick.go
```diff
@@ -0,0 +1,156 @@
...
+				acquired = true
+				break
+			case <-tick.C:
+				mgr.c.Cancel()
+			}
+			if acquired {
+				break
+			}
+		}
+	}
```
robskillington: Why not just use a blocking wait if it's not `force`? i.e.

```go
if force {
  mgr.tokenCh <- struct{}{}
} else {
  select {
    case mgr.tokenCh <- struct{}{}:
    default:
      return false
  }
}
```

robskillington: Also, perhaps starting with the token enqueued from `newTickManager` then reading it out seems more idiomatic?

storage/tick.go
```diff
@@ -0,0 +1,156 @@
...
+	end := mgr.nowFn()
+	duration := end.Sub(start)
+	mgr.metrics.tickDuration.Record(duration)
+
+	if duration > softDeadline {
+		mgr.metrics.tickDeadlineMissed.Inc(1)
+	} else {
+		mgr.metrics.tickDeadlineMet.Inc(1)
+		if mgr.c.IsCancelled() {
+			mgr.metrics.tickCancelled.Inc(1)
```
robskillington: Shouldn't this metric be checked regardless of which if we are in? 

Maybe it's best to just let the namespace return early if its cancelled, don't check it until the very end of this function so all code paths for the metric only flow through one way.  There might need an extra `if` to avoid the sleep if cancelled but that shouldn't be so bad?

storage/types.go
```diff
@@ -234,7 +234,7 @@ type databaseShard interface {
 	Close() error
 
 	// Tick performs any updates to ensure series drain their buffers and blocks are flushed, etc
-	Tick(softDeadline time.Duration) tickResult
+	Tick(softDeadline time.Duration, c context.Cancellable) tickResult
```
robskillington: Perhaps the cancellable should always come first (maybe second only to a context.Context) if we want to follow similar conventions?  Go linter now makes you always put the context as a first param of a method, so it seems kind of natural we would put a cancellable as first, and second only if there was already a context as the first arg.   Thoughts?

context/cancel.go
```diff
@@ -0,0 +1,45 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package context
+
+import "sync/atomic"
+
+type cancellable struct {
+	cancelled *int32
```
xichen2020: In general yes, however there is an optimization made in the go compiler that says if the receiver type fits in a machine word, then it will stored directly in the `data` field: https://research.swtch.com/interfaces

If we use *int32 as the field and use `cancellable` as the receiver type, the size of the struct value is 8 bytes, which fits in a single machine word and therefore directly stored in the `data` field. The value contains an *int32 that points to a 4-byte memory chunk on the heap.

If we use int32 as the field and use `*cancellable` as the receiver type, the size of the struct pointer is 8 bytes, which also fits in a single machine word and stored in the `data` field. The struct pointer then points to a struct allocated on the heap containing an `int32` object.

Therefore, the two approaches are essentially equivalent.

context/cancel.go
xichen2020: Sure thing

storage/bootstrap.go
xichen2020: Good catch, will remove

storage/fs.go
xichen2020: Sure thing

storage/fs.go
xichen2020: Sounds good

storage/mediator.go
xichen2020: I didn't write this code originally, but sure can change

storage/mediator.go
xichen2020: Will do

xichen2020: Let's experiement with it, if the load doesn't become too big a problem, I'd like to tick as fast as possible to avoid delaying bootstrapping

storage/mediator.go
xichen2020: meh I'm fine w/ this

storage/mediator.go
xichen2020: Sure thing

storage/shard.go
xichen2020: This is already the inner loop, the function passed to `forEachShardEntry` is applied to each batch instead of each entry, which means we are only checking for cancellation every 128 entries.

storage/tick.go
xichen2020: Sure thing

storage/tick.go
xichen2020: Not sure if I understood what you meant, the code you gave does the same thing as mine if `force` is false, and uses a blocking wait if `force` is true. However, when performing a tick during bootstrapping (i.e., when `force` is true), we'd need to cancel the ongoing tick if any instead of waiting for the current tick to finish to avoid incurring a full ticking delay (currently 10 minutes) for bootstrapping, and I don't see how a blocking wait would support cancellation?

storage/mediator.go
xichen2020: Yeah I removed it in a later commit, ta

storage/tick.go
xichen2020: Yeah that's what I did in a later commit, I also made the sleep more fine-grained by waking up periodically and checking for cancellation to make it more responsive in case a cancellation is requested while the current tick is sleeping

storage/types.go
xichen2020: Sure thing

storage/mediator.go
robskillington: Sure thing.

storage/tick.go
```diff
@@ -0,0 +1,161 @@
...
+	} else {
+		mgr.metrics.tickDeadlineMet.Inc(1)
+
+		// Throttle to reduce locking overhead during ticking
+		for d := time.Duration(0); d < softDeadline-duration; d += cancellationCheckInterval {
+			if mgr.c.IsCancelled() {
+				break
+			}
+			mgr.sleepFn(cancellationCheckInterval)
+		}
```
robskillington: Maybe just remove this extra sleep cycle.  It's going to be such a micro sleep as we would only sleep here if it is the last batch of series being slight faster than it should be (since we check and sleep every batch of series if we are ahead).  i.e. if we have 6MM series and we do batches of 128 and we do a 10min tick, then the most we would be ahead of the tick here would be `= 10mins / (6MM/128) = 12.8 milliseconds`.

PR #271, opened 2017-02-18
Add incremental bootstrapping for the peers bootstrapper

robskillington:
This change makes the first time range that's bootstrapped an incremental bootstrap which means if raw data is being read for blocks it is incrementally flushed and replaced with retrievable blocks in place of the full raw block.

The peers bootstrapper uses these semantics to flush raw block data as it finishes each shard.  This avoids having to hold the entire set of data in memory when doing a full peer stream.

### DIFF
--- a/storage/bootstrap/bootstrapper/commitlog/source.go
+++ b/storage/bootstrap/bootstrapper/commitlog/source.go
@@ -81,7 +81,7 @@ func (s *commitLogSource) Available(
 func (s *commitLogSource) Read(
 	namespace ts.ID,
 	shardsTimeRanges result.ShardTimeRanges,
-	opts bootstrap.RunOptions,
+	_ bootstrap.RunOptions,
 ) (result.BootstrapResult, error) {
 	if shardsTimeRanges.IsEmpty() {
 		return nil, nil
--- a/storage/bootstrap/bootstrapper/commitlog/source.go
+++ b/storage/bootstrap/bootstrapper/commitlog/source.go
@@ -81,6 +81,7 @@ func (s *commitLogSource) Available(
 func (s *commitLogSource) Read(
 	namespace ts.ID,
 	shardsTimeRanges result.ShardTimeRanges,
+	opts bootstrap.RunOptions,
 ) (result.BootstrapResult, error) {
 	if shardsTimeRanges.IsEmpty() {
 		return nil, nil
--- a/storage/bootstrap/bootstrapper/fs/source.go
+++ b/storage/bootstrap/bootstrapper/fs/source.go
@@ -354,7 +354,7 @@ func (s *fileSystemSource) bootstrapFromReaders(
 func (s *fileSystemSource) Read(
 	namespace ts.ID,
 	shardsTimeRanges result.ShardTimeRanges,
-	opts bootstrap.RunOptions,
+	_ bootstrap.RunOptions,
 ) (result.BootstrapResult, error) {
 	if shardsTimeRanges.IsEmpty() {
 		return nil, nil
--- a/storage/bootstrap/bootstrapper/fs/source.go
+++ b/storage/bootstrap/bootstrapper/fs/source.go
@@ -354,6 +354,7 @@ func (s *fileSystemSource) bootstrapFromReaders(
 func (s *fileSystemSource) Read(
 	namespace ts.ID,
 	shardsTimeRanges result.ShardTimeRanges,
+	opts bootstrap.RunOptions,
 ) (result.BootstrapResult, error) {
 	if shardsTimeRanges.IsEmpty() {
 		return nil, nil
--- a/storage/bootstrap/bootstrapper/peers/source.go
+++ b/storage/bootstrap/bootstrapper/peers/source.go
@@ -41,6 +41,20 @@ type peersSource struct {
 	nowFn clock.NowFn
 }
 
+type incrementalFlush struct {
+	namespace         ts.ID
+	persistManager    persist.Manager
+	shard             uint32
+	shardRetrieverMgr block.DatabaseShardBlockRetrieverManager
+	shardResult       result.ShardResult
+	timeRange         xtime.Range
+}
+
+type incrementalFlushedBlock struct {
+	id    ts.ID
+	block block.DatabaseBlock
+}
+
 func newPeersSource(opts Options) bootstrap.Source {
 	return &peersSource{
 		opts:  opts,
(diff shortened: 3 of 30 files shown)


storage/bootstrap/bootstrapper/peers/types.go
```diff
@@ -39,9 +41,47 @@ type Options interface {
...
+
+	// SetIncrementalBootstrapPersistMaxQueue sets the max queue for
+	// bootstrapping shards waiting in line to persist without blocking
+	// the concurrent shard fetchers.
+	SetIncrementalBootstrapPersistMaxQueue(value int) Options
+
+	// IncrementalBootstrapPersistMaxQueue returns the max queue for
+	// bootstrapping shards waiting in line to persist without blocking
+	// the concurrent shard fetchers.
+	IncrementalBootstrapPersistMaxQueue() int
```
xichen2020: Maybe call it `IncrementalPersistMaxQueueSize` to be more clear?

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -117,5 +199,103 @@ func (s *peersSource) Read(
 
 	wg.Wait()
 
+	close(incrementalQueue)
+	incrementalWg.Wait()
+
+	if incremental {
+		// Now cache the incremental results
+		shards := make([]uint32, 0, len(shardsTimeRanges))
+		for shard := range shardsTimeRanges {
```
xichen2020: Should we look at the (un)fulfilled time ranges in `result` to figure out which shards have we successfully bootstrapped instead?

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -117,5 +199,103 @@ func (s *peersSource) Read(
...
+			tmpCtx.Reset()
+			stream, err := bl.Stream(tmpCtx)
+			if err != nil {
+				blockErr = err // Need to call prepared.Close, avoid return
+				break
+			}
+
+			segment, err := stream.Segment()
+			if err != nil {
+				blockErr = err // Need to call prepared.Close, avoid return
```
xichen2020: Also need to call `tmpCtx.BlockingClose` here?

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -117,5 +199,103 @@ func (s *peersSource) Read(
...
+	namespace ts.ID,
+	persistManager persist.Manager,
+	shard uint32,
+	shardRetrieverMgr block.DatabaseShardBlockRetrieverManager,
+	shardResult result.ShardResult,
+	tr xtime.Range,
+) error {
+	var (
+		numSeries      = len(shardResult.AllSeries())
+		flushedBlocks  = make([]incrementalFlushedBlock, 0, numSeries)
```
xichen2020: This is a pretty big allocation per call, can we pass it in and reuse it across multiple calls of `incrementalFlush` instead to reduce memory allocation and GC?

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -117,5 +199,103 @@ func (s *peersSource) Read(
...
+		if blockErr != nil {
+			// A block error is worth bubbling up more than a close error
+			err = blockErr
+		}
+		if err != nil {
+			return err
+		}
+
+		// We can now make flushed blocks retrievable
+		for _, entry := range flushedBlocks {
```
xichen2020: I wonder instead of accumulating all these flushed blocks and holding onto them until the flush succeeds, would it better to just reiterate over the result again so we don't need this `flushedBlocks` slice at all at the expense of slightly more CPU cycles?

storage/bootstrap/bootstrapper/peers/types.go
```diff
@@ -39,9 +41,47 @@ type Options interface {
...
-	SetBootstrapShardConcurrency(value int) Options
+	// SetDefaultBootstrapShardConcurrency sets the concurrency for
+	// bootstrapping shards when performing a non-incremental bootstrap.
+	SetDefaultBootstrapShardConcurrency(value int) Options
 
-	// BootstrapShardConcurrency returns the concurrency for bootstrapping shards
-	BootstrapShardConcurrency() int
+	// DefaultBootstrapShardConcurrency returns the concurrency for
+	// bootstrapping shards when performing a non-incremental bootstrap.
+	DefaultBootstrapShardConcurrency() int
```
robskillington: Sure thing.

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -81,6 +81,7 @@ func (s *commitLogSource) Available(
 func (s *commitLogSource) Read(
 	namespace ts.ID,
 	shardsTimeRanges result.ShardTimeRanges,
+	opts bootstrap.RunOptions,
```
robskillington: Sure thing.

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -354,6 +354,7 @@ func (s *fileSystemSource) bootstrapFromReaders(
 func (s *fileSystemSource) Read(
 	namespace ts.ID,
 	shardsTimeRanges result.ShardTimeRanges,
+	opts bootstrap.RunOptions,
```
robskillington: Sure thing.

storage/bootstrap/bootstrapper/peers/types.go
robskillington: Sure thing.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: It doesn't hurt to try and cache the shard indices, it simply scans the available filesets.  At most this would save a few milliseconds of disk seeks.  It doesn't return an error if the filesets are not there as it is best effort.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: I thought about this, I'll need to add a lock for single access but can do.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: Good catch, cheers.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: Yeah that's a good call, then can avoid needing a lock for that large array too.  Will take this approach.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: I'll actually just loop again based on the below discussion

PR #273, opened 2017-02-19
Add an inbuilt facility to perform write and fetch retries

robskillington:
This change adds the ability to set a retrier for writes and fetches.  It allows for anyone to use an out of the box retrier from the `github.com/m3db/m3x/retry` package.  It also let's them create their own custom retrier that fits the `Retrier` interface and specify that instead.

### DIFF
--- a/client/fetch_attempt.go
+++ b/client/fetch_attempt.go
@@ -1,0 +1,95 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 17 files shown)


client/fetch_attempt.go
```diff
@@ -0,0 +1,95 @@
...
+func (p *fetchAttemptPool) Init() {
+	p.pool.Init(func() interface{} {
+		w := &fetchAttempt{session: p.session}
+		// NB(r): Bind attemptFn once to avoid creating receiver
+		// and function method pointer over and over again
+		w.attemptFn = w.perform
+		w.reset()
+		return w
+	})
+	p.initialized = true
```
xichen2020: This `initialized` flag seems to be only assigned to but never used?

client/options.go
```diff
@@ -74,7 +75,7 @@ const (
 	defaultIdentifierPoolSize = 8192
 
 	// defaultWriteOpPoolSize is the default write op pool size
-	defaultWriteOpPoolSize = 1000000
+	defaultWriteOpPoolSize = 262144
```
xichen2020: why reducing the size to 256K out of curiosity?

client/fetch_attempt.go
robskillington: I want to use this later to do deferred pool initialization, however the way the code is structured makes that difficult.  I can do this in a follow up PR that restructures a lot of the session.

client/options.go
robskillington: After looking at how many writes are open at any given time (i.e. num goroutines) we seem to peak at 60-80k, the original number was arbitrarily picked.

client/fetch_attempt.go
xichen2020: Yeah in that case would you mind removing the flag now and add it back in the followup PR then?

client/options.go
xichen2020: Makes sense

PR #274, opened 2017-02-19
More accurate filesystem rate limiting

xichen2020:

This PR achieves more accurate rate limiting by checking the rate limit every 128 writes as opposed to every 100 ms and throttle as needed. Since the number of series per shard is a lot higher than 128, checking the limit every 128 writes means throttling will be applied more frequently and accurately.


persist/fs/manager.go
```diff
@@ -82,21 +82,21 @@ func (pm *persistManager) persist(
 	rateLimitMbps := pm.rateLimitOpts.LimitMbps()
 	if pm.rateLimitOpts.LimitEnabled() && rateLimitMbps > 0.0 {
 		now := pm.nowFn()
-		if pm.lastCheck.IsZero() {
+		if pm.start.IsZero() {
 			pm.start = now
-			pm.lastCheck = now
-		} else if now.Sub(pm.lastCheck) >= pm.rateLimitOpts.LimitCheckInterval() {
-			pm.lastCheck = now
+		} else if pm.count == pm.rateLimitOpts.LimitCheckEvery() {
```
robskillington: Is it safer perhaps to `pm.count %  pm.rateLimitOpts.LimitCheckEvery() == 0` and only reset to zero on `reset()` call?

persist/fs/manager.go
```diff
@@ -82,21 +82,21 @@ func (pm *persistManager) persist(
 	rateLimitMbps := pm.rateLimitOpts.LimitMbps()
 	if pm.rateLimitOpts.LimitEnabled() && rateLimitMbps > 0.0 {
```
robskillington: Is it worth caching the limitmbps, limitenabled and limitcheckevery since we perform this once per series, per time window? i.e. MMs * 24

xichen2020: Meh these are literally one alone getters so the compiler should already inline these anyway

persist/fs/manager.go
xichen2020: Hmm It's a lot cheaper to check against 0 than doing a mod, though. I could perhaps use >= if you feel strongly

robskillington: If you wouldn't mind?  Seems more resilient.

xichen2020: Sure thing

PR #275, opened 2017-02-19
Open only the delta set of connections on topology change

robskillington:
This change only connects to new sets of hosts on a topology change and only closes the host queues belonging to hosts no longer in the topology.

This will reduce the large double connection count required for a topology change.  Since the topology can change a lot during shards becoming available one by one this should reduce a lot of overhead.


robskillington:
Yeah let me get one in for only opening new connections and one for only closing old connections.

PR #276, opened 2017-02-19
Disable ticking and flushing after bootstrapping

xichen2020:

This PR disables ticking and flushing after bootstrapping because we observed that it added large load on the cluster, causing commit log queues to back up. Instead we'll let the buffer draining and data flushing happen naturally to spread out the load across the cluster. This has shown to reduce the commit log queue spikes significantly.

### DIFF
--- a/storage/bootstrap.go
+++ b/storage/bootstrap.go
@@ -182,9 +182,6 @@ func (m *bootstrapManager) Bootstrap() error {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


robskillington:
Can you leave a comment before merging where you are taking it away from saying something like "a tick and a consequent flush will need to occur until this data is durable on disk, this was too intensive to call directly after a bootstrap however"?

PR #277, opened 2017-02-20
Report timing breakdowns during persistence

xichen2020:

This PR adds logic to report how much time is spent on doing actual work vs. sleeping during flushing so we have better visibility into what's happening during filesystem persistence.

### DIFF
--- a/persist/fs/persist_manager.go
+++ b/persist/fs/persist_manager.go
@@ -112,7 +112,7 @@ func NewPersistManager(opts Options) persist.Manager {
 	newFileMode := opts.NewFileMode()
 	newDirectoryMode := opts.NewDirectoryMode()
 	writer := NewWriter(blockSize, filePathPrefix, writerBufferSize, newFileMode, newDirectoryMode)
-	pm := &persistManager{
+	return &persistManager{
 		opts:           opts,
 		scope:          scope,
 		filePathPrefix: filePathPrefix,
@@ -123,9 +123,6 @@ func NewPersistManager(opts Options) persist.Manager {
 		metrics:        make(map[uint32]*shardMetrics),
 		segmentHolder:  make([]checked.Bytes, 2),
 	}
-
-	go pm.reportLoop()
-	return pm
 }
 
 func (pm *persistManager) persist(
@@ -164,21 +161,22 @@ func (pm *persistManager) persist(
 	end := pm.nowFn()
 	worked := end.Sub(start) - slept
 	currMetrics.recordWorked(worked)
-	currMetrics.recordSlept(slept)
+	if slept > 0 {
+		currMetrics.recordSlept(slept)
+	}
 
 	return err
 }
 
 func (pm *persistManager) close() error {
-	err := pm.writer.Close()
-	pm.reset()
-	return err
+	return pm.writer.Close()
 }
 
 func (pm *persistManager) reset() {
 	pm.start = timeZero
 	pm.count = 0
 	pm.bytesWritten = 0
+	pm.currMetrics.reset()
 }
 
 func (pm *persistManager) Prepare(namespace ts.ID, shard uint32, blockStart time.Time) (persist.PreparedPersist, error) {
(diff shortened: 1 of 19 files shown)


persist/fs/persist_manager.go
```diff
@@ -57,37 +105,51 @@ type persistManager struct {
...
 		filePathPrefix: filePathPrefix,
 		rateLimitOpts:  opts.RateLimitOptions(),
 		nowFn:          opts.ClockOptions().NowFn(),
 		sleepFn:        time.Sleep,
 		writer:         writer,
+		metrics:        make(map[uint32]*shardMetrics),
 		segmentHolder:  make([]checked.Bytes, 2),
 	}
+
+	go pm.reportLoop()
```
robskillington: Perhaps some way to stop the report loop would be good?

persist/fs/persist_manager.go
```diff
@@ -115,20 +182,40 @@ func (pm *persistManager) reset() {
...
+		end := pm.nowFn()
+		currMetrics.recordWorked(end.Sub(start))
 		return prepared, err
 	}
 
 	prepared.Persist = pm.persist
 	prepared.Close = pm.close
+
+	end := pm.nowFn()
+	currMetrics.recordWorked(end.Sub(start))
```
robskillington: Is it worth reporting any stats for checking exists and opening?  Perhaps we just want to look at the actual file writing/throughput?

persist/fs/persist_manager.go
```diff
@@ -99,6 +161,11 @@ func (pm *persistManager) persist(
 	pm.count++
 	pm.bytesWritten += int64(segment.Len())
 
+	end := pm.nowFn()
+	worked := end.Sub(start) - slept
+	currMetrics.recordWorked(worked)
+	currMetrics.recordSlept(slept)
```
robskillington: Maybe `if slept > 0 { currMetrics.recordSlept(slept) }` to avoid a second `AddInt64` for the usual case?

persist/fs/persist_manager.go
xichen2020: Possibly, we'd only stop it when the database is shutting down, so stopping this report loop would require threading the closing logic from the very top-level down to the persistence manager. I could do that if you feel strongly?

persist/fs/persist_manager.go
xichen2020: Sure thing

persist/fs/persist_manager.go
xichen2020: Possibly, could remove

storage/bootstrap.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/m3db/m3x/errors"
 	"github.com/m3db/m3x/log"
 	"github.com/m3db/m3x/time"
+	"github.com/uber-go/tally"
```
robskillington: nit: spare line above this?

storage/bootstrap.go
```diff
@@ -194,6 +160,46 @@ func (m *bootstrapManager) Bootstrap() error {
...
+func (m *bootstrapManager) Report() {
+	if m.IsBootstrapped() {
+		m.status.Update(1)
+	} else {
+		m.status.Update(0)
+	}
+
+}
+
+func (m *bootstrapManager) targetRanges(at time.Time) []bootstrap.TargetRange {
```
robskillington: Is it worth moving this function?  We'll lose all git blame ability to see lines that changed, etc.  

I only bring it up because its a nuanced one.

storage/bootstrap.go
```diff
@@ -194,6 +160,46 @@ func (m *bootstrapManager) Bootstrap() error {
 	return multiErr.FinalError()
 }
 
+func (m *bootstrapManager) Report() {
+	if m.IsBootstrapped() {
+		m.status.Update(1)
+	} else {
+		m.status.Update(0)
+	}
+
```
robskillington: nit: perhaps remove extra line?

storage/bootstrap.go
xichen2020: Sure thing

storage/bootstrap.go
xichen2020: Sure thing

storage/bootstrap.go
xichen2020: It's probably fine, I'd rather have all public APIs at the top of the function and the private implementation details at the bottom. If we really need to git blame this function, we can still do it with some extra effort.

PR #278, opened 2017-02-21
Add ability to dynamically set the rate limit options

robskillington:
This changes allows a HTTP or Thrift call to change the flushing rate limit options.  It also emits the writing and throttle durations for persist flushes as timer metrics.


persist/fs/persist_manager.go
```diff
@@ -85,43 +99,58 @@ type persistManager struct {
...
+	count        int
+	bytesWritten int64
+	worked       time.Duration
+	slept        time.Duration
+
+	metrics persistManagerMetrics
+}
+
+type persistManagerMetrics struct {
+	writeDuration tally.Timer
```
xichen2020: If these are recorded as timers, are we going to look at write-duration.Max for each node?

robskillington: Well the p50, p99, etc will all be the same as they happen just once in a 10s window (since its for all shards, etc).

xichen2020: Yeah just feels kinda strange to use a percentile for this, why not use a gauge?

robskillington: Can emit as a gauge, updating.

PR #279, opened 2017-02-21
Add load shedding when server is overloaded

xichen2020:

This PR adds load shedding when the server is overloaded. We use the number of "commit log queue full" errors as a proxy to measure whether the server is overwhelmed. In particular, if the number of such errors exceeds a given threshold, the server will return "database is overloaded" back to the client.

Load shedding is currently only applicable to `FetchBlocksMetadata` and `FetchBlocks` calls because they are less performance critical, but may be applied to other methods in the future.

### DIFF
--- a/storage/database.go
+++ b/storage/database.go
@@ -34,6 +34,7 @@ import (
 	"github.com/m3db/m3db/storage/block"
 	"github.com/m3db/m3db/storage/namespace"
 	"github.com/m3db/m3db/ts"
+	"github.com/m3db/m3db/x/counter"
 	xio "github.com/m3db/m3db/x/io"
 	"github.com/m3db/m3x/errors"
 	"github.com/m3db/m3x/instrument"
@@ -57,6 +58,9 @@ var (
 
 	// errCommitLogStrategyUnknown raised when trying to use an unknown commit log strategy
 	errCommitLogStrategyUnknown = errors.New("database commit log strategy is unknown")
+
+	// errDatabaseIsOverloaded raised when trying to process a request while the database is overloaded
+	errDatabaseIsOverloaded = errors.New("database is overloaded")
 )
 
 type databaseState int
@@ -99,6 +103,10 @@ type db struct {
 
 	scope   tally.Scope
 	metrics databaseMetrics
+
+	errors       xcounter.FrequencyCounter
+	errWindow    time.Duration
+	errThreshold int64
 }
 
 type databaseMetrics struct {
@@ -106,6 +114,7 @@ type databaseMetrics struct {
 	read                instrument.MethodMetrics
 	fetchBlocks         instrument.MethodMetrics
 	fetchBlocksMetadata instrument.MethodMetrics
+	overloaded          tally.Counter
 }
 
 func newDatabaseMetrics(scope tally.Scope, samplingRate float64) databaseMetrics {
@@ -114,6 +123,7 @@ func newDatabaseMetrics(scope tally.Scope, samplingRate float64) databaseMetrics
 		read:                instrument.NewMethodMetrics(scope, "read", samplingRate),
 		fetchBlocks:         instrument.NewMethodMetrics(scope, "fetchBlocks", samplingRate),
 		fetchBlocksMetadata: instrument.NewMethodMetrics(scope, "fetchBlocksMetadata", samplingRate),
+		overloaded:          scope.Counter("overloaded"),
 	}
 }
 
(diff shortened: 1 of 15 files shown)


storage/database.go
```diff
@@ -99,13 +103,18 @@ type db struct {
...
+	errWindow    time.Duration
+	errThreshold int64
 }
 
 type databaseMetrics struct {
 	write               instrument.MethodMetrics
 	read                instrument.MethodMetrics
 	fetchBlocks         instrument.MethodMetrics
 	fetchBlocksMetadata instrument.MethodMetrics
+	overloaded          tally.Counter
```
robskillington: Should we perhaps call this `overloadedRejected`?

storage/options.go
```diff
@@ -63,6 +64,12 @@ const (
 
 	// defaultRepairEnabled enables repair by default
 	defaultRepairEnabled = true
+
+	// defaultErrorWindowForLoad is the default error window for evaluating server load
+	defaultErrorWindowForLoad = 10 * time.Second
+
+	// defaultErrorThresholdForLoad is the default error threshold for considering server overloaded
+	defaultErrorThresholdForLoad = 100
```
robskillington: 100 in the past 10 seconds seems a little slow to react considering we can easily be doing 150,000 writes per second.  Can we do 100 every 1 second perhaps?

x/counter/frequency_counter_test.go
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: nit: 2017?

x/counter/frequency_counter.go
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: nit: 2017?

storage/database.go
xichen2020: Sure thing

storage/options.go
xichen2020: Yeah I was gonna set it in the config, but sure can change here too

x/counter/frequency_counter_test.go
xichen2020: Sure thing

x/counter/frequency_counter.go
xichen2020: Will do

x/counter/frequency_counter.go
```diff
@@ -0,0 +1,141 @@
...
+// FrequencyCounter keeps track of the frequency counts for the
+// previous interval and the current interval
+type FrequencyCounter struct {
+	nowFn    clock.NowFn
+	interval time.Duration
+	buckets  []frequencyBucket
+}
+
+// NewFrequencyCounter creates a new frequency counter
+func NewFrequencyCounter(opts Options) FrequencyCounter {
```
robskillington: Wonder if its worth moving this into m3x and generalizing as a `CircuitBreaker` or at least the main component of a `CircuitBreaker` that could be reused?

xichen2020: Sure I can put that up as a separate diff

storage/database.go
```diff
@@ -292,8 +308,13 @@ func (d *db) FetchBlocks(
 	starts []time.Time,
 ) ([]block.FetchBlockResult, error) {
 	callStart := d.nowFn()
-	n, err := d.namespaceFor(namespace)
+	if d.isOverloaded() {
```
robskillington: Thinking we should move this up a layer and just expose the `IsOverloaded()` method.  That way people building a general application on top of these components don't need to somehow circumvent it if they don't care about the backoff procedure.

i.e. Use this `IsOverloaded()` check before even beginning an RPC and making the decision at that level.

That makes the `Database` more flexible yet you can still easily do the if check if you want to on top of it.

xichen2020: I went back and forth on this, the nice thing about putting it in the database layer is that the RPC layer doesn't need to know what status the database is in since the logic is encapsulated in the database layer, and load shedding can be disabled via setting the options if so desired. I do however see the argument for the other approach too.

robskillington: I think just in terms of separation of concerns and making the API more flexible and less coupled for this specific case it makes sense to do it at the layer above.

We can always add a reuseable RPC layer in between our current RPC implementation and the database in the future to alleviate any concerns about duplicating this logic down the line.

PR #280, opened 2017-02-21
Fix fetch batch op pooling misuse

robskillington:
This change fixes the use of the op length when calling the call backs for the operation.  This lead to newly enqueued callbacks in the double used operation to be called, causing more than once callback semantics for a given callback, leading to unexpected results.

It also introduces ownership semantics and checking for fetch batch operations for further safety going forward.

cc @xichen2020 @cw9 @prateek @ben-lerner @kobolog

### DIFF
--- a/client/fetch_batch.go
+++ b/client/fetch_batch.go
@@ -124,6 +124,7 @@ func (p *fetchBatchOpPool) Put(f *fetchBatchOp) {
 	f.IncRef()
 	if cap(f.request.Ids) != p.capacity || cap(f.completionFns) != p.capacity {
 		// Grew outside capacity, do not return to pool
+		f.DecRef()
 		return
 	}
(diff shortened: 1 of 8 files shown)


client/fetch_batch.go
```diff
@@ -93,10 +121,12 @@ func (p *fetchBatchOpPool) Get() *fetchBatchOp {
 }
 
 func (p *fetchBatchOpPool) Put(f *fetchBatchOp) {
+	f.IncRef()
 	if cap(f.request.Ids) != p.capacity || cap(f.completionFns) != p.capacity {
 		// Grew outside capacity, do not return to pool
```
xichen2020: Maybe DecRef() here for consistency?

robskillington: Sure thing.

PR #284, opened 2017-02-25
Limit context closing concurrency to avoid goroutine explosion during topology changes

xichen2020:

This PR uses a worker pool to limit the concurrency when closing contexts to avoid the goroutine explosion problem during topology changes.


xichen2020:
Prod tests looking good, merging

PR #288, opened 2017-03-01
Use an insert queue to avoid series flood causing heavy lock contention

robskillington:
This change adds a shard series insert queue used to batch inserts into the shard series map without sacrificing delays to insert the series.

This is important as during floods of new IDs we want to avoid acquiring lock to insert each individual series and insert as many as possible all together acquiring the lock once.

It was experimented also sleeping for a very short duration, i.e. 1ms, during the insert loop and it added enough latency even just with 1ms that it hurt it more than the previous implementation of just acquiring the lock for each series insertion.

The batching as it is without any sleep and just relying on a notification trigger and hot looping when being flooded improved the overall perf by a factor of roughly 4x during floods of new series. (4x less perf degradation during a flood).

### DIFF
--- a/storage/shard_insert_queue.go
+++ b/storage/shard_insert_queue.go
@@ -24,8 +24,6 @@ import (
 	"errors"
 	"sync"
 	"time"
-
-	"github.com/m3db/m3db/ts"
 )
 
 var (
@@ -45,7 +43,7 @@ type dbShardInsertQueue struct {
 	sync.RWMutex
 
 	state              dbShardInsertQueueState
-	insertBatchFn      dbShardInsertBatchFn
+	insertEntryBatchFn dbShardInsertEntryBatchFn
 	insertBatchBackoff time.Duration
 
 	currBatch    *dbShardInsertBatch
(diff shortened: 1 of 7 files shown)


storage/shard_insert_queue.go
```diff
@@ -0,0 +1,165 @@
...
+		insertEntryBatchFn: insertEntryBatchFn,
+		currBatch:          currBatch,
+		notifyInsert:       make(chan struct{}, 1),
+	}
+}
+
+func (q *dbShardInsertQueue) insertLoop() {
+	freeBatch := &dbShardInsertBatch{}
+	freeBatch.reset()
+	for _ = range q.notifyInsert {
```
xichen2020: Could just be `for range q.notifyInsert`?

storage/shard_insert_queue.go
```diff
@@ -0,0 +1,165 @@
...
+	defer q.Unlock()
+
+	if q.state != dbShardInsertQueueStateOpen {
+		return errShardInsertQueueNotOpen
+	}
+
+	q.state = dbShardInsertQueueStateClosed
+
+	// Final flush
+	q.notifyInsert <- struct{}{}
```
xichen2020: Wonder if we should do a select here to avoid blocking

PR #291, opened 2017-03-05
Add a fast decode path when decoding multi iterators

robskillington:
This change removes all use of the iterator heap in the series iterator and multi iterator reader.  This avoids popping and pushing iterators constantly, which helps especially for the single iterator case in the multi iterator reader.

It also helps avoid the overhead of using the heap interface.


encoding/iterators.go
```diff
@@ -0,0 +1,182 @@
...
+		if dp.Timestamp.Before(i.earliestAt) {
+			i.earliest = i.values[idx]
+			i.earliestIdx = idx
+			i.earliestAt = dp.Timestamp
+		}
+	}
+
+	// Apply filter to new earliest if necessary
+	if i.filtering {
+		inFilter := i.earliestAt.Before(i.filterEnd) &&
```
xichen2020: I wonder if it's more efficient to introduce a `filteredNext` method on the inner iterators to return the next value whose timestamp fall within the time range (or nil/false otherwise) and use it to move inner iterators to their respective next values, so if the earliest iterator has many datapoints before the start time, we don't need to recurse many levels deep to find the actual earliest iterator

robskillington: As discussed, the level of recursion is bounded by the number of iterators themselves as you exhaust an iterator one at a time using `moveIteratorToFilterNext` with only one top level for loop.

I'd prefer to keep the logic of sorting and filtering to one place if possible.

encoding/iterators.go
```diff
@@ -0,0 +1,182 @@
...
+
+import (
+	"time"
+
+	"github.com/m3db/m3db/ts"
+	"github.com/m3db/m3x/time"
+)
+
+var (
+	timeMax = time.Unix(1<<63-62135596801, 999999999)
```
robskillington: Sure thing, sounds good.

encoding/multi_reader_iterator.go
```diff
@@ -125,54 +124,31 @@ func (it *multiReaderIterator) moveToNext() {
...
-	curr, _, _ := it.Current()
-	if curr.Timestamp.Equal(prev.Timestamp) {
-		// Dedupe
-		it.moveIteratorsToNext()
-	}
-}
+	for {
+		prev := it.iters.at()
+		next, err := it.iters.moveToValidNext()
+		if err != nil {
```
robskillington: So it's definitely not possible to get here without `it.err` being nil as `moveIteratorsToNext` is only called from `moveToNext` which is only called after checking `hasNext` which checks `!it.hasError()`.

I'll add a check anyhow though.

encoding/iterators.go
xichen2020: Sure thing

encoding/multi_reader_iterator.go
xichen2020: Sounds good

PR #293, opened 2017-04-05
Add option with runtime configuration to asynchronously write new series

robskillington:
This change adds the ability to configure the database to write values for new series asynchronously.  This can be set from config and also at runtime using RPC calls.

This is primarily useful when using the batch write RPC endpoint.  When inserting large volumes of new series the database will appear a lot slower as even existing series will take a much longer time to write to, because naturally they are batched together with the new series values too.

Making writes for new series means that these values will be eventually consistent.  It is still ensured they are written to the commit log so the writes are recoverable, they just will not be visible until just after the write, once the insertion into the shard lookup is complete for the new series.

### DIFF
--- a/network/server/tchannelthrift/node/service.go
+++ b/network/server/tchannelthrift/node/service.go
@@ -577,7 +577,8 @@ func (s *service) Truncate(tctx thrift.Context, req *rpc.TruncateRequest) (r *rp
 func (s *service) GetPersistRateLimit(
 	ctx thrift.Context,
 ) (*rpc.NodePersistRateLimitResult_, error) {
-	opts := s.db.Options().PersistManager().RateLimitOptions()
+	runtimeOptsMgr := s.db.Options().RuntimeOptionsManager()
+	opts := runtimeOptsMgr.Get().PersistRateLimitOptions()
 	limitEnabled := opts.LimitEnabled()
 	limitMbps := opts.LimitMbps()
 	limitCheckEvery := int64(opts.LimitCheckEvery())
@@ -593,9 +594,9 @@ func (s *service) SetPersistRateLimit(
 	ctx thrift.Context,
 	req *rpc.NodeSetPersistRateLimitRequest,
 ) (*rpc.NodePersistRateLimitResult_, error) {
-	persistManager := s.db.Options().PersistManager()
-
-	opts := persistManager.RateLimitOptions()
+	runtimeOptsMgr := s.db.Options().RuntimeOptionsManager()
+	runopts := runtimeOptsMgr.Get()
+	opts := runopts.PersistRateLimitOptions()
 	if req.LimitEnabled != nil {
 		opts = opts.SetLimitEnabled(*req.LimitEnabled)
 	}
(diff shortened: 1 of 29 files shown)


network/server/tchannelthrift/node/service.go
```diff
@@ -611,6 +611,24 @@ func (s *service) SetPersistRateLimit(
 	return s.GetPersistRateLimit(ctx)
 }
 
+func (s *service) GetWriteNewSeriesAsyncResult(
+	ctx thrift.Context,
+) (*rpc.NodeWriteNewSeriesAsyncResult_, error) {
+	value := s.db.RuntimeOptionsManager().Get().WriteNewSeriesAsync()
```
xichen2020: Perhaps we should do the same for rate limit values to be consistent? I'm fine with doing it in a separate diff also.

storage/types.go
```diff
@@ -616,4 +619,43 @@ type Options interface {
...
+	// Update updates the current runtime options
+	Update(value RuntimeOptions)
+
+	// Get returns the current values
+	Get() RuntimeOptions
+
+	// GetAndWatch begins a new watch also returning the current values,
+	// the returned watch will not notify on the channel until a new update
+	// is available
+	GetAndWatch() (RuntimeOptions, RuntimeOptionsWatch)
```
xichen2020: Perhaps just call it `Watch`?

storage/shard.go
```diff
@@ -199,6 +213,24 @@ func newDatabaseShard(
 	return d
 }
 
+func (s *dbShard) initRuntimeOptions(mgr RuntimeOptionsManager) {
```
xichen2020: I wonder if it's better to make this more generic. The runtime option setup is generic enough so it can be used on other types as well, so perhaps we should have a `RuntimeUpdater` interface:
```
type RuntimeUpdater interface{
  RuntimeOptions() RuntimeOptions
  SetRuntimeOptions()
}
```
and have `dbShard` implement the interface

storage/shard_insert_queue.go
```diff
@@ -24,6 +24,9 @@ import (
 	"errors"
 	"sync"
 	"time"
+
+	xtime "github.com/m3db/m3x/time"
+	"github.com/uber-go/tally"
```
xichen2020: Blank line in between?

storage/shard_insert_queue.go
```diff
@@ -48,24 +51,60 @@ type dbShardInsertQueue struct {
...
+	annotation []byte
 }
 
 func (b *dbShardInsertBatch) reset() {
 	b.wg = &sync.WaitGroup{}
 	// We always expect to be waiting for an insert
 	b.wg.Add(1)
-	for i := range b.entries {
-		b.entries[i] = nil
+	insertZeroed := dbShardInsert{pendingWrite: dbShardPendingWrite{}}
```
xichen2020: How often does `reset` get called? Should we perhaps make the empty `dbShardInsert` object a file-scope `var`?

storage/shard.go
```diff
@@ -441,27 +473,70 @@ func (s *dbShard) Write(
...
+				timestamp:  timestamp,
+				value:      value,
+				unit:       unit,
+				annotation: annotation,
+			},
+		})
+		if err != nil {
+			return err
+		}
+		commitLogSeriesID = result.copiedID
```
robskillington: Note to self: add comment here about why this needs to be a copied ID.

storage/shard.go
```diff
@@ -441,27 +473,70 @@ func (s *dbShard) Write(
...
+	var (
+		commitLogSeriesID          ts.ID
+		commitLogSeriesUniqueIndex uint64
+	)
+	if writable {
+		// Perform write
+		err = entry.series.Write(ctx, timestamp, value, unit, annotation)
+		// Load series metadata before decrementing the writer count
+		// to ensure this metadata is snapshotted at a consistent state
+		commitLogSeriesID = entry.series.ID()
```
xichen2020: Do we return id to pool? If so, and we decrement writer count before finishing using the id, the id might be corrupted.

network/server/tchannelthrift/node/service.go
robskillington: Yeah I can definitely update the rate limit values too.

storage/shard.go
robskillington: We explicitly don't pool IDs from series to protect against this very case, here is the comment from `series.go`:
```go
	// NB(r): We explicitly do not place this ID back into an
	// existing pool as high frequency users of series IDs such
	// as the commit log need to use the reference without the
	// overhead of ownership tracking.
	// Since series are purged so infrequently the overhead
	// of not releasing back an ID to a pool is amortized over
	// a long period of time.
```

I can add a similar comment here though, that's a good idea.

storage/shard_insert_queue.go
robskillington: Sure thing.

storage/shard_insert_queue.go
robskillington: Sure thing, will do.

robskillington:
@xichen2020 this is ready for review again, ta!

PR #294, opened 2017-04-25
Enforce max wired blocks using a LRU wired list

robskillington:
This change adds an LRU mechanism for evicting wired blocks after a limit is reached.  This ensures that heavy scans across huge datasets in the database doesn't cause the process to exceed a threshold of wired blocks, hence protecting the process from OOMing.

cc @xichen2020 @martin-mao @prateek @ben-lerner


storage/block/block.go
```diff
@@ -42,31 +42,49 @@ var (
...
 	retriever  DatabaseShardBlockRetriever
 	retrieveID ts.ID
 
+	// For use in a list of blocks, notably for the wired list. The
+	// doubly linked list pointers are on the block themselves to avoid
+	// creating a heap allocated struct per block.
+	// NB(r): Beware that the list controls the synchronization of these
+	// fields to avoid unnecessary lock contention.
+	// They should never be read or mutated by the block itself.
+	next *dbBlock
```
kobolog: As a design consideration: this pattern is commonly known as "intrusive container". In practice, you would expect a type that implements such a pattern to inherit from or embed a "hook" subtype that is implemented by the intrusive container package, e.g.:

```go
type dbBlock struct {
    wired.Hook
    // or if golang had templates =)
    wired.Hook<dbBlock>
    // or
    hook wired.Hook
}
```

In my opinion, this helps with readability and the separation of concerns in the code.

robskillington: Hey that definitely would be a step up, thanks for the feedback.  Will use an embedded type to encapsulate the next/prev.

robskillington: Yeah so this actually can't be done in Go as when you follow the pointer to the next element, there's no way to actually get a reference to the enclosing struct.  e.g.:

```go
type list struct {
  root listElem
}

type listElem struct {
  next *listElem

  // you could create an `interface{}` type field here that lets you reference back to the enclosing struct.. however that would cause slow and gross cast back to the enclosing struct
}

type foo struct {
  listElem

  x int
  // other fields 
}

func useList() {
  // create list
  l := &list{}
  l.root.next = &l.root.listElem

  // insert
  x, y, z := &foo{x: 1}, &foo{x: 2}, &foo{x: 3}
  l.root.next = &x.listElem
  x.next = &y.listElem
  y.next = &z.listElem

  // try to iterate and access the "foo" fields
  for elem := l.root.next; elem.next != &l.root.listElem; elem = elem.next {
    // only have a reference to *listElem now
  }
}
```

storage/block/block.go
```diff
@@ -141,6 +160,38 @@ func (b *dbBlock) OnRetrieveBlock(
 	b.resetSegmentWithLock(segment)
 }
 
+func (b *dbBlock) updateWiredList() {
+	list := b.opts.WiredList()
+	list.update(b)
+}
+
+func (b *dbBlock) wiredWithLock() bool {
```
xichen2020: `isWiredWithLock` might be better?

storage/block/block.go
```diff
@@ -172,43 +246,57 @@ func (b *dbBlock) Stream(blocker context.Context) (xio.SegmentReader, error) {
...
-		stream = newDatabaseMergedBlockReader(b.startWithLock(), stream, mergeStream, b.opts)
-	}
-
-	// Register the finalizer for the stream
-	blocker.RegisterFinalizer(stream)
-
 	return stream, nil
 }
 
 func (b *dbBlock) IsRetrieved() bool {
```
xichen2020: Should this be called `IsWired` then?

storage/block/wired_list.go
```diff
@@ -0,0 +1,192 @@
...
+
+import (
+	"sync"
+
+	"github.com/m3db/m3x/instrument"
+	"github.com/uber-go/tally"
+)
+
+const (
+	wiredListEventsChannelLength = 65536
```
xichen2020: Would it make sense to make these configurable?

storage/block/wired_list.go
```diff
@@ -0,0 +1,192 @@
...
+	wiredListSampleGaugesEvery   = 100
+)
+
+// WiredList is a database block wired list.
+type WiredList struct {
+	sync.RWMutex
+
+	maxWired  int
+	root      dbBlock
+	len       int
```
xichen2020: Call this `size` or `length` perhaps to avoid colliding with builtin function `len`?

storage/block/wired_list.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+const (
+	wiredListEventsChannelLength = 65536
+	wiredListSampleGaugesEvery   = 100
+)
+
+// WiredList is a database block wired list.
+type WiredList struct {
+	sync.RWMutex
```
xichen2020: Doesn't look like we ever use the reader lock, so `sync.Mutex` should be sufficient?

storage/block/wired_list.go
```diff
@@ -0,0 +1,192 @@
...
+func (l *WiredList) update(v *dbBlock) {
+	// Trigger for processing
+	l.updatesCh <- v
+}
+
+func (l *WiredList) processUpdateBlock(v *dbBlock) {
+	entry := v.wiredListEntry()
+
+	// If already in list
+	if l.exists(v) {
```
xichen2020: Could simplify this function to
```
entry := v.wiredListEntry()
if !entry.wired {
  l.remove(v)
  return
}
l.pushBack(v)
```

storage/block/wired_list.go
```diff
@@ -0,0 +1,192 @@
...
+		return
+	}
+
+	// Not in the list, if wired add it
+	if entry.wired {
+		l.pushBack(v)
+	}
+}
+
+func (l *WiredList) insert(v, at *dbBlock) {
```
xichen2020: might be better to call this `insertAfter`, since most list insert functions insert before the given element. Also can v or at be nil?

storage/block/wired_list.go
```diff
@@ -0,0 +1,192 @@
...
+	n.prev = v
+	l.len++
+
+	if l.maxWired <= 0 {
+		// Not enforcing max wired blocks
+		return
+	}
+
+	// Try to unwire all blocks possible
+	for bl := l.root.next; l.len > l.maxWired && bl != &l.root; bl = bl.next {
```
xichen2020: Can we perhaps do this asynchronously in batches in the background? Otherwise once we get to `maxWired`, every insert will need to do an unwire operation.

storage/block/types.go
```diff
@@ -317,12 +315,24 @@ type Options interface {
...
 
-	// DatabaseBlockAllocSize returns the databaseBlockAllocSize
+	// DatabaseBlockAllocSize returns the database block alloc size
 	DatabaseBlockAllocSize() int
 
+	// SetWiredList sets the database block wired list
+	SetWiredList(value *WiredList) Options
+
+	// WiredList returns the database block wired list
+	WiredList() *WiredList
```
prateek: General question: i don't know how uniform the load across the shards is, if we have hotspotting, is there value in making the LRUCache per shard?

storage/series/series.go
```diff
@@ -399,9 +399,11 @@ func (s *dbSeries) mergeBlock(
 		return
 	}
 
-	// We are performing this in a lock, cannot wait for the existing
-	// block potentially to be retrieved from disk, lazily merge the stream.
-	newBlock.Merge(existingBlock)
+	if err := newBlock.Merge(existingBlock); err != nil {
```
robskillington: Since there's no one to float the error back to we kind of have to do this here.  FWIW this is what we were doing before when merge could return an error instead of lazily be evaluated.

storage/block/block.go
robskillington: Sure thing.

storage/block/block.go
robskillington: Yeah probably, I can go through and update all the terminology.

storage/block/wired_list.go
robskillington: Good call, will remove.

storage/block/wired_list.go
robskillington: Good call yeah, will do.

storage/block/wired_list.go
robskillington: Sounds good yeah.

robskillington: Sounds good yeah.

storage/block/wired_list.go
robskillington: This is asynchronous in the background, all operations in the wired list is done after processing a request for an updated from a channel.

storage/block/types.go
robskillington: Good question, so this is definitely valid - that is why I made the API to update WiredList completely asynchronous.  You simply enqueue an update to the wired list and a background goroutine reads the update and updates the list.  Enqueuing to a channel can be contentions even itself however, so this enqueuing is done asynchronously after a read completes with the onStreamDoneFinalizer on the block.

This avoids the central contention and allows reads to still complete quickly.

storage/block/types.go
```diff
@@ -317,12 +315,24 @@ type Options interface {
...
+
+	// SetDatabaseBlockAllocSize sets the database block alloc size
 	SetDatabaseBlockAllocSize(value int) Options
 
-	// DatabaseBlockAllocSize returns the databaseBlockAllocSize
+	// DatabaseBlockAllocSize returns the database block alloc size
 	DatabaseBlockAllocSize() int
 
+	// SetWiredList sets the database block wired list
+	SetWiredList(value *WiredList) Options
```
robskillington: Thankfully it's relatively simple and is safe to use across different layers, i.e. we don't need to mock it.  Hence why I'm currently inclined to leave it as a pointer type.

storage/block/block.go
```diff
@@ -238,10 +344,10 @@ func (b *dbBlock) resetSegmentWithLock(seg ts.Segment) {
 	b.length = seg.Len()
 	b.checksum = digest.SegmentChecksum(seg)
 
-	b.retriever = nil
```
robskillington: The general idea now is that you should keep your retriever around so that if you want to be unloaded you can as you still have a reference to it.

storage/block/wired_list.go
```diff
@@ -138,22 +157,27 @@ func (l *WiredList) insert(v, at *dbBlock) {
 	at.next = v
```
xichen2020: Call this method `insertAfter`?

network/server/tchannelthrift/node/service.go
```diff
@@ -632,6 +632,47 @@ func (s *service) SetWriteNewSeriesAsync(
...
+	return s.GetMaxWiredBlocks(ctx)
+}
+
+func (s *service) GetWiredBlockExpiryAfterNotAccessedPeriod(
+	ctx thrift.Context,
+) (*rpc.NodeWiredBlockExpiryAfterNotAccessedPeriodResult_, error) {
+	runtimeOptsMgr := s.db.Options().RuntimeOptionsManager()
+	value := runtimeOptsMgr.Get().WiredBlockExpiryAfterNotAccessedPeriod()
+	return &rpc.NodeWiredBlockExpiryAfterNotAccessedPeriodResult_{
+		ExpiryMilliseconds: int64(value / time.Millisecond),
```
robskillington: Well other Thrift APIs actually use `timestamp` and `timestampType`.  I can change this to `duration` and `durationType` and allow SECONDS, MILLISECONDS, MICROSECONDS, NANOSECONDS to be used.

runtime/runtime_options.go
```diff
@@ -20,22 +20,40 @@
...
 
 const (
+	// defaultWriteNewSeriesAsync is false by default to give consistent
+	// writes by default
 	defaultWriteNewSeriesAsync = false
+
+	// defaultMaxWiredBlocks is 2MM by default which if using 2hr block sizes
+	// and writing every 10s at 1.4 point/byte (m3tsz) should use roughly 4gb:
+	// 1.4 * 6 * 120 * (2^21) = ~2gb
+	defaultMaxWiredBlocks = 2097152
```
robskillington: It's not bytes though, this is the max number of wired blocks we allow in process.

storage/block/wired_list.go
robskillington: Sure thing.

storage/block/wired_list.go
```diff
@@ -138,22 +157,27 @@ func (l *WiredList) insert(v, at *dbBlock) {
...
 	v.prev = at
 	v.next = n
+	v.nextPrevUpdatedAtUnixNano = time.Now().UnixNano()
```
robskillington: Good catch, thanks yeah.

runtime/runtime_options.go
xichen2020: Oh sorry misread the comment, thought it was referring to 2gb number

network/server/tchannelthrift/node/service.go
xichen2020: Yeah I see, don't feel strongly either way as long as it is consistent.

storage/block/wired_list.go
robskillington: This is used for start/stop

network/server/tchannelthrift/convert/convert.go
```diff
@@ -49,15 +50,30 @@ func ToTime(value int64, timeType rpc.TimeType) (time.Time, error) {
...
+func ToDuration(duration int64, durationType rpc.DurationType) (time.Duration, error) {
+	switch durationType {
+	case rpc.DurationType_SECONDS:
+		return time.Duration(duration) * time.Second, nil
+	case rpc.DurationType_MILLISECONDS:
+		return time.Duration(duration) * time.Millisecond, nil
+	case rpc.DurationType_MICROSECONDS:
+		return time.Duration(duration) * time.Microsecond, nil
+	case rpc.DurationType_NANOSECONDS:
+		return time.Duration(duration) * time.Microsecond, nil
```
xichen2020: Should be `time.Nanosecond`

robskillington: Good catch

storage/block/types.go
prateek: Makes sense.

richardartoul:
@r you can probably close this

PR #295, opened 2017-05-14
Fixing m3tsz int overflow edge case

martin-mao:
Fixing case where m3tsz int optimization overflows with very large ints.

### DIFF
--- a/encoding/m3tsz/roundtrip_test.go
+++ b/encoding/m3tsz/roundtrip_test.go
@@ -21,6 +21,7 @@
 package m3tsz
 
 import (
+	"math"
 	"math/rand"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


xichen2020:
I assume the failing test is unrelated to this change?

encoding/m3tsz/roundtrip_test.go
```diff
@@ -220,3 +225,20 @@ func generateMixedDatapoints(numPoints int, timeUnit time.Duration) []ts.Datapoi
 	}
 	return res
 }
+
+func generateOverflowDatapoints() []ts.Datapoint {
+	var startTime int64 = 1427162462
+	currentTime := time.Unix(startTime, 0)
+	largeInt := float64(math.MaxInt64 - 1)
```
robskillington: Shouldn't this be `float64(math.MaxInt64)-1` to be more safe about compiler semantics here?  Same with the variable below.

PR #296, opened 2017-05-15
Add `clone_fileset` tool

prateek:
Wired up a tool to clone filesets using `fs.FileSetReader|FileSetWriter`. 

/cc @robskillington

### DIFF
--- a/tools/clone_fileset/main/main.go
+++ b/tools/clone_fileset/main/main.go
@@ -9,14 +9,11 @@ import (
 	"github.com/m3db/m3db/persist/fs"
 	"github.com/m3db/m3db/ts"
 	xlog "github.com/m3db/m3x/log"
-	"github.com/m3db/m3x/pool"
 	xtime "github.com/m3db/m3x/time"
 )
 
 const (
-	defaultBufferSize      = 4096
-	defaultBufferCapacity  = 256
-	defaultBufferPoolCount = 3145728
+	defaultBufferSize = 4096
 )
 
 var (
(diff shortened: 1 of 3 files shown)


tools/clone_fileset/main/main.go
```diff
@@ -0,0 +1,98 @@
...
+		*optDestPathPrefix == "" ||
+		*optSrcNamespace == "" ||
+		*optDestNamespace == "" ||
+		*optSrcBlockstart <= 0 ||
+		*optDestBlockstart <= 0 {
+		flag.Usage()
+		os.Exit(1)
+	}
+
+	bytesPool := pool.NewCheckedBytesPool([]pool.Bucket{pool.Bucket{
```
robskillington: Hey as discussed, found for this case no pooling is roughly 3x faster (probably because GC never has enough time to kick in during execution as its less than a few seconds) .

Let's kick the pooling out?

prateek: sounds good, done.

tools/clone_fileset/main/main.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/m3db/m3db/persist/encoding/msgpack"
+	"github.com/m3db/m3db/persist/fs"
+	"github.com/m3db/m3db/ts"
+	xlog "github.com/m3db/m3x/log"
+	xtime "github.com/m3db/m3x/time"
+)
+
+const (
+	defaultBufferSize = 4096
```
robskillington: Let's use 64kb that gives a slight boost in write perf, also its what we use as default settings for M3DB and in production.  That is 65536 here.

PR #297, opened 2017-05-22
Add ability to read last read time from block

robskillington:
This change adds the ability to read the last read time for a block from the `FetchBlocksMetadata` API.  This allows for one to observe the analyze the read patterns of a cluster by using the client's `FetchBlocksMetadataFromPeers(...)` method.

cc @xichen2020 @prateek

### DIFF
--- a/storage/series/buffer.go
+++ b/storage/series/buffer.go
@@ -386,14 +386,12 @@ func (b *dbBuffer) FetchBlocksMetadata(
 		if opts.IncludeSizes {
 			resultSize = size
 		}
-		// NB(r): Ignore if opts.IncludeChecksum because we avoid
-		// calculating checksum since block is open and is being mutated
 		var resultLastRead time.Time
 		if opts.IncludeLastRead {
 			resultLastRead = bucket.lastRead()
 		}
-		// NB(xichen): intentionally returning nil checksums for buckets
-		// because they haven't been flushed yet
+		// NB(r): Ignore if opts.IncludeChecksum because we avoid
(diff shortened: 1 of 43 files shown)


storage/series/buffer.go
```diff
@@ -240,6 +240,10 @@ func bucketDrainAndReset(now time.Time, b *dbBuffer, idx int, start time.Time) i
 			mergedOutOfOrderBlocks = 1
 		}
 		if merged.block.Len() > 0 {
+			// If this block was read mark it as such
+			if lastRead := b.buckets[idx].lastRead(); !lastRead.IsZero() {
+				merged.block.SetLastReadTime(lastRead)
```
xichen2020: What happens if the block already has a last read time? Do we want to take the latter of the two read times?

robskillington: This block is newly created from the `discardMerged(...)` call so it doesn't have any last read time.

xichen2020: Cool yeah sounds good

storage/series/buffer.go
```diff
@@ -371,13 +382,23 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
 			return
 		}
-		var pSize *int64
-		if includeSizes {
-			pSize = &size
+		var resultSize int64
+		if opts.IncludeSizes {
+			resultSize = size
+		}
+		// NB(r): Ignore if opts.IncludeChecksum because we avoid
```
prateek: nit: merge this comment with Xi's from below

PR #298, opened 2017-05-25
Upgrade m3db vendored packages

prateek:
Upgrading vendored packages to latest versions. 

- Update Makefile to `m3db/ci-scripts` convention 
- Pull latest m3cluster version, corresponding changes
- Pull latest Tally version, corresponding changes

/cc @robskillington

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -12,7 +12,7 @@ import:
   - gomock
 
 - package: github.com/golang/protobuf
-  version: bf531ff1a004f24ee53329dfd5ce0b41bfdc17df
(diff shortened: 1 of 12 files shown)


glide.yaml
```diff
@@ -36,16 +36,16 @@ import:
   version: 5c3c0fce48842b2c0bbaa99b4e61b0175d84b47c
 
 - package: github.com/uber-go/tally
-  version: v1.1.0
+  version: v3.0.0
```
robskillington: Perhaps we make this a bit looser with "<4" so that any possible compatible library can be used.

prateek: sounds good, addressed.

PR #301, opened 2017-05-27
Extract integration data generation routines into sub-package

prateek:
- New `integration/generate` package to generate faked series data: in-memory and on-disk 

/cc @robskillington


prateek:
@robskillington updated code to include new `generate.Writer`. Please take another look

PR #309, opened 2017-06-25
Add bootstrapper configuration 

prateek:
Part of work to make `m3dbnode` buildable from GH.

/cc @robskillington

### DIFF
--- a/services/m3dbnode/config/bootstrap.go
+++ b/services/m3dbnode/config/bootstrap.go
@@ -46,8 +46,8 @@ type filesystemConfiguration struct {
 	NumProcessorsPerCPU float64 `yaml:"numProcessorsPerCPU" validate:"min=0.0"`
 }
 
-// Configuration captures the configuration for bootstrappers.
-type Configuration struct {
+// BootstrapConfiguration captures the configuration for bootstrappers.
+type BootstrapConfiguration struct {
 	// Bootstrappers is the list of bootstrappers, ordered by precedence in descending order
 	Bootstrappers []string `yaml:"bootstrappers" validate:"nonzero"`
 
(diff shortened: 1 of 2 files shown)


robskillington:
Hey this looks good but perhaps we should put this into a package named `config` at the location `services/m3dbnode/config`.  It really is specific for running that service (or eventually will be).

prateek:
@robskillington sure thing, done.

services/m3dbnode/config/bootstrap.go
```diff
@@ -0,0 +1,118 @@
...
+type BootstrapConfiguration struct {
+	// Bootstrappers is the list of bootstrappers, ordered by precedence in descending order
+	Bootstrappers []string `yaml:"bootstrappers" validate:"nonzero"`
+
+	// Filesystem bootstrapper configuration
+	FilesystemConfiguration *filesystemConfiguration `yaml:"fs"`
+}
+
+func (bsc BootstrapConfiguration) numProcessors() int {
+	if bsc.FilesystemConfiguration == nil {
```
xichen2020: nit: might be slightly cleaner to do
```
var defaultNumProcessorsPerCPU = 0.5

func (bsc BootstrapConfiguration) numProcessors() int {
  np := defaultNumProcessorsPerCPU
  if bsc.FilesystemConfiguration != nil {
    np = bsc.FilesystemConfiguration.NumProcessorsPerCPU
  }
  return int(math.Ceil(float64(runtime.NumCPU()) * np))
}
```
You don't need the `>0.0` check because of `validate:"min=0.0"`

prateek: sure thing

prateek: done

PR #315, opened 2017-07-14
Speed up commitlog bootstrapper by parallelizing bootstrapper and reader

richardartoul:
Speed up the commitlog bootstrapper with the following changes:

1) Parallelize the commitlog reader such that msgpack decoding is done in parallel
2) Parallelize the commitlog bootstrap such that M3TSZ encoding is done in parallel
3) Replace a frequently accessed map with a slice
4) Replace a map[time.Time] with a map[f64] for faster access
5) Update msgpack library to take advantage of speed improvements
6) Fix file descriptors not being immediately closed when switching from one file to the next

With a concurrency of 4 for the reader and a concurrency of 4 for the M3TSZ encoders I was able to get bootstrapping time down from 3 hours to 35 minutes. CPU utilization sits in the 600-700% range with occasional short-lived spikes in the 10000+% range (I assume due to garbage collection). Maximum utilized memory was ~42G towards the end of the bootstrapping.

I'll do some more experimentation with different concurrency numbers to see what our options there are.

Also, I git cherry-picked this pull request: https://github.com/m3db/m3db/pull/314 and ran the tests to make sure I hadn't subtly altered the behavior of the bootstrapper and the tests passed.

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -30,7 +30,7 @@ import:
   - thrift
 
 - package: gopkg.in/vmihailenco/msgpack.v2
-  version: master
+  version: f4f8982de4ef0de18be76456617cc3f5d8d8141e
 
 - package: github.com/willf/bitset
   version: 5c3c0fce48842b2c0bbaa99b4e61b0175d84b47c
--- a/glide.yaml
+++ b/glide.yaml
@@ -30,7 +30,7 @@ import:
   - thrift
 
 - package: gopkg.in/vmihailenco/msgpack.v2
-  version: a1382b1ce0c749733b814157c245e02cc1f41076
+  version: master
 
 - package: github.com/willf/bitset
   version: 5c3c0fce48842b2c0bbaa99b4e61b0175d84b47c
--- a/integration/commitlog_bootstrap_helpers.go
+++ b/integration/commitlog_bootstrap_helpers.go
@@ -103,6 +103,9 @@ func writeCommitLog(
 		// write this point
 		idx, ok := indexes[point.ID.String()]
 		require.True(t, ok)
+		// fmt.Println("ID: ", point.ID)
+		// fmt.Println("Shard: ", shardSet.Lookup(point.ID))
+		// fmt.Println("-------------------------------------")
 		cId := commitlog.Series{
 			Namespace:   namespace,
 			Shard:       shardSet.Lookup(point.ID),
--- a/integration/commitlog_bootstrap_helpers.go
+++ b/integration/commitlog_bootstrap_helpers.go
@@ -1,0 +1,132 @@
+// +build integration
+
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package integration
+
+import (
+	"testing"
+	"time"
+
+	"github.com/m3db/m3db/context"
+	"github.com/m3db/m3db/integration/generate"
+	"github.com/m3db/m3db/persist/fs/commitlog"
+	"github.com/m3db/m3db/storage/block"
+	"github.com/m3db/m3db/ts"
+	"github.com/m3db/m3x/time"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	defaultIntegrationTestFlushInterval = 10 * time.Millisecond
+)
+
+func computeMetricIndexes(timeBlocks generate.SeriesBlocksByStart) map[string]uint64 {
+	var idx uint64
+	indexes := make(map[string]uint64)
+	for _, blks := range timeBlocks {
+		for _, blk := range blks {
+			id := blk.ID.String()
+			if _, ok := indexes[id]; !ok {
+				indexes[id] = idx
+				idx++
+			}
+		}
+	}
+	return indexes
+}
+
+func writeCommitLog(
+	t *testing.T,
+	s *testSetup,
+	data generate.SeriesBlocksByStart,
+	namespace ts.ID,
+) {
+	indexes := computeMetricIndexes(data)
+	opts := s.storageOpts.CommitLogOptions()
+
+	// ensure commit log is flushing frequently
+	require.Equal(t, defaultIntegrationTestFlushInterval, opts.FlushInterval())
+	ctx := context.NewContext()
+
+	var (
+		shardSet  = s.shardSet
+		points    = generate.ToPointsByTime(data) // points are sorted in chronological order
+		blockSize = opts.RetentionOptions().BlockSize()
+		commitLog commitlog.CommitLog
+		now       time.Time
+	)
+
+	closeCommitLogFn := func() {
+		if commitLog != nil {
+			// wait a bit to ensure writes are done, and then close the commit log
+			time.Sleep(10 * defaultIntegrationTestFlushInterval)
+			require.NoError(t, commitLog.Close())
+		}
+
+	}
+
+	for _, point := range points {
+		pointTime := point.Timestamp
+
+		// check if this point falls in the current commit log block
+		if truncated := pointTime.Truncate(blockSize); truncated != now {
+			// close commit log if it exists
+			closeCommitLogFn()
+			// move time forward
+			now = truncated
+			s.setNowFn(now)
+			// create new commit log
+			commitLog = commitlog.NewCommitLog(opts)
+			require.NoError(t, commitLog.Open())
+		}
+
+		// write this point
+		idx, ok := indexes[point.ID.String()]
+		require.True(t, ok)
+		cId := commitlog.Series{
+			Namespace:   namespace,
+			Shard:       shardSet.Lookup(point.ID),
+			ID:          point.ID,
+			UniqueIndex: idx,
+		}
+		require.NoError(t, commitLog.WriteBehind(ctx, cId, point.Datapoint, xtime.Second, nil))
+	}
+
+	closeCommitLogFn()
+}
+
+func testSetupMetadatas(
+	t *testing.T,
+	testSetup *testSetup,
+	namespace ts.ID,
+	start time.Time,
+	end time.Time,
+) map[uint32][]block.ReplicaMetadata {
+	// Retrieve written data using the AdminSession APIs
+	// FetchMetadataBlocksFromPeers/FetchBlocksFromPeers
+	adminClient := testSetup.m3dbAdminClient
+	metadatasByShard, err := m3dbClientFetchBlocksMetadata(
+		adminClient, namespace, testSetup.shardSet.AllIDs(), start, end)
+	require.NoError(t, err)
+	return metadatasByShard
+}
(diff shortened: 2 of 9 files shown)


persist/fs/commitlog/reader.go
```diff
@@ -110,52 +156,121 @@ func (r *reader) Read() (
...
+		for true {
+			r.metadataLookupLock.Lock()
+			metadata, ok = r.metadataLookup[entry.Index]
+			r.metadataLookupLock.Unlock()
+
+			if ok {
+				break
+			}
+
+			// Race condition where the metadata is currently being processed by
```
richardartoul: I really don't like sleeping in here, but I couldn't think of a better way to synchronize the map.

The performance loss due to the sleep is negligible because this doesn't happen that often (less than 20 times per file)

glide.yaml
```diff
@@ -30,7 +30,7 @@ import:
   - thrift
 
 - package: gopkg.in/vmihailenco/msgpack.v2
-  version: a1382b1ce0c749733b814157c245e02cc1f41076
+  version: f4f8982de4ef0de18be76456617cc3f5d8d8141e
```
prateek: Did you need newer features from msgpack?

persist/fs/commitlog/reader.go
```diff
@@ -100,6 +143,11 @@ func (r *reader) Open(filePath string) (time.Time, time.Duration, int, error) {
 	start := time.Unix(0, info.Start)
 	duration := time.Duration(info.Duration)
 	index := int(info.Index)
+
+	go r.readLoop()
```
prateek: Any reason to start decoding at `Open()` instead of the first `Read()`?

persist/fs/commitlog/reader.go
```diff
@@ -165,16 +283,10 @@ func (r *reader) readChunk() ([]byte, error) {
...
 	}
 
-	// Extend buffer as necessary
-	if len(r.dataBuffer) < int(size) {
-		diff := int(size) - len(r.dataBuffer)
-		for i := 0; i < diff; i++ {
-			r.dataBuffer = append(r.dataBuffer, 0)
-		}
-	}
+	newBuf := make([]byte, size, size)
```
prateek: maybe worth investigating perf impact if you pooled this

richardartoul: Yeah I actually had a mental TODO to ask you about that. I've never done object pooling before. Do I just do a .Get() when I need a new one and a .Put() when I'm done with it?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -121,129 +207,144 @@ func (s *commitLogSource) Read(
...
-		}
+			<-mergeSemaphore
+			wg.Done()
+		}(shard, unmergedShard)
+	}
+	// Wait for all merge goroutines to complete
+	wg.Wait()
+
+	endMergeTime := time.Now()
+	s.log.Infof("Complete merge took: %fs\n", endMergeTime.Sub(startMergeTime).Seconds())
```
prateek: emit metrics for each phase rather than logging. The entry point into different bootstrappers (`bootstrap.Process`) already [logs](https://github.com/m3db/m3db/blob/master/storage/bootstrap/process.go#L109) the e2e time per bootstrapper,

persist/fs/commitlog/reader.go
richardartoul: woops forgot we discussed that, I can change that.

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -95,13 +108,86 @@ func (s *commitLogSource) Read(
...
+	encoderChans := []chan encoderArg{}
+	for i := uint32(0); i < numConc; i++ {
+		encoderChans = append(encoderChans, make(chan encoderArg, 1000))
+	}
+
+	// Spin up numConc background go-routines to handle M3TSZ encoding. This must
+	// happen before we start reading to prevent infinitely blocking writes to
+	// the encoderChans.
+	wg := sync.WaitGroup{}
+	workerErrs := make([]int, numConc, numConc)
```
prateek: Not related to your change, but considering you're reworking this part, it'd be good to refactor this method into smaller chunks.

glide.yaml
richardartoul: No I just saw in their commit history they had made performance improvements since the version we had pinned

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -95,13 +108,86 @@ func (s *commitLogSource) Read(
 	defer iter.Close()
 
 	var (
-		unmerged    = make(map[uint32]map[ts.Hash]encoderMap)
+		numShards = len(shardsTimeRanges)
+		unmerged  = make([]map[ts.Hash]encoderMap, numShards, numShards)
```
prateek: Add types for the different sub-parts of `unmerged, and make the operations used in the read/merge phase functions on those types. 

E.g. for the read side, it could be something like:

```
type shardEncoders map[ts.Hash]encoderMap
type shardSetEncoders map[uint32]shardEncoders
func (s *shardSetEncoders) GetOrAdd(shardID uint32) shardEncoders {} 
func (s *shardEncoders) GetOrAdd(seriesID ts.ID) encoderMap {} 
```

prateek: Revert to `make(map[uint32]map[ts.Hash]encoderMap)` instead of `[]map[ts.Hash]encoderMap`: shard allocations aren't contiguous or zero-based, so a node could have `numShards=10` and the shards could be `1,10,20,30,40..`

You could pre-allocate the map to have size `numShards` if that helps perf

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -95,13 +108,86 @@ func (s *commitLogSource) Read(
...
+						})
+						unmergedSeries.encoders[blockStart.UnixNano()] = unmergedBlock
+					}
+				}
+				if err != nil {
+					workerErrs[workerNum]++
+				}
+			}
+			wg.Done()
+		}(workerNum, encoderChan)
```
prateek: bug: don't use loop variables in go-routine calls, [reference](https://github.com/golang/go/wiki/CommonMistakes)

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -49,8 +50,20 @@ type encoder struct {
 }
 
 type encoderMap struct {
```
prateek: nit: rename this to something series specific, `seriesEncodersByTime` or some such

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Oh I see the issue now. I guess I didn't run into this issue because I was always using 8192 shards in my runs.

Instead of using a map, can I just initialize unmerged to MAX_SHARDS+1? I'm accessing by index, not appending, so then it should do the right thing?

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Are you sure this is wrong? The reference you linked seems to suggest that what I'm doing is correct. If I didn't pass them into the goroutine it would be broken, but since I pass them in I'm creating a new variable that will hold onto the old values even as the for loop continues on.

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: will do

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: will do

glide.yaml
prateek: Cool, only concern was backward compatibility with data written using older version. Not a big deal if it's breaking that, considering we're not reading that data in prod right now anyway.

persist/fs/commitlog/reader.go
prateek: Pretty much. Just add a `BytesPool` in the options (similar to [this](https://github.com/m3db/m3db/blob/master/storage/types.go#L573), initialise with a default in the options ctor like [this](https://github.com/m3db/m3db/blob/master/storage/options.go#L289-L298))

storage/bootstrap/bootstrapper/commitlog/source.go
prateek: Yep, you're right. The way it is works.

persist/fs/commitlog/reader.go
```diff
@@ -57,28 +59,69 @@ type commitLogReader interface {
...
-		bytesPool:       opts.BytesPool(),
-		chunkReader:     newChunkReader(opts.FlushSize()),
-		sizeBuffer:      make([]byte, binary.MaxVarintLen64),
-		logDecoder:      msgpack.NewDecoder(decodingOpts),
-		metadataDecoder: msgpack.NewDecoder(decodingOpts),
-		metadataLookup:  make(map[uint64]Series),
+	cancelCtx, cancelFunc := context.WithCancel(context.Background())
+	decoderBufs := []chan decoderArg{}
+	for i := 0; i < numConc; i++ {
+		decoderBufs = append(decoderBufs, make(chan decoderArg, 100))
```
prateek: nit: use an option or a constant instead of the magic number

glide.yaml
robskillington: We should probably verify by deploying this in an old cluster and upgrading in place.

The raw data blocks are encoded in msgpack too so it might break normal block bootstrapping (outside of commit log bootstrapper). 

This is 100% easily testable though thankfully.

integration/commitlog_bootstrap_helpers.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/m3db/m3x/time"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	defaultIntegrationTestFlushInterval = 10 * time.Millisecond
+)
+
+func computeMetricIndexes(timeBlocks generate.SeriesBlocksByStart) map[string]uint64 {
```
robskillington: We should probably find a slightly more descriptive name.  This could mean a whole bunch of things.

integration/commitlog_bootstrap_helpers.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/m3db/m3db/persist/fs/commitlog"
+	"github.com/m3db/m3db/storage/block"
+	"github.com/m3db/m3db/ts"
+	"github.com/m3db/m3x/time"
+
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	defaultIntegrationTestFlushInterval = 10 * time.Millisecond
```
robskillington: Maybe every 100ms?  10ms probably doesn't allow many  metrics at all to flush together.

integration/commitlog_bootstrap_helpers.go
```diff
@@ -0,0 +1,135 @@
...
+			commitLog = commitlog.NewCommitLog(opts)
+			require.NoError(t, commitLog.Open())
+		}
+
+		// write this point
+		idx, ok := indexes[point.ID.String()]
+		require.True(t, ok)
+		// fmt.Println("ID: ", point.ID)
+		// fmt.Println("Shard: ", shardSet.Lookup(point.ID))
+		// fmt.Println("-------------------------------------")
```
robskillington: Let's pull the commented code out?

integration/commitlog_bootstrap_helpers.go
```diff
@@ -0,0 +1,135 @@
...
+		shardSet  = s.shardSet
+		points    = generate.ToPointsByTime(data) // points are sorted in chronological order
+		blockSize = opts.RetentionOptions().BlockSize()
+		commitLog commitlog.CommitLog
+		now       time.Time
+	)
+
+	closeCommitLogFn := func() {
+		if commitLog != nil {
+			// wait a bit to ensure writes are done, and then close the commit log
```
robskillington: We should use a `sync.WaitGroup` to make this deterministic, we're avoiding test flakeyness by determinism as much as humanly possible these days.

integration/commitlog_bootstrap_helpers.go
```diff
@@ -0,0 +1,135 @@
...
+
+	var (
+		shardSet  = s.shardSet
+		points    = generate.ToPointsByTime(data) // points are sorted in chronological order
+		blockSize = opts.RetentionOptions().BlockSize()
+		commitLog commitlog.CommitLog
+		now       time.Time
+	)
+
+	closeCommitLogFn := func() {
```
robskillington: You can just defer this to avoid explicit call at bottom, i.e.

```go
defer func() {
  // ....
}()
```

persist/fs/commitlog/iterator.go
```diff
@@ -95,6 +95,7 @@ func (i *iterator) Next() bool {
 	var err error
 	i.read.series, i.read.datapoint, i.read.unit, i.read.annotation, err = i.reader.Read()
 	if err == io.EOF {
+		i.reader.Close()
```
robskillington: Good catch, is this only place where we close readers?  Maybe we should audit this code a little for these cases.

integration/commitlog_bootstrap_helpers.go
richardartoul: Woops, that snuck in when I was trying to debug the test suite. I'll rip it out thanks

PR #317, opened 2017-07-18
Add commit log integration test

richardartoul:
This is PR 1 in three PR's that together will improve the performance of the commitlog bootstrapper. This PR just adds the integration tests so that the other two PRs can be merged with confidence.

Adds a commit log integration test. Most of that code was written by prateek, but I made a few small changes to:

1) Remove an unnecessary sleep (the commitlogs Close() method waits for all writes to complete so the sleep was not necessary)
2) Increased defaultIntegrationTestFlushInterval to 100ms based on feedback from Rob
3) Increased the amount of data generated by the integration test such that it was enough (in my experience testing it) to detect concurrency / parallelism issues.
4) Rename computeMetricIndexes to generateUniqueMetricIndexes based on Rob's feedback
5) Replace named function with anonymous defer based on Rob's feedback

### DIFF
--- a/integration/commitlog_bootstrap_helpers.go
+++ b/integration/commitlog_bootstrap_helpers.go
@@ -37,10 +37,10 @@ import (
 )
 
 var (
-	defaultIntegrationTestFlushInterval = 10 * time.Millisecond
+	defaultIntegrationTestFlushInterval = 100 * time.Millisecond
 )
 
-func computeMetricIndexes(timeBlocks generate.SeriesBlocksByStart) map[string]uint64 {
+func generateUniqueMetricIndexes(timeBlocks generate.SeriesBlocksByStart) map[string]uint64 {
 	var idx uint64
 	indexes := make(map[string]uint64)
 	for _, blks := range timeBlocks {
@@ -61,7 +61,7 @@ func writeCommitLog(
 	data generate.SeriesBlocksByStart,
 	namespace ts.ID,
 ) {
-	indexes := computeMetricIndexes(data)
+	indexes := generateUniqueMetricIndexes(data)
 	opts := s.storageOpts.CommitLogOptions()
 
 	// ensure commit log is flushing frequently
@@ -76,14 +76,11 @@ func writeCommitLog(
 		now       time.Time
 	)
 
-	closeCommitLogFn := func() {
+	defer func() {
 		if commitLog != nil {
-			// wait a bit to ensure writes are done, and then close the commit log
-			time.Sleep(10 * defaultIntegrationTestFlushInterval)
 			require.NoError(t, commitLog.Close())
 		}
-
-	}
+	}()
 
 	for _, point := range points {
 		pointTime := point.Timestamp
--- a/integration/commitlog_bootstrap_helpers.go
+++ b/integration/commitlog_bootstrap_helpers.go
@@ -76,11 +76,12 @@ func writeCommitLog(
 		now       time.Time
 	)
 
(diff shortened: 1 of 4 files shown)


integration/commitlog_bootstrap_helpers.go
```diff
@@ -0,0 +1,130 @@
...
+
+	var (
+		shardSet  = s.shardSet
+		points    = generate.ToPointsByTime(data) // points are sorted in chronological order
+		blockSize = opts.RetentionOptions().BlockSize()
+		commitLog commitlog.CommitLog
+		now       time.Time
+	)
+
+	closeCommitLogFn := func() {
```
prateek: nit: just this instead –

```
defer func() {
  if commitLog != nil {
    require.NoError(t, commitLog.Close())
  }
} ()
```

integration/commitlog_bootstrap_helpers.go
```diff
@@ -0,0 +1,130 @@
...
+		cId := commitlog.Series{
+			Namespace:   namespace,
+			Shard:       shardSet.Lookup(point.ID),
+			ID:          point.ID,
+			UniqueIndex: idx,
+		}
+		require.NoError(t, commitLog.WriteBehind(ctx, cId, point.Datapoint, xtime.Second, nil))
+	}
+
+	closeCommitLogFn()
```
prateek: don't need this because you're already deferring above

richardartoul: Woops meant to delete that

integration/commitlog_bootstrap_helpers.go
richardartoul: But its also called in the loop, not just at the bottom of the function. Have to give it a name otherwise it cant be called in the loop

integration/commitlog_bootstrap_helpers.go
```diff
@@ -0,0 +1,128 @@
...
+			require.NoError(t, commitLog.Close())
+		}
+	}
+	defer closeCommitLogFn()
+
+	for _, point := range points {
+		pointTime := point.Timestamp
+
+		// check if this point falls in the current commit log block
+		if truncated := pointTime.Truncate(blockSize); truncated != now {
```
prateek: couple of things: 
- probably easier to iterate over the map which already has the data broken by blockStart ~  `generate.SeriesBlocksByStart` 
- iirc we can write multiple commit log files for a single blockStart under some condition, might be useful to write data split up like that

integration/commitlog_bootstrap_helpers.go
```diff
@@ -0,0 +1,128 @@
...
+	indexes := generateUniqueMetricIndexes(data)
+	opts := s.storageOpts.CommitLogOptions()
+
+	// ensure commit log is flushing frequently
+	require.Equal(t, defaultIntegrationTestFlushInterval, opts.FlushInterval())
+	ctx := context.NewContext()
+
+	var (
+		shardSet  = s.shardSet
+		points    = generate.ToPointsByTime(data) // points are sorted in chronological order
```
robskillington: You may want to write some % of out of order points within a valid time range bucket to exercise the commit log case where we need to combine out of order streams at the end of the commit log bootstrapping process.

prateek: done in #320

prateek:
Can't commandeer because the origin branch isn't off the `m3db` repo, continuing work in #320

PR #318, opened 2017-07-19
Parallelize commitlog reader

richardartoul:
This is PR 2 in three PR's that together will improve the performance of the commitlog bootstrapper. 
This PR parallelizes the commitlog bootstrapper (specifically it parallelizes the msgpack decoding)

### DIFF
--- a/persist/fs/commitlog/commit_log_test.go
+++ b/persist/fs/commitlog/commit_log_test.go
@@ -30,6 +30,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/fortytw2/leaktest"
 	"github.com/m3db/m3db/clock"
 	"github.com/m3db/m3db/context"
 	"github.com/m3db/m3db/persist/fs"
@@ -331,6 +332,9 @@ func TestCommitLogWrite(t *testing.T) {
 }
 
 func TestCommitLogReaderIsNotReusable(t *testing.T) {
+	// Make sure we're not leaking goroutines
+	defer leaktest.CheckTimeout(t, time.Second)()
+
 	opts, scope := newTestOptions(t, overrides{
 		strategy: StrategyWriteWait,
 	})
--- a/persist/fs/commitlog/commit_log_test.go
+++ b/persist/fs/commitlog/commit_log_test.go
@@ -103,8 +103,7 @@ type mockBitSet struct {
 }
 
 func (b *mockBitSet) test(i uint) bool {
-	// return b.indexTestReturn[i]
-	return true
+	return b.indexTestReturn[i]
 }
 func (b *mockBitSet) set(i uint) {}
 func (b *mockBitSet) clearAll()  {}
@@ -251,7 +250,7 @@ func writeCommitLogs(
 
 		// Wait for previous writes to enqueue
 		for getAllWrites() != preWrites+i {
-			time.Sleep(1 * time.Microsecond)
+			time.Sleep(time.Microsecond)
 		}
 
 		wg.Add(1)
@@ -272,7 +271,7 @@ func writeCommitLogs(
 
 	// Wait for all writes to enqueue
 	for getAllWrites() != preWrites+len(writes) {
-		time.Sleep(1 * time.Microsecond)
+		time.Sleep(time.Microsecond)
 	}
 
 	return &wg
@@ -395,9 +394,10 @@ func TestReadCommitLogMissingMetadata(t *testing.T) {
 	iter, err := commitLog.Iter()
 	assert.NoError(t, err)
 	for iter.Next() {
+		assert.NoError(t, iter.Err())
 	}
 	iter.Close()
-	commitLog.Close()
+	assert.NoError(t, commitLog.Close())
 }
 
 func TestCommitLogReaderIsNotReusable(t *testing.T) {
--- a/persist/fs/commitlog/commit_log_test.go
+++ b/persist/fs/commitlog/commit_log_test.go
@@ -380,7 +380,7 @@ func TestReadCommitLogMissingMetadata(t *testing.T) {
 	// Generate fake writes for each of the series
 	writes := []testWrite{}
 	for _, series := range allSeries {
-		for i := 0; i < 100; i++ {
+		for i := 0; i < 10; i++ {
 			writes = append(writes, testWrite{series, time.Now(), rand.Float64(), xtime.Second, []byte{1, 2, 3}, nil})
 		}
 	}
--- a/persist/fs/commitlog/commit_log_test.go
+++ b/persist/fs/commitlog/commit_log_test.go
@@ -330,6 +330,43 @@ func TestCommitLogWrite(t *testing.T) {
 	assertCommitLogWritesByIterating(t, commitLog, writes)
 }
 
+func TestCommitLogReaderIsNotReusable(t *testing.T) {
+	opts, scope := newTestOptions(t, overrides{
+		strategy: StrategyWriteWait,
+	})
+	defer cleanup(t, opts)
+
+	commitLog := newTestCommitLog(t, opts)
+
+	writes := []testWrite{
+		{testSeries(0, "foo.bar", 127), time.Now(), 123.456, xtime.Second, []byte{1, 2, 3}, nil},
+		{testSeries(1, "foo.baz", 150), time.Now(), 456.789, xtime.Second, nil, nil},
+	}
+
+	// Call write sync
+	writeCommitLogs(t, scope, commitLog, writes).Wait()
+
+	// Close the commit log and consequently flush
+	assert.NoError(t, commitLog.Close())
+
+	// Assert writes occurred by reading the commit log
+	assertCommitLogWritesByIterating(t, commitLog, writes)
+
+	// Assert commitlog file exists and retrieve path
+	fsopts := opts.FilesystemOptions()
+	files, err := fs.CommitLogFiles(fs.CommitLogsDirPath(fsopts.FilePathPrefix()))
+	assert.NoError(t, err)
+	assert.Equal(t, 1, len(files))
+
+	// Assert commitlog cannot be opened more than once
+	reader := newCommitLogReader(opts)
+	_, _, _, err = reader.Open(files[0])
+	assert.NoError(t, err)
+	reader.Close()
+	_, _, _, err = reader.Open(files[0])
+	assert.Equal(t, errCommitLogReaderIsNotReusable, err)
+}
+
 func TestCommitLogWriteBehind(t *testing.T) {
 	opts, scope := newTestOptions(t, overrides{
 		strategy: StrategyWriteBehind,
--- a/persist/fs/commitlog/commit_log_test.go
+++ b/persist/fs/commitlog/commit_log_test.go
@@ -366,7 +366,9 @@ func TestReadCommitLogMissingMetadata(t *testing.T) {
 		indexTestReturn: map[uint]bool{},
 	}
 
-	// Generate fake series, where approximately half will be missing metadata
+	// Generate fake series, where approximately half will be missing metadata.
+	// This works because the commitlog writer uses the bitset to determine if
+	// the metadata for a particular series had already been written to disk.
 	allSeries := []Series{}
 	for i := 0; i < 200; i++ {
 		willNotHaveMetadata := !(i%2 == 0)
--- a/persist/fs/commitlog/commit_log_test.go
+++ b/persist/fs/commitlog/commit_log_test.go
@@ -23,6 +23,7 @@ package commitlog
 import (
 	"fmt"
 	"io/ioutil"
+	"math/rand"
 	"os"
 	"strings"
 	"sync"
@@ -95,6 +96,20 @@ func newTestOptions(
 func cleanup(t *testing.T, opts Options) {
 	filePathPrefix := opts.FilesystemOptions().FilePathPrefix()
 	assert.NoError(t, os.RemoveAll(filePathPrefix))
+}
+
+type mockBitSet struct {
+	indexTestReturn map[uint]bool
+}
+
+func (b *mockBitSet) test(i uint) bool {
+	// return b.indexTestReturn[i]
+	return true
+}
+func (b *mockBitSet) set(i uint) {}
+func (b *mockBitSet) clearAll()  {}
+func (b *mockBitSet) getValues() []uint64 {
+	return []uint64{}
 }
 
 type testWrite struct {
@@ -236,7 +251,7 @@ func writeCommitLogs(
 
 		// Wait for previous writes to enqueue
 		for getAllWrites() != preWrites+i {
-			time.Sleep(10 * time.Millisecond)
+			time.Sleep(1 * time.Microsecond)
 		}
 
 		wg.Add(1)
@@ -257,7 +272,7 @@ func writeCommitLogs(
 
 	// Wait for all writes to enqueue
 	for getAllWrites() != preWrites+len(writes) {
-		time.Sleep(10 * time.Millisecond)
+		time.Sleep(1 * time.Microsecond)
 	}
 
 	return &wg
@@ -329,6 +344,58 @@ func TestCommitLogWrite(t *testing.T) {
 
 	// Assert writes occurred by reading the commit log
 	assertCommitLogWritesByIterating(t, commitLog, writes)
+}
+
+func TestReadCommitLogMissingMetadata(t *testing.T) {
+	readConc := 4
+	// Make sure we're not leaking goroutines
+	defer leaktest.CheckTimeout(t, 10*time.Second)()
+
+	opts, scope := newTestOptions(t, overrides{
+		strategy: StrategyWriteWait,
+	})
+	// Set read concurrency so that the parallel path is definitely tested
+	opts.SetReadConcurrency(readConc)
+	defer cleanup(t, opts)
+
+	// Replace bitset in writer with one that configurably returns true or false
+	// depending on the series
+	commitLog := newTestCommitLog(t, opts)
+	writer := commitLog.writer.(*writer)
+	mockBitSet := &mockBitSet{
+		indexTestReturn: map[uint]bool{},
+	}
+
+	// Generate fake series, where approximately half will be missing metadata
+	allSeries := []Series{}
+	for i := 0; i < 200; i++ {
+		willNotHaveMetadata := !(i%2 == 0)
+		allSeries = append(allSeries, testSeries(uint64(i), "hax", uint32(i%100)))
+		mockBitSet.indexTestReturn[uint(i)] = willNotHaveMetadata
+	}
+	writer.seen = mockBitSet
+
+	// Generate fake writes for each of the series
+	writes := []testWrite{}
+	for _, series := range allSeries {
+		for i := 0; i < 100; i++ {
+			writes = append(writes, testWrite{series, time.Now(), rand.Float64(), xtime.Second, []byte{1, 2, 3}, nil})
+		}
+	}
+
+	// Call write sync
+	writeCommitLogs(t, scope, commitLog, writes).Wait()
+
+	// Close the commit log and consequently flush
+	assert.NoError(t, commitLog.Close())
+
+	// Make sure we don't panic / deadlock
+	iter, err := commitLog.Iter()
+	assert.NoError(t, err)
+	for iter.Next() {
+	}
+	iter.Close()
+	commitLog.Close()
 }
 
 func TestCommitLogReaderIsNotReusable(t *testing.T) {
--- a/persist/fs/commitlog/iterator.go
+++ b/persist/fs/commitlog/iterator.go
@@ -95,6 +95,7 @@ func (i *iterator) Next() bool {
 	var err error
 	i.read.series, i.read.datapoint, i.read.unit, i.read.annotation, err = i.reader.Read()
 	if err == io.EOF {
+		i.reader.Close()
 		// Try the next reader
 		i.reader = nil
 		return i.Next()
--- a/persist/fs/commitlog/iterator.go
+++ b/persist/fs/commitlog/iterator.go
@@ -186,10 +186,10 @@ func (i *iterator) nextReader() bool {
 }
 
 func (i *iterator) closeAndResetReader() error {
-	if i.reader != nil {
-		err := i.reader.Close()
-		i.reader = nil
-		return err
+	if i.reader == nil {
+		return nil
 	}
-	return nil
+	reader := i.reader
+	i.reader = nil
+	return reader.Close()
 }
--- a/persist/fs/commitlog/iterator.go
+++ b/persist/fs/commitlog/iterator.go
@@ -95,7 +95,10 @@ func (i *iterator) Next() bool {
 	var err error
 	i.read.series, i.read.datapoint, i.read.unit, i.read.annotation, err = i.reader.Read()
 	if err == io.EOF {
-		i.closeAndResetReader()
+		closeErr := i.closeAndResetReader()
+		if closeErr != nil {
+			i.err = closeErr
+		}
 		// Try the next reader
 		return i.Next()
 	}
@@ -138,7 +141,11 @@ func (i *iterator) nextReader() bool {
 		return false
 	}
 
-	i.closeAndResetReader()
+	err := i.closeAndResetReader()
+	if err != nil {
+		i.err = err
+		return false
+	}
 
 	file := i.files[0]
 	i.files = i.files[1:]
--- a/persist/fs/commitlog/iterator.go
+++ b/persist/fs/commitlog/iterator.go
@@ -172,9 +172,11 @@ func (i *iterator) nextReader() bool {
 	return true
 }
 
-func (i *iterator) closeAndResetReader() {
+func (i *iterator) closeAndResetReader() error {
 	if i.reader != nil {
-		i.reader.Close()
+		err := i.reader.Close()
 		i.reader = nil
+		return err
 	}
+	return nil
 }
--- a/persist/fs/commitlog/iterator.go
+++ b/persist/fs/commitlog/iterator.go
@@ -129,6 +129,7 @@ func (i *iterator) Err() error {
 	return i.err
 }
 
+// TODO: Refactor codebase so that it can handle Close() returning an error
 func (i *iterator) Close() {
 	if i.closed {
 		return
--- a/persist/fs/commitlog/iterator.go
+++ b/persist/fs/commitlog/iterator.go
@@ -95,9 +95,8 @@ func (i *iterator) Next() bool {
 	var err error
 	i.read.series, i.read.datapoint, i.read.unit, i.read.annotation, err = i.reader.Read()
 	if err == io.EOF {
-		i.reader.Close()
+		i.closeAndResetReader()
 		// Try the next reader
-		i.reader = nil
 		return i.Next()
 	}
 	if err != nil {
@@ -127,10 +126,7 @@ func (i *iterator) Close() {
 		return
 	}
 	i.closed = true
-	if i.reader != nil {
-		i.reader.Close()
-		i.reader = nil
-	}
+	i.closeAndResetReader()
 }
 
 func (i *iterator) hasError() bool {
@@ -142,10 +138,7 @@ func (i *iterator) nextReader() bool {
 		return false
 	}
 
-	if i.reader != nil {
-		i.reader.Close()
-		i.reader = nil
-	}
+	i.closeAndResetReader()
 
 	file := i.files[0]
 	i.files = i.files[1:]
@@ -177,4 +170,11 @@ func (i *iterator) nextReader() bool {
 
 	i.reader = reader
 	return true
+}
+
+func (i *iterator) closeAndResetReader() {
+	if i.reader != nil {
+		i.reader.Close()
+		i.reader = nil
+	}
 }
--- a/persist/fs/commitlog/iterator.go
+++ b/persist/fs/commitlog/iterator.go
@@ -106,6 +106,11 @@ func (i *iterator) Next() bool {
 		// Try the next reader, this enables restoring with best effort from commit logs
 		i.metrics.readsErrors.Inc(1)
 		i.log.Errorf("commit log reader returned error, iterator moving to next file: %v", err)
+		i.err = err
+		closeErr := i.closeAndResetReader()
+		if closeErr != nil {
+			i.err = closeErr
+		}
 		return i.Next()
 	}
 	i.setRead = true
--- a/persist/fs/commitlog/options.go
+++ b/persist/fs/commitlog/options.go
@@ -49,6 +49,9 @@ const (
 
 	// defaultReadConcurrency is the default read concurrency
 	defaultReadConcurrency = 4
+
+	// defaultMetadataTimeout is the default metadata timeout
+	defaultMetadataTimeout = 100 * time.Millisecond
 )
 
 var (
@@ -76,6 +79,7 @@ type options struct {
 	backlogQueueSize int
 	bytesPool        pool.CheckedBytesPool
 	readConcurrency  int
+	metadataTimeout  time.Duration
 }
 
 // NewOptions creates new commit log options
@@ -94,6 +98,7 @@ func NewOptions() Options {
 			return pool.NewBytesPool(s, nil)
 		}),
 		readConcurrency: defaultReadConcurrency,
+		metadataTimeout: defaultMetadataTimeout,
 	}
 	o.bytesPool.Init()
 	return o
@@ -226,4 +231,14 @@ func (o *options) SetReadConcurrency(concurrency int) Options {
 
 func (o *options) ReadConcurrency() int {
 	return o.readConcurrency
+}
+
+func (o *options) SetMetadataTimeout(timeout time.Duration) Options {
+	opts := *o
+	opts.metadataTimeout = timeout
+	return &opts
+}
+
+func (o *options) MetadataTimeout() time.Duration {
+	return o.metadataTimeout
 }
--- a/persist/fs/commitlog/options.go
+++ b/persist/fs/commitlog/options.go
@@ -49,9 +49,6 @@ const (
 
 	// defaultReadConcurrency is the default read concurrency
 	defaultReadConcurrency = 4
-
-	// defaultMetadataTimeout is the default metadata timeout
-	defaultMetadataTimeout = 100 * time.Millisecond
 )
 
 var (
@@ -79,7 +76,6 @@ type options struct {
 	backlogQueueSize int
 	bytesPool        pool.CheckedBytesPool
 	readConcurrency  int
-	metadataTimeout  time.Duration
 }
 
 // NewOptions creates new commit log options
@@ -98,7 +94,6 @@ func NewOptions() Options {
 			return pool.NewBytesPool(s, nil)
 		}),
 		readConcurrency: defaultReadConcurrency,
-		metadataTimeout: defaultMetadataTimeout,
 	}
 	o.bytesPool.Init()
 	return o
@@ -231,14 +226,4 @@ func (o *options) SetReadConcurrency(concurrency int) Options {
 
 func (o *options) ReadConcurrency() int {
 	return o.readConcurrency
-}
-
-func (o *options) SetMetadataTimeout(timeout time.Duration) Options {
-	opts := *o
-	opts.metadataTimeout = timeout
-	return &opts
-}
-
-func (o *options) MetadataTimeout() time.Duration {
-	return o.metadataTimeout
 }
--- a/persist/fs/commitlog/options.go
+++ b/persist/fs/commitlog/options.go
@@ -46,6 +46,9 @@ const (
 
 	// defaultBlockSize is the default commit log block size
 	defaultBlockSize = 15 * time.Minute
+
+	// defaultReadConcurrency is the default read concurrency
+	defaultReadConcurrency = 4
 )
 
 var (
@@ -71,6 +74,7 @@ type options struct {
 	flushInterval    time.Duration
 	backlogQueueSize int
 	bytesPool        pool.CheckedBytesPool
+	readConcurrency  int
 }
 
 // NewOptions creates new commit log options
@@ -88,6 +92,7 @@ func NewOptions() Options {
 		bytesPool: pool.NewCheckedBytesPool(nil, nil, func(s []pool.Bucket) pool.BytesPool {
 			return pool.NewBytesPool(s, nil)
 		}),
+		readConcurrency: defaultReadConcurrency,
 	}
 	o.bytesPool.Init()
 	return o
@@ -207,4 +212,14 @@ func (o *options) SetBytesPool(value pool.CheckedBytesPool) Options {
 
 func (o *options) BytesPool() pool.CheckedBytesPool {
 	return o.bytesPool
+}
+
+func (o *options) SetReadConcurrency(concurrency int) Options {
+	opts := *o
+	opts.readConcurrency = concurrency
+	return &opts
+}
+
+func (o *options) ReadConcurrency() int {
+	return o.readConcurrency
 }
--- a/persist/fs/commitlog/options.go
+++ b/persist/fs/commitlog/options.go
@@ -61,6 +61,7 @@ var (
 	errBlockSizePositive              = errors.New("block size must be a positive duration")
 	errRetentionPeriodPositive        = errors.New("retention period must be a positive duration")
 	errRetentionGreaterEqualBlockSize = errors.New("retention period must be >= block size")
+	errReadConcurrencyPositive        = errors.New("read concurrency must be a positive integer")
 )
 
 type options struct {
@@ -110,6 +111,9 @@ func (o *options) Validate() error {
 	}
 	if o.RetentionPeriod() < o.BlockSize() {
 		return errRetentionGreaterEqualBlockSize
+	}
+	if o.ReadConcurrency() <= 0 {
+		return errReadConcurrencyPositive
 	}
 	return nil
 }
--- a/persist/fs/commitlog/reader.go
+++ b/persist/fs/commitlog/reader.go
@@ -51,65 +51,65 @@ var (
 	errCommitLogReaderMultipleReadloops         = errors.New("commitlog reader tried to open multiple readLoops, do not call Read() concurrently")
 )
 
-// type waitGroupWithTimeout struct {
-// 	sync.Mutex
-// 	waiters   []chan bool
-// 	doneCh    chan struct{}
-// 	timeoutCh <-chan time.Time
-// 	completed bool
-// 	timedout  bool
-// }
+type waitGroupWithTimeout struct {
+	sync.Mutex
+	waiters   []chan bool
+	doneCh    chan struct{}
+	timeoutCh <-chan time.Time
+	completed bool
+	timedout  bool
+}
 
-// func (w *waitGroupWithTimeout) Wait() (timedout bool) {
-// 	w.Lock()
+func (w *waitGroupWithTimeout) Wait() (timedout bool) {
+	w.Lock()
 
-// 	if w.completed {
-// 		w.Unlock()
-// 		return w.timedout
-// 	}
+	// Done() has already been called, just return
+	if w.completed {
+		w.Unlock()
+		return w.timedout
+	}
 
-// 	// If this is the first call to Wait(), setup the state
-// 	if len(w.waiters) == 0 {
-// 		// Timer starts after first call to Wait()
-// 		w.timeoutCh = time.After(5 * time.Second)
-// 		w.doneCh = make(chan struct{})
+	// If this is the first call to Wait(), setup the state
+	if len(w.waiters) == 0 {
+		// Timer starts after first call to Wait()
+		w.timeoutCh = time.After(5 * time.Second)
+		w.doneCh = make(chan struct{})
 
-// 		go func() {
-// 			select {
-// 			case <-w.doneCh:
-// 				w.timedout = false
-// 				for _, waiterCh := range w.waiters {
-// 					waiterCh <- false
-// 				}
-// 			case <-w.timeoutCh:
-// 				w.timedout = true
-// 				for _, waiterCh := range w.waiters {
-// 					waiterCh <- true
-// 				}
-// 			}
-// 		}()
-// 	}
+		go func() {
+			select {
+			case <-w.doneCh:
+				w.timedout = false
+				for _, waiterCh := range w.waiters {
+					waiterCh <- false
+				}
+			case <-w.timeoutCh:
+				w.timedout = true
+				for _, waiterCh := range w.waiters {
+					waiterCh <- true
+				}
+			}
+		}()
+	}
 
-// 	waitCh := make(chan bool)
-// 	w.waiters = append(w.waiters, waitCh)
-// 	w.Unlock()
-// 	timedout = <-waitCh
-// 	return timedout
-// }
+	waitCh := make(chan bool)
+	w.waiters = append(w.waiters, waitCh)
+	w.Unlock()
+	return <-waitCh
+}
 
-// func (w *waitGroupWithTimeout) Done() {
-// 	w.Lock()
-// 	defer w.Unlock()
+func (w *waitGroupWithTimeout) Done() {
+	w.Lock()
+	defer w.Unlock()
 
-// 	w.completed = true
-// 	// No one ever called Wait()
-// 	if len(w.waiters) == 0 {
-// 		w.timedout = false
-// 		return
-// 	}
+	w.completed = true
+	// No one ever called Wait()
+	if len(w.waiters) == 0 {
+		w.timedout = false
+		return
+	}
 
-// 	w.doneCh <- struct{}{}
-// }
+	w.doneCh <- struct{}{}
+}
 
 type commitLogReader interface {
 	// Open opens the commit log for reading
--- a/persist/fs/commitlog/reader.go
+++ b/persist/fs/commitlog/reader.go
@@ -30,13 +30,12 @@ import (
 	"sync"
 	"time"
 
-	"github.com/m3db/m3x/checked"
-
 	"github.com/m3db/m3db/digest"
 	"github.com/m3db/m3db/persist/encoding"
 	"github.com/m3db/m3db/persist/encoding/msgpack"
 	"github.com/m3db/m3db/persist/schema"
 	"github.com/m3db/m3db/ts"
+	"github.com/m3db/m3x/checked"
 	"github.com/m3db/m3x/pool"
 	xtime "github.com/m3db/m3x/time"
 )
@@ -367,8 +366,7 @@ func (r *reader) Close() error {
 	// and close the fd
 	r.cancelFunc()
 
-	<-r.shutdownChan
-	return nil
+	return <-r.shutdownChan
 }
 
 func (r *reader) close() error {
--- a/persist/fs/commitlog/reader.go
+++ b/persist/fs/commitlog/reader.go
@@ -442,20 +442,6 @@ func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan<- readResponse
 	}
 }
 
-func (r *reader) waitWithTimeout(wg *sync.WaitGroup, timeout time.Duration) (timedout bool) {
-	c := make(chan struct{})
-	go func() {
-		wg.Wait()
-		close(c)
-	}()
-	select {
-	case <-c:
-		return false
-	case <-time.After(timeout):
-		return true
-	}
-}
-
 func (r *reader) readChunk() (checked.Bytes, error) {
 	// Read size of message
 	size, err := binary.ReadUvarint(r.chunkReader)
--- a/persist/fs/commitlog/reader.go
+++ b/persist/fs/commitlog/reader.go
@@ -46,6 +46,7 @@ var (
 
 	errCommitLogReaderAlreadyOpen               = errors.New("commit log reader already open")
 	errCommitLogReaderChunkSizeChecksumMismatch = errors.New("commit log reader encountered chunk size checksum mismatch")
+	errCommitLogReaderChunkDataChecksumMismatch = errors.New("commit log reader encountered chunk data checksum mismatch")
 	errCommitLogReaderMissingLogMetadata        = errors.New("commit log reader encountered message missing metadata")
 )
 
@@ -423,6 +424,7 @@ func (r *chunkReader) Read(p []byte) (int, error) {
 
 		// Reset read target
 		p = p[read:]
+		size = len(p)
 
 		// Perform consecutive read(s)
 		n, err := r.Read(p)
--- a/persist/fs/commitlog/reader.go
+++ b/persist/fs/commitlog/reader.go
@@ -171,7 +171,17 @@ func (r *reader) Read() (
 		}
 	}
 	// Data is written into these channels in round-robin fashion, so reading them
-	// one at a time in the same order results in an ordered stream.
+	// one at a time in the same order results in an ordered stream. I.E with a
+	// concurrency of two then we'll start with an initial structure like this:
+	// 		r.outBufs:
+	// 			[0]: []
+	// 			[1]: []
+	// After several reads from the commitlog we end up with:
+	// 		r.outBufs:
+	// 			[0]: [a, c]
+	// 			[1]: [b, d]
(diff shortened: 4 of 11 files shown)


persist/fs/commitlog/iterator.go
```diff
@@ -95,6 +95,7 @@ func (i *iterator) Next() bool {
 	var err error
 	i.read.series, i.read.datapoint, i.read.unit, i.read.annotation, err = i.reader.Read()
 	if err == io.EOF {
+		i.reader.Close()
 		// Try the next reader
 		i.reader = nil
```
robskillington: We should probably add a method that encapsulates `i.reader = nil; i.reader.Close()` as it's two steps performed often in this file.

robskillington: It could also encapsulate the `if i.reader != nil { i.reader.Close(); i.reader = nil }` logic as that's also duplicative.

persist/fs/commitlog/reader.go
```diff
@@ -110,52 +157,132 @@ func (r *reader) Read() (
...
+			r.metadataLookupLock.Unlock()
+
+			if ok {
+				break
+			}
+
+			// Race condition where the metadata is currently being processed by
+			// another decoderLoop goroutine, but that routine has not updated the
+			// metadataLookup map yet.
+			if !ok && numMetadataMiss < 3 {
```
robskillington: Hm this smells a little, metadata should be written once and only once in an entire commit log.  Did you run into this?  Also sleeping/timeout & backoffs does not provide strong deterministic behavior that we will need.

persist/fs/commitlog/reader.go
```diff
@@ -110,52 +157,132 @@ func (r *reader) Read() (
...
-			ID:          ts.BinaryID(id),
-			Namespace:   ts.BinaryID(namespace),
-			Shard:       uint32(decoded.Shard),
+func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan *readResponse) {
+	decodingOpts := r.opts.FilesystemOptions().DecodingOptions()
+	decoder := msgpack.NewDecoder(decodingOpts)
+	metadataDecoder := msgpack.NewDecoder(decodingOpts)
+
+	for arg := range inBuf {
+		readResponse := &readResponse{}
```
robskillington: I think it would be best to make responses simply a `chan readResponse` rather than a `chan *readResponse` so we can avoid having to allocate these on the heap and simply pass by value.

persist/fs/commitlog/reader.go
```diff
@@ -165,16 +292,11 @@ func (r *reader) readChunk() ([]byte, error) {
...
 
-	// Extend buffer as necessary
-	if len(r.dataBuffer) < int(size) {
-		diff := int(size) - len(r.dataBuffer)
-		for i := 0; i < diff; i++ {
-			r.dataBuffer = append(r.dataBuffer, 0)
-		}
-	}
+	// TODO: Pool this
+	newBuf := make([]byte, size, size)
```
robskillington: Yeah we'll definitely have to pool or else the GC pressure will be intense.

persist/fs/commitlog/reader.go
```diff
@@ -110,52 +157,132 @@ func (r *reader) Read() (
...
+			go r.decoderLoop(decoderBuf, r.outBufs[i])
+		}
 	}
+	// Data is written into these channels in round-robin fashion, so reading them
+	// one at a time in the same order results in an ordered stream.
+	// TODO: This doesn't eventually deadlock because the reader goroutine will
+	// continue to fill these channels with EOFs until it is closed. We should
+	// probably avoid the deadlock (read from empty channel) scenario in a more
+	// graceful way.
+	rr := <-r.outBufs[r.nextIndex%r.numConc]
```
robskillington: Yeah we should `rr, ok := <- ....` then `if !ok { return nil, ts.Datapoint{}, xtime.Unit(0), ts.Annotation(nil), errUnexpectedEOF }` or something.

persist/fs/commitlog/reader.go
```diff
@@ -110,52 +157,132 @@ func (r *reader) Read() (
...
-		namespace.IncRef()
-		defer namespace.DecRef()
-		namespace.AppendAll(decoded.Namespace)
-
-		r.metadataLookup[entry.Index] = Series{
-			UniqueIndex: entry.Index,
-			ID:          ts.BinaryID(id),
-			Namespace:   ts.BinaryID(namespace),
-			Shard:       uint32(decoded.Shard),
+func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan *readResponse) {
```
robskillington: Can make the `outBuf` also take the required channel direction `outBuf chan<- ...`

persist/fs/commitlog/iterator.go
richardartoul: Sounds good

persist/fs/commitlog/reader.go
richardartoul: Correct the issue was not that metadata was written multiple times, the issue was the chunk containing the metadata and the chunks afterwards (which required the metadata but did not have it) were being processed in parallel and I needed the metadata chunk to be processed before the others could continue.

Agreed it should be deterministic, couldn't figure out how to do it and keep it fast but your solution looks good. For canceling any inflight pendingResult waitgroups, are you sure we need to do that? Seems like we can just allow them to finish. If a decoderloop is blocked on a pendingResult.Wait() then it means the metadata chunk must be being processed by a different decoder loop somewhere, so the call will eventually return and the decoderLoop will clean itself up because closing the reader causes the channel that the decoderLoop is looping through to close

persist/fs/commitlog/reader.go
richardartoul: will do

persist/fs/commitlog/reader.go
richardartoul: I forgot you could do that with channels, good call

persist/fs/commitlog/reader.go
richardartoul: I think originally I had this as a chan readResponse and I changed it after profiling showed a lot of time spent coping data through this channel and the reader became faster after I changed it as well. Its possible I'm thinking of a different channel in my head and I can try benchmarking it again

richardartoul: I could be wrong but I would imagine anything you pass through a channel has to be allocated on the heap because it has to be available after that function returns

persist/fs/commitlog/reader.go
richardartoul: The old code reused a single buffer over and over again (because there was no parallelism). I changed it to reuse a single buffer per "worker". I think that should be more efficient than pooling

persist/fs/commitlog/reader.go
richardartoul: Ok that makes a lot of sense actually, let me test it

persist/fs/commitlog/reader.go
```diff
@@ -22,11 +22,16 @@ package commitlog
...
+	"context"
 	"encoding/binary"
 	"errors"
+	"io"
 	"os"
+	"sync"
 	"time"
 
+	"github.com/m3db/m3x/checked"
+
```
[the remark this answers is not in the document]
richardartoul: will fix

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +159,166 @@ func (r *reader) Read() (
...
 	resultErr error,
 ) {
-	entry, err := r.readEntry()
-	if err != nil {
-		resultErr = err
-		return
+	if r.nextIndex == 0 {
+		go r.readLoop()
+		for i, decoderBuf := range r.decoderBufs {
+			go r.decoderLoop(decoderBuf, r.outBufs[i])
```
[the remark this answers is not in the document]
richardartoul: Are you sure this is wrong? I'm pretty sure that by invoking r.decoderLoop() I'm capturing the arguments and making a copy.

Example go playground: https://play.golang.org/p/nH8Vc0axGY

persist/fs/commitlog/reader.go
```diff
@@ -182,36 +340,45 @@ func (r *reader) readChunk() ([]byte, error) {
...
-	}
-	r.logDecoder.Reset(data)
-	return r.logDecoder.DecodeLogEntry()
+func (r *reader) Close() error {
+	// Shutdown the readLoop goroutine which will shut down the decoderLoops
+	// and close the fd
+	r.cancelFunc()
+
+	<-r.shutdownChan
+	return nil
```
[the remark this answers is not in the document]
richardartoul: Good catch, I changed this temporarily while I added some profiling code and forgot to put it back properly when I was done

persist/fs/commitlog/reader.go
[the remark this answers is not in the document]
robskillington: Ah this is bound before the routine runs, yes this should be just fine.  Apologies, didn't realize that binding occurs for `go <statement>` before the routine executes `<statement>`.

We use worker pools mostly everywhere so we always pre-declare for the lambda to capture, here though a pool is not necessary as there is a firm bound on concurrency.  Disregard, I'll delete this comment thread.

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +158,166 @@ func (r *reader) Read() (
...
+		var hasMetadata bool
+		var ok bool
+
+		// Try cheap read lock first
+		r.metadataLock.RLock()
+		metadata, hasMetadata = r.metadataLookup[entry.Index]
+		r.metadataLock.RUnlock()
+
+		// The required metadata hasn't been processed yet
+		r.metadataLock.Lock()
```
robskillington: You can move this write lock one level deeper to avoid a write lock if the first read lock succeeds, e.g.:
```go
rwmutex.RLock()
value, ok := r.lookup[key]
rwmutex.RUnlock()

if !ok {
  rwmutex.Lock()
  value, ok = r.lookup[key]
  if !ok {
    // insert
  }
  rwmutex.Unlock()
}
```

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +158,166 @@ func (r *reader) Read() (
...
+			}
+		}
+		r.metadataLock.Unlock()
+
+		// The required metadata hasn't been processed yet
+		if pendingResult != nil {
+			pendingResult.Wait()
+			r.metadataLock.Lock()
+			metadata, ok = r.metadataLookup[entry.Index]
+			r.metadataLock.Unlock()
```
robskillington: This can just be `RLock()` and `RUnlock()` right?

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +158,166 @@ func (r *reader) Read() (
...
-		Timestamp: time.Unix(0, entry.Timestamp),
-		Value:     entry.Value,
+		readResponse.series = metadata
+		readResponse.datapoint = ts.Datapoint{
+			Timestamp: time.Unix(0, entry.Timestamp),
+			Value:     entry.Value,
+		}
+		readResponse.unit = xtime.Unit(byte(entry.Unit))
+		// Copy annotation to prevent reference to pooled byte slice
+		readResponse.annotation = append(readResponse.annotation, entry.Annotation...)
```
robskillington: I would only perform this if there's any contents in the annotation, otherwise it might cause an alloc (although granted I haven't tested this).  Either way the behavior is probably undefined and may change between Go versions, etc.

i.e.
```go
if len(entry.Annotation) > 0 {
  readResponse.annotation = append([]byte(nil), entry.Annotation...)
}
```

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +158,166 @@ func (r *reader) Read() (
...
+		// DecRef delayed until this point because even after decoding, underlying
+		// byte slice is still referenced by metadata and annotation
+		arg.bytes.DecRef()
```
robskillington: It seems that we copy the annotation and we also copy the namespace and ID for metadata so we should actually be able to `arg.bytes.Finalize()` here?

persist/fs/commitlog/reader.go
```diff
@@ -182,36 +339,44 @@ func (r *reader) readChunk() ([]byte, error) {
...
 		return nil, err
 	}
-	return buffer, nil
+
+	// Copy message into pooled byte slice. Reading happens synchronously so its
+	// safe to reuse the same buffer repeatedly, but decoding happens in parallel
+	// so we need to make sure that the byte slice we return won't be mutated
+	// until the caller is done with it.
+	pooled := r.bytesPool.Get(int(size))
+	pooled.IncRef()
```
robskillington: Instead of allowing the caller to DecRef, it's better practice to:
```go
b.IncRef()
b.AppendAll(...)
b.DecRef()
```

And let the consumer then Inc and Dec during use.

persist/fs/commitlog/reader.go
```diff
@@ -182,36 +339,44 @@ func (r *reader) readChunk() ([]byte, error) {
...
 func (r *reader) readInfo() (schema.LogInfo, error) {
 	data, err := r.readChunk()
 	if err != nil {
 		return emptyLogInfo, err
 	}
-	r.logDecoder.Reset(data)
-	return r.logDecoder.DecodeLogInfo()
+	r.logDecoder.Reset(data.Get())
+	logInfo, err := r.logDecoder.DecodeLogInfo()
+	data.DecRef()
```
robskillington: This seems to leak the `checked.Bytes` here, need to `data.Finalize()` I believe.

persist/fs/commitlog/reader.go
```diff
@@ -56,28 +61,71 @@ type commitLogReader interface {
...
+		decoderBufs:        decoderBufs,
+		outBufs:            outBufs,
+		dataBuffer:         []byte{},
+		cancelCtx:          cancelCtx,
+		cancelFunc:         cancelFunc,
+		shutdownChan:       make(chan error),
+		metadataDecoder:    msgpack.NewDecoder(decodingOpts),
+		metadataLookup:     make(map[uint64]Series),
+		metadataWaitGroups: make(map[uint64]*sync.WaitGroup),
+		metadataLock:       sync.RWMutex{},
```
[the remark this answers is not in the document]
richardartoul: good idea

persist/fs/commitlog/reader.go
richardartoul: oh duh, thanks

persist/fs/commitlog/reader.go
richardartoul: yep good call

persist/fs/commitlog/reader.go
richardartoul: good idea

persist/fs/commitlog/reader.go
richardartoul: I wasn't clear about this. Does calling Finalize() mean "I'm done with this, allow it to be re-used"? If so, why do I need to call IncRef() and DecRef() manually?

persist/fs/commitlog/reader.go
richardartoul: Yeah I think this is falling out of me not understanding this paradigm properly

persist/fs/commitlog/reader.go
```diff
@@ -193,6 +198,9 @@ func (r *reader) readLoop() {
 			if err == io.EOF {
 				eofFound = true
 			}
+			if err == nil {
+				data.IncRef()
```
robskillington: I would let the consumer IncRef, the Inc/Dec is mainly to signify ownership.  This method is passing ownership of `data` to someone else so it should not Inc and leave it `+1` as it's ownership has finished when it puts it onto a decoder chan.

persist/fs/commitlog/reader.go
```diff
@@ -243,19 +252,19 @@ func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan<- *readRespons
...
 				UniqueIndex: entry.Index,
 				ID:          ts.BinaryID(id),
 				Namespace:   ts.BinaryID(namespace),
 				Shard:       decoded.Shard,
 			}
-			waitGroup, ok := r.metadataWaitGroups[entry.Index]
+			waitGroup, ok := r.metadata.wgs[entry.Index]
 			// Another goroutine is blocked waiting for this metadata
 			if ok {
 				waitGroup.Done()
```
robskillington: You probably want to wait outside holding the lock.  In fact this seems like a deadlock scenario, do we have a test for this branch?

How can you wait for the metadata to be available in the map while you're holding the lock if one can only put it in the map by acquiring the lock?

persist/fs/commitlog/reader.go
robskillington: Calling `Finalize()` will return it to the underlying pool.  Inc/Dec is to signify ownership.  When `Finalize()` is called it will make sure the Inc/Dec is zero (no one owns it any longer) or panic.

persist/fs/commitlog/reader.go
richardartoul: Ok cool that makes sense

persist/fs/commitlog/reader.go
richardartoul: You were right about this. I didn't generate flamegraphs, but I benchmarked for throughput and changed this to pass a struct instead of a pointer didn't seem to have any affect on the readers performance.

richardartoul: I'll update the code

persist/fs/commitlog/reader.go
richardartoul: I'm a little confused here,  waitgroup.Done() notifies all the goroutines that are waiting for this metadata that its now available, this part of the code is not waiting for the metadata to be available in the map, its the opposite its puting the metadata in the map and notifying everyone else that the data is available.

the call to waitGroup.Wait() is called on 301 and that method is invoked while the worker is not holding any locks.

persist/fs/commitlog/reader.go
```diff
@@ -56,28 +61,76 @@ type commitLogReader interface {
...
+		bytesPool:    opts.BytesPool(),
+		chunkReader:  newChunkReader(opts.FlushSize()),
+		logDecoder:   msgpack.NewDecoder(decodingOpts),
+		decoderBufs:  decoderBufs,
+		outBufs:      outBufs,
+		cancelCtx:    cancelCtx,
+		cancelFunc:   cancelFunc,
+		shutdownChan: make(chan error),
+		metadata: readerMetadata{
+			RWMutex: sync.RWMutex{},
```
robskillington: nit: No need for this line, no fields are being set.  For consistency with our other code just leave this declaration out, RWMutex will be initialized with default/zeroed values and will be just fine without this declaration.

persist/fs/commitlog/reader.go
```diff
@@ -182,38 +349,49 @@ func (r *reader) readChunk() ([]byte, error) {
...
+func (r *reader) Close() error {
+	// Shutdown the readLoop goroutine which will shut down the decoderLoops
+	// and close the fd
+	r.cancelFunc()
+
+	return <-r.shutdownChan
 }
 
-func (r *reader) Close() error {
+func (r *reader) close() error {
```
robskillington: Just briefly, is there a test that tests reusing the reader (i.e. Open .. then Close then Open .. Close again)?  Just wanting to make sure we're testing the lifecycle now that we have more complex inner statefulness.

richardartoul: I'll add a unit test for this

persist/fs/commitlog/reader.go
richardartoul: sure thing

richardartoul:
@robskillington I made the reader explicitly un-reusable and added an associated unit test. Let me know if there is anything else that needs to change before I land

storage/bootstrap/bootstrapper/commitlog/types.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 // Options represents the options for bootstrapping from commit logs
-type Options interface {
+type 		Options interface {
```
prateek: nit: remove whitespace

persist/fs/commitlog/reader.go
prateek: Maybe add a e2e unittest (mini-integration test) with the leak detection turned on to catch issues like this

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
+		arg.bytes.Finalize()
+		outBuf <- readResponse
 	}
-	unit = xtime.Unit(byte(entry.Unit))
-	annotation = entry.Annotation
-	return
 }
 
-func (r *reader) readChunk() ([]byte, error) {
+// Caller is responsible for calling DecRef()
```
prateek: Looks like you're calling DecRef at the end of the this, so Callers would have to call both IncRef/DecRef

persist/fs/commitlog/reader.go
richardartoul: Do you mean the race detector? or is there a memory leak detector I can turn on or something?

persist/fs/commitlog/reader.go
richardartoul: ah thats an old comment. I'm just gonna remove it

persist/fs/commitlog/iterator.go
```diff
@@ -177,3 +171,10 @@ func (i *iterator) nextReader() bool {
 	i.reader = reader
 	return true
 }
+
+func (i *iterator) closeAndResetReader() {
```
prateek: `i.reader.Close()` can return an error, could you bubble that up here

persist/fs/commitlog/commit_log_test.go
```diff
@@ -330,6 +330,43 @@ func TestCommitLogWrite(t *testing.T) {
...
+	// Assert writes occurred by reading the commit log
+	assertCommitLogWritesByIterating(t, commitLog, writes)
+
+	// Assert commitlog file exists and retrieve path
+	fsopts := opts.FilesystemOptions()
+	files, err := fs.CommitLogFiles(fs.CommitLogsDirPath(fsopts.FilePathPrefix()))
+	assert.NoError(t, err)
+	assert.Equal(t, 1, len(files))
+
+	// Assert commitlog cannot be opened more than once
```
prateek: maybe break the following lines into a separate unit test for better isolation

persist/fs/commitlog/reader.go
```diff
@@ -22,27 +22,32 @@ package commitlog
...
 	"github.com/m3db/m3db/persist/encoding"
 	"github.com/m3db/m3db/persist/encoding/msgpack"
 	"github.com/m3db/m3db/persist/schema"
 	"github.com/m3db/m3db/ts"
+	"github.com/m3db/m3x/checked"
 	"github.com/m3db/m3x/pool"
 	xtime "github.com/m3db/m3x/time"
 )
 
+const decoderInBufChanSize = 100
```
prateek: Would it be better to make this an `Options` parameter?

persist/fs/commitlog/commit_log_test.go
richardartoul: We already have a unit test for all the previous lines, but I needed to setup the files for bootstrapping and bootstrap them once before I could test trying to bootstrap them again

persist/fs/commitlog/iterator.go
richardartoul: yeah I can return it from this function, but it looked like the code was intentionally swallowing the errors before. Should I make this return the error and have the caller continue to ignore it for now?

persist/fs/commitlog/options.go
```diff
@@ -88,6 +92,7 @@ func NewOptions() Options {
 		bytesPool: pool.NewCheckedBytesPool(nil, nil, func(s []pool.Bucket) pool.BytesPool {
 			return pool.NewBytesPool(s, nil)
 		}),
+		readConcurrency: defaultReadConcurrency,
```
prateek: Could you add a check in `Validate()` to ensure this is `>0`

persist/fs/commitlog/reader.go
```diff
@@ -56,34 +61,84 @@ type commitLogReader interface {
...
-		bytesPool:       opts.BytesPool(),
-		chunkReader:     newChunkReader(opts.FlushSize()),
-		sizeBuffer:      make([]byte, binary.MaxVarintLen64),
-		logDecoder:      msgpack.NewDecoder(decodingOpts),
-		metadataDecoder: msgpack.NewDecoder(decodingOpts),
-		metadataLookup:  make(map[uint64]Series),
+	cancelCtx, cancelFunc := context.WithCancel(context.Background())
+
+	numConc := opts.ReadConcurrency()
+	decoderBufs := []chan decoderArg{}
```
prateek: nit: you can pre-allocate here

```
decoderBufs := make([]chan decoderArg, 0, numCon)
```

persist/fs/commitlog/reader.go
```diff
@@ -56,34 +61,84 @@ type commitLogReader interface {
...
+	for i := 0; i < numConc; i++ {
+		decoderBufs = append(decoderBufs, make(chan decoderArg, decoderInBufChanSize))
 	}
+	outBufs := []chan readResponse{}
```
prateek: same here, can pre-allocate

persist/fs/commitlog/options.go
richardartoul: Oh I didn't realize ya'll had a validate function for the options like that, yeah that sounds like a good idea.

persist/fs/commitlog/reader.go
```diff
@@ -56,34 +61,84 @@ type commitLogReader interface {
...
+	numConc       int
+	bytesPool     pool.CheckedBytesPool
+	chunkReader   *chunkReader
+	dataBuffer    []byte
+	logDecoder    encoding.Decoder
+	decoderBufs   []chan decoderArg
+	outBufs       []chan readResponse
+	cancelCtx     context.Context
+	cancelFunc    context.CancelFunc
+	shutdownChan  chan error
```
prateek: nit: I think we use `Ch` instead of `Chan` everywhere else in the codebase

persist/fs/commitlog/reader.go
richardartoul: Honestly I'm not sure. I just kind of made up a number that seemed large enough. Changing it shouldn't have too much of an effect on things since all the workers should be doing almost exactly the same amount of work and it seems a little too fine-grained for people to be tuning manually

persist/fs/commitlog/reader.go
richardartoul: I usually try to avoid doing this when its not performance sensitive (this only happens one time in the entire bootstrapping process) because I find its a really common source of bugs, but I can change it

persist/fs/commitlog/reader.go
```diff
@@ -56,34 +61,84 @@ type commitLogReader interface {
...
+		bytesPool:    opts.BytesPool(),
+		chunkReader:  newChunkReader(opts.FlushSize()),
+		logDecoder:   msgpack.NewDecoder(decodingOpts),
+		decoderBufs:  decoderBufs,
+		outBufs:      outBufs,
+		cancelCtx:    cancelCtx,
+		cancelFunc:   cancelFunc,
+		shutdownChan: make(chan error),
+		metadata: readerMetadata{
+			decoder: msgpack.NewDecoder(decodingOpts),
```
prateek: you create a new metadata decoder at the start of `readLoop()`, do you still need this?

persist/fs/commitlog/reader.go
prateek: could you add a [leaktest](https://github.com/fortytw2/leaktest) assert to the tests to ensure we're not leaking go-routines

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
+		select {
+		case <-r.cancelCtx.Done():
+			for _, decoderBuf := range r.decoderBufs {
+				close(decoderBuf)
+			}
+			r.shutdownChan <- r.close()
 			return
+		default:
+			if eofFound {
+				continue
```
prateek: Doesn't this cause an infinite loop?

persist/fs/commitlog/reader.go
richardartoul: Cool good to know, I'll change that

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
 	annotation ts.Annotation,
 	resultErr error,
 ) {
-	entry, err := r.readEntry()
-	if err != nil {
-		resultErr = err
-		return
+	if r.nextIndex == 0 {
+		go r.readLoop()
```
prateek: afaict, the code assumes there should only be a single `readLoop` running at any point (`readChunk` is appending data to the bytes array without a guard). Could you add a lock/guard here or in `readLoop` to ensure that only a single instance runs at a time

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
+			}
+			r.shutdownChan <- r.close()
 			return
+		default:
+			if eofFound {
+				continue
+			}
+			data, err := r.readChunk()
+			if err == io.EOF {
+				eofFound = true
```
prateek: Is `data` valid if `err == io.EOF`?

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
-		resultErr = err
-		return
+	if r.nextIndex == 0 {
+		go r.readLoop()
+		for i, decoderBuf := range r.decoderBufs {
+			go r.decoderLoop(decoderBuf, r.outBufs[i])
+		}
 	}
+	// Data is written into these channels in round-robin fashion, so reading them
+	// one at a time in the same order results in an ordered stream.
```
prateek: Not sure I follow entirely. We round-robin across the slice, and write data to each chan. And are subsequently reading decoding each channel's contents. That guarantees we maintain global order iff the readers of each chan are synchronised, I don't see how we're guaranteeing that. 

i.e. say we have concurrency 2, and initial structure -
```
r.outBufs:
  [0]: [] // using an empty slice to indicate channel contents here
  [1]: []
```

after we do a bunch of reads, lets say `a`, `b`, `c`, `d`, we'd end up like:

```
r.outBufs:
  [0]: [a, c] 
  [1]: [b, d]
```

The decoders are only going to maintain local order (i.e. `c` after `a`, and `d` after `b`), not global order. Is that going to cause issues?

persist/fs/commitlog/reader.go
richardartoul: data will be nil, but the code handles that path

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
 	}
-	unit = xtime.Unit(byte(entry.Unit))
-	annotation = entry.Annotation
-	return
 }
 
-func (r *reader) readChunk() ([]byte, error) {
+// Caller is responsible for calling DecRef()
+func (r *reader) readChunk() (checked.Bytes, error) {
 	// Read size of message
```
prateek: just curious - would increasing read perf here (e.g. `mmap`ing the chunk reader) help overall perf at all?

persist/fs/commitlog/reader.go
richardartoul: This works because when we read out of the channels, we also read in round-robin order so with the example you gave

```
r.outBufs:
  [0]: [a, c] 
  [1]: [b, d]
```

We'll read in this order:

1) 0
2) 1
3) 0
4) 1

which translates into a, b, c, d

persist/fs/commitlog/reader.go
richardartoul: Sure, what are you thinking? I guess I could do something like when readLoop starts it locks a global counter, makes sure that the counter is equal to zero and if its not...panic?

persist/fs/commitlog/reader.go
richardartoul: I think you're right, I'll get rid of it

persist/fs/commitlog/reader.go
richardartoul: I considered that, but I don't think so. At least not yet. As far as I can tell from my profiling all the bottlenecks in the code are around M3TSZ encoding and MSGPACK decoding.

This PR along with the next one makes the commitlog bootstrapper about 5-6x faster just by parallelizing those things.

Might be interesting to look at parallelizing the reader later to if we see that become a bottleneck, but there is some weird complexity there because I think we still want things to come out of the reader in the same order that they're found in the commitlog.

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
 	resultErr error,
 ) {
-	entry, err := r.readEntry()
-	if err != nil {
-		resultErr = err
-		return
+	if r.nextIndex == 0 {
+		go r.readLoop()
+		for i, decoderBuf := range r.decoderBufs {
+			go r.decoderLoop(decoderBuf, r.outBufs[i])
```
prateek: More of a style nit: could you capture `decoderBuf` and `r.outBufs[i]` in separate variables above this line, so that readers don't have to remember the closure argument evaluation semantics

persist/fs/commitlog/reader.go
richardartoul: yeah technically, but it doesn't really matter because it ends when you call Close() and its in a background routine, but you're right its poor form. I found a way to make it not an infinite loop and I'll change it.

persist/fs/commitlog/reader.go
richardartoul: sure

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
+				outBuf <- readResponse
+				continue
+			}
 
-	series = metadata
-	datapoint = ts.Datapoint{
-		Timestamp: time.Unix(0, entry.Timestamp),
-		Value:     entry.Value,
+			id := r.bytesPool.Get(len(decoded.ID))
+			id.IncRef()
```
prateek: could you put a comment describing lifecycle of these bytes here: you increment, `ts.BinaryID` increments, and then you decrement. You don't finalize because the metadata struct has a reference.

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
+			id.AppendAll(decoded.ID)
+
+			namespace := r.bytesPool.Get(len(decoded.Namespace))
+			namespace.IncRef()
+			namespace.AppendAll(decoded.Namespace)
+
+			r.metadata.Lock()
+			r.metadata.lookup[entry.Index] = Series{
```
prateek: Do you need to check if this would be overwriting an existing value?

persist/fs/commitlog/reader.go
richardartoul: yeah although I didn't write this code. When does this get finalized then? Seems like the way it is, it never ends up back in the pool?

persist/fs/commitlog/reader.go
```diff
@@ -56,34 +61,84 @@ type commitLogReader interface {
...
+type decoderArg struct {
+	bytes checked.Bytes
+	err   error
+}
+
+type readerMetadata struct {
+	sync.RWMutex
+	decoder encoding.Decoder
+	lookup  map[uint64]Series
+	wgs     map[uint64]*sync.WaitGroup
```
prateek: should this be an `sync.Cond` instead? i.e. when multiple routines are waiting on a metadata, you need a way to wake all of them

persist/fs/commitlog/reader.go
richardartoul: The existing code did not do that. I could add a check, although if there are duplicate values what should I do? Panic?

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
+			r.metadata.lookup[entry.Index] = Series{
+				UniqueIndex: entry.Index,
+				ID:          ts.BinaryID(id),
+				Namespace:   ts.BinaryID(namespace),
+				Shard:       decoded.Shard,
+			}
+			waitGroup, ok := r.metadata.wgs[entry.Index]
+			// Another goroutine is blocked waiting for this metadata
+			if ok {
+				waitGroup.Done()
```
prateek: can have multiple results waiting for a metadata, a single Done won't do the trick. Maybe use a `sync.Cond` instead

persist/fs/commitlog/reader.go
richardartoul: Oh dope that looks cool, I'll add it

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +165,171 @@ func (r *reader) Read() (
...
+					pendingResult.Add(1)
+					r.metadata.wgs[entry.Index] = pendingResult
+				}
+			}
+			r.metadata.Unlock()
+		}
+
+		// The required metadata hasn't been processed yet
+		if pendingResult != nil {
+			pendingResult.Wait()
```
prateek: This can cause an infinite wait if the input file is illformed and we don't receive a metadata entry for a series. We should guard against that case somehow (either timed or expect all metadata within a certain number of entries)

persist/fs/commitlog/reader.go
prateek: Here I mean a memory leak detector. I think there's a way to turn it on for `checked.Bytes`. @robskillington do you remember how?

persist/fs/commitlog/reader.go
richardartoul: I didn't know about sync.Cond when I was writing this, but I think the existing solution is correct. If you look at it, the waitgroup will never increment higher than 1 so a single call to .Done() is enough, and you can have multiple goroutines waiting on the same waitgroup as seen here: https://play.golang.org/p/ivmvhwksFp

persist/fs/commitlog/iterator.go
prateek: Yeah we shouldn't drop these errors silently, could you rewire w/e is necessary.

persist/fs/commitlog/commit_log_test.go
prateek: You're creating a new reader below as is, why does it matter if there's data on disk (i.e. all the preceeding lines) before you check reuse of the commit log reader?

persist/fs/commitlog/reader.go
prateek: Could you verify what impact it has in a couple of runs, can decide whether we should move it up after.

persist/fs/commitlog/reader.go
prateek: Yeah not the biggest difference perf wise but might as well whilst we're here.

persist/fs/commitlog/reader.go
prateek: yeah something like a bool w/ a guard or an atomic value in the reader struct. Lol, no panics, how bout just an error?

persist/fs/commitlog/reader.go
prateek: Feels wrong to bake in that assumption to the code. Why not just terminate if you're done?

persist/fs/commitlog/reader.go
prateek: fair enough.

persist/fs/commitlog/reader.go
prateek: I think the series eviction logic during ticks would finalize. /cc-ing @robskillington to confirm if that's true

persist/fs/commitlog/reader.go
richardartoul: Basically the iterator that is calling Read() on the Reader is expecting back: 
```
(
	series Series,
	datapoint ts.Datapoint,
	unit xtime.Unit,
	annotation ts.Annotation,
	resultErr error,
)
```
 on every functional call.

In order for me to satisfy that interface, when I encounter an io.EOF I have to propagate the error through the rest of the code so that the iterator can receive it. That means continuing through the code path one more time where I send the data through the channel with these lines:

```
r.decoderBufs[index%r.numConc] <- decoderArg{
				bytes: data,
				err:   err,
			}
```

Does that make sense?

I could copy that code into the if statement and return early immediately if you think that would be cleaner

richardartoul: I might be misunderstanding you if you want to talk in person

persist/fs/commitlog/reader.go
prateek: Nah no need to panic, just release the now redundant id/namespace allocs, and signal to any waiting routines

persist/fs/commitlog/reader.go
prateek: yeah but if multiple go routines are waiting on the same waitgroup, none will be released with a single `Done()`.

persist/fs/commitlog/reader.go
prateek: i'd do the same as what you're doing in the `errCommitLogReaderMissingLogMetadata` case below

persist/fs/commitlog/reader.go
prateek: Makes sense. Could you put a little blurb explaining this in the struct/method

persist/fs/commitlog/reader.go
richardartoul: Are you sure? I tested it in the go playground and you have multiple goroutines waiting on the same waitgroup with a single call to .Done() and it works: https://play.golang.org/p/ivmvhwksFp

persist/fs/commitlog/reader.go
richardartoul: In that case, whats the difference between just overwriting it?

persist/fs/commitlog/reader.go
richardartoul: Sure, should I just return the error from the Read() call? I wanted to add the check in the readLoop() call itself, but I can't return errors from that because I'm executing it async with go keyword.

I could either:

1) Do the check inside of the Read() function (protects callers from misuse, but not people writing code in this file)
2) do the check in the readLoop function, remove the go keyword in the caller, and make readLoop() start its own goroutine after validating that its safe to do so.

Do you have a preference?

persist/fs/commitlog/reader.go
prateek: Sounds good

persist/fs/commitlog/reader.go
prateek: You're able to save memory by releasing the id/namespace, which you wouldn't do if you overwrote.

persist/fs/commitlog/reader.go
prateek: only happens if you have more than 1 waiting go-routine, e.g. https://play.golang.org/p/mjcxBGv4Bn

persist/fs/commitlog/reader.go
richardartoul: will do

persist/fs/commitlog/reader.go
richardartoul: I'll add a comment explaining it

persist/fs/commitlog/reader.go
prateek: What I mean is, you're using `data == nil` and/or `err == io.EOF` as sentinels indicating the decoders should terminate. Would be cleaner to use another way to indicate termination to the decoders, e.g. a closeCh channel per decoder, or a field in the decoderArg

persist/fs/commitlog/reader.go
prateek: Either way is good with me.

persist/fs/commitlog/reader.go
richardartoul: ah I see what you're saying. ok cool

persist/fs/commitlog/reader.go
richardartoul: Hmm I'm still not quite following you. The decoders are closed by an explicit call to shutdown(), they're not waiting to see if err == io.EOF, they just pass those values along.

Let's sync in person on this, I might still be misunderstanding

persist/fs/commitlog/commit_log_test.go
richardartoul: Yeah I guess you're right it doesn't matter, I'll remove some of that stuff.

persist/fs/commitlog/reader.go
richardartoul: Yeah but the difference is that in your example the waitgroup was incremented multiple times (I.E you called add twice). In the reader code I have, even if there are 4 goroutines waiting, Add() is only called once and they're all released with a single call to Done()

persist/fs/commitlog/commit_log_test.go
prateek: Sounds good

persist/fs/commitlog/iterator.go
```diff
@@ -95,8 +95,8 @@ func (i *iterator) Next() bool {
 	var err error
 	i.read.series, i.read.datapoint, i.read.unit, i.read.annotation, err = i.reader.Read()
 	if err == io.EOF {
+		i.closeAndResetReader()
```
prateek: could you check if errs (and if so set it in `i.err`)

persist/fs/commitlog/iterator.go
```diff
@@ -126,10 +126,7 @@ func (i *iterator) Close() {
 		return
 	}
 	i.closed = true
-	if i.reader != nil {
-		i.reader.Close()
-		i.reader = nil
-	}
+	i.closeAndResetReader()
```
prateek: same here

richardartoul: cool I made this one return false if there is an error if thats ok

persist/fs/commitlog/commit_log_test.go
richardartoul: oh actually part of it was just so I would have a file to read

persist/fs/commitlog/commit_log_test.go
```diff
@@ -235,7 +251,7 @@ func writeCommitLogs(
 
 		// Wait for previous writes to enqueue
 		for getAllWrites() != preWrites+i {
-			time.Sleep(10 * time.Millisecond)
+			time.Sleep(1 * time.Microsecond)
```
prateek: s/`1 * time.Microsecond`/`time.Microsecond`

persist/fs/commitlog/commit_log_test.go
```diff
@@ -256,7 +272,7 @@ func writeCommitLogs(
 
 	// Wait for all writes to enqueue
 	for getAllWrites() != preWrites+len(writes) {
-		time.Sleep(10 * time.Millisecond)
+		time.Sleep(1 * time.Microsecond)
```
prateek: same here

persist/fs/commitlog/commit_log_test.go
```diff
@@ -330,6 +346,100 @@ func TestCommitLogWrite(t *testing.T) {
 	assertCommitLogWritesByIterating(t, commitLog, writes)
 }
 
+func TestReadCommitLogMissingMetadata(t *testing.T) {
+	readConc := 4
+	// Make sure we're not leaking goroutines
+	defer leaktest.CheckTimeout(t, 10*time.Second)()
```
prateek: nice!

persist/fs/commitlog/iterator.go
prateek: looks good. do you need to do the `closeErr` check on line 137 too?

persist/fs/commitlog/options.go
prateek: Sounds good

persist/fs/commitlog/commit_log_test.go
```diff
@@ -96,6 +98,20 @@ func cleanup(t *testing.T, opts Options) {
 	assert.NoError(t, os.RemoveAll(filePathPrefix))
 }
 
+type mockBitSet struct {
+	indexTestReturn map[uint]bool
+}
+
+func (b *mockBitSet) test(i uint) bool {
+	// return b.indexTestReturn[i]
```
prateek: do you mean to undo/remove this line?

persist/fs/commitlog/commit_log_test.go
```diff
@@ -330,6 +346,100 @@ func TestCommitLogWrite(t *testing.T) {
...
+
+	// Close the commit log and consequently flush
+	assert.NoError(t, commitLog.Close())
+
+	// Make sure we don't panic / deadlock
+	iter, err := commitLog.Iter()
+	assert.NoError(t, err)
+	for iter.Next() {
+	}
+	iter.Close()
```
prateek: do you need to check the return code for `iter.Err()`, and the other close methods too?

persist/fs/commitlog/reader.go
prateek: any word on this

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +198,265 @@ func (r *reader) Read() (
...
+			return
+		// outBuf is full
+		default:
+			r.metadata.Lock()
+			// Check if all the other decoderLoops are finished or blocked too
+			if r.metadata.numBlockedOrFinishedDecoders >= r.numConc-1 {
+				// If they are, then performing a blocking write would cause deadlock
+				// so we free all pending waiters
+				r.freeAllPendingWaiters()
+				r.metadata.Unlock()
```
prateek: feels sketch to be allowing this to fall thru and loop intentionally.

persist/fs/commitlog/reader.go
richardartoul: yep I messed around with it when I was debugging that deadlock issue and the code is still correct even if I set the value to 1, but I didn't notice any perf difference between 100 and 1000 so I think this is an ok value for now

persist/fs/commitlog/iterator.go
richardartoul: the current implementation of the Close() method doesn't return an error so there is nowhere to bubble it up to

richardartoul: I could leave a TODO with a note to refactor the rest of the codebase to support an error on Close()?

persist/fs/commitlog/commit_log_test.go
richardartoul: nope will undo, was just debugging

persist/fs/commitlog/commit_log_test.go
richardartoul: added it

persist/fs/commitlog/reader.go
richardartoul: Are you worried about too much unnecessary looping (could be solved with a short sleep?) or you just don't like the code structure?

This is the only way I know how to write a potentially blocked channel without deadlocking

persist/fs/commitlog/iterator.go
prateek: Yep TODO sounds good

persist/fs/commitlog/reader.go
prateek: Dope, sounds good

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +198,265 @@ func (r *reader) Read() (
...
 
-	metadata, ok := r.metadataLookup[entry.Index]
-	if !ok && len(entry.Metadata) != 0 {
-		r.metadataDecoder.Reset(entry.Metadata)
-		decoded, err := r.metadataDecoder.DecodeLogMetadata()
-		if err != nil {
-			resultErr = err
+func (r *reader) startBackgroundWorkers() error {
+	// Make sure background workers are never setup more than once
+	r.backgroundWorkersInit.Lock()
```
prateek: could you replace this with an atomic cas, something like:

```
set := atomic.CompareAndSwapInt64(&r.bgWorkersInitialised, 0, 1)
if !set {
  return errCommitLogReaderMultipleReadloops
}
// continue onwards
```

persist/fs/commitlog/reader.go
prateek: lets chat in person, have a couple of concerns about it

persist/fs/commitlog/iterator.go
prateek: +1 for the todo

persist/fs/commitlog/iterator.go
```diff
@@ -177,3 +183,12 @@ func (i *iterator) nextReader() bool {
 	i.reader = reader
 	return true
 }
+
+func (i *iterator) closeAndResetReader() error {
+	if i.reader != nil {
```
prateek: nit: reads better if you reverse the ordering, i.e.
```
if i.reader == nil {
    return nil 
}
reader := i.reader
i.reader = nil
return reader.Close()
```

persist/fs/commitlog/reader.go
richardartoul: cool lets do it

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +192,267 @@ func (r *reader) Read() (
...
+				metadata := Series{
+					UniqueIndex: entry.Index,
+					ID:          ts.BinaryID(id),
+					Namespace:   ts.BinaryID(namespace),
+					Shard:       decoded.Shard,
+				}
+				r.metadata.lookup[entry.Index] = metadata
+				pendingMetadata, ok := r.metadata.pending[entry.Index]
+				// One (or more) goroutines are blocked waiting for this metadata
+				if ok {
```
prateek: this ordering is weird, can you condense the ok checks as follows:

```
// One (or more) ...

namespace.DecRef()
id.DecRef()

if ok {
  delete(r.metadata.pending, entry.Index)
  for _, p := range pendingMetadata {
    p.done(metadata, nil)
  }
}
```

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +192,267 @@ func (r *reader) Read() (
...
+		entry, err := decoder.DecodeLogEntry()
+		if err != nil {
+			readResponse.resultErr = err
+			r.writeToOutBuf(outBuf, readResponse)
+			continue
 		}
-		ok = true
 
-		r.metadataLookup[entry.Index] = metadata
+		if len(entry.Metadata) != 0 {
```
prateek: can you flip the check and reduce the nesting here? something like:

```
if len(entry.Metadata) == 0 {
  continue 
}

// rest of the code
```

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +192,267 @@ func (r *reader) Read() (
...
+
+			r.metadata.Lock()
+			_, ok := r.metadata.lookup[entry.Index]
+			// If the metadata already exists, we can skip this step
+			if ok {
+				id.DecRef()
+				id.Finalize()
+				namespace.DecRef()
+				namespace.Finalize()
+			} else {
```
prateek: could you `continue` in the if branch to avoid this else?

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +192,267 @@ func (r *reader) Read() (
...
+	}
+	r.metadata.Unlock()
+	// Close the outBuf now that there is no more data to produce to it
+	close(outBuf)
+}
 
-	if !ok {
-		resultErr = errCommitLogReaderMissingLogMetadata
-		return
+func (r *reader) writeToOutBuf(outBuf chan<- readResponse, readResponse readResponse) {
```
prateek: good stuff, this function is much easier to grok now.

persist/fs/commitlog/reader.go
```diff
@@ -109,58 +192,267 @@ func (r *reader) Read() (
...
+}
+
+func (r *reader) shutdown() {
+	for _, decoderBuf := range r.decoderBufs {
+		close(decoderBuf)
+	}
+	r.shutdownCh <- r.close()
+}
+
+func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan<- readResponse) {
```
prateek: This function is doing a lot, do you think you can break it into smaller parts to make it easier to follow ~ for e.g. the metadata retrieval bits on line 344-364 can probably be another function on the `reader`, same for the pending metadata release from lines 320-344.

persist/fs/commitlog/reader.go
richardartoul: I don't think so because there is more code that needs to execute afterwards regardless

persist/fs/commitlog/reader.go
richardartoul: I don't think so for the same reason as above, I need to execute the code after the if statement regardless

PR #319, opened 2017-07-19
Parallelize commitlog bootstrapper

richardartoul:
This is PR 3 in three PR's that together will improve the performance of the commitlog bootstrapper. This PR parallelizes the commitlog bootstrapper itself.

### DIFF
--- a/storage/bootstrap/bootstrapper/commitlog/options.go
+++ b/storage/bootstrap/bootstrapper/commitlog/options.go
@@ -21,6 +21,8 @@
 package commitlog
 
 import (
+	"errors"
+
 	"github.com/m3db/m3db/persist/fs/commitlog"
 	"github.com/m3db/m3db/storage/bootstrap/result"
 )
@@ -28,6 +30,11 @@ import (
 const (
 	defaultEncodingConcurrency   = 4
 	defaultMergeShardConcurrency = 4
+)
+
+var (
+	errEncodingConcurrencyPositive   = errors.New("encoding concurrency must be positive")
+	errMergeShardConcurrencyPositive = errors.New("merge shard concurrency must be positive")
 )
 
 type options struct {
@@ -48,6 +55,12 @@ func NewOptions() Options {
 }
 
 func (o *options) Validate() error {
+	if o.encodingConcurrency <= 0 {
+		return errEncodingConcurrencyPositive
+	}
+	if o.mergeShardConcurrency <= 0 {
+		return errMergeShardConcurrencyPositive
+	}
 	return o.commitLogOpts.Validate()
 }
 
--- a/storage/bootstrap/bootstrapper/commitlog/options.go
+++ b/storage/bootstrap/bootstrapper/commitlog/options.go
@@ -25,6 +25,11 @@ import (
 	"github.com/m3db/m3db/storage/bootstrap/result"
 )
 
+const (
+	defaultEncodingConcurrency   = 4
+	defaultMergeShardConcurrency = 4
+)
+
 type options struct {
 	resultOpts            result.Options
 	commitLogOpts         commitlog.Options
@@ -37,8 +42,8 @@ func NewOptions() Options {
 	return &options{
 		resultOpts:            result.NewOptions(),
 		commitLogOpts:         commitlog.NewOptions(),
-		encodingConcurrency:   4,
-		mergeShardConcurrency: 4,
+		encodingConcurrency:   defaultEncodingConcurrency,
+		mergeShardConcurrency: defaultMergeShardConcurrency,
 	}
 }
 
--- a/storage/bootstrap/bootstrapper/commitlog/options.go
+++ b/storage/bootstrap/bootstrapper/commitlog/options.go
@@ -26,15 +26,19 @@ import (
 )
 
 type options struct {
-	resultOpts    result.Options
-	commitLogOpts commitlog.Options
+	resultOpts               result.Options
+	commitLogOpts            commitlog.Options
+	m3TSZEncodingConcurrency int
+	mergeShardConcurrency    int
 }
 
 // NewOptions creates new bootstrap options
 func NewOptions() Options {
 	return &options{
-		resultOpts:    result.NewOptions(),
-		commitLogOpts: commitlog.NewOptions(),
+		resultOpts:               result.NewOptions(),
+		commitLogOpts:            commitlog.NewOptions(),
+		m3TSZEncodingConcurrency: 4,
+		mergeShardConcurrency:    4,
 	}
 }
 
@@ -60,4 +64,24 @@ func (o *options) SetCommitLogOptions(value commitlog.Options) Options {
 
 func (o *options) CommitLogOptions() commitlog.Options {
 	return o.commitLogOpts
+}
+
+func (o *options) SetM3TSZEncodingConcurrency(value int) Options {
+	opts := *o
+	opts.m3TSZEncodingConcurrency = value
+	return &opts
+}
+
+func (o *options) M3TSZEncodingConcurrency() int {
+	return o.m3TSZEncodingConcurrency
+}
+
+func (o *options) SetMergeShardsConcurrency(value int) Options {
+	opts := *o
+	opts.mergeShardConcurrency = value
+	return &opts
+}
+
+func (o *options) MergeShardsConcurrency() int {
+	return o.mergeShardConcurrency
 }
--- a/storage/bootstrap/bootstrapper/commitlog/options.go
+++ b/storage/bootstrap/bootstrapper/commitlog/options.go
@@ -26,19 +26,19 @@ import (
 )
 
 type options struct {
-	resultOpts               result.Options
-	commitLogOpts            commitlog.Options
-	m3TSZEncodingConcurrency int
-	mergeShardConcurrency    int
+	resultOpts            result.Options
+	commitLogOpts         commitlog.Options
+	encodingConcurrency   int
+	mergeShardConcurrency int
 }
 
 // NewOptions creates new bootstrap options
 func NewOptions() Options {
 	return &options{
-		resultOpts:               result.NewOptions(),
-		commitLogOpts:            commitlog.NewOptions(),
-		m3TSZEncodingConcurrency: 4,
-		mergeShardConcurrency:    4,
+		resultOpts:            result.NewOptions(),
+		commitLogOpts:         commitlog.NewOptions(),
+		encodingConcurrency:   4,
+		mergeShardConcurrency: 4,
 	}
 }
 
@@ -66,14 +66,14 @@ func (o *options) CommitLogOptions() commitlog.Options {
 	return o.commitLogOpts
 }
 
-func (o *options) SetM3TSZEncodingConcurrency(value int) Options {
+func (o *options) SetEncodingConcurrency(value int) Options {
 	opts := *o
-	opts.m3TSZEncodingConcurrency = value
+	opts.encodingConcurrency = value
 	return &opts
 }
 
-func (o *options) M3TSZEncodingConcurrency() int {
-	return o.m3TSZEncodingConcurrency
+func (o *options) EncodingConcurrency() int {
+	return o.encodingConcurrency
 }
 
 func (o *options) SetMergeShardsConcurrency(value int) Options {
--- a/storage/bootstrap/bootstrapper/commitlog/source.go
+++ b/storage/bootstrap/bootstrapper/commitlog/source.go
@@ -400,6 +400,7 @@ func (s *commitLogSource) mergeShard(
 			block := blocksPool.Get()
 			if len(unmergedBlockEncoders) == 0 {
 				numShardEmptyErrs++
+				blocksPool.Put(block)
 				continue
 			} else if len(unmergedBlockEncoders) == 1 {
 				block.Reset(start, unmergedBlockEncoders[0].enc.Discard())
@@ -441,6 +442,7 @@ func (s *commitLogSource) mergeShard(
 				}
 
 				if err != nil {
+					blocksPool.Put(block)
 					continue
 				}
 
--- a/storage/bootstrap/bootstrapper/commitlog/source.go
+++ b/storage/bootstrap/bootstrapper/commitlog/source.go
@@ -231,21 +231,7 @@ func (s *commitLogSource) Read(
 	// Wait for all merge goroutines to complete
 	wg.Wait()
 
-	errSum = 0
-	for _, numErrs := range shardErrs {
-		errSum += numErrs
-	}
-	if errSum > 0 {
-		s.log.Errorf("error bootstrapping from commit log: %d merge out of order errors", errSum)
-	}
-
-	emptyErrSum := 0
-	for _, numEmptyErr := range shardEmptyErrs {
-		emptyErrSum += numEmptyErr
-	}
-	if emptyErrSum > 0 {
-		s.log.Errorf("error bootstrapping from commit log: %d empty unmerged blocks errors", emptyErrSum)
-	}
+	s.printMergeShardsOutcome(shardErrs, shardEmptyErrs)
 
 	return bootstrapResult, nil
 }
@@ -460,4 +446,22 @@ func (s *commitLogSource) mergeShard(
 		}
 	}
 	return shardResult, numShardEmptyErrs, numErrs
+}
+
+func (s *commitLogSource) printMergeShardsOutcome(shardErrs []int, shardEmptyErrs []int) {
+	errSum := 0
+	for _, numErrs := range shardErrs {
+		errSum += numErrs
+	}
+	if errSum > 0 {
+		s.log.Errorf("error bootstrapping from commit log: %d merge out of order errors", errSum)
+	}
+
+	emptyErrSum := 0
+	for _, numEmptyErr := range shardEmptyErrs {
+		emptyErrSum += numEmptyErr
+	}
+	if emptyErrSum > 0 {
+		s.log.Errorf("error bootstrapping from commit log: %d empty unmerged blocks errors", emptyErrSum)
+	}
 }
--- a/storage/bootstrap/bootstrapper/commitlog/source.go
+++ b/storage/bootstrap/bootstrapper/commitlog/source.go
@@ -197,17 +197,7 @@ func (s *commitLogSource) Read(
 	// Block until all data has been read and encoded by the worker goroutines
 	wg.Wait()
 
-	// TODO: Function?
-	errSum := 0
-	for _, numErrs := range workerErrs {
-		errSum += numErrs
-	}
-	if errSum > 0 {
-		s.log.Errorf("error bootstrapping from commit log: %d block encode errors", errSum)
-	}
-	if err := iter.Err(); err != nil {
-		s.log.Errorf("error reading commit log: %v", err)
-	}
+	s.printM3TSZEncodingOutcome(workerErrs, iter)
 
 	shardErrs := make([]int, numShards, numShards)
 	shardEmptyErrs := make([]int, numShards, numShards)
@@ -389,6 +379,19 @@ func (s *commitLogSource) shouldEncodeSeries(
 	}
 
 	return true
+}
+
+func (s *commitLogSource) printM3TSZEncodingOutcome(workerErrs []int, iter commitlog.Iterator) {
+	errSum := 0
+	for _, numErrs := range workerErrs {
+		errSum += numErrs
+	}
+	if errSum > 0 {
+		s.log.Errorf("error bootstrapping from commit log: %d block encode errors", errSum)
+	}
+	if err := iter.Err(); err != nil {
+		s.log.Errorf("error reading commit log: %v", err)
+	}
 }
 
 func (s *commitLogSource) mergeShard(
--- a/storage/bootstrap/bootstrapper/commitlog/source.go
+++ b/storage/bootstrap/bootstrapper/commitlog/source.go
@@ -124,7 +124,7 @@ func (s *commitLogSource) Read(
 		blopts       = bopts.DatabaseBlockOptions()
 		blockSize    = ns.Options().RetentionOptions().BlockSize()
 		encoderPool  = bopts.DatabaseBlockOptions().EncoderPool()
-		workerErrs   = make([]int, numConc, numConc)
+		workerErrs   = make([]int, numConc)
 	)
 
 	encoderChans := []chan encoderArg{}
@@ -198,41 +198,7 @@ func (s *commitLogSource) Read(
 	wg.Wait()
 	s.printM3TSZEncodingOutcome(workerErrs, iter)
 
-	shardErrs := make([]int, numShards, numShards)
-	shardEmptyErrs := make([]int, numShards, numShards)
-	bootstrapResult := result.NewBootstrapResult()
-	blocksPool := bopts.DatabaseBlockOptions().DatabaseBlockPool()
-	multiReaderIteratorPool := blopts.MultiReaderIteratorPool()
-	// Controls how many shards can be merged in parallel
-	mergeSemaphore := make(chan bool, s.opts.MergeShardsConcurrency())
-	bootstrapResultLock := sync.Mutex{}
-	wg = &sync.WaitGroup{}
-
-	for shard, unmergedShard := range unmerged {
-		mergeSemaphore <- true
-		wg.Add(1)
-		go func(shard int, unmergedShard map[ts.Hash]encodersByTime) {
-			var shardResult result.ShardResult
-			shardResult, shardEmptyErrs[shard], shardErrs[shard] = s.mergeShard(
-				unmergedShard, blocksPool, multiReaderIteratorPool, encoderPool, blopts)
-			if len(shardResult.AllSeries()) > 0 {
-				// Prevent race conditions while updating bootstrapResult from multiple go-routines
-				bootstrapResultLock.Lock()
-				// Shard is a slice index so conversion to uint32 is safe
-				bootstrapResult.Add(uint32(shard), shardResult, nil)
-				bootstrapResultLock.Unlock()
-			}
-			<-mergeSemaphore
-			wg.Done()
-		}(shard, unmergedShard)
-	}
-
-	// Wait for all merge goroutines to complete
-	wg.Wait()
-
-	s.printMergeShardsOutcome(shardErrs, shardEmptyErrs)
-
-	return bootstrapResult, nil
+	return s.mergeShards(numShards, bopts, blopts, encoderPool, unmerged), nil
 }
 
 func (s *commitLogSource) startM3TSZEncodingWorker(
@@ -377,6 +343,48 @@ func (s *commitLogSource) printM3TSZEncodingOutcome(workerErrs []int, iter commi
 	if err := iter.Err(); err != nil {
 		s.log.Errorf("error reading commit log: %v", err)
 	}
+}
+
+func (s *commitLogSource) mergeShards(
+	numShards int,
+	bopts result.Options,
+	blopts block.Options,
+	encoderPool encoding.EncoderPool,
+	unmerged []map[ts.Hash]encodersByTime,
+) result.BootstrapResult {
+	shardErrs := make([]int, numShards, numShards)
+	shardEmptyErrs := make([]int, numShards, numShards)
+	bootstrapResult := result.NewBootstrapResult()
+	blocksPool := bopts.DatabaseBlockOptions().DatabaseBlockPool()
+	multiReaderIteratorPool := blopts.MultiReaderIteratorPool()
+	// Controls how many shards can be merged in parallel
+	mergeSemaphore := make(chan bool, s.opts.MergeShardsConcurrency())
+	bootstrapResultLock := sync.Mutex{}
+	wg := &sync.WaitGroup{}
+
+	for shard, unmergedShard := range unmerged {
+		mergeSemaphore <- true
+		wg.Add(1)
+		go func(shard int, unmergedShard map[ts.Hash]encodersByTime) {
+			var shardResult result.ShardResult
+			shardResult, shardEmptyErrs[shard], shardErrs[shard] = s.mergeShard(
+				unmergedShard, blocksPool, multiReaderIteratorPool, encoderPool, blopts)
+			if len(shardResult.AllSeries()) > 0 {
+				// Prevent race conditions while updating bootstrapResult from multiple go-routines
+				bootstrapResultLock.Lock()
+				// Shard is a slice index so conversion to uint32 is safe
+				bootstrapResult.Add(uint32(shard), shardResult, nil)
+				bootstrapResultLock.Unlock()
+			}
+			<-mergeSemaphore
+			wg.Done()
+		}(shard, unmergedShard)
+	}
+
+	// Wait for all merge goroutines to complete
+	wg.Wait()
+	s.printMergeShardsOutcome(shardErrs, shardEmptyErrs)
+	return bootstrapResult
 }
 
 func (s *commitLogSource) mergeShard(
--- a/storage/bootstrap/bootstrapper/commitlog/source.go
+++ b/storage/bootstrap/bootstrapper/commitlog/source.go
@@ -119,7 +119,7 @@ func (s *commitLogSource) Read(
 		numShards    = len(shardsTimeRanges)
 		unmerged     = []map[ts.Hash]encodersByTime{}
 		unmergedLock = &sync.RWMutex{}
-		numConc      = s.opts.M3TSZEncodingConcurrency()
+		numConc      = s.opts.EncodingConcurrency()
 		bopts        = s.opts.ResultOptions()
 		blopts       = bopts.DatabaseBlockOptions()
 		blockSize    = ns.Options().RetentionOptions().BlockSize()
--- a/storage/bootstrap/bootstrapper/commitlog/source.go
+++ b/storage/bootstrap/bootstrapper/commitlog/source.go
@@ -224,7 +224,6 @@ func (s *commitLogSource) startM3TSZEncodingWorker(
 		// has been resized already
 		unmerged := *unmergedPtr
 		unmergedShard := unmerged[series.Shard]
-		unmergedLock.RUnlock()
 		if unmergedShard == nil {
 			unmergedShard = make(map[ts.Hash]encodersByTime)
 			unmerged[series.Shard] = unmergedShard
@@ -237,6 +236,7 @@ func (s *commitLogSource) startM3TSZEncodingWorker(
 				encoders: make(map[int64][]encoder)}
 			unmergedShard[series.ID.Hash()] = unmergedSeries
 		}
+		unmergedLock.RUnlock()
 
 		var (
 			err           error
--- a/storage/bootstrap/bootstrapper/commitlog/source.go
+++ b/storage/bootstrap/bootstrapper/commitlog/source.go
@@ -196,7 +196,6 @@ func (s *commitLogSource) Read(
 
 	// Block until all data has been read and encoded by the worker goroutines
 	wg.Wait()
-
 	s.printM3TSZEncodingOutcome(workerErrs, iter)
 
 	shardErrs := make([]int, numShards, numShards)
@@ -239,7 +238,7 @@ func (s *commitLogSource) Read(
 func (s *commitLogSource) startM3TSZEncodingWorker(
 	workerNum int,
 	ec <-chan encoderArg,
-	// Pass as pointer so slice resizing is reflected
+	// Pass as pointer so slice resizing is reflected elsewhere
 	unmergedPtr *[]map[ts.Hash]encodersByTime,
 	unmergedLock *sync.RWMutex,
 	encoderPool encoding.EncoderPool,
@@ -342,17 +341,18 @@ func (s *commitLogSource) shouldEncodeSeries(
 	series commitlog.Series,
 	dp ts.Datapoint,
 ) bool {
-	// check if the series belongs to current namespace being bootstrapped
+	// Check if the series belongs to current namespace being bootstrapped
 	if !namespace.Equal(series.Namespace) {
 		return false
 	}
 
+	// Check if the shard is one of the shards we're trying to bootstrap
 	ranges, ok := shardsTimeRanges[series.Shard]
 	if !ok {
-		// Not bootstrapping this shard
 		return false
 	}
 
+	// Check if the block corresponds to the time-range that we're trying to bootstrap
 	blockStart := dp.Timestamp.Truncate(blockSize)
 	blockEnd := blockStart.Add(blockSize)
 	blockRange := xtime.Range{
@@ -360,7 +360,6 @@ func (s *commitLogSource) shouldEncodeSeries(
 		End:   blockEnd,
 	}
 	if !ranges.Overlaps(blockRange) {
-		// Data in this block does not match the requested ranges
 		return false
 	}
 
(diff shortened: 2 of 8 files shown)


storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -49,9 +52,21 @@ type encoder struct {
...
 }
 
-type encoderMap struct {
-	id       ts.ID
-	encoders map[time.Time][]encoder
+type encodersByTime struct {
+	id ts.ID
+	// int64 instead of time.Time because there is an optimized map access pattern
+	// for i64's
+	encoders map[int64][]encoder
```
richardartoul: I'm currently creating a type alias for this called UnixNano to be used throughout the codebase wherever we would normally use a map[time.Time]. I'll replace this int64 with the appropriate type in a subsequent PR

storage/bootstrap/bootstrapper/commitlog/options.go
```diff
@@ -61,3 +65,23 @@ func (o *options) SetCommitLogOptions(value commitlog.Options) Options {
...
 	return o.commitLogOpts
 }
+
+func (o *options) SetM3TSZEncodingConcurrency(value int) Options {
+	opts := *o
+	opts.m3TSZEncodingConcurrency = value
+	return &opts
+}
+
+func (o *options) M3TSZEncodingConcurrency() int {
```
robskillington: Lets just call this `EncodingConcurrency` since all the encoder/decoder interfaces are generic and aren't necessarily using M3TSZ encoding dependening on the encoder pool specified.

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -96,56 +111,132 @@ func (s *commitLogSource) Read(
...
+		unmergedLock = &sync.RWMutex{}
+		numConc      = s.opts.M3TSZEncodingConcurrency()
+		bopts        = s.opts.ResultOptions()
+		blopts       = bopts.DatabaseBlockOptions()
+		blockSize    = ns.Options().RetentionOptions().BlockSize()
+		encoderPool  = bopts.DatabaseBlockOptions().EncoderPool()
+		workerErrs   = make([]int, numConc)
 	)
+
+	encoderChans := []chan encoderArg{}
```
robskillington: nit: Create with upfront size to save allocation here, i.e. `make([]chan encoderArg, 0, numConc)`

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -96,56 +111,132 @@ func (s *commitLogSource) Read(
...
 
-		ranges, ok := shardsTimeRanges[series.Shard]
-		if !ok {
-			// Not bootstrapping this shard
-			continue
+		// We use the unmerged slice for lookup because its faster than a map.
+		// Shard ID's map directly to indices, but we don't know the highest shard
+		// number upfront so we resize when necessary. We don't need to acquire a
+		// lock when we check the length because this is the only goroutine that
+		// change the size of the slice.
```
robskillington: Technically you could a single range over `shardTimeRanges` to check what the highest shard is up front yeah?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -96,56 +111,132 @@ func (s *commitLogSource) Read(
...
+		// approximately numShards / numConc shards. This also means that all
+		// datapoints for a given shard/series will be processed in a serialized
+		// manner.
+		encoderChans[series.Shard%uint32(numConc)] <- struct {
+			series     commitlog.Series
+			dp         ts.Datapoint
+			unit       xtime.Unit
+			annotation ts.Annotation
+			blockStart time.Time
+		}{
```
robskillington: Better to actually used the declared type that inline declare a type, e.g. `<- encoderArg{ /* fields */ }`

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -96,56 +111,132 @@ func (s *commitLogSource) Read(
...
+
+	return s.mergeShards(numShards, bopts, blopts, encoderPool, unmerged), nil
+}
+
+func (s *commitLogSource) startM3TSZEncodingWorker(
+	workerNum int,
+	ec <-chan encoderArg,
+	// Pass as pointer so slice resizing is reflected elsewhere
+	unmergedPtr *[]map[ts.Hash]encodersByTime,
+	unmergedLock *sync.RWMutex,
```
robskillington: If we range over `shardTimeRanges` and allocate the slice up front knowing the largest shard we'll ever encounter, and we also modulo using the current shard for the series then each worker has their own independent set of shards they're decoding for.

Using this approach we can get rid of all locking w.r.t. to unmerged encoders.

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +257,182 @@ func (s *commitLogSource) Read(
...
+	series commitlog.Series,
+	dp ts.Datapoint,
+) bool {
+	// Check if the series belongs to current namespace being bootstrapped
+	if !namespace.Equal(series.Namespace) {
+		return false
+	}
+
+	// Check if the shard is one of the shards we're trying to bootstrap
+	ranges, ok := shardsTimeRanges[series.Shard]
```
robskillington: If we allocate unmerged upfront and only allocate maps for shards in that slice we can do a slice lookup here which is much faster than a map lookup.

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +257,182 @@ func (s *commitLogSource) Read(
...
+	blockEnd := blockStart.Add(blockSize)
+	blockRange := xtime.Range{
+		Start: blockStart,
+		End:   blockEnd,
+	}
+
+	return ranges.Overlaps(blockRange)
+}
+
+func (s *commitLogSource) printM3TSZEncodingOutcome(workerErrs []int, iter commitlog.Iterator) {
```
robskillington: Can we change print... to log... to be more clear (i.e. we're not just printing to stdout/stderr, depending on where our log goes it goes to wherever it is directed to).  We could call this simply `logEncodingOutcome(..)` again because encoders could technically also be an encoder other than a M3TSZ encoder.

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +257,182 @@ func (s *commitLogSource) Read(
...
+) result.BootstrapResult {
+	shardErrs := make([]int, numShards)
+	shardEmptyErrs := make([]int, numShards)
 	bootstrapResult := result.NewBootstrapResult()
 	blocksPool := bopts.DatabaseBlockOptions().DatabaseBlockPool()
 	multiReaderIteratorPool := blopts.MultiReaderIteratorPool()
+	// Controls how many shards can be merged in parallel
+	mergeSemaphore := make(chan bool, s.opts.MergeShardsConcurrency())
+	bootstrapResultLock := sync.Mutex{}
+	wg := &sync.WaitGroup{}
```
robskillington: Don't necessarily need to make this a pointer type, it'll be captured by ref anyhow by the goroutine with just `var wg sync.WaitGroup`

robskillington: Also would replace all these definitions with a var block:

```
var (
	shardErrs = make([]int, numShards)
	shardEmptyErrs = make([]int, numShards)
	bootstrapResult = result.NewBootstrapResult()
	blocksPool = bopts.DatabaseBlockOptions().DatabaseBlockPool()
	multiReaderIteratorPool = blopts.MultiReaderIteratorPool()
	mergeSemaphore = make(chan bool, s.opts.MergeShardsConcurrency())
	bootstrapResultLock  sync.Mutex{}
	wg sync.WaitGroup{}
)
```

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +257,182 @@ func (s *commitLogSource) Read(
...
-						defer stream.Finalize()
-					}
+		mergeSemaphore <- true
+		wg.Add(1)
+		go func(shard int, unmergedShard map[ts.Hash]encodersByTime) {
+			var shardResult result.ShardResult
+			shardResult, shardEmptyErrs[shard], shardErrs[shard] = s.mergeShard(
+				unmergedShard, blocksPool, multiReaderIteratorPool, encoderPool, blopts)
+			if len(shardResult.AllSeries()) > 0 {
+				// Prevent race conditions while updating bootstrapResult from multiple go-routines
```
robskillington: If we separated out a set of workers that each got a certain set of shards similar to the encoder workers we could avoid locking here too.

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +257,182 @@ func (s *commitLogSource) Read(
...
+		if blocks.Len() > 0 {
+			shardResult.AddSeries(unmergedBlocks.id, blocks)
 		}
 	}
-	if errs > 0 {
-		s.log.Errorf("error bootstrapping from commit log: %d merge out of order errors", errs)
+	return shardResult, numShardEmptyErrs, numErrs
+}
+
+func (s *commitLogSource) printMergeShardsOutcome(shardErrs []int, shardEmptyErrs []int) {
```
robskillington: Similar nit, lets call this `logMergeShardsOutcome(...)`

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +257,182 @@ func (s *commitLogSource) Read(
...
+				}
+				if iterErr := iter.Err(); iterErr != nil {
+					if err == nil {
+						err = iter.Err()
 					}
+					numErrs++
+				}
 
-					block.Reset(start, enc.Discard())
+				iter.Close()
```
robskillington: Think you need to check this for an error.

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +257,182 @@ func (s *commitLogSource) Read(
...
+				numShardEmptyErrs++
+				continue
+			} else if len(unmergedBlock) == 1 {
+				block.Reset(startInNano, unmergedBlock[0].enc.Discard())
+			} else {
+				readers := make([]io.Reader, len(unmergedBlock))
+				for i := range unmergedBlock {
+					stream := unmergedBlock[i].enc.Stream()
+					readers[i] = stream
+					defer stream.Finalize()
```
robskillington: Should not perform defer inside a huge for loop for large dynamic allocation of deferred fn calls and also just not finalizing for so long.

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Yeah good call, I have no idea why I did that

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: nice I like it.

richardartoul: ah actually, I still need access to the shard's ranges to perform the range overlap check at the end. In that case, it actually becomes slower because we do a slice lookup followed by a map lookup. I could probably copy the ranges into the unmerged slice though so that they're both available for quick lookup. Might be worth it since this is on the hot path

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Yeah I think I tried removing this at one point and was surprised that it didn't seem to have any performance impact. It was probably done this way originally because when you put the streams in the readers slice you lose access to the Finalize() method, but I was able to move it to the end manually by doing a type assertion. Let me know if it looks like I'm finalizing it at the right place, because its not happening earlier than it used to (by virtue of doing it in the middle of the loop instead of at the very end when the function returns)

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: ended up creating the encodersAndRanges struct and doing it that way

storage/bootstrap/bootstrapper/commitlog/source.go
prateek: could you update this to use the new type

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -96,56 +117,118 @@ func (s *commitLogSource) Read(
...
-		blockStart := dp.Timestamp.Truncate(blockSize)
-		blockEnd := blockStart.Add(blockSize)
-		blockRange := xtime.Range{
-			Start: blockStart,
-			End:   blockEnd,
-		}
-		if !ranges.Overlaps(blockRange) {
-			// Data in this block does not match the requested ranges
-			continue
+		// Distribute work such that each encoder goroutine is responsible for
```
prateek: nice

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +249,187 @@ func (s *commitLogSource) Read(
...
+	multiReaderIteratorPool encoding.MultiReaderIteratorPool,
+	encoderPool encoding.EncoderPool,
+	blopts block.Options,
+) (shardResult result.ShardResult, numShardEmptyErrs int, numErrs int) {
+	shardResult = result.NewShardResult(len(unmergedShard.encodersBySeries), s.opts.ResultOptions())
+	for _, unmergedBlocks := range unmergedShard.encodersBySeries {
+		blocks := block.NewDatabaseSeriesBlocks(len(unmergedBlocks.encoders))
+		for start, unmergedBlockEncoders := range unmergedBlocks.encoders {
+			startInNano := time.Unix(0, start)
+			block := blocksPool.Get()
```
prateek: there's a couple of continues in the loops below which skip over the block, should we be putting the block back into the pool in those cases?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +249,187 @@ func (s *commitLogSource) Read(
...
+	unmergedShard encodersAndRanges,
+	blocksPool block.DatabaseBlockPool,
+	multiReaderIteratorPool encoding.MultiReaderIteratorPool,
+	encoderPool encoding.EncoderPool,
+	blopts block.Options,
+) (shardResult result.ShardResult, numShardEmptyErrs int, numErrs int) {
+	shardResult = result.NewShardResult(len(unmergedShard.encodersBySeries), s.opts.ResultOptions())
+	for _, unmergedBlocks := range unmergedShard.encodersBySeries {
+		blocks := block.NewDatabaseSeriesBlocks(len(unmergedBlocks.encoders))
+		for start, unmergedBlockEncoders := range unmergedBlocks.encoders {
```
prateek: could you refactor this inner loop into smaller methods. E.g.

```
type []encoder encoders

func (e encoders) merge(...) (block.Block, error) {
// ...
}
```

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: yeah will do

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: I think you're right. The block is only added to "blocks" if no error occurs, so if any error occurs and we continue, then it should be safe to return it to the pool. I'll add that.

richardartoul: Also it looks like the iterator is never returned to the multiiteratorpool after its done. Its safe to do so once the Close() method is called right?

richardartoul: Nevermind the Close() method will return it to the pool

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: I managed to factor out the innermost loop into a new function called mergeSeries(). It's still a fairly long method, but I removed a lot of the identation and its a lot easier to read now I think

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -96,57 +117,121 @@ func (s *commitLogSource) Read(
...
+		encoderPool  = bopts.DatabaseBlockOptions().EncoderPool()
+		workerErrs   = make([]int, numConc)
 	)
-	for iter.Next() {
-		series, dp, unit, annotation := iter.Current()
 
-		// check if the series belongs to current namespace being bootstrapped
-		if !namespace.Equal(series.Namespace) {
-			continue
+	fmt.Println("Highest shard: ", highestShard)
```
prateek: do you need to drop this println?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +251,225 @@ func (s *commitLogSource) Read(
...
+	unmerged []encodersAndRanges,
+) result.BootstrapResult {
+	var (
+		shardErrs               = make([]int, numShards)
+		shardEmptyErrs          = make([]int, numShards)
+		bootstrapResult         = result.NewBootstrapResult()
+		blocksPool              = bopts.DatabaseBlockOptions().DatabaseBlockPool()
+		multiReaderIteratorPool = blopts.MultiReaderIteratorPool()
+		// Controls how many shards can be merged in parallel
+		mergeSemaphore      = make(chan bool, s.opts.MergeShardsConcurrency())
```
prateek: I think you can use a `m3x/sync.WorkerPool` for the semantics you're after.

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: ...yes. Removed it. Maybe I should write a linter for that to...

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +251,225 @@ func (s *commitLogSource) Read(
...
+		seriesBlocks, numSeriesEmptyErrs, numSeriesErrs := s.mergeSeries(
+			unmergedBlocks,
+			blocksPool,
+			multiReaderIteratorPool,
+			encoderPool,
+			blopts,
+		)
+
+		if seriesBlocks.Len() > 0 {
+			shardResult.AddSeries(unmergedBlocks.id, seriesBlocks)
```
prateek: could you change this to lazy alloc instead. i.e. only do the `result.NewShardResult(...)` call when you first attempt to add a series' blocks for a shard, don't need to allocate anything for any empty shard

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +251,225 @@ func (s *commitLogSource) Read(
...
+}
+
+func (s *commitLogSource) mergeSeries(
+	unmergedBlocks encodersByTime,
+	blocksPool block.DatabaseBlockPool,
+	multiReaderIteratorPool encoding.MultiReaderIteratorPool,
+	encoderPool encoding.EncoderPool,
+	blopts block.Options,
+) (seriesBlocks block.DatabaseSeriesBlocks, numEmptyErrs int, numErrs int) {
+	seriesBlocks = block.NewDatabaseSeriesBlocks(len(unmergedBlocks.encoders))
```
prateek: same here, could you lazy alloc for `seriesBlocks` too

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +251,225 @@ func (s *commitLogSource) Read(
...
+
+	for startNano, encoders := range unmergedBlocks.encoders {
+		start := startNano.ToTime()
+		block := blocksPool.Get()
```
prateek: would make lifecycle easier to track if you only Get a `block` once you're sure you're going to need it. e.g.

```
for ... {
  start := .. 
  
  if len(encoders) == 0 {
    numEmptyErrs++
    continue
  }

  if len(encoders) == 1 {
    block := blocksPool.Get()
    block.Reset(start, ...)
    seriesBlocks.AddBlock(block)
    continue
  }
  
  // .. all the []encoder => encoder conversions 
 
  // .. encoder/iter cleanup 
  
  block := blocksPool.Get(..)
  // .... 
}
```

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +251,225 @@ func (s *commitLogSource) Read(
...
+		}
+
+		if len(encoders) == 1 {
+			block.Reset(start, encoders[0].enc.Discard())
+			seriesBlocks.AddBlock(block)
+			continue
+		}
+
+		// Convert encoders to readers so we can use iteration helpers
+		readers := make([]io.Reader, len(encoders))
```
prateek: Can you break this up a little further by making some utility methods like:

```
type encoders []encoder

type ioReaders []io.Reader

func (e encoders) NewReaders() ioReaders {
  readers := make(ioReaders, len(e))
  for i := range e {
    readers[i] = e[i].enc.Stream()    
  }
  return readers
}

func (e encoders) Close() {
  for i := range e {
    e[i].enc.Close()
  }
}

func (ir ioReaders) Close() {
  for _, r := range ir {
    r.(xio.SegmentReader).Finalize()
  }
}

```

storage/bootstrap/bootstrapper/commitlog/source.go
prateek: lol. I usually put a `TODO(prateek): ` when i leave something around and then it's obvious I have to clean it up.

storage/bootstrap/bootstrapper/commitlog/source.go
prateek: Yeah it's way easier to grok

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: This required an extra nil check in the caller, but I added it

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: This required an extra nil check in the caller, but I added it

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: good call

storage/bootstrap/bootstrapper/commitlog/options.go
```diff
@@ -26,15 +26,19 @@ import (
...
 }
 
 // NewOptions creates new bootstrap options
 func NewOptions() Options {
 	return &options{
-		resultOpts:    result.NewOptions(),
-		commitLogOpts: commitlog.NewOptions(),
+		resultOpts:            result.NewOptions(),
+		commitLogOpts:         commitlog.NewOptions(),
+		encodingConcurrency:   4,
```
prateek: could you declare consts for these vars at the top of the file

```
const (
  defaultEncodingConcurrency = 4
  ... 
)
```

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,235 @@
...
+	"io/ioutil"
+	"reflect"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/leanovate/gopter"
+	"github.com/leanovate/gopter/gen"
+	"github.com/leanovate/gopter/prop"
+	"github.com/m3db/m3db/context"
```
prateek: nit: separate m3db org's import group from 3rd party ones

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+
+	idx = metricIdx.idx
+	metricIdx.idx++
+	metricIdx.idToIdx[s] = idx
+	return idx
+}
+
+// idToShard ensures that each string series ID maps to exactly one shard
+func idToShard(s string) uint32 {
```
prateek: Why do the id->shard assignment like this instead of `hash(id) % numShards` ?

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,235 @@
...
+var metricIdx = globalMetricIdx{
+	idToIdx: make(map[string]uint64),
+}
+
+var metricShard = globalMetricShard{
+	idToShard: make(map[string]uint32),
+}
+
+// idToIDX ensures that each string series ID maps to exactly one UniqueIndex
+func idToIdx(s string) uint64 {
```
prateek: nit: rename to something like `func seriesUniqueIndex(series string) uint64`

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,235 @@
...
+		start := inputs[0].(time.Time)
+		numDatapoints := inputs[1].(int)
+		return genPropTestInput(start, numDatapoints, ns)
+	}
+	return gopter.CombineGens(
+		// Runs iterations of the test starting 1000 hours in the past/future
+		gen.TimeRange(time.Now().Add(-1000*time.Hour), 1000*time.Hour),
+		// Run iterations of the test with between 0 and 1000 datapoints
+		gen.IntRange(0, 1000),
+	).FlatMap(curriedGenPropTestInput, reflect.TypeOf(propTestInput{}))
```
prateek: nifty

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,235 @@
...
+func TestCommitLogSourcePropCorrectlyBootstrapsFromCommitlog(t *testing.T) {
+	parameters := gopter.DefaultTestParameters()
+	parameters.Rng.Seed(123456789)
+	parameters.MinSuccessfulTests = 100
+	props := gopter.NewProperties(parameters)
+
+	props.Property("Commitlog bootstrapping properly bootstraps the entire commitlog", prop.ForAll(
+		func(input propTestInput) (bool, error) {
+			// Make sure we have a new directory for each test run
+			dir, err := ioutil.TempDir("", "foo")
```
prateek: do you need to cleanup this dir once you're done?

prateek: Won't this cause issues because it generates a random directory id internally. i.e. gopter shrinking/repro should fail no?

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,235 @@
...
+	curriedGenPropTestInput := func(input interface{}) gopter.Gen {
+		inputs := input.([]interface{})
+		start := inputs[0].(time.Time)
+		numDatapoints := inputs[1].(int)
+		return genPropTestInput(start, numDatapoints, ns)
+	}
+	return gopter.CombineGens(
+		// Runs iterations of the test starting 1000 hours in the past/future
+		gen.TimeRange(time.Now().Add(-1000*time.Hour), 1000*time.Hour),
+		// Run iterations of the test with between 0 and 1000 datapoints
```
prateek: maybe have a higher lower bound; [100,1000] instead of [0, 1000]?

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,235 @@
...
+			for _, write := range input.writes {
+				log.Write(context.NewContext(), write.series, write.datapoint, write.unit, write.annotation)
+			}
+			err = log.Close()
+			if err != nil {
+				return false, err
+			}
+
+			// Instantiate a commitlog source
+			source := newCommitLogSource(bootstrapOpts)
```
prateek: Could you use the exported ctor here instead?

storage/bootstrap/bootstrapper/commitlog/options.go
```diff
@@ -25,16 +25,25 @@ import (
...
 
 // NewOptions creates new bootstrap options
 func NewOptions() Options {
 	return &options{
-		resultOpts:    result.NewOptions(),
-		commitLogOpts: commitlog.NewOptions(),
+		resultOpts:            result.NewOptions(),
+		commitLogOpts:         commitlog.NewOptions(),
+		encodingConcurrency:   defaultEncodingConcurrency,
+		mergeShardConcurrency: defaultMergeShardConcurrency,
```
prateek: maybe check if the `concurrencies are >= 1` in `Validate()`

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,235 @@
+package commitlog
```
prateek: could you add the license header

storage/bootstrap/bootstrapper/commitlog/options.go
richardartoul: done

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: 1) Regarding the temp dir, do I? I looked through existing test cases and they don't look like they cleanup old directories. I figured the O.S would just reclaim it eventually

2) Does the random directory ID matter? This test writes the commitlog files and reads them back all in one go. If gopter is trying to repro, it will just be producing new data into new folders which should be fine. Maybe I'm misunderstanding something?

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: done

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: done

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: done

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: i actually thought zero datapoints was an interesting edge-case to include

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: done

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: done

richardartoul: (I was just being lazy before)

richardartoul:
@prateek Just an FYI this PR failed C.I because master is broken, not because there is anything wrong with this P.R

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -49,9 +54,50 @@ type encoder struct {
 	enc         encoding.Encoder
 }
 
-type encoderMap struct {
-	id       ts.ID
-	encoders map[time.Time][]encoder
+type encodersAndRanges struct {
```
prateek: super nit: could you move the helper type decls and methods to the bottom of the file. Couple of rules of thumb (apply with discretion obv) -
- Try to have related code close by in a single file (i.e. two methods on the same object, or the exported+helper methods)
- Order the contents of a file's types/methods in decreasing order of importance (so primary type/method at the top, helper types/methods at the bottom).

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -96,57 +142,119 @@ func (s *commitLogSource) Read(
...
-			unmergedShard = make(map[ts.Hash]encoderMap)
-			unmerged[series.Shard] = unmergedShard
+	// Block until all data has been read and encoded by the worker goroutines
+	wg.Wait()
+	s.logEncodingOutcome(workerErrs, iter)
+
+	return s.mergeShards(int(highestShard+1), bopts, blopts, encoderPool, unmerged), nil
+}
+
+func (s *commitLogSource) findHighestShard(shardsTimeRanges result.ShardTimeRanges) uint32 {
```
prateek: per above, move towards bottom of the file

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -96,57 +142,119 @@ func (s *commitLogSource) Read(
...
 	var (
-		namespace   = ns.ID()
-		unmerged    = make(map[uint32]map[ts.Hash]encoderMap)
-		bopts       = s.opts.ResultOptions()
-		blopts      = bopts.DatabaseBlockOptions()
-		blockSize   = ns.Options().RetentionOptions().BlockSize()
-		encoderPool = bopts.DatabaseBlockOptions().EncoderPool()
-		errs        = 0
+		namespace    = ns.ID()
+		highestShard = s.findHighestShard(shardsTimeRanges)
```
prateek: maybe just make this `numShards = s.findHighestShard(...) + 1` and use numShards in the two places below?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -166,93 +274,228 @@ func (s *commitLogSource) Read(
...
+		pooledBlock := blocksPool.Get()
+		pooledBlock.Reset(start, enc.Discard())
+		if seriesBlocks == nil {
+			seriesBlocks = block.NewDatabaseSeriesBlocks(len(unmergedBlocks.encoders))
+		}
+		seriesBlocks.AddBlock(pooledBlock)
 	}
-	if errs > 0 {
-		s.log.Errorf("error bootstrapping from commit log: %d merge out of order errors", errs)
+	return seriesBlocks, numEmptyErrs, numErrs
```
prateek: Weird to be using named return vars and overriding them instead of a naked return. Could you pick one convention

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
prateek: Cool, fair enough

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
prateek: Yep you're right. It's an issue if you're using a SUT, not with this flavor of gopter.

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,255 @@
...
+}
+
+type globalMetricIdx struct {
+	sync.Mutex
+
+	idx     uint64
+	idToIdx map[string]uint64
+}
+
+type globalMetricShard struct {
```
prateek: do you still need this type?

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,255 @@
...
+
+	shard     uint32
+	idToShard map[string]uint32
+}
+
+var metricIdx = globalMetricIdx{
+	idToIdx: make(map[string]uint64),
+}
+
+var metricShard = globalMetricShard{
```
prateek: this one too?

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,255 @@
...
+				return false, err
+			}
+			err = log.Open()
+			if err != nil {
+				return false, err
+			}
+
+			// Write all the datapoints to the commitlog
+			for _, write := range input.writes {
+				log.Write(context.NewContext(), write.series, write.datapoint, write.unit, write.annotation)
```
prateek: do you need to check the error returned here?

richardartoul: fixed

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: fixed

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: fixed

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: good idea, fixed

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: fixed

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: fixed

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -0,0 +1,247 @@
...
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/leanovate/gopter"
+	"github.com/leanovate/gopter/gen"
+	"github.com/leanovate/gopter/prop"
+	"github.com/spaolacci/murmur3"
+
+	"github.com/m3db/m3db/context"
```
prateek: super nit: order internal imports above 3rd party ones

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: I actually kind of like this pattern. Naked returns are generally frowned upon for long functions, but I do like being able to name the returns because it makes the function signature cleaner

richardartoul: Can change it to a naked return if you feel strongly

storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: fixed

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -97,56 +97,108 @@ func (s *commitLogSource) Read(
...
-		errs        = 0
+		workerErrs  = make([]int, numConc)
 	)
-	for iter.Next() {
-		series, dp, unit, annotation := iter.Current()
 
-		// check if the series belongs to current namespace being bootstrapped
-		if !namespace.Equal(series.Namespace) {
-			continue
+	// +1 so we can use the shard number as an index throughout without constantly
```
prateek: nit: move comment adjacent to numShards (better) or delete

richardartoul: fixed

storage/bootstrap/bootstrapper/commitlog/source.go
prateek: Yeah call me c99 bent but I don't like named returns ever (and naked returns as a result). Could you declare the vars instead of using named returns, and then leave the return as is? 

The concern is if someone adds a return to early terminate the function without realising the lifecycle of the named return vars

prateek:
@richardartoul mainly looks good, could you rebase/merge master to pickup the ci/broken-test fixes

richardartoul:
@prateek Yep rebasing right now to fix conflicts and tests

prateek:
@richardartoul once you're done with the changes. might be worthwhile to attempt a commit log bootstrap on one of the prod app/hist nodes and see how long it takes

PR #324, opened 2017-07-31
Add prop tests for Commit Log

prateek:
- e2e prop test for commit log writer/iterator 

/cc @robskillington

### DIFF
--- a/persist/fs/commitlog/read_write_prop_test.go
+++ b/persist/fs/commitlog/read_write_prop_test.go
@@ -41,11 +41,8 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var (
-	defaultCommitLogSeed     int64  = 12345678
-	defaultShardNum          uint32 = 1234
-	defaultTestNamespace            = ts.StringID("testNamespace")
-	defaultTestFlushInterval        = time.Millisecond
+const (
+	defaultTestFlushInterval = time.Millisecond
 )
 
 func TestCommitLogReadWrite(t *testing.T) {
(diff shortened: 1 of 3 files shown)


persist/fs/commitlog/read_write_prop_test.go
```diff
@@ -0,0 +1,369 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package commitlog_test
```
robskillington: Hm shouldn't this just be `commitlog`?

persist/fs/commitlog/read_write_prop_test.go
```diff
@@ -0,0 +1,369 @@
...
+	return &commands.ProtoCommands{
+		NewSystemUnderTestFunc: func(initialState commands.State) commands.SystemUnderTest {
+			return initialState
+		},
+		DestroySystemUnderTestFunc: func(sut commands.SystemUnderTest) {
+			state := sut.(*clState)
+			if cl := state.cLog; cl != nil {
+				cl.Close()
+			}
+			os.RemoveAll(state.opts.FilesystemOptions().FilePathPrefix())
```
robskillington: Seems you removeall here and also the directory containing these files in the tests themselves?

persist/fs/commitlog/read_write_prop_test.go
prateek: Using `_test` makes go treat this file as if it were in an external package forcing me to rely on only exported methods/types/... It feels more general to keep the test like this.

persist/fs/commitlog/read_write_prop_test.go
prateek: Yep so need this one because gopter can re-create the same system multiple times, want each run to have a clean directory to work with. 

The other one is me being paranoid, in case there's a gopter failure mode we haven't run into yet.

persist/fs/commitlog/read_write_prop_test.go
robskillington: Ok sure thing, in future we might want to write up our recommended best practices for prop tests.

prateek: Will do, opened #326 to track.

PR #325, opened 2017-08-07
Use a more efficient mechanism writing commit log series metadata

robskillington:
This change removes the use of bitset series metadata written cache in the commit log writer.  Using a fixed size array and then mapping the series into that array with potential collisions allows to know when a series was definitely seen or not.  This means some series may write its metadata multiple times during collisions to the commit log, however should reserve the vast majority of space savings made by keeping a metadata written cache.

FWIW this adds roughly 10-25% increase to the size of commit logs, the read speed should be the same as it doesn't decode the metadata multiple times even if it appears multiple times.  Completely disabling the seen cache at all increased the size of commit logs by 800%+ so this seems like a good trade off that avoids any allocations whatsoever during times of heavy inserts.

I've also refactored some of the shard insert logic to clean it up and avoid any asynchronous inserting during bootstrapping in preparation for adding insert queue backoff and limiting in the next change.

cc @xichen2020 @martin-mao @prateek


persist/fs/commitlog/reader.go
```diff
@@ -142,7 +143,9 @@ func (r *reader) Read() (
 		}
 	}
 
-	metadata, ok := r.metadataLookup[entry.Index]
+	if !ok {
+		metadata, ok = r.metadataLookup[entry.Index]
```
xichen2020: nit: Move this to the `if` block above to save a map lookup
```
if !ok && len(entry.Metadata) != 0 {
  ...
  metadata = Series {...}
  r.metadataLookup[entry.Index] = metadata
  ok = true
}
```
and get rid of 
```
if !ok {
  metadata, ok = r.metadataLookup[entry.Index]
}
```

persist/fs/commitlog/writer.go
```diff
@@ -154,19 +157,38 @@ func (w *writer) isOpen() bool {
...
-		// If "idx" hasn't been written to commit log
+	logEntry.Index = logEntryIndex
+
+	// NB(r): Calculate whether metadata for this series
+	// has been written before. Once we have more unique series than
+	// the size of the seen cache this can cause some metadata
+	// to be written multiple times. The vast majority
+	// of IDs should not collide frequently which preserves
+	// the size of the resulting commit log staying small.
+	seenIdx := logEntryIndex % uint64(w.seenCacheLen)
```
xichen2020: Just throwing it out there, would a bloom filter help reduce the false positive rate here?

persist/fs/commitlog/writer.go
```diff
@@ -191,9 +213,9 @@ func (w *writer) Write(
 		return err
 	}
 
-	if !seen {
-		// Record we have seen this series
-		w.bitset.set(logEntry.Index)
+	if !definitelySeen {
+		// Record we have seen this series at the current seenIdx
+		w.seenCache[seenIdx] = logEntryIndex
```
xichen2020: Any reason why we do a separate check here as opposed to just doing this in the `if !definitelySeen {...}` block?

persist/fs/commitlog/writer.go
```diff
@@ -154,19 +157,38 @@ func (w *writer) isOpen() bool {
 	return w.chunkWriter.fd != nil
 }
 
+func (w *writer) logEntryIndex(seriesUniqueIndex uint64) uint64 {
+	// Special case the zeroth value as we compare the value
+	// at the entry of the seen cache with the index being put
+	// at the location - since all entries in the seen cache when
+	// not set are zero the value being used to compare whether
+	// this result has been seen or not cannot be zero itself
+	return seriesUniqueIndex + 1
```
prateek: I don't follow, won't the '%' during `seenIdx` computation wrap around anyway?

persist/fs/commitlog/writer.go
```diff
@@ -154,19 +157,38 @@ func (w *writer) isOpen() bool {
...
+}
+
 func (w *writer) Write(
 	series Series,
 	datapoint ts.Datapoint,
 	unit xtime.Unit,
 	annotation ts.Annotation,
 ) error {
+	logEntryIndex := w.logEntryIndex(series.UniqueIndex)
```
prateek: Would it be worthwhile to also make the `series -> UniqueIndex` computation deterministic (using a hash for e.g.)? As long we didn't see many collisions, we'd be less likely to write extra metadata if we did.

persist/fs/commitlog/writer.go
robskillington: Yes but that's the index when looking up into the `seenCache`.  We use the return value of this method to check the value in the `seenCache` matches it.  i.e. was this the last series to use this slot.

persist/fs/commitlog/reader.go
robskillington: Sure, can do.

persist/fs/commitlog/writer.go
robskillington: The problem with bloom filters is that they can return false positives.  It will sometimes return "this might be in this set".  For our use case we can't afford to miss the case of it actually not being in the set as that means we won't attach metadata for that series and it will be missing when we try to read the commit log.

Hence me using this approach, which is kind of similar to the approach others have taken in this scenario:
https://www.somethingsimilar.com/2012/05/21/the-opposite-of-a-bloom-filter/

persist/fs/commitlog/writer.go
robskillington: We only want to add it to the `seenCache` if there was no error writing it to the log.

persist/fs/commitlog/writer.go
xichen2020: Yeah agreed.

persist/fs/commitlog/writer.go
xichen2020: Yeah makes sense.

persist/fs/commitlog/writer.go
robskillington: That would only redistribute the layout of ID -> index in the seen cache, which for the cost of a hash doesn't seem worth it.  I hashed 100 million IDs in a row and it took a long time vs just doing a modulus and an array lookup/insertion was hundreds of milliseconds, i.e. hash approach took 10s for 100 million and modulus lookup/insertion was 400ms.

storage/shard.go
```diff
@@ -786,15 +854,29 @@ func (s *dbShard) Bootstrap(
...
+			// the insert queue potential delayed insert
+			entry, err = s.insertSeriesSync(dbBlocks.ID, insertSyncIncWriterCount)
+			if err != nil {
+				multiErr = multiErr.Add(err)
+				continue
+			}
+		}
 
 		err = entry.series.Bootstrap(dbBlocks.Blocks)
+		if err != nil {
```
prateek: nit: multiErr is already doing the [nil check](https://github.com/m3db/m3x/blob/master/errors/errors.go#L215), could make this a little simpler by doing:

```
multiErr = multiErr.Add(entry.series.Bootstrap(dbBlocks.Blocks)
```

storage/shard.go
```diff
@@ -625,40 +627,109 @@ func (s *dbShard) tryRetrieveWritableSeries(id ts.ID) (
...
+func (s *dbShard) insertSeriesBatch(inserts []dbShardInsert) error {
+	anyPendingWrites := false
+
 	s.Lock()
 	for i := range inserts {
+		// If we are going to write to this entry then increment the
+		// writer count so it does not look empty immediately after
+		// we release the write lock
+		hasPendingWrite := inserts[i].hasPendingWrite
+		if hasPendingWrite && !anyPendingWrites {
```
prateek: nit: can avoid the branching by always or-ing:

```
anyPendingWrites = anyPendingWrites || hasPendingWrite
```

storage/shard.go
```diff
@@ -673,16 +744,13 @@ func (s *dbShard) insertSeriesEntries(inserts []dbShardInsert) error {
...
-		if inserts[i].hasPendingWrite {
-			entry.incrementWriterCount()
-		}
 	}
 	s.Unlock()
 
+	if !anyPendingWrites {
+		return nil
+	}
+
```
prateek: I know this isn't in your changes, but should this method capture and return the errors from `entry.series.Write` in the loop below this line?

xichen2020:
changes LGTM, are the failing tests related?

persist/fs/commitlog/writer.go
prateek: Ah makes sense. Maybe re-word the comment to clarify why we're doing this, it took me a minute to figure out what you meant. Something like: `This value is used as the series' unique index in the seenCache. The seenCache is zero initialised, so we add one to ensure the value we set in the slice is non-zero. This lets us distinguish un-initialised v initialised seenCache entries.`

persist/fs/commitlog/writer.go
prateek: I was more concerned about the thrashing in the seenCache once go past capacity. Using the modulo would lead to a cascade of thrashing, any series could thrash an entry owned by another one. The hash approach would minimise that thrashing by being deterministic about which ids thrashed each other. Then again, it could also cause a lot more thrashing because two ids with the same hash could be written to very frequently. 

Might be best to just do the simple thing (like in the PR) and track a metric for thrashing, and address if it becomes an issue?

persist/fs/commitlog/writer.go
robskillington: Sure thing.

persist/fs/commitlog/writer.go
robskillington: Sounds good re: metric tracking thrashes.  Yeah this should be deterministic as well as two IDs that resolve to the same entry when modulo their unique index by the seen cache will collide and override each other.  

Whether you hash it or not it is deterministic placing, the only thing that would help fix the frequently appearing series overriding each other would be an LRU - but that is super expensive by comparison.

storage/shard.go
robskillington: Sure thing, that's how it was before - just wanted to be more Go-ish but I'll change this back =]

storage/shard.go
robskillington: Sure thing - looks better.

storage/shard.go
robskillington: Async inserts by their very nature do not really have a good place to surface.  I can add a metric however.

robskillington:
@prateek any ideas on the failing prop test?  Doesn't seem to fail locally for me.  Maybe it just takes too long?  Sometimes it takes 7-8 seconds locally for me, maybe travis is just too slow?

Link to failing part of log:
https://travis-ci.org/m3db/m3db/builds/262040390#L1503
```
=== RUN   TestCommitLogPropTest
panic: test timed out after 1m0s

...
goroutine 364 [running]:
...
```

prateek:
@robskillington Yeah I think it's the combination of Travis being slow, and the prop tests running in `-race` mode. Weird that they didn't fail in the earlier PR. Could you try reducing the number of successful test runs (search for "parameters.MinSuccessfulTests" or increasing the timeout in `ci-scripts/test-cover.sh`).

PR #327, opened 2017-08-08
Add a runtime configurable limit and backoff for new series inserts

robskillington:
This change adds a runtime configurable new series insert limit and also a backoff to avoid new series inserts starving the shard map lookup for existing series writes.  By default the backoff is installed as 2ms to allow for 500 batched inserts per second at peak new insert throughput.

The idea is to set the limit in config so that we can divide a global limit by the number of shards a topology has.  Have tested with these settings and they look to be optimal.

cc @xichen2020 @prateek @martin-mao


storage/shard.go
```diff
@@ -198,12 +198,21 @@ func newDatabaseShard(
...
-		metrics:               newDbShardMetrics(scope),
+		metrics:               newDatabaseShardMetrics(scope),
 	}
-	d.insertQueue = newDbShardInsertQueue(d.insertSeriesBatch, scope)
-	d.insertQueue.Start()
+	d.insertQueue = newDatabaseShardInsertQueue(d.insertSeriesBatch,
+		d.nowFn, scope)
+
+	registerRuntimeOptionsListener := func(listener runtime.OptionsListener) {
+		elem := opts.RuntimeOptionsManager().RegisterListener(d)
```
xichen2020: Should this be `RegisterListener(listener)`?

robskillington: Ha, good catch!  Cheers.

xichen2020:
Changes LGTM, failing test is not related?

Also do we have a cluster-wide endpoint for setting the rate limits? That might come in handy if we need to set it on all nodes in emergency situations.

robskillington:
Missing a newly generated file, reran mock-gen with the latest definitions for the runtime options package.

PR #330, opened 2017-08-10
Add `metalinter: misspell, unconvert` (3/n)

prateek:
PR before this: #329

/cc @robskillington

### DIFF
--- a/encoding/m3tsz/iterator.go
+++ b/encoding/m3tsz/iterator.go
@@ -255,7 +255,7 @@ func (it *readerIterator) readIntValDiff() {
 		sign = 1.0
 	}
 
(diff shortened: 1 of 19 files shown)


encoding/m3tsz/iterator.go
```diff
@@ -255,7 +255,7 @@ func (it *readerIterator) readIntValDiff() {
 		sign = 1.0
 	}
 
-	it.intVal += sign * float64(uint64(it.readBits(int(it.sig))))
+	it.intVal += sign * float64(it.readBits(int(it.sig)))
```
robskillington: This is because readBits already returns `uint64` yeah?

prateek: yessir

PR #331, opened 2017-08-10
Add `metalinter: gosimple` (4/n)

prateek:
PR before this: #330 

/cc @robskillington


client/session_fetch_bulk_blocks_test.go
```diff
@@ -589,7 +589,7 @@ func TestSelectBlocksForSeriesFromPeerBlocksMetadataAllPeersSucceed(t *testing.T
 	session := s.(*session)
 
 	var (
-		metrics          = session.streamFromPeersMetricsForShard(0, resultTypeTest)
+		metrics          = session.streamFromPeersMetricsForShard(0, resultTypeRaw)
```
jeromefroe: Are these changes from the `metalinter` as well?

prateek: Yep.`resultTypeTest` was only being used by tests, changed it to `resultTypeRaw` and eliminated `resultTypeTest`

PR #332, opened 2017-08-10
Annotate entries with commit log seen state to remove seen cache

robskillington:
This removes the seen cache altogether from the commit log and keeps the commit log series metadata written state in the already allocated DB shard entry for the series.

cc @prateek @xichen2020 @martin-mao


prateek:
LGTM

PR #336, opened 2017-08-18
Revert the seen series store to a bitset for the commit log

robskillington:
Reverting the seen series storage for a given commit log to be a bitset.  This bitset implementation is a derivation of the previously used bitset library.  It makes it both a little simpler and also enforces that the growth of the bitset is 2x required length, instead of dependent on the library version.

Even with the changes to fix the memory locality when testing whether a series has already been written to the current commit log, the updating of that data still has very poor memory locality as the source of truth still sits on the heap across a large part of the memory of the process.  The bitset implementation originally set in place allows for much better locality with the caches close to the CPU, even if it is not as clean as keeping the relevant data next to the series themselves.

cc @xichen2020 @prateek @martin-mao

### DIFF
--- a/persist/fs/commitlog/bitset.go
+++ b/persist/fs/commitlog/bitset.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Uber Technologies, Inc.
+// Copyright (c) 2017 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/persist/fs/commitlog/bitset.go
+++ b/persist/fs/commitlog/bitset.go
@@ -1,0 +1,64 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package commitlog
+
+const (
+	wordSize      = uint(64)
+	logWordSize   = uint(6) // lg(wordSize)
+	totalCapacity = ^uint(0)
+)
+
+type bitSet struct {
+	values []uint64
+}
+
+func newBitSet(size uint) *bitSet {
+	return &bitSet{values: make([]uint64, bitSetIndexOf(size)+1)}
+}
+
+func (b *bitSet) test(i uint) bool {
+	idx := bitSetIndexOf(i)
+	if idx >= len(b.values) {
+		return false
+	}
+	return b.values[idx]&(1<<(i&(wordSize-1))) != 0
+}
+
+func (b *bitSet) set(i uint) {
+	idx := bitSetIndexOf(i)
+	currLen := len(b.values)
+	if idx >= currLen {
+		newValues := make([]uint64, 2*(idx+1))
+		copy(newValues, b.values)
+		b.values = newValues
+	}
+	b.values[idx] |= 1 << (i & (wordSize - 1))
+}
+
+func (b *bitSet) clearAll() {
+	for i := range b.values {
+		b.values[i] = 0
+	}
+}
+
+func bitSetIndexOf(i uint) int {
+	return int(i >> logWordSize)
+}
--- a/persist/fs/commitlog/bitset.go
+++ b/persist/fs/commitlog/bitset.go
@@ -21,9 +21,8 @@
 package commitlog
 
 const (
-	wordSize      = uint(64)
-	logWordSize   = uint(6) // lg(wordSize)
-	totalCapacity = ^uint(0)
+	wordSize    = uint(64)
+	logWordSize = uint(6) // lg(wordSize)
 )
 
 type bitSet struct {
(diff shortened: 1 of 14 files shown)


persist/fs/commitlog/bitset_test.go
```diff
@@ -0,0 +1,176 @@
...
+				if !ok {
+					return gopter.NewEmptyResult(reflect.PtrTo(reflect.TypeOf(bitSetTestState{})))
+				}
+				state := bitSetTestState{
+					set:     newBitSet(v),
+					currSet: make(map[uint]struct{}),
+				}
+				return gopter.NewGenResult(state, gopter.NoShrinker)
+			}),
+		InitialPreConditionFunc: func(state commands.State) bool {
```
prateek: don't need to specify the pre/post condition function if it's always true

persist/fs/commitlog/bitset_test.go
```diff
@@ -0,0 +1,176 @@
...
+	currSet map[uint]struct{}
+}
+
+var genBitSetTestCommand = gen.
+	UIntRange(0, bitSetPropTestMaxValueLength).
+	Map(func(v uint) commands.Command {
+		return &commands.ProtoCommand{
+			Name: "Test",
+			PreConditionFunc: func(state commands.State) bool {
+				return true
```
prateek: same here

persist/fs/commitlog/bitset_test.go
```diff
@@ -0,0 +1,176 @@
...
+		}
+	})
+
+var genBitSetSetCommand = gen.
+	UIntRange(0, bitSetPropTestMaxValueLength).
+	Map(func(v uint) commands.Command {
+		return &commands.ProtoCommand{
+			Name: "Set",
+			PreConditionFunc: func(state commands.State) bool {
+				return true
```
prateek: here too

persist/fs/commitlog/bitset_test.go
```diff
@@ -0,0 +1,176 @@
...
+					Error:  result.(error),
+				}
+			},
+		}
+	})
+
+var genBitSetClearAllCommand = gen.Const(&commands.ProtoCommand{
+	Name: "ClearAll",
+	PreConditionFunc: func(state commands.State) bool {
+		return true
```
prateek: here too

persist/fs/commitlog/bitset_test.go
```diff
@@ -0,0 +1,176 @@
...
+			delete(s.currSet, k)
+		}
+		return s
+	},
+	PostConditionFunc: func(state commands.State, result commands.Result) *gopter.PropResult {
+		s := state.(bitSetTestState)
+		for i := range s.set.values {
+			if s.set.values[i] != 0 {
+				return &gopter.PropResult{
+					Status: gopter.PropFalse,
```
prateek: I'd suggest using `gopter.PropError`. With `PropFalse` `gopter` will fail but won't log the error

persist/fs/commitlog/bitset.go
```diff
@@ -0,0 +1,63 @@
...
+
+package commitlog
+
+const (
+	wordSize    = uint(64)
+	logWordSize = uint(6) // lg(wordSize)
+)
+
+type bitSet struct {
+	values []uint64
```
prateek: Could you emit a metric to indicate bitSet size

persist/fs/commitlog/bitset.go
```diff
@@ -0,0 +1,63 @@
...
+}
+
+func newBitSet(size uint) *bitSet {
+	return &bitSet{values: make([]uint64, bitSetIndexOf(size)+1)}
+}
+
+func (b *bitSet) test(i uint) bool {
```
prateek: - Could you add sanity unit tests in addition to the prop tests just so we're sure the prop tests themselves aren't missing anything
- Could you add the benchmarks from https://github.com/willf/bitset/pull/45/files too

persist/fs/commitlog/bitset.go
```diff
@@ -0,0 +1,63 @@
...
+	idx := bitSetIndexOf(i)
+	if idx >= len(b.values) {
+		return false
+	}
+	return b.values[idx]&(1<<(i&(wordSize-1))) != 0
```
xichen2020: Shouldn't this just be `b.values[idx] & (i & (wordSize - 1))`? e.g., if i is 127, you want to bit and with 63, not 1<<63, no?

Also would suggest making `wordSize - 1` a const and name it `wordMask`, I don't trust compilers but your call.

persist/fs/commitlog/bitset.go
```diff
@@ -0,0 +1,63 @@
...
+
+func (b *bitSet) set(i uint) {
+	idx := bitSetIndexOf(i)
+	currLen := len(b.values)
+	if idx >= currLen {
+		newValues := make([]uint64, 2*(idx+1))
+		copy(newValues, b.values)
+		b.values = newValues
+	}
+	b.values[idx] |= 1 << (i & (wordSize - 1))
```
xichen2020: Same, shouldn't it be `b.value[idx] |= i & (wordSize -1)`?

persist/fs/commitlog/bitset.go
```diff
@@ -0,0 +1,63 @@
...
+// THE SOFTWARE.
+
+package commitlog
+
+const (
+	wordSize    = uint(64)
+	logWordSize = uint(6) // lg(wordSize)
+)
+
+type bitSet struct {
```
prateek: could we save some space by changing this a type alias to `[]uint64`?

prateek: turns out, nope. the go compiler optimises the storage and access to the be same in either case.

persist/fs/commitlog/bitset.go
robskillington: Done both sounds good.

persist/fs/commitlog/bitset_test.go
robskillington: Done sounds good.

persist/fs/commitlog/bitset_test.go
robskillington: Done, sounds good.

persist/fs/commitlog/bitset_test.go
robskillington: Done, sounds good.

persist/fs/commitlog/bitset_test.go
robskillington: Done, sounds good.

persist/fs/commitlog/bitset_test.go
robskillington: Done, sounds good.

persist/fs/commitlog/bitset.go
robskillington: Sure thing.

persist/fs/commitlog/bitset.go
robskillington: Don't really want to as it will add an atomic op each commit log write, we can easily heapdump to see the size of bitset on demand.  Seems a bit overkill to also be graphing this.

prateek: I was thinking more like the commit log emits `len(bitSet.values)` every 10 seconds or every time we flush, which ever is easier. Would be useful to have on the top level dash if it isn't too expensive to do.

persist/fs/commitlog/bitset.go
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
xichen2020: Should probably be `2017`?

robskillington: Sure thing.

PR #338, opened 2017-08-24
Increase test coverage for namespace kv config serialisation/deserialisation

prateek:
- Usability tweak for kv configuration ~ using `mins` rather than `nanos` in the config parameters.
- Adding tests (unit/prop)


robskillington:
I'm not opposed to the tests, but the change from nanos to minutes change seems to remove the ability to use sub-minute block sizes, etc.  Which might be useful in embedded cases, etc.

It'd be better to simply make the tooling better rather than the underlying datastructures more friendly (i.e. make the tooling have an hours, minutes, seconds, etc flag or take a time.Duration using time.ParseDuration to be able to specify "1m10s" etc).

prateek:
@robskillington that's fair, there's no need to loose generality here. addressed in the new patch

PR #340, opened 2017-08-30
Update m3cluster version

cw9:
Pick up some recent interface changes

@robskillington @xichen2020 @prateek


glide.lock
```diff
@@ -206,6 +213,8 @@ imports:
   - internal/log
   - internal/modules
   - internal/remote_api
+- name: google.golang.org/appengine/datastore
```
xichen2020: Need to pin dependencies whose commits SHA have changed?

cw9: This one is already pinned in glide.yaml, not sure why it's not showing up in old glide.lock, gonna pin the other dependencies.

PR #341, opened 2017-08-31
Update m3cluster with more refactoring

cw9:
@xichen2020 @robskillington @prateek


cw9:
This PR updates interfaces only

Issue #290, opened 2017-03-02
change map[time.Time]XXX to map[int64]XXX


cw9:
There were cases where we could not find a hit in map for the same time but different timezone, we should use UnixNanoSeconds as the key to represent a time

robskillington:
We need to audit a lot of the code to ensure we are ensuring UTC in <- and UTC out -> and then we can begin to use map[int64]... once we are sure everything comes in as UTC time.

richardartoul:
My profiling over the last three days shows that you'll get a slight performance boost once you do this since the go compiler has a specialized access pattern for map[int64]

PR #345, opened 2017-09-10
[travis] Add go 1.8 and 1.9 to travis tests

jeromefroe:
This PR adds Go 1.8 and 1.9 to the versions of Go we test against in Travis.

Update: Go 1.9 changed the internal representation of `time.Time` such that `Time` values, even if they represent the same instant, are unlikely to be equal according to `reflect.DeepEqual` which the `testify` package uses under the hood to test if two values are equal. There's an [open PR](https://github.com/stretchr/testify/pull/394) to address this problem by first checking if the values being compared have an `Equal` method before falling back to comparing basic types. Admittedly the PR hasn't gotten much traction though the development of packages such as https://github.com/google/go-cmp, which aim to provide a more powerful alternative to `reflect.DeepEqual`, may help the cause. 

With that being said, I started updating the tests which compare times using `assert.Equal` to instead use `assert.True` and call the `Equal` method on the values being compared, when I noticed we have a more insidious problem: we use `time.Time` as keys in maps (a grep of `map[time.Time]` revealed 60 such occurrences). [According to the spec](https://golang.org/ref/spec#Comparison_operators) struct values are compared by all of their fields and not by an `Equal` method if one is defined, so all of such maps are likely to break in Go 1.9. 

One possible solution is that we can convert all maps with `time.Time` keys to instead use `int64` and then use the `Unix` or `UnixNano` functions to convert the times into seconds or nanoseconds from the Unix epoch respectively. Admittedly, this will take a non-trivial amount of refactoring, though we would perhaps be helped by defining a common `TimeMap` type which would do the conversions for us automatically.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 go:
-  - 1.7
-  - 1.8
+  - 1.7.6
+  - 1.8.3
   - 1.9
 install: make install-ci
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: go
 go:
   - 1.7
+  - 1.8
+  - 1.9
 install: make install-ci
 env:
  # Set higher timeouts for Travis


.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: go
 go:
   - 1.7
+  - 1.8
```
prateek: same as other PR, change to latest point releases

prateek:
We have #290 to track the map work. Per your comments, looks like it'll be a blocker for 1.9. 

Lets hold off adding the travis target for 1.9 until we fix the issue.

richardartoul:
@jeromefroe I think you can close this pr, I'll be tackling it in #385

jeromefroe:
@richardartoul sweet!

PR #347, opened 2017-09-11
Check checkpoint file writer error

prateek:
- Catches an error we were dropping in the `persist/fs.Writer.Close()`

Making a test for this is more involved, opened #348 as a follow up. 

/cc @robskillington

### DIFF
--- a/persist/fs/write.go
+++ b/persist/fs/write.go
@@ -262,8 +262,11 @@ func (w *writer) writeCheckpointFile() error {
 	if err != nil {
 		return err
 	}
-	defer fd.Close()
-	return w.digestBuf.WriteDigestToFile(fd, w.digestFdWithDigestContents.Digest().Sum32())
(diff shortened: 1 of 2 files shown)


persist/fs/write.go
```diff
@@ -262,8 +262,11 @@ func (w *writer) writeCheckpointFile() error {
 	if err != nil {
 		return err
 	}
-	defer fd.Close()
-	return w.digestBuf.WriteDigestToFile(fd, w.digestFdWithDigestContents.Digest().Sum32())
+	if err := w.digestBuf.WriteDigestToFile(fd, w.digestFdWithDigestContents.Digest().Sum32()); err != nil {
```
robskillington: nit: Extract the digest calculation to line before to reduce line length and improve overall readability?

prateek: sure thing

prateek: done

PR #349, opened 2017-09-12
Add nil namespace key propagation check

prateek:
Found an edge case while testing namespace kv propagation ~ m3dbnode was segfaulting if  it received a nil value from KV. This PR ensure that won't happen. 

/cc @robskillington


integration/dynamic_namespace_delete_test.go
```diff
@@ -0,0 +1,176 @@
...
+	// delete namespace key, ensure update propagates
+	numInvalid := numInvalidNamespaceUpdates(scope)
+	_, err = kvStore.Delete(dynamicOpts.NamespaceRegistryKey())
+	require.NoError(t, err)
+	log.Infof("deleted namespace key from kv")
+	for {
+		invalidUpdates := numInvalidNamespaceUpdates(scope)
+		if invalidUpdates > numInvalid {
+			break
+		}
```
robskillington: nit: Can rewrite this as `for numInvalidNamespaceUpdates(scope) <= numInvalid { time.Sleep(time.Second) }`

integration/dynamic_namespace_delete_test.go
```diff
@@ -0,0 +1,176 @@
...
+	for start, testData := range seriesMaps {
+		testSetup.setNowFn(start)
+		require.NoError(t, testSetup.writeBatch(ns0.ID(), testData))
+	}
+	log.Infof("test data is now written")
+
+	// Advance time and sleep for a long enough time so data blocks are sealed during ticking
+	testSetup.setNowFn(testSetup.getNowFn().Add(2 * blockSize))
+	later := testSetup.getNowFn()
+	time.Sleep(testOpts.TickInterval() * 10)
```
robskillington: Hm, at some point we should find some way to hint/trigger a manual tick instead of having to sleep for a tick cycle here.

storage/namespace/dynamic/registry.go
```diff
@@ -181,6 +182,12 @@ func (r *dynamicRegistry) run() {
 		}
 
 		val := r.kvWatch.Get()
+		if val == nil || reflect.ValueOf(val).IsNil() {
```
robskillington: Hm the way we've previously been protecting against an interface with nil ptr value is to make sure to return true `nil` instead of an interface with nil ptr for instance at the callsite.  e.g. see this type safe watch wrapping a `m3x/watch.Watch`:
https://github.com/m3db/m3db/blob/3d1c61958b68e0456c6d4936abb8176c9d710ba5/topology/map.go#L149-L155

Can't we do the same instead of using `reflect` here?

integration/dynamic_namespace_delete_test.go
prateek: sure, will just make it use `waitUntil` to follow the pattern below

integration/dynamic_namespace_delete_test.go
prateek: I've gone back and forth on that. Doing it this way sucks cause we're just polling & hoping, but it does mimic what the process would be doing in production. If we manually trigger a tick, it'd deviate from production behaviour so we might missing something.

storage/namespace/dynamic/registry.go
prateek: Yep, so this arises because I've got a stub implementation of `kv.ValueWatch` (same file at the bottom) which doesn't do this. I'll see if I can use the standard implementation or make the fix in my stub impl. 

Either way, won't need the reflect check.

integration/dynamic_namespace_delete_test.go
prateek: done

storage/namespace/dynamic/registry.go
prateek: Done. I replaced my stub implementation with `kv.ValueWatchable`, and removed the reflect call.

integration/dynamic_namespace_delete_test.go
prateek: Sure thing, I'll make the change to all integration tests where that's applicable.

robskillington: Thanks, big help!

prateek: sure thing, made the change in the latest patch

PR #353, opened 2017-09-14
Pull latest m3x/m3em/m3cluster

prateek:
/cc @jeromefroe @cw9 @robskillington

### DIFF
--- a/client/connection_pool.go
+++ b/client/connection_pool.go
@@ -32,7 +32,7 @@ import (
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	nchannel "github.com/m3db/m3db/network/server/tchannelthrift/node/channel"
 	"github.com/m3db/m3db/topology"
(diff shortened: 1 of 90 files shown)


jeromefroe:
@prateek looks like one of the integration tests failed, maybe it's flaky? Changes look good to me otherwise

client/connection_pool.go
```diff
@@ -32,7 +32,7 @@ import (
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	nchannel "github.com/m3db/m3db/network/server/tchannelthrift/node/channel"
 	"github.com/m3db/m3db/topology"
-	"github.com/m3db/m3x/close"
+	xclose "github.com/m3db/m3x/close"
```
cw9: Should we get rid of the name override?

prateek: I prefer keeping it to indicate we're using a `m3x` package

PR #362, opened 2017-10-10
add FetchMetadataBatchRaw rpc interface with a stubbed endpoint

jskelcy:
Motivation:
This diff is resurrecting an old diff @robskillington  put up to fetch batch metadata via ids rather than fetching for an entire shard. While the FetchBlocksMetadataRawRequest is useful for generating reports about metric utilization for an entire namespace this new endpoint will be helpful for utilization of adhoc queries.

Implementation:
Rob's initial diff can be found [here](https://github.com/m3db/m3db/compare/master...r/add-fetch-metadata-batch-rpc#diff-e3df571604b8901511d8dcd2a1f842f3). However I plan on taking a much more piecemeal approach. In this pr I am just adding the changes to the .thift file so we can agree on the network interface, as well as stubbing out functions to satisfy interfaces.


jskelcy:
it would appear that while I am able to generate go while compiles because I am not using the correct version of thrift gen the code churn in massive. Going to work on getting my thrift version to be correct.

storage/series/series_mock.go
```diff
@@ -26,14 +26,13 @@ package series
 import (
 	time "time"
 
+	gomock "github.com/golang/mock/gomock"
 	context "github.com/m3db/m3db/context"
 	persist "github.com/m3db/m3db/persist"
 	block "github.com/m3db/m3db/storage/block"
 	ts "github.com/m3db/m3db/ts"
 	io "github.com/m3db/m3db/x/io"
 	time0 "github.com/m3db/m3x/time"
```
jskelcy: I had to edit this generated mock by hand. For some reason, it was trying to import itself.

network/server/tchannelthrift/node/service.go
```diff
@@ -300,6 +302,13 @@ func (s *service) FetchBatchRaw(tctx thrift.Context, req *rpc.FetchBatchRawReque
 	return result, nil
 }
 
+// stubb function
+func (s *service) FetchMetadataBatchRaw(ctx thrift.Context, req *rpc.FetchMetadataBatchRawRequest) (*rpc.FetchMetadataBatchRawResult_, error) {
```
jskelcy: This will be implemented in future PRs.

network/server/tchannelthrift/node/service.go
```diff
@@ -57,28 +57,30 @@ var (
...
-	fetch               instrument.MethodMetrics
-	write               instrument.MethodMetrics
-	fetchBlocks         instrument.MethodMetrics
-	fetchBlocksMetadata instrument.MethodMetrics
-	repair              instrument.MethodMetrics
-	truncate            instrument.MethodMetrics
-	fetchBatchRaw       instrument.BatchMethodMetrics
-	writeBatchRaw       instrument.BatchMethodMetrics
-	overloadRejected    tally.Counter
+	fetch                 instrument.MethodMetrics
```
jskelcy: most gofmt changes in this file.

jskelcy:
Still investigating but it seems like the changes to most of the mock files are just changes to how the import statements are aliased and then using the new import alias. Not sure why this was changed, looking into versions.

jskelcy:
Closing this commit out for #364

PR #363, opened 2017-10-10
Add tool for testing bootstrapping from commitlog

richardartoul:
Replace: https://github.com/m3db/m3db/pull/313


tools/bootstrap_commitlogs/main/main.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+		// Or just handled up to N (shard-count) shards
+	} else if shardsCount > 0 {
+		for i := uint32(0); i < uint32(shardsCount); i++ {
+			rng := xtime.Range{Start: startInclusive, End: endExclusive}
+			shardTimeRanges[i] = xtime.NewRanges().AddRange(rng)
+			shardsAll = append(shardsAll, i)
+		}
+	} else {
+		flag.Usage()
```
prateek: might be better to tell the user that either shardCount or the comma list has to be set, and then print usage.

tools/bootstrap_commitlogs/main/main.go
```diff
@@ -0,0 +1,233 @@
...
+	fsOpts := fs.NewOptions().
+		SetInstrumentOptions(instrumentOpts).
+		SetFilePathPrefix(pathPrefix)
+
+	commitLogOpts := commitlog.NewOptions().
+		SetInstrumentOptions(instrumentOpts).
+		SetFilesystemOptions(fsOpts).
+		SetFlushSize(flushSize).
+		SetBlockSize(blockSize).
+		SetReadConcurrency(8)
```
prateek: maybe take a flag for this too

tools/bootstrap_commitlogs/main/main.go
```diff
@@ -0,0 +1,233 @@
...
+
+	opts := commitlogsrc.NewOptions().
+		SetResultOptions(resultOpts).
+		SetCommitLogOptions(commitLogOpts).
+		SetM3TSZEncodingConcurrency(10).
+		SetMergeShardsConcurrency(10)
```
prateek: and these two as well?

tools/bootstrap_commitlogs/main/main.go
```diff
@@ -0,0 +1,233 @@
...
+
+	retentionOpts := retention.NewOptions().
+		SetBlockSize(blockSize).
+		SetRetentionPeriod(bootstrapRetention).
+		SetBufferPast(1 * time.Minute).
+		SetBufferFuture(1 * time.Minute)
+
+	blockOpts := block.NewOptions()
+
+	encoderPool := encoding.NewEncoderPool(nil)
```
prateek: might be worthwhile to set the defaults to the pool options to the same thing we have for the nodes in production

richardartoul: Is that in a yaml file or...? How do i find that? Or is it just the default options and I just need to instantiate it?

tools/bootstrap_commitlogs/main/main.go
```diff
@@ -0,0 +1,233 @@
...
+	currentUnixTimestampParsed, err := strconv.Atoi(currentUnixTimestamp)
+	if err != nil {
+		log.Fatalf("could not parse unix timestamp: '%s': %v", currentUnixTimestamp, err)
+	}
+
+	now := time.Unix(int64(currentUnixTimestampParsed), 0)
+	// Round current time down to nearest blocksize (2h) and then decrease blocksize (2h)
+	startInclusive := now.Truncate(blockSize).Add(-blockSize)
+	// Round current time down to nearest blocksize (2h) and then add blocksize (2h)
+	endExclusive := now.Truncate(blockSize).Add(blockSize * 2)
```
prateek: would it be better to take a start/end time (or start time + duration)?

tools/bootstrap_commitlogs/main/main.go
```diff
@@ -0,0 +1,233 @@
...
+var (
+	pathPrefixArg           = flag.String("path-prefix", "/var/lib/m3db", "Path prefix - must contain a folder called 'commitlogs'")
+	namespaceArg            = flag.String("namespace", "metrics", "Namespace")
+	blockSizeArg            = flag.Duration("block-size", 10*time.Minute, "Block size")
+	flushSizeArg            = flag.Int("flush-size", 524288, "Flush size of commit log")
+	bootstrapRetentionArg   = flag.Duration("retention", 48*time.Hour, "Retention")
+	shardsCountArg          = flag.Int("shards-count", 8192, "Shards count - set number too bootstrap all shards in range")
+	shardsArg               = flag.String("shards", "", "Shards - set comma separated list of shards")
+	debugListenAddressArg   = flag.String("debug-listen-address", "", "Debug listen address - if set will expose pprof, i.e. ':8080'")
+	currentUnixTimestampArg = flag.String("current-unix-timestamp", currentUnixTimestampString(), "Current unix timestamp - If set will perform the bootstrap as if this was the current time, defaults to current time")
```
prateek: nit: mention which unit (seconds/millis/...) you're expecting

prateek: Any reason to not just make the Var a int64 instead of a string?

tools/bootstrap_commitlogs/main/main.go
richardartoul: I actually have to take these out for now because that code is not merged yet

tools/bootstrap_commitlogs/main/main.go
richardartoul: I actually have to take these out for now because that code is not merged yet

tools/bootstrap_commitlogs/main/main.go
richardartoul: Does it not need to be in increments of blocksize?

tools/bootstrap_commitlogs/main/main.go
richardartoul: good idea will fix

tools/bootstrap_commitlogs/main/main.go
richardartoul: Found em

tools/bootstrap_commitlogs/main/main.go
richardartoul: done

tools/bootstrap_commitlogs/main/main.go
richardartoul: done

tools/bootstrap_commitlogs/main/main.go
richardartoul: done

Makefile
```diff
@@ -43,6 +43,7 @@ TOOLS :=         \
 	read_index_ids \
 	clone_fileset  \
 	dtest          \
+	bootstrap_commitlogs \
```
robskillington: Perhaps we should call this `verify_commitlogs`

tools/bootstrap_commitlogs/main/main.go
```diff
@@ -0,0 +1,282 @@
...
+	"flag"
+	"io"
+	"net/http"
+	_ "net/http/pprof"
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/m3db/m3x/pool"
```
robskillington: nit: Remove space below.

tools/bootstrap_commitlogs/main/main.go
```diff
@@ -0,0 +1,282 @@
...
+var (
+	pathPrefixArg           = flag.String("path-prefix", "/var/lib/m3db", "Path prefix - must contain a folder called 'commitlogs'")
+	namespaceArg            = flag.String("namespace", "metrics", "Namespace")
+	blockSizeArg            = flag.Duration("block-size", 10*time.Minute, "Block size")
+	flushSizeArg            = flag.Int("flush-size", 524288, "Flush size of commit log")
+	bootstrapRetentionArg   = flag.Duration("retention", 48*time.Hour, "Retention")
+	shardsCountArg          = flag.Int("shards-count", 8192, "Shards count - set number too bootstrap all shards in range")
+	shardsArg               = flag.String("shards", "", "Shards - set comma separated list of shards")
+	debugListenAddressArg   = flag.String("debug-listen-address", "", "Debug listen address - if set will expose pprof, i.e. ':8080'")
+	currentUnixTimestampArg = flag.Int64("current-unix-timestamp", time.Now().Unix(), "Current unix timestamp (Seconds) - If set will perform the bootstrap as if this was the current time, defaults to current time")
```
robskillington: Instead of specifying current timestamp, it'd be more flexible to allow users to specify a start and an end time.

Also if they don't specify either it'd be nice if it simply tried to verify all commit logs by looking at the very first commit log and very last and work it ot.

richardartoul: @robskillington Do I need to look at the commitlog files on disk, or can I just set the start/end times to time.Unix(0, 0) and time.Unix(1<<63-62135596801, 999999999) respectively (https://stackoverflow.com/questions/25065055/what-is-the-maximum-time-time-in-go)

tools/verify_commitlogs/main/main.go
```diff
@@ -0,0 +1,279 @@
...
+	flushSizeArg          = flagParser.Int("flush-size", 524288, "Flush size of commit log")
+	bootstrapRetentionArg = flagParser.Duration("retention", 48*time.Hour, "Retention")
+	shardsCountArg        = flagParser.Int("shards-count", 8192, "Shards count - set number too bootstrap all shards in range")
+	shardsArg             = flagParser.String("shards", "", "Shards - set comma separated list of shards")
+	debugListenAddressArg = flagParser.String("debug-listen-address", "", "Debug listen address - if set will expose pprof, i.e. ':8080'")
+	startUnixTimestampArg = flagParser.Int64("start-unix-timestramp", 0, "Start unix timestamp (Seconds) - If set will bootstrap all data after this timestamp up to end-unix-timestamp, defaults to reading from the beginning of the first commitlog")
+	// 1<<63-62135596801 is the largest possible time.Time that can be represented
+	// without causing overflow when passed to functions in the time package
+	endUnixTimestampArg = flagParser.Int64("end-unix-timestramp", 1<<63-62135596801, "End unix timestamp (Seconds) - If set will bootstrap all data from start-unix-timestamp up to this timestamp, defaults to reading up to the end of the last commitlog")
```
robskillington: Haha nice, I like it.

PR #364, opened 2017-10-11
Add FetchMetadataBatchRaw RPC interface with a stubbed endpoint

jskelcy:
Motivation:
This diff is resurrecting an old diff @robskillington put up to fetch batch metadata via ids rather than fetching for an entire shard. While the FetchBlocksMetadataRawRequest is useful for generating reports about metric utilization for an entire namespace this new endpoint will be helpful for utilization of adhoc queries.

Implementation:
Rob's initial diff can be found here. However I plan on taking a much more piecemeal approach. In this pr I am just adding the changes to the .thift file so we can agree on the network interface, as well as stubbing out functions to satisfy interfaces.

### DIFF
--- a/generated/thrift/rpc.thrift
+++ b/generated/thrift/rpc.thrift
@@ -50,6 +50,7 @@ service Node {
 	FetchBatchRawResult fetchBatchRaw(1: FetchBatchRawRequest req) throws (1: Error err)
 	FetchBlocksRawResult fetchBlocksRaw(1: FetchBlocksRawRequest req) throws (1: Error err)
 	FetchBlocksMetadataRawResult fetchBlocksMetadataRaw(1: FetchBlocksMetadataRawRequest req) throws (1: Error err)
+	FetchMetadataBatchRawResult fetchMetadataBatchRaw(1: FetchMetadataBatchRawRequest req) throws (1: Error err)
 	void writeBatchRaw(1: WriteBatchRawRequest req) throws (1: WriteBatchRawErrors err)
 	void repair() throws (1: Error err)
 	TruncateResult truncate(1: TruncateRequest req) throws (1: Error err)
@@ -91,6 +92,22 @@ struct WriteRequest {
 	2: required string id
 	3: required Datapoint datapoint
 }
+
+struct FetchMetadataBatchRawRequest {
+	1: required binary nameSpace
+	2: required list<binary> ids
+}
+
+struct FetchMetadataBatchRawResult {
+	1: required list<FetchMetadataRawResult> elements
+}
+
+struct FetchMetadataRawResult {
+	1: required bool exists
+	2: optional i64 lastRead
+	3: optional TimeType lastReadTimeType = TimeType.UNIX_SECONDS
+}
+
 
 struct FetchBatchRawRequest {
 	1: required i64 rangeStart
@@ -144,6 +161,8 @@ struct Block {
 	2: optional Segments segments
 	3: optional Error err
 	4: optional i64 checksum
+	5: optional i64 lastRead
+	6: optional TimeType lastReadTimeType = TimeType.UNIX_SECONDS
 }
 
 struct FetchBlocksMetadataRawRequest {
--- a/network/server/tchannelthrift/node/service.go
+++ b/network/server/tchannelthrift/node/service.go
@@ -57,28 +57,30 @@ var (
 )
 
 type serviceMetrics struct {
-	fetch               instrument.MethodMetrics
-	write               instrument.MethodMetrics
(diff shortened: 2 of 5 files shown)


jskelcy:
Coverage decreased because I did not include a test for stub function.

generated/thrift/rpc.thrift
```diff
@@ -144,6 +161,8 @@ struct Block {
 	2: optional Segments segments
 	3: optional Error err
 	4: optional i64 checksum
+	5: optional i64 lastRead
```
jskelcy: This has been included so that when a peer bootstraps and blocks are passed between peers we can also pass along last read time.

generated/thrift/rpc.thrift
```diff
@@ -92,6 +93,22 @@ struct WriteRequest {
 	3: required Datapoint datapoint
 }
 
+struct FetchMetadataBatchRawRequest {
+	1: required binary nameSpace
```
prateek: nit: rename to `namespace`

generated/thrift/rpc.thrift
```diff
@@ -92,6 +93,22 @@ struct WriteRequest {
...
+	2: required list<binary> ids
```
prateek: could you rename to what type of id this is, i.e. `seriesIDs` or `shardIDs`

generated/thrift/rpc.thrift
```diff
@@ -92,6 +93,22 @@ struct WriteRequest {
...
+}
+
+struct FetchMetadataBatchRawResult {
+	1: required list<FetchMetadataRawResult> elements
+}
+
+struct FetchMetadataRawResult {
```
prateek: this struct doesn't have a field to indicate which series it's talking about, do you need that?

generated/thrift/rpc.thrift
```diff
@@ -92,6 +93,22 @@ struct WriteRequest {
 	3: required Datapoint datapoint
 }
 
+struct FetchMetadataBatchRawRequest {
+	1: required binary nameSpace
+	2: required list<binary> ids
+}
+
+struct FetchMetadataBatchRawResult {
+	1: required list<FetchMetadataRawResult> elements
```
prateek: `elements` is a bit weird, how about `results` or `metadatas`?

generated/thrift/rpc.thrift
prateek: Does this conflict have to be resolved in m3dbnode? Why not resolve it during the report processing.

generated/thrift/rpc.thrift
jskelcy: I added an `id` field here. But I took a look at the `FetchBatchRaw` endpoint which accepts a list of ids and returns a list of segments which are in the same order as the list of requested ids. I dont personally love this pattern but if we want to maintain this pattern for consistency that is fine.

generated/thrift/rpc.thrift
robskillington: It's not a conflict, peers that do not have a block at all (i.e. bootstrapping nodes) need this data or otherwise it is lost when the block is exchanged.

generated/thrift/rpc.thrift
robskillington: So the problem is that thrift does not allow the keyword `namespace` as its reserved... another reason to move to GRPC sometime in the future.. or rename namespace to be table for a shorter term fix, ha.

generated/thrift/rpc.thrift
robskillington: Yeah `id` is our standard name for series IDs when it comes to the RPC layer.

generated/thrift/rpc.thrift
robskillington: Yeah to save space (and allocs, run time, etc) we do not return ID's, simply the order in which the data was requested.  The raw and batch endpoints aren't meant to be as most efficient as can possibly be.

network/server/tchannelthrift/node/service.go
```diff
@@ -300,6 +302,13 @@ func (s *service) FetchBatchRaw(tctx thrift.Context, req *rpc.FetchBatchRawReque
 	return result, nil
 }
 
+// stubb function
```
robskillington: Perhaps update this to be more standard, i.e. `// TODO(jskelcy): implement the stubbed functions`

generated/thrift/rpc.thrift
jskelcy: I will remove this then.Thanks for the tip rob.

generated/thrift/rpc.thrift
prateek: haha damn, that's terrible

PR #365, opened 2017-10-12
Only read commitlog files that match shardTimeRanges

richardartoul:


persist/fs/commitlog/types.go
```diff
@@ -162,4 +163,11 @@ type Options interface {
 
 	// BytesPool returns the checked bytes pool
 	BytesPool() pool.CheckedBytesPool
+
+	// SetShardTimeRanges sets the ShardTimeRanges that determines which
+	// commitlogs to include in the bootstrap
+	SetShardTimeRanges(value result.ShardTimeRanges) Options
+
+	// ShardTimeRanges returns the ShardTimeRanges
+	ShardTimeRanges() result.ShardTimeRanges
```
prateek: If these are only used during Bootstrap, could you move them to`storage/bootstrap/bootstrapper/commitlog`

persist/fs/commitlog/iterator.go
```diff
@@ -177,3 +184,21 @@ func (i *iterator) nextReader() bool {
...
+// getFilesBasedOnShardTimeRanges determines which commit log files include
+// the time ranges that we're trying to bootstrap, and returns those
+func getFilesBasedOnShardTimeRanges(
+	fsopts fs.Options,
+	shardTimeRanges result.ShardTimeRanges,
+) ([]string, error) {
+	path := fs.CommitLogsDirPath(fsopts.FilePathPrefix())
+
+	if shardTimeRanges.IsEmpty() {
+		fmt.Println("IS EMPTY")
```
prateek: drop these printlns

persist/fs/commitlog/types.go
richardartoul: I tried doing that and then I realized that persis/fs/commit_log.go has an Iter() method which would mean that I would also need to pass the bootstrapping options into the CommitLog itself which seems weird

persist/fs/commitlog/iterator.go
```diff
@@ -177,3 +183,19 @@ func (i *iterator) nextReader() bool {
...
+
+// getFilesBasedOnShardTimeRanges determines which commit log files include
+// the time ranges that we're trying to bootstrap, and returns those
+func getFilesBasedOnShardTimeRanges(
+	fsopts fs.Options,
+	shardTimeRanges result.ShardTimeRanges,
+) ([]string, error) {
+	path := fs.CommitLogsDirPath(fsopts.FilePathPrefix())
+
+	if shardTimeRanges.IsEmpty() {
```
richardartoul: @prateek Is this preferable than returning an empty slice?

persist/fs/commitlog/iterator.go
```diff
@@ -177,3 +183,19 @@ func (i *iterator) nextReader() bool {
...
+		return fs.CommitLogFiles(path)
+	}
+
+	start, end := shardTimeRanges.MinMax()
```
richardartoul: @prateek I'm actually starting to question if this logic is correct because it doesn't account for the write back / write forward buffer. Am I correct that I need to include the commit logs for the Min/Max timestamps, and then add one more commitlog on both ends to account for the forward/backwards writes?

prateek: lol yeah it gets a little wonky. theres a good explanation here ~ https://github.com/m3db/m3db/blob/master/storage/cleanup.go#L184-L198

persist/fs/commitlog/types.go
prateek: Shouldn't introduce a dependency on the `result` package here. Here's an alternative: you can get this information using a new ctor for `Iterator` which takes a list of times (min/max if that's easier), or a predicate which returns a bool indicating if a commit log block start time should be read. 

What do you think?

richardartoul: New ctor for Iterator seems reasonable, although it leaves open the question of how to handle the Iter method in commit_log.go:

```
func (l *commitLog) Iter() (Iterator, error) {
	return NewIterator(l.opts)
}
```

Do I just make that method accept the min/max as well? I actually wouldn't be opposed t removing that method since it barely does anything

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -256,3 +260,31 @@ func (s *commitLogSource) Read(
...
+	return func(commitLogTime time.Time) bool {
+		// Minimum and maximum times for which we want to bootstrap
+		shardMin, shardMax := shardsTimeRanges.MinMax()
+
+		// How far into the past or future a commitlog might contain a write for a
+		// previous or future block
+		bufferPast := ns.Options().RetentionOptions().BufferPast()
+		bufferFuture := ns.Options().RetentionOptions().BufferFuture()
+
+		blockSize := opts.CommitLogOptions().BlockSize()
```
richardartoul: @prateek This is correct right? I care about the blockSize of the commitlog, not of the retention?

integration/commitlog_bootstrap_test.go
```diff
@@ -82,7 +82,24 @@ func TestCommitLogBootstrap(t *testing.T) {
...
+	log.Info("finished writing data")
+
+	// The datapoints in this generated data are within the retention period and
+	// would oridinarly be bootstrapped, however, we intentionally write them to a
+	// commitlog file that has a timestamp outside of the retention period. This
+	// allows us to verify the commitlog bootstrapping logic will not waste time
+	// reading commitlog files that are outside of the retention period.
+	log.Info("generating data")
+	seriesMapsExpiredCommitlog := generateSeriesMaps(30, now.Add(-2*blockSize), now.Add(-blockSize))
+	log.Info("writing data to commitlog file with out of range timestamp")
```
robskillington: Might be better to add this as a separate independent test so we can tell what part of the commit log is failing based on test name, rather than add this to a even larger compound integration test.

persist/fs/commitlog/iterator.go
```diff
@@ -62,15 +64,29 @@ type iteratorRead struct {
 	annotation []byte
 }
 
+// ReadCommitLogPredicate is a predicate that allows the caller to determine
+// which commitlogs the iterator should read from
+type ReadCommitLogPredicate func(commitLogTime time.Time) bool
+
+// ReadAllCommitLogs can be passed as the ReadCommitLogPredicate for callers
+// that want a convenient way to read all the commitlogs
+func ReadAllCommitLogs(commitLogTime time.Time) bool { return true }
```
robskillington: Perhaps we can add a more readable option:
```go
func ReadAllPredicate() ReadEntryPredicate {
  return func(entryTime time.Time) bool { return true }
}
```

That way the calling code looks more readable and its more clear looking at the godocs how to get a type that matches `ReadCommitLogPredicate` easily, for instance `ReadAllCommitLogs` just looking at godoc is not clear it can be used as a predicate.  Also trying to avoid including the package name in exported types is a Good Thing(tm). e.g.:
```go
iter, err := commitlog.NewIterator(opts, commitlog.ReadAllPredicate())
```

Rather than current:
```go
iter, err := commitlog.NewIterator(opts, commitlog.ReadAllCommitLogs)
```

persist/fs/commitlog/iterator.go
```diff
@@ -185,6 +201,22 @@ func (i *iterator) nextReader() bool {
 	return true
 }
 
+func filterFiles(files []string, pred ReadCommitLogPredicate) ([]string, error) {
+	filteredFiles := []string{}
+	var multiErr xerrors.MultiError
+	for _, file := range files {
+		fileTime, err := fs.TimeFromFileName(file)
```
robskillington: It might be worth actually reading the info struct instead of relying on file name?  File name is usually only for ordering, not for relying on data in the actual file itself.

persist/fs/files.go
```diff
@@ -245,17 +245,11 @@ func CommitLogFilesForTime(commitLogsDir string, t time.Time) ([]string, error)
...
 	if err != nil {
-		multiErr = multiErr.Add(err)
-	}
-	res, err := filesBefore(commitLogs, t)
-	if err != nil {
-		multiErr = multiErr.Add(err)
+		return nil, err
 	}
-	return res, multiErr.FinalError()
+	return filesBefore(commitLogs, t)
```
robskillington: Hm, it might also be worth fixing `filesBefore` to actually read the info header too.  Seems it has the same issue.

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -256,3 +260,33 @@ func (s *commitLogSource) Read(
...
+		shardRange := xtime.Range{
+			Start: shardMin,
+			End:   shardMax,
+		}
+
+		// How far into the past or future a commitlog might contain a write for a
+		// previous or future block
+		bufferPast := ns.Options().RetentionOptions().BufferPast()
+		bufferFuture := ns.Options().RetentionOptions().BufferFuture()
+		blockSize := opts.CommitLogOptions().BlockSize()
```
robskillington: Maybe move all these vars before the lambda so we only do these once?

persist/fs/commitlog/iterator.go
richardartoul: Hey Rob, I did this refactoring. Couple of notes:

1) I wasn't able to share too much of the logic between the reader itself and the new ReadLogInfo function. Thats primarily because the Reader methods are highly entangled with its state (pooling, shared buffers, etc etc). I tried doing it and it just made the code unnecessarily hard to grok. I think its ok though because the new ReadLogInfo function is fairly short, straightforward, and doesn't have too much business logic in int. Its just setting up a chunk reader, opening/closing a file, instantiating a msgpack decoder etc etc.

2) Previously we were relying on the Options struct to tell us how long of a period of time the commitlog represented (via the Blocksize). I think this would have caused trouble for people who tried to change this value and restart a node, so now that we get that info from the commitlog itself I think the logic is more sound.

3) I was unable to move the call to ReadLogInfo into the fs.go file because there is already a dependent in the commitlog reader on the fs package so it would have become cyclic. With that said, it seems reasonable for the FS package to have functions for determining information about the commitlog from the filename, and for the commitlog package to have functions for determining information about the commitlog by parsing it.

4) Presumably we wanted to make this change because we trust the loginfo struct more than we trust the filename? Should I look at both? We could make the predicate pass both values so that if there is a discrepancy you could choose to read the file anyways, but that might be overkill.

integration/commitlog_bootstrap_test.go
richardartoul: Yeah fair enough, I just did it because these tests can take awhile to run and this seemed like a fairly small feature to test in an integration test, but I can break it out

persist/fs/commitlog/chunk_reader.go
```diff
@@ -0,0 +1,137 @@
...
+	if err != nil {
+		return err
+	}
+
+	sizeStart, sizeEnd :=
+		0, chunkHeaderSizeLen
+	checksumSizeStart, checksumSizeEnd :=
+		sizeEnd, sizeEnd+chunkHeaderSizeLen
+	checksumDataStart, checksumDataEnd :=
+		checksumSizeEnd, checksumSizeEnd+chunkHeaderChecksumDataLen
```
robskillington: Perhaps move these constants to the top of this file? i.e. `chunkHeaderSizeLen` and `chunkHeaderChecksumDataLen`

persist/fs/commitlog/chunk_reader.go
```diff
@@ -0,0 +1,137 @@
...
+	size := endianness.Uint32(header[sizeStart:sizeEnd])
+	checksumSize := digest.
+		Buffer(header[checksumSizeStart:checksumSizeEnd]).
+		ReadDigest()
+	checksumData := digest.
+		Buffer(header[checksumDataStart:checksumDataEnd]).
+		ReadDigest()
+
+	// Verify size checksum
+	if digest.Checksum(header[:4]) != checksumSize {
```
robskillington: Maybe make this `header[: chunkHeaderSizeLen]` for consistency?

persist/fs/commitlog/iterator.go
```diff
@@ -185,6 +203,22 @@ func (i *iterator) nextReader() bool {
 	return true
 }
 
+func filterFiles(opts Options, files []string, predicate ReadEntryPredicate) ([]string, error) {
+	filteredFiles := []string{}
```
robskillington: Perhaps allocate this up front for single alloc?  Even though this is a low frequency called method, is still reasonable practice when you can very easily calculate the upper bound.

i.e.
```go
filteredFiles := make([]string, 0, len(files))
```

persist/fs/files.go
robskillington: This is resolved now 👍

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -256,3 +260,33 @@ func (s *commitLogSource) Read(
...
+		shardMin, shardMax := shardsTimeRanges.MinMax()
+		shardRange := xtime.Range{
+			Start: shardMin,
+			End:   shardMax,
+		}
+
+		// How far into the past or future a commitlog might contain a write for a
+		// previous or future block
+		bufferPast := ns.Options().RetentionOptions().BufferPast()
+		bufferFuture := ns.Options().RetentionOptions().BufferFuture()
```
robskillington: Hey want to move all these vars to above so that they get calculated/called once?

e.g.
```go
		// Minimum and maximum times for which we want to bootstrap
		shardMin, shardMax := shardsTimeRanges.MinMax()
		shardRange := xtime.Range{
			Start: shardMin,
			End:   shardMax,
		}
 
		// How far into the past or future a commitlog might contain a write for a
		// previous or future block
		bufferPast := ns.Options().RetentionOptions().BufferPast()
		bufferFuture := ns.Options().RetentionOptions().BufferFuture()
		return func(entryTime time.Time, entryDuration time.Duration) bool {
			return xtime.Range{
				Start: entryTime.Add(-bufferPast),
				End:   entryTime.Add(entryDuration).Add(bufferFuture),
			}.Overlaps(shardRange)
		}
	}
```

persist/fs/files.go
richardartoul: is this resolved in some other PR? I ended up not touching this package much at all

robskillington:
Hm seems `TestDeleteInactiveFiles` is failing, think this is a new one added by @katezaps - probably need to find out what's going on here.

persist/fs/files.go
robskillington: Yeah I think we were thinking we should update this fn to use your new `ReadLogInfo(...)` method, do you mind doing that? 

We should also rename it to `commitLogFilesBefore(...)` or just inline it here, I checked and this is the only function that calls `filesBefore(...)`.

richardartoul:
@robskillington Re: "Hm seems TestDeleteInactiveFiles is failing, think this is a new one added by @katezaps - probably need to find out what's going on here."

That test is actually just broken on master. It managed to get land when our C.I was broken and was reporting successes even when the tests failed (now fixed). #389 is tracking it

persist/fs/files.go
robskillington: Ok sure thing, maybe we can open an issue and come back to it?

richardartoul: #391

PR #366, opened 2017-10-13
Fix race condition in Terminate which fixes flaky integration test MixedModeReadWrite

richardartoul:
The test was flaky because after the Database had shutdown, background goroutines would continue running (sometime between the call to Terminate() completing and the actual process exiting) and when they called GetOwnedNamespaces() they would receive an empty list.

The empty list is bad, because when the list of namespaces is empty, the logic in [this function](https://github.com/m3db/m3db/blob/master/storage/cleanup.go#L143) allows M3DB to delete all commitlog files which triggers data loss / test failure.

### DIFF
--- a/storage/cleanup.go
+++ b/storage/cleanup.go
@@ -89,7 +89,9 @@ func (m *cleanupManager) Cleanup(t time.Time) error {
 			"encountered errors when cleaning up fileset files for %v: %v", t, err))
 	}
 
-	commitLogStart, commitLogTimes := m.commitLogTimes(t)
+	commitLogStart, commitLogTimes, err := m.commitLogTimes(t)
+	multiErr = multiErr.Add(fmt.Errorf(
+		"encountered errors when cleaning up commit logs: %v", err))
 	if err := m.cleanupCommitLogs(commitLogStart, commitLogTimes); err != nil {
 		multiErr = multiErr.Add(fmt.Errorf(
 			"encountered errors when cleaning up commit logs for commitLogStart %v commitLogTimes %v: %v",
@@ -112,10 +114,11 @@ func (m *cleanupManager) Report() {
 }
 
 func (m *cleanupManager) cleanupFilesetFiles(t time.Time) error {
-	var (
-		multiErr   = xerrors.NewMultiError()
-		namespaces = m.database.GetOwnedNamespaces()
-	)
+	multiErr := xerrors.NewMultiError()
+	namespaces, err := m.database.GetOwnedNamespaces()
+	if err != nil {
+		return err
+	}
 	for _, n := range namespaces {
 		earliestToRetain := retention.FlushTimeStart(n.Options().RetentionOptions(), t)
 		multiErr = multiErr.Add(n.CleanupFileset(earliestToRetain))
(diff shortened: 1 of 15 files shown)


storage/cleanup.go
```diff
@@ -89,7 +89,9 @@ func (m *cleanupManager) Cleanup(t time.Time) error {
 			"encountered errors when cleaning up fileset files for %v: %v", t, err))
 	}
 
-	commitLogStart, commitLogTimes := m.commitLogTimes(t)
+	commitLogStart, commitLogTimes, err := m.commitLogTimes(t)
+	multiErr = multiErr.Add(fmt.Errorf(
```
prateek: Need to check `err != nil` here

storage/database.go
```diff
@@ -551,10 +554,13 @@ func (d *db) ownedNamespacesWithLock() []databaseNamespace {
 	return namespaces
 }
 
-func (d *db) GetOwnedNamespaces() []databaseNamespace {
+func (d *db) GetOwnedNamespaces() ([]databaseNamespace, error) {
```
prateek: could you add a test to verify this functions as you're expecting

client/client_mock.go
```diff
@@ -38,10 +37,9 @@ import (
 	instrument "github.com/m3db/m3x/instrument"
 	pool "github.com/m3db/m3x/pool"
 	retry "github.com/m3db/m3x/retry"
-	time "github.com/m3db/m3x/time"
-
-	gomock "github.com/golang/mock/gomock"
+	time0 "github.com/m3db/m3x/time"
```
prateek: could you drop any mock files that only have import reorderings in this diff

storage/database.go
richardartoul: yep done

storage/cleanup.go
```diff
@@ -89,7 +89,12 @@ func (m *cleanupManager) Cleanup(t time.Time) error {
 			"encountered errors when cleaning up fileset files for %v: %v", t, err))
 	}
 
-	commitLogStart, commitLogTimes := m.commitLogTimes(t)
+	commitLogStart, commitLogTimes, err := m.commitLogTimes(t)
+	if err != nil {
+		multiErr = multiErr.Add(fmt.Errorf(
+			"encountered errors when cleaning up commit logs: %v", err))
```
prateek: i'd suggest returning `multiErr.FinalError()` after this line

richardartoul: yeah that makes sense

storage/database.go
```diff
@@ -165,12 +168,14 @@ func NewDatabase(
 		return nil, err
 	}
 
-	// wait till first value is received
+	// Wait till first namespaces value is received and set the value.
+	// Its important that this happens before the mediator is started to prevent
```
prateek: Might make sense to move the mediator construction below the namespace calls, what do you think?

richardartoul: I think practically it has no effect (because tick routines don't start until Open() is called), but conceptually it makes sense yeah.

Also in the terminate method we briefly discussed re-ordering things so the mediator is closed and then we close the namespace watcher instead of the opposite. Again, I don't think it has any effect but might make more conceptual sense. What do you think?

prateek: Yeah makes sense. Could you do the reordering in Close/Terminate too.

richardartoul: cool done

storage/cleanup.go
```diff
@@ -89,7 +89,13 @@ func (m *cleanupManager) Cleanup(t time.Time) error {
 			"encountered errors when cleaning up fileset files for %v: %v", t, err))
 	}
 
-	commitLogStart, commitLogTimes := m.commitLogTimes(t)
+	commitLogStart, commitLogTimes, err := m.commitLogTimes(t)
```
prateek: Could you verify we have unit test coverage to verify that `Cleanup()` behaves as expected in both cases ~ when GetOwnedNamespaces returns error, and when it doesn't.

richardartoul: Looks like happy path was already tested so I added a unit test for the case where commitLogTimes returns an error

PR #367, opened 2017-10-16
[WIP] Extend cleanup manager to delete shard filesets from shards no longer owned

katezaps:


katezaps:
rebase got too complicated so i copied logic to https://github.com/m3db/m3db/pull/381

PR #368, opened 2017-10-17
Add YAML config driven source file for building and running m3dbnode

robskillington:
Still pending:
- [x] Move `ClientConfiguration` into `client` package
- [x] Move KV keys into their own package
- [x] Cleanup some structs in the main file that should live in `services/m3dbnode/config`
- [x] Move `services/m3dbnode/server` package near the integration package as its the sole user now


services/m3dbnode/main/main.go
```diff
@@ -23,114 +23,383 @@ package main
...
+	serverGracefulCloseTimeout = 10 * time.Second
 )
 
 var (
-	idArg                  = flag.String("id", "", "Node host ID")
-	httpClusterAddrArg     = flag.String("clusterhttpaddr", "0.0.0.0:9000", "Cluster HTTP server address")
-	tchannelClusterAddrArg = flag.String("clustertchanneladdr", "0.0.0.0:9001", "Cluster TChannel server address")
-	httpNodeAddrArg        = flag.String("nodehttpaddr", "0.0.0.0:9002", "Node HTTP server address")
-	tchannelNodeAddrArg    = flag.String("nodetchanneladdr", "0.0.0.0:9003", "Node TChannel server address")
+	configFile = flag.String("f", "", "configuration file")
```
prateek: could you include a sample config and write a test which verifies the service node is able to startup with the it

robskillington: I was thinking more about a test that verifies a YAML transforms to the given struct expected in the config package.

For integration test of an actual node coming up and being accessible we'll need to wait for an embedded etcd server solution, which will come in a subsequent change. i.e. ability to startup the node running an embedded etcd server.

client/types.go
```diff
@@ -65,17 +68,48 @@ const (
 func (l ConnectConsistencyLevel) String() string {
 	switch l {
 	case ConnectConsistencyLevelAny:
-		return "ConnectConsistencyLevelAny"
+		return "any"
```
richardartoul: Out of curiosity, is there any reason the consistency levels are iota's instead of strings? If they were strings you wouldn't have to do all this manual switching. Is it just a perf thing?

robskillington: Yeah it's a perf thing, don't want to have to encode strings over the wire when making RPCs.  We may not even push these over the wire just yet, but it was a "best practices" thing considering we weren't sure if we were going to use the nodes as coordinators or not at that point.

services/m3dbnode/config/hostid.go
```diff
@@ -0,0 +1,76 @@
...
+import (
+	"fmt"
+	"os"
+)
+
+type hostIDResolver string
+
+const (
+	hostnameHostIDResolver    hostIDResolver = "hostname"
+	configHostIDResolver      hostIDResolver = "config"
```
prateek: nit: i think you only need the type decl on the first var in the const block

services/m3dbnode/config/hostid.go
```diff
@@ -0,0 +1,76 @@
...
+	environmentHostIDResolver hostIDResolver = "environment"
+)
+
+// HostIDConfiguration is the configuration for resolving the local host ID.
+type HostIDConfiguration struct {
```
prateek: should we move this to m3x?

services/m3dbnode/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+	Repair repairPolicy `yaml:"repair"`
+
+	// The pooling policy
+	PoolingPolicy PoolingPolicy `yaml:"poolingPolicy"`
+
+	// The configuration for config service client
+	ConfigService etcdclient.Configuration `yaml:"configService"`
+}
+
+type commitLogPolicy struct {
```
prateek: Could you make this (& other config structs) exported? Will be useful for when we're overriding properties programmatically, e.g. dtest runs

services/m3dbnode/config/fs.go
```diff
@@ -0,0 +1,109 @@
...
+		return DefaultNewFileMode, nil
+	}
+
+	str := *p.NewFileMode
+	if len(str) != 3 {
+		return 0, fmt.Errorf("file mode must be 3 chars long")
+	}
+
+	// NB(r): There's no way next line can be ineffectual as it prepends str..
+	str = "0" + str // nolint: ineffassign
```
prateek: isn't this ineffectual? 

str is a value type - you're dereferencing `*p.NewFileMode` above - so this would change the value of str, which is never used below this line, i.e. ineffectual yea?

services/m3dbnode/config/fs.go
```diff
@@ -0,0 +1,109 @@
...
+	ReadBufferSize int `yaml:"readBufferSize" validate:"min=0"`
+
+	// Disk flush throughput limit in Mb/s
+	ThroughputLimitMbps float64 `yaml:"throughputLimitMbps" validate:"min=0.0"`
+
+	// Disk flush throughput check interval
+	ThroughputCheckEvery int `yaml:"throughputCheckEvery" validate:"nonzero"`
+
+	// NewFileMode is the new file permissions mode to use when
+	// creating files - be sure to specifying leading zero, e.g. 0666.
```
prateek: the comment says to include "0" in the value, but you're checking length = 3 and prepending "0" whilst parsing. Could you change one or other

services/m3dbnode/config/fs.go
```diff
@@ -0,0 +1,109 @@
...
+		return DefaultNewDirectoryMode, nil
+	}
+
+	str := *p.NewDirectoryMode
+	if len(str) != 3 {
+		return 0, fmt.Errorf("file mode must be 3 chars long")
+	}
+
+	// NB(r): There's no way next line can be ineffectual as it prepends str..
+	str = "0" + str // nolint: ineffassign
```
prateek: same point as above

services/m3dbnode/config/pooling.go
```diff
@@ -0,0 +1,128 @@
...
+
+package config
+
+// PoolingPolicy specifies the pooling policy.
+type PoolingPolicy struct {
+	// The initial alloc size for a block
+	BlockAllocSize int `yaml:"blockAllocSize"`
+
+	// The general pool type: simple or native.
+	Type string `yaml:"type" validate:"regexp=(^simple$|^native$)"`
```
prateek: could you define a type for this, the same as you're doing for `CalculationType`

x/tchannel/logger.go
```diff
@@ -0,0 +1,45 @@
...
+
+var noLog noopLogger
+
+type noopLogger struct{}
+
+// NewNoopLogger returns a default tchannel no-op logger.
+func NewNoopLogger() tchannel.Logger { return noLog }
+
+func (noopLogger) Enabled(_ tchannel.LogLevel) bool                         { return false }
+func (noopLogger) Fatal(msg string)                                         { os.Exit(1) }
```
prateek: might be good to log the message here at least

services/m3dbnode/main/main.go
```diff
@@ -23,114 +23,383 @@ package main
...
+	if err != nil {
+		log.Fatalf("could not parse new file mode: %v", err)
+	}
+
+	newDirectoryMode, err := cfg.Filesystem.ParseNewFileMode()
+	if err != nil {
+		log.Fatalf("could not parse new directory mode: %v", err)
+	}
+
+	fsopts := fs.NewOptions().
```
prateek: The main would get simplified a lot if we could just get a `storage.Options` from the top level config struct. Could you make `Options()` a method on the config structs and do this stuff there?

services/m3dbnode/main/main.go
```diff
@@ -23,114 +23,383 @@ package main
...
+
+	opts = opts.SetRuntimeOptionsManager(runtimeOptsMgr)
+
+	opts = withEncodingAndPoolingOptions(logger, opts, cfg.PoolingPolicy)
+
+	// Set the block retriever manager
+	retrieverOpts := fs.NewBlockRetrieverOptions().
+		SetBytesPool(opts.BytesPool()).
+		SetSegmentReaderPool(opts.SegmentReaderPool()).
+		SetFetchConcurrency(2)
```
prateek: nit: make 2 a const

services/m3dbnode/main/main.go
```diff
@@ -139,3 +408,308 @@ func interrupt() error {
...
+		}
+	}
+	if scope != nil {
+		opts = opts.SetInstrumentOptions(opts.InstrumentOptions().
+			SetMetricsScope(scope))
+	}
+	return opts
+}
+
+func capacityPoolOptions(
```
prateek: could you move this to a method on the `CapacityPoolPolicy` struct

services/m3dbnode/config/hostid.go
robskillington: Only for enums, tried this and they reverted to string type

services/m3dbnode/config/fs.go
robskillington: Haha, ok `str` is meant to be the arg passed to Sscanf, will fix this up.

services/m3dbnode/config/fs.go
robskillington: Yeah I'lll remove this comment.

services/m3dbnode/config/hostid.go
robskillington: Sure thing.

services/m3dbnode/config/pooling.go
robskillington: Sure thing.

services/m3dbnode/main/main.go
robskillington: As discussed, this is a big change that we should probably defer considering there are other bigger issues at hand.  We can always revisit this when there is some more time.

services/m3dbnode/config/hostid.go
```diff
@@ -0,0 +1,76 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package config
```
prateek: I think you still need to delete and pick this up from m3x

x/tchannel/logger.go
prateek: *bump*

services/m3dbnode/config/bootstrap.go
```diff
@@ -41,7 +41,7 @@ var (
 	defaultNumProcessorsPerCPU = 0.5
 )
 
-type filesystemConfiguration struct {
+type bootstrapFilesystemConfiguration struct {
```
prateek: could you make this an exported type too

services/m3dbnode/server/server_test.go
```diff
@@ -0,0 +1,476 @@
...
+		assert.Equal(t, expectAt, dp.Timestamp)
+		assert.Equal(t, v.unit, unit)
+	}
+
+	// Wait for server to stop
+	interruptCh <- fmt.Errorf("test complete")
+	serverWg.Wait()
+}
+
+var testConfig = `
```
prateek: this is dope af! 

how would you feel about making this an exported type in a non `_test.go` file in the same package?

services/m3dbnode/server/server.go
```diff
@@ -0,0 +1,748 @@
...
+		SetInstrumentOptions(iopts).
+		SetConfigServiceClient(configSvcClient).
+		SetNamespaceRegistryKey(kvconfig.NamespacesKey)
+	nsInit := namespace.NewDynamicInitializer(dynamicOpts)
+
+	opts = opts.
+		SetTickInterval(cfg.TickInterval).
+		SetNamespaceInitializer(nsInit)
+
+	newFileMode, err := cfg.Filesystem.ParseNewFileMode()
```
prateek: super nit: could you move the cheaper validations above the more expensive ones. e.g. `newFileMode`/`hostid`/etc before connecting to kv. In the event of a bad config, it'd be good to fail earlier.

robskillington: Sure thing, done.

services/m3dbnode/server/server_test.go
robskillington: i.e. some kind of method that takes options and spits out a YAML?
```go
type ConfigYAMLOptions struct {
  ServicePort int
  //...
}
func ConfigYAML(opts ConfigYAMLOptions) {
  //...
}
```

The only issue with this is that if you truly want to be able to print any type of YAML (i.e. make it flexible) then you need `ConfigYAMLOptions` to be as expressive as `config.Configuration` which parses the YAML to be.  As such it would be better to actually create a `config.Configuration` then marshal it to YAML for such a purpose.

What's the intended use case?  Then we can understand the best ergonomics to meet that use case.

prateek: Yep so pretty much the use-case you're thinking: use this template, and write a tool in the m3db repo which populates it with the values to spit out a YAML. I hear your point about it needing to be as expressive as the struct itself in the general case. IMO the base version you have here would go a long way to get people started. Consider the template we use internally, most of it is static too.

Don't have to write the tool now, I figure just exporting the struct is a good enough starting point.

PR #372, opened 2017-10-21
Fix race condition in Shard insert v bootstrap

prateek:
- Fixes race condition in async shard insert during bootstrap

/cc @robskillington


robskillington:
Hey, I think the more preferable approach which is race safe even during async inserts would  be this:
https://github.com/m3db/m3db/compare/r/simple-new-series-bootstrapped-logic

PR #375, opened 2017-10-23
PR 1 of many to replace map[time.Time] with map[int64]

richardartoul:
Partially addresses #290


x/time/time.go
```diff
@@ -0,0 +1,17 @@
+package time
+
+import "time"
+
+// UnixNano is used to indicate that an int64 stores a unix timestamp at
+// nanosecond resolution
+type UnixNano int64
+
+// Time returns a time.Time from a UnixNano
+func (u UnixNano) Time() time.Time {
```
xichen2020: nit: `ToTime` might be a better name?

x/time/time.go
```diff
@@ -0,0 +1,17 @@
...
+	return time.Unix(0, int64(u))
+}
+
+// NewUnixNano returns a UnixNano from a time.Time
+func NewUnixNano(t time.Time) UnixNano {
```
xichen2020: nit: `fromTime` or `toUnixNanos` might be a better name?

storage/block/block.go
```diff
@@ -335,32 +336,34 @@ func (dbb *databaseSeriesBlocks) MaxTime() time.Time {
...
 	dbb.min, dbb.max = timeZero, timeZero
 	if len(dbb.elems) == 0 {
 		return
 	}
-	for k := range dbb.elems {
-		if dbb.min == timeZero || dbb.min.After(k) {
-			dbb.min = k
+	for key := range dbb.elems {
+		keyTime := key.Time()
+		if dbb.min == timeZero || dbb.min.After(keyTime) {
```
xichen2020: Might make sense to convert `dbb.min` and `dbb.max` to use `UnixNano` as well?

x/time/time.go
richardartoul: Yeah thats actually what I called it originally, I can change it

x/time/time.go
richardartoul: I like ToUnixNano, I'll make it that

storage/block/block.go
richardartoul: I can change this, but I think using the existing time.Time interface is nicer just because people are familiar with it.

prateek:
@richardartoul I think you need to pull in the latest submodule change

storage/block/block.go
richardartoul: Basically my rationale is that in this case the comparison are fairly simple, but in other cases as I make these changes there will be things that will be using more complicated functions like Truncate() and Round() and it will just keep everything cleaner if I only do the conversions when putting things in and pulling things out of the map

richardartoul:
@prateek You're right, pulled it in.

storage/block/block.go
```diff
@@ -30,6 +30,7 @@ import (
 	"github.com/m3db/m3db/digest"
 	"github.com/m3db/m3db/ts"
 	xio "github.com/m3db/m3db/x/io"
+	m3dbTime "github.com/m3db/m3db/x/time"
```
prateek: nit: could you change the casing here, `m3dbtime` instead of `m3dbTime`

richardartoul: Oh yeah I forgot its always supposed to be lowercase

x/time/time.go
```diff
@@ -0,0 +1,17 @@
+package time
+
+import "time"
+
+// UnixNano is used to indicate that an int64 stores a unix timestamp at
+// nanosecond resolution
+type UnixNano int64
+
+// ToTime returns a time.ToTime from a UnixNano
+func (u UnixNano) ToTime() time.Time {
```
robskillington: Might be worth putting this into `m3x/time` and using package name of `xtime` so its more similar to our other shared types.  At some point `m3aggregator` will probably want to use the same approach and they may as well reference the same type.

storage/block/block.go
```diff
@@ -335,32 +336,34 @@ func (dbb *databaseSeriesBlocks) MaxTime() time.Time {
...
 	dbb.min, dbb.max = timeZero, timeZero
 	if len(dbb.elems) == 0 {
 		return
 	}
-	for k := range dbb.elems {
-		if dbb.min == timeZero || dbb.min.After(k) {
-			dbb.min = k
+	for key := range dbb.elems {
+		keyTime := key.ToTime()
+		if dbb.min == timeZero || dbb.min.After(keyTime) {
```
prateek: How do you feel about exposing the time.Time methods in use  (`After`/`Before`/etc) on the `UnixNano` type? That way we could avoid the extra time.Time alloc

storage/block/block.go
```diff
@@ -287,15 +288,15 @@ func (b *dbBlock) resetMergeTargetWithLock() {
 }
 
 type databaseSeriesBlocks struct {
-	elems map[time.Time]DatabaseBlock
+	elems map[xtime.UnixNano]DatabaseBlock
 	min   time.Time
```
prateek: could make min/max `UnixNano` too

richardartoul: I tried changing them and they're actually exposed in public interfaces. Might be best to just keep them as time.Time for now. I did a bunch of benchmarking and even with the conversions back and forth, this code should be faster https://gist.github.com/richardartoul/2ce3c5258df0fb6f47a030950eafe93e

storage/block/block.go
prateek: talked offline, going to keep the scope to only map[time.Time] for now.

PR #376, opened 2017-10-24
PR 2 of many to replace map[time.Time] with map[int64] - File Seek Manager

richardartoul:
Partially addresses #290


x/time/time.go
```diff
@@ -0,0 +1,17 @@
+package time
```
richardartoul: This file will be merged in PR1, its just here until that one is merged

prateek: think you need to rebase this and the other prs

richardartoul: yep was just waiting for PR1 to get approved so I could do that, thanks for stamping it! Will rebase everything.

PR #377, opened 2017-10-24
PR 3 of many to replace map[time.Time] with map[int64] - Repair

richardartoul:
Partially addresses #290


x/time/time.go
```diff
@@ -0,0 +1,17 @@
+package time
```
richardartoul: This file won't be merged, just waiting to merge the first PR

PR #378, opened 2017-10-24
PR 4 of many to replace map[time.Time] with map[int64] - Shard

richardartoul:
Partially addresses #290


x/time/time.go
```diff
@@ -0,0 +1,17 @@
+package time
```
richardartoul: Wont be merged, just waiting for first PR to get merged first

PR #379, opened 2017-10-24
PR 5 of many to replace map[time.Time] with map[int64] - Integration + MISC unit tests

richardartoul:
Partially addresses #290


x/time/time.go
```diff
@@ -0,0 +1,17 @@
+package time
```
richardartoul: This file won't be merged, just waiting for the first PR to be merged

prateek:
We need some way to exercise the case where time on different hosts are set to different timezones. Egs: m3dbingester/query is running with Eastern time, and the m3dbnode is running in UTC; or vice versa; or two m3dbnodes have different timezones. Maybe some way to simulate that with integration tests. What do you think? 

Maybe also add in a go1.9 travis target.

richardartoul:
We have an outstanding PR for go 1.9 that I'll get merged once all these PRs are in. I can't add it until we've fixed everything.

Agreed it would be good to have that test, although I'm not sure if this PR is the place for it. With these changes we should be strictly better off right? We're removing the concept of timezones at least from some of our code. Once the map issue is resolved, what other types of issues do you see us having with wrong timezones? The only thing I can think of is if we tried to do something like

```golang
time1 := time.Time{}
time2 := time.Time{}
if time1 == time2 {
    ....
}
```

richardartoul:
@prateek https://github.com/m3db/m3db/pull/345

integration/roundtrip_test.go
```diff
@@ -27,7 +27,7 @@ import (
 	"time"
 
 	"github.com/m3db/m3db/integration/generate"
-
+	xtime "github.com/m3db/m3x/time"
```
prateek: nit: newline after this

prateek:
Mainly looks good. 

Re: #345 ~ Sure. Can keep this change specific to only map[time.Time] if you prefer. 

Re the new integration tests: I was thinking of something where we spin up a single node using some random TZ, and we use a client which writes values in a different timezone. Would verify that we're treating timezones correctly internally.

richardartoul:
@prateek Ok I think I see what you're saying. Can I file an issue for that and make it a separate PR? Feels slightly orthogonal to this PR

prateek:
Feels related because the `time.Time` keys would resolve time differently from `UnixNano` if a timezone were ever set. Afaik it's never been set, but a test to ensure we're doing the right thing would make me feel more confident of that.

integration/write_read_timezone_test.go
```diff
@@ -0,0 +1,126 @@
...
+	iters, err := session.FetchAll(namespace, []string{"some-id-1", "some-id-2"}, startNy, startNy.Add(1*time.Hour))
+	require.NoError(t, err)
+
+	// Assert datapoints match what we wrote
+	for i, iter := range iters.Iters() {
+		for j := 0; iter.Next(); j++ {
+			dp, _, _ := iter.Current()
+			expectedDatapoint := writeSeries[i].datapoints[j]
+			// Datapoints will comeback with the timezone set to the local timezone
+			// of the machine that the client is runnign on. The Equal() method ensures
```
prateek: typo:s/runnign/running

integration/write_read_timezone_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+// Make sure that everything works properly end-to-end even if the client issues
+// a write in a timezone other than that of the server.
+func TestWriteReadTimezone(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
+	// Ensure that the test is running with the local timezone set to US/Pacific
```
prateek: could you put 1-2 lines explaining why you prefer doing this over changing the clock.Now

Issue #304, opened 2017-06-16
Cleanup files for shards no longer owned


robskillington:
After shards move around the cluster the node doesn't cleanup the files for shards it no longer owns.

PR #381, opened 2017-10-25
Extend the the file cleanup manager to handle inactive filesets from shards no longer owned

katezaps:


persist/fs/files.go
```diff
@@ -232,6 +245,32 @@ func FilesetBefore(filePathPrefix string, namespace ts.ID, shard uint32, t time.
 	return filesBefore(matched, t)
 }
 
+// DeleteInactiveFilesets deletes any filesets that are not currently owned by the namespace
+func DeleteInactiveFilesets(filePathPrefix string, namespace ts.ID, activeShards []uint32) error {
+	var toDelete []string
+	dirs := make(map[string]bool)
```
robskillington: You can make this just a `map[string]struct{}` when constructing sets.  This is what we traditionally use, so perhaps for consistency update to use that?

i.e.
```go
dirs := make(map[string]struct{})

// set
dirs[dir] = struct{}{}

// check
if _, ok := dirs[dir]; !ok {
  // ...
}
```

persist/fs/files.go
```diff
@@ -232,6 +245,32 @@ func FilesetBefore(filePathPrefix string, namespace ts.ID, shard uint32, t time.
...
+	}
+
+	//can be extracted to a function perhaps?
+	for _, dir := range activeShardDirs {
+		dirs[dir] = true
+	}
+
+	//is there a general slice function that can handle this?
+	for _, dir := range allShardDirs {
+		ok := dirs[dir]
```
robskillington: To make this a little more robust perhaps we should compare the basename?  i.e. in case there's trailing slashes, or some symlink thing going on?
https://golang.org/pkg/path/#Base

robskillington:
Overall looks great, perhaps we should also write an integration test?  There's a few in `integration/` that aren't too insanely complex that could be copied and slightly altered I think.

~
```diff
@@ -0,0 +1,425 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
prateek: do you need to delete this file

client/client_mock.go
```diff
@@ -24,8 +24,7 @@
 package client
 
 import (
-	time0 "time"
-
+	gomock "github.com/golang/mock/gomock"
```
prateek: could you drop the mocks that haven't been altered

storage/cleanup.go
```diff
@@ -89,6 +89,11 @@ func (m *cleanupManager) Cleanup(t time.Time) error {
 			"encountered errors when cleaning up fileset files for %v: %v", t, err))
 	}
 
+	if err := m.deleteInactiveFilesetFiles(); err != nil {
```
prateek: Maybe also address the case of deleting files for namespaces that are no longer active.

robskillington: Good catch, yeah we should do this.

~
katezaps: I have no idea what file I deleted 🤷‍♀️

katezaps: OH the `~` ? interesting, good catch thanks

storage/cleanup.go
katezaps: should I include this as part of this PR or separately?

persist/fs/files.go
katezaps: I use https://golang.org/pkg/os/#File.Readdirnames
which gives back the names of the dirs instead of the full path.

katezaps: I wanted to keep it simple by re-attaching the osPath to the dirnames, but I will remove the logic that handles that

katezaps: Actually, it might make sense to make a small struct of shard `uint32` and its associated directory

storage/cleanup.go
robskillington: You could do this as a separate PR if you feel like it won't fundamentally change this code much?

katezaps: I think it would be a separate function to attach as a similar flow and the findDirectories function can be used in either case, just `DeleteInactiveNamespaces` which would probably then just become `DeleteInactiveFiles` instead of Filesets or Namespaces.

persist/fs/files.go
```diff
@@ -269,6 +306,24 @@ func findFiles(fileDir string, pattern string, fn toSortableFn) ([]string, error
 	return matched, nil
 }
 
+func findDirectories(osPath string) (map[string]string, error) {
```
robskillington: Perhaps `type fileBaseNamesToPaths map[string]string` to pass back a more readable type here?

storage/cleanup.go
```diff
@@ -117,6 +122,19 @@ func (m *cleanupManager) Report() {
 	}
 }
 
+func (m *cleanupManager) deleteInactiveFilesetFiles() error {
+	multiErr := xerrors.NewMultiError()
+	namespaces, err := m.database.GetOwnedNamespaces()
+	if err != nil {
+		return err
+	}
+	for _, n := range namespaces {
```
prateek: I think you can localise the code a bit more. For e.g. I don't think you need any `namespace.go` changes if you do the following: 

- add `deleteInactiveFilesetsFn` to the cleanup manager instead of namespace, 
- rewrite the `deleteInactiveFilesetFiles()` function to retrieve the shards from the database namespace, something like:

```
func (m *cleanupManager) deleteInactiveFilesetFiles() error {
	namespaces, err := m.database.GetOwnedNamespaces()
 	if err != nil {
		return err
	}
	for _, n := range namespaces {
            id := n.ID()
            shards := n.Shards()
            multiErr = multiErr.Add(m.deleteInactiveFilesetsFn(m.filePathPrefix, id, shards)))
        }
        return multiErr.FinalError()
}
```

persist/fs/files.go
```diff
@@ -269,6 +306,26 @@ func findFiles(fileDir string, pattern string, fn toSortableFn) ([]string, error
...
 }
 
+type directoryNamesToPaths map[string]string
+
+func findSubDirectoriesAndPaths(directoryPath string) (directoryNamesToPaths, error) {
+	parent, err := os.Open(directoryPath)
+	if err != nil {
+		return nil, err
+	}
+	defer parent.Close()
```
prateek: iirc `Close()` returns an error, could you check it instead of dropping silently

persist/fs/files.go
```diff
@@ -232,6 +245,30 @@ func FilesetBefore(filePathPrefix string, namespace ts.ID, shard uint32, t time.
...
+func DeleteInactiveFilesets(filePathPrefix string, namespace ts.ID, activeShards []uint32) error {
+	var toDelete []string
+	activeDirNames := make(map[string]struct{})
+	namespaceDirPath := NamespaceDirPath(filePathPrefix, namespace)
+	allShardDirs, err := findSubDirectoriesAndPaths(namespaceDirPath)
+	if err != nil {
+		return nil
+	}
+
+	//create shard set, might also be useful to just send in as strings?
```
prateek: nit: could you drop the question in the comment. Minor nit: convention is to put a space after `//`, ie `// Create ...` instead of `//create ...`

persist/fs/files.go
katezaps: ahh thanks!

storage/cleanup.go
katezaps: I think it makes sense for a namespace to handle how its deleting the files, particularly since the cleanup function is called wrt a namespace.

If that's the case then the cleanup manager should handle all forms of deletion, vs being called when a namespace is removed.

storage/cleanup.go
katezaps: Hey so I did some poking around, and it looks like when a namespace no longer has shards its deleted immediately after. I'm not sure if its necessary to do the deletion here, and if it is,

then I think 2 things need to happen

1) the deletion function is localized to the cleanup manager
2) namespaces don't try to delete associated files when they lose all their shards

would there be any case where a namespace has files that don't get deleted after (2) happens?

storage/cleanup.go
prateek: I'm in two minds about this: 

On the one hand, we already have `namespace.CleanupFileset` and `shard.CleanupFileset` which perform fs operations, so sticking with that way of doing things, the way you have the code makes sense. 

That said, the way the code is now, `databaseNamespace` and `databaseShard` both perform fs operations. Once you add the cleanup for unknown namespaces, `databaseCleanupManager` will also perform fs operations. This feels like bad code structure. It'd be easier to reason about/modify if all fs three operations were performed in the `cleanupManager`. 

What do you think?

storage/cleanup.go
prateek: I was more worried about the case where a namespace is removed from a database. It won't loose any of its shards when this happens, so there'd be zombie dirs/files we'd need to cleanup in the namespace's base directory.

storage/cleanup.go
robskillington: I can get on board with this, it should make it more central and pragmatic to make changes (can make changes in one place vs across a bunch of places).

katezaps: Alright, I like the idea of normalizing /all/ deletion to the cleanup manager. I'll update this PR with those changes.

integration/disk_cleanup_for_inactive_fileset_files_test.go
```diff
@@ -0,0 +1,85 @@
...
+	fileTimes := make([]time.Time, numTimes)
+	now := testSetup.getNowFn()
+	for i := 0; i < numTimes; i++ {
+		fileTimes[i] = now.Add(time.Duration(i) * blockSize)
+	}
+	writeFilesetFiles(t, testSetup.storageOpts, md, shard, fileTimes)
+	writeCommitLogs(t, filePathPrefix, fileTimes)
+
+	namespaceDir := newNamespaceDir(testSetup.storageOpts, md)
+	fmt.Println("this is the namespace", namespaceDir)
```
robskillington: We can remove this now right?

integration/disk_cleanup_helpers.go
```diff
@@ -41,6 +42,14 @@ var (
 )
 
 // nolint: deadcode
+func newNamespaceDir(storageOpts storage.Options, md namespace.Metadata) string {
+	fsOpts := storageOpts.CommitLogOptions().FilesystemOptions()
+	filePathPrefix := fsOpts.FilePathPrefix()
+	fmt.Println(fs.NamespaceDirPath(filePathPrefix, md.ID()))
```
robskillington: This too?

persist/fs/files.go
robskillington: +1

persist/fs/files.go
robskillington: +1

persist/fs/files_test.go
```diff
@@ -164,6 +164,31 @@ func TestDeleteFiles(t *testing.T) {
...
 }
 
+func TestDeleteInactiveFiles(t *testing.T) {
+	tempPrefix := "temp/"
+	namespaceDir := NamespaceDirPath(tempPrefix, testNs1ID)
+
+	shards := []uint32{uint32(4), uint32(5), uint32(6)}
+	for _, shard := range shards {
+		shardDir := ShardDirPath(tempPrefix, testNs1ID, shard)
+		err := os.MkdirAll(shardDir, os.FileMode(uint32(0777)))
```
robskillington: Better to use `defaultNewDirectoryMode` instead of 777

persist/fs/write.go
```diff
@@ -85,6 +86,7 @@ func (w *writer) Open(
 	blockStart time.Time,
 ) error {
 	shardDir := ShardDirPath(w.filePathPrefix, namespace, shard)
+	fmt.Println("shard dir path", shardDir)
```
robskillington: Can remove this yeah?

storage/namespace.go
```diff
@@ -670,6 +675,17 @@ func (n *dbNamespace) CleanupFileset(earliestToRetain time.Time) error {
...
 
+func (n *dbNamespace) DeleteInactiveFilesets() error {
+	var shardIds []uint32
+	filePathPrefix := n.opts.CommitLogOptions().FilesystemOptions().FilePathPrefix()
+	activeShards := n.getOwnedShards()
+	for _, shard := range activeShards {
+		shardIds = append(shardIds, shard.ID())
+	}
+
+	return n.deleteInactiveFilesetsFn(filePathPrefix, n.id, shardIds)
```
robskillington: We still going to move this into cleanup mgr directly as per other comment thread?

katezaps: yes, that's the plan

katezaps: i'll add a comment in the code

persist/fs/write.go
katezaps: yeah i caught a few others, this was actually causing the integration test to fail i think

katezaps:
fixes #304

PR #387, opened 2017-11-06
Refactor cleanup to manage all deletion

katezaps:


storage/cleanup.go
```diff
@@ -122,14 +131,38 @@ func (m *cleanupManager) Report() {
...
 	if err != nil {
 		return err
 	}
 	for _, n := range namespaces {
-		multiErr = multiErr.Add(n.DeleteInactiveFilesets())
+		var activeShards []string
+		//replace with injection later
+		namespaceDirPath := fs.NamespaceDirPath(filePathPrefix, n.ID())
+		for _, s := range n.Shards() {
+			shard := fmt.Sprintf("%v", s.ID())
```
robskillington: Let's use the more specific `%d` here, is good hygiene.

robskillington:
Other than the 1 comment looking good 🎉

storage/types.go
```diff
@@ -230,10 +230,6 @@ type databaseNamespace interface {
 	// CleanupFileset cleans up fileset files
 	CleanupFileset(earliestToRetain time.Time) error
```
prateek: will function this be removed too?

persist/fs/fs_mock.go
```diff
@@ -27,8 +27,8 @@ import (
 	gomock "github.com/golang/mock/gomock"
 	ts "github.com/m3db/m3db/ts"
 	checked "github.com/m3db/m3x/checked"
-	time "github.com/m3db/m3x/time"
-	time0 "time"
+	time0 "github.com/m3db/m3x/time"
```
prateek: could you drop unchanged files.

katezaps: Isn't this changed with the addition to `fs`?

prateek: The only changes I see in this file are for the import alias rename, no signature changes beyond that yea?

katezaps: Fair point, I just assumed it changed bc I edited the `fs` file itself. I'll remove it.

integration/disk_cleanup_deletes_inactive_directories_test.go
```diff
@@ -75,7 +76,29 @@ func TestDiskCleansupInactiveFilesets(t *testing.T) {
 	require.NoError(t, err)
 	testSetup.db.AssignShardSet(shardSet)
 
+	// So some gnarly stuff here to make the testing package a bit more synchronous
+	var fsCleanupErr chan error
+	var nsCleanupErr chan error
 	// Check if files have been deleted
 	waitTimeout := 30 * time.Second
```
robskillington:
```

File: integration/disk_cleanup_deletes_inactive_directories_test.go [hunk 5]
```diff
@@ -75,7 +76,29 @@ func TestDiskCleansupInactiveFilesets(t *testing.T) {
 	require.NoError(t, err)
 	testSetup.db.AssignShardSet(shardSet)
 
+	// So some gnarly stuff here to make the testing package a bit more synchronous
+	var fsCleanupErr chan error
```
robskillington: var (

robskillington:
```

Comment (robskillington) [inline integration/disk_cleanup_deletes_inactive_directories_test.go]: These would be good in a var block:

```go
var (
  fsCleanupErr chan error
  nsCleanupErr chan error
  waitTimeout = 30 * time.Second
)
```

integration/disk_cleanup_deletes_inactive_directories_test.go
```diff
@@ -75,7 +76,29 @@ func TestDiskCleansupInactiveFilesets(t *testing.T) {
...
+	}()
+
+	// Assigning a shardset of length one will mean there are extra namespaces
+	// that will need to be removed via cleanup
+	shardSet, err = sharding.NewShardSet(shards[:1], shardSet.HashFn())
+	require.NoError(t, err)
+	testSetup.db.AssignShardSet(shardSet)
+
+	// Needs an extra 30 seconds due to race conditions
+	waitTimeout = 160 * time.Second
```
robskillington: Better to use two different variables to pass to the different goroutines rather than modifying this at the top level, it's not guarenteed they won't read them in a non-racy way.

integration/disk_cleanup_deletes_inactive_directories_test.go
```diff
@@ -75,7 +76,29 @@ func TestDiskCleansupInactiveFilesets(t *testing.T) {
...
+	go func() {
+		nsCleanupErr <- waitUntilNamespacesCleanedUp(testSetup, filePathPrefix,
+			testNamespaces[1], waitTimeout)
+		require.NoError(t, <-nsCleanupErr)
+	}()
```
robskillington: You'll need to wait for both the goroutines that got launched here to finish before the test finishes, otherwise it might look like it passed when some of those routines haven't finished yet.

Also is there a reason you're using channels here when you read straight off the channel afterwards?

integration/disk_cleanup_helpers.go
```diff
@@ -151,11 +151,47 @@ func waitUntilDataCleanedUpExtended(
...
+func waitUntilNamespacesCleanedUp(testSetup *testSetup, filePathPrefix string, namespace ts.ID, waitTimeout time.Duration) error {
+
+	// The idea would be to have a channel that has notifs (ie, --> func if val 1, 2, or 2)
+	// and pass that to a function that does a function based on that state
+	// this would involve locks on the channel to check the functions aren't called
+	// more than once
+
+	c := make(chan error)
+	go func() {
+		c <- waitUntilNamespacesHaveReset(testSetup)
```
robskillington: Are you wanting to wait for this asynchronously and use the result or synchronously here?

integration/disk_cleanup_deletes_inactive_directories_test.go
```diff
@@ -50,6 +50,7 @@ func TestDiskCleansupInactiveFilesets(t *testing.T) {
 	log.Debug("disk cleanup test")
 	require.NoError(t, testSetup.startServer())
 	log.Debug("server is now up")
+	testSetup.waitUntilServerIsUp()
```
robskillington: This is superfluous I believe no?  `startServer` waits by calling `waitUntilServerIsUp` already.

integration/disk_cleanup_helpers.go
```diff
@@ -151,11 +151,47 @@ func waitUntilDataCleanedUpExtended(
...
+	}
+
+	if waitUntil(dataCleanedUp, waitTimeout) {
+		return nil
+	}
+	return errDataCleanupTimedOut
+}
+
+// nolint: deadcode, unused
+func waitUntilNamespacesHaveReset(testSetup *testSetup) error {
```
robskillington: Seems `stopServer` is not called here so `waitUntilServerIsDown()` will never finish I believe?

storage/namespace_test.go
```diff
@@ -429,36 +429,31 @@ func TestNamespaceFlushAllShards(t *testing.T) {
...
-	for i := range errs {
-		shard := NewMockdatabaseShard(ctrl)
-		shard.EXPECT().CleanupFileset(earliestToRetain).Return(errs[i])
-		ns.shards[testShardIDs[i].ID()] = shard
-	}
-
-	require.Error(t, ns.CleanupFileset(earliestToRetain))
-}
-
+//func testnamespacecleanupFilesetDontNeedCleanup(t *testing.T) {
```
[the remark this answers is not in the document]
robskillington: We probably still want at least one test that ensures each shard in a namespace gets the `CleanupFileset()` call, even if just for test coverage.  Integration tests don't add to our test coverage unfortunately as they don't run from the package itself.

storage/types.go
```diff
@@ -153,8 +153,11 @@ type Namespace interface {
 	// NumSeries returns the number of series in the namespace
 	NumSeries() int64
 
-	// Shards returns the shards
+	// Shards returns the shard as storange shards
```
robskillington: `Shards returns the namespace's storage shards` perhaps?

storage/types.go
```diff
@@ -153,8 +153,11 @@ type Namespace interface {
...
 	Shards() []Shard
+
+	// GetOwnedShards returns the database shards
+	GetOwnedShards() []databaseShard
```
prateek: can this be part of `databaseNamespace` instead of `Namespace`?

robskillington: +1

integration/disk_cleanup_deletes_inactive_directories_test.go
```diff
@@ -23,59 +23,83 @@
...
+	testSetup, err := newTestSetup(t, testOpts, nil)
 	require.NoError(t, err)
-	defer testSetup.close()
 
 	md := testSetup.namespaceMetadataOrFail(testNamespaces[0])
-	blockSize := md.Options().RetentionOptions().BlockSize()
 	filePathPrefix := testSetup.storageOpts.CommitLogOptions().FilesystemOptions().FilePathPrefix()
 
-	// Start the server
+	// Start tte server
```
robskillington: Super minor nit: typo here

integration/disk_cleanup_helpers.go
```diff
@@ -151,11 +152,50 @@ func waitUntilDataCleanedUpExtended(
...
+
+	if waitUntil(dataCleanedUp, waitTimeout) {
+		return nil
+	}
+	return errDataCleanupTimedOut
+}
+
+// nolint: deadcode, unused
+func waitUntilNamespacesHaveReset(testSetup *testSetup, newNamespaces []namespace.Metadata, newShardSet sharding.ShardSet) (*testSetup, error) {
+	testSetup.stopServer()
```
robskillington: Should also check the error return from stopServer too yeah?

PR #390, opened 2017-11-07
Fix broken test

### LINKED ISSUE #389: Fix broken test TestDeleteInactiveFiles

richardartoul:
Closes #389

### DIFF
--- a/persist/fs/files_test.go
+++ b/persist/fs/files_test.go
@@ -165,7 +165,8 @@ func TestDeleteFiles(t *testing.T) {
 }
 
 func TestDeleteInactiveFiles(t *testing.T) {
-	tempPrefix := os.TempDir()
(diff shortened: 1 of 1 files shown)


persist/fs/files_test.go
```diff
@@ -165,27 +165,29 @@ func TestDeleteFiles(t *testing.T) {
 }
 
 func TestDeleteInactiveFiles(t *testing.T) {
-	tempPrefix := "temp/"
+	tempPrefix, err := ioutil.TempDir("", "foo")
+	require.NoError(t, err)
+	defer func() {
```
katezaps: so is this the trick to getting it to eval later?

robskillington: `temp` would've only been available if you made it on your own machine, `ioutil.TempDir(...)` creates a temporary directory for use in a directory the stdlib manages.

PR #394, opened 2017-11-09
Add FetchBlocksMetadataRawV2 endpoint

richardartoul:


network/server/tchannelthrift/node/service.go
```diff
@@ -478,6 +485,139 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
...
 
+func (s *service) FetchBlocksMetadataRawV2(tctx thrift.Context, req *rpc.FetchBlocksMetadataRawV2Request) (*rpc.FetchBlocksMetadataRawV2Result_, error) {
+	if s.db.IsOverloaded() {
+		s.metrics.overloadRejected.Inc(1)
+		return nil, tterrors.NewInternalError(errServerIsOverloaded)
+	}
+
+	var err error
+	callStart := s.nowFn()
+	defer func() {
```
richardartoul: Is this path "cold" enough to use a defer?

network/server/tchannelthrift/node/service.go
```diff
@@ -478,6 +485,139 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
...
+		ctx, nsID, uint32(req.Shard), start, end, req.Limit, pageToken.ShardIndex, opts)
+	if err != nil {
+		s.metrics.fetchBlocksMetadata.ReportError(s.nowFn().Sub(callStart))
+		return nil, convert.ToRPCError(err)
+	}
+	ctx.RegisterFinalizer(context.FinalizerFn(fetchedMetadata.Close))
+
+	result, err := s.getFetchBlocksMetadataRawV2Result(nextShardIndex, opts, fetchedMetadata)
+	// Finalize pooled datastructures regardless of errors
+	ctx.RegisterFinalizer(context.FinalizerFn(func() {
```
richardartoul: Similar to the comment above, is it ok to create this anonymous function everytime? Originally had this as a separate struct like the original endpoint but seemed like a lot of boilerplate

generated/proto/pagetoken/pagetoken.proto
```diff
@@ -0,0 +1,6 @@
+syntax = "proto3";
+package pagetoken;
+
+message PageToken {
+	int64 shardIndex = 1;
+}
```
robskillington: nit: Add an endline to the end of this file.

network/server/tchannelthrift/node/service.go
robskillington: Single defer is not so bad, it's been sped up in 1.8 https://golang.org/doc/go1.8#defer - it's a batch call so it's probably ok.  

Although you shouldn't always `ReportSuccess(...)` you can use "ReportSuccessOrError(...)` if you want.

network/server/tchannelthrift/node/service.go
```diff
@@ -478,6 +485,139 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
...
+	if s.db.IsOverloaded() {
+		s.metrics.overloadRejected.Inc(1)
+		return nil, tterrors.NewInternalError(errServerIsOverloaded)
+	}
+
+	var err error
+	callStart := s.nowFn()
+	defer func() {
+		if err != nil {
+			s.metrics.fetchBlocksMetadata.ReportSuccess(s.nowFn().Sub(callStart))
```
robskillington: Instead of an if/else here, better to use `ReportSuccessOrError(...)`

network/server/tchannelthrift/node/service.go
```diff
@@ -478,6 +485,139 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
...
+		start     = time.Unix(0, req.RangeStart)
+		end       = time.Unix(0, req.RangeEnd)
+		pageToken = &pt.PageToken{}
+	)
+	err = proto.Unmarshal(req.PageToken, pageToken)
+	if err != nil {
+		return nil, tterrors.NewBadRequestError(errInvalidPageToken)
+	}
+
+	fetchedMetadata, nextShardIndex, err := s.db.FetchBlocksMetadata(
```
robskillington: Would this `err` get a new definition at this block level?  i.e. not 100% sure that the defer would capture if this `err` is not nil or not, probably can test with a simple playground:

network/server/tchannelthrift/node/service.go
robskillington: Yeah this would actually be a heap lambda alloc per iteration, defer has been speed up recently so it could just be stack allocated but this is definitely heap allocated.  It'd be better to put the method on the struct if possible.

network/server/tchannelthrift/node/service.go
```diff
@@ -478,6 +485,139 @@ func (s *service) FetchBlocksMetadataRaw(tctx thrift.Context, req *rpc.FetchBloc
...
+	if err != nil {
+		s.metrics.fetchBlocksMetadata.ReportError(s.nowFn().Sub(callStart))
+		return nil, convert.ToRPCError(err)
+	}
+	ctx.RegisterFinalizer(context.FinalizerFn(fetchedMetadata.Close))
+
+	result, err := s.getFetchBlocksMetadataRawV2Result(nextShardIndex, opts, fetchedMetadata)
+	// Finalize pooled datastructures regardless of errors
+	ctx.RegisterFinalizer(context.FinalizerFn(func() {
+		s.finalizeFetchBlocksMetadataRawV2Result(result)
```
robskillington: Should also only register finalizer err is not nil right?

richardartoul: I updated the comment / code to be more clear, but even in the error case there are pooled objects that need to be returned

generated/proto/pagetoken/pagetoken.proto
```diff
@@ -0,0 +1,6 @@
+syntax = "proto3";
+package pagetoken;
```
prateek: would it be better to have this be `client` ?

richardartoul: the package name? I'm open to renaming it, but client seems like a weird name because the token is actually opaque to the client, whereas its not opaque to the server

robskillington:
Hey @richardartoul seems metalint is failing?
```
network/server/tchannelthrift/node/service.go:881:19:warning: func (*service).finalizeFetchBlocksMetadataRawV2Result is unused (U1000) (unused)
```

richardartoul:
@robskillington Yeah I just forgot to remove that when I went back to using the struct that implements the Finalize() interface. Should be good now

robskillington:
You could land the `badtime` PR first and then this last test that is failing will get skipped and metalint should also be fixed.

richardartoul:
@robskillington Sounds good, thanks for fixing that other test. By switching it to adhoc we're skipping it entirely right? Is it an important test?

robskillington:
Yeah, its unfortunate we have to skip it.  It's mainly just to see that the server boots from a config file.

Travis kills the test with SIGKILL (you see "signal: killed" in logs) and its not clear why, I don't think its memory as it only ever hits 900mb on my machine for that test and Travis has 7.5gb (also integration tests get up to 3-3.5gb so that supports the case its not running out of memory).

PR #395, opened 2017-11-10
Update fileset index format to sorted string order with summaries and bloom filter

robskillington:
This is the beginning of a set of changes to move to remove all block references from memory except those that are being written to and those that have been loaded in recently as cached blocks due to reads.

This will write out the new file format and support reading it.

For now the seeker still loads the ID and offsets into memory, the next change will update it to load the summaries file and resolve IDs by reading the index entries on disk after searching the summaries for a jumping off point.


persist/encoding/decoder.go
```diff
@@ -42,3 +51,110 @@ type Decoder interface {
 	// DecodeLogEntry decodes commit log entry
 	DecodeLogEntry() (schema.LogEntry, error)
 }
+
+// DecoderStream is a data stream that is read by the decoder,
```
richardartoul: This comment is a little confusing. After reading it, I expected the constructor for DecoderStream to look like:

```golang
NewDecoderStream(b []byte, r io.Reader) DecoderStream {}
```

but instead it only accepts []byte, and the type of the internal reader is *bytes.Reader which is not an interface.

How would you use this interface with a ReaderWithDigest interface?

persist/encoding/msgpack/decoder.go
```diff
@@ -141,13 +153,45 @@ func (dec *decoder) decodeIndexInfo() schema.IndexInfo {
 	indexInfo.Start = dec.decodeVarint()
 	indexInfo.BlockSize = dec.decodeVarint()
 	indexInfo.Entries = dec.decodeVarint()
+	indexInfo.MajorVersion = dec.decodeVarint()
+	indexInfo.Summaries = dec.decodeIndexSummariesInfo()
+	indexInfo.BloomFilter = dec.decodeIndexBloomFilterInfo()
```
richardartoul: index files are per shard / block so we have a bloomfilter per shard/block right?

persist/fs/commitlog/commit_log_test.go
```diff
@@ -372,9 +360,11 @@ func TestReadCommitLogMissingMetadata(t *testing.T) {
 	for i := 0; i < 200; i++ {
 		willNotHaveMetadata := !(i%2 == 0)
 		allSeries = append(allSeries, testSeries(uint64(i), "hax", uint32(i%100)))
-		mockBitSet.indexTestReturn[uint(i)] = willNotHaveMetadata
+		if willNotHaveMetadata {
+			bitSet.Set(uint(i))
```
richardartoul: good call, I probably should have realized this myself

persist/fs/commitlog/writer.go
```diff
@@ -90,7 +91,7 @@ type writer struct {
 	chunkReserveHeader []byte
 	buffer             *bufio.Writer
 	sizeBuffer         []byte
-	seen               bitSet
+	seen               *xsets.BitSet
```
richardartoul: were you worried about the performance impact of having an interface here, or it just wasn't necessary?

persist/fs/mmap.go
```diff
@@ -0,0 +1,88 @@
...
+package fs
+
+import (
+	"fmt"
+	"os"
+)
+
+type mmapFileDesc struct {
+	// file is the *os.File ref to store
+	file **os.File
```
richardartoul: why is this a double ref?

persist/fs/mmap.go
```diff
@@ -0,0 +1,88 @@
...
+	// bytes is the []byte slice ref to store the mmap'd address
+	bytes *[]byte
```
richardartoul: same here, confused why we're using a pointer to a "pointer type"

persist/fs/mmap_linux.go
```diff
@@ -18,58 +18,52 @@
...
+	if length < hugePageSize {
+		return b, nil
 	}
-	b.values[idx] |= 1 << (i & wordMask)
-}
 
-func (b *bitSetImpl) clearAll() {
-	for i := range b.values {
-		b.values[i] = 0
+	if err := syscall.Madvise(b, syscall.MADV_HUGEPAGE); err != nil {
```
richardartoul: A comment explaining this would be nice. I found these two links helpful personally: 

1) http://man7.org/linux/man-pages/man2/madvise.2.html
2) https://www.kernel.org/doc/Documentation/vm/transhuge.txt

Seems like the gist of it is that you can reduce TLB misses by making the pages larger, at the risk of wasting a little extra memory. Is that basically it?

Also just out of curiosity, did this make a big difference in your benchmarks or is it just kind of a "best practices" thing?

persist/fs/options.go
```diff
@@ -30,6 +31,12 @@ import (
 )
 
 const (
+	// defaultIndexSummariesPercent is the default percent to save summaries for when writing summary files
```
richardartoul: This comment could be a little more clear: "defaultIndexSummariesPercent is the default percent of series for which an entry will be written into the metadata summary" or something like that. It took me awhile to figure out what this was until you explained it in person in the meeting

persist/fs/write.go
```diff
@@ -54,25 +64,55 @@ type writer struct {
 	err        error
 }
 
+type indexEntry struct {
```
richardartoul: Is there any benefit in optimizing the size of this struct?

From what I can tell, offsetForSummary never stores a value that is different than offset, so you could probably replace it with a bool. That doesn't actually change the size of the struct, but if you then replace size (or offset but size seems safer) with uint32 and re-order the fields, you can save 8 bytes per struct making it 40 bytes instead of 48 (if this is meaningful at all):

https://play.golang.org/p/dXGawT5YxL

Might be premature, I don't know how hot this is

persist/fs/write.go
```diff
@@ -21,14 +21,18 @@
 package fs
```
richardartoul: There is a lot going on in this file. I count 6 different file descriptors. Seems like generally speaking now (or sometime soon with the subsequent P.R's) would be a good time to write a few paragraphs about our new file formats (kind of what we have in commitlog.md) and then link to that documentation at the top of this page.

Otherwise as someone unfamiliar with the codebase its pretty difficult to decipher the what / why from code that is just manipulating bytes on disk if that makes sense.

persist/fs/write.go
richardartoul: Although it might make your sort faster too just by virtue of being able to fit more of them in the cache lines

persist/fs/write.go
```diff
@@ -195,33 +235,141 @@ func (w *writer) writeAll(
...
+	summariesApprox := float64(len(w.indexEntries)) * w.summariesPercent
+	summaryEvery := 0
+	if summariesApprox > 0 {
+		summaryEvery = int(math.Floor(float64(len(w.indexEntries)) / summariesApprox))
+	}
+
+	// Write the index entries and calculate the bloom filter
+	n, p := uint(w.currIdx), w.bloomFilterFalsePositivePercent
+	m, k := xsets.BloomFilterEstimate(n, p)
+	bloomFilter := xsets.NewBloomFilter(m, k)
```
richardartoul: Ok interesting so we don't even create the bloom filter until we're ready to flush to disk? I guess I thought we'd be creating it as we saw things, but I guess we don't need it until data has been flushed to disk, because if its still in memory we can just check.

persist/fs/write.go
```diff
@@ -195,33 +235,141 @@ func (w *writer) writeAll(
...
+
+	var offset int64
+	for i := range w.indexEntries {
+		id := w.indexEntries[i].id.Data().Get()
+
+		entry := schema.IndexEntry{
```
richardartoul: Kind of annoying we have to do this conversion because of a single field difference but I dont know if there is much that can be done about it

persist/fs/write.go
```diff
@@ -195,33 +235,141 @@ func (w *writer) writeAll(
...
-	}
+
+	w.indexEntries = append(w.indexEntries, entry)
 	w.currIdx++
 
 	return nil
 }
 
-func (w *writer) close() error {
+func (w *writer) writeIndexEntriesSummariesBloomFilter() error {
```
richardartoul: Seems like this function does more than the name implies. Maybe: "writeIndexEntriesAndIndexSummaryAndBloomFilter" or something. I know that's verbose but the existing name makes it seem like its writing one file when its writing 3

persist/fs/write.go
```diff
@@ -195,33 +235,141 @@ func (w *writer) writeAll(
...
+		},
 	}
 
 	w.encoder.Reset()
 	if err := w.encoder.EncodeIndexInfo(info); err != nil {
 		return err
 	}
 
-	if _, err := w.infoFdWithDigest.WriteBytes(w.encoder.Bytes()); err != nil {
+	_, err = w.infoFdWithDigest.Write(w.encoder.Bytes())
```
richardartoul: So the info about the index is stored in a separate file than the index itself?

persist/fs/write.go
```diff
@@ -195,33 +235,141 @@ func (w *writer) writeAll(
...
+	// leaking memory as we've released all refs while writing summaries
+	w.indexEntries = w.indexEntries[:0]
+
+	// Write the bloom filter bitset out
+	err := bloomFilter.BitSet().Write(w.bloomFilterFdWithDigest)
+	if err != nil {
+		return err
+	}
+
 	info := schema.IndexInfo{
```
richardartoul: Should we have constructors for these msgpack structs? Seems like it would be easy to leave out a field by accident and corrupt the whole thing

x/sets/bloom_filter.go
```diff
@@ -0,0 +1,151 @@
...
+type BloomFilter struct {
+	m   uint64
+	k   uint64
+	set *BitSet
+}
+
+// NewBloomFilter creates a new bloom filter that can represent
+// m elements with k hashes.
+func NewBloomFilter(m uint, k uint) *BloomFilter {
+	if m < 1 {
```
richardartoul: I wonder if it should return an error if these values are invalid...might be annoying to add the error value everywhere though

richardartoul:
Looks good to me for the most part. My main feedback is that it would be really nice to document these new file formats and how they interact (even a few paragraphs) and then link to them from the write.go file

persist/encoding/decoder.go
robskillington: `NewDecoderStream(...)` in this package is just a default offering for users to use if they don't need to provide the reader themselves.

There is a private `newDecoderStream(...)` in the `persist/fs package` for that use case.  This is so that users who don't need to and already have just a byte slice can still easily pass a buffer to decode into the msgpack decoder.

persist/encoding/msgpack/decoder.go
robskillington: Correct yup.

persist/fs/commitlog/writer.go
robskillington: I was thinking that this might get moved to `m3x` at some point and if so to help avoid breaking changes when we add methods to things it might be worth keeping the pretty simple data structures that don't need to get mocked out as just structs.  

So this choice was mainly for reducing cross repo breaking changes.  

If it was a more complicated/logic based structured then we probably would've want to mock it.

persist/fs/mmap.go
robskillington: This is so you can pass in the ptr you wish to have set.

persist/fs/mmap.go
robskillington: Yeah this is so that the API call to `mmapFiles(...)` can populate refs for you.

persist/fs/mmap_linux.go
robskillington: > Seems like the gist of it is that you can reduce TLB misses by making the pages larger, at the risk of wasting a little extra memory. Is that basically it?

persist/fs/options.go
robskillington: For sure, will update the comment.  Thanks for the note.

persist/fs/write.go
robskillington: The two offset values are actually different.  One is the offset into the data file, the `offsetForSummary` is the offset of this record in the index file that the summary entry will point to.  I'll update these to be `dataOffset` and `indexOffset` perhaps.

Yeah size could be reduced to be a uint32 for sure, offsets probably shouldn't be in case we ever have a file >4gb.  

I'm pretty sure that the size of the struct won't benefit from just a single field being a uint32 though due to alignment.

persist/fs/write.go
robskillington: Yup sounds good, I'll write a `storage.md` and reference it from this module.  And maybe a short paragraph in the source file itself.

persist/fs/write.go
robskillington: Right yeah.  We'll experiment with how much time this takes to do sequentially all during a flush vs amortizing it during series insertions.

persist/fs/write.go
robskillington: Yeah, not too big a deal it'll just be a memcopy to a stack allocated piece of memory though.

persist/fs/write.go
robskillington: Yeah I might break some of this up into smaller fns and have a second attempt at naming this.

persist/fs/write.go
robskillington: Correct, right now all files are either just repeated records or a single record.  i.e. always a single type in one file.

persist/fs/write.go
robskillington: There's usually only a single place where we ever construct these and the unit tests are pretty rigorous.  Since there's only one piece of code constructing these msgpack structs I'd prefer the readability when setting the fields rather than the safety added of a constructor (where you don't know what each arg is when reading the code necessarily and maybe you specify the wrong local variable to the wrong argument).

If tons of different places constructed these types then I would agree with you.

x/sets/bloom_filter.go
robskillington: Possibly, although am definitely trying to keep the generic datastructures as simple as possible to use so its more ergonomic.  I think in this case keeping both bitset and bloomfilter without an error is more ergonomic currently.  If there was some other args that could truly screw up things then at that point I'd lean towards an error here.  

Overall a good comment but I think just for ergonomics we should probably avoid returning an error for this API.

storage/options.go
```diff
@@ -147,6 +147,7 @@ func newOptions(poolOpts pool.ObjectPoolOptions) Options {
 	})
 	bytesPool.Init()
 	seriesOpts := series.NewOptions()
+	pm, _ := fs.NewPersistManager(fs.NewOptions())
```
prateek: could you check error here and add a `Validate()`/bubble up/panic for it.

robskillington: Yeah =[  

It's going to be not ideal because will have to cache the error (to be able to return it from Validate) and then return it if its not nil or something like that.  

Will do.

digest/reader.go
```diff
@@ -164,3 +165,80 @@ func (r *fdWithDigestContentsReader) ReadDigest() (uint32, error) {
 	}
 	return r.digestBuf.ReadDigest(), nil
 }
+
+// ReaderWithDigest is a reader that that calculates a digest
```
xichen2020: Any chance of code sharing with `FdWithDigestReader`? The implementation here looks very similar.

persist/encoding/decoder.go
```diff
@@ -42,3 +51,110 @@ type Decoder interface {
...
+// it takes both a reader and the underlying backing bytes.
+// This is constructed this way since the decoder needs access
+// to the backing bytes when taking refs directly for decoding byte
+// slices without allocating bytes itself but also needs to progress
+// the reader (for instance when a reader is a ReaderWithDigest that
+// is calculating a digest as its being read).
+type DecoderStream interface {
+	io.Reader
+
+	Reset(b []byte)
```
xichen2020: nit: add some comments here for consistency?

persist/encoding/decoder.go
```diff
@@ -42,3 +51,110 @@ type Decoder interface {
...
+	Skip(length int64) error
+
+	// Remaining returns the remaining bytes in the stream.
+	Remaining() int64
+}
+
+type decoderStream struct {
+	reader       *bytes.Reader
+	bytes        []byte
+	lastReadByte int
```
xichen2020: Any reason these are `int` instead of `byte`?

xichen2020: Why do we keep track of the last unread byte / read byte manually instead of using the APIs provided by `bytes.Reader`?

persist/fs/read.go
```diff
@@ -63,43 +66,159 @@ type reader struct {
...
+	metadataRead           int
+	prologue               []byte
+	decoder                encoding.Decoder
+	digestBuf              digest.Buffer
+	bytesPool              pool.CheckedBytesPool
+
+	indexEntriesByOffset []schema.IndexEntry
+}
+
+type indexEntriesByOffset []schema.IndexEntry
```
xichen2020: nit: `indexEntriesByOffsetAsc` might be more clear.

glide.yaml
```diff
@@ -31,8 +31,10 @@ import:
 - package: github.com/spaolacci/murmur3
```
prateek: do we need both this and the stack version?

.travis.yml
```diff
@@ -3,6 +3,9 @@ go:
   - 1.7.6
   - 1.8.3
   - 1.9.2
+branches:
```
prateek: do you want to land this too?

storage/options.go
```diff
@@ -206,6 +210,16 @@ func (o *options) Validate() error {
 		}
 	}
 
+	// validate that persist manager is present, if not return
+	// error if error occurred during default creation otherwise
+	// it was set to nil by a caller
+	if o.persistManager == nil {
+		if err := o.persistManagerErr; err != nil {
```
prateek: How about just not setting persistManager by default, and always requiring users to override it. Would be able to avoid this ugliness

persist/schema/types.go
```diff
@@ -20,14 +20,33 @@
 
 package schema
 
+// MajorVersion is the major schema version for a set of fileset files,
+// this is only incremented when breaking changes are introduced and
+// tooling needs to upgrade older files to newer files before a server restart
+const MajorVersion = 1
```
prateek: +1

persist/schema/types.go
```diff
@@ -20,14 +20,33 @@
...
+	BlockSize    int64
+	Entries      int64
+	MajorVersion int64
+	Summaries    IndexSummariesInfo
+	BloomFilter  IndexBloomFilterInfo
+}
+
+// IndexSummariesInfo stores metadata about the summaries
+type IndexSummariesInfo struct {
+	Summaries int64
```
prateek: Should this be `NumSummaries`?

.metalinter.json
```diff
@@ -8,8 +8,8 @@
       "Command": "gosimple -tags integration",
       "Pattern": "PATH:LINE:COL:MESSAGE"
     },
-    "maptime": {
-      "Command": "maptime -tags integration",
+    "badtime": {
+      "Command": "badtime -tags integration",
```
prateek: could you put all the build tags we use in m3dbnode (`-tags integration,adhoc` per https://github.com/m3db/m3db/pull/399/files/c3dad4b4f019058fb6a02bbb68e458395f5b9616#diff-deadcda208014667c463990fd3af6db1)

persist/fs/write.go
prateek: +1, I found https://code.uberinternal.com/P79869 helpful while reviewing the code

persist/fs/write.go
prateek: https://github.com/dominikh/go-tools/tree/master/cmd/structlayout might be helpful if you wanna go down this road

persist/fs/write.go
prateek: +1, breaking into smaller fns would help readability.

persist/fs/read.go
```diff
@@ -318,5 +466,24 @@ func (r *reader) EntriesRead() int {
...
 
 func (r *reader) Close() error {
-	return r.dataFdWithDigest.Close()
+	for i := 0; i < len(r.indexEntriesByOffset); i++ {
+		r.indexEntriesByOffset[i].ID = nil
+	}
+	r.indexEntriesByOffset = r.indexEntriesByOffset[:0]
+
+	if err := munmap(r.indexMmap); err != nil {
+		return err
```
prateek: Might be better semantics to collect all errors and return once all operations have been attempted.

persist/fs/read.go
```diff
@@ -302,7 +437,20 @@ func (r *reader) entryID(id []byte) ts.ID {
 // NB(xichen): Validate should be called after all data are read because the
 // digest is calculated for the entire data file.
 func (r *reader) Validate() error {
-	return r.dataFdWithDigest.Validate(r.expectedDataDigest)
+	err := r.indexReader.readerWithDigest.Validate(r.expectedIndexDigest)
+	if err != nil {
+		return fmt.Errorf("could not validate index file: %v", err)
+	}
+	// TODO(r): Make sure to validate checksum when using seeker now that
```
prateek: Is the TODO still applicable?

digest/reader.go
prateek: +1, could FdWithDigestReader use ReaderWithDigest under the hood?

persist/encoding/decoder.go
```diff
@@ -20,19 +20,28 @@
...
+import (
+	"bytes"
+	"fmt"
+	"io"
+
+	"github.com/m3db/m3db/persist/schema"
+)
 
 // Decoder decodes persisted data
 type Decoder interface {
```
prateek: Would you mind breaking this into smaller interfaces just to separate the methods a bit. Something like:

```
type FilesetDecoder interface {

}

type CommitlogDecoder interface {
}

type Decoder {
  FilesetDecoder
  CommitlogDecoder
}
```

persist/encoding/decoder.go
```diff
@@ -42,3 +51,110 @@ type Decoder interface {
...
+
+type decoderStream struct {
+	reader       *bytes.Reader
+	bytes        []byte
+	lastReadByte int
+	unreadByte   int
+}
+
+// NewDecoderStream creates a new decoder stream from a bytes ref.
+func NewDecoderStream(b []byte) DecoderStream {
```
prateek: Could you add a couple of simple tests for this (ones that don't use the `msgpack/` package)

glide.yaml
robskillington: One is faster for writing, and the other is faster for reading concurrently hehe (stack based approach needs no hash object pooling and hence no read/write locks, but is slightly slower in constant time - when we write out the bloom filter its sequential so no need to for concurrent hashing).

persist/fs/mmap.go
```diff
@@ -0,0 +1,88 @@
...
+	// options specifies options to use when mmaping a file
+	options mmapOptions
+}
+
+type mmapOptions struct {
+	// read is whether to make mmap bytes ref readable
+	read bool
+	// write is whether to make mmap bytes ref writable
+	write bool
```
prateek: Might be better to not support write at all considering it's not something we're going to use in production. E.g. concern: we'd have to be careful about `msync` calls to ensure memory writes are synced to the filesystem. Don't need to spend the time figuring this out considering we're not using it.

persist/fs/read.go
```diff
@@ -63,43 +66,159 @@ type reader struct {
...
+type decoderStream struct {
+	bytesReader      *bytes.Reader
+	readerWithDigest digest.ReaderWithDigest
+	backingBytes     []byte
+	buf              [8]byte
+	lastReadByte     int
+	unreadByte       int
+}
+
+func newDecoderStream() *decoderStream {
```
prateek: How come you're not reusing `encoding.NewDecoderStream` here?

persist/fs/read.go
```diff
@@ -63,43 +66,159 @@ type reader struct {
...
-	entries     int
-	entriesRead int
-	prologue    []byte
-	decoder     encoding.Decoder
-	digestBuf   digest.Buffer
-	bytesPool   pool.CheckedBytesPool
+
+	indexFd     *os.File
+	indexMmap   []byte
+	indexReader *decoderStream
```
prateek: nit: maybe better to call it `indexDecoderStream` ?

persist/fs/read.go
```diff
@@ -188,19 +310,27 @@ func (r *reader) readDigest() error {
 	if r.expectedIndexDigest, err = r.digestFdWithDigestContents.ReadDigest(); err != nil {
 		return err
 	}
+	if _, err := r.digestFdWithDigestContents.ReadDigest(); err != nil {
+		// Skip the summaries digest
```
prateek: How come?

glide.yaml
prateek: Sigh, lol.

.travis.yml
robskillington: Yup we'd like to build PRs and just the master branch.  Not when you push, that way during PRs it'll reduce the amount of times we test the head of the PR branch from two to just once.

persist/fs/write.go
robskillington: This one is optimized, in future we can always look at this - agreed.

persist/fs/write.go
robskillington: Ha thanks for that, removed the URL from your comment though to avoid leakage of non-public assets.

persist/fs/write.go
robskillington: Done.

digest/reader.go
robskillington: Sounds good, done.

persist/encoding/decoder.go
robskillington: Following the standard as used by the stdlib, it's nice because you can tell if -1 is set as the value or not without converting byte into a signed int8 just to check the value (which unless you're actually using UnreadByte often is 99% of the case it'll be -1).

robskillington: `bytes.Reader` does not offer that API, only `bufio.Reader` does but we can't use it as we need to know the position of where into the buffer we are ourselves explicitly.

persist/fs/read.go
robskillington: Sure thing, sounds good.

storage/options.go
robskillington: Sounds good, will do.

persist/schema/types.go
robskillington: To match the `Entries int64` it does not have the num prefix.

.metalinter.json
robskillington: Sure thing.

persist/fs/read.go
robskillington: Sure thing.

persist/fs/read.go
robskillington: It was and I have implemented that now, thanks for the catch =]

persist/encoding/decoder.go
robskillington: Sure thing.

persist/encoding/decoder.go
robskillington: Good call, will do.

persist/fs/mmap.go
robskillington: Fair, sure I'll remove this.

persist/fs/read.go
robskillington: Need special logic to call the reader with digest.  The encoding `NewDecoderStream(...)` is for more simple use cases.

persist/fs/read.go
robskillington: Sure thing.

persist/fs/read.go
robskillington: We're not using these files yet, that will come in a later PR to keep this one a reasonable size.

.travis.yml
prateek: Cool, sounds good.

digest/reader.go
```diff
@@ -57,79 +56,66 @@ type FdWithDigestReader interface {
...
 }
 
 func (r *fdWithDigestReader) ReadAllAndValidate(b []byte, expectedDigest uint32) (int, error) {
-	n, err := r.readBytes(b)
+	n, err := r.Read(b)
 	if err != nil {
 		return n, err
 	}
-	_, err = r.reader.ReadByte()
+	_, err = r.Read(r.single[:])
```
prateek: Maybe add a quick comment about why this is necessary.

persist/encoding/stream.go
```diff
@@ -0,0 +1,148 @@
...
+	Skip(length int64) error
+
+	// Remaining returns the remaining bytes in the stream.
+	Remaining() int64
+}
+
+type decoderStream struct {
+	reader       *bytes.Reader
+	bytes        []byte
+	lastReadByte int
```
prateek: can this just be a byte?

persist/encoding/stream.go
```diff
@@ -0,0 +1,148 @@
...
+	s.lastReadByte = -1
+	s.unreadByte = -1
+}
+
+func (s *decoderStream) Read(p []byte) (int, error) {
+	ref := p
+
+	var numUnreadByte int
+	if s.unreadByte >= 0 {
+		p[0] = byte(s.unreadByte)
```
prateek: do you need to guarantee that `len(p)` > 0 here?

persist/encoding/stream.go
prateek: Ah I see why it can't, you're using the negative portions to denote un-initialised. I'll delete the comment.

persist/encoding/stream.go
```diff
@@ -0,0 +1,148 @@
...
+	// Reset resets the decoder stream for decoding a new byte slice.
+	Reset(b []byte)
+
+	// ReadByte reads the next byte.
+	ReadByte() (byte, error)
+
+	// UnreadByte unreads the last read byte or returns error if none read
+	// yet. Only a single byte can be unread at a time, a consecutive call
+	// to UnreadByte will result in an error.
+	UnreadByte() error
```
prateek: One thing that's weird about this is the implementation of calling `UnreadByte()` captures the unread byte in the decoderStream. So calls to `Bytes()` right after a successful `UnreadByte()` return a byte slice without the "unread byte". Is that kosher?

persist/fs/commitlog/info.go
```diff
@@ -49,7 +50,7 @@ func ReadLogInfo(filePath string) (start time.Time, duration time.Duration, inde
 		return time.Time{}, 0, 0, err
 	}
 	logDecoder := msgpack.NewDecoder(nil)
-	logDecoder.Reset(bytes)
+	logDecoder.Reset(encoding.NewDecoderStream(bytes))
```
prateek: Should we pool Decoder/Reader/DecoderStream here?

persist/fs/commitlog/reader.go
```diff
@@ -501,7 +509,7 @@ func (r *reader) readInfo() (schema.LogInfo, error) {
 		return emptyLogInfo, err
 	}
 	data.IncRef()
-	r.logDecoder.Reset(data.Get())
+	r.logDecoder.Reset(encoding.NewDecoderStream(data.Get()))
```
prateek: same here, should we pool `DecoderStream` somehow?

persist/fs/mmap_linux.go
```diff
@@ -18,17 +18,50 @@
...
+		prot = prot | syscall.PROT_READ
+	}
+
+	flags := syscall.MAP_SHARED
+	b, err := syscall.Mmap(fd, int64(offset), length, prot, flags)
+	if err != nil {
+		return nil, err
+	}
+
+	if length < hugePageSize {
```
prateek: Not sure I follow, why not use a huge page for 1.5MB index file?

persist/fs/mmap.go
```diff
@@ -0,0 +1,86 @@
...
+
+	if firstErr == nil {
+		return nil
+	}
+
+	// If we have encountered an error when opening the files,
+	// close the ones that have been opened.
+	for _, desc := range files {
+		if *desc.file != nil {
+			(*desc.file).Close()
```
prateek: do you also need to `munmap` here?

persist/fs/read.go
prateek: Cool makes sense.

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -138,7 +129,11 @@ func (s *fileSystemSource) enqueueReaders(
 
 		readers := make([]fs.FileSetReader, 0, len(files))
 		for i := 0; i < len(files); i++ {
-			r := readerPool.get()
+			r, err := readerPool.get()
+			if err != nil {
+				s.log.Errorf("unable to get reader from pool")
```
prateek: should this be bubbled up so we know that we've bootstrapped from partial data.

persist/fs/write.go
prateek: lol cheers

persist/schema/types.go
prateek: sounds good

persist/fs/read.go
```diff
@@ -108,28 +233,42 @@ func (r *reader) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
...
 	}
 
 	r.infoFdWithDigest.Reset(infoFd)
-	r.indexFdWithDigest.Reset(indexFd)
-	r.dataFdWithDigest.Reset(dataFd)
 	r.digestFdWithDigestContents.Reset(digestFd)
 
 	defer func() {
 		// NB(r): We don't need to keep these FDs open as we use these up front
 		r.infoFdWithDigest.Close()
```
prateek: Would it be better to not retain these fds in the struct?

persist/fs/read.go
prateek: afaict, most of the code except `Skip()` is the same. Any chance of re-using those bits?

persist/fs/read.go
```diff
@@ -63,43 +67,164 @@ type reader struct {
...
+	metadataRead           int
+	prologue               []byte
+	decoder                encoding.Decoder
+	digestBuf              digest.Buffer
+	bytesPool              pool.CheckedBytesPool
+
+	indexEntriesByOffsetAsc []schema.IndexEntry
+}
+
+type indexEntriesByOffsetAsc []schema.IndexEntry
```
prateek: nit: move `indexEntriesByOffsetAsc` and `decoderStream` to the bottom of the file

persist/fs/read.go
```diff
@@ -209,28 +345,34 @@ func (r *reader) readInfo(size int) error {
...
-func (r *reader) readIndex(size int) error {
-	// NB(r): need to entirely consume index file so we don't stripe
-	// between index and data contents as we read the data contents.
-	r.prepareUnreadBuf(size)
-	n, err := r.indexFdWithDigest.ReadAllAndValidate(r.unreadBuf[:size], r.expectedIndexDigest)
-	if err != nil {
-		return err
+func (r *reader) readIndex() error {
+	r.decoder.Reset(r.indexDecoderStream)
+	for i := 0; i < r.entries; i++ {
```
prateek: Looks like we're still keeping all of the index file in memory, is the plan to do this lazily in subsequent PRs?

digest/reader.go
robskillington: Sure thing.

persist/encoding/stream.go
robskillington: Yes, I'll add what I was going to originally do with a `if len(p) == 0 { return 0, nil }` at the beginning of the method (but was upset that I'd have to add a test for branch coverage ha).

persist/encoding/stream.go
robskillington: `Bytes()` actually always gives you the entire underlying byte slice without any offsets.  `Remaining()` which is used to calculate position into the current byte slice however does not take into account whether you have unread a byte or not, so that is a good observation.  I'll fix this.

persist/fs/commitlog/info.go
robskillington: This is only O(files) of commit log files, not O(entries).  The amount of files should be less than a hundred (worst-ish case) so this is fine, `ReadLogInfo(...)` is not in the hot path.

persist/fs/commitlog/reader.go
robskillington: I can reuse the decoderstream for the reader, sure thing.

persist/fs/mmap_linux.go
robskillington: Well we definitely don't want to use huge pages for anything less than 4096 or even "a few" 4096 files in length, otherwise you do cause memory fragmentation (it's a problem with huge pages).  Maybe I'll use > 32kb which is 8 pages.

persist/fs/mmap.go
robskillington: Good call.

storage/bootstrap/bootstrapper/fs/source.go
robskillington: Sure thing, restructured this to bubble it up.

persist/fs/read.go
robskillington: We don't actually retain the fds in the struct, we just retain the fd with digest wrappers so that we don't need to create them every time we open the file again (i.e. pooling the fd with digest wrappers).

persist/fs/read.go
robskillington: Sure thing.

persist/fs/read.go
robskillington: Correct, however we'll implement this in the seeker and keep the reader as is.  We'll just never use the reader outside needing to do batch jobs.

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -24,52 +24,41 @@ import (
...
 )
 
 type newFileSetReaderFn func(
-	filePathPrefix string,
-	dataReaderBufferSize int,
-	infoReaderBufferSize int,
 	bytesPool pool.CheckedBytesPool,
-	decodingOpts msgpack.DecodingOptions,
-) fs.FileSetReader
+	opts fs.Options,
```
prateek: massive +1 for this change

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -138,7 +129,12 @@ func (s *fileSystemSource) enqueueReaders(
 
 		readers := make([]fs.FileSetReader, 0, len(files))
 		for i := 0; i < len(files); i++ {
-			r := readerPool.get()
+			r, err := readerPool.get()
+			if err != nil {
+				s.log.Errorf("unable to get reader from pool")
+				readersCh <- shardReaders{err: err}
```
prateek: do you need a `readerPool.put(r)` after this?

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -400,14 +413,9 @@ func (s *fileSystemSource) Read(
...
-		return s.newReaderFn(
-			s.filePathPrefix,
-			s.dataReaderBufferSize,
-			s.infoReaderBufferSize,
-			s.opts.ResultOptions().DatabaseBlockOptions().BytesPool(),
-			s.decodingOpts,
-		)
+	bytesPool := s.opts.ResultOptions().DatabaseBlockOptions().BytesPool()
+	readerPool := newReaderPool(func() (fs.FileSetReader, error) {
+		return s.newReaderFn(bytesPool, s.fsopts)
```
prateek: Any reason to have this here instead of `Options`?

persist/fs/commitlog/reader.go
```diff
@@ -270,7 +272,9 @@ func (r *reader) shutdown() {
 func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan<- readResponse) {
 	decodingOpts := r.opts.FilesystemOptions().DecodingOptions()
```
prateek: nit: could you put this in a var block

persist/fs/read_test.go
```diff
@@ -407,3 +421,143 @@ func TestReadValidate(t *testing.T) {
...
+		if b1 != b2 {
+			t.Fatalf("incorrect byte after unread: got %q, want %q", b1, b2)
+		}
+	}
+	if got != want {
+		t.Errorf("got %q, want %q", got, want)
+	}
+}
+
+func TestDecoderStreamUnreadByteMultiple(t *testing.T) {
```
prateek: +1 for this keeping this test. 

Feels like you're having to test a lot of weird combination of conditions, how would you feel about adding a few property tests for the decoder?

persist/fs/read_decode_stream.go
```diff
@@ -0,0 +1,141 @@
...
+	"fmt"
+	"io"
+
+	"github.com/m3db/m3db/digest"
+	"github.com/m3db/m3db/persist/encoding"
+)
+
+var (
+	// Ensure readerDecoderStream implements encoding.DecoderStream
+	_ encoding.DecoderStream = &readerDecoderStream{}
```
prateek: could you avoid this by changing the ctor to the following signature: `func newReaderDecoderStream() encoding.DecoderStream {...}`

persist/fs/read.go
prateek: ah makes sense. would you mind putting a note down in the struct about the lifecycle of the wrappers

persist/fs/mmap.go
```diff
@@ -0,0 +1,102 @@
...
+			break
+		}
+
+		b, err := mmapFile(fd, desc.options)
+		if err != nil {
+			multiErr = multiErr.Add(err)
+			break
+		}
+
+		*desc.file = fd
```
prateek: pretty soon we'll be writing `desc->file = fd` :rofl:

storage/bootstrap/bootstrapper/fs/source.go
robskillington: Ah no because reader is nil here (the err was from trying to get it from the pool hehe).

storage/bootstrap/bootstrapper/fs/source.go
robskillington: Yeah you don't really want to allocate and keep around readers outside of the bootstrapping process, hence why its created on demand each time.

persist/fs/read_test.go
robskillington: I'll open an issue for it, in the interest of expediting this change we'll look at this later.

robskillington: Opened #403

persist/fs/read_decode_stream.go
robskillington: Sounds good, done

persist/fs/mmap.go
robskillington: Ha, aye

PR #396, opened 2017-11-11
Update M3DB client to support new peer bootstrapping endpoint

richardartoul:


client/session.go
```diff
@@ -1224,79 +1230,99 @@ func (s *session) FetchBlocksMetadataFromPeers(
...
+		errCh <- s.streamBlocksMetadataFromPeers(
+			namespace, shard, peers, start, end, metadataCh, m, false)
 		close(metadataCh)
 		close(errCh)
 	}()
 
 	return newMetadataIter(metadataCh, errCh), nil
 }
 
+// FetchBootstrapBlocksFromPeers will fetch the specified blocks from peers for
```
richardartoul: TODO: I'm gonna add a Markdown documentation doc

client/session.go
```diff
@@ -1548,29 +1590,164 @@ func (s *session) streamBlocksMetadataFromPeer(
...
+			pageToken = pageToken[:nextPageTokenLen]
+		} else {
+			// No further results
+			moreResults = false
+		}
+
+		for _, elem := range result.Elements {
+			// Error occurred retrieving block metadata, use default values
+			blockStart := time.Unix(0, elem.Start)
+			if elem.Err != nil {
```
richardartoul: This error handling doesn't seem very good, but this is how the existing code handles it

richardartoul: I guess it kind of works because the block is so far back in the past (since elem.Start is 0?) that it just ignores the this one entirely? Seems like we should emit a stat or something

Makefile
```diff
@@ -20,6 +20,7 @@ metalint_exclude     := .excludemetalint
 mockgen_package      := github.com/golang/mock/mockgen
 mocks_output_dir     := generated/mocks/mocks
 mocks_rules_dir      := generated/mocks
+proto_output_dir     := generated/proto
```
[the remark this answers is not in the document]
richardartoul: I think the auto-gen script in ci-scripts requires it, but it wasn't defined for M3DB yet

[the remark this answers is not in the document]
richardartoul: its used in the proto-gen command

client/options.go
```diff
@@ -130,6 +130,9 @@ const (
 	// defaultFetchSeriesBlocksMetadataBatchTimeout is the default series blocks metadata fetch timeout
 	defaultFetchSeriesBlocksMetadataBatchTimeout = 60 * time.Second
 
+	// defaultFetchSeriesBlocksMetadataBatchBackoff is the default series blocks metadata fetch retry backoff
+	defaultFetchSeriesBlocksMetadataBatchBackoff = 1 * time.Second
```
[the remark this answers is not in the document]
richardartoul: good call

client/session.go
```diff
@@ -1224,79 +1230,105 @@ func (s *session) FetchBlocksMetadataFromPeers(
...
+	streamFunc := s.streamAndGroupCollectedBlocksMetadata
+	if isV2 {
+		streamFunc = s.streamAndGroupCollectedBlocksMetadataV2
+	}
+	s.streamBlocksFromPeers(nsMetadata, shard, peers,
+		metadataCh, opts, result, progress, streamFunc)
 
-	if err := waitDone(); err != nil {
+	// Check if an error occurred during the metadata streaming
+	if streamBlocksMetadataErr != nil {
```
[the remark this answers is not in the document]
richardartoul: You're right I think this is wrong. In practice it probably always works fine because its extremely unlikely that streamBlocksFromPeers would complete inbetween the call to close(metadataCh) and settings streamBlocksMetadataErr, but I'll fix it to use an error channel.

Good catch

client/session.go
```diff
@@ -1548,29 +1590,164 @@ func (s *session) streamBlocksMetadataFromPeer(
...
+	namespace ts.ID,
+	shard uint32,
+	peer peer,
+	start, end time.Time,
+	metadataCh chan<- blocksMetadata,
+	progress *streamFromPeersMetrics,
+) error {
+	var (
+		pageToken []byte
+		retrier   = xretry.NewRetrier(xretry.NewOptions().
```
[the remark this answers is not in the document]
richardartoul: Ok I took a quick glance through the retrier code and it looks thread safe so this should be fine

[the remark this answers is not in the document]
richardartoul: actually there is already a fetch retrier in the struct, I'll just reuse that

client/session.go
```diff
@@ -1224,79 +1230,105 @@ func (s *session) FetchBlocksMetadataFromPeers(
...
+// In terms of error handling, if an error occurs during the metadata streaming
+// portion, then this function will return an error. However, if something goes
+// wrong during the data streaming portion, it will not return an error, and
+// the function will just return as much data as it can.
 func (s *session) FetchBootstrapBlocksFromPeers(
 	nsMetadata namespace.Metadata,
 	shard uint32,
 	start, end time.Time,
 	opts result.Options,
+	isV2 bool,
```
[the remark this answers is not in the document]
richardartoul: sure

client/session.go
```diff
@@ -1224,79 +1230,105 @@ func (s *session) FetchBlocksMetadataFromPeers(
...
+				return
+			default:
+				progress.fetchBlocksFromPeers.Update(1)
+				time.Sleep(gaugeReportInterval)
+			}
 		}
-		m.fetchBlocksFromPeers.Update(0)
+	}()
+	defer func() {
+		close(doneCh)
```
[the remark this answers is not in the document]
richardartoul: good call

client/session.go
```diff
@@ -1224,79 +1230,105 @@ func (s *session) FetchBlocksMetadataFromPeers(
...
-		err := s.streamBlocksFromPeers(nsMetadata, shard, peers,
-			metadataCh, opts, result, m, s.streamAndGroupCollectedBlocksMetadata)
-		onDone(err)
-	}()
+	// Begin consuming metadata and making requests. This will block until all
+	// data has been streamed (or failed to stream). Note that this function does
+	// not return an error and if anything goes wrong here we won't report it to
+	// the caller, but metrics and logs are emitted internally. Also note that the
+	// streamAndGroupCollectedBlocksMetadata function is injected.
+	streamFunc := s.streamAndGroupCollectedBlocksMetadata
```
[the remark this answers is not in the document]
richardartoul: done

client/session.go
```diff
@@ -1385,8 +1417,9 @@ func (s *session) streamBlocksMetadataFromPeers(
 	shard uint32,
 	peers []peer,
 	start, end time.Time,
-	ch chan<- blocksMetadata,
-	m *streamFromPeersMetrics,
+	metadataCh chan<- blocksMetadata,
+	progress *streamFromPeersMetrics,
+	isV2 bool,
```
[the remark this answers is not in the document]
richardartoul: done

client/session.go
```diff
@@ -1437,13 +1478,14 @@ func (s *session) streamBlocksMetadataFromPeer(
 		retrier   = xretry.NewRetrier(xretry.NewOptions().
```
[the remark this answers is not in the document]
richardartoul: See other comment re: retrier

client/session.go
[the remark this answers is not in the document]
richardartoul: Ok so that retrier had some defaults that were different than what we hard-coded before, so to be safe I created a streamBlocksRetrier in the options which has the same defaults as what used to be hard-coded

client/session.go
```diff
@@ -1224,79 +1244,104 @@ func (s *session) FetchBlocksMetadataFromPeers(
...
 		errCh      = make(chan error, 1)
-		m          = s.streamFromPeersMetricsForShard(shard, resultTypeMetadata)
+		m          = s.newPeerMetadataStreamingProgressMetrics(shard, resultTypeMetadata)
 	)
 
 	go func() {
-		errCh <- s.streamBlocksMetadataFromPeers(namespace, shard, peers,
-			start, end, metadataCh, m)
+		errCh <- s.streamBlocksMetadataFromPeers(
+			namespace, shard, peers, start, end, metadataCh, m, FetchBlocksMetadataV1)
```
richardartoul: Is it cool to leave this on the old endpoint only? Presumably if it works for the bootstrap it will also work for this

client/session.go
```diff
@@ -2672,8 +2873,10 @@ func (qs peerBlocksQueues) closeAll() {
 }
 
 type blocksMetadata struct {
-	peer   peer
-	id     ts.ID
+	peer peer
+	id   ts.ID
+	// TODO:
```
prateek: suggestion(take/leave): change this to `TODO(user): message`, e.g. `TODO(rartoul): Make this not a slice once we delete the V1 code path`. It's a lot easier to parse the output of `grep -i TODO(rartoul)` if you follow that convention

richardartoul: yeah totally fair, I thinking I was just following the convention I saw in some of the file I was working on. I'll update this

richardartoul: done

Makefile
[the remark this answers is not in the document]
robskillington: Ah right, looks good then.

client/session.go
```diff
@@ -1224,79 +1246,88 @@ func (s *session) FetchBlocksMetadataFromPeers(
...
 
-	// Begin consuming metadata and making requests
-	go func() {
-		err := s.streamBlocksFromPeers(nsMetadata, shard, peers,
-			metadataCh, opts, result, m, s.streamAndGroupCollectedBlocksMetadata)
-		onDone(err)
-	}()
+	// Begin consuming metadata and making requests. This will block until all
+	// data has been streamed (or failed to stream). Note that this function does
+	// not return an error and if anything goes wrong here we won't report it to
```
robskillington: Seems strange that error is not bubbled up here, not sure what's changed that we don't want to bubble up the error anymore - we used to have the behavior described here but allowed some very heavy errors to continue to bubble up.

client/session.go
```diff
@@ -1370,10 +1401,10 @@ func (s *session) FetchBlocksFromPeers(
 
 	// Begin consuming metadata and making requests
 	go func() {
-		err := s.streamBlocksFromPeers(nsMetadata, shard, peers,
-			metadataCh, opts, result, m, s.passThruBlocksMetadata)
+		s.streamBlocksFromPeers(nsMetadata, shard, peers,
+			metadataCh, opts, result, progress, s.passThruBlocksMetadata)
 		close(outputCh)
-		onDone(err)
+		onDone(nil)
```
robskillington: Hm why don't we return error here too?

client/session.go
richardartoul: If you look at the source code for streamBlocksFromPeers on master, the function signature returns an error, but there is only a single call to return in that function in which it explicitly returns nil, so this was a minor refactoring where I updated the signature of that function since it was not possible for it to ever return an error

client/session.go
richardartoul: same comment as above

robskillington: Right, yeah so perhaps we should change the `onDone(...)` to just be `func()` instead of `func(error)` because it seems like this is the only caller of `onDone` in this method?

client/session.go
```diff
@@ -1224,79 +1246,88 @@ func (s *session) FetchBlocksMetadataFromPeers(
...
+// FetchBootstrapBlocksFromPeers will fetch the specified blocks from peers for
+// bootstrapping purposes. Refer to peer_bootstrapping.md for more details.
 func (s *session) FetchBootstrapBlocksFromPeers(
 	nsMetadata namespace.Metadata,
 	shard uint32,
 	start, end time.Time,
 	opts result.Options,
+	version FetchBlocksMetadataEndpointVersion,
 ) (result.ShardResult, error) {
+	if version != FetchBlocksMetadataEndpointV1 && version != FetchBlocksMetadataEndpointV2 {
```
robskillington: Maybe we should move this closer to the definition of:

robskillington: In fact, maybe we don't need to do this here.

Below you can do

client/session.go
robskillington: I see, ok yeah makes sense.  If we want to add the possibility to float up an error later from that method we can just change this back.

client/session.go
```diff
@@ -1397,22 +1429,28 @@ func (s *session) streamBlocksMetadataFromPeers(
...
-			err := s.streamBlocksMetadataFromPeer(namespace, shard, peer,
-				start, end, ch, m)
+			var err error
+			if version == FetchBlocksMetadataEndpointV2 {
+				err = s.streamBlocksMetadataFromPeerV2(
+					namespace, shard, peer, start, end, metadataCh, progress)
+			} else {
+				err = s.streamBlocksMetadataFromPeer(
+					namespace, shard, peer, start, end, metadataCh, progress)
+			}
```
robskillington: Can we do a switch here too to verify that a valid support one was set?  i.e. switch V1, V2 and default to setting an error.

client/session.go
```diff
@@ -1541,7 +1577,137 @@ func (s *session) streamBlocksMetadataFromPeer(
...
+			if delta > 0 {
+				for i := 0; i < delta; i++ {
+					pageToken = append(pageToken, '\x00')
+				}
+			}
+			// Copy the pageToken so we don't have to keep the result around
+			copy(pageToken, result.NextPageToken)
+			// Make sure we don't include extraneous bytes if the new page token is
+			// shorter than the previous one (for whateve reason)
+			pageToken = pageToken[:nextPageTokenLen]
```
robskillington: Hey instead of all this code, better to just do:

client/session.go
```diff
@@ -1693,6 +1852,47 @@ func (s *session) streamAndGroupCollectedBlocksMetadata(
 	}
 }
 
+func (s *session) streamAndGroupCollectedBlocksMetadataV2(
+	peersLen int,
+	metadataCh <-chan blocksMetadata,
+	enqueueCh *enqueueChannel,
+) {
+	metadata := make(map[hashAndBlockStart]*receivedBlocks)
```
robskillington: Maybe we should make this `map[hashAndBlockStart]receivedBlocks` to avoid heap alloc per entry and just update it in place?

client/session.go
richardartoul: yeah exactly, better to have a useful function signature in the meantime I think

client/session.go
```diff
@@ -1693,6 +1852,47 @@ func (s *session) streamAndGroupCollectedBlocksMetadata(
...
+		if !ok {
+			break
+		}
+
+		key := hashAndBlockStart{
+			hash:       m.id.Hash(),
+			blockStart: m.blocks[0].start.UnixNano(),
+		}
+		received, ok := metadata[key]
+		if !ok || received.submitted {
```
robskillington: If it's already  been submitted we probably don't want to add another entry?  Otherwise we'll submit it twice.

doc/peer_bootstrapping.md
```diff
@@ -0,0 +1,94 @@
...
+in, we begin determining which peer is the best source to stream the rest of
+of the data from, and then we will begin streaming data from that peer all
+the while we continue to receive metadata.
+
+In terms of error handling, if an error occurs during the metadata streaming
+portion, then this function will return an error. However, if something goes
+wrong during the data streaming portion, it will not return an error, and
+the function will just return as much data as it can.
+
+The diagram below depicts the control flow and concurrency (goroutines and channels) in detail:
```
robskillington: Brilliant, great illustration 👍

client/session.go
richardartoul: I can do this, the only problem is that its passed into the peerBlocksIter which has this signature:

```golang
func newPeerBlocksIter(
	inputC <-chan peerBlocksDatapoint,
	errC <-chan error,
) *peerBlocksIter {
	return &peerBlocksIter{
		inputCh: inputC,
		errCh:   errC,
	}
}

func (it *peerBlocksIter) Current() (topology.Host, ts.ID, block.DatabaseBlock) {
	return it.current.peer, it.current.id, it.current.block
}

func (it *peerBlocksIter) Err() error {
	return it.err
}

func (it *peerBlocksIter) Next() bool {
	if it.done || it.err != nil {
		return false
	}
	m, more := <-it.inputCh

	if !more {
		it.err = <-it.errCh
		it.done = true
		return false
	}

	it.current = m
	return true
}
```

I can re-write the iter to never return an err / not accept an errCh but its a bigger change.

What do you think?

client/session.go
richardartoul: I think waiting until we reach the streamFn portion of the code is too late because the streamBlocksMetadataFromPeers goroutine is not pre-emptible and will just run until its done.

How about I replace the lines:

```golang
if version != FetchBlocksMetadataEndpointV1 && version != FetchBlocksMetadataEndpointV2 {
		return nil, errInvalidFetchBlocksMetadataVersion
	}
```

with the new IsValidFetchBlocksMetadataEndpoint to remove the code duplication? I can also add the switch statement as specified as an extra sanity check

client/session.go
richardartoul: Sure added the switch / default case. I can't really return the err since its an anonymous goroutine, but I can handle it the way we handle the other errs. Should be sufficient considering this should NeverHappenAnywaysTM

client/session.go
richardartoul: that...is much better haha thanks

client/session.go
richardartoul: I thought about changing this, but the existing code uses a pointer. I think this codepath is hot enough that I'm concerned about changing it while we're also trying to measure lots of other perf-related stuff.

richardartoul: Maybe a TODO or task to investigate would be sufficient?

client/session.go
richardartoul: This is similarly following the precedent set by the existing code. Is re-submitting it bad? I may have an incomplete understanding here but I thought resubmitting would be an idempotent operation

client/session.go
```diff
@@ -1224,79 +1246,88 @@ func (s *session) FetchBlocksMetadataFromPeers(
...
 		errCh      = make(chan error, 1)
-		m          = s.streamFromPeersMetricsForShard(shard, resultTypeMetadata)
+		m          = s.newPeerMetadataStreamingProgressMetrics(shard, resultTypeMetadata)
 	)
 
 	go func() {
-		errCh <- s.streamBlocksMetadataFromPeers(namespace, shard, peers,
-			start, end, metadataCh, m)
+		errCh <- s.streamBlocksMetadataFromPeers(
+			namespace, shard, peers, start, end, metadataCh, m, FetchBlocksMetadataEndpointV1)
```
richardartoul: @robskillington Its ok to just leave this as V1 only right?

client/session.go
richardartoul: Spoke in person. We settled on leaving the onDone function signature as is because we want to leave open the possibility of needed to return an error from the peerBlocksIter interface in the future

client/session.go
richardartoul: spoke in person, decided to leave this in for now to not pollute our perf measurements

client/session.go
robskillington: Yup sounds good, we can update this code path when we deprecate V1.

PR #399, opened 2017-11-13
Upgrade maptime --> badtime

richardartoul:


robskillington:
Looks like the big unit test is failing, I removed it in my PR.  Perhaps either way until mine is landed or I can merge that change into master separately and you can rebase perhaps?

richardartoul:
@robskillington Sounds good to me, although this is an existing test right? why did it suddenly start failing

richardartoul:
@robskillington Also I've already moved this into m3x and m3cluster :)

services/m3dbnode/server/server_test.go
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-// +build big
+// +build adhoc
```
prateek: @richardartoul do we need to include these tags in the metalinter config too?

richardartoul: I copied this out of Rob's pending PR and I think he wants to intentionally skip this test right now because its fails in Travis

richardartoul: Oh I see what you're saying, we could include it, but I think this is a temporary hack we don't want to keep.

@robskillington

prateek: Cool, I'll put a note in the other PR then.

PR #400, opened 2017-11-20
Support configurable seed for murmur32

richardartoul:


glide.lock
```diff
@@ -237,7 +237,7 @@ imports:
 - name: github.com/shirou/w32
   version: bb4de0191aa41b5507caa14b0650cdbddcd9280b
 - name: github.com/spaolacci/murmur3
-  version: 0d12bf811670bf6a1a63828dfbd003eded177fce
+  version: 9f5d223c60793748f04a9d5b4b4eacddfc1f755d
```
richardartoul: Had to update dependency to get the new seedable function

sharding/shardset.go
```diff
@@ -148,7 +148,19 @@ func validateShards(shards []shard.Shard) error {
 
 // DefaultHashGen generates a HashFn based on murmur32
 func DefaultHashGen(length int) HashFn {
```
robskillington: Maybe we should rename this to `DefaultHashFn(length int) HashFn` as it returns a `HashFn` not a `HashGen`.

robskillington:
We also need to add this field to the `client/Configuration` struct too so that users using it to create clients have the ability to specify the seed.

sharding/shardset.go
richardartoul: sure thing

client/config.go
```diff
@@ -76,6 +77,9 @@ type Configuration struct {
 	// BackgroundHealthCheckFailThrottleFactor is the factor of the host connect
 	// time to use when sleeping between a failed health check and the next check.
 	BackgroundHealthCheckFailThrottleFactor float64 `yaml:"backgroundHealthCheckFailThrottleFactor" validate:"min=0,max=10"`
+
+	// The configuration for hashing
+	HashingConfiguration HashingConfiguration `yaml:"hashing"`
```
robskillington: For consistency perhaps start the comment with the symbol name?  i.e. `// HashingConfiguration is the configuration for hashing of IDs to vnodes`

client/config.go
```diff
@@ -193,3 +198,9 @@ func (c Configuration) NewAdminClient(
 
 	return NewAdminClient(opts)
 }
+
+// HashingConfiguration is the configuration for hashing
+type HashingConfiguration struct {
```
robskillington: I would place the struct below `Configuration` and above the parsing methods, just for readability of the file.

richardartoul:
@robskillington How do you feel about adding a check in m3dbnode/server.go to make sure that the hash seed for the client configuration and the hash seed for the node configuration are the same? The YAML configures them independently, but it seems like you would always want them to be the same right?

robskillington:
@richardartoul so the server actually uses whatever topology descriptions are present in the top level config and ignores the client level one so that what you described isn't possible:

```go
// services/m3dbnode/server.go

// topo options are parsed from top level config
	topoOpts := topology.NewDynamicOptions().
		SetConfigServiceClient(configSvcClient).
		SetServiceID(serviceID).
		SetQueryOptions(services.NewQueryOptions().SetIncludeUnhealthy(true)).
		SetInstrumentOptions(opts.InstrumentOptions())

	topoInit := topology.NewDynamicInitializer(topoOpts)
	topo, err := topoInit.Init()
	if err != nil {
		logger.Fatalf("could not initialize m3db topology: %v", err)
	}

	hostID, err := cfg.HostID.Resolve()
	if err != nil {
		logger.Fatalf("could not resolve local host ID: %v", err)
	}

	m3dbClient, err := cfg.Client.NewAdminClient(
		client.ConfigurationParameters{
			InstrumentOptions: iopts.
				SetMetricsScope(iopts.MetricsScope().SubScope("m3dbclient")),
			TopologyInitializer: topoInit,
		},
		// ...
	)
```

So you can see that we explicitly pass the `topoInit` to the `cfg.Client.NewAdminClient(...)` method which will correctly ignore any settings they have on the client configuration itself:

```go
// client/config.go

// NewAdminClient creates a new M3DB admin client using
// specified params and custom options.
func (c Configuration) NewAdminClient(
	params ConfigurationParameters,
	custom ...CustomAdminOption,
) (AdminClient, error) {
	// ... other stuff

	topoInit := params.TopologyInitializer
	if topoInit == nil {
		// ... use client level configuration for sharding/discovery settings
	}
```

richardartoul:
@robskillington Oh wow nice, Ok I'm gonna go ahead and merge this then. Thanks!

PR #405, opened 2017-12-10
Stabilize ticks with constant per series sleep and minimum tick interval

robskillington:


integration/options.go
```diff
@@ -54,8 +54,8 @@ const (
 	// defaultWorkerPoolSize is the default number of workers in the worker pool.
 	defaultWorkerPoolSize = 10
 
-	// defaultTickInterval is the default tick interval.
-	defaultTickInterval = 1 * time.Second
+	// defaultTickMinimumInterval is the default tick interval.
```
richardartoul: nit: is the default minimum tick interval

runtime/types.go
```diff
@@ -72,6 +72,32 @@ type Options interface {
 	// limit is primarily offered to defend against unintentional bursts of new
 	// time series being inserted.
 	WriteNewSeriesLimitPerShardPerSecond() int
+
+	// SetTickPerSeriesSleepDuration sets the tick sleep per series value that
+	// provides a constant duration to sleep per series at the end of processing
+	// a batch of series during a background tick, this can directly effect how
```
richardartoul: "batch of series" is a little ambigous. This sleep is after processing a single series isn't it?

runtime/types.go
```diff
@@ -72,6 +72,32 @@ type Options interface {
...
+
+	// TickPerSeriesSleepDuration returns the tick sleep per series value that
+	// provides a constant duration to sleep per series at the end of processing
+	// a batch of series during a background tick, this can directly effect how
+	// fast a block is persisted after is rotated from the mutable series buffer
+	// to a series block (since all series need to be merged/processed before a
+	// persist can occur).
+	TickPerSeriesSleepDuration() time.Duration
+
+	// SetTickMinimumInterval sets the minimum tick interval to run ticks, this
```
richardartoul: Might be nice to explain that the tick duration is unbounded, but this allows you to control the wait inbetween ticks

services/m3dbnode/config/config.go
```diff
@@ -71,6 +71,14 @@ type Configuration struct {
 	// Write new series backoff between batches of new series insertions.
 	WriteNewSeriesBackoffDuration time.Duration `yaml:"writeNewSeriesBackoffDuration"`
 
+	// Tick per series sleep at the completion of a tick batch, only used
```
richardartoul: whats a tick batch, a group of series?

services/m3dbnode/config/config.go
```diff
@@ -71,6 +71,14 @@ type Configuration struct {
...
+	// if the value is greater than zero otherwise the default is used.
+	TickPerSeriesSleepDuration time.Duration `yaml:"tickPerSeriesSleepDuration"`
+
+	// Tick minimum interval controls the minimum tick interval for the node,
```
richardartoul: Since these are "public" config I wonder if its worth making these as explicit as the other comments? Presumably this is what people will look at when configuring their nodes

storage/namespace.go
```diff
@@ -109,10 +120,17 @@ type dbNamespace struct {
 
 	tickWorkers            xsync.WorkerPool
 	tickWorkersConcurrency int
+	statsLastTick          databaseNamespaceStatsLastTick
 
 	metrics databaseNamespaceMetrics
 }
 
+type databaseNamespaceStatsLastTick struct {
+	sync.RWMutex
```
richardartoul: This seems like the sort of thing we'd normally use an atomic for, is contention low here?

storage/namespace.go
```diff
@@ -828,10 +882,16 @@ func (n *dbNamespace) initShards(needBootstrap bool) {
 
 func (n *dbNamespace) Close() error {
 	n.Lock()
+	if n.closed {
```
richardartoul: Did we have a race condition before, or are you just protecting against duplicate calls?

storage/series/buffer.go
```diff
@@ -236,21 +236,28 @@ func (b *dbBuffer) DrainAndReset() drainAndResetResult {
...
+		if err != nil {
 			log := b.opts.InstrumentOptions().Logger()
-			log.Errorf("buffer drain tried to drain empty stream for bucket: %v",
-				start.String())
+			log.Errorf("buffer merge encode error: %v", err)
+		} else {
+			if result.merges > 0 {
+				mergedOutOfOrderBlocks++
+			}
+			if result.block.Len() < 1 {
```
richardartoul: could this be == 0? Are you just protecting against accidental negative values?

I think `if !result.block.Len() > 0` might be clearer, `< 1` is jarring when talking about a length

storage/shard.go
```diff
@@ -51,9 +51,14 @@ import (
 )
 
 const (
-	shardIterateBatchPercent     = 0.01
-	expireBatchLength            = 1024
-	defaultTickSleepIfAheadEvery = 128
+	shardMinimumTickPerSeriesSleeps        = 32
```
richardartoul: a comment on this variable would be good

storage/shard.go
```diff
@@ -51,9 +51,14 @@ import (
...
+	shardMinimumTickPerSeriesSleepDuration = time.Microsecond
```
richardartoul: this one too

storage/shard.go
```diff
@@ -446,6 +457,11 @@ func (s *dbShard) tickAndExpire(
 		s.purgeExpiredSeries(expired)
 	}
 
+	minSleep := shardMinimumTickPerSeriesSleeps * tickSleepPerSeries
```
richardartoul: having a hard time grokking what "shardMinimumTickPerSeriesSleeps" is

richardartoul: Also generally confused about this code, shouldn't you just be sleeping by max(shardMinimumTickPerSeriesSleepDuration, tickSleepPerSeries)? Why the multiplication

storage/tick.go
```diff
@@ -85,9 +93,19 @@ func newTickManager(database database, opts Options) databaseTickManager {
 		c:        context.NewCancellable(),
 		tokenCh:  tokenCh,
 	}
+
+	runtimeOptsMgr := opts.RuntimeOptionsManager()
+	runtimeOptsMgr.RegisterListener(mgr)
+	return mgr
+}
+
+func (mgr *tickManager) SetRuntimeOptions(opts runtime.Options) {
```
richardartoul: Why is this updateable after instantiation? Is there a KV config that I missed

richardartoul: also seems a little weird that this function accepts a runtime.Options and uses hat to update a single field instead of just wholesale replacing the internal object reference (which seems to be of a different type?). Could we just use the same type (and wrap it in a mutex if we need the lock)

storage/tick.go
```diff
@@ -125,36 +143,39 @@ func (mgr *tickManager) Tick(softDeadline time.Duration, forceType forceType) er
...
+		n.Tick(mgr.c)
 	}
-	for i, n := range namespaces {
-		deadline := float64(softDeadline) * (float64(sizes[i]) / float64(totalSize))
-		n.Tick(mgr.c, time.Duration(deadline))
+
+	// NB(r): Always sleep for some constant period since ticking
+	// is variable with num series. With a really small amount of series
+	// the per shard amount of constant sleeping is only:
+	// = num shards * min sleeps per shard (32 constant) * sleep per series (100 microseconds default)
```
richardartoul: what is min sleeps per shard?

storage/tick.go
```diff
@@ -125,36 +143,39 @@ func (mgr *tickManager) Tick(softDeadline time.Duration, forceType forceType) er
...
+	// This number can be quite small if configuring to have small amount of
+	// shards (say 10 shards):
+	// = 10 num shards * 32 min sleeps * 100 microseconds default sleep per series
+	// = 10*32*0.1 milliseconds
+	// = 32ms
+	// Because of this we always sleep at least some fixed constant amount.
```
richardartoul: Whats the benefit of having this minimum sleep in-between ticks versus just "always be ticking"? Presumably if you have a lot of series, you're always gonna be ticking anyways (say it takes 15 minutes to tick and then you wait one minute inbetween, you're ticking the vast majority of the time). Would it be better to have the performance be constant vs spiky?

robskillington: Yeah this is to catch the super small embedded case, e.g. M3DB as a local time series stores on your application host.

storage/shard.go
robskillington: I'm removing this concept, yeah.

integration/options.go
robskillington: Good catch, will updated.

runtime/types.go
robskillington: Its after a group, not at the end of each one, primarily to go easy on the scheduler.  A sleep at the end of each one would be a lot more yielding/continuation than doing it in groups of 4096 (current value).

runtime/types.go
robskillington: Sure thing.

services/m3dbnode/config/config.go
robskillington: Group of series yeah, I can make that more explicit by using group rather than batch.

services/m3dbnode/config/config.go
robskillington: You mean just more descriptive?  These should not need to be overridden TBH, in fact I'm kind of tempted to make them only override-able using config KV values.

storage/namespace.go
robskillington: Yes, the only reader will be a background goroutine and the only writer is at the end of a tick (i.e. not in the hot path).

storage/namespace.go
robskillington: Just protecting against duplicate calls, seemed strange to be returning a potential error here but not actually returning one too.

storage/series/buffer.go
robskillington: Ha, yeah its the negative thing - a habit I picked up a while ago defending against values passed by users.  I'm not sure how Len() is implemented under the hood for a block when I was writing the code I think.

storage/shard.go
robskillington: I've removed this concept thankfully.

storage/shard.go

storage/tick.go
robskillington: I've removed this concept thankfully.

services/m3dbnode/server/server.go
```diff
@@ -174,13 +174,24 @@ func Run(runOpts RunOptions) {
...
-	runtimeOptsMgr := m3dbruntime.NewOptionsManager(m3dbruntime.NewOptions().
+	runtimeOpts := m3dbruntime.NewOptions().
 		SetPersistRateLimitOptions(ratelimit.NewOptions().
 			SetLimitEnabled(true).
 			SetLimitMbps(cfg.Filesystem.ThroughputLimitMbps).
 			SetLimitCheckEvery(cfg.Filesystem.ThroughputCheckEvery)).
 		SetWriteNewSeriesAsync(cfg.WriteNewSeriesAsync).
-		SetWriteNewSeriesBackoffDuration(cfg.WriteNewSeriesBackoffDuration))
+		SetWriteNewSeriesBackoffDuration(cfg.WriteNewSeriesBackoffDuration)
+	if value := cfg.Tick.SeriesBatchSize; value > 0 {
```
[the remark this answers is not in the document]
robskillington: Yeah, I mean I can make `Tick TickConfiguration` be a pointer type `Tick *TickConfiguration` and hence nil-able field on `Configuration` perhaps - and as such able to easily to specify to use the default configuration instead of what is listed.

That might help avoid what you're calling out here.

storage/namespace.go
```diff
@@ -146,9 +165,19 @@ type databaseNamespaceTickMetrics struct {
 	errors                 tally.Counter
 }
 
+// databaseNamespaceStatusMetrics are metrics emitted at a fixed interval
+// so that summing the value of gauges across hosts when graphed summarizing
+// values at the same fixed intervals can show meaningful results (vs variably
+// emitted values that can be aggregated across hosts to see a snapshot).
+type databaseNamespaceStatusMetrics struct {
```
prateek: nice!

storage/namespace.go
```diff
@@ -234,10 +268,27 @@ func newDatabaseNamespace(
 	}
 
 	n.initShards(nopts.NeedsBootstrap())
+	go n.reportStatusLoop()
 
 	return n, nil
 }
 
+func (n *dbNamespace) reportStatusLoop() {
+	ticker := time.NewTicker(namespaceReportLoopInterval)
```
prateek: Could you use the report interval from the instrument opts instead of using a const?

storage/namespace.go
```diff
@@ -234,10 +268,27 @@ func newDatabaseNamespace(
 	}
 
 	n.initShards(nopts.NeedsBootstrap())
+	go n.reportStatusLoop()
```
prateek: Could you add a leaktest check for ns.Open()+ns.Close() to validate we don't leak this in the future

storage/series/buffer.go
```diff
@@ -226,6 +232,33 @@ func bucketNeedsDrain(now time.Time, b *dbBuffer, idx int, start time.Time) int
...
+func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
+	// Perform a drain and reset if necessary
+	mergedOutOfOrderBlocks := bucketDrainAndReset(now, b, idx, start)
+
+	if b.buckets[idx].needsMerge() {
+		// Try to merge any out of order encoders to amortize the cost of a drain
+		r, err := b.buckets[idx].merge()
+		if err != nil {
+			log := b.opts.InstrumentOptions().Logger()
+			log.Errorf("buffer merge encode error: %v", err)
```
prateek: would it be useful to log the related series here too?

storage/series/buffer.go
```diff
@@ -592,39 +632,33 @@ func (b *dbBufferBucket) resetBootstrapped() {
...
-		b.encoders = b.encoders[:0]
-		return discardMergedResult{newBlock, 0}
-	}
+func (b *dbBufferBucket) hasJustSingleEncoder() bool {
+	return len(b.encoders) == 1 && len(b.bootstrapped) == 0
+}
 
-	if (len(b.encoders) == 0 ||
+func (b *dbBufferBucket) hasJustSingleBootstrappedBlock() bool {
+	encodersEmpty := len(b.encoders) == 0 ||
```
prateek: do the other bits of the code guarantee we won't have multiple empty encoders?

storage/series/buffer.go
```diff
@@ -660,20 +688,75 @@ func (b *dbBufferBucket) discardMerged() discardMergedResult {
...
 	for iter.Next() {
 		dp, unit, annotation := iter.Current()
 		if err := encoder.Encode(dp, unit, annotation); err != nil {
-			log := b.opts.InstrumentOptions().Logger()
-			log.Errorf("buffer merge encode error: %v", err)
-			break
+			return mergeResult{}, err
 		}
+		lastWriteAt = dp.Timestamp
+	}
```
prateek: do you need to check `iter.Err()` after this?

storage/tick.go
```diff
@@ -125,36 +143,41 @@ func (mgr *tickManager) Tick(softDeadline time.Duration, forceType forceType) er
...
+	// This number can be quite small if configuring to have small amount of
+	// shards (say 10 shards) with 32 series per shard (total of 320 series):
+	// = 10 num shards * ~32 series per shard * 100 microseconds default sleep per series
+	// = 10*32*0.1 milliseconds
+	// = 32ms
+	// Because of this we always sleep at least some fixed constant amount.
+	took := mgr.nowFn().Sub(start)
+	mgr.metrics.tickWorkDuration.Record(took)
+
+	mgr.runtimeOpts.RLock()
```
prateek: nit: could you make this get a method on the runtimeOpts struct

storage/shard.go
```diff
@@ -108,7 +106,9 @@ type dbShard struct {
 }
 
 type dbShardRuntimeOptions struct {
-	writeNewSeriesAsync bool
+	writeNewSeriesAsync      bool
```
prateek: Any reason to reuse the shard's lock instead of giving the struct a lock?

storage/tick.go
```diff
@@ -85,9 +93,19 @@ func newTickManager(database database, opts Options) databaseTickManager {
 		c:        context.NewCancellable(),
 		tokenCh:  tokenCh,
 	}
+
+	runtimeOptsMgr := opts.RuntimeOptionsManager()
+	runtimeOptsMgr.RegisterListener(mgr)
```
prateek: more a curiosity: is there a way to inspect the current runtime value in-use by the process (either log/metric)?

storage/namespace.go
robskillington: Sure thing, can do.

storage/namespace.go
robskillington: Sure thing, can do.

storage/series/buffer.go
robskillington: Yeah so for a bucket that returns `true from `canRead()` its guarenteed that there's data that is written.

storage/series/buffer.go
robskillington: Good catch, cheers.

storage/series/buffer.go
robskillington: Because buffer is a composition of a series (i.e. series has a buffer) and there is no cyclical link back there is no way to access that ID unless we add an ID ptr to the buffer too.  This may be a little expensive considering that is yet another 16 bytes per series to add.  In all honesty it should never happen unless the encoder is unable to encode already encoded data in the streams.  There is a pretty serious application error and if we ever encountered this we'd need to find, fix the bug and completely end of life that version.

storage/namespace.go
robskillington: Hey so this is already covered with `TestNamespaceCloseWithShard`.  It checks no more goroutines are running after a namespace close.

robskillington: Separated into two tests:

storage/tick.go
robskillington: Sure thing.

storage/shard.go
robskillington: There is a pretty good reason for this, so when we perform a write we get the read lock to check whether a series exists or not.  This requires just a single read lock, and to avoid needing to require multiple read locks we put this under the same read lock.

storage/tick.go
robskillington: Not currently no, it would be pretty trivial however to implement such a method as an RPC method (with an autogenerated HTTP/JSON endpoint).

storage/shard.go
prateek: Makes sense. Could you put that a NB blurb somewhere so we can capture for posterity

storage/namespace.go
prateek: Noice 👍

storage/namespace_test.go
```diff
@@ -876,15 +874,44 @@ func TestNamespaceCloseWithShard(t *testing.T) {
...
+	require.NoError(t, ns.Close())
+
+	// Check the namespace no long owns any shards
+	require.Empty(t, ns.GetOwnedShards())
+}
+
+func TestNamespaceCloseDoesNotLeak(t *testing.T) {
+	// Need to generate leaktest at top of test as that is when
+	// goroutines that are interesting are captured
+	leakCheck := leaktest.Check(t)
```
prateek: lol love the line up of defers here

services/m3dbnode/server/server.go
```diff
@@ -33,6 +33,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/grafana/grafana/pkg/cmd/grafana-cli/logger"
```
prateek: looks like this got pulled in randomly

storage/shard.go

services/m3dbnode/server/server.go
robskillington: Every time... thanks for the catch.

storage/series/buffer.go
prateek: Yeah fair. Only matters if finding which ID caused this would be particularly helpful in repro, sounds like we'd have bigger problems on hand than a single ID if this were to happen.

PR #406, opened 2017-12-11
Read Bloom Filters into memory inside of seeker

richardartoul:
- [x] Add a function for reading out the bloom filter from disk using an anonymous mmap region + validate the digest + write appropriate unit tests
- [x] Add container object for BloomFilter for lifecycle management (freeing the mmap'd region when we're done with it)
- [x] Read bloom filter in seeker with lifecycle management ( construct in Open() and release resources in Close() )


persist/fs/mmap_linux.go
```diff
@@ -26,6 +26,18 @@ import (
 )
 
 func mmap(fd, offset, length int64, opts mmapOptions) ([]byte, error) {
+	// MAP_SHARED because we want writes to the byte slice to be reflected back in the underlying file
+	return mmapBase(fd, offset, length, syscall.MAP_SHARED, opts)
```
robskillington: Hm, so for file based mmap'ing we actually do not want to allow writes or to sync writes to the underlying file.

Maybe we need to rethink the abstractions to make that clear.

We should only ever mmap immutable data.

persist/fs/read_bloom_filter.go
```diff
@@ -0,0 +1,70 @@
...
+
+	// Validate the bytes on disk using the digest, and read them into
+	// the mmap'd region
+	_, err = bloomFilterFdWithDigest.ReadAllAndValidate(
+		anonMmap, expectedDigest)
+	if err != nil {
+		return nil, err
+	}
+
+	// TODO: Use conccurrent one?
```
robskillington: Yeah I think we want the concurrent bloom filter here as we'll want to check it before enqueuing a fetch. (i.e. before locking).

persist/fs/read_fileset_digests.go
```diff
@@ -0,0 +1,64 @@
...
+	)
+
+	if fsDigests.infoDigest, err = digestFdWithDigestContents.ReadDigest(); err != nil {
+		return fsDigests, err
+	}
+	if fsDigests.indexDigest, err = digestFdWithDigestContents.ReadDigest(); err != nil {
+		return fsDigests, err
+	}
+	if fsDigests.summariesDigest, err = digestFdWithDigestContents.ReadDigest(); err != nil {
+		// Skip the summaries digest
```
robskillington: We're not skipping it per say though in this method yeah?

persist/fs/read_write_test.go
```diff
@@ -70,18 +71,43 @@ func readTestData(t *testing.T, r FileSetReader, shard uint32, timestamp time.Ti
...
+	assert.Equal(t, expectedK, bloomFilter.K())
+	// EstimateFalsePositiveRate always returns at least 1, so skip this check
+	// if len entries is 0
+	if len(entries) > 0 {
+		assert.Equal(t, expectedM, bloomFilter.M())
+	}
+
 	for i := 0; i < r.Entries(); i++ {
-		id, data, checksum, err := r.Read()
+		idFromRead, data, checksumFromRead, err := r.Read()
```
robskillington: Maybe you should Finalize the idFromRead too?

persist/fs/seek.go
```diff
@@ -129,8 +146,9 @@ func newSeeker(opts seekerOpts) fileSetSeeker {
 		filePathPrefix:             opts.filePathPrefix,
 		infoFdWithDigest:           digest.NewFdWithDigestReader(opts.infoBufferSize),
 		indexFdWithDigest:          digest.NewFdWithDigestReader(opts.dataBufferSize),
-		dataReader:                 bufio.NewReaderSize(nil, opts.seekBufferSize),
 		digestFdWithDigestContents: digest.NewFdWithDigestContentsReader(opts.infoBufferSize),
+		bloomFilterFdWithDigest:    digest.NewFdWithDigestContentsReader(opts.dataBufferSize),
```
robskillington: Here too, think you want `NewFdWithDigestReader` instead.

persist/fs/seek.go
```diff
@@ -198,7 +218,16 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
...
 	s.dataFd = dataFd
-	return nil
+
+	s.bloomFilter, err = readBloomFilter(
+		bloomFilterFd,
+		s.bloomFilterFdWithDigest,
+		s.expectedBloomFilterDigest,
+		uint(s.bloomFilterInfo.NumElementsM),
+		uint(s.bloomFilterInfo.NumHashesK),
+	)
```
robskillington: Might want to move this above the keepunreadbuf stanza and if `err != nil` then perform `s.Close(); return err` like the other bailout cases?

persist/fs/read_bloom_filter.go
```diff
@@ -0,0 +1,70 @@
...
+	bloomFilter := bloom.NewReadOnlyBloomFilter(numElementsM, numHashesK, anonMmap)
+	closeFn := func() error {
+		return munmap(anonMmap)
+	}
+
+	return block.NewManagedBloomFilter(bloomFilter, closeFn), nil
```
robskillington: Hmm maybe we should make the managed bloom filter managed the mmap lifecycle somehow, could we make it a type in the `fs` package?  i.e. `BloomFilterFile` or something?  That way there doesn't have to be two allocations, one for the `closeFn` anonymous function and one for the actual wrapper itself.

persist/fs/mmap_linux.go
richardartoul: Yeah I knew that, not sure why I left it the way it is. Writes will still fail because we're setting the read prop but not the write prop.

Anyways if we never want anything to propagate we should just use anonymous, not sure we need to rethink the abstractions here

persist/fs/read.go
```diff
@@ -67,8 +68,9 @@ type reader struct {
 	start          time.Time
 	blockSize      time.Duration
 
-	infoFdWithDigest           digest.FdWithDigestReader
-	digestFdWithDigestContents digest.FdWithDigestContentsReader
+	infoFdWithDigest               digest.FdWithDigestReader
+	digestFdWithDigestContents     digest.FdWithDigestContentsReader
+	bloomFilterFWithDigestContents digest.FdWithDigestContentsReader
```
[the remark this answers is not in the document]
richardartoul: Oh interesting, I wonder why it worked. I think the interface is a superset? Anyways I'll update this

persist/fs/read_bloom_filter.go
richardartoul: Yeah totally, I think originally I didn't want to do that because we were putting references to the boom filter all over the place (in shardresult and dbshard etc etc) and I didn't want everything importing the fs package, but now that its just the seeker this is a good idea

persist/fs/read_fileset_digests.go
richardartoul: yep old comment from before when I extracted this method out, will fix

persist/fs/seek.go
richardartoul: will fix

persist/fs/read_write_test.go
richardartoul: will fix

persist/fs/seek.go
richardartoul: yeah good call

persist/fs/bloom_filter.go
```diff
@@ -0,0 +1,126 @@
...
+	// Close closes the ManagedBloomFilter, releasing any held resoures
+	Close() error
+}
+
+// managedConcurrentBloomFilter is the same as managedBloomFilter, with the
+// addition that the implementation must be safe for concurrent use
+type managedConcurrentBloomFilter interface {
+	managedBloomFilter
+
+	IsConcurrent()
```
robskillington: Haha - I like where you went here, forcing Go's type system to be useful can be challenging sometimes.  

Considering all these types are internal to just this package it might be simpler to just make a single struct `type managedConcurrentBloomFilter struct { ... }` (there won't be a no need to mock these I don't think, so the value of an interface is not as high).

richardartoul: Yeah its a cool little pattern I've seen in some open-source code before. I think you're right though, I probably overdid it in this case. I'll just make it the one struct

persist/fs/read.go
```diff
@@ -315,6 +324,16 @@ func (r *reader) ReadMetadata() (id ts.ID, length int, checksum uint32, err erro
 	return r.entryID(entry.ID), int(entry.Size), uint32(entry.Checksum), nil
 }
 
+func (r *reader) ReadBloomFilter() (managedConcurrentBloomFilter, error) {
```
robskillington: If this method is exported, we'll need to return an exported type.

persist/fs/types.go
```diff
@@ -67,6 +67,10 @@ type FileSetReader interface {
 	// Use either Read or ReadMetadata to progress through a volume, but not both.
 	ReadMetadata() (id ts.ID, length int, checksum uint32, err error)
 
+	// ReadBloomFilter returns the bloom filter stored on disk in a container object that is safe
+	// for concurrent use and has a Close() method for releasing resources when done.
+	ReadBloomFilter() (managedConcurrentBloomFilter, error)
```
robskillington: This method won't be useful outside of this package if we don't return an exported type.

persist/fs/seek.go
```diff
@@ -364,10 +402,16 @@ func (s *seeker) Close() error {
 	for key := range s.indexMap {
 		delete(s.indexMap, key)
 	}
-	if s.dataFd == nil {
-		return nil
+
+	multiErr := xerrors.NewMultiError()
+	if s.dataFd != nil {
+		multiErr.Add(s.dataFd.Close())
```
robskillington: Missing `multiErr = multiErr.Add(...)` here.  

I might open an issue that we should add a metalinter for that perhaps sometime haha.

persist/fs/seek.go
```diff
@@ -364,10 +402,16 @@ func (s *seeker) Close() error {
...
+		s.dataFd = nil
+	}
+	if s.bloomFilter != nil {
+		multiErr.Add(s.bloomFilter.Close())
```
robskillington: Same here.

persist/fs/mmap_other.go
```diff
@@ -27,7 +27,22 @@ import (
...
+
+// mmapAnon requests a private (non-shared) region of anonymous (not backed by a file) memory from the O.S
+func mmapAnon(length int64, opts mmapOptions) ([]byte, error) {
+	// offset is 0 because we're not indexing into a file
+	// fd is -1 and MAP_ANON because we're asking for an anonymous region of memory not tied to a file
+	// MAP_PRIVATE because we don't plan on sharing this region of memory with other processes
+	return mmapBase(-1, 0, length, syscall.MAP_ANON|syscall.MAP_PRIVATE, opts)
+}
+
+func mmapBase(fd, offset, length int64, flags int, opts mmapOptions) ([]byte, error) {
```
robskillington: Maybe call this `mmapFd`?  Seems less OO-ish and more Go-ish, he.

robskillington: Actually on second thought, it might make it better to specialize the other two.  i.e. `mmap(...)` -> `mmapFd(...)` and `mmapAnon(...)` -> `mmapBytes(...)` and then `mmapBase(...)` -> `mmap(...)`

persist/fs/read.go
richardartoul: Its an unexported type with exported methods so people can still use it

persist/fs/types.go
richardartoul: same comment as above

persist/fs/seek.go
richardartoul: oh wow, I didn't know that..good catch

persist/fs/seek.go
richardartoul: will fix

persist/fs/mmap_other.go
richardartoul: sounds good

persist/fs/read_write_test.go
```diff
@@ -70,18 +71,44 @@ func readTestData(t *testing.T, r FileSetReader, shard uint32, timestamp time.Ti
...
```
robskillington: nit: It might be best to wrap this up so that the defer's are executed inside the loop, just to avoid tests using tons of memory.  It's fine that it will be a little slower because of the fn declaration, which will hopefully be inlined.

i.e.
```go
for i := 0; i < r.Entries(); i++ {
  func() {
    x := ...
    defer x.Finalize()

    y := ...
    defer y.Finalize()
    
    ...
  }()
}
```

PR #409, opened 2017-12-12
Fix race condition allowing multiple Tick() calls to a single Shard

prateek:


storage/shard.go
```diff
@@ -354,27 +361,61 @@ func (s *dbShard) Close() error {
...
-	return s.tickAndExpire(c, softDeadline, tickPolicyRegular)
+	return s.tickAndExpire(c, softDeadline, tickPolicyRegular, defaultSkipTickIfShardClosing)
 }
 
 func (s *dbShard) tickAndExpire(
 	c context.Cancellable,
 	softDeadline time.Duration,
 	policy tickPolicy,
-) tickResult {
+	skipIfClosing bool,
```
robskillington: Can you combine `tickPolicy` and `skipIfClosing`?  i.e. just change `tickPolicyForceExpiry` to `tickPolicyCloseShard`

storage/shard.go
```diff
@@ -354,27 +361,61 @@ func (s *dbShard) Close() error {
...
+) (tickResult, error) {
+	// ensure only one tick can execute at a time
+	if s.ticking.Swap(true) {
+		// i.e. we were previously ticking
+		return tickResult{}, errShardAlreadyTicking
+	}
```
robskillington: If the tickPolicy is `tickPolicyRegular` and `s.isClosing()` can we also return err?

storage/shard.go
```diff
@@ -354,27 +361,61 @@ func (s *dbShard) Close() error {
...
+	s.tickWg.Wait()
+
 	// NB(r): Asynchronously we purge expired series to ensure pressure on the
 	// GC is not placed all at one time.  If the deadline is too low and still
 	// causes the GC to impact performance when closing shards the deadline
 	// should be increased.
 	cancellable := context.NewNoOpCanncellable()
 	softDeadline := s.opts.TickInterval()
-	s.tickAndExpire(cancellable, softDeadline, tickPolicyForceExpiry)
+	s.tickAndExpire(cancellable, softDeadline, tickPolicyForceExpiry, false)
```
robskillington: Probably want to check if this returns an error and return it yeah?

storage/shard.go

storage/shard.go

storage/shard.go
prateek: nice catch, will do

storage/shard.go

storage/shard.go

storage/namespace.go
```diff
@@ -435,6 +438,8 @@ func (n *dbNamespace) Tick(c context.Cancellable) {
 	n.metrics.tick.madeUnwiredBlocks.Inc(int64(r.madeUnwiredBlocks))
 	n.metrics.tick.mergedOutOfOrderBlocks.Inc(int64(r.mergedOutOfOrderBlocks))
 	n.metrics.tick.errors.Inc(int64(r.errors))
+
+	return multiErr.FinalError()
```
robskillington: I don't think we want to emit stats or set the last tick stats unless err == nil.

prateek: fixed up

glide.yaml
```diff
@@ -89,6 +89,9 @@ import:
 - package: github.com/spf13/pflag
   version: 4f9190456aed1c2113ca51ea9b89219747458dc1
 
+- package: go.uber.org/atomic
```
robskillington: Do we need to pull this in anymore?

storage/namespace.go
```diff
@@ -406,18 +407,24 @@ func (n *dbNamespace) Tick(c context.Cancellable) {
...
+			multiErr = multiErr.Add(err)
 			l.Unlock()
 		})
 	}
 
 	wg.Wait()
 
 	if c.IsCancelled() {
-		return
+		return multiErr.FinalError()
```
robskillington: Is it worth actually returning here?  As long as err == nil even if it just got cancelled we should probably inc stats, etc.

glide.yaml
prateek: doh I remember thinking ill forget to revert that :|

storage/namespace.go
```diff
@@ -406,18 +407,24 @@ func (n *dbNamespace) Tick(c context.Cancellable) {
...
+	}
+
+	if err := multiErr.FinalError(); err != nil {
```
robskillington: Maybe combine these two together? e.g. `if err := multiErr.FinalError(); err != nil || c.IsCancelled() { return err }`

storage/shard_test.go
```diff
@@ -476,6 +487,145 @@ func TestShardWriteAsync(t *testing.T) {
 	require.True(t, ok)
 }
 
+// This tests a race in shard ticking with an empty series pending expiration.
+func TestShardTickRace(t *testing.T) {
```
robskillington: Good set of tests 👍

storage/namespace.go
prateek: sure thing

PR #410, opened 2017-12-13
Amortize cost of out of order merges by performing on each tick

robskillington:


storage/series/buffer.go
```diff
@@ -660,20 +688,78 @@ func (b *dbBufferBucket) discardMerged() discardMergedResult {
 		}
 	}
 
+	var lastWriteAt time.Time
 	iter := b.opts.MultiReaderIteratorPool().Get()
 	iter.Reset(readers)
+
+	defer func() {
+		iter.Close()
+		for _, stream := range readers {
```
prateek: nice, good catch!

storage/series/buffer.go
```diff
@@ -660,20 +688,78 @@ func (b *dbBufferBucket) discardMerged() discardMergedResult {
...
+		b.resetEncoders()
+		b.resetBootstrapped()
+		b.empty = true
+	}()
+
+	if b.hasJustSingleEncoder() {
+		// Already merged as a single encoder
+		encoder := b.encoders[0].encoder
+		newBlock := b.opts.DatabaseBlockOptions().DatabaseBlockPool().Get()
+		newBlock.Reset(b.start, encoder.Discard())
```
prateek: Do we have to put the encoder back into the pool here?

storage/series/buffer.go
```diff
@@ -660,20 +688,78 @@ func (b *dbBufferBucket) discardMerged() discardMergedResult {
...
+	result, err := b.merge()
+	if err != nil {
+		return discardMergedResult{}, err
+	}
+
+	merged := b.encoders[0].encoder
 
 	newBlock := b.opts.DatabaseBlockOptions().DatabaseBlockPool().Get()
-	newBlock.Reset(b.start, encoder.Discard())
+	newBlock.Reset(b.start, merged.Discard())
```
prateek: same question as above for the lifecycle of the encoder here

storage/series/buffer.go
robskillington: `encoder.Discard()` will recycle the encoder, it calls `.Close()` after stealing the buffer =]

storage/series/buffer.go
robskillington: `.Discard()` recycles the encoder 👍

PR #411, opened 2017-12-13
Refactor seeker to use bloom filter and summaries file

richardartoul:
- [X] mmap summaries file into memory, and implement life-cycle management inside of the seeker
- [X] read summaries file entries offsets into in-memory data-structure that can be binary searched and implement life-cycle management inside of seeker
- [X] modify seeker to use new binary-searchable data structure to find nearest index for the requested ID in the index file, and then scan forward until the desired ID is found or we're confident the ID does not exist
- [X] modify the block retriever to use the concurrent bloom filter in the seeker to quickly determine if its possible a given ID exists before attempting to seek it out on disk
- [X] implement a robust property test to verify the behavior of using the new data structure / summaries files


persist/fs/index_lookup.go
```diff
@@ -0,0 +1,197 @@
...
+	// the index file itself. Because of that, the binary search that we're
+	// performing is "optimistic". We're trying to find either an exact match,
+	// OR the nearest match that is to the left of the series we're searching
+	// for (so we keep track of it everytime we move right).
+	bestMatchSoFar := int64(-1)
+
+	for {
+		if min > max {
+			if bestMatchSoFar == -1 {
+				return -1, false, nil
```
robskillington: Hm, this is dependent on the first index entry always being available as a summary yeah?  I wonder if we should depend on this or not and just start searching from the start of the index if there's no match?

persist/fs/index_lookup.go
```diff
@@ -0,0 +1,197 @@
...
+
+func newNearestIndexOffsetLookup(
+	summaryIDsOffsets []encoding.IndexSummaryToken,
+	summariesBytes []byte,
+	decoder encoding.Decoder,
+	decoderStream encoding.DecoderStream,
+) *nearestIndexOffsetLookup {
+	return &nearestIndexOffsetLookup{
+		summaryIDsOffsets: summaryIDsOffsets,
+		summariesBytes:    summariesBytes,
```
robskillington: Considering we munmap with close, perhaps better to name this byte buffer so it's explicit and not weird when reading the code?  i.e. `summariesMmap`?

persist/fs/index_lookup.go
```diff
@@ -0,0 +1,197 @@
...
+		decoderStream:     decoderStream,
+		decoder:           decoder,
+		msgpackDecoder:    msgpack.NewDecoder(nil),
```
robskillington: Maybe we should make the `encoding.Decoder` be able to return a `IndexOffsetLookup`?  That way we can reuse msgpack decoder from it?  (Also means that `persist/fs/` package doesn't need to know about the fact that the `encoding.Decoder` is msgpack encoded which seems very leaky that this is here).

persist/fs/index_lookup.go
```diff
@@ -0,0 +1,197 @@
...
+func (il *nearestIndexOffsetLookup) close() error {
+	return munmap(il.summariesBytes)
+}
+
+// readNearestIndexOffsetLookupFromSummaries creates an nearestIndexOffsetLookup
+// from an index summaries file by reading the summaries file into an anonymous
+// mmap'd region, and also creating the slice of summaries offsets which is
+// required to binary search the data structure. It will also make sure that
+// the summaries file is sorted (which it always should be).
+func readNearestIndexOffsetLookupFromSummaries(
```
robskillington: nit: Might make sense to just prefix this with `new` since its really kind of a specialized constructor?

persist/fs/index_lookup_prop_test.go
```diff
@@ -0,0 +1,240 @@
...
+			summariesFile, summariesFdWithDigest, expectedSummariesDigest, decoder, len(writes))
+		if err != nil {
+			return false, fmt.Errorf("err reading index lookup from summaries file: %v, ", err)
+		}
+
+		// Make sure it returns the correct index offset for every ID
+		for id, expectedOffset := range expectedIndexFileOffsets {
+			foundOffset, ok, err := indexLookup.getNearestIndexFileOffset(ts.StringID(id))
+			if err != nil {
+				return false, fmt.Errorf("Err locating index file offset for: %s, err: %v", id, err)
```
robskillington: nit: Lowercase the first char of the error (similar to our log output naming convention)

persist/fs/index_lookup_prop_test.go
```diff
@@ -0,0 +1,240 @@
...
+			}
+			if !ok {
+				return false, fmt.Errorf("Unable to locate index file offset for: %s", id)
```
robskillington: nit: Lowercase the first char of the error (similar to our log output naming convention)

persist/fs/index_lookup_prop_test.go
```diff
@@ -0,0 +1,240 @@
...
+			}
+			if expectedOffset != foundOffset {
+				return false, fmt.Errorf(
```
robskillington: nit: Lowercase the first char of the error (similar to our log output naming convention)

persist/fs/retriever.go
```diff
@@ -179,13 +189,21 @@ func (r *blockRetriever) fetchLoop(seekerMgr FileSetSeekerManager) {
...
-		// all written.
+		// all written. Note that this sort does NOT mean that we're going to stripe
+		// through different files at once as you might expect at first, but simply
+		// that since all the fileset files are written at the end of a block period
+		// those files are more likely to be physically located close to each other
+		// on disk. In other words, instead of accessing files like this:
+		// 		shard1T1 --> shard1T2 --> shard1T3 --> shard2T1 --> shard2T2 --> shard2T3
+		// its probably faster to access them like this:
+		// 		shard1T1 --> shard2T1 --> shard1T2 --> shard2T2 --> shard1T3 --> shard2T3
+		// so we re-arrange the order of the requests to achieve that
```
robskillington: Good call on the notes here, the explanation I left was not super detailed =]

persist/fs/retriever.go
```diff
@@ -285,6 +308,23 @@ func (r *blockRetriever) Stream(
 	startTime time.Time,
 	onRetrieve block.OnRetrieveBlock,
 ) (xio.SegmentReader, error) {
+	// This should never happen
+	if len(r.seekerMgrs) < 1 {
```
robskillington: It might be cleaner to assign a single seeker mgr to be the "main" one for bloom filter lookups and use that one here?  We can also assign a no-op one to it (like a mock seeker mgr) by default so that if we haven't opened the block retriever or w/e it just returns similarly to a "not found".

persist/fs/retriever.go
```diff
@@ -285,6 +308,23 @@ func (r *blockRetriever) Stream(
...
+	// just have multiple for concurrency reasons), so we use the first one
+	// because it's guaranteed to be there (concurrency cannot be < 1)
+	seeker, err := r.seekerMgrs[0].Seeker(shard, startTime)
+	if err != nil {
+		return nil, err
+	}
+	// If the ID is not in the seeker's bloom filter, then it's definitely not on
+	// disk and we can return immediately
+	if !seeker.ConcurrentIDBloomFilter().Test(id.Data().Get()) {
+		return nil, errSeekIDNotFound
```
robskillington: I'm not sure if we treat an error here as if it's a not found error.  I think traditionally we call back the `req.onRetrieved` with an empty segment (which is what it looks like by looking at `fetchBatch` call).

.travis.yml
```diff
@@ -15,6 +15,7 @@ env:
     - PACKAGE=github.com/m3db/m3db
   matrix:
     - MAKE_TARGET="test-ci-unit"
-    - MAKE_TARGET="test-ci-integration"
+    - MAKE_TARGET="test-ci-integration-mem-policy-all"
+    - MAKE_TARGET="test-ci-integration-mem-policy-default"
```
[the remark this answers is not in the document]
richardartoul: yeah fair, I'll remove the makefile and setup changes. I'm gonna leave in a few minor tweaks to the integration changes cause I noticed that FilesystemOptions was not being propagated properly everywhere

persist/fs/index_lookup.go
richardartoul: sure

persist/fs/index_lookup.go
richardartoul: You're right, it would be much better not to depend on that being true. If thats the case, then I can probably get rid of the bool in the return value ("found"), and just have it return 0 cause this is the only case where it can happen.

persist/fs/index_lookup.go
richardartoul: sure

persist/fs/index_lookup.go
richardartoul: Ended up nuking the encoding package (deleted most interfaces) and moving a few of the files that were in there into the msgpack folder, and the msgpack folder is now a subpackage of persist/fs (mostly just so I don't have to add 10 files all pre-fixed with msgpack_ to the folder)

persist/fs/index_lookup_prop_test.go
richardartoul: done

persist/fs/index_lookup_prop_test.go
richardartoul: done

persist/fs/index_lookup_prop_test.go
richardartoul: done

persist/fs/retriever.go
richardartoul: I think you're right. Unfortunately the existing behavior for this method is not documented with a test, so I ended up writing one but it looks like I didn't understand the existing behavior properly. I'll try and write a better test that captures the existing codes intentions

persist/fs/retriever.go
richardartoul: I don't like the idea that if you trying and Stream() before Open() you get the equivalent of a "not found", I think that should probably be an "use before open" error or something. The existing code kind of captures that, although I can make one of the seeker managers the "main" one. Should I just add another field to the struct for that?

persist/fs/retriever.go
richardartoul: OK I think I got this all figured out, will update it soon. Out of curiosity, how come this didn't manifest in production? Is it because we use the index to lookup IDs first, so we almost always know an ID exists before we query for it?

richardartoul: Gonna add a new integration test for this because we don't have one for querying an ID that does not exist

richardartoul: @robskillington I tried to add an integration test for this, and I don't think I can because it relies too heavily on the ReadEncoded path which you're heavily modifying in your P.R. Could you add an integration test to your P.R that tests reading an ID from disk that doesn't exist and make sure it returns an empty array and not an error? Could also modify an existing integration test, I'll handle the unit test on my side

persist/fs/index_lookup.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// newNearestIndexOffsetLookupFromSummariesFile creates an nearestIndexOffsetLookup
+// from an index summaries file by reading the summaries file into an anonymous
+// mmap'd region, and also creating the slice of summaries offsets which is
+// required to binary search the data structure. It will also make sure that
+// the summaries file is sorted (which it always should be).
+func newNearestIndexOffsetLookupFromSummariesFile(
+	summariesFd *os.File,
+	summariesFdWithDigest digest.FdWithDigestReader,
```
robskillington: Hm, why pass both of these?  Perhaps just the second arg and expect that it's already been reset?

You can get the `*os.File` ref back with `summariesFdWithDigest.Fd()`

persist/fs/index_lookup.go
```diff
@@ -0,0 +1,200 @@
...
+
+// nearestIndexOffsetLookup provides a way of quickly determining the nearest offset of an
+// ID in the index file. It is not safe for concurrent use
+type nearestIndexOffsetLookup struct {
+	summaryIDsOffsets []xmsgpack.IndexSummaryToken
+	// bytes from file mmap'd into anonymous region
+	summariesMmap []byte
+	// reusable decoder stream
+	decoderStream  xmsgpack.DecoderStream
+	decoder        *xmsgpack.Decoder
```
robskillington: The `decoder *xmsgpack.Decoder` doesn't actually look used?

persist/fs/index_lookup.go
robskillington: I'm also having a hard time working out whether this should be a standalone method or a method on the seeker itself?  It seems a lot of the variables have lifecycles related to the seeker so it would make more sense there, but that's just off the vibe I get reading it.

persist/fs/index_lookup_prop_test.go
```diff
@@ -0,0 +1,236 @@
+// +build big
```
robskillington: Can't run this normally? =[

persist/fs/retriever.go
```diff
@@ -242,16 +261,33 @@ func (r *blockRetriever) fetchBatch(
...
+		// If the ID does not exist, don't try to seek to it or we'll get a checksum
+		// mismatch error because default offset value for indexEntry is zero.
+		if req.doesNotExist {
+			data, err = nil, nil
+		} else {
+			data, err = seeker.SeekByIndexEntry(req.indexEntry)
+			if err != nil && err != errSeekIDNotFound {
+				req.onError(err)
+				continue
+			}
```
robskillington: How about just `if !req.doesNotExist { ... }` then you don't have to set data and err to nil explicitly?

persist/fs/retriever.go
```diff
@@ -285,18 +321,39 @@ func (r *blockRetriever) Stream(
...
+	// It doesn't matter which seekerManager we use (they're all identical, we
+	// just have multiple for concurrency reasons), so we use the first one
+	// because it's guaranteed to be there (concurrency cannot be < 1)
+	seeker, err := r.seekerMgrs[0].Seeker(shard, startTime)
+	if err != nil {
+		return nil, err
+	}
+	// If the ID is not in the seeker's bloom filter, then it's definitely not on
+	// disk and we can return immediately
+	if !seeker.ConcurrentIDBloomFilter().Test(id.Data().Get()) {
```
robskillington: Sorry to flip flop on this, I thought about this again and it definitely would cause a lot of extra concurrency and scheduling for not much gain to keep the contract the same for both bloom filter lookups that are negative and actual lookups that are negative after searching disk.

I think the issue before was that we were returning `errSeekIDNotFound` which is impossible to test for equality as its not an exported error from the `fs` package.  We should either update the interface to be xplicilt

persist/fs/retriever.go
```diff
@@ -437,9 +497,10 @@ func (req *retrieveRequest) resetForReuse() {
 	req.id = nil
 	req.start = time.Time{}
 	req.onRetrieve = nil
-	req.seekOffset = -1
+	req.indexEntry = IndexEntry{}
 	req.reader = nil
 	req.err = nil
+	req.doesNotExist = false
```
robskillington: nit: Super minor nit, perhaps `notFound` is more terse/Go-ish?

persist/fs/seek.go
```diff
@@ -221,14 +213,39 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
...
+		s.expectedSummariesDigest,
+		s.decoder,
+		int(s.summariesInfo.Summaries),
+	)
+	if err != nil {
+		s.Close()
+		return err
+	}
+
 	if !s.keepUnreadBuf {
```
robskillington: Do we need this unreadbuf anymore?  Can we kill this whole concept?

persist/fs/seek.go
```diff
@@ -221,14 +213,39 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
...
 		// NB(r): Free the unread buffer and reset the decoder as unless
 		// using this seeker in the seeker manager we never use this buffer again
 		s.unreadBuf = nil
 		s.decoder.Reset(nil)
 	}
 
+	// Make sure the indexFd is at the beginning of the file
+	_, err = indexFd.Seek(0, 0)
```
robskillington: Do we need to seek to zero here?  It doesn't look like anything uses the `indexFd` before the mmap call?

persist/fs/seek.go
```diff
@@ -221,14 +213,39 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
...
+		return err
+	}
+	indexMmapResult, err := mmapFile(indexFd, mmapOptions{read: true})
+	if err != nil {
+		s.Close()
+		return err
+	}
+
 	s.dataFd = dataFd
+	s.indexMmap = indexMmapResult.result
```
robskillington: Should we also log the warning?

persist/fs/seek.go
```diff
@@ -278,51 +296,27 @@ func (s *seeker) readInfo(size int) error {
...
-	if !exists {
-		return nil, errSeekIDNotFound
-	}
-
-	_, err := s.dataFd.Seek(entry.offset, 0)
+// SeekByIndexEntry is similar to Seek, but uses the provided IndexEntry
+// instead of looking it up on its own. Useful in cases where you've already
+// obtained an entry and don't want to waste resources looking it up again.
+func (s *seeker) SeekByIndexEntry(entry IndexEntry) (checked.Bytes, error) {
+	_, err := s.dataFd.Seek(entry.Offset, 0)
```
robskillington: Hm, I thought we'd probably be using mmap for the data file too to reduce syscalls and reuse page caches (if some of them begin being read very frequently).  Thoughts?

persist/fs/seek.go
```diff
@@ -368,25 +362,58 @@ func (s *seeker) Seek(id ts.ID) (checked.Bytes, error) {
...
+
+	// Prevent panic's when we're scanning to the end of the buffer
+	for stream.Remaining() != 0 {
+		entry, err := s.decoder.DecodeIndexEntry()
+		// Should never happen, either something is really wrong with the code or
+		// the file on disk was corrupted
+		if err != nil {
+			return IndexEntry{}, err
+		}
+		comparison := bytes.Compare(entry.ID, id.Data().Get())
```
robskillington: Can we move the `id.Data().Get()` call to before this for loop and keep the ref in a local var?  Doing ref counting for each lookup might be expensive.

tools/read_index_ids/main/main.go
```diff
@@ -47,19 +44,23 @@ func main() {
 	})
 	bytesPool.Init()
 
-	seeker := fs.NewSeeker(*optPathPrefix, defaultDataBufferReadSize, defaultInfoBufferReadSize, defaultSeekBufferReadSize, bytesPool, nil)
-
-	err := seeker.Open(ts.StringID(*optNamespace), *optShard, time.Unix(0, *optBlockstart))
+	reader, err := fs.NewReader(bytesPool, fs.NewOptions())
 	if err != nil {
-		log.Fatalf("unable to open file: %v", err)
+		log.Fatalf("Could not create new reader: %v", err)
```
robskillington: nit: Use lower case log message characters for first character.

persist/fs/index_lookup.go
richardartoul: I didn't know about the Fd() method, condensed it to a single argument.

I think its a little cleaner as a stand-alone method (I did the same thing for the bloom filter) and could be potentially re-used in the reader if we ever wanted to

persist/fs/index_lookup.go
richardartoul: You're right it was used in the new function but didn't need to be persisted in the struct. Removed

persist/fs/index_lookup_prop_test.go
richardartoul: Test takes about 20 seconds to run and if I run it normally it'll timeout.

I could reduce the number of tests but then its not much of a prop test. To be clear, with the "big" build tag it will still be run in C.I, it's just run serially with all the other big tests at the end

persist/fs/retriever.go
richardartoul: good call

persist/fs/retriever.go
richardartoul: good call, it should match the error name

persist/fs/seek.go
richardartoul: Looks like its used to read the info file, so not sure if this P.R affects that

persist/fs/seek.go
richardartoul: you're right, I'm not sure why I put this there. Maybe just being overly precautious

persist/fs/index_lookup_prop_test.go
robskillington: Ah yes, forgot about the fact that we run these at the end.  The `adhoc` tag was the one I was thinking about.

persist/fs/seek.go
richardartoul: yep good call

persist/fs/retriever.go
robskillington: After syncing looks like we can do this without any goroutine scheduling as we will no longer launch a goroutine for onRetrieve if there is no data.

persist/fs/seek.go
richardartoul: good call

persist/fs/seek.go
richardartoul: Spoke in person, decided to handle this in a subsequent P.R

persist/fs/retriever.go
richardartoul: Spoke in person, decided that the existing implementation is fast enough, but that we should not call `go req.onRetrieve.OnRetrieveBlock()` if there is no data (Not found in files or not in bloom filter)

persist/fs/index_lookup.go
```diff
@@ -0,0 +1,195 @@
...
+		// We ignore the entry itself because we don't need any information from it
+		entry, entryMetadata, err := decoder.DecodeIndexSummary()
+		if err != nil {
+			munmap(summariesMmap)
+			return nil, err
+		}
+
+		// Make sure that all the IDs are sorted as we iterate, and return an error
+		// if they're not. This should never happen as files should be sorted on disk.
+		if lastReadID != nil && bytes.Compare(lastReadID, entry.ID) != -1 {
```
robskillington: `lastReadID` is not actually set at the bottom of this loop it seems?  

Can we add a test to catch this in the future too perhaps?

persist/fs/msgpack/stream.go
```diff
@@ -59,11 +59,23 @@ type DecoderStream interface {
...
+	Offset() int
 }
 
+// DecoderStream is a data stream that is read by the decoder,
+// it takes both a reader and the underlying backing bytes.
+// This is constructed this way since the decoder needs access
+// to the backing bytes when taking refs directly for decoding byte
+// slices without allocating bytes itself but also needs to progress
+// the reader (for instance when a reader is a ReaderWithDigest that
+// is calculating a digest as its being read).
```
robskillington: Does the comment need to be duplicated here considering it's already on the interface?

persist/fs/retriever.go
```diff
@@ -285,18 +321,45 @@ func (r *blockRetriever) Stream(
...
+	// because it's guaranteed to be there (concurrency cannot be < 1)
+	seeker, err := seekerMgr.Seeker(shard, startTime)
+	if err != nil {
+		return nil, err
+	}
+	// If the ID is not in the seeker's bloom filter, then it's definitely not on
+	// disk and we can return immediately
+	if !seeker.ConcurrentIDBloomFilter().Test(id.Data().Get()) {
+		// No need to call req.onRetrieve.OnRetrieveBlock if there is no data
+		req.resultWg.Done()
```
robskillington: Instead of calling `req.resultWg.Done()` directly, for better encapsulation perhaps is better instead to `req.onRetrieved(ts.Segment{})`

persist/fs/index_lookup.go
richardartoul: Good catch, added the unit test

persist/fs/msgpack/stream.go
richardartoul: nope I meant to delete that, thanks

persist/fs/retriever.go
richardartoul: yeah fair

Issue #418, opened 2017-12-19
Write property test for series buffer to validate ownership semantics during rotation or merging


robskillington:
There was a bad change to ownership semantics where encoders were being free'd after they were discard (because `resetEncoders()` was called which registered the encoder as a closer on the buffer bucket after the encoder in the bucket had already been discarded and hence the buffer was still in use).

Here's the PR that introduces the bug:
https://github.com/m3db/m3db/pull/410

Here's the PR with the fix:
https://github.com/m3db/m3db/pull/419

We need property tests on the delicate ownership semantics in the series buffer to catch these types of changes.

PR #419, opened 2017-12-19
Fix buffer ownership semantics freeing buffers twice

robskillington:
Opened an issue to add property tests to catch future bad ownership semantic changes, issue #418.


richardartoul:
LGTM

PR #423, opened 2017-12-22
Add support for static topology

benraskin92:
Couple comments/questions: 

1. Where should I put the `StaticConfiguration` struct- it's in the topology package for now?
2. Need to add some tests, but getting this error: `?   	github.com/m3db/m3db/services/m3dbnode/server	[no test files]`  - will look into this
3. I have an example config file for static topology, so wondering if I should put that somewhere specific

### DIFF
--- a/services/m3dbnode/config/config.go
+++ b/services/m3dbnode/config/config.go
@@ -97,10 +97,15 @@ type Configuration struct {
 	ConfigService *etcdclient.Configuration `yaml:"configService"`
 
 	// The static configuration for config service client.
-	StaticConfigService *topology.StaticConfiguration `yaml:"staticConfigService"`
+	StaticConfig *StaticConfiguration `yaml:"static"`
 
 	// The configuration for hashing
 	HashingConfiguration HashingConfiguration `yaml:"hashing"`
+}
+
+// StaticConfiguration is used for running M3DB with a static config
+type StaticConfiguration struct {
+	TopologyConfig *topology.StaticConfiguration `yaml:"topology"`
 }
 
 // TickConfiguration is the tick configuration for background processing of
--- a/services/m3dbnode/config/config.go
+++ b/services/m3dbnode/config/config.go
@@ -25,6 +25,7 @@ import (
 
 	etcdclient "github.com/m3db/m3cluster/client/etcd"
 	"github.com/m3db/m3db/client"
+	"github.com/m3db/m3db/topology"
 	"github.com/m3db/m3x/config/hostid"
 	"github.com/m3db/m3x/instrument"
 	xlog "github.com/m3db/m3x/log"
@@ -92,8 +93,11 @@ type Configuration struct {
 	// The pooling policy.
 	PoolingPolicy PoolingPolicy `yaml:"poolingPolicy"`
 
-	// The configuration for config service client.
-	ConfigService etcdclient.Configuration `yaml:"configService"`
+	// The dynamic etcd configuration for config service client.
+	ConfigService *etcdclient.Configuration `yaml:"configService"`
+
+	// The static configuration for config service client.
+	StaticConfigService *topology.StaticConfiguration `yaml:"staticConfigService"`
 
 	// The configuration for hashing
 	HashingConfiguration HashingConfiguration `yaml:"hashing"`
--- a/services/m3dbnode/config/config.go
+++ b/services/m3dbnode/config/config.go
@@ -105,7 +105,34 @@ type Configuration struct {
 
 // StaticConfiguration is used for running M3DB with a static config
 type StaticConfiguration struct {
-	TopologyConfig *topology.StaticConfiguration `yaml:"topology"`
+	Namespaces     []StaticNamespaceConfiguration `yaml:"namespaces"`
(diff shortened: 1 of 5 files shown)


services/m3dbnode/config/config.go
```diff
@@ -92,8 +93,11 @@ type Configuration struct {
 	// The pooling policy.
 	PoolingPolicy PoolingPolicy `yaml:"poolingPolicy"`
 
-	// The configuration for config service client.
-	ConfigService etcdclient.Configuration `yaml:"configService"`
+	// The dynamic etcd configuration for config service client.
+	ConfigService *etcdclient.Configuration `yaml:"configService"`
+
+	// The static configuration for config service client.
+	StaticConfigService *topology.StaticConfiguration `yaml:"staticConfigService"`
```
robskillington: So I think we should leave room for other static configuration in the future (not purely topology based).

Might be better to create a new struct here for that purpose, also "static" is more terse than "staticConfigService" and better describes the data described.

This could work well:

```go
StaticConfig *StaticConfiguration `yaml:"static"`
```

services/m3dbnode/server/server.go
```diff
@@ -226,38 +229,76 @@ func Run(runOpts RunOptions) {
...
+		shardSet, hostShardSets, err := createStaticShardSet(cfg.StaticConfig.TopologyConfig.Shards, cfg.ListenAddress)
+		if err != nil {
+			logger.Fatalf("unable to create shard set for static config: %v", err)
+		}
+		staticOptions := topology.NewStaticOptions().
+			SetReplicas(1).
+			SetHostShardSets(hostShardSets).
+			SetShardSet(shardSet)
+
+		md, err := createDefaultMetaData("metrics")
```
robskillington: I think this should be in config, in fact it should probably be able to have multiple namespaces as well.  

We can expand `StaticConfiguration` in the config package to be something like:

services/m3dbnode/server/server.go
```diff
@@ -226,38 +234,80 @@ func Run(runOpts RunOptions) {
...
+			nsList = append(nsList, md)
+		}
+		nsInitStatic := namespace.NewStaticInitializer(nsList)
+		topoInit = topology.NewStaticInitializer(staticOptions)
+		opts = opts.SetNamespaceInitializer(nsInitStatic)
+
+		kv = m3clusterkvmem.NewStore()
+
+	default:
+		logger.Fatal("configService or staticConfigService required")
```
robskillington: This looks outdated, should be `config service or static configuration required` perhaps?

services/m3dbnode/server/server.go
```diff
@@ -790,3 +835,60 @@ func capacityPoolOptions(
...
+	}
+
+	host := topology.NewHost("localhost", listenAddress)
+	hostShardSet := topology.NewHostShardSet(host, shardSet)
+	hostShardSets = append(hostShardSets, hostShardSet)
+
+	return shardSet, hostShardSets, nil
+}
+
+func createMetaData(cfg config.StaticNamespaceConfiguration) (namespace.Metadata, error) {
```
robskillington: nit: "metadata" is one word so this should be `createMetadata` also probably better to be specific and mention the word "namespace" in the method name.  Another thing is by default all our "constructor-like" functions are prefixed with "new" so for consistency we should probably apply the same rule here. 
 `newNamespaceMetadata(...)` seems to incorporate all that as a name.

services/m3dbnode/server/server.go
```diff
@@ -790,3 +835,60 @@ func capacityPoolOptions(
 	}
 	return opts
 }
+
+func createStaticShardSet(numShards int, listenAddress string) (sharding.ShardSet, []topology.HostShardSet, error) {
```
robskillington: nit: To keep consistent with our constructor-like methods being named with "new" probably should rename this to `newStaticShardSet(...)`.

services/m3dbnode/server/server.go
```diff
@@ -790,3 +835,60 @@ func capacityPoolOptions(
...
+		namespace.NewOptions().
+			SetNeedsBootstrap(cfg.Options.NeedsBootstrap).
+			SetNeedsFilesetCleanup(cfg.Options.NeedsFilesetCleanup).
+			SetNeedsFlush(cfg.Options.NeedsFlush).
+			SetNeedsRepair(cfg.Options.NeedsRepair).
+			SetWritesToCommitLog(cfg.Options.WritesToCommitLog).
+			SetRetentionOptions(
+				retention.NewOptions().
+					SetBlockSize(cfg.Retention.BlockSize).
+					SetRetentionPeriod(cfg.Retention.RetentionPeriod)))
```
robskillington: Missing setting the following retention options from the config:
```go
	BufferFuture                        time.Duration `yaml:"bufferFuture"`
	BufferPast                          time.Duration `yaml:"bufferPast"`
	BlockDataExpiry                     bool          `yaml:"blockDataExpiry"`
	BlockDataExpiryAfterNotAccessPeriod time.Duration `yaml:"blockDataExpiryAfterNotAccessPeriod"`
```

PR #424, opened 2018-01-04
Fix mmap functions for linux by replacing MADV_HUGEPAGES with MAP_HUGETLB

richardartoul:


persist/fs/mmap.go
```diff
@@ -41,64 +41,79 @@ type mmapOptions struct {
...
 	// that support it
 	enabled bool
 	// threshold determines if the size being mmap'd is greater or equal
-	// to this value to use or not use transparent huge pages if enabled
+	// to this value to use or not use the huge TLB flag if enabled
 	threshold int64
 }
 
-func mmapFiles(opener fileOpener, files map[string]mmapFileDesc) error {
+func mmapFiles(opener fileOpener, files map[string]mmapFileDesc) (error, error) { // nolint: golint
```
robskillington: I think we should create another struct here, return two untype errors is quite weird.  Also would rather not escape the linter if we don't have to of course.

i.e. 

```go
type mmapFilesResult struct {
  warning error
}

func mmapFiles(...) (mmapFilesResult, error) { 
  ...
}
```

persist/fs/mmap_linux.go
```diff
@@ -56,22 +56,42 @@ func mmap(fd, offset, length int64, flags int, opts mmapOptions) ([]byte, error)
...
+		b, err = syscall.Mmap(int(fd), offset, int(length), prot, flagsWithoutHugeTLB)
+		// If it still failed, then return an error
+		if err != nil {
+			return mmapResult{}, fmt.Errorf("mmap error: %v", err)
+		}
+
+		// Otherwise, return success but with an appropriate warning that we were
+		// not able to enable the HUGETLB flag.
+		// TODO:
+		return mmapResult{result: b}, nil
```
robskillington: You can get rid of line 80 - 87 here as the same code sits below.

i.e.
```go
var (b []byte
if err != nil && shouldUseHugeTLB {
  b, err = syscall.Mmap(int(fd), offset, int(length), prot, flagsWithoutHugeTLB)
}

```

richardartoul: oh duh, thanks!

persist/fs/mmap_linux.go
```diff
@@ -56,22 +56,48 @@ func mmap(fd, offset, length int64, flags int, opts mmapOptions) ([]byte, error)
...
+	// and mmap without the hugeTLB flag.
+	if err != nil && shouldUseHugeTLB {
+		b, err = syscall.Mmap(int(fd), offset, int(length), prot, flagsWithoutHugeTLB)
+		// If we succeeded the second time, then proceed but make sure the caller
+		// receives a warning that includes the error from when we tried to use the
+		// hugeTLB flag.
+		if err == nil {
+			warning = fmt.Errorf(
+				"error while trying to mmap with hugeTLB flag: %s, hugeTLB disabled", withTLBErr.Error())
+		}
```
robskillington: nit: You could reorder this to save the need for `withTLBErr`:
```go
b, err = syscall.Mmap(int(fd), offset, int(length), prot, flags)
if err != nil && shouldUseHugeTLB {
  warning = err
  b, err = syscall.Mmap(int(fd), offset, int(length), prot, flagsWithoutHugeTLB)
}
```

persist/fs/mmap_linux.go
```diff
@@ -56,22 +56,48 @@ func mmap(fd, offset, length int64, flags int, opts mmapOptions) ([]byte, error)
...
 
-	// Reduce the number of pagefaults when scanning through large files
-	if err := syscall.Madvise(b, syscall.MADV_HUGEPAGE); err != nil {
-		return nil, fmt.Errorf("madvise error, check platform support: %v", err)
+	if err != nil {
+		return mmapResult{}, fmt.Errorf("mmap error: %v", err)
 	}
 
-	return b, nil
+	return mmapResult{result: b}, nil
```
robskillington: Missing actually setting the warning here, should this be `mmapResult{result: b, warning: warning}`?

richardartoul: yep, fixed

PR #425, opened 2018-01-08
Improve handling of duplicate metadata

richardartoul:


client/session.go
```diff
@@ -1977,29 +2001,33 @@ func (s *session) emitDuplicateMetadataLog(received *receivedBlocks, metadata bl
 		"Received metadata, but peer metadata has already been submitted")
 }
 
+// emitDuplicateMetadataLogV2 emits a log with the details of the duplicate metadata
+// event. Note that we're unable to log the blocks themselves because they're contained
+// in a slice that is not safe for concurrent access (I.E logging them here would be
+// racey because other code could be modifying the slice)
 func (s *session) emitDuplicateMetadataLogV2(received *receivedBlocks, metadata blocksMetadata) {
 	fields := make([]xlog.Field, len(received.results)+1)
```
robskillington: This should be `make([]xlog.Field, 0, len(received.results)+1)` or avoid using append, although probably better to keep using append.

robskillington: Need to make the same change to `emitDuplicateMetadataLog(...)` fn.

richardartoul: yeah forgot to propagate that fix from other P.R's. Will update

richardartoul: @robskillington Does the V1 endpoint have the same issue with duplicate metadata?

robskillington: Nope just the fact that it also initializes the slice to a non-zero length.

PR #426, opened 2018-01-08
Make Mmap files handle mmap warning appropriately

richardartoul:
Should not break in warning case, otherwise file descriptor pointer doesn't get updated

### DIFF
--- a/persist/fs/mmap.go
+++ b/persist/fs/mmap.go
@@ -27,8 +27,6 @@ import (
 	xerrors "github.com/m3db/m3x/errors"
 )
 
-type mmapFdFuncType func(fd, offset, length int64, opts mmapOptions) (mmapResult, error)
-
 // Package-level global for easy mocking
 var mmapFdFunc = mmapFd
 
--- a/persist/fs/mmap.go
+++ b/persist/fs/mmap.go
@@ -29,6 +29,7 @@ import (
 
 type mmapFdFuncType func(fd, offset, length int64, opts mmapOptions) (mmapResult, error)
 
+// Package-level global for easy mocking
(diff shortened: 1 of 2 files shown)


persist/fs/mmap.go
```diff
@@ -81,7 +81,6 @@ func mmapFiles(opener fileOpener, files map[string]mmapFileDesc) (mmapFilesResul
 		}
 		if result.warning != nil {
 			multiWarn = multiWarn.Add(errorWithFilename(filePath, result.warning))
-			break
```
robskillington: Ha, interesting.  Maybe we should add a unit test for this?

We can use some package level indirection to mock this out.

e.g. 
```go
var mmapFileFn = mmapFile

// ...
func TestMmapFiles(t *testing.T) {
  currMmapFileFn := mmapFileFn
  mmapFileFn = func(..) {
    // mock..
  }
  defer func() {
    mmapFileFn = currMmapFileFn
  }()
}
```

richardartoul: I try and avoid those if I can cause they break parallel tests. I'll add a unit test though

robskillington: We only run package tests in parallel, but sure, you can probably avoid it if you'd like with:

persist/fs/mmap.go
```diff
@@ -27,6 +27,9 @@ import (
 	xerrors "github.com/m3db/m3x/errors"
 )
 
+// Package-level global for easy mocking
+var mmapFdFunc = mmapFd
```
robskillington: nit: For consistency make this `mmapFdFn`, we use `fn` as a suffix everywhere vs `func`

PR #428, opened 2018-01-08
Add ability to specify static configuration

benraskin92:


services/m3dbnode/config/config.go
```diff
@@ -93,48 +92,13 @@ type Configuration struct {
 	// The pooling policy.
 	PoolingPolicy PoolingPolicy `yaml:"poolingPolicy"`
 
-	// The dynamic etcd configuration for config service client.
-	ConfigService *etcdclient.Configuration `yaml:"configService"`
-
-	// The static configuration for config service client.
-	StaticConfig *StaticConfiguration `yaml:"static"`
+	// The environment (static or dynamic) configuration.
+	EnvironmentConfig *environment.Configuration `yaml:"config"`
```
robskillington: You could make this non pointer type to avoid having to check for nil-ability of this.

client/config.go
```diff
@@ -78,8 +77,11 @@ type Configuration struct {
 	// time to use when sleeping between a failed health check and the next check.
 	BackgroundHealthCheckFailThrottleFactor float64 `yaml:"backgroundHealthCheckFailThrottleFactor" validate:"min=0,max=10"`
 
-	// HashingConfiguration is the configuration for hashing of IDs to shards
+	// HashingConfiguration is the configuration for hashing of IDs to shards.
 	HashingConfiguration HashingConfiguration `yaml:"hashing"`
+
+	// The host and port on which m3db listens.
+	ListenAddress string `yaml:"listenAddress"`
```
robskillington: Did you need to add this?

benraskin92: Yes, because `NewTopologyInitializer()` for the static config needs a listen address

environment/config.go
```diff
@@ -0,0 +1,231 @@
...
+		nsInitStatic := namespace.NewStaticInitializer(nsList)
+		return nsInitStatic, nil
+
+	default:
+		return nil, errMissingNamespaceConfig
+	}
+}
+
+// NewTopologyInitializer creates a new topology initializer
+func (c Configuration) NewTopologyInitializer(
```
robskillington: So on thinking this through a fair bit, we should probably just make this one method yeah.  

With respect to returning things callers don't need: the namespace initializer isn't expensive, it's just an initializer.  It doesn't acquire any resources to create.

services/m3dbnode/server/server.go
```diff
@@ -33,14 +32,12 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/m3db/m3db/environment"
+
```
robskillington: nit: No space between this new import and the existing ones.

client/config.go
```diff
@@ -148,32 +150,39 @@ func (c Configuration) NewAdminClient(
 	writeRequestScope := iopts.MetricsScope().SubScope("write-req")
 	fetchRequestScope := iopts.MetricsScope().SubScope("fetch-req")
 
-	topoInit := params.TopologyInitializer
-	if topoInit == nil {
-		if c.ConfigService == nil {
-			return nil, errConfigurationMustSupplyConfigService
-		}
+	configureResults := &environment.ConfigureResults{
```
robskillington: nit: Call this `envCfg` so that when reading the result it seems pragmatic, i.e. `envCfg.NamespaceInitializer`

client/config.go
```diff
@@ -148,32 +150,39 @@ func (c Configuration) NewAdminClient(
...
+
+			configureResults, err = c.EnvironmentConfig.NewConfigureResults(configSvcClient, iopts, "", c.HashingConfiguration.Seed)
+			if err != nil {
+				err = fmt.Errorf("unable to create dynamic topology initializer, err: %v", err)
+				return nil, err
+			}
+
+		case c.EnvironmentConfig.Static != nil:
+			var err error
+			configureResults, err = c.EnvironmentConfig.NewConfigureResults(nil, nil, c.ListenAddress, 0)
```
robskillington: Yeah let's move `ListenAddress` into `environment.StaticConfiguration` as it's pretty strange to specify it here.

environment/config.go
```diff
@@ -0,0 +1,215 @@
...
+
+// ConfigureResults stores initializers and kv store for dynamic and static configs
+type ConfigureResults struct {
+	NamespaceInitializer namespace.Initializer
+	TopologyInitializer  topology.Initializer
+	KVStore              kv.Store
+}
+
+// NewConfigureResults creates a new ConfigureResults
+func (c Configuration) NewConfigureResults(configSvcClient client.Client,
```
robskillington: Let's call this just `Configure(...)` because it's doing more than just new'ing up a few things.  

Also calling something `NewConfigureResults(...)` is weird, it's kind of like calling a method that returns say a type `UserResults` as `NewUserResults() UserResults` instead of just `Users() UserResults`.

client/config.go
```diff
@@ -148,32 +147,36 @@ func (c Configuration) NewAdminClient(
...
-				SetEnvironment(c.ConfigService.Env).
-				SetZone(c.ConfigService.Zone)).
-			SetQueryOptions(services.NewQueryOptions().SetIncludeUnhealthy(true)).
-			SetInstrumentOptions(iopts).
-			SetHashGen(sharding.NewHashGenWithSeed(c.HashingConfiguration.Seed))
-
-		topoInit = topology.NewDynamicInitializer(initOpts)
+		case c.EnvironmentConfig.Static != nil:
+
+			envCfg, err = c.EnvironmentConfig.Configure(nil, c.HashingConfiguration.Seed)
```
robskillington: Is there a reason `iopts` isn't passed here?

environment/config.go
```diff
@@ -0,0 +1,221 @@
+package environment
```
robskillington: Missing the copyright at top of this file.

environment/config.go
```diff
@@ -0,0 +1,221 @@
...
+
+// ConfigureResults stores initializers and kv store for dynamic and static configs
+type ConfigureResults struct {
+	NamespaceInitializer namespace.Initializer
+	TopologyInitializer  topology.Initializer
+	KVStore              kv.Store
+}
+
+// Configure creates a new ConfigureResults
+func (c Configuration) Configure(iopts instrument.Options, hashingSeed uint32) (ConfigureResults, error) {
```
robskillington: Can we make the argument to this method a struct so if it changes it's not a breaking change to callers in the future?  i.e.

PR #429, opened 2018-01-09
Replace scanning through dataFD (using seek) with an mmap

richardartoul:
Instead of scanning through data files in the seeker using a file descriptor and the Seek() syscall, we can use mmap instead.

### DIFF
--- a/persist/fs/seek.go
+++ b/persist/fs/seek.go
@@ -182,7 +182,6 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
 	}
 
 	// Mmap necessary files
-	// TODO: Handle hugeTLB
 	mmapResult, err := mmapFiles(os.Open, map[string]mmapFileDesc{
 		filesetPathFromTime(shardDir, blockStart, indexFileSuffix): mmapFileDesc{
 			file:  &indexFd,
--- a/persist/fs/seek.go
+++ b/persist/fs/seek.go
@@ -64,9 +64,9 @@ type seeker struct {
 	summariesFdWithDigest      digest.FdWithDigestReader
 	digestFdWithDigestContents digest.FdWithDigestContentsReader
 
-	dataFd     *os.File
 	dataReader *bufio.Reader
 
+	dataMmap  []byte
 	indexMmap []byte
 
 	// Expected digests for each file read from the digests file
@@ -168,6 +168,8 @@ func (s *seeker) ConcurrentIDBloomFilter() *ManagedConcurrentBloomFilter {
 func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error {
 	shardDir := ShardDirPath(s.filePathPrefix, namespace, shard)
 	var infoFd, indexFd, dataFd, digestFd, bloomFilterFd, summariesFd *os.File
+
+	// Open necessary files
 	if err := openFiles(os.Open, map[string]**os.File{
 		filesetPathFromTime(shardDir, blockStart, infoFileSuffix):        &infoFd,
 		filesetPathFromTime(shardDir, blockStart, indexFileSuffix):       &indexFd,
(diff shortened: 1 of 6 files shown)


persist/fs/seek.go
```diff
@@ -320,11 +340,7 @@ func (s *seeker) SeekByID(id ts.ID) (checked.Bytes, error) {
 // instead of looking it up on its own. Useful in cases where you've already
 // obtained an entry and don't want to waste resources looking it up again.
 func (s *seeker) SeekByIndexEntry(entry IndexEntry) (checked.Bytes, error) {
-	_, err := s.dataFd.Seek(entry.Offset, 0)
-	if err != nil {
-		return nil, err
-	}
-	s.dataReader.Reset(s.dataFd)
+	s.dataReader.Reset(bytes.NewReader(s.dataMmap[entry.Offset:]))
```
robskillington: We should definitely cache the `*bytes.Reader`, perhaps create one in the seeker ctor?

persist/fs/seek.go
```diff
@@ -179,6 +181,37 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
...
+		hugeTLB: mmapHugeTLBOptions{
+			enabled:   s.opts.MmapEnableHugeTLB(),
+			threshold: s.opts.MmapHugeTLBThreshold(),
+		},
+	}
+	mmapResult, err := mmapFiles(os.Open, map[string]mmapFileDesc{
+		filesetPathFromTime(shardDir, blockStart, indexFileSuffix): mmapFileDesc{
+			file:  &indexFd,
+			bytes: &s.indexMmap,
+			// Index files are small enough that they don't warrant hugeTLB
```
robskillington: This comment no longer relevant no?

persist/fs/seek.go
```diff
@@ -64,9 +64,9 @@ type seeker struct {
 	summariesFdWithDigest      digest.FdWithDigestReader
 	digestFdWithDigestContents digest.FdWithDigestContentsReader
 
-	dataFd     *os.File
 	dataReader *bufio.Reader
```
robskillington: bufio.Reader no longer necessary with the mmap yeah?  Can just make this a `*bytes.Reader`.

persist/fs/seek.go
robskillington: As discussed, we can go without any readers for the data mmap thankfully (yay)

persist/fs/seek.go
```diff
@@ -320,63 +339,53 @@ func (s *seeker) SeekByID(id ts.ID) (checked.Bytes, error) {
...
-	if err != nil {
-		return nil, err
-	} else if n != cap(s.prologue) {
+	// We'll treat the dataView similar to a reader interface in that we will continue to reslice
+	// it such that the first byte is always the next byte, and we'll focus instead on how much
+	// we want to read.
+	dataView := s.dataMmap[entry.Offset:]
+
+	// Should never happen, but prevents panics in the case of malformed data
+	if len(dataView) < cap(s.prologue) {
```
robskillington: Hm, do we need `s.prologue` any more?  Maybe we can remove this field?

persist/fs/seek.go
```diff
@@ -320,63 +339,53 @@ func (s *seeker) SeekByID(id ts.ID) (checked.Bytes, error) {
...
-	s.dataReader.Reset(s.dataFd)
 
-	n, err := s.dataReader.Read(s.prologue)
-	if err != nil {
-		return nil, err
-	} else if n != cap(s.prologue) {
+	// We'll treat the dataView similar to a reader interface in that we will continue to reslice
+	// it such that the first byte is always the next byte, and we'll focus instead on how much
+	// we want to read.
+	dataView := s.dataMmap[entry.Offset:]
```
robskillington: minor nit: Maybe just `data`?  For terse-ness.

persist/fs/seek.go
```diff
@@ -320,63 +339,53 @@ func (s *seeker) SeekByID(id ts.ID) (checked.Bytes, error) {
...
-		}
-		n += remainder
-	}
-
-	if n != int(entry.Size) {
-		return nil, errReadNotExpectedSize
-	}
+	// Copy the actual data into the underlying buffer
+	underlyingBuf := buffer.Get()
+	copy(underlyingBuf, dataView[:entry.Size])
```
robskillington: Should probably check `len(data) >= entry.Size` first yeah?

robskillington: Alternatively you can check this once before comparing against the marker, e.g. `if len(data) < markerLen + entry.Size { return nil, fmt.Errorf("not long enough...") }`

persist/fs/seek.go
```diff
@@ -179,6 +179,37 @@ func (s *seeker) Open(namespace ts.ID, shard uint32, blockStart time.Time) error
...
+	if err != nil {
+		s.Close()
+		return err
+	}
+	if warning := mmapResult.warning; warning != nil {
+		s.opts.InstrumentOptions().Logger().Warnf(
+			"warning while mmaping files in seeker: %s", warning.Error())
+	}
+
+	// Setup digest readers
```
robskillington: Perhaps move the `...Reset()` calls and the `defer func() { ...Close()  }` stanza's above the `mmapFiles` call because it could return `if err != nil { ... }` from the `mmapFiles` call?

richardartoul: good call

persist/fs/seek.go
richardartoul: is it 

```golang
len(data) < markerLen
```

or

```golang
len(data) < markerLen + idxLen
```

PR #431, opened 2018-01-11
Refactor Retriever/SeekerManager to not duplicate Seeker's based on the fetch concurrency

richardartoul:
Currently, the BlockRetriever will create a SeekerManager per fetchLoop / fetch concurrency. This results in duplication of the underlying indexLookup and Mmaps in the seekers themselves, wasting memory. In addition, we want to perform the index digest in the Open() method of the seeker, but if we added it with the existing code we would calculate the digest once for each fetch concurrency which would waste a lot of time.

This P.R refactors the code so that the BlockRetriever still has multiple independent fetch loops, but only a single SeekerManager. The SeekerManager is also concurrency aware and creates one seeker per shard/blockStart, but then "clones" it such that multiple seekers for the same shard/blockStart can seek independently, while sharing the same underlying resources and pre-calculated index digest.

- [X] Refactor nearestIndexOffsetLookup to support concurrentClone()
- [X] Refactor Seeker to support ConcurrentClone()
- [X] Refactor SeekerManager to support arbitrary fetch concurrency
- [X] Refactor Retriever to use a single SeekerManager
- [X] Improve Test coverage significantly for SeekerManager
- [X] Improve lifecycle management ( Close() ) for Retriever and SeekerManager so that they don't leak resources on shutdown
- [X] Calculate index file digest on seeker open (now that we're not doing it twice)


persist/fs/index_lookup_test.go
```diff
@@ -77,6 +83,137 @@ func TestNewNearestIndexOffsetDetectsUnsortedFiles(t *testing.T) {
...
+	for _, clone := range clones {
+		require.NoError(t, clone.close())
+	}
+}
+
+func newIndexLookupWithValidData(t *testing.T, indexSummaries []schema.IndexSummary) *nearestIndexOffsetLookup {
+	// Create a temp file
+	file, err := ioutil.TempFile("", "index-lookup-sort")
+	require.NoError(t, err)
+	defer os.Remove(file.Name())
```
robskillington: Ha this works?  I assume its because the file descriptor is still active with the mmap, but it's funny that we delete the file but return a data-structure here that still has a ref to it (with the mmap) from the deleted file.

Probably fine to keep it this way, I just hope it's not OS dependent.

persist/fs/index_lookup_test.go
```diff
@@ -77,6 +83,137 @@ func TestNewNearestIndexOffsetDetectsUnsortedFiles(t *testing.T) {
...
+
+	// Write out the out-of-order summaries into the temp file
```
robskillington: This is not always used for out of order tests though yeah?

persist/fs/index_lookup_test.go
```diff
@@ -99,3 +236,19 @@ func calculateExpectedDigest(t *testing.T, digestReader digest.FdWithDigestReade
...
+}
+
+func (s sortableSummaries) Less(i, j int) bool {
+	return bytes.Compare(s[i].ID, s[j].ID) < 0
+}
+
+func (s sortableSummaries) Swap(i, j int) {
+	temp := s[i]
+	s[i] = s[j]
+	s[j] = temp
```
robskillington: This can just be `s[i], s[j] = s[j], s[i]` btw.

persist/fs/retriever.go
```diff
@@ -58,6 +60,7 @@ type blockRetrieverStatus int
 type newSeekerMgrFn func(
 	bytesPool pool.CheckedBytesPool,
 	opts Options,
+	fetchConcurrency int,
```
robskillington: Does this need to be passed in?  Can't the seeker mgr just read `opts.FetchConcurrency()`?

persist/fs/retriever.go
```diff
@@ -101,6 +106,10 @@ func NewBlockRetriever(
 		bytesPool:      opts.BytesPool(),
 		status:         blockRetrieverNotOpen,
 		notifyFetch:    make(chan struct{}, 1),
+		// We just close this channel when the fetchLoops should shutdown, so no
+		// buffering is required
+		fetchLoopsShouldShutdownCh: make(chan struct{}),
+		fetchLoopsHaveShutdownCh:   make(chan struct{}, opts.FetchConcurrency()),
```
robskillington: If you're just waiting for this channel to be shut then it doesn't need to be of a specific size.

persist/fs/retriever.go
```diff
@@ -313,6 +315,13 @@ func (r *blockRetriever) fetchBatch(
 
 		req.onRetrieved(seg)
 	}
+
+	err = seekerMgr.Return(shard, blockStart, seeker)
+	if err != nil {
+		r.fsOpts.InstrumentOptions().Logger().Errorf(
```
robskillington: Maybe put the logger on the `blockRetriever` struct so you can just `r.logger.Errorf(..)`?  There's only one per namespace so don't need to worry about saving space.

persist/fs/retriever.go
```diff
@@ -313,6 +315,13 @@ func (r *blockRetriever) fetchBatch(
...
+			"err returning seeker for shard: %d, and blockStart: %d, err: %s",
+			shard, blockStart.Unix(), err.Error())
```
robskillington: nit: Maybe use `WithFields(...)` here?  Having a lot of vars in the fmt string doesn't seem fantastic.

persist/fs/retriever.go
```diff
@@ -423,12 +427,17 @@ func (r *blockRetriever) shardRequests(
 
 func (r *blockRetriever) Close() error {
 	r.Lock()
-	defer r.Unlock()
-
 	r.nsMetadata = nil
 	r.status = blockRetrieverClosed
```
robskillington: Perhaps test if `if r.status == blockRetrieverClosed { return errBlockRetrieverAlreadyClosed }`?  Especially now that `close(...)` will cause a panic now if this is called twice.

persist/fs/retriever.go
```diff
@@ -423,12 +427,17 @@ func (r *blockRetriever) shardRequests(
...
+	r.Unlock()
 
-	return nil
+	close(r.fetchLoopsShouldShutdownCh)
+	for i := 0; i < r.opts.FetchConcurrency(); i++ {
+		<-r.fetchLoopsHaveShutdownCh
+	}
+
+	err := r.seekerMgr.Close()
+	return err
```
robskillington: nit: Just `return r.seekerMgr.Close()` here?

persist/fs/retriever_test.go
```diff
@@ -92,7 +93,26 @@ type streamResult struct {
...
+// with high concurrent seekers and calls cache shard indices at the beginning.
+// This means that the seekers will be opened all at once in the beginning and
+// by the time ConcurrentIDBloomFilter() is called, they seekers will already be
+// open.
+func TestBlockRetrieverHighConcurrentSeeksCacheShardIndices(t *testing.T) {
+	testBlockRetrieverHighConcurrentSeeks(t, true)
+}
+
+func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices bool) {
+	defer leaktest.CheckTimeout(t, 2*time.Minute)()
```
robskillington: Hm 2 minutes seems like a long time, any reason not to just use the default?

persist/fs/seek_manager.go
```diff
@@ -137,75 +170,198 @@ func (m *seekerManager) CacheShardIndices(shards []uint32) error {
...
+	byTime := m.seekersByTime(shard)
+
+	byTime.Lock()
+	// Track accessed to precache in open/close loop
+	byTime.accessed = true
+
+	startNano := xtime.ToUnixNano(start)
+	seekersAndBloom, err := m.getOrOpenSeekersWithLock(startNano, byTime)
+	if err != nil {
+		return nil, err
```
robskillington: Hm, need to `byTime.Unlock()` here yeah?  Maybe we could pay the defer cost and `defer byTime.Unlock()` after locking here.

persist/fs/seek_manager.go
```diff
@@ -137,75 +170,198 @@ func (m *seekerManager) CacheShardIndices(shards []uint32) error {
...
+	// Should not occur in the case of a well-behaved caller
+	if availableSeekerIdx == -1 {
 		byTime.Unlock()
+		return nil, errNoAvailableSeekers
 	}
 
-	return multiErr.FinalError()
+	availableSeeker.isBorrowed = true
+	seekers[availableSeekerIdx] = availableSeeker
+	byTime.seekers[startNano] = seekersAndBloom
```
robskillington: Is this required to be set here?  It seems its not modified/changed since it was returned from ` m.getOrOpenSeekersWithLock(...)`.  We modify an entry in `seekers` but that should be copy into the existing slice.

persist/fs/seek_manager.go
```diff
@@ -137,75 +170,198 @@ func (m *seekerManager) CacheShardIndices(shards []uint32) error {
...
+			found = true
+			compareSeeker.isBorrowed = false
+			seekersAndBloom.seekers[i] = compareSeeker
+			break
+		}
 	}
+	// Should never happen with a well behaved caller. Either they are trying to return a seeker
+	// that we're not managing, or they provided the wrong shard/start.
+	if !found {
+		return errReturnedUnmanagedSeeker
```
robskillington: Need Unlock here to yeah?  Maybe we should use defer in this fn too.

persist/fs/seek_manager.go
```diff
@@ -137,75 +170,198 @@ func (m *seekerManager) CacheShardIndices(shards []uint32) error {
...
+	}
+
+	byTime.seekers[startNano] = seekersAndBloom
```
robskillington: Same here, you modify the entry in place of the `seekers` slice so shouldn't need to set `seekersAndBloom` back again.

persist/fs/seek_manager.go
```diff
@@ -137,75 +170,198 @@ func (m *seekerManager) CacheShardIndices(shards []uint32) error {
...
+	seekers, ok := byTime.seekers[start]
+	if ok && seekers.wg == nil {
+		// Seekers are already open
+		return seekers, nil
+	}
+
+	if seekers.wg != nil {
+		// Seekers are being initialized / opened, wait for the that to complete
+		byTime.Unlock()
+		seekers.wg.Wait()
```
robskillington: Seems racy to perform `seekers.wg.Wait()` outside of lock as it may be nil'd out between the Unlock and the wait call no?

robskillington: Hm, actually seekers is just a value struct not a ref here?  Maybe it's safe in that case.

persist/fs/seek_manager_test.go
```diff
@@ -21,14 +21,22 @@
...
 
+	xtime "github.com/m3db/m3x/time"
+
+	"github.com/fortytw2/leaktest"
+	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 )
 
 func TestSeekerManagerCacheShardIndices(t *testing.T) {
+	defer leaktest.CheckTimeout(t, 1*time.Minute)()
```
robskillington: Any reason why 1minute here too?

persist/fs/seek_manager_test.go
```diff
@@ -58,3 +66,214 @@ func TestSeekerManagerCacheShardIndices(t *testing.T) {
...
+			}
+		}
+	}()
+
+	// Restore previous interval once the openCloseLoop ends
+	require.NoError(t, m.Close())
+	// Make sure there are no goroutines still trying to write into the tickCh
+	// to prevent the test itself from interfering with the goroutine leak test
+	close(cleanupCh)
+	seekManagerCloseInterval = oldInterval
```
robskillington: Might be worth doing the restore of this in a defer at the top so at least that logic sticks together (and one doesn't get deleted without the other).

persist/fs/index_lookup_test.go
richardartoul: you're right, I didn't realize what I was doing at the time haha. Cool that it works though

persist/fs/index_lookup_test.go
richardartoul: yeah, I mean really since you're providing the writes as an argument validData shouldn't even be in the name. I'll fix it

persist/fs/index_lookup_test.go
richardartoul: oh nice thanks

persist/fs/retriever.go
richardartoul: can do

persist/fs/retriever.go
richardartoul: yeah fair

persist/fs/retriever.go
richardartoul: good idea

persist/fs/retriever.go
richardartoul: sure

persist/fs/retriever_test.go
richardartoul: it times out in travis haha. I think the primary culprit is the race detector making it slow.

richardartoul: the checkTimeout start at the beginning of the test unfortunately, not when the test is complete.

richardartoul: I could probably move it to a separate file and add the "big" tag or whatever, but doesn't seem worth it right now

persist/fs/seek_manager.go
richardartoul: yeah you're right good catch, I'll use defer

persist/fs/seek_manager.go
richardartoul: you're right its not, I'll remove it

persist/fs/seek_manager.go
richardartoul: nice catch, yeah I'll convert to defer

persist/fs/seek_manager.go
richardartoul: nice, will remove

persist/fs/seek_manager.go
richardartoul: yeah you had me worried for a second, but because its a value struct it should be fine. When the .wg field is set to nil, our's will still point to the wg and when we call .Wait() on it we should just proceed immediately

persist/fs/seek_manager.go
```diff
@@ -282,11 +438,33 @@ func (m *seekerManager) seekersByTime(shard uint32) *seekersByTime {
 
 func (m *seekerManager) Close() error {
 	m.Lock()
-	defer m.Unlock()
+
```
richardartoul: based on previosu comments, I'm gonna conver to defer m.Unlock() here

richardartoul: actually nevermind, you end up holding the lock for way too long because of the openCloseLoopDoneCh

persist/fs/seek_manager_test.go
richardartoul: same reason as before, travis + race detector is sloooooow and timeout starts from beginning of test not the end

persist/fs/seek_manager_test.go
richardartoul: I actually think this is completely unnecessary now that I'm using a custom sleepFn, just gonna make it a constant again

persist/fs/retriever.go
```diff
@@ -313,6 +319,15 @@ func (r *blockRetriever) fetchBatch(
 
 		req.onRetrieved(seg)
 	}
+
+	err = seekerMgr.Return(shard, blockStart, seeker)
+	if err != nil {
+		r.logger.WithFields(
+			log.NewField("shard", shard),
+			log.NewField("blockStart", blockStart.Unix()),
+			log.NewField("err", err),
```
robskillington: nit: Might be worth `err.Error()` here on account of this coming out usually as a pointer value when printed instead of the actual err string.

persist/fs/retriever_test.go
robskillington: Ah I see, yeah - time from start of the test is pretty unfortunate.

persist/fs/seek_manager.go
```diff
@@ -374,20 +550,39 @@ func (m *seekerManager) openCloseLoop() {
...
 		m.RUnlock()
 
 		// Close after releasing lock so any IO is done out of lock
 		for _, seeker := range closing {
-			seeker.Close()
+			err := seeker.seeker.Close()
+			if err != nil {
+				m.logger.
+					WithFields(
+						log.NewField("err", err),
```
robskillington: nit: `err.Error()` here too.

persist/fs/seek_manager.go
```diff
@@ -374,20 +550,39 @@ func (m *seekerManager) openCloseLoop() {
...
 		}
 		m.RUnlock()
 
 		// Close after releasing lock so any IO is done out of lock
 		for _, seeker := range closing {
-			seeker.Close()
+			err := seeker.seeker.Close()
+			if err != nil {
+				m.logger.
+					WithFields(
```
robskillington: super nit: Is it worth it to put WithFields on its own line?

persist/fs/seek_manager.go
```diff
@@ -396,12 +591,25 @@ func (m *seekerManager) openCloseLoop() {
...
-		for _, seeker := range byTime.seekers {
-			seeker.Close()
+		for _, seekersByTime := range byTime.seekers {
+			for _, seeker := range seekersByTime.seekers {
+				// We don't need to check if the seeker is borrowed here because we don't allow the
+				// SeekerManager to be closed if any seekers are still outstanding.
+				err := seeker.seeker.Close()
+				if err != nil {
+					m.logger.
+						WithFields(
```
robskillington: super nit: Is it worth it to put WithFields on its own line?

persist/fs/seek_manager.go
```diff
@@ -396,12 +591,25 @@ func (m *seekerManager) openCloseLoop() {
...
+							log.NewField("err", err),
```
robskillington: nit: `err.Error()` here too.

persist/fs/seek_manager_test.go
robskillington: Makes sense if from beginning of test, strange design choice there for the API... np.

PR #432, opened 2018-01-12
Add more detail to error message on server close

katezaps:


tools/dtest/harness/harness.go
```diff
@@ -475,7 +475,7 @@ func (dt *DTestHarness) newHeartbeatRouter() node.HeartbeatRouter {
 				dt.logger.Fatalf("could not create heartbeat server: %v", err)
 			}
 			// we're closing the server, which will trigger this path. we don't want to error on it
-			dt.logger.Infof("stopping heartbeatserver, err: ", err)
+			dt.logger.Infof("stopping heartbeatserver, server closed or inaccessible. err:  ", err)
```
jskelcy: Does this need to be:
`+			dt.logger.Infof("stopping heartbeatserver, server closed or inaccessible. err: %s ", err.Error())
`

PR #433, opened 2018-01-12
 Add caching of series block for the recently read cache policy

robskillington:
Remaining:
- [x] Unit tests for shard and series OnRetrieveBlock call lifecycles
- [x] Setup integration tests to run for all caching policies


storage/series/reader.go
```diff
@@ -179,11 +177,17 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
...
-	res := make([]block.FetchBlockResult, 0, len(starts))
-
-	var clonedID ts.ID
-	cachePolicy := r.opts.CachePolicy()
-
+	var (
+		// TODO(r): pool these results arrays
+		res         = make([]block.FetchBlockResult, 0, len(starts))
+		cachePolicy = r.opts.CachePolicy()
+		// NB(r): Always use nil for OnRetrieveBlock so we don't cache the
```
richardartoul: Comment makes sense to me, this method is only used during peer bootstrapping then?

robskillington: Correct, yeah fetch blocks is only for the peer bootstrapping case.

storage/series/reader.go
```diff
@@ -133,17 +137,11 @@ func (r Reader) readersWithBlocksMapAndBuffer(
...
-				}
-				// TODO(r): Specify a non-nil OnRetrieveBlock callback to cache the
-				// block back to a series in the shard on a read.
-				var onRetrieve block.OnRetrieveBlock
-				stream, err := r.retriever.Stream(clonedID, blockAt, onRetrieve)
+				// Clone ID as the block retriever uses the ID async so we cannot
+				// be sure about owner not finalizing the cloneableID passed to
+				// the Reader
+				clonedID := r.opts.IdentifierPool().Clone(r.cloneableID)
+				stream, err := r.retriever.Stream(clonedID, blockAt, r.onRetrieve)
```
richardartoul: What is the lifecycle management for the clonedID now that its not finalized by the ctx?

storage/series/reader.go
```diff
@@ -179,11 +177,17 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
...
+	var (
+		// TODO(r): pool these results arrays
+		res         = make([]block.FetchBlockResult, 0, len(starts))
+		cachePolicy = r.opts.CachePolicy()
+		// NB(r): Always use nil for OnRetrieveBlock so we don't cache the
+		// series after fetching it from disk, the fetch blocks API is called
+		// during streaming so to cache it in memory would mean we would
+		// eventually cache all series in memory when we stream results to a
+		// peer.
+		onRetrieve block.OnRetrieveBlock
```
richardartoul: This makes sense to me, its just a little confusing because its not clear that a method called FetchBlocks would only be used during peer-bootstrapping. Not sure how to improve it though

storage/series/reader.go
```diff
@@ -210,19 +214,10 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
...
-				// NB(r): Always use nil for onRetrieveBlock so we don't cache the
-				// series after fetching it from disk, the fetch blocks API is called
-				// during streaming so to cache it in memory would mean we would
-				// eventually cache all series in memory when we stream results to a
-				// peer.
-				var onRetrieve block.OnRetrieveBlock
+				// Clone ID as the block retriever uses the ID async so we cannot
+				// be sure about owner not finalizing the cloneableID passed to
+				// the Reader
+				clonedID := r.opts.IdentifierPool().Clone(r.cloneableID)
```
richardartoul: Regarding my previous question, you're able to remove the RegisterFinalizer() call because the onRetrieve method that caches the data will actually finalize it right?

richardartoul: Actually you made sure that onRetrieve was nil, so won't this leak the clonedID?

storage/series/series.go
```diff
@@ -156,44 +155,56 @@ func (s *dbSeries) updateBlocksWithLock() updateBlocksResult {
...
-						Checksum: currBlock.Checksum(),
-					}
-					currBlock.ResetRetrievable(start, retriever, metadata)
-				default:
-					// Remove the block and it will be looked up later
-					s.blocks.RemoveBlockAt(start)
-					currBlock.Close()
+		// Potentially unwire
+		var unwired, shouldUnwire bool
+		if retriever.IsBlockRetrievable(start) {
```
richardartoul: Is the IsBlockRetrievable just making sure that its been flushed to disk?

storage/series/series.go
```diff
@@ -220,6 +231,13 @@ func (s *dbSeries) IsEmpty() bool {
 	return false
 }
 
+func (s *dbSeries) NumActiveBlocks() int {
+	s.RLock()
+	value := s.blocks.Len() + s.buffer.Stats().wiredBlocks
```
richardartoul: can you refresh me on active/wired/open?

Confused as to why we would be adding these up here

storage/series/series.go
```diff
@@ -412,6 +432,35 @@ func (s *dbSeries) Bootstrap(blocks block.DatabaseSeriesBlocks) error {
...
 
+func (s *dbSeries) OnRetrieveBlock(
+	id ts.ID,
+	startTime time.Time,
+	segment ts.Segment,
+) {
+	s.Lock()
+	defer s.Unlock()
+
+	if !id.Equal(s.id) {
```
richardartoul: is just this a sanity test?

storage/series/series.go
```diff
@@ -412,6 +432,35 @@ func (s *dbSeries) Bootstrap(blocks block.DatabaseSeriesBlocks) error {
...
+		Checksum: digest.SegmentChecksum(segment),
+	}
+	b.ResetRetrievable(startTime, s.blockRetriever, metadata)
+	b.OnRetrieveBlock(id, startTime, segment)
+
+	// NB(r): Blocks retrieved have been triggered by a read, so set the last
+	// read time as now so caching policies are followed.
+	b.SetLastReadTime(s.now())
+
+	// If we retrieved this from disk then we directly emplace it
```
richardartoul: had to google emplace, thought it was a typo ha

storage/shard.go
```diff
@@ -297,8 +309,54 @@ func (s *dbShard) IsBlockRetrievable(blockStart time.Time) bool {
...
+	startTime time.Time,
+	segment ts.Segment,
+) {
+	s.RLock()
+	entry, _, err := s.lookupEntryWithLock(id)
+	if entry != nil {
+		entry.incrementReaderWriterCount()
+	}
+	s.RUnlock()
+	defer func() {
```
richardartoul: Not sure how hot this will be in the grand schema of things, but could avoid the defer with a helper function like

```golang
onRetrieveBlockCleanup(entry *dbShardEntry, id ts.ID) {
  if entry != nil {
    entry.decrementReaderWriterCount()
  }
  id.Finalize()
}
```

storage/series/reader.go
robskillington: Yeah so basically the callee (OnRetrieveBlock) now needs to finalize it, the issue with cloning it and pinning it to the lifecycle of the context is that it might not be safe to release the ID once the request finishes because the OnRetrieveBlock callee needs to use it asynchronously.

storage/series/reader.go
robskillington: Yeah good catch, for FetchBlocks it might actually make sense to continue using `ctx.RegisterFinalizer(...)` for cleaning up the IDs.

storage/series/series.go
robskillington: Correct yeah.

storage/series/series.go
robskillington: So Active == wired + unwired
Open == wired (but only mutable wired blocks)

storage/series/series.go
robskillington: Correct yeah, potentially once upon a time it was possible to list yourself as the `OnRetrieveBlock(...)` and then by the time the ref was actually used you had been recycled and were in use for another series or just sitting in the object pool.

This isn't the case anymore since `databaseShard` is the one that gets called back and uses an RLock to make sure only a valid ref would ever get called with `OnRetrieveBlock(...)` though.

storage/shard.go
robskillington: Could, however since go 1.8 it's pretty cheap, roughly 25-30 nanoseconds, which is fine since this only happens on reads which is already ripe for optimization in other places first.

storage/series/reader.go
robskillington: You know what, I might move this solely into the retriever - let me attempt this.

PR #441, opened 2018-01-24
Use truncate instead of divide to decide bucket index

cw9:
time.Truncate does not equal to a simple math division.
 
For example:
```
now := time.Now()
blockSize := 5 * 24 * time.Hour
fmt.Println(now)
fmt.Println(now.Truncate(blockSize))
fmt.Println(time.Unix(0, now.UnixNano()/int64(blockSize)*int64(blockSize)))
```
returns:
```
2018-01-24 14:29:31.624265 -0500 EST m=+0.003810489
2018-01-21 19:00:00 -0500 EST
2018-01-23 19:00:00 -0500 EST
```

That is because `Truncate returns the result of rounding t down to a multiple of d (since the zero time).`, where as `UnixNano returns t as a Unix time, the number of nanoseconds elapsed since January 1, 1970 UTC.`.

Since the start of each bucket are aligned using Truncate, we need to use Truncate when looking up which bucket to write to.

@xichen2020 @robskillington

### DIFF
--- a/storage/series/series_test.go
+++ b/storage/series/series_test.go
@@ -531,3 +531,32 @@ func TestSeriesOutOfOrderWritesAndRotate(t *testing.T) {
 	require.NoError(t, it.Err())
 	require.Equal(t, expected, actual)
 }
+
+func TestSeriesWriteReadFromTheSameBucket(t *testing.T) {
+	opts := newSeriesTestOptions()
+	opts = opts.SetRetentionOptions(opts.RetentionOptions().
+		SetRetentionPeriod(40 * 24 * time.Hour).
+		SetBlockSize(5 * 24 * time.Hour).
+		SetBufferFuture(10 * time.Minute).
+		SetBufferPast(20 * time.Minute))
+	curr := time.Now()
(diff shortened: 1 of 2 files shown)


storage/series/series_test.go
```diff
@@ -531,3 +531,32 @@ func TestSeriesOutOfOrderWritesAndRotate(t *testing.T) {
 	require.NoError(t, it.Err())
 	require.Equal(t, expected, actual)
 }
+
+func TestSeriesWriteReadFromTheSameBucket(t *testing.T) {
+	opts := newSeriesTestOptions()
+	opts = opts.SetRetentionOptions(opts.RetentionOptions().
+		SetRetentionPeriod(40 * 24 * time.Hour).
+		SetBlockSize(5 * 24 * time.Hour).
```
robskillington: Could you mention in a comment above the test that this uses a block size (5d) that is not equally as divisible as seconds from time zero and time epoch and tests this compatibility with reads?

cw9: Sure

PR #442, opened 2018-01-29
Add support for LRU caching policy

richardartoul:


storage/types.go
```diff
@@ -309,6 +309,11 @@ type Shard interface {
 type databaseShard interface {
 	Shard
 
+	// OnEvictedFromWiredList is the same as block.Owner. Had to duplicate
+	// it here because mockgen chokes on embedded interfaces sometimes:
+	// https://github.com/golang/mock/issues/10
```
prateek: hm, did their suggestion of using reflected mode not work too?

storage/shard.go
```diff
@@ -352,6 +352,26 @@ func (s *dbShard) OnRetrieveBlock(
...
+	s.RLock()
+	entry, _, err := s.lookupEntryWithLock(id)
+	s.RUnlock()
+
+	if err != nil && err != errShardEntryNotFound {
+		return // Shard is probably closing
+	}
+
+	if entry == nil {
+		// This could happen in a situation where the WiredList had a reference
```
prateek: Not sure I follow. I thought Tick's couldn't flush retrieved blocks?

storage/block/types.go
```diff
@@ -185,6 +186,29 @@ type DatabaseBlock interface {
...
 	// Close closes the block.
 	Close()
+
+	// SetOwner sets the owner of the block
+	SetOwner(Owner)
+
+	// Owner returns the owner of the block
+	Owner() Owner
+
+	// Private methods because only the Wired List itself should use them.
```
prateek: maybe break these out into a separate interface:

```
type DatabaseBlock interface {
 ...
}

type dbBlock interface {
  DatabaseBlock 

 // private methods
}
```

storage/block/types.go
```diff
@@ -185,6 +186,29 @@ type DatabaseBlock interface {
...
+	setPrev(block DatabaseBlock)
+	nextPrevUpdatedAtUnixNano() int64
+	setNextPrevUpdatedAtUnixNano(value int64)
+	wiredListEntry() wiredListEntry
+}
+
+// Owner is implemented by an "owner" of a block that should
+// be notified when a block is evicted from the wired list (and thus unavailable
+// for reads.)
+type Owner interface {
```
prateek: Would `Listener` by a more apt name?

runtime/runtime_options.go
```diff
@@ -34,6 +34,10 @@ const (
 	defaultTickSeriesBatchSize                  = 512
 	defaultTickPerSeriesSleepDuration           = 100 * time.Microsecond
 	defaultTickMinimumInterval                  = time.Minute
+	// defaultMaxWiredBlocks is 2MM by default which if using 2hr block sizes
+	// and writing every 10s at 1.4 point/byte (m3tsz) should use roughly 4gb:
+	// 1.4 * 6 * 120 * (2^21) = ~2gb
+	defaultMaxWiredBlocks = 2097152
```
prateek: maybe just `1<<31` then?

storage/shard.go
```diff
@@ -1019,7 +1039,9 @@ func (s *dbShard) FetchBlocks(
 	retriever := s.seriesBlockRetriever
 	onRetrieve := s.seriesOnRetrieveBlock
 	opts := s.seriesOpts
-	reader := series.NewReaderUsingRetriever(id, retriever, onRetrieve, opts)
+	// Nil for onRead callback because we don't want peer bootstrapping to impact
```
prateek: super nit: maybe set a var to nil and pass that as the arg

storage/types.go
richardartoul: Yeah I think I tried that as well, I could spend some more time on it though

storage/shard.go
richardartoul: I'm not sure if when I wrote this I meant flushed as in written to a fileset file, or I meant flushed as in evicted.

Regardless I think you're right that this shouldn't happen. If a block hasn't been flushed to disk yet then it shouldn't be in the WiredList. If it has been flushed and its in the wired list because it was retrieved from disk, then the series shouldn't be evicted from memory because it has a block.

It still seems like something we should defend against. Maybe I should emit a warning log?

storage/block/types.go
richardartoul: I thought Owner was a little more clear cause it conveyed the fact that the block isnt owned by the wired list...maybe EvictionListener?

runtime/runtime_options.go
```diff
@@ -34,6 +34,10 @@ const (
 	defaultTickSeriesBatchSize                  = 512
 	defaultTickPerSeriesSleepDuration           = 100 * time.Microsecond
 	defaultTickMinimumInterval                  = time.Minute
+	// defaultMaxWiredBlocks is 2MM by default which if using 2hr block sizes
+	// and writing every 10s at 1.4 point/byte (m3tsz) should use roughly 4gb:
+	// 1.4 * 6 * 120 * (2^21) = ~2gb
+	defaultMaxWiredBlocks = uint(1 << 31)
```
robskillington: Let's set this to ~256k (2^18) for now perhaps, this initial 2MM number was picked when wiredlist still included mutable blocks (that were skipped over when eviction was occurred).

services/m3dbnode/config/cache.go
```diff
@@ -40,5 +40,6 @@ func (c CacheConfigurations) SeriesConfiguration() SeriesCacheConfiguration {
 
 // SeriesCacheConfiguration is the series cache configuration.
 type SeriesCacheConfiguration struct {
-	Policy series.CachePolicy `yaml:"policy"`
+	Policy       series.CachePolicy `yaml:"policy"`
+	MaxLRUBlocks uint               `yaml:"max_lru_blocks"`
```
robskillington: To make this a little more consistent for if we add future options for each individual policy we can make it it a nullable composite struct for each policy?

i.e.

storage/block/types.go
```diff
@@ -137,7 +137,13 @@ type NewDatabaseBlockFn func() DatabaseBlock
 // DatabaseBlock represents a data block.
 type DatabaseBlock interface {
 	OnRetrieveBlock
+	DatabaseBlockPublic
+	// Private methods because only the Wired List itself should use them.
+	databaseBlockPrivate
+}
 
+// DatabaseBlockPublic is the public portion of the DatabaseBlock interface
+type DatabaseBlockPublic interface {
```
robskillington: Yeah instead of adding a `public` and `private` suffix naming convention that we don't use anywhere else, perhaps we can add just have `type databaseBlock interface { /* private methods */ }` and we can just keep the `DatabaseBlockPublic` methods inlined on the `DatabaseBlock` interface.

storage/block/types.go
```diff
@@ -185,6 +192,31 @@ type DatabaseBlock interface {
...
+
+	// SetOwner sets the owner of the block
+	SetOwner(Owner)
+
+	// Owner returns the owner of the block
+	Owner() Owner
+}
+
+// databaseBlockPrivate is the private portion of the DatabaseBlock interface
+type databaseBlockPrivate interface {
```
robskillington: Does this interface need to exist btw?  Wired list could just concretely be tied to `*dbBlock`.  Unless this interface truly does help with testing, we could just tie it concretely to the type.

storage/block/types.go
robskillington: Could just call this:

storage/block/wired_list.go
```diff
@@ -0,0 +1,269 @@
...
+			evicted: scope.Counter("evicted"),
+			// Incremented when a block is "pushed back" in the list, I.E
+			// it was already in the list
+			pushedBack: scope.Counter("pushed-back"),
+			// Incremented when a block is inserted into the list, I.E
+			// it wasn't already present
+			inserted: scope.Counter("inserted"),
+			// Measure how much time blocks spend in the list before being evicted
+			evictedAfterDuration: scope.Timer("evicted-after-duration"),
+		},
```
robskillington: Just for consistency, might be worth creating a `newWiredListMetrics(...) wiredListMetrics { return /* ... */ }` just to follow the pattern of the other components.

storage/block/wired_list.go
```diff
@@ -0,0 +1,269 @@
...
+func (l *WiredList) SetRuntimeOptions(value runtime.Options) {
+	atomic.StoreInt64(&l.maxWired, int64(value.MaxWiredBlocks()))
+}
+
+// Start starts processing the wired list
+func (l *WiredList) Start() {
+	l.Lock()
+	defer l.Unlock()
+	if l.updatesCh != nil {
+		return
```
robskillington: Might want to return an error here, like `errAlreadyStarted`.

storage/block/wired_list.go
```diff
@@ -0,0 +1,269 @@
...
+			}
+			i++
+		}
+		l.doneCh <- struct{}{}
+	}()
+}
+
+// Stop stops processing the wired list
+func (l *WiredList) Stop() {
+	close(l.updatesCh)
```
robskillington: Considering there's a check in `Start()` already, best to probably also lock, check updatesCh is not nil and if it is nil return an err to make the API symmetrical in safety/behavior.

storage/block/wired_list.go
```diff
@@ -0,0 +1,269 @@
...
+func (l *WiredList) Update(v DatabaseBlock) {
+	l.updatesCh <- v
+}
+
+// processUpdateBlock inspects a block that has been modified or read recently
+// and determines what outcome its state should have on the wired list.
+func (l *WiredList) processUpdateBlock(v DatabaseBlock) {
+	entry := v.wiredListEntry()
+	// The WiredList should never receive closed blocks or blocks that were not retrieved
+	// from disk, but we include the sanity check for posterity.
```
robskillington: If that's so, we should probably emit a metric and/or log a warning when this happens.

storage/series/series.go
```diff
@@ -162,14 +164,17 @@ func (s *dbSeries) updateBlocksWithLock() updateBlocksResult {
 			continue
 		}
 
-		if !currBlock.IsRetrieved() {
+		if cachePolicy == CacheAllMetadata && !currBlock.IsRetrieved() {
```
robskillington: Bugger, should probably deprecate this policy now.  I'll open an issue for it, maybe can spend some cycles myself too unwrapping it.

storage/series/series.go
```diff
@@ -180,8 +185,18 @@ func (s *dbSeries) updateBlocksWithLock() updateBlocksResult {
 			case CacheRecentlyRead:
 				sinceLastRead := now.Sub(currBlock.LastReadTime())
 				shouldUnwire = sinceLastRead >= wiredTimeout
+			case CacheLRU:
+				// The tick is responsible for managing the lifecycle of blocks that were not
+				// read from disk (not retrieved), and the WiredList will manage those that were
+				// retrieved from disk.
+				if !currBlock.WasRetrievedFromDisk() {
+					shouldUnwire = true
```
robskillington: Technically speaking, if you're in the `IsBlockRetrievable(...)` code path then it definitely is retrievable from disk.  So instead of doing this check we can probably just do `shouldUnwire = true` here.

storage/series/series.go
```diff
@@ -180,8 +185,18 @@ func (s *dbSeries) updateBlocksWithLock() updateBlocksResult {
...
+				}
+			default:
+				panic(fmt.Sprintf("Unhandled cache policy in series tick: %s", cachePolicy))
```
robskillington: nit: Use lower case first letter for panic to match logs, etc.

storage/series/series.go
```diff
@@ -370,14 +381,19 @@ func (s *dbSeries) mergeBlock(
...
 	}
 
 	// We are performing this in a lock, cannot wait for the existing
 	// block potentially to be retrieved from disk, lazily merge the stream.
 	newBlock.Merge(existingBlock)
-	blocks.AddBlock(newBlock)
+	s.addBlock(newBlock)
+}
+
+func (s *dbSeries) addBlock(b block.DatabaseBlock) {
```
robskillington: Maybe we should add suffix `...WithLock(...)` to methods `mergeBlock` and `addBlock` to make it clear.

storage/series/series.go
```diff
@@ -452,14 +468,68 @@ func (s *dbSeries) OnRetrieveBlock(
...
+	if list := s.opts.DatabaseBlockOptions().WiredList(); list != nil {
+		// Need to update the WiredList so blocks that were read from disk
+		// can enter the list (OnReadBlock is only called for blocks that
+		// were read from memory, regardless of whether the data originated
+		// from disk or a buffer rotation.)
+		list.Update(b)
+	}
+}
+
+// OnReadBlock is only called
```
robskillington: Comment seems not to be 100% complete here?

storage/series/series.go
```diff
@@ -452,14 +468,68 @@ func (s *dbSeries) OnRetrieveBlock(
...
+func (s *dbSeries) OnReadBlock(b block.DatabaseBlock) {
+	if list := s.opts.DatabaseBlockOptions().WiredList(); list != nil {
+		// The WiredList is only responsible for managing the lifecycle of blocks
+		// retrieved from disk.
+		if b.WasRetrievedFromDisk() {
+			// Need to update the WiredList so it knows which blocks have been
+			// most recently read.
+			list.Update(b)
```
robskillington: Does this happen synchronously during a read?  Are we worried about the locking synchronization overhead affecting p99s here?

storage/series/series.go
```diff
@@ -452,14 +468,68 @@ func (s *dbSeries) OnRetrieveBlock(
...
+		}
+	}
+}
+
+func (s *dbSeries) OnEvictedFromWiredList(id ident.ID, blockStart time.Time) {
+	s.Lock()
+
```
robskillington: Can probably just `defer s.Unlock()` since there's multiple returns.  Also this happens asynchronously to a read/write so won't add to tail latencies hopefully.

storage/series/series.go
```diff
@@ -514,12 +585,35 @@ func (s *dbSeries) Close() {
...
+		// by the WiredList and should not be closed here. They will eventually
+		// be evicted and closed by  the WiredList when it needs to make room
+		// for new blocks.
+		for _, block := range s.blocks.AllBlocks() {
+			if !block.WasRetrievedFromDisk() {
+				block.Close()
+			}
+		}
+	default:
+		s.blocks.RemoveAll()
```
robskillington: Is this necessary if `blocks.Reset()` is about to get called?

storage/shard.go
richardartoul: replaced with an error log, will throw this up on the test cluster to test our theory that this doesn't happen

runtime/types.go
```diff
@@ -115,6 +115,20 @@ type Options interface {
 	// helps throttle the tick when the amount of series is low and the sleeps
 	// on a per series basis is short.
 	TickMinimumInterval() time.Duration
+
+	// SetMaxWiredBlocks sets the max blocks to keep wired; zero is used
+	// to specify no limit. Wired blocks that are in the buffer, I.E are
+	// being written to, cannot be unwired. Similarly, blocks which have
+	// just been rotated out of the buffer but have not been flushed yet
+	// can also not be unwired. This means that the limit is best effort.
+	SetMaxWiredBlocks(value int) Options
```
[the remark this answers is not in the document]
richardartoul: done

storage/shard.go
richardartoul: done

runtime/runtime_options.go
richardartoul: not sure why I had this at 1 << 31, that was way too high. 1 << 18 ~~ 256k right?

services/m3dbnode/config/cache.go
richardartoul: yeah good idea

storage/block/types.go
richardartoul: It can't be tied to the concrete type because when you put things in the wired list you're dealing with an interface, not a concrete implementation.

richardartoul: We could do a forced type assertion, but seems kind of gross

storage/block/wired_list.go
richardartoul: done

storage/block/wired_list.go
richardartoul: done

storage/block/wired_list.go
richardartoul: done

storage/block/wired_list.go
richardartoul: done

storage/series/series.go
richardartoul: sounds good

storage/series/series.go
richardartoul: is that 100% true? Seems like something could be retrievable because it was flushed to disk, but still did not originate from disk because the tick that evicts it post-flush hasn't occurred yet

storage/series/series.go
richardartoul: replaced with a call to log.fatal

storage/series/series.go
richardartoul: good call

storage/series/series.go
richardartoul: yeah looks like I included it as an inline comment above, fixed it.

storage/series/series.go
richardartoul: In practice it didn't seem to affect the p99 reads

I could register it as a finalizer, I just don't like the idea that the wiredlist performs updates asynchronously and we're also going to schedule work asynchronously

I can change it if you think it will make a big difference. If I register a finalizer does that mean one extra goroutine? Or just one more task in the goroutine thats already created when the context closes?

storage/series/series.go
richardartoul: yep cause Reset doesn't explicitly close the underlying blocks

storage/series/series.go
richardartoul: @r let me know if you agree with me, and if so I'll add a comment

services/m3dbnode/server/server.go
```diff
@@ -132,6 +132,9 @@ func Run(runOpts RunOptions) {
 			SetLimitCheckEvery(cfg.Filesystem.ThroughputCheckEvery)).
 		SetWriteNewSeriesAsync(cfg.WriteNewSeriesAsync).
 		SetWriteNewSeriesBackoffDuration(cfg.WriteNewSeriesBackoffDuration)
+	if lruConfig := cfg.Cache.SeriesConfiguration().LRU; lruConfig != nil {
```
robskillington: nit: We have usually been naming local vars with the word "config" in them to "cfg" so this would be `lruCfg` to be slightly more consistent.  Don't feel too strongly about it, but would be more consistent.

storage/block/block.go
```diff
@@ -56,10 +56,24 @@ type dbBlock struct {
 	retriever  DatabaseShardBlockRetriever
 	retrieveID ident.ID
 
+	owner OnEvictedFromWiredList
```
robskillington: nit: I would rename `owner` to `onEvicted` if I was making this change, just to make it consistent *shrug*.

storage/block/wired_list.go
```diff
@@ -0,0 +1,304 @@
...
+	// Try to unwire all blocks possible
+	for bl := l.root.next(); l.length > maxWired && bl != &l.root; bl = bl.next() {
+		// Evict the block before closing it so that callers of series.ReadEncoded()
+		// don't get errors about trying to read from a closed block.
+		if onEvict := bl.OnEvictedFromWiredList(); onEvict != nil {
+			wlEntry := bl.wiredListEntry()
+			onEvict.OnEvictedFromWiredList(wlEntry.retrieveID, wlEntry.startTime)
+		}
+
+		bl.Close()
```
robskillington: Hm I see a race between `bl.Close()` and the `bl = bl.next()` iteration here.

storage/series/series.go
robskillington: Ah yes, ok I see so you don't want to unwire blocks brought from disk because the LRU does that.

storage/series/series.go
robskillington: It only creates one more task in the goroutine that's already created when the context closes.  But yeah since the wired list performs the updates asynchronous and not using the lock this is probably fine.

storage/series/series.go
robskillington: Hey so what about the blocks that get into the wired list just after being rotated into the series from the buffer when they are read (with the ReadEncoded and then the series `OnReadBlock(...)` call before a flush occurs.

richardartoul: Good call on the nit, I'll adjust that.

Regarding your other comment, blocks that are created by rotating into the series from the buffer will never get into the WiredList.

See the OnReadBlock function on the Series

```
// OnReadBlock is only called for blocks that were read from memory, regardless of
// whether the data originated from disk or buffer rotation.
func (s *dbSeries) OnReadBlock(b block.DatabaseBlock) {
	if list := s.opts.DatabaseBlockOptions().WiredList(); list != nil {
		// The WiredList is only responsible for managing the lifecycle of blocks
		// retrieved from disk.
		if b.WasRetrievedFromDisk() {
			// Need to update the WiredList so it knows which blocks have been
			// most recently read.
			list.Update(b)
		}
	}
}
```

This simplifies things a lot at the cost that after rotation we basically force these blocks to be read from disk, even if they were being read periodically while they were in the buffer. In practice this worked fine from a performance perspective.

richardartoul: Anything else and it gets too messy I think

storage/series/series.go
richardartoul: Yeah it would only matter if the wiredlists channel was full, and then you're probably under extremely heavy load anyways

services/m3dbnode/server/server.go
richardartoul: done

storage/block/block.go
richardartoul: yeah good call I missed this during my rename

storage/series/series.go
```diff
@@ -149,7 +151,32 @@ func (s *dbSeries) updateBlocksWithLock() updateBlocksResult {
...
+			//
+			// Also note that while technically the DatabaseBlock protects against double
+			// closes, they can be problematic due to pooling. I.E if the following sequence
+			// of actions happens:
+			// 		1) Tick closes expired block
+			// 		2) Block is re-inserted into pool
+			// 		3) Block is pulled out of pool and used for critical data
+			// 		4) WiredList tries to close the block, not knowing that it has
+			// 		   already been closed, and re-opened / re-used leading to
+			// 		   unexpected behavior or data loss.
```
robskillington: Solid documentation.

PR #444, opened 2018-01-29
Add tag constructs to `ts` package

prateek:
- Add `TagName`, `TagValue`, and `Tag` types to `ts/` package

/cc @robskillington @richardartoul @jeromefroe

### DIFF
--- a/ts/identifier_pool.go
+++ b/ts/identifier_pool.go
@@ -55,6 +55,17 @@ func (p *simpleIdentifierPool) GetBinaryID(ctx context.Context, v checked.Bytes)
 	return id
 }
 
+func (p *simpleIdentifierPool) GetBinaryTag(
+	ctx context.Context,
+	name checked.Bytes,
+	value checked.Bytes,
+) Tag {
+	return Tag{
+		Name:  TagName(p.GetBinaryID(ctx, name)),
+		Value: TagValue(p.GetBinaryID(ctx, value)),
+	}
+}
+
 func (p *simpleIdentifierPool) GetStringID(ctx context.Context, v string) ID {
 	data := p.bytesPool.Get(len(v))
 	data.IncRef()
@@ -63,9 +74,21 @@ func (p *simpleIdentifierPool) GetStringID(ctx context.Context, v string) ID {
 	return p.GetBinaryID(ctx, data)
 }
 
+func (p *simpleIdentifierPool) GetStringTag(ctx context.Context, name string, value string) Tag {
+	return Tag{
+		Name:  TagName(p.GetStringID(ctx, name)),
+		Value: TagValue(p.GetStringID(ctx, value)),
+	}
+}
+
 func (p *simpleIdentifierPool) Put(v ID) {
 	v.Reset()
 	p.pool.Put(v)
+}
+
+func (p *simpleIdentifierPool) PutTag(t Tag) {
+	p.Put(t.Name)
+	p.Put(t.Value)
 }
 
 func (p *simpleIdentifierPool) Clone(existing ID) ID {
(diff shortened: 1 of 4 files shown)


ts/tag_matcher.go
```diff
@@ -0,0 +1,53 @@
...
+package ts
+
+import (
+	"fmt"
+
+	"github.com/golang/mock/gomock"
+)
+
+// TagMatcher is a gomock.Matcher that matches Tag
+type TagMatcher interface {
```
richardartoul: Should this be in a test file?

ts/identifier_pool.go
```diff
@@ -55,6 +55,17 @@ func (p *simpleIdentifierPool) GetBinaryID(ctx context.Context, v checked.Bytes)
 	return id
 }
 
+func (p *simpleIdentifierPool) GetBinaryTag(
+	ctx context.Context,
+	name checked.Bytes,
+	value checked.Bytes,
+) Tag {
+	return Tag{
+		Name:  TagName(p.GetBinaryID(ctx, name)),
```
richardartoul: Do we want to keep referring to the id struct as an id if we're also using it to represent tags and their values?

The id struct also contains space for a Hash, do we need that for the tags and values?

ts/tag_matcher.go
prateek: nah, need it to be import-able cause its used by external packages for testing. Similar to `IDMatcher` in the same package. 

Could move it to `ts/testutil` or some such if you feel strongly.

ts/identifier_pool.go
prateek: Re: id struct reuse, I don't mind the internal struct being reused because it's package-local. Tho I was pondering renaming the external interface – currently, we have `ts.ID` which is used as a metric ID and namespace ID. It's pooled in the `ts.IdentifierPool`. This PR adds `ts.Tag(Name|Value)` as thin wrappers around it. I instead considered renaming `ts.ID` to something (`ts.Identifier`?) and having `ts.ID` and `ts.Tag(Name|Value)` all wrap that. Didn't see any value in doing that so avoided it. Thoughts?

prateek: Re: hashing, yeah we do need it for Tags. E.g. In the index, one of the components of a mutable segment is a 2-d map from tag name -> value -> collection of document ids.

robskillington: Hm yeah, we should probably move all this stuff to a new package like `ident`.  Then we would have `ident.ID` everywhere which is almost as short.

Also could be an opportunity to move it to to `m3x` so if you need to use this in `m3ninx` natively you can.

ts/tag_matcher.go
richardartoul: nah that makes sense

ts/identifier_pool.go
prateek: Hmm do you mean we'd use `ident.ID` instead of `ts.ID` everywhere? Or wrap `ts.ID` around `ident.ID`? How about `ts.Tag`?

prateek: Re: moving to `m3x`, couple of things to consider:
- The pooling code is pretty tied to m3db/context, so we'd have to break `ident` into only the non-pooling bits in `m3x` 
- The datamodel we use in m3db might not match general case. E.g. `m3metrics` assumes the `metric.ID` comprises all the different tags within it - ref: https://github.com/m3db/m3metrics/blob/master/metric/id/id.go

robskillington: Re: packaging.  It would be good if `m3ninx` shared the same underlying ID type, otherwise there will have to be casting and conversion between the boundaries which will be super unoptimal.

There is already an `id` package in `m3x` with some ID hashing related logic anyhow (I'd be in favor of renaming that `ident` so its not `id.ID` which is bad naming when it comes to Go conventions).

Re: the `m3metrics` case, I think that's ok if it retains its own view of what an ID is, it's modeling something else. which is a metric. 

Basically we need a shared package that let's us model:
- Single ID for primary key lookups (string-like)
- Supports pooling
- Doesn't sit in a package called `ts` which implies the ID is time series in nature, for `m3ninx` it could be used in other software that doesn't deal with time series at all

prateek: Fair enough. I'll do the m3x migration under https://github.com/m3db/m3x/pull/120 and then rebase this guy to update m3db

prateek:
Doing the work in https://github.com/m3db/m3x/pull/120, https://github.com/m3db/m3x/pull/121 and https://github.com/m3db/m3db/pull/449

PR #445, opened 2018-01-29
Make the integration tests not assume 1024 shards

richardartoul:


integration/server/serve.go
```diff
@@ -41,19 +41,21 @@ import (
 )
 
 const (
+	// DefaultNumShards is the default number of shards
+	DefaultNumShards     = 12
```
prateek: per offline chat, make this a testOption

integration/cluster_add_one_node_test.go
```diff
@@ -112,9 +117,9 @@ func TestClusterAddOneNode(t *testing.T) {
 		str   string
 		shard uint32
 	}{
-		{"foobarqux", 902},
-		{"bar", 397},
-		{"baz", 234},
+		{"jgfadfs", 10},
+		{"foobarqux", 2},
+		{"bar", 5},
```
robskillington: Ha - I knew this test was going to be susceptible to stuff like this, good job fixing these IDs too.

integration/server/serve.go
robskillington: Agreed, that would be ideal.

robskillington: P.S. If it makes it easier you can move everything in the `integration/server` package into just `integration`.  It was already moved from somewhere else that used to actually use it for running the server, so it's not really necessary to be in its own package here.

prateek: Dunno how much work it would be, but can we make the `integration` server share startup code with `services/m3dbnode`?

integration/cluster_add_one_node_test.go
```diff
@@ -112,9 +116,9 @@ func TestClusterAddOneNode(t *testing.T) {
 		str   string
 		shard uint32
 	}{
-		{"foobarqux", 902},
-		{"bar", 397},
-		{"baz", 234},
+		{"jgfadfs", 10},
```
prateek: use a rng to pick a number from `[0, opts.NumShards())` so this doesn't break if we run with fewer shards in the future

PR #446, opened 2018-01-30
client.Session methods use `ts.ID`

prateek:
Changes:
- Changes all API methods in `client/Session` to use `ts.ID` (instead of string) for both namespace and metric ID
- Adds a ts.ID iterator type

/cc @robskillington @richardartoul @jeromefroe

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -275,13 +275,13 @@ func newSession(opts Options) (clientSession, error) {
 	return s, nil
 }
 
-func (s *session) ShardID(id string) (uint32, error) {
+func (s *session) ShardID(id ident.ID) (uint32, error) {
 	s.RLock()
 	if s.state != stateOpen {
 		s.RUnlock()
 		return 0, errSessionStateNotOpen
 	}
-	value := s.topoMap.ShardSet().Lookup(ident.StringID(id))
+	value := s.topoMap.ShardSet().Lookup(id)
 	s.RUnlock()
 	return value, nil
 }
@@ -721,15 +721,14 @@ func (s *session) newHostQueue(host topology.Host, topoMap topology.Map) hostQue
 }
 
 func (s *session) Write(
-	namespace, id string,
+	namespace, id ident.ID,
 	t time.Time,
 	value float64,
 	unit xtime.Unit,
 	annotation []byte,
 ) error {
 	w := s.writeAttemptPool.Get()
-	w.args.namespace, w.args.id =
-		namespace, id
+	w.args.namespace, w.args.id = namespace, id
 	w.args.t, w.args.value, w.args.unit, w.args.annotation =
 		t, value, unit, annotation
 	err := s.writeRetrier.Attempt(w.attemptFn)
@@ -749,7 +748,7 @@ func (s *session) WriteTagged(
 }
 
 func (s *session) writeAttempt(
-	namespace, id string,
+	namespace, id ident.ID,
 	t time.Time,
 	value float64,
 	unit xtime.Unit,
@@ -758,9 +757,13 @@ func (s *session) writeAttempt(
 	var (
 		enqueued int32
 		majority = atomic.LoadInt32(&s.majority)
-		ctx      = s.contextPool.Get()
-		nsID     = s.idPool.GetStringID(ctx, namespace)
-		tsID     = s.idPool.GetStringID(ctx, id)
+		// NB(prateek): We retain an individual copy of the namespace, ID per
+		// writeState, as each writeState tracks the lifecycle of it's resources in
+		// use in the various queues. Tracking per writeAttempt isn't sufficient as
+		// we may enqueue multiple writeStates concurrently depending on retries
+		// and consistency level checks.
+		nsID = s.idPool.Clone(namespace)
+		tsID = s.idPool.Clone(id)
 	)
 
 	timeType, timeTypeErr := convert.ToTimeType(unit)
@@ -784,7 +787,7 @@ func (s *session) writeAttempt(
 
 	// todo@bl: Can we combine the writeOpPool and the writeStatePool?
 	state.op, state.majority = s.writeOpPool.Get(), majority
-	state.ctx, state.nsID, state.tsID = ctx, nsID, tsID
+	state.nsID, state.tsID = nsID, tsID
 
 	state.op.namespace = nsID
 	state.op.request.ID = tsID.Data().Get()
@@ -838,11 +841,12 @@ func (s *session) writeAttempt(
 }
 
 func (s *session) Fetch(
-	namespace string,
-	id string,
+	namespace ident.ID,
+	id ident.ID,
 	startInclusive, endExclusive time.Time,
 ) (encoding.SeriesIterator, error) {
-	results, err := s.FetchAll(namespace, []string{id}, startInclusive, endExclusive)
+	tsIDs := ident.NewIDsIterator(id)
+	results, err := s.FetchIDs(namespace, tsIDs, startInclusive, endExclusive)
 	if err != nil {
 		return nil, err
 	}
@@ -855,14 +859,14 @@ func (s *session) Fetch(
 	return iter, nil
 }
 
-func (s *session) FetchAll(
-	namespace string,
-	ids []string,
+func (s *session) FetchIDs(
+	namespace ident.ID,
+	ids ident.Iterator,
 	startInclusive, endExclusive time.Time,
 ) (encoding.SeriesIterators, error) {
 	f := s.fetchAttemptPool.Get()
-	f.args.namespace, f.args.ids, f.args.start, f.args.end =
-		namespace, ids, startInclusive, endExclusive
+	f.args.namespace, f.args.ids = namespace, ids
+	f.args.start, f.args.end = startInclusive, endExclusive
 	err := s.fetchRetrier.Attempt(f.attemptFn)
 	result := f.result
 	s.fetchAttemptPool.Put(f)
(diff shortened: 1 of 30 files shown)


client/write.go
```diff
@@ -96,7 +95,6 @@ type writeState struct {
 	session           *session
 	topoMap           topology.Map
 	op                *writeOp
-	ctx               context.Context
```
prateek: NB: can remove this as the writeState is no longer responsible for the lifecycle of the nsID+tsID, it's responsibility is pushed up to the user of the client/Session API

prateek:
@benraskin92 @nikunjgit just FYI, this refactoring will probably affect m3coordinator usage a bit.

ts/identifier_iterator.go
```diff
@@ -0,0 +1,66 @@
...
+package ts
+
+type idSliceIter struct {
+	backingSlice []ID
+	currentIdx   int
+	currentID    ID
+}
+
+// NewIDsIterator returns a new IDsIterator over the given IDs.
+func NewIDsIterator(ids ...ID) IDsIterator {
```
prateek: probably makes sense to take a slice of strings instead of ID

client/session.go
```diff
@@ -870,7 +867,7 @@ func (s *session) fetchAllAttempt(
 		resultErrs             int32
 		majority               int32
 		fetchBatchOpsByHostIdx [][]*fetchBatchOp
-		nsID                   = []byte(namespace)
+		nsID                   = namespace.Data().Get()
```
robskillington: Hm it's kind of unsafe to take a ref to the ts.ID underlying bytes and keep it around for a while.

In the `f.append(...)` call maybe we should make it take the `ts.ID` refs.  

Even then we can't guarantee the lifecycle of the ts.ID as this method might return early (say due to error or othe reason) but requests will be in flight with refs to the ts.IDs (both namespace and series IDs).

Maybe we should copy all the IDs and namespace ID at this point... either that or we need to use contexts to finalize of these properly and do the ref counting that way.

client/session.go
```diff
@@ -954,7 +954,7 @@ func (s *session) fetchAllAttempt(
 				successIters := results[:success]
 				resultsLock.RUnlock()
 				iter := s.seriesIteratorPool.Get()
-				iter.Reset(ids[idx], startInclusive, endExclusive, successIters)
+				iter.Reset(tsID.String(), startInclusive, endExclusive, successIters)
```
robskillington: This is quite expensive to cast back to string each time.  Better to either copy the ID and give the series a ts.ID itself, or do ref counting with the ts.ID provided by caller.

client/session_fetch_high_concurrency_test.go
```diff
@@ -201,3 +202,11 @@ func TestSessionFetchAllHighConcurrency(t *testing.T) {
 
 	require.NoError(t, session.Close())
 }
+
+func stringIDsToTsIDsIter(ids []string) ts.IDsIterator {
+	tsIDs := make([]ts.ID, 0, len(ids))
+	for _, id := range ids {
+		tsIDs = append(tsIDs, ts.StringID(id))
+	}
+	return ts.NewIDSliceIter(tsIDs)
```
robskillington: Hm, for convenience we should probably just create a copy of this method in the `ts` package.  i.e. `func NewStringsIDSliceIter(ids []string)`

client/session.go
```diff
@@ -735,7 +735,7 @@ func (s *session) Write(
 }
 
 func (s *session) writeAttempt(
-	namespace, id string,
+	nsID, tsID ts.ID,
```
robskillington: Yeah we need to be careful of the lifetime of these ts.IDs as well. 

We probably need to either copy the IDs using idPool to clone them (easiest) or make user supply a context, make them use the context to finalize the IDs (kind of annoying contract to have to know about), then keep our own context here, have their context depend on our context.

client/write.go
robskillington: Yeah so, issue is sometimes `Write(..)` returns before RPCs to all replicas have finished (quorum writes return early once quorum is reached).

prateek: Nice catch, didn't think of that. Will address in next patch.

client/session_fetch_high_concurrency_test.go
prateek: Yep will do

prateek:
The original code feels better than either of the two new ways we'd end up with:

- Context: The implicit contract required feels really brittle.
- Copy: This requires an extra alloc per ID for each RPC over the original code: one from the thrift request -> ts.ID input into client.Session.Write/Fetch, and an additional copy internally in the Session.* method. 

Would it be better to live with the native types instead of `ts.ID`?

prateek:
talked offline, @robskillington pointed out copying wouldn't be very expensive if we're pooling the IDs, only the extra memcpy/memset. Will try a build with that approach.

prateek:
The latest version does the following:
1) Allocate a ts.ID for ns and metric ID inside session.Write(), and `Finalize()` these along with the writeAttempt cleanup
2) Allocate a ts.ID for ns and metrics ID inside session.Fetch*, and rely on the GC to clean these up. 

For (2), because of the async nature of the fetch we need to either rely on the GC, or accept a context as an arg (and allocate all our ts.IDs off that context). Will test perf impact of this change on a production host and report back.

robskillington:
For (2): Why don't you transfer ownership to the `SeriesIterator`?  It can finalize the ID when it is closed itself. 

Right now `SeriesIterator` has type `ID()` as `string` which it really shouldn't be.

prateek:
@robskillington updated to the following semantics:
- Copy series and namespace ID on Write (backed by a pool), and release along with other resources in `writeState`
- Copy namespace ID and series ID(s) at the start of any Fetch* calls (backed by a pool). These are released once all corresponding fetch requests to the various replicas return. We also make an additional copy of the series ID(s) (still backed by the pool), and transfer ownership of it to `SeriesIterator`. The `SeriesIterator` calls `id.Finalize()` when `SeriesIterator.Close()` is called.

client/session.go
```diff
@@ -718,15 +718,18 @@ func (s *session) newHostQueue(host topology.Host, topoMap topology.Map) hostQue
...
 	unit xtime.Unit,
 	annotation []byte,
 ) error {
-	w := s.writeAttemptPool.Get()
-	w.args.namespace, w.args.id =
-		namespace, id
+	var (
+		w    = s.writeAttemptPool.Get()
+		nsID = s.idPool.Clone(namespace)
+		tsID = s.idPool.Clone(id)
```
robskillington: Hm, I think these will get finalized by the first write attempt at the end of the `writeState`.

A retry would probably cause a use after free.  Might need to clone it for every write attempt (simplest fix)?

client/session.go
```diff
@@ -1012,16 +1042,22 @@ func (s *session) fetchAllAttempt(
 			if atomic.AddInt32(&resultsAccessors, -1) == 0 {
 				s.iteratorArrayPool.Put(results)
 			}
+			if atomic.AddInt32(&idAccessors, -1) == 0 {
+				tsID.Finalize()
```
robskillington: Hm just trying to remember why we duplicate the finalization in both completion callbacks?

client/session.go
```diff
@@ -841,23 +843,33 @@ func (s *session) Fetch(
...
+
+func (s *session) fetchOwnedIDs(
+	namespace ts.ID,
+	ids ts.IDs,
 	startInclusive, endExclusive time.Time,
 ) (encoding.SeriesIterators, error) {
 	f := s.fetchAttemptPool.Get()
-	f.args.namespace, f.args.ids, f.args.start, f.args.end =
-		namespace, ids, startInclusive, endExclusive
+	f.args.namespace, f.args.ids = namespace, ids
```
robskillington: Same problem as the write attempt with here, the first attempt will finalize these IDs and a retry would cause use after free.

client/session.go
prateek: Yikes, nice catch. Added a test and fixed with by making a copy per writeState in the next revision.

client/session.go
prateek: Addressed this one the same way as above.

client/session.go
```diff
@@ -871,10 +875,19 @@ func (s *session) fetchAllAttempt(
...
 		fetchBatchOpsByHostIdx [][]*fetchBatchOp
-		nsID                   = []byte(namespace)
 		success                = false
 	)
 
+	// NB(prateek): need to make a copy of inputNamespace and inputIDs to control
+	// their life-cycle within this function.
+	namespace := s.idPool.Clone(inputNamespace)
+	idsClone := inputIDs.Clone()
+	ids := s.idPool.CloneIDs(idsClone)
```
robskillington: Strange, what are we cloning 2x here?  The slice and then the IDs or something like that?  

Any reason the idPool wouldn't do the slice cloning?

client/session.go
```diff
@@ -955,12 +978,20 @@ func (s *session) fetchAllAttempt(
...
+				seriesID := s.idPool.Clone(tsID)
+				iter.Reset(seriesID, startInclusive, endExclusive, successIters)
 				iters.SetAt(idx, iter)
 			}
 			if atomic.AddInt32(&resultsAccessors, -1) == 0 {
 				s.iteratorArrayPool.Put(results)
 			}
+			if atomic.AddInt32(&idAccessors, -1) == 0 {
+				tsID.Finalize()
+			}
```
robskillington: Think you need to add another block here?
```go
			if atomic.AddInt32(&namespaceAccessors, -1) == 0 {
				namespace.Finalize()
			}
```

client/session.go
prateek: `idsClone := inputIDs.Clone()` clones the iterator (only the struct referencing the underlying slice, not the slice itself). Need this to be able to iterate the original iterator multiple times in case of retries.

`ids := s.idPool.CloneIDs(..)` actually clones the ids in the slice.

client/session.go
prateek: Nah we don't need to at this time: `allCompletionFn` doesn't have a reference to `nsID`, only `completionFn` does. Tho this will change once SeriesIterator includes a reference to the namespace it's referencing. I'll make both changes and add to this PR. Can decide once you see it.

prateek: done

client/session.go
robskillington: Right I see, makes sense.

PR #450, opened 2018-02-02
RPC/Interface changes for tagged client APIs

prateek:
Initial PR for the addition of Tagged methods in the `client` package:
- Describe required changes in the `thrift.rpc` IDL
- Corresponding interface changes for API methods, input/output objects.

### DIFF
--- a/client/types.go
+++ b/client/types.go
@@ -31,9 +31,9 @@ import (
 	"github.com/m3db/m3db/generated/thrift/rpc"
 	"github.com/m3db/m3db/storage/block"
 	"github.com/m3db/m3db/storage/bootstrap/result"
+	"github.com/m3db/m3db/storage/index"
 	"github.com/m3db/m3db/storage/namespace"
 	"github.com/m3db/m3db/topology"
-	"github.com/m3db/m3ninx/index/segment"
 	"github.com/m3db/m3x/context"
 	"github.com/m3db/m3x/ident"
 	"github.com/m3db/m3x/instrument"
@@ -208,10 +208,10 @@ type Session interface {
 	FetchAll(namespace string, ids []string, startInclusive, endExclusive time.Time) (encoding.SeriesIterators, error)
 
 	// FetchTagged resolves the provided query to known IDs, and fetches the data for them.
-	FetchTagged(q Query, opts QueryOptions) (i encoding.SeriesIterators, t QueryResultsPageToken, moreResults bool, err error)
+	FetchTagged(index.Query, index.QueryOptions) (index.QueryResults, error)
 
 	// FetchTaggedIDs resolves the provided query to known IDs.
-	FetchTaggedIDs(q Query, opts QueryOptions) (i TaggedIDsIter, t QueryResultsPageToken, moreResults bool, err error)
+	FetchTaggedIDs(index.Query, index.QueryOptions) (index.QueryResults, error)
 
 	// ShardID returns the given shard for an ID for callers
 	// to easily discern what shard is failing when operations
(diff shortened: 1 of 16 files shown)


ts/types.go
```diff
@@ -41,6 +41,21 @@ type ID interface {
 	Reset()
 }
 
+// TagName is the name of a timeseries tag.
+type TagName ID
```
prateek: will move to m3x/ident once the other PRs are landed.

prateek: done

client/types.go
```diff
@@ -197,12 +198,21 @@ type Session interface {
...
 	Fetch(namespace string, id string, startInclusive, endExclusive time.Time) (encoding.SeriesIterator, error)
 
 	// FetchAll values from the database for a set of IDs
 	FetchAll(namespace string, ids []string, startInclusive, endExclusive time.Time) (encoding.SeriesIterators, error)
 
+	// FetchTagged resolves the provided query to known IDs, and fetches the data for them.
+	FetchTagged(q Query, startInclusive, endExclusive time.Time) (encoding.SeriesIterators, error)
+
+	// FetchTaggedIDs resolves the provided query to known IDs.
+	FetchTaggedIDs(q Query, startInclusive, endExclusive time.Time) (encoding.SeriesIterators, error)
```
robskillington: Hm, we should change the return type yeah?  IDs will need to be `[]ts.ID` or something yeah?  Also do we want to include limits, etc?

generated/thrift/rpc.thrift
```diff
@@ -44,11 +44,13 @@ exception WriteBatchRawErrors {
 service Node {
 	// Friendly not highly performant read/write endpoints
 	FetchResult fetch(1: FetchRequest req) throws (1: Error err)
+	FetchTaggedResult fetchTagged(1: FetchTaggedRequest req) throws (1: Error err)
```
robskillington: Hm, `FetchTaggedResult` has an optional `FetchRawResult` field, so it should be `fetchTaggedRaw` I think.

prateek: Ah nice catch. Migrated to using the uncompressed `datapoints` type in the response.

client/types.go
prateek: sure thing. 
- added a new iterator type in `client` for the first point,
- added corresponding fields for limit+pagetoken

generated/thrift/rpc.thrift
```diff
@@ -106,6 +110,48 @@ struct FetchBatchRawResult {
...
+	1: required binary tagName
+	2: required binary tagValueFilter
+	3: required bool negate
+	4: required bool regexp
+}
+
+struct FetchTaggedResult {
+	1: required list<FetchTaggedIDResult> elements
+	2: required bool hasMore
+	3: optional binary pageToken
```
robskillington: Might be simple to just make the semantics similar to fetch blocks metadata where pageToken is `nil` if no more results, otherwise if not `nil` then there are more?  The `hasMore` fields is kind of duplicative.

generated/thrift/rpc.thrift
robskillington: Hm shouldn't we return raw data?  It seems we're using raw IDs (binary instead of string).

The "Friendly" endpoints all use string IDs and datapoints and the "Performant" endpoints all use raw compressed buffers and binary IDs.

storage/index/types.go
```diff
@@ -0,0 +1,65 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package index
+
+import (
+	"time"
+
+	inx "github.com/m3db/m3ninx/index/segment"
```
robskillington: Do you feel strongly about renaming this import?  Usually we only do this for collisions.

generated/thrift/rpc.thrift
prateek: fair, next patch makes it 100% friendly.

generated/thrift/rpc.thrift
prateek: I reworked this to simplify the initial implementation - no paging token. Input only has a limit and on the return side a bool to indicate if the results are exhaustive. Can add the paging once we implement that query side in the indexing layer.

storage/index/types.go
prateek: Sure, undid the alias.

PR #452, opened 2018-02-09
Add static topology

benraskin92:
This allows us to run a multi-node m3db cluster with static topology until we figure out how we want to proceed with the embedded etcd. I will put the example config files in the m3coordinator repo.


environment/config.go
```diff
@@ -160,13 +160,13 @@ func (c Configuration) Configure(cfgParams ConfigurationParameters) (ConfigureRe
...
 
-		shardSet, hostShardSets, err := newStaticShardSet(c.Static.TopologyConfig.Shards, c.Static.ListenAddress, cfgParams.HostID)
+		shardSet, hostShardSets, err := newStaticShardSet(c.Static.TopologyConfig.Shards, c.Static.TopologyConfig.Hosts)
 		if err != nil {
 			err = fmt.Errorf("unable to create shard set for static config: %v", err)
 			return emptyConfig, err
 		}
 		staticOptions := topology.NewStaticOptions().
-			SetReplicas(1).
+			SetReplicas(len(c.Static.TopologyConfig.Hosts)).
```
robskillington: This means each host is a replica of each other.  This isn't a good idea, ha.  If this is truly what we want though we should think about expressing it better in the config as this is not clear what's being done.

benraskin92: So should we let users pass in `replicas` in the config and then if they don't specify, we just default to number hosts?

example/README.md
```diff
@@ -0,0 +1,48 @@
+## WARNING: This is Alpha software and not intended for use until a stable release.
+
+# Running M3DB on GCP
+
+Setup GCP for m3db:
+
+    1. Download necessary packages
+        $ sudo apt-get install golang golint make git golang-glide
+    2. Configure git
```
prateek: Do they need to do this?

example/README.md
```diff
@@ -0,0 +1,48 @@
...
+        $ mkdir -p /home/<user_name>/code/src/github.com/m3db
+    5. cd into m3db directory and git clone m3db
+        $ git clone https://github.com/m3db/m3db
+    6. Checkout `braskin/static_topology` (until this is landed in master)
+        $ git checkout origin/braskin/static_topology
+    7. Build m3db
+        $ git submodule update --init --recursive
+        $ glide install
+        $ make services
+    8. There are three config files in this directory (`config1.yaml`, `config2.yaml`, `config3.yaml`) - you will use one config per m3db node
```
prateek: naming nit: would be better to include service name, and the fact that it's for a different server in the filename. e.g. `m3dbnode-server1-config.yaml`

prateek: 'one config per m3db node' is a little vague, might be clearer to say per GCP instance?

example/README.md
```diff
@@ -0,0 +1,48 @@
+## WARNING: This is Alpha software and not intended for use until a stable release.
+
+# Running M3DB on GCP
+
+Setup GCP for m3db:
```
prateek: Would be useful to draw a picture of the different instances and services they'll end up with once they're done running this guide. I suggest https://monodraw.helftone.com/

example/config1.yaml
```diff
@@ -0,0 +1,198 @@
+logging:
```
prateek: looks like the three files are identical except for the hostID section. Any chance of reusing them somehow (e.g. provide a 1 file go utility to generate the log based on input)?

example/README.md
benraskin92: No- just if they want to make git changes. I'll remove.

example/README.md
benraskin92: Done

example/README.md
benraskin92: Cool- I'll work on something

example/config1.yaml
benraskin92: Sounds good- will address this in a subsequent pr.

example/README.md
robskillington: Hm, so I was very keen to not advise users to use the static topology.  Any reason why we're adding this?

environment/config.go
```diff
@@ -204,9 +207,11 @@ func newStaticShardSet(numShards int, listenAddress, hostID string) (sharding.Sh
 		return nil, nil, err
```
robskillington: Actually btw, the hash fn needs to be `sharding.DefaultHashFn(len(shards))` because it's creating a hashing function that needs to know how many shards its sharding over.

environment/config.go
```diff
@@ -160,16 +160,19 @@ func (c Configuration) Configure(cfgParams ConfigurationParameters) (ConfigureRe
...
 		if err != nil {
 			err = fmt.Errorf("unable to create shard set for static config: %v", err)
 			return emptyConfig, err
 		}
 		staticOptions := topology.NewStaticOptions().
-			SetReplicas(1).
 			SetHostShardSets(hostShardSets).
 			SetShardSet(shardSet)
 
+		if c.Static.TopologyConfig.Replicas != 0 {
```
robskillington: As discussed, let's default to replicas 0 and make sure replicas always matches `len(hosts)` or else users are going to have a very strange experience, because we only support static topology for clusters where each host holds the full set of data with this current config setup.

example/m3db-node-config.yaml
```diff
@@ -0,0 +1,195 @@
...
+          hosts:
+              - host: m3db_server
+                listenAddress: "0.0.0.0:9000"
+      namespaces:
+          - name: metrics
+            options:
+            retention:
+              retentionPeriod: 24h
+              blockSize: 4h
+              bufferPast: 2h
```
robskillington: blockSize: 2h, bufferPast: 1h is probably better defaults for users testing.

topology/types.go
```diff
@@ -202,7 +202,15 @@ func (l *ConsistencyLevel) UnmarshalYAML(unmarshal func(interface{}) error) erro
...
 type StaticConfiguration struct {
-	Shards int `yaml:"shards"`
+	Shards   int               `yaml:"shards"`
+	Replicas int               `yaml:"replicas"`
+	Hosts    []HostShardConfig `yaml:"hosts"`
+}
+
+// HostShardConfig stores host information for fanout
+type HostShardConfig struct {
+	Host          string `yaml:"host"`
```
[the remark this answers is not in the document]
benraskin92: I can change it to `HostID` - it's used to create a new host in `newStaticShardSet()`

PR #453, opened 2018-02-09
Add storage/index abstraction

prateek:
Changes:
- Add `databaseIndex` to `storage/` package, with stubbed implementation
- Re-wire database, dbNamespace and dbShard to pass appropriate references around to the index
- Add package `storage/index` for some common types used in the index (this will also be used in the `client` changes)

/cc @robskillington

### DIFF
--- a/storage/index/types.go
+++ b/storage/index/types.go
@@ -21,18 +21,45 @@
 package index
 
 import (
+	"time"
+
 	inx "github.com/m3db/m3ninx/index/segment"
 	"github.com/m3db/m3x/ident"
 )
 
-// Query ...
+// Query is a rich end user query to describe a set of constraints on required IDs.
 type Query struct {
 	inx.Query
 }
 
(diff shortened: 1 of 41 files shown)


storage/index/types.go
```diff
@@ -0,0 +1,65 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package index
+
+import (
+	"time"
+
+	inx "github.com/m3db/m3ninx/index/segment"
```
robskillington: Same comment here, we don't need to rewrite this yeah?

prateek: sure thing.

PR #461, opened 2018-02-20
Fix memory leak and unnecessary memory pressure during peer bootstrapping

richardartoul:


storage/series/series.go
```diff
@@ -520,7 +520,7 @@ func (s *dbSeries) Close() {
 	// a long period of time.
 	s.id = nil
 	s.buffer.Reset(s.opts)
-	s.blocks.Close()
+	s.blocks.Reset()
```
prateek: should this be `blocks.Close()`?

prateek: Ah disregard, i see you add the series back to the pool below. This needs to be `Reset()`. Maybe add clarifying comment.

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -363,6 +360,21 @@ func (s *peersSource) incrementalFlush(
 			// A block error is more interesting to bubble up than a close error
 			err = blockErr
 		}
+
+		if seriesCachePolicy != series.CacheAll && seriesCachePolicy != series.CacheAllMetadata {
+			// If we're not going to keep all of the data, or at least all of the metadata in memory
```
prateek: can this be done asynchronously?

storage/series/series.go
richardartoul: yeah I can add a comment

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: I think we probably could but it might complicate the logic a lot. Basically immediately after this executes, this shard result will be merged into the bootstrap result, and at that point messing with it could lead to lots of concurrency issues.

That said we spoke offline and this P.R results in a huge improvement to peer bootstrapping performance in general, so whatever penalty we're paying here shouldn't be an issue

client/session.go
```diff
@@ -1774,12 +1780,20 @@ func (s *session) streamBlocksMetadataFromPeerV2(
 
 		for _, elem := range result.Elements {
 			blockStart := time.Unix(0, elem.Start)
+
+			// Reset the reused ID data
+			reusedIDBytes.Resize(0)
+			reusedIDBytes.AppendAll(elem.ID)
+
+			// Clone the ID
+			clonedID := idPool.Clone(reusedID)
```
prateek: For some unrelated changes, I'm adding some methods to the ident.Pool to make this easier. Do me a solid and drop a `TODO(prateek): undo reusedIDBytes .. once https://github.com/m3db/m3x/pull/125 is merged` here

richardartoul:
@prateek Done

PR #463, opened 2018-02-20
Migrate `config_test` from integration => services/main

prateek:
/cc @robskillington


robskillington:
Interesting, any idea why coverage didn't improve?  It might need to be in the `server` package actually.

prateek:
@robskillington yep the short skip helped. leaving it in this package adds 0.05-0.1% ([multiple](https://coveralls.io/builds/15620875) [runs](https://coveralls.io/builds/15620641)), moving it to `server` made it [drop](https://coveralls.io/builds/15620787) -0.1%

PR #464, opened 2018-02-20
Refactor writeOp in `client`

prateek:
/cc @robskillington

### DIFF
--- a/client/write_state.go
+++ b/client/write_state.go
@@ -52,7 +52,6 @@ type writeState struct {
 	op                writeOp
 	nsID              ident.ID
 	tsID              ident.ID
-	tags              ident.Tags
 	majority, pending int32
 	success           int32
 	errors            []error
@@ -72,7 +71,6 @@ func (w *writeState) close() {
 
 	w.nsID.Finalize()
 	w.tsID.Finalize()
-	w.tags.Finalize()
 
 	w.op, w.majority, w.pending, w.success = nil, 0, 0, 0
 	w.nsID, w.tsID = nil, nil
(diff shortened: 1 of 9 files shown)


client/write_state.go
```diff
@@ -22,69 +22,24 @@ package client
...
-
-type writeOp struct {
-	namespace    ident.ID
-	shardID      uint32
-	request      rpc.WriteBatchRawRequestElement
-	datapoint    rpc.Datapoint
-	completionFn completionFn
-}
+// writableOp represents a generic write operation
+type writableOp interface {
```
robskillington: As discussed, maybe to avoid the overloaded term we can go with `type writeOp interface` and `type writeOperation struct` to separate implementation vs interface.

prateek: sure thing. done in next patch

client/write_state.go
```diff
@@ -112,11 +68,11 @@ func (w *writeState) reset() {
...
-	w.op.reset()
-	w.session.writeOpPool.Put(w.op)
+	w.op.Close()
 
 	w.nsID.Finalize()
 	w.tsID.Finalize()
+	w.tags.Finalize()
 
 	w.op, w.majority, w.pending, w.success = nil, 0, 0, 0
 	w.nsID, w.tsID = nil, nil
```
robskillington: We should add `w.tags = nil` here yeah?

client/write_state.go
```diff
@@ -94,7 +49,7 @@ type writeState struct {
 
 	session           *session
```
robskillington: As discussed maybe we can remove this now and replace with direct ref to the remaining pool and the consistency write level.

client/write_state.go
prateek: nice catch, removed.

PR #465, opened 2018-02-20
Add WriteTaggedBatch RPC types

prateek:
/cc @robskillington

### DIFF
--- a/generated/thrift/rpc.thrift
+++ b/generated/thrift/rpc.thrift
@@ -101,9 +101,9 @@ struct WriteRequest {
 struct WriteTaggedRequest {
 	1: required string nameSpace
 	2: required string id
-	3: required Datapoint datapoint
-	4: required list<string> tagNames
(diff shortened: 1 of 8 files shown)


generated/thrift/rpc.thrift
```diff
@@ -101,9 +101,9 @@ struct WriteRequest {
 struct WriteTaggedRequest {
 	1: required string nameSpace
 	2: required string id
-	3: required Datapoint datapoint
-	4: required list<binary> tagNames
-	5: required list<binary> tagValues
+	3: required list<string> tagNames
+	4: required list<string> tagValues
```
robskillington: Hm maybe we should create another type:
```
struct TagString {
  1: required string name
  2: required string value
}

struct TagRaw {
  1: required binary name
  2: required binary value
}
```

prateek: sure thing, done in next patch

PR #468, opened 2018-02-22
Reduce maximum memory usage during peer bootstrapping

richardartoul:


storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -90,32 +92,40 @@ func (s *peersSource) Read(
...
-				xlog.NewField("namespace", namespace.String()),
-			).Infof("peers bootstrapper resolving block retriever")
 
-			r, err := retrieverMgr.Retriever(nsMetadata)
-			if err != nil {
-				return nil, err
-			}
+		// Neither of these should ever happen
+		if seriesCachePolicy != series.CacheAll && retrieverMgr == nil {
+			panic("Tried to perform incremental flush without retriever manager")
```
robskillington: Let's use `logger.Fatalf(...)` instead, also nit: use lowercase to begin message.

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -90,32 +92,40 @@ func (s *peersSource) Read(
...
+		}
+		if seriesCachePolicy != series.CacheAll && persistManager == nil {
+			panic("Tried to perform incremental flush without persist manager")
```
robskillington: Same here.

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -146,39 +157,8 @@ func (s *peersSource) Read(
...
-				}
-
-				// Make unfulfilled
-				lock.Lock()
-				result.Add(flush.shard, nil, xtime.Ranges{}.AddRange(tr))
-				lock.Unlock()
-			}
-		}()
+		go s.startIncrementalQueueWorker(
+			incrementalWorkerDoneCh, incrementalQueue, persistFlush, result, &lock)
```
robskillington: Perhaps to keep the flow readable from this function, do the closing of the worker channel inline here?

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -127,14 +137,15 @@ func (s *peersSource) Read(
...
 	var (
-		lock                sync.Mutex
-		wg                  sync.WaitGroup
-		incrementalWg       sync.WaitGroup
-		incrementalMaxQueue = s.opts.IncrementalPersistMaxQueueSize()
-		incrementalQueue    = make(chan incrementalFlush, incrementalMaxQueue)
-		bopts               = s.opts.ResultOptions()
-		count               = len(shardsTimeRanges)
-		concurrency         = s.opts.DefaultShardConcurrency()
+		lock                    sync.Mutex
```
robskillington: Perhaps we should rename this `resultLock` so it's more clear?

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -188,79 +168,152 @@ func (s *peersSource) Read(
...
+	blockSize time.Duration,
+) {
+	it := ranges.Iter()
+	for it.Next() {
+		currRange := it.Value()
+
+		for start := currRange.Start; start.Before(currRange.End); start = start.Add(blockSize) {
+			version := s.opts.FetchBlocksMetadataEndpointVersion()
+			shardResult, err := session.FetchBootstrapBlocksFromPeers(nsMetadata,
+				shard, start, start.Add(blockSize), bopts, version)
```
robskillington: Perhaps add a `blockStart, blockEnd := start, start.Add(blockSize)` here so you can reuse here and when you specify `timeRange: xtime.Range{Start: blockStart, End: blockEnd}` you can reuse the var rather than doing the maths again (which isn't much overhead, but code more readable using the `blockStart`, `blockEnd` naming semantics.

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -320,7 +375,7 @@ func (s *peersSource) incrementalFlush(
 			segment, err := stream.Segment()
 			if err != nil {
 				tmpCtx.BlockingClose()
-				blockErr = err // Need to call prepared.Close, avoid return
+				blockErr = err // Need to  call prepared.Close, avoid return
```
robskillington: Any reason to add an extra space here?  (just want to avoid git blame being muddied if we don't have to touch this line)

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: I'm cool with the rename, but not sure what you're saying about the doneCh.

If I move this line:

```
<- incrementalWorkerDoneCh
```

into the goroutine like you have it, then the code won't wait for the incremental queue worker loop to complete before continuing.

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: done

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: done

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: done

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: sure

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: nah that was a mistake haha

client/session.go
```diff
@@ -1774,12 +1780,20 @@ func (s *session) streamBlocksMetadataFromPeerV2(
...
 		for _, elem := range result.Elements {
 			blockStart := time.Unix(0, elem.Start)
+
+			// Reset the reused ID data
+			reusedIDBytes.Resize(0)
+			reusedIDBytes.AppendAll(elem.ID)
+
+			// Clone the ID
+			clonedID := idPool.Clone(reusedID)
+
```
robskillington: Hey to clean up this code we should probably directly use the pool from `resultOpts`.

e.g.
```go
bytesPool := resultOpts.DatabaseBlockOptions().BytesPool()
```

robskillington:
LGTM other than last comment

PR #479, opened 2018-03-01
Add new tag serialization package

prateek:


serialize/decoder.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func (d *decoder) Close() {
+	d.release()
+	d.data = nil
+	d.err = nil
+	d.remaining = 0
+	if d.checkedData != nil {
+		d.checkedData.DecRef()
+		d.checkedData = nil
```
robskillington: Should also `d.checkedData.Finalize()` after the decref so it get's returned to pool.

robskillington: Hm guess you'll need to know if this one got duplicated, and if so, not to finalize I guess.

serialize/decoder.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (d *decoder) Finalize() {
+	d.Close()
+	if d.pool == nil {
+		return
+	}
+	d.pool.Put(d)
+}
+
+func (d *decoder) Duplicate() ident.TagIterator {
```
robskillington: Is this truly required?  Makes the lifecycle a lot harder to reason about to have a duplicate method.

serialize/decoder_pool.go
```diff
@@ -0,0 +1,57 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package serialize
```
robskillington: To have top level constructs such as `Decoder` and `Encoder` maybe this should be a sub-package of serialize?  i.e. `serialize/serializetag`

serialize/encoder.go
```diff
@@ -0,0 +1,177 @@
...
+	if numTags > MaxNumberTags {
+		return fmt.Errorf("too many tags to encode (%d), limit is: %d", numTags, MaxNumberTags)
+	}
+
+	if _, err := e.buf.Write(headerMagicBytes); err != nil {
+		e.buf.Reset()
+		return err
+	}
+
+	if _, err := e.buf.Write(encodeUInt16(uint16(numTags))); err != nil {
```
robskillington: Maybe for sanity check numTags isn't greater than 2^16?

serialize/encoder.go
```diff
@@ -0,0 +1,177 @@
...
+
+	return nil
+}
+
+func (e *encoder) Data() checked.Bytes {
+	return e.checkedBytes
+}
+
+func (e *encoder) Reset() {
+	if e.checkedBytes.NumRef() == 0 {
```
robskillington: Hm, strange?  Should we also check that NumRef == 0 after final DecRef if we fall through?

serialize/encoder.go
robskillington: I take it back, I see you are doing this above.

serialize/types.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Encoder encodes provided Tag iterators.
+type Encoder interface {
+	// Encode encodes the provided iterator into it's internal byte stream.
+	// NB: leaves the original iterator un-modified.
+	Encode(ident.TagIterator) error
+
+	// Data returns the encoded bytes.
+	// NB: Data() requires Encode() be called first, and it returning no
+	// error. Calling Data() without condition will cause a panic.
```
robskillington: Hm, perhaps better to return an error rather than panic.  We have tried very hard to make sure, even if it makes the API awkward, to always return err instead of panic.

serialize/types.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/m3db/m3x/checked"
+	"github.com/m3db/m3x/ident"
+)
+
+var (
+	// MaxNumberTags is the maximum number of tags that can be encoded.
+	MaxNumberTags = math.MaxUint16
+
+	// MaxTagLiteralLength is the maximum length of a tag Name/Value.
+	MaxTagLiteralLength uint16 = 4096
```
richardartoul: Somewhat unrelated, and maybe we should just file an issue, but can we implement this for IDs too?

serialize/types.go
```diff
@@ -0,0 +1,95 @@
...
+
+	// HeaderMagicNumber is an internal header used to denote the beginning of
+	// an encoded stream.
+	HeaderMagicNumber uint16 = 10101
+)
+
+// Encoder encodes provided Tag iterators.
+type Encoder interface {
```
richardartoul: Do we need an interface? Seems like there shouldn't be many situations where you can't just use the concrete type, even in a test, except to mock out really weird error cases?

richardartoul: I guess it allows us to swap out the encoder for a different encoder type if we want to have multiple formats or something...

richardartoul:
LGTM other than the question

serialize/decoder.go
prateek: Yeah this is a pretty ugly side-effect of passing around iterators by reference everywhere. 

We need to be able to iterate over the underlying values multiple times (e.g. in the case of retries). So it's either Duplication of some kind, which makes it easier for consumers of the type to reason about the lifecycles of the duplicated objects. Or, put some kind of "Reset to head" method on the iterator; but that makes it much harder to reason about concurrent use. 

I went the route of making it easier to use - `Duplicate()`. Do you prefer the other trade-off?

prateek: One thing to call out: don't want this type to be backed by a slice of items (so as to avoid []slice allocs), this makes it in-efficient to support random access with methods like `At()`

serialize/decoder.go
prateek: Good call. I can check `NumRef()==0` and Finalize() based on that. Sound good?

prateek: done

serialize/types.go
prateek: Made the change.

serialize/types.go
prateek: https://github.com/m3db/m3db/issues/489

serialize/decoder.go
robskillington: Maybe we should move to the random access pattern of using `Len()` and `ValueAt(idx int)`?

serialize/decoder_pool.go
robskillington: Maybe just call them `TagEncoder` and `TagDecoder` now then?  Even just in this package the names are ambiguous already.

serialize/types.go
robskillington: Makes sense, although I don't see any panics in checked bytes?

serialize/decoder_test.go
```diff
@@ -0,0 +1,205 @@
...
+
+	d := newTestDecoder()
+	d.Reset(wrapAsCheckedBytes(b))
+	require.NoError(t, d.Err())
+
+	require.True(t, d.Next())
+	require.False(t, d.Next())
+	require.Error(t, d.Err())
+}
+func TestDecodeCloneLifecycle(t *testing.T) {
```
robskillington: Can we perhaps add a check in this test that the checked bytes gets returned to pool somehow?

serialize/types.go
robskillington: Hm, this is frustrating but could we make this an option perhaps?  I can see someone using this in a generic TSDB case wanting to having quite long values they want to index on and the 4kb might need to be configurable.

serialize/decoder.go
prateek: Per above: avoiding `At()` because it forces us to be slice-backed to be efficient. 

Might be easier to talk offline. Will ping you in the AM.

serialize/decoder_pool.go
prateek: Sure

serialize/types.go
prateek: `checked.Bytes` uses ReadWriteRef which throws panics if the ref counting isn't done correctly. e.g. https://github.com/m3db/m3x/blob/master/checked/ref.go#L98

serialize/decoder.go
prateek: talked offline: going to retain the `Duplicate()` semantics and add a SUT prop test for lifecycle checks.

serialize/types.go
prateek: sure

serialize/decoder_pool.go
prateek: done

serialize/decoder_test.go
prateek: Sure, added a couple of mock based tests and a SUT prop test for `Decoder` lifecycle guarantees.

serialize/types.go
prateek: done

serialize/decoder_pool.go
prateek: done

serialize/decoder.go
prateek: done (actually caught some weird edge conditions so this was a good call).

PR #482, opened 2018-03-01
Implement WriteBatchTagged

prateek:


client/host_queue.go
```diff
@@ -177,16 +183,20 @@ func (q *queue) rotateOpsWithLock() []op {
 
 func (q *queue) drain() {
 	var (
-		currWriteOpsByNamespace      = make(map[ident.Hash][]op)
-		currBatchElementsByNamespace = make(map[ident.Hash][]*rpc.WriteBatchRawRequestElement)
-		writeBatchSize               = q.opts.WriteBatchSize()
+		currWriteOpsByNamespace            = make(map[ident.Hash][]op)
+		currBatchElementsByNamespace       = make(map[ident.Hash][]*rpc.WriteBatchRawRequestElement)
+		currTaggedWriteOpsByNamespace      = make(map[ident.Hash][]op)
+		currTaggedBatchElementsByNamespace = make(map[ident.Hash][]*rpc.WriteTaggedBatchRawRequestElement)
```
robskillington: Arg... I really need to get the generalized map ready.

client/session.go
```diff
@@ -781,24 +834,64 @@ func (s *session) writeAttempt(
...
+	default:
+		// should never happen
+		return errUnknownWriteAttemptType
+	}
+
+	// todo@bl: Can we combine the writeOpPool and the writeStatePool?
 	state.op, state.majority = op, majority
-	state.nsID, state.tsID = nsID, tsID
+	state.nsID, state.tsID, state.tagEncoder = nsID, tsID, tagEncoder
+	op.SetCompletionFn(state.completionFn)
```
robskillington: Hm, this is going to allocate as it will need to take a ref to the state ptr as well as the fn ptr.

client/write_batch_tagged_element_array_pool.go
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: nit: 2018.

client/write_batch_tagged_pool.go
```diff
@@ -0,0 +1,65 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: nit: 2018.

client/write_tagged_op.go
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: nit: 2018.

client/write_tagged_test.go
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: nit: 2018.

generated/thrift/rpc.thrift
```diff
@@ -287,7 +287,7 @@ struct WriteTaggedBatchRawRequest {
 
 struct WriteTaggedBatchRawRequestElement {
 	1: required binary id
-	2: required list<TagRaw> tags
+	2: required binary encodedTags
```
robskillington: Is `TagRaw` used anymore?  If not can we remove it?  And if so maybe we can also rename `TagString` to just `Tag`.

client/host_queue.go
prateek: :)

client/session.go
prateek: I don't follow what you mean by allocate. 

I ran the escape analysis compiler dump using `go build -gcflags '-m -l'`  on both the `op.SetCompletionFn(state.completionFn)` (the new way) and `op.completionFn = state.completionFn` (the old way); they both mark that `state.completionFn` escapes. 

Is there another allocation you're concerned about?

generated/thrift/rpc.thrift
prateek: sure thing, both done.

robskillington:
Hmm seems a prop test is failing:
```
--- FAIL: TestCommitLogReadWrite (0.21s)
	<autogenerated>:1: 
                        
	Error Trace:	read_write_prop_test.go:111
		
	Error:		Not equal: "aoszmkGGedebafoikLM0wxdxvzzt6lop0Io8whdyumed9jhhazos6taQkhGbmctBd3c" (expected)
			        != "henejf60Jtnqrxn1bFztwppgVlwdpanfiomtdsmflo6jwjxuOcvkjfjqp" (actual)
```

prateek:
@robskillington scarily enough, that's in the commit log code path (un-changed in this PR). Opened https://github.com/m3db/m3db/issues/497 to investigate

client/session.go
robskillington: This is fine, I think there is a small alloc here but looking at an `alloc_space` flame graph it looks like it's miniscule.

client/host_queue_test.go
```diff
@@ -678,6 +1175,46 @@ func testHostQueueFetchBatches(
...
+		Value:             value,
+		Timestamp:         timestamp,
+		TimestampTimeType: timeType,
+	}
+	w.request.EncodedTags = testEncode(tags)
+	w.completionFn = completionFn
+	return w
+}
+
+func testEncode(tags map[string]string) []byte {
```
robskillington: Hm interesting, don't want to use a real tag encoder here like the one in the PR #479?

client/session.go
```diff
@@ -781,24 +837,68 @@ func (s *session) writeAttempt(
...
+		wop.request.Datapoint.Annotation = annotation
+		op = wop
+	case taggedWriteAttemptType:
+		wop := s.writeTaggedOperationPool.Get()
+		wop.namespace = nsID
+		wop.shardID = s.topoMap.ShardSet().Lookup(tsID)
+		wop.request.ID = tsID.Data().Get()
+		encodedTagBytes, ok := tagEncoder.Data()
+		if !ok {
+			return errUnableToEncodeTags
```
robskillington: Hmm missing `RUnlock()` here.  Getting kinda worried about missing an `RUnlock()` in this entire block of code, maybe we could use a sub-method somewhere that does this work so we can more easily do explicit lock/unlock?

client/session.go
```diff
@@ -781,24 +837,68 @@ func (s *session) writeAttempt(
...
+		}
+		wop.request.EncodedTags = encodedTagBytes.Get()
+		wop.request.Datapoint.Value = value
+		wop.request.Datapoint.Timestamp = timestamp
+		wop.request.Datapoint.TimestampTimeType = timeType
+		wop.request.Datapoint.Annotation = annotation
+		op = wop
+	default:
+		// should never happen
+		return errUnknownWriteAttemptType
```
robskillington: Also missing `RUnlock()` here too.

client/host_queue_test.go
prateek: Nah explicitly chose to use mocks/stubbed code to keep the package level unit tests separate.

client/session.go
prateek: sure thing.

client/session.go
prateek: addressed by the `...WithRLock` refactor

client/options.go
```diff
@@ -77,6 +78,9 @@ const (
 	// defaultWriteOpPoolSize is the default write op pool size
 	defaultWriteOpPoolSize = 65536
 
+	// defaultWriteTaggedOpPoolSize is the default write tagged op pool size
+	defaultWriteTaggedOpPoolSize = 65536
```
prateek: Wanted to call this out. This will allocate a bunch more memory by default which doesn't seem great. 

To address I want to set the defaults to 0, and introduce a new method in the client package `func FillTagPools(Options) Options` which we'd expect people to use for index enabled workloads. Sound good?

client/options.go
```diff
@@ -120,6 +124,9 @@ const (
 	// defaultSeriesIteratorPoolSize is the default size of the series iterator pools
 	defaultSeriesIteratorPoolSize = 65536
 
+	// defaultTagEncoderPoolSize is the default size of the tag encoder pool.
+	defaultTagEncoderPoolSize = 4096
```
prateek: this would be another field set by the `FillTagPools` method

client/session.go
```diff
@@ -407,13 +426,27 @@ func (s *session) Open() error {
...
+
+	writeTaggedOperationPoolOpts := pool.NewObjectPoolOptions().
+		SetSize(s.opts.WriteTaggedOpPoolSize()).
+		SetInstrumentOptions(s.opts.InstrumentOptions().SetMetricsScope(
+			s.scope.SubScope("write-op-tagged-pool"),
+		))
+	s.writeTaggedOperationPool = newWriteTaggedOpPool(writeTaggedOperationPoolOpts)
+	s.writeTaggedOperationPool.Init()
+
+	writeStatePoolSize := s.opts.WriteOpPoolSize()
```
prateek: also have to do these checks because of the two kinds of pools using the writestate

client/session.go
```diff
@@ -697,16 +730,30 @@ func (s *session) newHostQueue(host topology.Host, topoMap topology.Map) hostQue
 	// For purposes of simplifying the options for pooling the write op pool size
 	// represents the number of ops to pool not including replication, this is due
 	// to the fact that the ops are shared between the different host queue replicas.
+	writeOpPoolSize := s.opts.WriteOpPoolSize()
```
prateek: same here

client/session.go
```diff
@@ -697,16 +730,30 @@ func (s *session) newHostQueue(host topology.Host, topoMap topology.Map) hostQue
...
 	writeBatchRequestPoolOpts := pool.NewObjectPoolOptions().
 		SetSize(hostBatches).
 		SetInstrumentOptions(s.opts.InstrumentOptions().SetMetricsScope(
 			s.scope.SubScope("write-batch-request-pool"),
 		))
 	writeBatchRequestPool := newWriteBatchRawRequestPool(writeBatchRequestPoolOpts)
 	writeBatchRequestPool.Init()
+
+	writeTaggedBatchRequestPoolOpts := pool.NewObjectPoolOptions().
+		SetSize(hostBatches).
```
prateek: Could avoid these allocs until someone explicitly calls `FillTaggedPools`

client/session.go
```diff
@@ -715,7 +762,20 @@ func (s *session) newHostQueue(host topology.Host, topoMap topology.Map) hostQue
 	writeBatchRawRequestElementArrayPool := newWriteBatchRawRequestElementArrayPool(
 		writeBatchRawRequestElementArrayPoolOpts, s.opts.WriteBatchSize())
 	writeBatchRawRequestElementArrayPool.Init()
-	hostQueue := s.newHostQueueFn(host, writeBatchRequestPool, writeBatchRawRequestElementArrayPool, s.opts)
+
+	writeTaggedBatchRawRequestElementArrayPoolOpts := pool.NewObjectPoolOptions().
```
prateek: here too

client/options.go
prateek: Did what we talked about and deployed an ingester node with these changes, it only bumped memory by like 1% with the extra pools. So I take this back. Lets leave the sensible defaults in there and not complicate anything further.

client/session.go
```diff
@@ -744,27 +806,29 @@ func (s *session) WriteTagged(
...
-		// use in the various queues. Tracking per writeAttempt isn't sufficient as
-		// we may enqueue multiple writeStates concurrently depending on retries
-		// and consistency level checks.
-		nsID = s.idPool.Clone(namespace)
-		tsID = s.idPool.Clone(id)
-	)
+	if wType != untaggedWriteAttemptType && wType != taggedWriteAttemptType {
+		// should never happen
+		return errUnknownWriteAttemptType
+	}
```
robskillington: Since this is already checked in `writeAttemptWithRLock(...)` perhaps it's not necessary?  (the switch on the `wtype`.

robskillington:
LGTM other than minor nit checking wtype twice in write attempt call stack.

client/session.go
prateek: Yeah this was more paranoia than necessary. Removed in the next patch.

PR #485, opened 2018-03-05
Embedded KV cluster

justinjc:
Enable an embedded etcd cluster to be run in M3DB via configuration.

The KV config can be minimally specified with the service defaulting required arguments, e.g.:
```
config:
    service:
        env: production
        zone: embedded
        service: m3db
        cacheDir: /var/lib/m3kv
    embeddedServer:
        initialCluster: host1=http://1.1.1.1:2380,host2=http://1.1.1.2:2380,host3=http://1.1.1.3:2380
```
Or more fully as something like:
```
config:
    service:
        env: production
        zone: embedded
        service: m3db
        cacheDir: /var/lib/m3kv
        etcdClusters: # defaults to endpoints parsed from initialCluster
            - zone: embedded
              endpoints:
                  - 1.1.1.1:2379
                  - 1.1.1.2:2379
                  - 1.1.1.3:2379
    embeddedServer:
        listenPeerUrls: # defaults to below
            - http://0.0.0.0:2380
        listenClientUrls: # defaults to below
            - http://0.0.0.0:2379
        dir: '/var/lib/etcd' # defaults to `fs.filePathPrefix` + '/etcd'
        initialAdvertisePeerUrls: # defaults to finding corresponding endpoint in initialCluster
            - http://1.1.1.1:2380
        advertiseClientUrls: # defaults to finding corresponding endpoint in initialCluster
            - http://1.1.1.1:2379
        initialCluster: host1=http://1.1.1.1:2380,host2=http://1.1.1.2:2380,host3=http://1.1.1.3:2380
        name: host1 # defaults to hostID
```

### DIFF
--- a/client/config.go
+++ b/client/config.go
@@ -154,7 +154,7 @@ func (c Configuration) NewAdminClient(
 	var err error
 	if envCfg.TopologyInitializer == nil {
 		switch {
-		case c.EnvironmentConfig.Service != nil:
+		case c.EnvironmentConfig.Static == nil && c.EnvironmentConfig.KV != nil:
 
 			envCfg, err = c.EnvironmentConfig.Configure(environment.ConfigurationParameters{
 				InstrumentOpts: iopts,
@@ -165,7 +165,7 @@ func (c Configuration) NewAdminClient(
 				return nil, err
 			}
 
-		case c.EnvironmentConfig.Static != nil:
+		case c.EnvironmentConfig.Static != nil && c.EnvironmentConfig.KV != nil:
 
 			envCfg, err = c.EnvironmentConfig.Configure(environment.ConfigurationParameters{})
 			if err != nil {
--- a/client/config.go
+++ b/client/config.go
@@ -153,27 +153,24 @@ func (c Configuration) NewAdminClient(
 
 	var err error
 	if envCfg.TopologyInitializer == nil {
-		switch {
-		case c.EnvironmentConfig.Static == nil && c.EnvironmentConfig.KV != nil:
-
+		if c.EnvironmentConfig.Service != nil {
 			envCfg, err = c.EnvironmentConfig.Configure(environment.ConfigurationParameters{
 				InstrumentOpts: iopts,
 				HashingSeed:    c.HashingConfiguration.Seed,
 			})
+
 			if err != nil {
 				err = fmt.Errorf("unable to create dynamic topology initializer, err: %v", err)
 				return nil, err
 			}
-
-		case c.EnvironmentConfig.Static != nil && c.EnvironmentConfig.KV != nil:
-
+		} else if c.EnvironmentConfig.Static != nil {
 			envCfg, err = c.EnvironmentConfig.Configure(environment.ConfigurationParameters{})
+
 			if err != nil {
 				err = fmt.Errorf("unable to create static topology initializer, err: %v", err)
 				return nil, err
 			}
-
-		default:
+		} else {
 			return nil, errConfigurationMustSupplyConfig
 		}
 	}
--- a/environment/config.go
+++ b/environment/config.go
@@ -65,13 +65,24 @@ type Configuration struct {
 
 // EmbeddedKV defines specific fields for the embedded kv server
 type EmbeddedKV struct {
-	Dir            string   `yaml:"dir"`
-	APUrls         []string `yaml:"initialAdvertisePeerUrls"`
-	ACUrls         []string `yaml:"advertiseClientUrls"`
-	LPUrls         []string `yaml:"listenPeerUrls"`
-	LCUrls         []string `yaml:"listenClientUrls"`
-	InitialCluster string   `yaml:"initialCluster"`
-	Name           string   `yaml:"name"`
+	Dir                string     `yaml:"dir"`
+	APUrls             []string   `yaml:"initialAdvertisePeerUrls"`
+	ACUrls             []string   `yaml:"advertiseClientUrls"`
+	LPUrls             []string   `yaml:"listenPeerUrls"`
+	LCUrls             []string   `yaml:"listenClientUrls"`
+	InitialCluster     string     `yaml:"initialCluster"`
+	Name               string     `yaml:"name"`
+	ClientSecurityJSON KVSecurity `json:"client-transport-security"`
+	PeerSecurityJSON   KVSecurity `json:"peer-transport-security"`
+}
+
+type KVSecurity struct {
+	CAFile        string `json:"ca-file"`
+	CertFile      string `json:"cert-file"`
+	KeyFile       string `json:"key-file"`
+	CertAuth      bool   `json:"client-cert-auth"`
+	TrustedCAFile string `json:"trusted-ca-file"`
+	AutoTLS       bool   `json:"auto-tls"`
 }
 
 // StaticConfiguration is used for running M3DB with a static config
--- a/environment/config.go
+++ b/environment/config.go
@@ -27,7 +27,6 @@ import (
 
 	etcdclient "github.com/m3db/m3cluster/client/etcd"
 	"github.com/m3db/m3cluster/kv"
-	m3clusterkvmem "github.com/m3db/m3cluster/kv/mem"
 	"github.com/m3db/m3cluster/services"
 	"github.com/m3db/m3cluster/shard"
 	"github.com/m3db/m3db/kvconfig"
@@ -39,6 +38,10 @@ import (
 	"github.com/m3db/m3x/instrument"
 )
 
+const (
+	defaultSDTimeout = 10 * time.Minute
+)
+
 var (
 	errNilRetention  = errors.New("namespace retention options cannot be empty")
 	errMissingConfig = errors.New("must supply service or static config")
@@ -46,11 +49,33 @@ var (
 
 // Configuration is a configuration that can be used to create namespaces, a topology, and kv store
 type Configuration struct {
-	// Service is used when a topology initializer is not supplied.
-	Service *etcdclient.Configuration `yaml:"service"`
-
 	// StaticConfiguration is used for running M3DB with a static config
 	Static *StaticConfiguration `yaml:"static"`
+
+	// KV sets the kv configuration for either a clustered or embedded etcd
+	KV *KVConfig `yaml:"kv"`
+}
+
+// KVConfig sets the configs for the KV client and server
+type KVConfig struct {
+	Client *etcdclient.Configuration `yaml:"client"`
+
+	// Presence of a (etcd) server in this config denotes an embedded cluster
+	Server *EmbeddedKV `yaml:"server"`
+}
+
+// EmbeddedKV defines specific fields for the embedded kv server
+type EmbeddedKV struct {
+	Dir            string   `yaml:"dir"`
+	APUrls         []string `yaml:"initialAdvertisePeerUrls"`
+	ACUrls         []string `yaml:"advertiseClientUrls"`
+	LPUrls         []string `yaml:"listenPeerUrls"`
+	LCUrls         []string `yaml:"listenClientUrls"`
+	InitialCluster string   `yaml:"initialCluster"`
+	Name           string   `yaml:"name"`
+
+	// NamespaceTimeout is the timeout duration for setting a namespace
+	NamespaceTimeout time.Duration `yaml:"namespaceTimeout"`
 }
 
 // StaticConfiguration is used for running M3DB with a static config
@@ -95,36 +120,42 @@ type ConfigureResults struct {
 
 // ConfigurationParameters are options used to create new ConfigureResults
 type ConfigurationParameters struct {
-	InstrumentOpts instrument.Options
-	HashingSeed    uint32
-	HostID         string
+	InstrumentOpts   instrument.Options
+	HashingSeed      uint32
+	HostID           string
+	NamespaceTimeout time.Duration
 }
 
 // Configure creates a new ConfigureResults
 func (c Configuration) Configure(cfgParams ConfigurationParameters) (ConfigureResults, error) {
-
 	var emptyConfig ConfigureResults
 
-	switch {
-	case c.Service != nil:
-		configSvcClientOpts := c.Service.NewOptions().
-			SetInstrumentOptions(cfgParams.InstrumentOpts)
-		configSvcClient, err := etcdclient.NewConfigServiceClient(configSvcClientOpts)
-		if err != nil {
-			err = fmt.Errorf("could not create m3cluster client: %v", err)
-			return emptyConfig, err
-		}
+	if c.KV.Client.SDConfig.InitTimeout == 0 {
+		c.KV.Client.SDConfig.InitTimeout = defaultSDTimeout
+	}
 
+	configSvcClientOpts := c.KV.Client.NewOptions().
+		SetInstrumentOptions(cfgParams.InstrumentOpts).
+		SetServiceDiscoveryConfig(c.KV.Client.SDConfig)
+	configSvcClient, err := etcdclient.NewConfigServiceClient(configSvcClientOpts)
+	if err != nil {
+		err = fmt.Errorf("could not create m3cluster client: %v", err)
+		return emptyConfig, err
+	}
+
+	switch {
+	case c.Static == nil:
 		dynamicOpts := namespace.NewDynamicOptions().
 			SetInstrumentOptions(cfgParams.InstrumentOpts).
 			SetConfigServiceClient(configSvcClient).
-			SetNamespaceRegistryKey(kvconfig.NamespacesKey)
+			SetNamespaceRegistryKey(kvconfig.NamespacesKey).
+			SetInitTimeout(cfgParams.NamespaceTimeout)
 		nsInit := namespace.NewDynamicInitializer(dynamicOpts)
 
 		serviceID := services.NewServiceID().
-			SetName(c.Service.Service).
-			SetEnvironment(c.Service.Env).
-			SetZone(c.Service.Zone)
+			SetName(c.KV.Client.Service).
+			SetEnvironment(c.KV.Client.Env).
+			SetZone(c.KV.Client.Zone)
 
 		topoOpts := topology.NewDynamicOptions().
 			SetConfigServiceClient(configSvcClient).
@@ -189,7 +220,11 @@ func (c Configuration) Configure(cfgParams ConfigurationParameters) (ConfigureRe
 
 		topoInit := topology.NewStaticInitializer(staticOptions)
 
-		kv := m3clusterkvmem.NewStore()
+		kv, err := configSvcClient.KV()
+		if err != nil {
+			err = fmt.Errorf("could not create KV client, %v", err)
+			return emptyConfig, err
+		}
 
 		configureResults := ConfigureResults{
 			NamespaceInitializer: nsInitStatic,
--- a/environment/config.go
+++ b/environment/config.go
@@ -143,13 +143,14 @@ type ConfigurationParameters struct {
 func (c Configuration) Configure(cfgParams ConfigurationParameters) (ConfigureResults, error) {
 	var emptyConfig ConfigureResults
 
-	if c.Service.SDConfig.InitTimeout == 0 {
-		c.Service.SDConfig.InitTimeout = defaultSDTimeout
+	sdTimeout := defaultSDTimeout
+	if initTimeout := c.Service.SDConfig.InitTimeout; initTimeout != nil && *initTimeout != 0 {
+		sdTimeout = *initTimeout
 	}
 
 	configSvcClientOpts := c.Service.NewOptions().
 		SetInstrumentOptions(cfgParams.InstrumentOpts).
-		SetServiceDiscoveryConfig(c.Service.SDConfig)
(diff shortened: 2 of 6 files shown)


environment/config.go
```diff
@@ -39,18 +38,44 @@ import (
 	"github.com/m3db/m3x/instrument"
 )
 
+const (
+	defaultSDTimeout = 10 * time.Minute
```
prateek: 10 minutes is pretty high. Any reason this can't be order of seconds?

cw9: +1

client/config.go
```diff
@@ -165,7 +165,7 @@ func (c Configuration) NewAdminClient(
 				return nil, err
 			}
 
-		case c.EnvironmentConfig.Static != nil:
+		case c.EnvironmentConfig.Static != nil && c.EnvironmentConfig.KV != nil:
```
cw9: should check `KV == nil` ?

environment/config.go
```diff
@@ -39,18 +38,44 @@ import (
...
+	// KV sets the kv configuration for either a clustered or embedded etcd
+	KV *KVConfig `yaml:"kv"`
+}
+
+// KVConfig sets the configs for the KV client and server
+type KVConfig struct {
+	Client *etcdclient.Configuration `yaml:"client"`
+
+	// Presence of a (etcd) server in this config denotes an embedded cluster
+	Server *EmbeddedKV `yaml:"server"`
```
cw9: this should be mandatory?

environment/config.go
```diff
@@ -39,18 +38,44 @@ import (
...
 	// StaticConfiguration is used for running M3DB with a static config
 	Static *StaticConfiguration `yaml:"static"`
+
+	// KV sets the kv configuration for either a clustered or embedded etcd
+	KV *KVConfig `yaml:"kv"`
+}
+
+// KVConfig sets the configs for the KV client and server
+type KVConfig struct {
+	Client *etcdclient.Configuration `yaml:"client"`
```
cw9: this should be mandatory?

services/m3dbnode/config/config_test.go
```diff
@@ -224,19 +224,30 @@ pooling:
...
+            zone: embedded
+            service: m3db
+            cacheDir: /var/lib/m3kv
+            etcdClusters:
+                - zone: embedded
+                  endpoints:
+                    - 1.1.1.1:2379
+                    - 1.1.1.2:2379
+                    - 1.1.1.3:2379
+        server:
```
cw9: @schallert can you take a look at these etcd configs?

environment/config.go
prateek: Lets keep server/client configs in different structs: leave `Service` where it was; and add EmbeddedKV to the top-level `Configuration` struct.

environment/config.go
```diff
@@ -39,18 +38,44 @@ import (
...
+
+	// Presence of a (etcd) server in this config denotes an embedded cluster
+	Server *EmbeddedKV `yaml:"server"`
+
+	// NamespaceTimeout is the timeout duration for setting a namespace
+	NamespaceTimeout time.Duration `yaml:"namespaceTimeout"`
+}
+
+// EmbeddedKV defines specific fields for the embedded kv server
+type EmbeddedKV struct {
```
prateek: add an `Enabled bool` field to this to be explicit

services/m3dbnode/config/config.go
```diff
@@ -190,3 +203,134 @@ type HashingConfiguration struct {
...
+	kvCfg := cfg.EnvironmentConfig.KV.Server
+
+	dir := kvCfg.Dir
+	if dir == "" {
+		dir = path.Join(cfg.Filesystem.FilePathPrefix, defaultETCDDirSuffix)
+	}
+	newKVCfg.Dir = dir
+
+	// listen-peer-urls
+	LPUrls, err := convertToURLsWithDefault(kvCfg.LPUrls, defaultETCDListenHost+defaultETCDServerPort)
```
prateek: could you use a function for the string concatenation:

```
func newURL(host string, port int) string { 
  return fmt.Sprintf("http://%s:%d", host, port)
}
```

services/m3dbnode/config/config.go
```diff
@@ -190,3 +203,134 @@ type HashingConfiguration struct {
 	// Murmur32 seed value.
 	Seed uint32 `yaml:"seed"`
 }
+
+// GetETCDConfig creates a new embedded etcd config from kv config.
+func GetETCDConfig(cfg Configuration) (*embed.Config, error) {
```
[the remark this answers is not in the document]
prateek: also a naming nit: `ETCDConfig()` instead of `Get..`

services/m3dbnode/config/config.go
```diff
@@ -190,3 +203,134 @@ type HashingConfiguration struct {
...
+			return endpoint[:colonIdx], nil
+		}
+	}
+
+	return "", errors.New("host not in initialCluster list")
+}
+
+// InitialClusterToETCDEndpoints converts the etcd config initialCluster into etcdClusters config
+// for m3cluster.
+func InitialClusterToETCDEndpoints(initialCluster string) ([]string, error) {
```
prateek: does this need to be an exported type? any reason to not accept initialCluster as `initialClusterHosts []string` if you're having to do this anyway?

prateek:
Could you add a test which uses the end to end code path (can probably re-work `services/m3dbnode/main/main_test.go` to use the new code you've added in).

prateek:
Would be good to expose TLS/SSL options so people can run ETCD securely

services/m3dbnode/config/config.go
justinjc: I thought about this too, but decided to keep the same format as the `initial-cluster` config from etcd, thinking that it'll make it easier for users to construct the config.

environment/config.go
justinjc: No longer applicable after restructuring this.

justinjc:
>Could you add a test which uses the end to end code path (can probably re-work `services/m3dbnode/main/main_test.go` to use the new code you've added in).

>Would be good to expose TLS/SSL options so people can run ETCD securely

👍 I'll work on these.

services/m3dbnode/config/config.go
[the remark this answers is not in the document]
justinjc: `embed.Config` takes in `url.URL` structs for a number of variables, which doesn't allow inputting them via JSON. Unfortunately, the helper functions and structs to help construct this config are unexported and only takes in a path for a yaml file instead.

environment/config.go
robskillington: I'm fine with not adding an explicit Enabled, seems strange that you would add it and then not enable it.

environment/config.go
```diff
@@ -51,6 +55,34 @@ type Configuration struct {
...
+	// Presence of a (etcd) server in this config denotes an embedded cluster
+	EmbeddedServer *EmbeddedKV `yaml:"embeddedServer"`
+
+	// NamespaceTimeout is the timeout duration for setting a namespace
+	NamespaceTimeout time.Duration `yaml:"namespaceTimeout"`
+}
+
+// EmbeddedKV defines specific fields for the embedded kv server
+type EmbeddedKV struct {
+	Dir                string     `yaml:"dir"`
```
robskillington: Would `RootDir` be a better description?  It's a bit ambiguous otherwise.

environment/config.go
```diff
@@ -51,6 +55,34 @@ type Configuration struct {
...
+	APUrls             []string   `yaml:"initialAdvertisePeerUrls"`
+	ACUrls             []string   `yaml:"advertiseClientUrls"`
+	LPUrls             []string   `yaml:"listenPeerUrls"`
+	LCUrls             []string   `yaml:"listenClientUrls"`
```
robskillington: Maybe we get rid of the acronymization of these variables?  Seems like they are inconsistent with the other long form variables below.

environment/config.go
```diff
@@ -51,6 +55,34 @@ type Configuration struct {
...
+	InitialCluster     string     `yaml:"initialCluster"`
+	Name               string     `yaml:"name"`
+	ClientSecurityJSON KVSecurity `json:"client-transport-security"`
+	PeerSecurityJSON   KVSecurity `json:"peer-transport-security"`
```
robskillington: Hm what's the need for these fields that are only parseable using JSON?  How would we use them?

environment/config.go
```diff
@@ -51,6 +55,34 @@ type Configuration struct {
 
 	// StaticConfiguration is used for running M3DB with a static config
 	Static *StaticConfiguration `yaml:"static"`
+
+	// Presence of a (etcd) server in this config denotes an embedded cluster
+	EmbeddedServer *EmbeddedKV `yaml:"embeddedServer"`
+
+	// NamespaceTimeout is the timeout duration for setting a namespace
```
robskillington: For setting a namespace?  Do you mean namespace creation or resolving a namespace from KV?

environment/config.go
```diff
@@ -95,112 +127,131 @@ type ConfigureResults struct {
...
+	if err != nil {
+		err = fmt.Errorf("could not create m3cluster client: %v", err)
+		return emptyConfig, err
+	}
+
+	if c.Service != nil {
+		return c.newDynamicConfigRes(configSvcClient, cfgParams)
+	}
+
+	if c.Static != nil {
```
robskillington: Should probably check that they are not both set too?  Should be invalid to specify both.

environment/config.go
```diff
@@ -39,6 +39,10 @@ import (
 	"github.com/m3db/m3x/instrument"
 )
 
+const (
+	defaultSDTimeout = 5 * time.Second
```
robskillington: Perhaps 10s is a better timeout, also this might cause problems for newly created clusters that don't have a namespace setup, especially if we need to call endpoints on the nodes themselves to create the first namespace (humans don't want to race this timestamp to create a cluster).

environment/config.go
```diff
@@ -95,112 +127,131 @@ type ConfigureResults struct {
...
+	}
+
+	if c.Static != nil {
+		return c.newStaticConfigRes(configSvcClient, cfgParams)
+	}
+
+	return emptyConfig, errMissingConfig
+}
+
+func (c Configuration) newDynamicConfigRes(configSvcClient client.Client, cfgParams ConfigurationParameters) (ConfigureResults, error) {
```
robskillington: minor nit: `configureDynamic(..)` and `configureStatic(..)` would be more consistent naming as children methods of the top level `Configure(...)` method.

services/m3dbnode/config/config.go
```diff
@@ -190,3 +205,146 @@ type HashingConfiguration struct {
 	// Murmur32 seed value.
 	Seed uint32 `yaml:"seed"`
 }
+
+// ETCDConfig creates a new embedded etcd config from kv config.
+func ETCDConfig(cfg Configuration) (*embed.Config, error) {
```
robskillington: minor nit: Perhaps a better name would be `NewEtcdEmbedConfiguration`?  Also this could be instead a member method on `Configuration` itself which would be more similar to our other sub-component creations off of `Configuration`.

e.g.

environment/config.go
```diff
@@ -51,6 +55,34 @@ type Configuration struct {
 
 	// StaticConfiguration is used for running M3DB with a static config
 	Static *StaticConfiguration `yaml:"static"`
+
+	// Presence of a (etcd) server in this config denotes an embedded cluster
+	EmbeddedServer *EmbeddedKV `yaml:"embeddedServer"`
```
robskillington: Could we call this `seedNodeServer`?  It more accurately describes what we're doing, we should name everything around the nodes that run etcd as "seed nodes".

environment/config.go
```diff
@@ -51,6 +55,34 @@ type Configuration struct {
...
+}
+
+// EmbeddedKV defines specific fields for the embedded kv server
+type EmbeddedKV struct {
+	Dir                string     `yaml:"dir"`
+	APUrls             []string   `yaml:"initialAdvertisePeerUrls"`
+	ACUrls             []string   `yaml:"advertiseClientUrls"`
+	LPUrls             []string   `yaml:"listenPeerUrls"`
+	LCUrls             []string   `yaml:"listenClientUrls"`
+	InitialCluster     string     `yaml:"initialCluster"`
```
robskillington: Instead of making this comma separated endpoints, can we instead make this a `[]string`?  It looks much more respectable and readable in YAML as a `[]string` rather than a comma separated string.

services/m3dbnode/config/config.go
```diff
@@ -190,3 +205,146 @@ type HashingConfiguration struct {
...
+	nodesStr := strings.Split(initialCluster, ",")
+
+	for _, nodeStr := range nodesStr {
+		nodeData := strings.Split(nodeStr, "=")
+
+		if len(nodeData) != 2 {
+			return "", errors.New("invalid initialCluster format")
+		}
+
+		if hostID == nodeData[0] {
```
robskillington: As discussed at some point the hostID may not be the same as the IP:port listed in the seed nodes list.

Can we instead add the concept of every seed host with:

services/m3dbnode/config/config.go
robskillington: See my comment in `getHostFromHostID(...)`.  Unfortunately the hostID we resolve is not always going to match the raw etcd ip:port or host:port that we give to etcd when we configured it.

environment/config.go
cw9: hmm seems this timeout is applied to the service watch not the namespace one? Also do we plan to manually create the initial placement? could that info somehow be passed in through config and the process creates the placement itself?

client/config.go
```diff
@@ -153,27 +153,24 @@ func (c Configuration) NewAdminClient(
 
 	var err error
 	if envCfg.TopologyInitializer == nil {
-		switch {
-		case c.EnvironmentConfig.Service != nil:
-
+		if c.EnvironmentConfig.Service != nil {
```
robskillington: Just out of curiosity, what's the reasoning for switch to if change?

environment/config.go
```diff
@@ -51,6 +55,41 @@ type Configuration struct {
...
+	ListenClientUrls         []string               `yaml:"listenClientUrls"`
+	InitialCluster           []SeedNode             `yaml:"initialCluster"`
+	Name                     string                 `yaml:"name"`
+	ClientTransportSecurity  SeedNodeSecurityConfig `json:"clientTransportSecurity"`
+	PeerTransportSecurity    SeedNodeSecurityConfig `json:"peerTransportSecurity"`
+}
+
+// SeedNode represents a seed node for the cluster
+type SeedNode struct {
+	HostID   string `yaml:"hostId"`
```
robskillington: nit: We usually keep the same field names except lower camel cased, so perhaps `hostID` here?  Also think we use `hostID` instead of `hostId` in other structs we use for YAML parsing.

environment/config.go
```diff
@@ -51,6 +55,41 @@ type Configuration struct {
...
+type SeedNodeConfig struct {
+	RootDir                  string                 `yaml:"rootDir"`
+	InitialAdvertisePeerUrls []string               `yaml:"initialAdvertisePeerUrls"`
+	AdvertiseClientUrls      []string               `yaml:"advertiseClientUrls"`
+	ListenPeerUrls           []string               `yaml:"listenPeerUrls"`
+	ListenClientUrls         []string               `yaml:"listenClientUrls"`
+	InitialCluster           []SeedNode             `yaml:"initialCluster"`
+	Name                     string                 `yaml:"name"`
+	ClientTransportSecurity  SeedNodeSecurityConfig `json:"clientTransportSecurity"`
+	PeerTransportSecurity    SeedNodeSecurityConfig `json:"peerTransportSecurity"`
```
robskillington: Is there a reason we're using `json` field tags here?  Do we ever parse from JSON?  If not can we change these to `yaml` tags?

services/m3dbnode/config/config_test.go
```diff
@@ -225,19 +226,35 @@ pooling:
...
+        advertiseClientUrls:
+            - http://1.1.1.1:2379
+        initialCluster:
+            - hostId: host1
+              endpoint: http://1.1.1.1:2380
+            - hostId: host2
+              endpoint: http://1.1.1.2:2380
+            - hostId: host3
+              endpoint: http://1.1.1.3:2380
+        name: host1
```
robskillington: Hm this is meant to be the local `HostID`? Of this node?  If so we'll need to pass this in, as the top level config already has this and we won't want user's to have to define the local host ID twice:
https://github.com/m3db/m3db/blob/master/services/m3dbnode/config/config.go#L57

client/config.go
justinjc: Purely stylistic. There was a previous PR comment noting that using switch here felt a little odd, which I agreed with.

environment/config.go
justinjc: Nice catch, thanks. Typo'd here.

services/m3dbnode/config/config_test.go
justinjc: Yep, makes sense. Etcd config requires that the "seed node" `name` must equal one of the `name`s specified in the initialCluster config. After splitting the initialCluster into hostID/endpoint values, it makes more sense to just use the HostID of the node.

environment/config.go
```diff
@@ -95,112 +134,135 @@ type ConfigureResults struct {
...
+
+	return emptyConfig, errInvalidConfig
+}
 
+func (c Configuration) configureDynamic(configSvcClient client.Client, cfgParams ConfigurationParameters) (ConfigureResults, error) {
+	dynamicOpts := namespace.NewDynamicOptions().
+		SetInstrumentOptions(cfgParams.InstrumentOpts).
+		SetConfigServiceClient(configSvcClient).
+		SetNamespaceRegistryKey(kvconfig.NamespacesKey).
+		SetInitTimeout(cfgParams.NamespaceResolutionTimeout)
```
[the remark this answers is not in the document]
justinjc: I was able to set up a cluster of seed nodes talking to each other, although the flat duration wait does indeed make you "race" against the timeout duration. When the m3db node is started, it first sets up the embedded etcd (if it's a seed node) and the topology initializer will wait (block) the specified duration for a topology, during which you can issue your setup commands.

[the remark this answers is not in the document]
robskillington: Ah I see yeah.  Ok great, would be good to get this changed later (once this has landed) to avoid the need to race against the clock.  Good to know it doesn't block etcd startup.

services/m3dbnode/server/server.go
```diff
@@ -111,6 +117,61 @@ func Run(runOpts RunOptions) {
...
+		if len(clusters) == 0 {
+			endpoints, err := config.InitialClusterEndpoints(cfg.EnvironmentConfig.SeedNode.InitialCluster)
+
+			if err != nil {
+				logger.Fatalf("unable to create etcd clusters: %v", err)
+			}
+
+			zone := cfg.EnvironmentConfig.Service.Zone
+
+			logger.Infof("setting client etcd cluster to zone [%s], endpoints %v", zone, endpoints)
```
robskillington: nit: This is definitely a nit, but we tend to use `key=value, key=value` if we are embedding the info in the message itself.

For instance:
```go 
logger.Infof("using seed nodes etcd cluster: zone=%s, endpoints=%v", zone, endpoints)
```

services/m3dbnode/server/server.go
```diff
@@ -111,6 +117,61 @@ func Run(runOpts RunOptions) {
...
+
+			cfg.EnvironmentConfig.Service.ETCDClusters = []etcd.ClusterConfig{etcd.ClusterConfig{
+				Zone:      zone,
+				Endpoints: endpoints,
+			}}
+		}
+
+		if config.IsSeedNode(cfg.EnvironmentConfig.SeedNode.InitialCluster, cfg.EnvironmentConfig.SeedNode.Name) {
+			logger.Info("is a seed node; starting etcd server")
```
robskillington: nit: Use commas instead of semi colons, or if possible avoid.
```go
logger.Info("starting etcd server, running as seed node")
```

environment/config.go
```diff
@@ -51,6 +55,40 @@ type Configuration struct {
...
+}
+
+// SeedNodeSecurityConfig contains the data used for security in seed nodes
+type SeedNodeSecurityConfig struct {
+	CAFile        string `json:"caFile"`
+	CertFile      string `json:"certFile"`
+	KeyFile       string `json:"keyFile"`
+	TrustedCAFile string `json:"trustedCaFile"`
+	CertAuth      bool   `json:"clientCertAuth"`
+	AutoTLS       bool   `json:"autoTls"`
```
robskillington: These fields are all `json:` too still, update them to be `yaml:`?

environment/config.go
```diff
@@ -51,6 +55,40 @@ type Configuration struct {
 
 	// StaticConfiguration is used for running M3DB with a static config
 	Static *StaticConfiguration `yaml:"static"`
+
+	// Presence of a (etcd) server in this config denotes an embedded cluster
+	SeedNode *SeedNodeConfig `yaml:"seedNode"`
```
robskillington: nit: Because this is actually describes the seed nodes for the entire cluster, perhaps better to call it something like `seedNodes` with type `SeedNodesConfig`?

PR #491, opened 2018-03-08
Fix read_index_ids to use nanos as block start and use std bytes pool

robskillington:


richardartoul:
I think this was all addressed in #487

PR #492, opened 2018-03-09
Add tool to read data files

prateek:
/cc @richardartoul @robskillington

### DIFF
--- a/tools/read_data_files/README.md
+++ b/tools/read_data_files/README.md
@@ -1,0 +1,28 @@
+# read_data_files
+
+`read_data_files` is a utility to extract data for all timeseries' present in a TSDB file set.
+
+# Usage
+```
+$ git clone git@github.com:m3db/m3db.git
+$ make tools
+$ ./bin/read_data_files
+Usage: read_data_files [-b value] [-n value] [-p value] [-s value] [parameters ...]
+ -b, --block-start=value
+       Block Start Time [in nsec]
+ -f, --id-filter=value
+       ID Contains Filter [e.g. xyz]
+ -n, --namespace=value
+       Namespace [e.g. metrics]
+ -p, --path-prefix=value
+       Path prefix [e.g. /var/lib/m3db]
+ -s, --shard-id=value
+       Shard ID [expected format uint32]
+
+# example usage
+# read_data_files -b1480960800000000000 -n metrics -p /var/lib/m3db -s 451 -f 'metric-name' > /tmp/sample-data.out
+```
+
+# TBH
+- The tool outputs the identifiers to `stdout`, remember to redirect as desired.
+- The code currently assumes the data layout under the hood is `<path-prefix>/data/<namespace>/<shard-id>/...<block-start>-[index|...].db`. If this is not the file structure under the hood, replicate it to use this tool. Remember to copy checkpoint files along with each index file.
--- a/tools/read_data_files/README.md
+++ b/tools/read_data_files/README.md
@@ -5,7 +5,7 @@
 # Usage
 ```
 $ git clone git@github.com:m3db/m3db.git
-$ make tools
+$ make read_data_files
 $ ./bin/read_data_files
 Usage: read_data_files [-b value] [-n value] [-p value] [-s value] [parameters ...]
  -b, --block-start=value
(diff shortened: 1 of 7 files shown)


tools/read_data_files/README.md
```diff
@@ -0,0 +1,28 @@
+# read_data_files
+
+`read_data_files` is a utility to extract data for all timeseries' present in a TSDB file set.
+
+# Usage
+```
+$ git clone git@github.com:m3db/m3db.git
+$ make tools
```
richardartoul: Might be nice to just specify the exact command for building this particular tool

tools/read_data_files/main/main.go
```diff
@@ -0,0 +1,83 @@
...
+	"time"
+
+	"github.com/m3db/m3db/encoding"
+	"github.com/m3db/m3db/encoding/m3tsz"
+	"github.com/m3db/m3db/persist/fs"
+	"github.com/m3db/m3db/tools"
+	"github.com/m3db/m3x/ident"
+	xlog "github.com/m3db/m3x/log"
+
+	"github.com/pborman/getopt"
```
richardartoul: flags not good enough for you eh?

tools/read_data_files/main/main.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+func main() {
```
richardartoul: Possibly out of scope, but it would be nice to have the option to generate a valid CSV file with id, timestamp, dp

Doesn't need to be anything fancy, just printing properly formatted lines would be fine

tools/read_data_files/main/main.go
```diff
@@ -0,0 +1,83 @@
...
+			dp, _, _ := iter.Current()
+			// Use fmt package so it goes to stdout instead of stderr
+			fmt.Printf("{id: %s, dp: %+v}\n", id.String(), dp)
+		}
+		if err := iter.Err(); err != nil {
+			log.Fatalf("unable to iterate original data: %v", err)
+		}
+		iter.Close()
+
+		data.DecRef()
```
richardartoul: super nit, but might be worth benching it once to see if it runs faster without the finalizing

tools/read_data_files/README.md
```diff
@@ -0,0 +1,28 @@
...
+Usage: read_data_files [-b value] [-n value] [-p value] [-s value] [parameters ...]
+ -b, --block-start=value
+       Block Start Time [in nsec]
+ -f, --id-filter=value
+       ID Contains Filter [e.g. xyz]
+ -n, --namespace=value
+       Namespace [e.g. metrics]
+ -p, --path-prefix=value
+       Path prefix [e.g. /var/lib/m3db]
+ -s, --shard-id=value
```
richardartoul: I think our other tools does just call this shard? If so, would be nice to be consistent

tools/read_data_files/README.md
prateek: sure thing

tools/read_data_files/main/main.go
prateek: meh, this doesn't require an extra var decl so I like it better.

tools/read_data_files/main/main.go
prateek: Yeah the output right now isn't meant for anything but human consumption. E.g.

```
{id: servers.<redacted>memory.<..>, dp: {Timestamp:2018-03-07 00:09:00 +0000 UTC Value:2.04443207378e+11}}
```

It'd be nice to ensure we do the right thing with quoting/escaping for both json/csv output so we could pipe to jq/*nix tools, but meh. We can add that if/when we need it.

tools/read_data_files/main/main.go
prateek: With id filtering it runs within a few seconds for a 24h block so fast enough for debugging purposes already. If we needed more in depth checking, we'd probably be scripting in go anyway. Going to leave as is for now, don't want to spend more cycles tuning this.

tools/read_data_files/README.md
prateek: sure thing

tools/read_data_files/README.md
prateek: done

tools/read_data_files/README.md
prateek: turns out we had a weird mix of shard and shard-id in various tools, standardised to `shard` as suggested.

PR #498, opened 2018-03-13
[WIP] Pull in m3cluster with interface changes

cw9:
@prateek @robskillington

### DIFF
--- a/topology/types.go
+++ b/topology/types.go
@@ -255,10 +255,10 @@ type DynamicOptions interface {
 	ServiceID() services.ServiceID
 
 	// SetServiceOptions sets the options for service discovery.
-	SetServiceOptions(opts services.Options) DynamicOptions
+	SetServiceOptions(opts services.OverrideOptions) DynamicOptions
 
 	// ServicesOptions returns the options for service discovery.
-	ServicesOptions() services.Options
+	ServicesOptions() services.OverrideOptions
 
 	// SetQueryOptions sets the ConfigService query options
 	SetQueryOptions(value services.QueryOptions) DynamicOptions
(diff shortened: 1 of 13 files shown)


topology/types.go
```diff
@@ -255,10 +255,10 @@ type DynamicOptions interface {
 	ServiceID() services.ServiceID
 
 	// SetServiceOptions sets the options for service discovery.
-	SetServiceOptions(opts services.Options) DynamicOptions
+	SetServiceOptions(opts services.OverrideOptions) DynamicOptions
 
 	// ServicesOptions returns the options for service discovery.
-	ServicesOptions() services.Options
+	ServicesOptions() services.OverrideOptions
```
prateek: The name feels weird. Why `OverrideOptions` instead of `ServicesOptions`?

cw9: Because we used to have etcd.Options, now that all the code inside etcd/ got moved to services/, I have to rename that to services.Options because that contains all the options to create the SD client. the original services.Options is just an option that allows users to override namespace and stuff, thus the rename, also services.ServicesOptions stutters.

cw9:
I'll wait until the outstanding GH update diff in statsdex landed first before I land this thing

topology/types.go
prateek: Hm should we rename the field in this interface then (`OverrideOptions()` instead of `ServicesOptions()`)?

cw9: sounds good

topology/types.go
```diff
@@ -254,11 +254,11 @@ type DynamicOptions interface {
 	// ServiceID returns the ServiceID for service discovery
 	ServiceID() services.ServiceID
 
-	// SetServiceOptions sets the options for service discovery.
-	SetServiceOptions(opts services.Options) DynamicOptions
+	// SetServicesOverrideOptions sets the override options for service discovery.
+	SetServicesOverrideOptions(opts services.OverrideOptions) DynamicOptions
```
prateek: LGTM

PR #499, opened 2018-03-15
Set max finalizer pool capacity to 4

richardartoul:
Task to investigate different values later: https://github.com/m3db/m3db/issues/500

### DIFF
--- a/services/m3dbnode/config/config_test.go
+++ b/services/m3dbnode/config/config_test.go
@@ -143,6 +143,7 @@ pooling:
         size: 524288
         lowWatermark: 0.01
         highWatermark: 0.02
+        maxFinalizerCapacity: 4
     segmentReaderPool:
         size: 16384
         lowWatermark: 0.01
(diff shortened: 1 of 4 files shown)


services/m3dbnode/server/server.go
```diff
@@ -679,7 +679,8 @@ func withEncodingAndPoolingOptions(
 	contextPoolOpts := poolOptions(policy.ContextPool, scope.SubScope("context-pool"))
 	contextPool := context.NewPool(context.NewOptions().
 		SetContextPoolOptions(contextPoolOpts).
-		SetFinalizerPoolOptions(closersPoolOpts))
+		SetFinalizerPoolOptions(closersPoolOpts).
+		SetMaxPooledFinalizerCapacity(4))
```
prateek: mind making this config driven while you're here

prateek:
@richardartoul looks like the config unit test needs some changes (build is failing still)

services/m3dbnode/config/config_test.go
```diff
@@ -143,6 +143,7 @@ pooling:
         size: 524288
         lowWatermark: 0.01
         highWatermark: 0.02
+        maxFinalizerCapacity: 4
```
prateek: nit: could you change it to the non-default value

And add some a test to ensure we use the default value when nothing is set

PR #502, opened 2018-03-16
Fix Race Conditions in Shard

prateek:
- `tickAndExpire`s usual path of calling `purgeExpired` within the `forEachShardEntry` always skipped cleanup for entries present in the underlying batch; it cleaned up stuff in the lagging if check at the end. 
- A race in `forEachShardEntry`: took a reference to the front of the list but didn't increment the entry's reference count immediately, and released a lock. This was the bug allowing ticks to cleanup data while something else held a reference to it.
- Another race in `forEachShardEntry`: similar to the point above, we held onto an entry pointer with zero reference counts while transitioning from one batch to another, this pointer could get invalidated during a tick and cause panics. 
- Tests to ensure we are guarding against these issues

### DIFF
--- a/storage/shard.go
+++ b/storage/shard.go
@@ -57,9 +57,13 @@ import (
 
 const (
 	shardIterateBatchPercent = 0.01
-	expireBatchLength        = 1024
 )
 
+var (
+	// NB(prateek): the following is set to a var instead of a const because we
+	// override it in tests
+	expireBatchLength = 1024
+)
 var (
 	errShardEntryNotFound         = errors.New("shard entry not found")
 	errShardNotOpen               = errors.New("shard is not open")
@@ -182,6 +186,8 @@ func (entry *dbShardEntry) decrementReaderWriterCount() {
 }
 
 type dbShardEntryWorkFn func(entry *dbShardEntry) bool
+
+type dbShardEntryBatchWorkFn func(entries []*dbShardEntry) bool
 
 type shardFlushState struct {
 	sync.RWMutex
@@ -379,17 +385,49 @@ func (s *dbShard) OnEvictedFromWiredList(id ident.ID, blockStart time.Time) {
 }
 
 func (s *dbShard) forEachShardEntry(entryFn dbShardEntryWorkFn) error {
+	return s.forEachShardEntryBatch(func(currEntries []*dbShardEntry) bool {
+		for _, entry := range currEntries {
+			if continueForEach := entryFn(entry); !continueForEach {
+				return false
+			}
+		}
+		return true
+	})
+}
+
+func (s *dbShard) forEachShardEntryBatch(entriesBatchFn dbShardEntryBatchWorkFn) error {
 	// NB(r): consider using a lockless list for ticking.
 	s.RLock()
 	elemsLen := s.list.Len()
 	batchSize := int(math.Ceil(shardIterateBatchPercent * float64(elemsLen)))
-	nextElem := s.list.Front()
 	s.RUnlock()
 
-	// TODO(xichen): pool or cache this.
-	currEntries := make([]*dbShardEntry, 0, batchSize)
-	for nextElem != nil {
+	decRefElem := func(e *list.Element) {
+		if e == nil {
+			return
+		}
+		e.Value.(*dbShardEntry).decrementReaderWriterCount()
+	}
+
+	var (
+		currEntries = make([]*dbShardEntry, 0, batchSize) // TODO(xichen): pool or cache this.
+		first       = true
+		nextElem    *list.Element
+	)
+
+	for nextElem != nil || first {
 		s.RLock()
+		// NB(prateek): release held reference on the next element pointer now
+		// that we have the read lock and are guaranteed it cannot be changed
+		// from under us.
+		decRefElem(nextElem)
+
+		// lazily pull from the head of the list at first
+		if first {
+			nextElem = s.list.Front()
+			first = false
+		}
+
 		elem := nextElem
 		for ticked := 0; ticked < batchSize && elem != nil; ticked++ {
 			nextElem = elem.Next()
@@ -398,22 +436,26 @@ func (s *dbShard) forEachShardEntry(entryFn dbShardEntryWorkFn) error {
 			currEntries = append(currEntries, entry)
 			elem = nextElem
 		}
-		s.RUnlock()
-		for _, entry := range currEntries {
-			if continueForEach := entryFn(entry); !continueForEach {
-				// Abort early, decrement reader writer count for all entries first
-				for _, e := range currEntries {
-					e.decrementReaderWriterCount()
-				}
-				return nil
-			}
+
+		// NB(prateek): inc a reference to the next element while we have a lock,
+		// to guarantee the element pointer cannot be changed from under us.
+		if nextElem != nil {
+			nextElem.Value.(*dbShardEntry).incrementReaderWriterCount()
 		}
+		s.RUnlock()
+
+		continueExecution := entriesBatchFn(currEntries)
 		for i := range currEntries {
 			currEntries[i].decrementReaderWriterCount()
 			currEntries[i] = nil
 		}
 		currEntries = currEntries[:0]
+		if !continueExecution {
+			decRefElem(nextElem)
+			return nil
+		}
 	}
+
 	return nil
 }
 
@@ -523,7 +565,6 @@ func (s *dbShard) tickAndExpire(
 
 	var (
 		r                             tickResult
-		expired                       []series.DatabaseSeries
 		terminatedTickingDueToClosing bool
 		i                             int
 		slept                         time.Duration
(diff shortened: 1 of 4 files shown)


storage/shard.go
```diff
@@ -357,13 +361,19 @@ func (s *dbShard) forEachShardEntry(entryFn dbShardEntryWorkFn) error {
...
+		currEntries = make([]*dbShardEntry, 0, batchSize) // TODO(xichen): pool or cache this.
+		first       = true
+		nextElem    *list.Element
+	)
+	for nextElem != nil || first {
 		s.RLock()
+		if first {
+			nextElem = s.list.Front()
+			first = false
+		}
```
robskillington: If you wanted to get cute, this seems to work:


Checked using the playground:

prateek: Missing the paste/links here

storage/shard.go
```diff
@@ -537,7 +547,7 @@ func (s *dbShard) tickAndExpire(
 			err = series.ErrSeriesAllDatapointsExpired
 		}
 		if err == series.ErrSeriesAllDatapointsExpired {
-			expired = append(expired, entry.series)
+			expired = append(expired, entry)
```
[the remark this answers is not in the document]
prateek: Good catch on the issue! 

I'm thinking of doing it slightly differently: expose a new fn `forEachShardEntryBatch` which is used in the implementation of `forEachShardEntry`. Re-work `tickAndExpire` to use the new Batch fn, and then simplify the lifecycle of expired slice in `tickAndExpire` to be contained within the tick batch (as opposed to across batches). It'll make the code simpler in a couple of different places. 

One negative of this is: it doesn't allow accumulation of expired slices greater than the length of a tickBatch. But I think that's an OK trade-off cause we already tried to clear the expired slice every 1024 elements, it going down to 512 (the default tickBatchSize) isn't that big a deal.

storage/shard.go
```diff
@@ -353,17 +359,49 @@ func (s *dbShard) OnRetrieveBlock(
...
+	var (
+		currEntries = make([]*dbShardEntry, 0, batchSize) // TODO(xichen): pool or cache this.
+		first       = true
+		nextElem    *list.Element
+	)
+
+	for nextElem != nil || first {
 		s.RLock()
+		// NB(prateek): release held reference on the next element pointer now
+		// that we have the read lock and are guarantee it cannot be chaned
```
richardartoul: nit: changed

storage/shard.go
```diff
@@ -353,17 +359,49 @@ func (s *dbShard) OnRetrieveBlock(
...
 	s.RLock()
 	elemsLen := s.list.Len()
 	batchSize := int(math.Ceil(shardIterateBatchPercent * float64(elemsLen)))
-	nextElem := s.list.Front()
 	s.RUnlock()
 
-	// TODO(xichen): pool or cache this.
-	currEntries := make([]*dbShardEntry, 0, batchSize)
-	for nextElem != nil {
+	decRefElem := func(e *list.Element) {
```
richardartoul: nit: an allocation ever 1024 entries isn't much, but seems like you could just make this a method that you re-used since you're not creating any closures here

storage/shard.go
richardartoul: nit: guaranteed

storage/shard.go
```diff
@@ -506,96 +547,107 @@ func (s *dbShard) tickAndExpire(
...
 			continue
 		}
-		entry := elem.Value.(*dbShardEntry)
 		// If this series is currently being written to or read from, we don't
 		// remove it even though it's empty in that it might become non-empty
 		// soon.
 		// We avoid closing the series during a read to ensure a consistent view
 		// of the series if they manage to take a reference to it.
-		if entry.readerWriterCount() > 0 {
+		if entry.readerWriterCount() > 1 {
```
richardartoul: I think @r asked about this too, but why 1 instead of 0?

storage/shard_race_prop_test.go
```diff
@@ -0,0 +1,145 @@
...
+	seed := time.Now().UnixNano()
+	parameters.MinSuccessfulTests = 200
+	parameters.MaxSize = 40
+	parameters.MaxDiscardRatio = 20
+	parameters.Rng = rand.New(rand.NewSource(seed))
+	properties := gopter.NewProperties(parameters)
+
+	properties.Property("Concurrent Tick and Shard Fn doesn't panic", prop.ForAll(
+		func(ids []ident.ID, tickBatchSize uint8, expireBatchLength uint8, fn testShardReadFn) bool {
+			// fmt.Printf("ids - %d; tickBatchSize: %d ; expireLen: %d\n", len(ids), tickBatchSize, expireBatchLength)
```
richardartoul: nit: remove?

storage/shard.go
prateek: re the allocation: I don't think it's ever going to grow much beyond 2K entries with our current prod workload (app/hist) so this won't ever be a bottleneck, but didn't want to drop the TODO without definitive proof. 

re the method: I considered doing that but liked having the method be self contained. I verified the function is marked safe for inlining so there's no performance cost: 

```
❯ go build -gcflags '-m' ./storage 2>&1 | grep 'shard.go' | sort | grep -e ':379'
storage/shard.go:379:16: (*dbShard).forEachShardEntryBatch func literal does not escape
storage/shard.go:379:16: can inline (*dbShard).forEachShardEntryBatch.func1
storage/shard.go:379:24: leaking param content: e
```

storage/shard.go
prateek: This addresses a bug in the original code: When iterating over a batch of elements, `forEachShardEntry' (and now `forEachShardEntryBatch`) inc the `readerWriterCount` of the element, and then dec the count once they're done with a batch. The original `tickAndExpire` called this function `purgeExpired...` within the function passed to `forEachShardEntry`. Any elements in the `expired` list being purged, which were in the current underlying batch used by `forEachShardEntry` were guaranteed to have a readerWriterCount > 0 (at least 1) and we'd skip expiring them pessimistically. TBH: the original code maintained the `expired` list across `forEachShardEntry` calls, so it wasn't entirely fked. 

The new code maintains `expired` only within each underlying batch used by `forEachShardEntryBatch`. This is why we can use the `> 1` check and be more accurate about cleanup.

storage/shard_race_prop_test.go
prateek: sure

storage/shard.go
prateek: fixed

storage/shard.go
```diff
@@ -532,96 +573,107 @@ func (s *dbShard) tickAndExpire(
...
 	s.RUnlock()
-	s.forEachShardEntry(func(entry *dbShardEntry) bool {
-		if i > 0 && i%tickSleepBatch == 0 {
-			// NB(xichen): if the tick is cancelled, we bail out immediately.
-			// The cancellation check is performed on every batch of entries
-			// instead of every entry to reduce load.
-			if c.IsCancelled() {
-				return false
+	s.forEachShardEntryBatch(func(currEntries []*dbShardEntry) bool {
+		var expired []*dbShardEntry
```
robskillington: `expired = expired[:0]` and reuse this slice perhaps?

storage/shard.go
```diff
@@ -532,96 +573,107 @@ func (s *dbShard) tickAndExpire(
...
-				// and ensure smooth rate of elements being returned to pools.
-				// This method does not run using a lock so this is safe to
-				// perform inline.
-				s.purgeExpiredSeries(expired)
-				for i := range expired {
-					expired[i] = nil
+			if err == series.ErrSeriesAllDatapointsExpired {
+				expired = append(expired, entry)
+				r.expiredSeries++
+				if len(expired) >= expireBatchLength {
```
robskillington: As discussed, let's perhaps remove the limit of expiring in smaller batches, if we interrupt less that's probably better anyhow.

storage/shard.go
richardartoul: I'm still not totally following, but could you add a comment explaining it? It looks like something that would be tempting for someone to "fix" right now

storage/shard.go
prateek: sure thing.

storage/shard.go

storage/shard.go

storage/shard.go

storage/shard.go

storage/shard.go
prateek: Put down a verbose comment and a failing if block to explicitly callout the non-expected condition. Does it make more sense now?

storage/shard.go
```diff
@@ -523,105 +569,118 @@ func (s *dbShard) tickAndExpire(
...
 		if !exists {
 			continue
 		}
-		entry := elem.Value.(*dbShardEntry)
+
+		count := entry.readerWriterCount()
+		// The contract requires all entries to have count >= 1.
+		if count < 1 {
+			s.logger.Errorf(`observed series [%s] with readerWriterCount [%d]. This violates
+				expected guarantees in the code.`, series.ID().String(), count)
```
robskillington: Hm this will actually print an extra end line in the logs and extra spacing in front of word "expected".

Better to just split with concat perhaps?

storage/shard.go
```diff
@@ -379,17 +381,57 @@ func (s *dbShard) OnEvictedFromWiredList(id ident.ID, blockStart time.Time) {
...
+				return false
+			}
+		}
+		return true
+	})
+}
+
+func iterateBatchSize(elemsLen int) int {
+	if elemsLen < shardIterateBatchMinSize {
+		return elemsLen
```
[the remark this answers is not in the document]
prateek: We do have that: there's a math.Max in the return path. i.e. we're guaranteed to have a batch size of at least `shardIterateBatchMinSize`  if `elemsLen>=shardIterateBatchMinSize`. There's a test that checks exactly the situation you're worried about btw - 
```
func TestShardIterateBatchSize(t *testing.T) {
...
	require.Equal(t, shardIterateBatchMinSize, iterateBatchSize(shardIterateBatchMinSize+1))
...
}
```

storage/shard.go
prateek: Sure thing. Reworked to do the concat with the `Errorf` format string in the next patch.

storage/shard.go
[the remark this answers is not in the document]
robskillington: Ah yes, my bad - I did not look closely at the fallthrough below (and missed the `math.Max`).  Thanks!

storage/shard.go
```diff
@@ -523,105 +569,118 @@ func (s *dbShard) tickAndExpire(
...
 		elem, exists := s.lookup[hash]
 		if !exists {
 			continue
 		}
-		entry := elem.Value.(*dbShardEntry)
+
+		count := entry.readerWriterCount()
+		// The contract requires all entries to have count >= 1.
+		if count < 1 {
+			s.logger.Errorf("observed series [%s] with readerWriterCount [%d]. %s",
```
robskillington: super super nit: I'd usually use `WithFields(..)` here, at the very least for the series ID as that's what I'd want if I was searching this as structured logs, but anyway not a big deal.

prateek: Fair point. I'll make the change.

PR #503, opened 2018-03-17
Include build age in ld-flags

prateek:


prateek:
#510 pulls these changes

PR #506, opened 2018-03-19
[WIP] Shard Tick/Write race

prateek:
Wrote up a test that inserts and ticks at the same time. Can get it to hang after a few iterations of execution. Not sure what the issue: bad test setup / code race ?

### DIFF
--- a/storage/shard.go
+++ b/storage/shard.go
@@ -868,6 +868,11 @@ func (s *dbShard) writableSeries(id ident.ID) (*dbShardEntry, error) {
 
 		// Wait for the insert attempt
 		result.wg.Wait()
+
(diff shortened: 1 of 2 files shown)


robskillington:
Hm interesting, I can dig in here.

richardartoul:
@prateek didn't we resolve this?

prateek:
@richardartoul Nah, we resolved a couple of races during concurrent Read/Tick; but there's something funky going on during concurrent Write/Tick as well - the attached test fails periodically.

robskillington:
As per https://github.com/m3db/m3/pull/2128 this can finally be closed, yay!

PR #507, opened 2018-03-20
Integrate `databaseIndex` into WriteTagged path 

prateek:
- Wires up `databaseIndex` to the WriteTagged path in Database -> Namespace -> Shard 

NB: In an effort to keep reviews small, i'm using `FOLLOWUP(prateek)` littered in the code to indicate changes that will follow in subsequent PRs.


storage/shard.go
```diff
@@ -762,21 +814,36 @@ func (s *dbShard) Write(
...
 		// overhead of ownership tracking. This makes taking a ref here safe.
 		commitLogSeriesID = entry.series.ID()
+		commitLogSeriesTags = entry.series.Tags()
 		commitLogSeriesUniqueIndex = entry.index
-		entry.decrementReaderWriterCount()
 		if err != nil {
+			entry.decrementReaderWriterCount()
 			return err
 		}
+		needsIndex := indexWriteFn != nil && entry.needsIndexUpdate(timestamp)
```
prateek: NB: this function (`writeAndIndex`) is pretty involved so I tried breaking it into smaller pieces, can see the changes it'd take [here](https://github.com/m3db/m3db/commit/668bd3e92b1b40a3634a469ac89d1a8429b6f1cb). 

But that suffers from poor in-lining. Even with `-l=4` it still won't inline any function that calls an interface so it'd affect performance. How much impact I don't know. Can test impact if people think it's worth going down the road.

persist/fs/commitlog/types.go
```diff
@@ -99,6 +99,9 @@ type Series struct {
 	// ID is the series identifier
 	ID ident.ID
 
+	// Tags are the series tags
+	Tags ident.Tags // FOLLOWUP(prateek): wire Tags to commit log writer
```
richardartoul: We should sync on this as well because this will probably involve changing the commit log format

richardartoul: Or can you just throw it in the "annotations" section?

storage/namespace.go
```diff
@@ -268,7 +278,7 @@ func newDatabaseNamespace(
 		log:                    logger,
 		increasingIndex:        increasingIndex,
 		commitLogWriter:        commitLogWriter,
-		reverseIndex:           nil, // FOLLOWUP(prateek): will be done in https://github.com/m3db/m3db/pull/507
+		reverseIndex:           index,
```
richardartoul: little weird to call this field reverse index when we dont use the word reverse anywhere else

storage/namespace.go
```diff
@@ -941,11 +955,15 @@ func (n *dbNamespace) readableShardAtWithRLock(shardID uint32) (databaseShard, e
...
+	var indexFn namespaceIndexWriteFn
+	if n.reverseIndex != nil {
+		indexFn = n.reverseIndex.Write
+	}
 	shards := n.shardSet.AllIDs()
 	dbShards := make([]databaseShard, n.shardSet.Max()+1)
 	for _, shard := range shards {
 		dbShards[shard] = newDatabaseShard(n.metadata, shard, n.blockRetriever,
-			n.namespaceReaderMgr, n.increasingIndex, n.commitLogWriter, n.reverseIndex,
+			n.namespaceReaderMgr, n.increasingIndex, n.commitLogWriter, indexFn,
```
richardartoul: Is this literally just a function? Might be easier to maintain in the future if its a narrow interface

storage/options.go
```diff
@@ -555,6 +582,7 @@ func (o *options) MultiReaderIteratorPool() encoding.MultiReaderIteratorPool {
 
 func (o *options) SetIdentifierPool(value ident.Pool) Options {
 	opts := *o
+	opts.indexOpts = opts.indexOpts.SetIdentifierPool(value)
```
richardartoul: This is non-standard, right? Any reason you're updating both and not just expecting IndexOptions with a correctly configured identifier pool?

storage/shard.go
```diff
@@ -168,6 +167,12 @@ type dbShardEntry struct {
 	series         series.DatabaseSeries
 	index          uint64
 	curReadWriters int32
+	reverseIndex   struct {
```
richardartoul: same comment about the name, field name is weird since we just refer to the index itself as an index

storage/shard.go
```diff
@@ -732,7 +777,13 @@ func (s *dbShard) Write(
 	// If no entry and we are not writing new series asynchronously
 	if !writable && !opts.writeNewSeriesAsync {
 		// Avoid double lookup by enqueueing insert immediately
-		result, err := s.insertSeriesAsyncBatched(id, dbShardInsertAsyncOptions{})
+		result, err := s.insertSeriesAsyncBatched(id, tags, dbShardInsertAsyncOptions{
+			hasPendingIndexing: indexWriteFn != nil,
```
richardartoul: the nil check seems a little weird, can we just throw an error if its nil?

storage/shard.go
```diff
@@ -720,6 +751,20 @@ func (s *dbShard) Write(
 	value float64,
 	unit xtime.Unit,
 	annotation []byte,
+) error {
+	return s.writeAndIndex(ctx, id, ident.EmptyTagIterator, timestamp,
+		value, unit, annotation, nil)
+}
+
+func (s *dbShard) writeAndIndex(
```
richardartoul: This function is pretty similar to the write() function, but I assume its hard to DRY it up? Maybe just leave a comment on them that if you make a change to one to look at the other

storage/shard.go
```diff
@@ -906,18 +975,37 @@ func (s *dbShard) tryRetrieveWritableSeries(id ident.ID) (
 	return nil, opts, nil
 }
 
-func (s *dbShard) newShardEntry(id ident.ID) *dbShardEntry {
+func (s *dbShard) newShardEntry(id ident.ID, tags ident.TagIterator) (*dbShardEntry, error) {
+	clonedTags, err := s.cloneTags(tags)
```
richardartoul: Why do the tags need to be cloned? Can we just not finalize them with the context?

persist/fs/commitlog/types.go
prateek: Talked offline: going to change the format. Will sync again when we reach the point of making this change.

storage/namespace.go
prateek: happy to rename, any suggestions?

storage/namespace.go
prateek: happy to rename, any suggestions?

storage/options.go
prateek: we do stuff like this in a couple of places in this file - e.g. `SetClockOptions`: 

```
func (o *options) SetClockOptions(value clock.Options) Options {
	opts := *o
	opts.clockOpts = value
	opts.commitLogOpts = opts.commitLogOpts.SetClockOptions(value)
	opts.seriesOpts = NewSeriesOptionsFromOptions(&opts, nil)
	return &opts
}
```

storage/shard.go
prateek: I wanted to use `index` but the entry already uses that for the `UniqueIndex` int64. What would you like it called instead?

storage/shard.go
prateek: tl;dr - Unfortunately, no. 

Longer version - The `indexWriteFn` value can be nil in two situations: 
1) With indexing enabled, and a user calling `db.Write()` (not `db.WriteTagged()`). Need to indicate don't need to index in those situations. 
2) With indexing disabled, and we have code that calls `shard.WriteTagged()` directly. This will never happen in usual circumstances as `namespace.WriteTagged` checks if indexing is enabled before calling `shard.WriteTagged()`.

storage/shard.go
prateek: I don't follow. `Write()` and `WriteTagged` both call `writeAndIndex()` internally for re-use.

Which write() are you talking about?

storage/shard.go
prateek: same reason we need clone the ID (two lines below this): the lifecycle of the underlying bytes is tied to the context which is finalised after the server side RPC response. Need to clone to have manage the lifecycle independently.

storage/shard.go
```diff
@@ -980,13 +1074,29 @@ func (s *dbShard) insertSeriesSync(
...
+	}
+
 	if s.newSeriesBootstrapped {
 		if err := entry.series.Bootstrap(nil); err != nil {
 			entry = nil // Don't increment the writer count for this series
 			return nil, err
 		}
 	}
+
+	if fn := s.indexWriteFn; fn != nil {
```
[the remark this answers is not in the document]
prateek: I don't have strong feelings either way. Will change as suggested.

storage/namespace.go
prateek: The namespace owns the full reverse index type (with all the write/read methods). The shard is only given a `databaseIndexWriteFn`, defined as:

```
type databaseIndexWriteFn func(
	namespace ident.ID,
	id ident.ID,
	tags ident.Tags,
	fns indexInsertLifecycleHooks,
) error
```

That's kinda what you were going for, yea?

storage/shard.go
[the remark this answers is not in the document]

prateek:
@richardartoul re: the integration tests. Nah we can add write only tests to ensure that doesn't mess up at least. Going to put up a couple in a subsequent PR (which wires up the server side WriteTagged RPC path to the storage WriteTagged path): [1] [2]

[1]: https://github.com/m3db/m3db/blob/67b89229bd340199e52f9935ba935c42d56ed37f/services/m3dbnode/main/main_index_test.go
[2]: https://github.com/m3db/m3db/blob/67b89229bd340199e52f9935ba935c42d56ed37f/integration/index_large_cardinality_high_concurrency_test.go

storage/namespace.go
richardartoul: can we just call it `index`? I'm not opposed to calling it `reverseIndex` just seems like it should be consistent with whatever we're calling the index interface

storage/namespace.go
richardartoul: I mean something like

```golang
type databaseIndex interface {
    Write(
        namespace ident.ID,
	id ident.ID,
	tags ident.Tags,
	fns indexInsertLifecycleHooks,
    ) error
}
```

And then if in the future we ever need to add another method (for whatever reason) its easier

storage/options.go
richardartoul: Yeah fair enough, especially after our debugging session today

storage/shard.go
richardartoul: ah....in that case maybe keep calling this `reverseIndex` and rename the other one `uniqueIndex` so its not ambiguous. Although now that you've pointed that out, I'm not sure if I want you to rename the reverseIndex field in the namespace anymore

storage/shard.go
richardartoul: Cool, can you just throw in a comment on when this can be nil then

storage/shard.go
richardartoul: Looking back at the code, I have no idea what I was talking about. Maybe I was looking at an old version or something

storage/shard.go
richardartoul: Yeah it just seems like we could not tie the lifecycle of the bytes to the context (and just let them leak in the rare error case) but I understand that this is safer

storage/shard.go
prateek: sure

storage/shard.go
prateek: lol yeah this pretty much the thought process that led to me to name both of them `reverseIndex`. Will leave as is unless there's a better suggestion.

storage/namespace.go
prateek: done. also renamed the associated types to `reverseIndexWriter` and `reverseIndexWriterFn` to align better with the naming in shard.go

storage/shard.go
prateek: The full write path allocation is a little sketchier than my original comment lets on: the bytes underlying the ident's here (in `newShardEntry`) are actually allocated by tchannel, we wrap them up into checked.Bytes (and then an ident) using a CheckedBytesWrapperPool in the m3db server side code. The context finalizing just releases the references the ident.ID and the checked.Bytes have to the underlying tchannel bytes slice.

prateek: Going to stick to the simple version of cloning (as we do now), and enhance in subsequent PRs based on benchmarks (as needed)

storage/shard.go
prateek: Lol fair enough

storage/shard.go

storage/namespace.go

storage/namespace.go
prateek: leaving as `reverseIndex` per the other discussion

prateek:
Haven't rolled out a version to the test cluster yet. Will do so after the WriteTagged RPC path is wired up (next PR after these two land)

storage/series/series.go
```diff
@@ -658,6 +667,7 @@ func (s *dbSeries) Reset(
 	defer s.Unlock()
 
 	s.id = id
+	s.tags = tags
```
robskillington: Can you also set `s.tags = nil` in the series Close method?

storage/shard.go
```diff
@@ -182,6 +187,22 @@ func (entry *dbShardEntry) decrementReaderWriterCount() {
...
+// ensure dbShardEntry satisfies the `onIndexSeries` interface.
+var _ onIndexSeries = &dbShardEntry{}
+
+func (entry *dbShardEntry) OnIndexSuccess(nextWriteTime time.Time) {
+	atomic.StoreInt64(&entry.reverseIndex.nextWriteTimeNanos, nextWriteTime.UnixNano())
+}
+
+func (entry *dbShardEntry) OnIndexFinalize() {
+	// indicate the index has released held reference for provided write
+	entry.decrementReaderWriterCount()
```
robskillington: Can we perhaps have a synchronous API to `OnIndexFinalize(...)` and we can keep it private too so others don't need to see it, i.e. `onIndexPrepare(...) { entry.incrementReaderWriterCount() }`?   That way its clear why you would decrement here.

storage/shard.go
```diff
@@ -732,7 +777,13 @@ func (s *dbShard) Write(
 	// If no entry and we are not writing new series asynchronously
 	if !writable && !opts.writeNewSeriesAsync {
 		// Avoid double lookup by enqueueing insert immediately
-		result, err := s.insertSeriesAsyncBatched(id, dbShardInsertAsyncOptions{})
+		result, err := s.insertSeriesAsyncBatched(id, tags, dbShardInsertAsyncOptions{
+			hasPendingIndexing: reverseIndexWriteFn != nil,
+			pendingIndex: dbShardPendingIndex{
+				timestamp: timestamp,
+				fn:        reverseIndexWriteFn,
```
robskillington: Yeah this would be nicer if you just pass a `databaseIndex` interface here rather than a plain fn, much more flexible in the future if there are other methods we need to call on the index.

storage/shard.go
```diff
@@ -762,21 +814,36 @@ func (s *dbShard) Write(
...
 		}
+		needsIndex := reverseIndexWriteFn != nil && entry.needsIndexUpdate(timestamp)
+		if needsIndex {
+			// NB(prateek): we retain the ref count on the entry while the indexing is pending,
+			// the callback executed on the entry once the indexing is completed releases this
+			// reference.
+			reverseIndexWriteFn(entry.series.ID(), entry.series.Tags(), entry)
+		} else {
+			entry.decrementReaderWriterCount()
+		}
```
robskillington: To simplify the code a little above, this looks slightly cleaner:
```go
		indexSeries := reverseIndexWriteFn != nil && entry.needsIndexUpdate(timestamp)
		if err == nil && indexSeries {
			// OnIndexFinalize will decrement the reader/writer
			reverseIndexWriteFn(entry.series.ID(), entry.series.Tags(), entry)
		} else {
			entry.decrementReaderWriterCount()
		}
		if err != nil {
			return err
		}
```

storage/shard.go
```diff
@@ -1040,28 +1153,44 @@ func (s *dbShard) insertSeriesBatch(inserts []dbShardInsert) error {
...
+
+		if inserts[i].opts.hasPendingIndexing {
+			pendingIndex := inserts[i].opts.pendingIndex
+			releaseEntryRef = true
+			// only index any entry that hasn't crossed the nextIndexTime
+			if entry.needsIndexUpdate(pendingIndex.timestamp) {
+				// NB(prateek): we index entries asynchronously, so if we are going to
+				// index the entry, we inc an extra ref. It is released by the indexing
+				// function callback - `entry.OnIndexFinalize`.
+				entry.incrementReaderWriterCount()
```
robskillington: This is where it would be nice if we had that synchronous method I mentioned, i.e. `entry.onIndexPrepare(...)`

storage/shard.go
robskillington: Actually on second thoughts, why don't we leave the code like it was but do a second increment if we need to index, similar to the async batch insert?  Seems better perhaps.  Then it would also be more clean if we also added the synchronous method name, i.e. `entry.onIndexPrepare()` which would increment the reader/writer count.

storage/shard.go
```diff
@@ -1040,28 +1153,44 @@ func (s *dbShard) insertSeriesBatch(inserts []dbShardInsert) error {
...
+				pendingIndex.fn(entry.series.ID(), entry.series.Tags(), entry)
```
robskillington: Is there any reason you need to pass the fn into the shard insert queue?  Couldn't you instead just `s.reverseIndex.Write(id, tags, entry)` here?

storage/namespace.go
prateek: The query will be transparent to users of the index, the write method will hide it.

storage/series/series.go
prateek: sure thing

storage/shard.go

storage/shard.go
prateek: Sure, will do

storage/shard.go
prateek: sure, will do

storage/shard.go
prateek: Yep, you're right. Will do

storage/shard.go

storage/shard.go

storage/shard.go

storage/series/series.go
prateek: done

storage/shard.go

storage/shard.go
prateek: Can avoid entirely here because of your other point

PR #510, opened 2018-03-20
Update ci-scripts; regen proto

justinjc:
Update to latest ci-scripts to use gogo/protobuf for code generation. Checking in modified files after running `make all-gen`.


prateek:
@justinjc is stuff serialized with `protoc-gen-gogo` compatible with `protoc-gen-go` and vice versa?

justinjc:
@prateek Unfortunately, not in all cases. However, inspecting the generated code here, there's just a change in variable names and the addition of some new methods, e.g. `MarshalTo`, `Size`. These methods are the ones I'm interested in (for pulling into m3coordinator).

prateek:
API/code changes I'm not super worried about. 

Serialisation/[]byte incompatibility is a bigger deal. We store protobuf in a []byte serialized form for namespace config in KV, and in the peer bootstrapping codepath. So if []byte output from `protoc-gen-go` doesn't work with the `protoc-gen-gogo` structs, the nodes won't be able to use existing config/talk to each other correctly. Could you confirm that's the case?

justinjc:
Good point. I've verified this for `namespace` and `pagetoken` by:
1. Generating `.pb.go` files using both `protoc-gen-go` and `protoc-gen-gogo`
1. Marshalling a `go` object and unmarshalling it into a `gogo` object and comparing fields
1. Vice versa of the above
Looks good compatibility-wise!

PR #512, opened 2018-03-21
Prevent peer incremental peer bootstrapper from panic'ing when encountering corrupt fileset files

### LINKED ISSUE #511: Make peer bootstrapper incremental flush handle case where fileset files already exist
Currently, if something goes wrong when trying to read a set of fileset files, that shard/block combination will be marked as unfulfilled, and the peer bootstrapper will try to incrementally flush it.

If the Prepare() method is called for a shard/block combination that already exists, the persist object that is returned will be nil and M3DB will panic during the peer bootstrapping process because it doesn't do the nil check.

We should make the following changes:

1) Re-write the method to return (prepared.Persist, bool, error) instead of just (prepared.Persist, error)
2) Refactor the peer bootstrapper to handle this situation and overwrite the existing files
3) Address other places in the code where this could potentially happen

richardartoul:
fixes #511

### DIFF
--- a/persist/fs/files.go
+++ b/persist/fs/files.go
@@ -256,6 +256,15 @@ func FilesetAt(filePathPrefix string, namespace ident.ID, shard uint32, t time.T
 	return filesAt(matched, t)
 }
 
+// DeleteFilesetAt deletes all fileset files whose timestamps exactly match a given time.
+func DeleteFilesetAt(filePathPrefix string, namespace ident.ID, shard uint32, t time.Time) error {
+	matched, err := FilesetAt(filePathPrefix, namespace, shard, t)
+	if err != nil {
+		return err
+	}
+	return DeleteFiles(matched)
+}
+
 // DeleteInactiveDirectories deletes any directories that are not currently active, as defined by the
 // inputed active directories within the parent directory
 func DeleteInactiveDirectories(parentDirectoryPath string, activeDirectories []string) error {
--- a/persist/fs/files.go
+++ b/persist/fs/files.go
@@ -47,6 +47,8 @@ const (
 
 type fileOpener func(filePath string) (*os.File, error)
 
+type filterFilesByTimePred func(t time.Time) (include, continueScanning bool)
+
 func openFiles(opener fileOpener, fds map[string]**os.File) error {
 	var firstErr error
 	for filePath, fdPtr := range fds {
@@ -245,6 +247,15 @@ func FilesetBefore(filePathPrefix string, namespace ident.ID, shard uint32, t ti
 	return filesBefore(matched, t)
 }
 
+// FilesetAt returns all the fileset files whose timestamps exactly match a given time.
+func FilesetAt(filePathPrefix string, namespace ident.ID, shard uint32, t time.Time) ([]string, error) {
+	matched, err := filesetFiles(filePathPrefix, namespace, shard, filesetFilePattern)
+	if err != nil {
+		return nil, err
+	}
+	return filesAt(matched, t)
+}
+
 // DeleteInactiveDirectories deletes any directories that are not currently active, as defined by the
 // inputed active directories within the parent directory
 func DeleteInactiveDirectories(parentDirectoryPath string, activeDirectories []string) error {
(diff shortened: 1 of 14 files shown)


storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -347,17 +353,40 @@ func (s *peersSource) incrementalFlush(
...
 
 	for start := tr.Start; start.Before(tr.End); start = start.Add(blockSize) {
-		prepared, err := flush.Prepare(nsMetadata, shard, start)
+		prepared, ok, err := flush.Prepare(nsMetadata, shard, start)
 		if err != nil {
 			return err
 		}
 
+		if !ok {
+			// If we reach this point then it means that the fileset files already
```
prateek: Few different things:
- Deleting files unconditionally in a bootstrapper scares me. Could you include a knob in the bootsrapper options to allow users to configure whether data should be overwritten or not.
- This assumes a bunch about how we're doing bootstrapping today (always doing FS before Peers). What if we decide to run with other bootstrapping configurations in the future?

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -347,17 +353,40 @@ func (s *peersSource) incrementalFlush(
...
 	}
 
 	for start := tr.Start; start.Before(tr.End); start = start.Add(blockSize) {
-		prepared, err := flush.Prepare(nsMetadata, shard, start)
+		prepared, ok, err := flush.Prepare(nsMetadata, shard, start)
 		if err != nil {
 			return err
 		}
 
+		if !ok {
```
prateek: would be to make this explicit: either have `flush.Prepare` return a typed error (`ErrFilesetAlreadyExists`), or  check `fs.FilesetExistsAt` here

persist/types.go
```diff
@@ -53,7 +53,7 @@ type Flush interface {
 	// Prepare prepares writing data for a given (shard, blockStart) combination,
 	// returning a PreparedPersist object and any error encountered during
 	// preparation if any.
-	Prepare(ns namespace.Metadata, shard uint32, blockStart time.Time) (PreparedPersist, error)
+	Prepare(ns namespace.Metadata, shard uint32, blockStart time.Time) (PreparedPersist, bool, error)
```
prateek: would it be better to encode the exists semantics by returning a typed error?  

i.e. `Prepare(...)` returns `nil, ErrFilesetExists` when a file exists, so err != nil is sufficient to guarantee the returned `PreparedPersist` is valid.

services/m3dbnode/config/bootstrap.go
```diff
@@ -106,14 +114,7 @@ func (bsc BootstrapConfiguration) New(
...
 		case fs.FileSystemBootstrapperName:
-			fsopts := opts.CommitLogOptions().FilesystemOptions()
-			filePathPrefix := fsopts.FilePathPrefix()
-			fsbopts := fs.NewOptions().
-				SetResultOptions(rsopts).
-				SetFilesystemOptions(fsopts).
-				SetNumProcessors(bsc.fsNumProcessors()).
-				SetDatabaseBlockRetrieverManager(opts.DatabaseBlockRetrieverManager())
-			bs = fs.NewFileSystemBootstrapper(filePathPrefix, fsbopts, bs)
+			bs = fs.NewFileSystemBootstrapper(filePathPrefix, bfsOpts, bs)
```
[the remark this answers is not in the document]
richardartoul: sure, can change that

persist/fs/files.go
```diff
@@ -245,6 +247,24 @@ func FilesetBefore(filePathPrefix string, namespace ident.ID, shard uint32, t ti
 	return filesBefore(matched, t)
 }
 
+// FilesetAt returns all the fileset files whose timestamps exactly match a given time.
+func FilesetAt(filePathPrefix string, namespace ident.ID, shard uint32, t time.Time) ([]string, error) {
```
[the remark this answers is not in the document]
richardartoul: sure

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: 1) I can add the knob, although I'm not sure what the value in it is...
2) I'm not sure if it really assumes that much. This code path only executes in the scenario where we've streamed all the data for a given shard/block from our peers and we're ready to write that information to disk.

One alternative would be to make the FS bootstrapper do the deletion when it detects a fileset file it can't read which is less leaky (or to archive the files instead of deleting them) but then you're deleting files for which you don't yet have a replacement which seems less good.

Also I'm open to completely alternative solutions, I just couldn't think of any

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: I think this is really explicit, the ok boolean does nothing but tell you whether the files already exist or not. Returning a typed error seems worse to me because then all callers have to know to handle that specific type of error, whereas with the boolean you're force to handle that situation because its part of the type signature

persist/types.go
richardartoul: We could, but I think it leads to people not handling that error case properly. The signature I've defined here forces the caller to handle it or at least acknowledge that they've ignored that situation

richardartoul:
spoke with @prateek in person and we decided to move the actual deletion into the Prepare() method itself and make the caller specify whether or not the existing files should be overwritten

persist/fs/files.go
```diff
@@ -338,22 +358,54 @@ func commitlogFiles(commitLogsDir string, pattern string) ([]string, error) {
...
 func filesBefore(files []string, t time.Time) ([]string, error) {
+	return filterFilesByTime(files, func(ft time.Time) (bool, bool) {
+		if ft.Before(t) {
+			// Include and continue scanning
+			return true, true
+		}
+
+		// Don't include and don't continue scanning because files
+		// are sorted by their timestamps in ascending order so we
+		// know we're done.
```
robskillington: Hm how are we sure that the files are guaranteed to be ordered by time?  It seems they’re just passed in here as an array.  Perhaps instead of passing an array of string we can pass an array of some struct that includes the time and then we can sort it explicitly to be safe/resilient to contracts between functions changing, etc.  Sort on presorted data will be v fast.

persist/fs/files.go
```diff
@@ -338,22 +358,54 @@ func commitlogFiles(commitLogsDir string, pattern string) ([]string, error) {
...
+	})
+}
+
+func filterFilesByTime(files []string, pred filterFilesByTimePred) ([]string, error) {
 	var (
 		j        int
 		multiErr xerrors.MultiError
 	)
-	// Matched files are sorted by their timestamps in ascending order.
+
```
robskillington: Similar to above, can we explicitly sort here to guarantee?

persist/types.go
```diff
@@ -53,7 +53,7 @@ type Flush interface {
...
```
robskillington: Yeah agreed, that’s what caused the issue in the first place, misunderstanding API could return nil for a success / non error case.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: The other alternative is asking an operator to take action, ie fail the bootstrap altogether and crash (which would start crash looping the node).

Maybe that’s ok if there was an easy way for you to decide whether to delete the bad files or not and/or stream them in with peers on next process start, etc.

PR #514, opened 2018-03-25
Add databaseIndex implementation in`storage`

prateek:
- Adds `databaseIndex` and `databaseIndexInsertQueue` construct and implementation. 

NB: In an effort to keep reviews small, i'm using `FOLLOWUP(prateek):...` littered in the code to indicate changes that will follow in subsequent PRs.

/cc @robskillington @richardartoul @jeromefroe @xichen2020


storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+// nolint: deadcode
+func newDatabaseIndex(
+	fn newDatabaseIndexInsertQueueFn,
+	opts index.Options,
+) (databaseIndex, error) {
+	if err := opts.Validate(); err != nil {
+		return nil, err
+	}
+
+	seg, err := mem.New(1, opts.MemSegmentOptions())
```
richardartoul: SUPER nit, but seems like 1 is a unique identifier for this particular index? I would say maybe make it a constant, but if we're moving to namespace-level indices...

Whats the id for?

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	idx := &dbIndex{
+		insertMode: dbIndexInsertAsync,
```
richardartoul: might be nice to make this wireable from the YAML config

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+}
+
+func (i *dbIndex) writeBatch(inserts []dbIndexInsert) error {
+	// NB(prateek): we use a read lock to guard against mutation of the
+	// dbIndexBlock currently active, mutations within the underlying
+	// dbIndexBlock are guarded by primitives internal to it.
+	i.RLock()
+	defer i.RUnlock()
+
+	if !i.isOpenWithRLock() {
```
richardartoul: Is skipping the OnFinalize calls really necessary? Seems like a potential source of shutdown-bugs depending on what people are putting in the OnIndexFinalize method

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+	opts        index.Options
 }
 
-func newDatabaseIndex(o Options) (databaseIndex, error) {
-	return &dbIndex{
-		opts: o,
-	}, nil
+// nolint: deadcode
+func newDatabaseIndex(
+	fn newDatabaseIndexInsertQueueFn,
```
richardartoul: A slightly more verbose name here would be nice, a spent a little time trying to track down what fn was below. Couldn't figure it out until I opened your branch in my editor

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+
+func (i *dbIndex) writeBatch(inserts []dbIndexInsert) error {
+	// NB(prateek): we use a read lock to guard against mutation of the
+	// dbIndexBlock currently active, mutations within the underlying
+	// dbIndexBlock are guarded by primitives internal to it.
+	i.RLock()
+	defer i.RUnlock()
+
+	if !i.isOpenWithRLock() {
+		// NB(prateek): skip calling any of the `OnIndexFinalize` methods
```
richardartoul: Did this make a noticeable difference in performance at any point? I think I understand what you're trying to do, but I spent enough time trying to decipher this comment that I wonder if its even worth doing this.

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+	}
+
+	if numErr != 0 {
+		i.metrics.AsyncInsertErrors.Inc(numErr)
+	}
+
+	return nil
 }
 
 func (i *dbIndex) Write(
```
richardartoul: The "normal path" will not use this method right? It'll go shard insert queue --> indexQueue --> writeBatch right?

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+	}
 
-func (n dbIndexNoOp) Write(context.Context, ident.ID, ident.ID, ident.TagIterator) error {
-	return nil
+	i.state = dbIndexStateClosed
+	return i.insertQueue.Stop()
+}
+
+func (i *dbIndex) doc(ns, id ident.ID, tags ident.Tags) (doc.Document, error) {
+	fields := make([]doc.Field, 0, 1+len(tags))
```
richardartoul: Any opportunity for pooling this, or does the index own it forever?

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+	fields = append(fields, doc.Field{
+		Name:      index.ReservedFieldNameNamespace,
```
richardartoul: Even if this namespace is really long, the index will end up compressing the re-used namespace pretty well right?

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+		Fields: fields,
+	}, nil
 }
 
-func (n dbIndexNoOp) Query(context.Context, index.Query, index.QueryOptions) (index.QueryResults, error) {
-	return index.QueryResults{}, nil
+// NB(prateek): we take an independent copy of the bytes underlying
+// any ids provided, as we need to maintain the lifecycle of the indexed
+// bytes separately from the rest of the storage subsystem.
+func (i *dbIndex) clone(id ident.ID) []byte {
```
richardartoul: Do you have to do this manual clone to prevent sharing pooled byte slices with the index?

storage/index_insert_queue.go
```diff
@@ -0,0 +1,258 @@
...
+	sleepFn      func(time.Duration)
+	notifyInsert chan struct{}
+
+	metrics dbIndexInsertQueueMetrics
+}
+
+type newDatabaseIndexInsertQueueFn func(
+	dbIndexInsertBatchFn, clock.NowFn, tally.Scope) databaseIndexInsertQueue
+
+// FOLLOWUP(prateek): subsequent PR to wire up rate limiting to runtime.Options
```
richardartoul: Lets sync on where you want to put this. I still need to add runtime configuration for LRU size, and I think we want to standardize KV watcher locations

storage/index_insert_queue.go
```diff
@@ -0,0 +1,258 @@
...
+	}
+
+	q.state = dbIndexInsertQueueStateOpen
+	go q.insertLoop()
+	return nil
+}
+
+func (q *dbIndexInsertQueue) Stop() error {
+	q.Lock()
+	defer q.Unlock()
```
richardartoul: Do you want to hold the lock for the whole duration here? Seems like you'd want to:

1. Grab the lock
2. Mark the insert queue as closed
3. Release the lock

That way callers that are trying to call Write() can receive an immediate error instead of waiting for a while on the lock only to be told the insert queue is closed

storage/database.go
```diff
@@ -138,21 +138,15 @@ func NewDatabase(
...
 	logger := iopts.Logger()
 
-	index := databaseIndexNoOp
-	if opts.IndexingEnabled() {
-		index, err = newDatabaseIndex(opts)
-		if err != nil {
-			return nil, err
-		}
-	}
+	var idx databaseIndex
```
robskillington: Hm this is just nil yeah?  Is it worth explicitly setting just the default nil value on the `db` struct?

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+	seg, err := mem.New(1, opts.MemSegmentOptions())
+	if err != nil {
+		return nil, err
+	}
+
+	idx := &dbIndex{
+		insertMode: dbIndexInsertAsync,
+		active: dbIndexBlock{
+			segment:    seg,
+			expiryTime: maxTime, // FOLLOWUP(prateek): undo hard-coding to infinite retention
```
robskillington: Now that we are going to make this per namespace, can we just use the namespace retention options here?

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+			numErr++
+		} else {
+			insert.fns.OnIndexSuccess(i.active.expiryTime)
+		}
+		// NB: we need to release held resources so we un-conditionally execute the Finalize.
+		insert.fns.OnIndexFinalize()
+	}
+
+	if numErr != 0 {
+		i.metrics.AsyncInsertErrors.Inc(numErr)
```
robskillington: Might it be worth adding a `OnIndexError(...)` callback too?

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+	if err != nil {
+		fns.OnIndexFinalize()
+		return err
+	}
+
+	i.RLock()
+	if !i.isOpenWithRLock() {
+		i.RUnlock()
+		i.metrics.InsertAfterClose.Inc(1)
+		fns.OnIndexFinalize()
```
robskillington: Hm why do we finalize on error here but not in write batch method?  Also do we need a write batch and and non batched one?  Does this ever get used?

storage/index.go
```diff
@@ -21,61 +21,260 @@
...
+
+func (i *dbIndex) doc(ns, id ident.ID, tags ident.Tags) (doc.Document, error) {
+	fields := make([]doc.Field, 0, 1+len(tags))
+	fields = append(fields, doc.Field{
+		Name:      index.ReservedFieldNameNamespace,
+		Value:     i.clone(ns),
+		ValueType: doc.StringValueType,
+	})
+	for j := 0; j < len(tags); j++ {
+		t := tags[j]
```
robskillington: nit: Any reason not to `for j, t := range tags { ... }`?

storage/index/iterator.go
```diff
@@ -0,0 +1,126 @@
...
+
+func (i *idsIter) wrapBytes(bytes []byte) checked.Bytes {
+	return i.wrapperPool.Get(bytes)
+}
+
+func (i *idsIter) parseAndStore(d doc.Document) {
+	i.currentID = i.idPool.BinaryID(i.wrapBytes(d.ID))
+	nsFound := false
+	for _, f := range d.Fields {
+		if !nsFound && bytes.Equal(f.Name, ReservedFieldNameNamespace) {
```
robskillington: Yeah if the fields were separated from normal fields then we wouldn't have to do this search, or at the very least be search a smaller bucket of tags that are definitely the reserved tags.

storage/index/iterator.go
```diff
@@ -0,0 +1,126 @@
...
+			nsFound = true
+			continue
+		}
+		i.currentTags = append(i.currentTags,
+			i.idPool.BinaryTag(i.wrapBytes(f.Name), i.wrapBytes(f.Value)))
+	}
+
+	// should have found ns
+	if !nsFound {
+		i.err = errInvalidResultMissingNamespace
```
robskillington: Shouldn't this return `false` in `Next()` if this happens?  Seems like `i.err` is not checked after calling `i.parseAndStore(....)`, perhaps we do that and add a test to catch this?

storage/index_insert_queue.go
robskillington: There's no blocking in the code below, there's is a best effort to kick off the final flush, so whether you release it slightly earlier there will be no real difference in how fast this method executes.

storage/index_insert_queue.go
```diff
@@ -0,0 +1,258 @@
...
+	dbIndexInsertQueueStateOpen
+	dbIndexInsertQueueStateClosed
+)
+
+var (
+	defaultIndexBatchBackoff   = time.Second
+	defaultIndexPerSecondLimit = 10000
+)
+
+type dbIndexInsertQueue struct {
```
robskillington: Could you possibly use Genny to condense needing a copy?  After looking at a lot of stuff, actually doesn't seem like much is different between the two at all.

storage/types.go
```diff
@@ -411,6 +413,40 @@ type databaseIndex interface {
...
+	// inserts to the index asynchronously. It executes the provided callbacks
+	// based on the result of the execution. The returned wait group can be used
+	// if the insert is required to be synchronous.
+	Insert(d doc.Document, s indexInsertLifecycleHooks) (*sync.WaitGroup, error)
+}
+
+// indexInsertLifecycleHooks provides a set of callback hooks to allow the
+// databaseIndex to do lifecycle management of any resources retained
+// during indexing.
+type indexInsertLifecycleHooks interface {
```
robskillington: Yeah we can we rename this something like `OnIndexSeries`?  This would be similar to `OnRetrieveBlock(...)` and `OnReadBlock` type, etc:
https://github.com/m3db/m3db/blob/352f2364f7202e591c69017b33cf78e6b0885f8d/storage/block/types.go#L219-L230

storage/index.go
prateek: It's the unique identifier, I'll rework to use created at nanos.

storage/index.go
prateek: sure thing, renamed to `newIndexQueueFn`

storage/index.go
prateek: Nah it will. I'll draw out the call chains for the different paths here:

- For new entry (previously unseen in the shard):
```
shard.WriteTagged()
  => shardInsertQueue.Insert()
  => shard.writeBatch()
  => index.Write()
  => indexQueue.Insert()
  => index.writeBatch() 
```

- For entry which exists in the shard, but needs indexing (either past the TTL or the last indexing hasn't happened/failed):
```
shard.WriteTagged()
  => index.Write()
  => indexQueue.Insert()
  => index.writeBatch() 
```

storage/index.go
prateek: The latter, it's a completely different lifecycle. Going to rely on the GC to clear as far as we can. Will change if needed down the road.

storage/index_insert_queue.go
prateek: Sure i'll bring it up in person tomorrow.

storage/index_insert_queue.go
prateek: The latency spike during the close path is such an edge case, don't need think about it while there's lots of core functionality to be developed.

storage/index.go
prateek: Kinda -t he storage code finalises IDs when it expires stuff, so we can't rely on the ident passed to the index to be valid for the entirety of the lifecycle of the index. The clone allows us to bypass that restriction. 

Ideally, we'd maintain only a single reference for the ident and do the ref counting correctly to only release once both the storage and index systems were done with it, but that's a much more pervasive change than I wanted to make to the storage side. Did this as a comprise.

storage/index.go
prateek: Sure, i'll make it an option for now. Will do the YAML config when making structs for it in a subsequent PR.

storage/index.go
prateek: done

storage/index.go
prateek: done

storage/index.go
prateek: The original intent was to point out to readers of the code that skipping `OnIndexSuccess` was deliberate. Don't want to loose that. If you think the comment doesn't help, I can rephrase it to be clearer.

prateek: done

storage/index.go
prateek: Haven't done any benchmarking for it, but you can imagine worst case is it's in the order of 10s to 100s of milliseconds. I can rephrase the comment to be clearer. What about it was confusing?

storage/index.go
prateek: Yeah either implementation of segment will keep a bitset of documents per unique (fieldname, fieldvalue)

storage/database.go
prateek: Sure, moved to the namespace and made the change.

storage/index.go
prateek: sure, will take as a ctor parameter and pass down.

storage/index.go
prateek: I didn't do it because it doesn't enable anything special in the shard entry once you're notified there's an indexing error. We emit a metric for it in the index-insert-queue/index already. And we'd retry the indexing on the next write for the shard entry as is.

Is there something else you'd like to do in the shard entry on `OnIndexError`?

storage/index.go
prateek: At this point the write has not made it to the index queue. We've received the write and realised the index has been closed so we terminate early.

prateek: Re: batch v non-batch. This thread explains the call chains: https://github.com/m3db/m3db/pull/514#discussion_r177185849 

Don't need batch v non-batch because all writes are batched internally by the queue.

storage/index.go
prateek: None, I'll make the change.

storage/index.go
prateek: done

storage/index.go
prateek: done

storage/index/iterator.go
prateek: For sure.

storage/index/iterator.go
prateek: Good catch. Addressed and added a test.

storage/index_insert_queue.go
prateek: Yeah they definitely share over 90% of code right now. I'm hesitant to make the change just now - I'd like to see how things are after we've got a fully functioning index. I'll open an issue to track for now, can come back and cleanup after.

prateek: https://github.com/m3db/m3db/issues/517

storage/types.go
prateek: Sure thing.

storage/index.go
richardartoul: ok makes sense.

This was super helpful, can we stick that call chain somewhere

storage/index.go
richardartoul: Not sure I'm totally following. I guess my question was if you insert a million document, we won't have a million different ident.ID's referring to the same namespace right?

storage/index.go
richardartoul: Fair enough, I'll have to look back at your code to see what you are cloning, but I think the series.ID is never finalized and would probably be safe to share with the index (several parts of the codebase depend on series.ID not being manually finalized)

storage/index_insert_queue.go
richardartoul: fair enough

storage/index.go
```diff
@@ -21,61 +21,255 @@
...
+	return i.insertQueue.Stop()
 }
 
-func (n dbIndexNoOp) Query(context.Context, index.Query, index.QueryOptions) (index.QueryResults, error) {
-	return index.QueryResults{}, nil
+func (i *nsIndex) doc(id ident.ID, tags ident.Tags) (doc.Document, error) {
+	fields := make([]doc.Field, 0, 1+len(tags))
+	fields = append(fields, doc.Field{
+		Name:      index.ReservedFieldNameID,
+		Value:     i.clone(id),
```
richardartoul: This is the series ID? I think we can possibly avoid cloning this because it never gets finalized (WiredList and CommitLog and possibly some other things also depend on this behavior)

richardartoul: @robskillington

storage/index_insert_queue.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/m3db/m3db/clock"
+	"github.com/m3db/m3ninx/doc"
+
+	"github.com/uber-go/tally"
+)
+
+var (
+	errIndexInsertQueueNotOpen             = errors.New("index insert queue is not open")
+	errIndexInsertQueueAlreadyOpenOrClosed = errors.New("index insert queue already open or is closed")
+	errNewSeriesIndexRateLimitExceeded     = errors.New("indexing new series eclipses rate limit")
```
richardartoul: extremely nit: eclipses seems like an expensive word where exceeds would suffice

storage/index_insert_queue.go
```diff
@@ -0,0 +1,258 @@
...
+
+var (
+	defaultIndexBatchBackoff   = time.Second
+	defaultIndexPerSecondLimit = 10000
+)
+
+type nsIndexInsertQueue struct {
+	sync.RWMutex
+	state nsIndexInsertQueueState
+	// rate limits
```
richardartoul: super nit, if these have distinction "sections" a space would help along with the comment

storage/index.go
prateek: done

storage/index.go
prateek: I think you're right but I'd still prefer to explicitly copy. It makes the lifecycle a lot easier to reason about, and avoids issues if we ever do change to start finalizing ids.

storage/index_insert_queue.go
prateek: talked offline. we should do this when one of us has cycles.

storage/index.go
prateek: same as above:  I think you're right but I'd still prefer to explicitly copy. It makes the lifecycle a lot easier to reason about, and avoids issues if we ever do change to start finalizing ids.

richardartoul: I guess if the index block size is different than m3db's blocksize that makes sense

storage/database.go
```diff
@@ -104,16 +107,23 @@ type databaseMetrics struct {
...
+	indexDisabledScope := scope.SubScope("index-disabled")
 	return databaseMetrics{
 		unknownNamespaceRead:                unknownNamespaceScope.Counter("read"),
 		unknownNamespaceWrite:               unknownNamespaceScope.Counter("write"),
 		unknownNamespaceWriteTagged:         unknownNamespaceScope.Counter("write-tagged"),
 		unknownNamespaceFetchBlocks:         unknownNamespaceScope.Counter("fetch-blocks"),
 		unknownNamespaceFetchBlocksMetadata: unknownNamespaceScope.Counter("fetch-blocks-metadata"),
+		unknownNamespaceQueryIDs:            unknownNamespaceScope.Counter("query-ids"),
+		queryIDsIndexDisabled:               indexDisabledScope.Counter("query-ids"),
+		writeTaggedIndexDisabled:            indexDisabledScope.Counter("write-tagged"),
```
robskillington: nit: rename these to be explicit about them being error cases?  `errQueryIDsIndexDisabled`, etc?  And for the names perhaps "err-query-ids" and "err-write-tagged"?

storage/index_insert_queue.go
prateek: lol it's a copy of the existing code from the shard insert queue. I'll change both.

storage/index_insert_queue.go
prateek: sure thing

storage/index.go
robskillington: Yeah copying should be ok, it's only on insert thankfully.

storage/index.go
```diff
@@ -21,61 +21,272 @@
...
+}
+
+// NB(prateek): including the call chains leading to this point:
+//
+// - For new entry (previously unseen in the shard):
+//     shard.WriteTagged()
+//       => shardInsertQueue.Insert()
+//       => shard.writeBatch()
+//       => index.Write()
+//       => indexQueue.Insert()
```
robskillington: Is there a reason we don't just `index.writeBatch` from `shard.writeBatch`?  Seems like a second level of queuing that may not be necessary?

storage/index.go
robskillington: Hm probably not at this time, just seemed like having a symmetrical API would've made it clear we don't do anything when we encounter an error on insert.

storage/index/options.go
```diff
@@ -0,0 +1,136 @@
...
+	wrapperPool    xpool.CheckedBytesWrapperPool
+}
+
+// NewOptions returns a new index.Options object with default properties.
+func NewOptions() Options {
+	idPool := ident.NewPool(nil, nil)
+	wrapperPool := xpool.NewCheckedBytesWrapperPool(defaultCheckedBytesPoolInitialSize)
+	wrapperPool.Init()
+	return &opts{
+		insertMode:     InsertAsync,
```
robskillington: Hm I believe that for defaults for series inserts it is synchronous, we should match the defaults of that with this - if writes are sync by default, so too should be indexing yeah?

robskillington: Also can you make this a default const at the top of this file so it's more clear. i.e. `defaultInsertMode = XYZ`

storage/index_insert_queue.go
```diff
@@ -0,0 +1,258 @@
...
+	nsIndexInsertQueueStateOpen
+	nsIndexInsertQueueStateClosed
+)
+
+var (
+	defaultIndexBatchBackoff   = time.Second
+	defaultIndexPerSecondLimit = 10000
+)
+
+type nsIndexInsertQueue struct {
```
robskillington: What was the outcome on determining if `genny` could be used here or not?

storage/namespace.go
```diff
@@ -495,6 +498,17 @@ func (n *dbNamespace) WriteTagged(
 	return err
 }
 
+func (n *dbNamespace) QueryIDs(
+	ctx context.Context,
+	query index.Query,
+	opts index.QueryOptions,
+) (index.QueryResults, error) {
+	callStart := n.nowFn()
+	res, err := n.reverseIndex.Query(ctx, query, opts)
```
robskillington: Hm this will panic at this point as `reverseIndex` is nil, should we instead for now `return fmt.Errorf("not implemented")` or something?

storage/index.go
prateek: talked offline, going to keep the two level indexing going to allow for batch'd writes to the index.

storage/index/options.go
prateek: sure, will do

storage/index_insert_queue.go
prateek: Going to do it once the APIs stabilise a bit more. Have  https://github.com/m3db/m3db/issues/517 to track.

storage/namespace.go
prateek: This can't happen right now because the only caller to this method is `db.QueryIDs` which does the equivalent check before calling `ns.QueryIDs`.

prateek: Added an error check and log statement to be defensive.

storage/index/iterator.go
```diff
@@ -0,0 +1,131 @@
...
+
+func (i *idsIter) wrapBytes(bytes []byte) checked.Bytes {
+	return i.wrapperPool.Get(bytes)
+}
+
+func (i *idsIter) parseAndStore(d doc.Document) error {
+	idFound := false
+	for _, f := range d.Fields {
+		if !idFound && bytes.Equal(f.Name, ReservedFieldNameID) {
+			i.currentID = i.idPool.BinaryID(i.wrapBytes(f.Value))
```
robskillington: aside: This whole wrap bytes thing makes me think maybe we should've been using `checked.Bytes` in the m3ninx code, but that's neither here nor there now.

storage/index/types.go
```diff
@@ -23,8 +23,27 @@ package index
...
 	"github.com/m3db/m3ninx/index/segment"
+	"github.com/m3db/m3ninx/index/segment/mem"
 	"github.com/m3db/m3x/ident"
+	"github.com/m3db/m3x/instrument"
+)
+
+var (
+	// ReservedFieldNameID is the field name used to index the ID in the
+	// m3ninx subsytem.
+	ReservedFieldNameID = []byte("_m3db-id")
```
robskillington: nit: Can we use the same special character here, i.e. `_m3db_id`?  Just when debugging things typing "_m3db_id" over and over again is much easier than "_m3db-id" (I've noticed this when debugging strange JSON structures, thank you human brain...).

prateek: haha sure thing

storage/index/iterator.go
prateek: Not so sure about that. If we're going to have external consumers of `m3ninx`, the `[]byte` interface is a lot more familiar to Go users than `checked.Bytes` would be. 

Can chat more offline.

PR #515, opened 2018-03-27
Codecov enhancements

prateek:
- Use `codecov.io` for coverage reports
- Separate big unit tests into their own make target 
- Add coverage reporting for integration tests under separate codecov flags
- Exclude `integration/`, `generated/`, `tools` and a couple of other files from coverage reports

/cc @richardartoul @robskillington

### DIFF
--- a/.excludecoverage
+++ b/.excludecoverage
@@ -1,0 +1,6 @@
+_mock.go
+_matcher.go
+generated/
+tools/
+vendor/
+integration/
--- a/Makefile
+++ b/Makefile
@@ -166,11 +166,17 @@ test-single-integration:
 	TEST_NATIVE_POOLING=false make test-base-single-integration name=$(name)
 
 .PHONY: test-ci-unit
-test-ci-unit: test-base-ci-unit
+test-ci-unit: test-base
+	$(codecov_push) -f $(coverfile) -F unittests
+
+.PHONY: test-ci-big-unit
+test-ci-big-unit: test-big-base
+	$(codecov_push) -f $(coverfile) -F unittests
 
 .PHONY: test-ci-integration
 test-ci-integration:
 	INTEGRATION_TIMEOUT=4m TEST_NATIVE_POOLING=false TEST_SERIES_CACHE_POLICY=$(cache_policy) make test-base-ci-integration
+	$(codecov_push) -f $(coverfile) -F integration
 
 .PHONY: clean
 clean:
--- a/integration/commitlog_bootstrap_merge_test.go
+++ b/integration/commitlog_bootstrap_merge_test.go
@@ -56,6 +56,10 @@ import (
 // - ns1 block at t0 and t1
 // - commit log blocks from [t1, t3]
 func TestCommitLogAndFSMergeBootstrap(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
+	}
+
 	// Test setup
 	var (
 		rOpts              = retention.NewOptions().SetRetentionPeriod(12 * time.Hour)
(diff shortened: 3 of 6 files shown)


integration/commitlog_bootstrap_merge_test.go
```diff
@@ -56,6 +56,10 @@ import (
 // - ns1 block at t0 and t1
 // - commit log blocks from [t1, t3]
 func TestCommitLogAndFSMergeBootstrap(t *testing.T) {
+	if testing.Short() {
+		t.SkipNow()
```
robskillington: Hm I wish there was a way to enforce this.

robskillington: Either way, can open that up as a separate task.

Makefile
```diff
@@ -166,11 +167,20 @@ test-single-integration:
 	TEST_NATIVE_POOLING=false make test-base-single-integration name=$(name)
 
 .PHONY: test-ci-unit
-test-ci-unit: test-base-ci-unit
+test-ci-unit: test-base
+	cat $(coverfile) | egrep -v -f $(coverage_exclude) > tmp_$(coverfile) && mv tmp_$(coverfile) $(coverfile)
+	bash <(curl -s https://codecov.io/bash) -f $(coverfile) -F unittests
```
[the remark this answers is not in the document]
richardartoul: doesn't seem much worse than `go get <random_go_tool` and then running that

.excludecoverage
```diff
@@ -0,0 +1,5 @@
+_mock.go
```
richardartoul: can we add _matcher.go to this so we can start excluding all our custom matchers from coverage

richardartoul:
LGTM as long as you and @r decide what you want to do with that bash file

Makefile
```diff
@@ -167,15 +168,18 @@ test-single-integration:
 
 .PHONY: test-ci-unit
 test-ci-unit: test-base
+	cat $(coverfile) | egrep -v -f $(coverage_exclude) | sponge $(coverfile)
```
xichen2020: Does this work actually? I think `egrep` matches the full expression whereas in the cover file the lines look like `xxx_mock.go:94,37,97.2 2 0`.

integration/commitlog_bootstrap_merge_test.go
prateek: lol tried migrating to that but aggregator uses go1.8 for ci tests still :| there's a TODO in the ci-scripts module to address.

.excludecoverage
prateek: sure

Makefile
prateek: Yep:
```
❯ cat test.out
try/something.go:13432
random_notmatch.go:1234223423
this-should-work:23423

❯ cat filters.out
try/

❯ cat test.out | egrep -v -f filters.out
random_notmatch.go:1234223423
this-should-work:23423
```

Makefile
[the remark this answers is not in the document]
prateek: Yeah I was in two minds about doing this. It's a terrible practice in general - I refuse to run scripts like this on my laptop. It's much worse than go get (e.g. [hijack](https://www.idontplaydarts.com/2016/04/detecting-curl-pipe-bash-server-side/)). I put it in cause it's only going to run on the ci servers but happy to change to avoid doing in general.

Makefile
xichen2020: Ah cool.

Makefile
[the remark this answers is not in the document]
prateek: put the bash script in `m3db/ci-scripts`

.excludecoverage
prateek: done

PR #518, opened 2018-04-02
Change peers bootstrapper to use consistency level and indefinitely retry when not achieved

robskillington:


prateek:
@robskillington looks like there's some compilation issues:

<details>

```
integration/client.go:211:25:warning: cannot use result.NewOptions() (value of type github.com/m3db/m3db/storage/bootstrap/result.Options) as github.com/m3db/m3db/client.ReadConsistencyLevel value in argument to session.FetchBlocksMetadataFromPeers (unconvert)integration/client.go:211:46:warning: cannot use version (variable of type github.com/m3db/m3db/client.FetchBlocksMetadataEndpointVersion) as github.com/m3db/m3db/storage/bootstrap/result.Options value in argument to session.FetchBlocksMetadataFromPeers: missing method ClockOptions (unconvert)
integration/client.go:211:53:warning: too few arguments in call to session.FetchBlocksMetadataFromPeers (unconvert)
integration/client.go:211:25:warning: unused variable or constant cannot use result.NewOptions() (value of type github.com/m3db/m3db/storage/bootstrap/result.Options) as github.com/m3db/m3db/client.ReadConsistencyLevel value in argument to session.FetchBlocksMetadataFromPeers (varcheck)
integration/client.go:211:46:warning: unused variable or constant cannot use version (variable of type github.com/m3db/m3db/client.FetchBlocksMetadataEndpointVersion) as github.com/m3db/m3db/storage/bootstrap/result.Options value in argument to session.FetchBlocksMetadataFromPeers: missing method ClockOptions (varcheck)
integration/client.go:211:53:warning: unused variable or constant too few arguments in call to session.FetchBlocksMetadataFromPeers (varcheck)
integration/client.go:211:25:warning: cannot use result.NewOptions() (value of type github.com/m3db/m3db/storage/bootstrap/result.Options) as github.com/m3db/m3db/client.ReadConsistencyLevel value in argument to session.FetchBlocksMetadataFromPeers (maligned)
integration/client.go:211:46:warning: cannot use version (variable of type github.com/m3db/m3db/client.FetchBlocksMetadataEndpointVersion) as github.com/m3db/m3db/storage/bootstrap/result.Options value in argument to session.FetchBlocksMetadataFromPeers: missing method ClockOptions (maligned)
integration/client.go:211:53:warning: too few arguments in call to session.FetchBlocksMetadataFromPeers (maligned)
integration/client.go:211:25:warning: unused struct field cannot use result.NewOptions() (value of type github.com/m3db/m3db/storage/bootstrap/result.Options) as github.com/m3db/m3db/client.ReadConsistencyLevel value in argument to session.FetchBlocksMetadataFromPeers (structcheck)
integration/client.go:211:46:warning: unused struct field cannot use version (variable of type github.com/m3db/m3db/client.FetchBlocksMetadataEndpointVersion) as github.com/m3db/m3db/storage/bootstrap/result.Options value in argument to session.FetchBlocksMetadataFromPeers: missing method ClockOptions (structcheck)
integration/client.go:211:53:warning: unused struct field too few arguments in call to session.FetchBlocksMetadataFromPeers (structcheck)
integration/client.go:211:25:warning: cannot use result.NewOptions() (value of type github.com/m3db/m3db/storage/bootstrap/result.Options) as github.com/m3db/m3db/client.ReadConsistencyLevel value in argument to session.FetchBlocksMetadataFromPeers (unused)
integration/client.go:211:25:warning: cannot use result.NewOptions() (value of type github.com/m3db/m3db/storage/bootstrap/result.Options) as github.com/m3db/m3db/client.ReadConsistencyLevel value in argument to session.FetchBlocksMetadataFromPeers (badtime)
integration/client.go:211:46:warning: cannot use version (variable of type github.com/m3db/m3db/client.FetchBlocksMetadataEndpointVersion) as github.com/m3db/m3db/storage/bootstrap/result.Options value in argument to session.FetchBlocksMetadataFromPeers: missing method ClockOptions (badtime)
integration/client.go:211:53:warning: too few arguments in call to session.FetchBlocksMetadataFromPeers (badtime)
integration/client.go:211:25:warning: cannot use result.NewOptions() (value of type github.com/m3db/m3db/storage/bootstrap/result.Options) as github.com/m3db/m3db/client.ReadConsistencyLevel value in argument to session.FetchBlocksMetadataFromPeers (gosimple)
# github.com/m3db/m3db/integration
/tmp/go-build193866661/github.com/m3db/m3db/integration/_test/_obj_test/client.go:247:52: not enough arguments in call to session.FetchBlocksMetadataFromPeers
	have (ident.ID, uint32, "time".Time, "time".Time, result.Options, "github.com/m3db/m3db/client".FetchBlocksMetadataEndpointVersion)
	want (ident.ID, uint32, "time".Time, "time".Time, "github.com/m3db/m3db/client".ReadConsistencyLevel, result.Options, "github.com/m3db/m3db/client".FetchBlocksMetadataEndpointVersion)
make[1]: *** [test-base-ci-integration] Error 2
```

</details>

storage/repair.go
```diff
@@ -119,8 +119,9 @@ func (r shardRepairer) Repair(
 	metadata.AddLocalMetadata(origin, localIter)
 
 	// Add peer metadata
-	peerIter, err := session.FetchBlocksMetadataFromPeers(namespace,
-		shard.ID(), start, end, result.NewOptions(), client.FetchBlocksMetadataEndpointV2)
+	level := client.ReadConsistencyLevelMajority
```
prateek: should this be moved to `storage/repair/Options`?

client/session.go
```diff
@@ -2344,23 +2479,8 @@ func (s *session) selectBlocksForSeriesFromPeerBlocksMetadata(
...
-				s.log.WithFields(
-					xlog.NewField("id", currID.String()),
-					xlog.NewField("start", earliestStart),
-					xlog.NewField("attempted", currUnselected.reattempt.attempt),
-					xlog.NewField("attemptErrs", xerrors.Errors(currUnselected.reattempt.errs).Error()),
-				).Error("retries failed for streaming blocks from peers")
-			}
-
+			// No current eligible peers to select from, retry all of them again
+			currEligible = currStart[:]
```
prateek: could you emit a metric/log indicating this is happening, will be useful if we ever see bootstrapping stuck for a while.

client/session.go
```diff
@@ -1420,27 +1437,45 @@ func (s *session) Truncate(namespace ident.ID) (int64, error) {
...
+		result.peers = append(result.peers, newPeer(s, host))
 	})
 	s.RUnlock()
-	if err != nil {
-		return nil, err
+	if resultErr := xerrors.FirstError(err, lookupErr); resultErr != nil {
+		return peers{}, resultErr
 	}
-	return peers, nil
+	return result, nil
```
prateek: Any concern about sharing this object outside the RLock? 

Main concern is if it can cause panics in weird situations: e.g. if we're bootstrapping during a topology change (for w/e reason).

client/session.go
```diff
@@ -1625,74 +1661,163 @@ func (s *session) FetchBlocksFromPeers(
...
 	version FetchBlocksMetadataEndpointVersion,
 ) error {
 	var (
 		wg       sync.WaitGroup
-		errLock  sync.Mutex
-		errLen   int
-		pending  int64
-		multiErr = xerrors.NewMultiError()
+		errLock  sync.RWMutex
+		errors   = make(map[int]error)
```
prateek: nit: could you break the `errors` and `abortError` vars and associated methods into a helper struct, would help readability a bunch.

storage/repair.go
richardartoul: Yeah options seems like the right place, although I wonder if we would even want to allow a non-quorum repair...

client/session.go
```diff
@@ -1293,61 +1300,71 @@ func (s *session) fetchIDsAttempt(
...
-	switch s.writeLevel {
+	if !s.writeConsistencyAchieved(level, int(majority), int(enqueued), int(success)) {
+		return newConsistencyResultError(level, int(enqueued), int(responded), errs)
+	}
+	return nil
+}
+
+func (s *session) writeConsistencyAchieved(level topology.ConsistencyLevel, majority, enqueued, success int) bool {
+	switch level {
 	case topology.ConsistencyLevelAll:
```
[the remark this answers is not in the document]
richardartoul: also a default panic would be nice

client/session.go
richardartoul: @prateek how would this cause a panic?

client/session.go
```diff
@@ -1625,74 +1661,163 @@ func (s *session) FetchBlocksFromPeers(
...
+			errLock.Unlock()
+		}
+		getAbortError = func() error {
+			errLock.RLock()
+			result := abortError
+			errLock.RUnlock()
+			return result
+		}
+		pending   = int64(len(peers.peers))
+		majority  = atomic.LoadInt32(&s.majority)
```
richardartoul: Why do we need an atomic here? In case it changes during the bootstrap?

client/session.go
```diff
@@ -1625,74 +1661,163 @@ func (s *session) FetchBlocksFromPeers(
...
+		enqueued  = int32(len(peers.peers))
```
richardartoul: any reason pending is an int64 and pending is an int32? Would be nice to just use an int since these numbers should never be anywhere near max for an int32

client/session.go
```diff
@@ -1625,74 +1661,163 @@ func (s *session) FetchBlocksFromPeers(
...
+		responded int32
+		success   int32
 	)
+	if peers.selfExcluded {
+		state, err := peers.selfHostShardSet.ShardSet().LookupStateByID(shardID)
+		if err == nil && state == shard.Available {
+			// If we excluded ourselves from fetching, we basically treat ourselves
```
richardartoul: This comment seems a little misleading. We can't really bootstrap from ourselves because being down for any amount of time implies that we are missing data.

Maybe re-phrase to say that as long as we read metadata from quorum nodes (including ourselves) we're guaranteed to not miss any confirmed writes? Although that implies that we read our own commit log.

client/session.go
richardartoul: +1

client/session.go
```diff
@@ -1625,74 +1661,163 @@ func (s *session) FetchBlocksFromPeers(
...
+
+				// Decrement pending
+				progress.metadataFetches.Update(float64(atomic.AddInt64(&pending, -1)))
+
+				// Mark done
+				wg.Done()
+			}()
+
+			var currPageToken pageToken
+			condition := func() bool {
```
richardartoul: Could this be a helper method called `shouldContinueStreamingMetadata` or something? This method is already pretty noisy, would be nice to tuck this away

client/session.go
```diff
@@ -1625,74 +1661,163 @@ func (s *session) FetchBlocksFromPeers(
...
+
+	errs := getErrors()
+	return s.readConsistencyResult(consistencyLevel, majority, enqueued,
+		atomic.LoadInt32(&responded), int32(len(errs)), errs)
 }
 
+// pageToken is just an opaque type that needs to be downcasted to expected
+// page token type, this makes it easy to use the page token across the two
+// versions
+type pageToken interface{}
```
richardartoul: Could you add a TODO to delete this along with the V1 endpoint like the other comments we have

client/session.go
richardartoul: yeah both a log and metric would be useful here just so its easy to debug whats wrong both ways

client/session.go
```diff
@@ -1625,74 +1661,163 @@ func (s *session) FetchBlocksFromPeers(
...
+			var currPageToken pageToken
+			condition := func() bool {
+				level := consistencyLevel
+				majority := int(majority)
+				enqueued := int(enqueued)
+				success := int(atomic.LoadInt32(&success))
+				return !s.readConsistencyAchieved(level, majority, enqueued, success) &&
+					getAbortError() == nil
 			}
+			for condition() {
```
richardartoul: Do we want to include any additional back-off here? We have exponential backoff in the streamBlocksMetadataFromPeer method, but when it errors out we will reset the retrier and start hitting it again.

I guess its technically really cheap for the node to tell us its overloaded, so maybe this is ok?

client/session.go
```diff
@@ -1625,74 +1661,163 @@ func (s *session) FetchBlocksFromPeers(
...
+					setError(idx, err)
+					return
+				}
 
-			if err != nil {
-				errLock.Lock()
-				defer errLock.Unlock()
-				errLen++
-				multiErr = multiErr.Add(err)
+				// Set error or success if err is nil
```
richardartoul: This comment is a little confusing because it makes it seem like setError will increment the atomic success variable. I think it would be clearer if you left the comment out or just did

```golang
if err != nil {
  setError(idx, err)
}
```

then there is no ambiguity even if it behaves exactly the same way

client/session.go
```diff
@@ -1625,74 +1661,163 @@ func (s *session) FetchBlocksFromPeers(
...
+				atomic.AddInt32(&responded, 1)
+
+				// Decrement pending
+				progress.metadataFetches.Update(float64(atomic.AddInt64(&pending, -1)))
+
+				// Mark done
+				wg.Done()
+			}()
+
+			var currPageToken pageToken
```
richardartoul: A comment here explaining that this starts off nil, but will be updated with the last seen page token when errors occur for the sake of resuming would help explain this code.

Actually maybe calling it `lastSeenPageToken` or something would also help

client/session.go
richardartoul: To clarify, if all of our peers error out enough to cause "final errors" this will just restart the retry loop again, right?

Could you add a comment to that effect, this method is pretty confusing if you've never worked with it before.

richardartoul:
This P.R seems like a good chance to improve our documentation, since the stuff I wrote in `peer_stream.md` will be inaccurate once this P.R lands

prateek:
@robskillington I was just deploying the test cluster in a super un-safe way: stopped all nodes, deleted all files, and restarted all nodes. All the nodes quickly got overloaded with the pending writes and started returning error to each other while bootstrapping. 

In the event of some unexpected failures (multiple/all replicas down) the changes in this PR would make recovery that much harder. Could we have mechanism to put a hard-cap on retries in case we need to recover partial state?

client/session.go
robskillington: Yeah this isn't an issue, that's why the `peer` interface only lets you borrow a connection (and if topology occurs and its no longer in the topology then borrowing connection will just fail):
```go
type peer interface {
	// Host gets the host
	Host() topology.Host

	// BorrowConnection will borrow a connection and execute a user function
	BorrowConnection(fn withConnectionFn) error
}
```

client/session.go
robskillington: Sure thing.

client/session.go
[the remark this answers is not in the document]
robskillington: Yeah I can add a panic.. I prefer to steer away from panics at all possible times.

client/session.go
```diff
@@ -1293,61 +1300,71 @@ func (s *session) fetchIDsAttempt(
...
-	switch s.readLevel {
+	if !s.readConsistencyAchieved(level, int(majority), int(enqueued), int(success)) {
+		return newConsistencyResultError(level, int(enqueued), int(responded), errs)
+	}
+	return nil
+}
+
+func (s *session) readConsistencyAchieved(level ReadConsistencyLevel, majority, enqueued, success int) bool {
+	switch level {
 	case ReadConsistencyLevelAll:
```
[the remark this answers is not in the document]
robskillington: Ditto my previous comment: I'll make it return an error on "default" case, if consistency level ever gets passed in by the user per operation then a panic would take their whole service down, which is highly undesirable.

client/session.go
robskillington: It *could* change during topology change (add replica) but very unlikely.

client/types.go
```diff
@@ -283,6 +283,7 @@ type AdminSession interface {
 		namespace ident.ID,
 		shard uint32,
 		start, end time.Time,
+		consistencyLevel ReadConsistencyLevel,
```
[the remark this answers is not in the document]
robskillington: Mainly because I assume that caller might want a different consistency level and this is a pretty generic API call here.

storage/repair.go
robskillington: Sure thing, can add there.

client/session.go
robskillington: Yeah that's the idea, I'll add a comment to that effect and the metric.  I'll also add an ability to cut this circuit if we need to (i.e. for test cluster).

client/session.go
robskillington: It actually needs to nil if it's no longer error for this replica so we don't report an old error when we call `getErrors()`.

client/session.go
robskillington: Sure thing, I'll add a comment.

client/session.go
robskillington: Yeah this should be fine, we'll have already performed pretty reasonable exponential backoff already.

client/session.go
robskillington: Sure thing.

client/session.go
robskillington: It's pretty difficult because it captures so many variables (5 in total), I've removed the errors, etc so it should be much better now.

client/session.go
robskillington: It's been rephrased somewhat, let me know if you don't think it's descriptive enough.

client/session.go
robskillington: These are only int32/etc for atomic purposes.

client/consistency_level.go
```diff
@@ -0,0 +1,83 @@
...
+
+import (
+	"github.com/m3db/m3db/topology"
+)
+
+// queryableReadConsistencyLevel is a queryable value for a
+// read consistency level, this supports it being able to change
+// dynamically or it can be just static if not required to be changed
+// during an operation.
+type queryableReadConsistencyLevel interface {
```
richardartoul: super nit: I wonder if "dynamic" or "runtime" would be more consistent with our existing naming conventions and thus more obvious to developers familiar with our codebase that this is a runtime-changeable value

client/options.go
```diff
@@ -298,6 +305,26 @@ func (o *options) Validate() error {
 	if o.readerIteratorAllocate == nil {
 		return errNoReaderIteratorAllocateSet
 	}
+	if err := topology.ValidateConsistencyLevel(
```
richardartoul: Might be nice to have the topology expose a "ValidateAllConsistencyLevels" method?

client/session.go
```diff
@@ -835,14 +872,14 @@ func (s *session) writeAttempt(
 		return timestampErr
 	}
 
-	if s.RLock(); s.state != stateOpen {
-		s.RUnlock()
-		return errSessionStateNotOpen
+	if s.state.RLock(); s.state.status != statusOpen {
```
richardartoul: This might just be personal preference and I know this was already like this, but my first instinct when I see this line is that you're using some variant of an RLock which returns immediately whether or not it was able to acquire the lock.

Might be a little easier to parse if we just move the RLock() call above the conditional

client/session.go
```diff
@@ -1522,18 +1631,8 @@ func (s *session) FetchBootstrapBlocksFromPeers(
...
-		streamFn = s.streamAndGroupCollectedBlocksMetadata
-	case FetchBlocksMetadataEndpointV2:
-		streamFn = s.streamAndGroupCollectedBlocksMetadataV2
-	// Should never happen
-	default:
-		return nil, errInvalidFetchBlocksMetadataVersion
-	}
-	s.streamBlocksFromPeers(
-		nsMetadata, shard, peers, metadataCh, opts, result, progress, streamFn)
+	s.streamBlocksFromPeers(nsMetadata, shard, peers, metadataCh, opts,
```
richardartoul: are you deleting the old codepath in this P.R as well?

client/session.go
richardartoul: ah ok fair enough. That might be worth a one line comment

client/session.go
richardartoul: nevermind I see that you just pushed the switch statement down

client/session.go
```diff
@@ -1751,22 +1925,27 @@ func (s *session) streamBlocksMetadataFromPeer(
...
-					blockMetas = append(blockMetas, blockMetadata{
-						start: blockStart,
-					})
 					s.log.WithFields(
 						xlog.NewField("shard", shard),
 						xlog.NewField("peer", peerStr),
 						xlog.NewField("block", blockStart),
 						xlog.NewField("error", err),
 					).Error("error occurred retrieving block metadata")
+					// Enqueue with a zerod checksum which cause a fanout fetch
```
richardartoul: super nit: zeroed or zero'd or anything other than zerod

richardartoul: causes*

richardartoul: maybe "which triggers a fanout fetch"?

client/session.go
```diff
@@ -1912,19 +2102,20 @@ func (s *session) streamBlocksMetadataFromPeerV2(
...
-						{start: blockStart},
-					},
-				}
 				s.log.WithFields(
 					xlog.NewField("shard", shard),
 					xlog.NewField("peer", peerStr),
 					xlog.NewField("block", blockStart),
 					xlog.NewField("error", err),
 				).Error("error occurred retrieving block metadata")
+				// Enqueue with a zerod checksum which cause a fanout fetch
```
richardartoul: same comment

client/session.go
```diff
@@ -2197,295 +2345,300 @@ func (s *session) streamAndGroupCollectedBlocksMetadataV2(
...
-			// No current eligible peers to select from
-			finalError := true
-			if currUnselected.reattempt.failAllowed != nil && atomic.AddInt32(currUnselected.reattempt.failAllowed, -1) > 0 {
-				// Some peers may still return results so we don't report error here
-				finalError = false
+		level := consistencyLevel.value()
+		achievedConsistencyLevel := s.readConsistencyAchieved(level,
+			majority, enqueued, success)
+		if achievedConsistencyLevel {
+			if success > 0 {
```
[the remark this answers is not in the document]
robskillington: Correct yes.

client/session.go
```diff
@@ -2197,295 +2345,300 @@ func (s *session) streamAndGroupCollectedBlocksMetadataV2(
...
-				m.fetchBlockFinalError.Inc(1)
-				s.log.WithFields(
-					xlog.NewField("id", currID.String()),
-					xlog.NewField("start", earliestStart),
-					xlog.NewField("attempted", currUnselected.reattempt.attempt),
-					xlog.NewField("attemptErrs", xerrors.Errors(currUnselected.reattempt.errs).Error()),
-				).Error("retries failed for streaming blocks from peers")
-			}
+			// No success, inform operator that although consistency level achieved
+			// there was a no successful fetches. This can happen if consistency
```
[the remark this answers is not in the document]
robskillington: Thanks 👍

client/session.go
robskillington: Sure thing 👍

client/session.go
robskillington: Sure thing 👍

client/session.go
robskillington: 👍

client/session.go
robskillington: Yeah, I'm more of a fan of idiomatic Go and advised I think when this was first authored to do what you're suggesting.  I'll clean this up now.

client/options.go
robskillington: Hm, I think it's a little cleaner like this because we may want to return the specialization of which consistency level is invalid.

client/consistency_level.go
robskillington: Sure thing.

PR #523, opened 2018-04-11
Wire WriteTagged RPC -> storage path

prateek:


integration/index_large_cardinality_high_concurrency_test.go
```diff
@@ -0,0 +1,112 @@
+// +build integration_disabled
```
robskillington: Do we want to have this disabled by default?

services/m3dbnode/main/main_index_test.go
```diff
@@ -0,0 +1,422 @@
...
+					Regexp:           true,
+				},
+			},
+		},
+	}, index.QueryOptions{
+		StartInclusive: fetchStart,
+		EndExclusive:   fetchEnd,
+	})
+
+	// FOLLOWUP(prateek): implement FetchTagged check once the `client/fetch*` APIs are implemented.
```
robskillington: Nice.  Looks good.

integration/index_large_cardinality_high_concurrency_test.go
prateek: Yep it's too heavy to run on the ci atm. Leaving it in as more a benchmarking utility for now.

PR #528, opened 2018-04-13
Update m3x and use code generated maps instead of ident.Hash based maps

robskillington:


client/received_blocks_map.go
```diff
@@ -0,0 +1,303 @@
+// This file was automatically generated by genny.
+// Any changes will be lost if this file is regenerated.
+// see https://github.com/cheekybits/genny
+
+// Copyright (c) 2018 Uber Technologies, Inc.
```
prateek: few nits:
- automate generation of the genny output files (`m3db/generated/genny/generate.go`, `make genny-gen`, etc)
- filter out the generated files from the coverage report (can include filename patterns in `.excludecoverage`)

client/received_blocks_map.go
```diff
@@ -0,0 +1,303 @@
...
+	return bytes.Equal(value.Bytes(), v)
+}
+
+func (v id) Reset() {
+	for i := range v {
+		v[i] = byte(0)
+	}
+}
+
+func (v id) Finalize() {
```
prateek: would it be better to force users to only use the non-copy version of `Set`, and then panic inside this method?

persist/fs/retriever_concurrent_test.go
```diff
@@ -1,4 +1,3 @@
-// +build big
```
prateek: is this intentionally removed?

Makefile
```diff
@@ -20,6 +21,9 @@ protoc_go_package    := github.com/golang/protobuf/protoc-gen-go
 thrift_gen_package   := github.com/uber/tchannel-go
 thrift_output_dir    := generated/thrift/rpc
 thrift_rules_dir     := generated/thrift
+m3x_package          := github.com/m3db/m3x
+m3x_package_path     := $(gopath_prefix)/$(m3x_package)
```
prateek: Would it be better to rely on the vendor'd version in m3db?

Makefile
```diff
@@ -135,10 +139,122 @@ proto-gen: install-proto-bin install-license-bin
...
+map-storage-database-namespaces-gen: install-m3x-repo
+	cd $(m3x_package_path) && make idhashmap-gen \
+		pkg=storage \
+		value_type=databaseNamespace \
+		out_dir=$(m3db_package_path)/storage \
+		rename_type_prefix=databaseNamespaces \
+		rename_constructor=newDatabaseNamespacesMap \
+		rename_constructor_options=databaseNamespacesMapOptions
+	# Rename both generated map and constructor files
+	mv -f $(m3db_package_path)/storage/map_gen.go $(m3db_package_path)/storage/namespace_map_gen.go
```
prateek: more a question: this target `map-storage-database-namespaces-gen` and `map-storage-shard-gen` both generate temporary files with the same name. Does that cause issues when you're running targets in parallel?

storage/shard.go
```diff
@@ -219,6 +219,8 @@ type dbShardEntryWorkFn func(entry *dbShardEntry) bool
 
 type dbShardEntryBatchWorkFn func(entries []*dbShardEntry) bool
 
+type listElement *list.Element
```
prateek: suggestion(take/leave): rename to `shardListElement` to be explicit about how the type is used.

storage/repair.go
```diff
@@ -184,7 +184,7 @@ type repairState struct {
 
 type namespaceRepairStateByTime map[xtime.UnixNano]repairState
 
-type repairStatesByNs map[ident.Hash]namespaceRepairStateByTime
+type repairStatesByNs map[string]namespaceRepairStateByTime
```
prateek: How come you're not using a generated map here?

storage/namespace_map_gen.go
```diff
@@ -0,0 +1,259 @@
...
+	finalize bool
+}
+
+// Key returns the map entry key.
+func (e databaseNamespacesMapEntry) Key() ident.ID {
+	return e.key.key
+}
+
+// databaseNamespace returns the map entry value.
+func (e databaseNamespacesMapEntry) databaseNamespace() databaseNamespace {
```
prateek: how about just calling this `Value()` instead? would make the usage of the generated map a little more standard - always have to use `Key()/Value()` instead of type specific method names to access

storage/bootstrap/result/result.go
```diff
@@ -90,38 +90,41 @@ func MergedBootstrapResult(i, j BootstrapResult) BootstrapResult {
...
 
 // NewShardResult creates a new shard result.
 func NewShardResult(capacity int, opts Options) ShardResult {
 	return &shardResult{
-		opts:   opts,
-		blocks: make(map[ident.Hash]DatabaseSeriesBlocks, capacity),
+		opts: opts,
+		blocks: NewMap(MapOptions{
+			InitialSize: capacity,
+			KeyCopyPool: opts.DatabaseBlockOptions().BytesPool().BytesPool(),
```
prateek: nit: This reads weird, could you rename `opts.DatabaseBlockOptions().BytesPool()` to `opts.DatabaseBlockOptions().CheckedBytesPool()`.

storage/bootstrap_test.go
```diff
@@ -65,6 +66,62 @@ func TestDatabaseBootstrapWithBootstrapError(t *testing.T) {
 	require.Equal(t, bootstrapped, bsm.state)
 }
 
+func TestDatabaseBootstrapSubsequentCallsQueued(t *testing.T) {
```
prateek: +1

client/received_blocks_new_map.go
```diff
@@ -0,0 +1,54 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package client
+
+import (
+	"github.com/cespare/xxhash"
+	"github.com/m3db/m3x/ident"
```
prateek: nit: import order

client/session.go
```diff
@@ -2278,20 +2282,22 @@ func (s *session) streamAndGroupCollectedBlocksMetadata(
 	peersLen int,
 	metadataCh <-chan receivedBlockMetadata,
 	enqueueCh enqueueChannel,
+	pool pool.BytesPool,
 ) {
-	metadata := make(map[hashAndBlockStart]receivedBlocks)
+	metadata := newReceivedBlocksMap(pool)
+	defer metadata.Reset() // Delete all the keys and return slices to pools
```
prateek: lol `the_lengths_our_code_goes_to_fight_the_gc++`. nice change

client/received_blocks_new_map.go
```diff
@@ -0,0 +1,54 @@
...
+)
+
+func newReceivedBlocksMap(pool pool.BytesPool) *receivedBlocksMap {
+	return _receivedBlocksMapAlloc(_receivedBlocksMapOptions{
+		hash: func(k idAndBlockStart) receivedBlocksMapHash {
+			return receivedBlocksMapHash(xxhash.Sum64(k.id.Bytes())) +
+				receivedBlocksMapHash(k.blockStart)
+		},
+		equals: func(x, y idAndBlockStart) bool {
+			return x.id.Equal(y.id) && x.blockStart == y.blockStart
```
prateek: super nit: can change order of equality checks to make this a little bit quicker due to short circuiting

client/received_blocks_new_map.go
```diff
@@ -0,0 +1,54 @@
...
+import (
+	"github.com/cespare/xxhash"
+	"github.com/m3db/m3x/ident"
+	"github.com/m3db/m3x/pool"
+)
+
+func newReceivedBlocksMap(pool pool.BytesPool) *receivedBlocksMap {
+	return _receivedBlocksMapAlloc(_receivedBlocksMapOptions{
+		hash: func(k idAndBlockStart) receivedBlocksMapHash {
+			return receivedBlocksMapHash(xxhash.Sum64(k.id.Bytes())) +
```
prateek: hm might be better for distribution to offset the addition, i.e. 

```
hash: func(k idAndBlockStart) receivedBlocksMapHash {
  const SOME_PRIME_NUMBER = 7
  return receivedBlocksMapHash(k.blockStart * SOME_PRIME_NUMBER ) +
    receivedBlocksMapHash(xxhash.Sum64(k.id.Bytes()))
 }

File: client/received_blocks_new_map.go [hunk 15]
```diff
@@ -0,0 +1,54 @@
...
+			keyLen := len(bytes)
+			pooled := pool.Get(keyLen)[:keyLen]
+			copy(pooled, bytes)
+			return idAndBlockStart{
+				id:         ident.BytesID(pooled),
+				blockStart: k.blockStart,
+			}
+		},
+		finalize: func(k idAndBlockStart) {
+			if slice, ok := k.id.(ident.BytesID); ok {
```
prateek: why is this check needed?

client/host_queue.go
```diff
@@ -183,64 +183,56 @@ func (q *queue) rotateOpsWithLock() []op {
...
 				}
 
-				currTaggedWriteOps = append(currTaggedWriteOps, ops[i])
-				currTaggedBatchElements = append(currTaggedBatchElements, &v.request)
-				currTaggedWriteOpsByNamespace[namespaceKey] = currTaggedWriteOps
-				currTaggedBatchElementsByNamespace[namespaceKey] = currTaggedBatchElements
+				currTaggedWriteOpsByNamespace.appendAt(idx, ops[i], &v.request)
 
-				if len(currTaggedWriteOps) == writeBatchSize {
+				if currTaggedWriteOpsByNamespace.lenAt(idx) == writeBatchSize {
```
prateek: noticed a bug unrelated to your changes: could you use `writeTaggedBatchSize` here

client/host_queue.go
```diff
@@ -584,3 +578,105 @@ func errQueueUnknownOperation(hostID string) error {
 func errQueueFetchNoResponse(hostID string) error {
 	return fmt.Errorf("host operation queue did not receive response for given fetch for host: %s", hostID)
 }
+
+// ops container types
+
+type namespaceWriteBatchOps struct {
```
prateek: opened https://github.com/m3db/m3db/issues/531 to track de-duplication. mind putting a `// TODO: use genny to make namespaceWriteBatchOps and namespaceWriteTaggedBatchOps share code (https://github.com/m3db/m3db/issues/531)` and elsewhere as you find applicable

Makefile
```diff
@@ -135,10 +139,122 @@ proto-gen: install-proto-bin install-license-bin
 	@echo Generating protobuf files
 	PACKAGE=$(m3db_package) $(auto_gen) $(proto_output_dir) $(proto_rules_dir)
 
+.PHONY: install-m3x-repo
+install-m3x-repo: install-glide install-generics-bin
```
prateek: same as above, might be better to use the version of m3x in `m3db/vendor`

Makefile
```diff
@@ -135,10 +139,122 @@ proto-gen: install-proto-bin install-license-bin
...
+	# Check if repository exists, if not get it
+	test -d $(m3x_package_path) && test -d $(m3x_package_path)/vendor || ( \
+		go get -u $(m3x_package)                                          && \
+		cd $(m3x_package_path)                                            && \
+		glide install                                                        \
+	)
+	# If does exist but not at min version then update it
+	(cd $(m3x_package_path) && git cat-file -t $(m3x_package_min_ver) > /dev/null) || go get -u $(m3x_package)
+
+# Map generation rule for client/receivedBlocksMap
```
prateek: suggestion(take/leave): move the new generation make targets to a separate makefile (`gennygen.mk` or something)

Makefile
```diff
@@ -135,10 +139,122 @@ proto-gen: install-proto-bin install-license-bin
...
+.PHONY: map-client-received-blocks-gen
```
prateek: suggestion(take/leave): rename new targets to include a `genny` prefix. will be useful to have `genny` as the toplevel classification for cases when we have non map types codegen'd.

Makefile
```diff
@@ -135,10 +139,122 @@ proto-gen: install-proto-bin install-license-bin
...
+		value_type=ReplicaSeriesBlocksMetadata \
+		out_dir=$(m3db_package_path)/storage/repair
+
+# Map generation rule for all generated maps
+.PHONY: map-all-gen
+map-all-gen: map-client-received-blocks-gen map-storage-block-retriever-gen map-storage-bootstrap-result-gen map-storage-gen map-storage-namespace-metadata-gen map-storage-repair-gen
+
+# Tests that all currently generated maps match their contents if they were regenerated now
+.PHONY: test-map-all-gen
+test-map-all-gen: map-all-gen
```
prateek: should we do this for all codegen? can open an issue to track if so

storage/bootstrap/result/result.go
robskillington: That would change every pool option in the project unfortunately.

client/received_blocks_new_map.go
robskillington: Sure thing.

client/received_blocks_new_map.go
robskillington: Sure thing.

client/received_blocks_new_map.go
robskillington: Fair, I'll use the standard Java composite key approach.

client/received_blocks_new_map.go
robskillington: Someone might use `SetUnsafe(...)` with NoCopyKey so they have a user defined type as the ID passed in here.  This is safe under that case, also type assertion to a concrete type `BytesID` which is a type alias of `[]byte` is super fast:
https://github.com/teh-cmc/go-internals/blob/master/chapter2_interfaces/README.md#type-assertions

client/host_queue.go
robskillington: Ah good call, sure thing I'll update this.

robskillington: Hm, actually there's no such option `WriteTaggedBatchSize` - is there any reason why we wouldn't just have a single value that applies to both?  We don't want our config to be overly complex whenever possible.

client/host_queue.go
robskillington: Sure thing.

Makefile
robskillington: Same response as first response.

Makefile
robskillington: Sure thing.

Makefile
robskillington: Sure thing, opened issue: 
https://github.com/m3db/m3db/issues/533

Makefile
prateek: Cool, makes sense

storage/repair.go
prateek: Yeah i'm fine with leaving as is. Mind putting a quick note explaining this rationale above the type.

client/received_blocks_new_map.go
prateek: Makes sense. 

PS that's a great reference, thanks for the link!

client/host_queue.go
prateek: Yeah I can see the value in having them the same. Opened an issue to consolidate any others that need to be: https://github.com/m3db/m3db/issues/534

Makefile
prateek: Hm main concern is if someone has a dirty version of `m3x` in their workspace. Could you add checks to fail if m3x is not on `master`

storage/bootstrap/result/result.go
prateek: Cool opened https://github.com/m3db/m3db/issues/535 to track

client/received_blocks_new_map.go
prateek: hm not sure I follow. That should only matter if we're resetting the map after each pagination no?

Makefile
robskillington: Sure thing, added check.

storage/repair.go
robskillington: Sure thing.

client/received_blocks_new_map.go
prateek: talked offline: such a micro optimisation that it doesn't really matter

generated-source-files.mk
```diff
@@ -0,0 +1,124 @@
...
+m3x_package_path     := $(gopath_prefix)/$(m3x_package)
+m3x_package_min_ver  := 29bc232d9ad2e6c4a1804ccce095bb730fb1c6bc
+
+.PHONY: install-m3x-repo
+install-m3x-repo: install-glide install-generics-bin
+	# Check if repository exists, if not get it
+	test -d $(m3x_package_path) || go get -u $(m3x_package)
+	test -d $(m3x_package_path)/vendor || (cd $(m3x_package_path) && glide install)
+	test "$(shell cd $(m3x_package_path) && git rev-parse --abbrev-ref HEAD)" = "master" || ( \
+		echo "m3x not on branch master" && \
```
prateek: thought you didn't want to do this?

PR #529, opened 2018-04-16
Pull latest m3ninx changes

prateek:


client/session_test.go
```diff
@@ -197,7 +197,7 @@ func testSessionClusterConnectConsistencyLevel(
 	expected outcome,
 ) {
 	opts := newSessionTestOptions()
-	opts = opts.SetClusterConnectTimeout(3 * clusterConnectWaitInterval)
+	opts = opts.SetClusterConnectTimeout(10 * clusterConnectWaitInterval)
```
prateek: to address flakyness in CI, e.g. 

```
--- FAIL: TestSessionClusterConnectConsistencyLevelMajority (2.57s)
	<autogenerated>:1: 
                        
	Error Trace:	session_test.go:231
		
			session_test.go:154
		
	Error:		Received unexpected error "timed out establishing min connections to cluster"
```

storage/index.go
```diff
@@ -207,8 +210,16 @@ func (i *nsIndex) Query(
 		return index.QueryResults{}, errDbIndexUnableToQueryClosed
 	}
 
+	reader, err := i.active.segment.Reader()
+	if err != nil {
+		return index.QueryResults{}, err
+	}
+
+	readers := []m3ninxindex.Reader{reader}
+	exec := executor.NewExecutor(readers)
```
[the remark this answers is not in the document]
prateek: sure thing, addressed in next patch

storage/index.go
```diff
@@ -207,15 +210,28 @@ func (i *nsIndex) Query(
...
+		exec.Close()
+		reader.Close()
 		return index.QueryResults{}, err
 	}
 
 	return index.QueryResults{
-		Iterator:   index.NewIterator(i.nsID, iter, i.opts),
+		Iterator: index.NewIterator(i.nsID, iter, i.opts, func() {
+			exec.Close()
+			reader.Close()
```
jeromefroe: We can update this, and it wouldn't actually matter because we don't check the error here, but the `executor` actually takes ownership of the `readers` so it will be closed by the `exectutor`.

prateek: cheers, will do

PR #530, opened 2018-04-16
`client` Fetch implementation

prateek:


client/fetch_tagged_state.go
```diff
@@ -0,0 +1,471 @@
...
+
+// TODO(prateek): delete the nolint below once we finish implementation
+// nolint
+type fetchTaggedResultAccumulator struct {
+	// NB(prateek): a fetchTagged request requires we fan out to each shard in the
+	// topology. As a result, we model the response consistency per shard.
+	// Length of this slice == 1 + max shard id in topology
+	shardResults []fetchTaggedShardConsistencyResult // TODO(prateek): should this be moved to the fetchTaggedState(?)
+
+	responses  fetchTaggedIDResultByIDMap
```
prateek: talked offline: going to try using `[]*rpc.FetchTaggedIDResult_` + sorting instead of this map. It'll be much easier for pooling, allow stable ordered iteration, not require hashing, etc.

client/fetch_tagged_state.go
```diff
@@ -0,0 +1,471 @@
...
+	response *rpc.FetchTaggedResult_,
+	err error,
+) {
+	f.Lock()
+	defer func() {
+		f.Unlock()
+		f.decRef()
+	}()
+
+	if f.fatalError != nil {
```
prateek: add field to early terminate if success too.

client/fetch_tagged_state.go
```diff
@@ -0,0 +1,471 @@
...
+			if state != shard.Available {
+				// skip any shards that are not AVAILABLE
+				continue
+			}
+			f.results.shardResults[id].enqueued++
+		}
+	}
+
+	// TODO(prateek): should be migrated to a function in `session_util.go`
+	minimumReplicasRequired := int8(0)
```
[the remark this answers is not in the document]
prateek: done

client/fetch_tagged_state.go
```diff
@@ -0,0 +1,471 @@
...
+
+			if err == nil {
+				shardResult.success++
+			} else {
+				shardResult.errors++
+			}
+
+			if readConsistencyAchieved(f.consistencyLevel, f.majority, int(shardResult.enqueued), int(shardResult.success)) {
+				shardResult.done = true
+			} else if readConsistencyImpossible(f.consistencyLevel, f.majority, int(shardResult.enqueued), int(shardResult.errors)) {
```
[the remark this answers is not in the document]
prateek: We don't. Removed for simplicity (actually will help load too)

client/fetch_tagged_state.go
prateek: done

client/fetch_state.go
```diff
@@ -0,0 +1,164 @@
...
+const (
+	maxUint = ^uint(0)
+	maxInt  = int(maxUint >> 1)
+)
+
+var (
+	errFetchStateStillProcessing = errors.New("[invariant violated] fetch state is still processing, unable to create response")
+)
+
+type fetchState struct {
```
richardartoul: This struct could benefit from a few lines explaining what it does. I think by the time I finished reading it in-depth I had an idea, but my reading would have benefited a lot from coming in with an idea of what it was for

client/fetch_state.go
```diff
@@ -0,0 +1,164 @@
...
+
+	op                   *fetchTaggedOp
+	tagResultAccumulator fetchResultsAccumulator
+	err                  error
+	done                 bool
+
+	pool fetchStatePool
+}
+
+func (f *fetchState) reset() {
```
richardartoul: Is this method only used by the pool? If so, could we get rid of it and just use a constructor function? As is its kind of confusing because its behavior is more like init() than reset() from what I can tell

client/fetch_state.go
```diff
@@ -0,0 +1,164 @@
...
+	f.done = false
+	f.tagResultAccumulator.Clear()
+
+	if f.pool == nil {
+		return
+	}
+	f.pool.Put(f)
+}
+
+func (f *fetchState) Reset(
```
richardartoul: What about setting err = nil and done = false and decref'ing the existing f.op etc etc

client/fetch_state.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+func (f *fetchState) completionFn(
+	result interface{},
+	resultErr error,
+) {
+	f.Lock()
+	defer func() {
+		f.Unlock()
+		f.decRef() // release ref held onto by the hostQueue (via op.completionFn)
```
richardartoul: Seems like this function will be called (potentially) multiple times, are multiple decRef calls ok?

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+type fetchResultsAccumulator interface {
+	// Reset resets the accumulator for use.
+	Reset(
+		startTime time.Time,
+		endTime time.Time,
+		topoMap topology.Map,
+		majority int,
+		consistencyLevel topology.ReadConsistencyLevel)
+
+	// Accumulate add the provided result to the accumulator.
```
richardartoul: adds*

Also this method might be clearer if it was just named add, but I don't have strong feelings about it

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+
+	// AsEncodingSeriesIterators converts the underlying response into an encoding.SeriesIterators
+	AsEncodingSeriesIterators(limit int, pools fetchTaggedPools) (
+		iters encoding.SeriesIterators, exhaustive bool, err error)
+
+	// AsIndexQueryResults converts the underlying response into an index.QueryResults
+	AsIndexQueryResults(limit int, pools fetchTaggedPools) (index.QueryResults, error)
+
+	// Clear releases any held resources.
+	Clear()
```
richardartoul: Is this conceptually different than Close?

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+}
+
+var _ fetchResultsAccumulator = &fetchTaggedResultAccumulator{}
+
+type fetchTaggedResultAccumulator struct {
+	// NB(prateek): a fetchTagged request requires we fan out to each shard in the
+	// topology. As a result, we model the response consistency per shard.
+	// Length of this slice == 1 + max shard id in topology
```
richardartoul: +1 for more convenient indexing yeah?

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+		iters encoding.SeriesIterators, exhaustive bool, err error)
+
+	// AsIndexQueryResults converts the underlying response into an index.QueryResults
+	AsIndexQueryResults(limit int, pools fetchTaggedPools) (index.QueryResults, error)
+
+	// Clear releases any held resources.
+	Clear()
+}
+
+var _ fetchResultsAccumulator = &fetchTaggedResultAccumulator{}
```
richardartoul: Whats this?

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+	errors   int8
+	done     bool
+}
+
+type fetchTaggedResultAccumulatorOpts struct {
+	host     topology.Host
+	response *rpc.FetchTaggedResult_
+}
+
+func (accum *fetchTaggedResultAccumulator) Accumulate(result interface{}, resultErr error) (bool, error) {
```
richardartoul: why is result an empty interface?

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+
+	startTime        time.Time
+	endTime          time.Time
+	majority         int
+	consistencyLevel topology.ReadConsistencyLevel
+	topoMap          topology.Map
+}
+
+type fetchTaggedShardConsistencyResult struct {
+	enqueued int8
```
richardartoul: Are these numbers just bounded by the replication factor? Probably safe but I think you can bump them to uint16 and still keep the whole struct under a single word

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+	iters := pools.IteratorArray().Get(numElems)[:numElems]
+	for idx, elem := range elems {
+		slicesIter := pools.ReaderSliceOfSlicesIterator().Get()
+		slicesIter.Reset(elem.Segments)
+		multiIter := pools.MultiReaderIterator().Get()
+		multiIter.ResetSliceOfSlices(slicesIter)
+		iters[idx] = multiIter
+	}
+
+	elem := elems[0] // pick the first element as they're all identical
```
richardartoul: Can you explain the guarantee that this slice wont be empty

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+
+type fetchTaggedIDResults []*rpc.FetchTaggedIDResult_
+
+// the returned bool indicates if the iteration should be continued
+type forEachFetchTaggedIDFn func(responsesForSingleID fetchTaggedIDResults) bool
+
+// forEach iterates over the provide results, and calls `fn` on each
+// group of responses with the same ID.
+// NB: assumes the results array being operated upon has been sorted
+func (results fetchTaggedIDResults) forEach(fn forEachFetchTaggedIDFn) {
```
richardartoul: do you think there is value in calling this `forEachSorted` or is that just more confusing?

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+	var (
+		startIdx = 0
+		lastID   []byte
+	)
+	for i := 0; i < len(results); i++ {
+		elem := results[i]
+		if !bytes.Equal(elem.ID, lastID) {
+			// skip calling function when we first see an element
```
richardartoul: This comment is kind of misleading. I would leave it out or say: "We only want to call the function once we reach the last element for a given ID, but the first element in the results list needs to be special-cased" or something like that

client/fetch_tagged_results_accumulator.go
richardartoul: I take it back after reading the code. Now I think `forEachGroup` would be better.

Also, would it be too expensive to wrap the results in a struct that contains a `sorted` bool and then have this function panic if its called on an unsorted list?

client/session.go
```diff
@@ -1031,15 +1062,123 @@ func (s *session) FetchIDs(
...
+	fetchState, err := s.fetchTaggedAttemptWithRLock(opts.StartInclusive, opts.EndExclusive, req)
+	s.state.RUnlock()
+
+	if err != nil {
+		return nil, false, err
+	}
+
+	// it's safe to Wait() here, as we still own the lock on fetchState, after it's
+	// returned from fetchTaggedAttemptWithRLock.
+	fetchState.Wait()
```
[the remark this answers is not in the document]
richardartoul: oh its two separate locks ignore me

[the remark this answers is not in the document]
richardartoul: maybe change your comment from "own the lock" to "hold the lock" and then on line 1079 add a comment saying that the returned fetchState is already locked

client/session.go
```diff
@@ -1031,15 +1062,123 @@ func (s *session) FetchIDs(
...
+	}
+
+	fetchState, err := s.fetchTaggedAttemptWithRLock(opts.StartInclusive, opts.EndExclusive, req)
+	s.state.RUnlock()
+
+	if err != nil {
+		return index.QueryResults{}, err
+	}
+
+	// it's safe to Wait() here, as we still own the lock on fetchState, after it's
```
richardartoul: same comments

client/fetch_state.go
prateek: sure, i'll add a one pager `fetchTagged.md` to this change.

client/fetch_state.go
prateek: Fair. Will do to `fetch_state.go` and `write_state.go`

prateek: done

client/fetch_state.go
prateek: Don't need to do that because of the guarantees provided by the pool. We always retrieve a fetchState from the fetchStatePool prior to calling Reset. This retrieval is guaranteed to have a valid fetchState (err=nil/..) because of `fetchState.close()` for re-used states, or `newFetchState(...)` for new allocs

client/fetch_state.go
prateek: Yep. It will be called by each of the different hostQueues, hence the corresponding `incRef()` when we enqueue an op (refer `fetchTaggedAttemptWithRLock` in session.go )

client/fetch_tagged_results_accumulator.go
prateek: Both read the same to me, will rename if `Add()` reads better to you.

client/fetch_tagged_results_accumulator.go
prateek: It's similar to the degree that both Close() and Clear() release internal state.

The difference is close() releases the object that it's called upon back into a pool. Clear() does not.  This is because the user of this (fetchState) retains the reference to the accumulator, and it's pooled at the fetchState level, not the the accumulator level.

client/fetch_tagged_results_accumulator.go
prateek: yep exactly. lets me use the slice as a map basically.

client/fetch_tagged_results_accumulator.go
prateek: Incantation to make the compiler ensure the concrete type `&fetchTaggedResultAccumulator{}` implements the `fetchResultsAccumulator` interface.

client/fetch_tagged_results_accumulator.go
prateek: that's the interface for a completionFn: https://github.com/m3db/m3db/blob/master/client/op.go#L23

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,352 @@
...
+		// should never happen, as we've taken a reference to the
+		// topology when beginning the request, and the var is immutable.
+		doneAccumulating := true
+		err := fmt.Errorf(
+			"[invariant violated] missing host shard in fetchState completionFn: %s", host.ID())
+		return doneAccumulating, xerrors.NewNonRetryableError(err)
+	}
+
+	for _, hs := range hostShardSet.ShardSet().All() {
+		if hs.State() != shard.Available {
```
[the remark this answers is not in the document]
prateek: Yep, hostShardSet is per host.

And yeah, we'll be able to serve responses from other replicas based on the consistency level. Have tests for this in `integration/fetch_tagged_quorum_test.go`

client/fetch_tagged_results_accumulator.go
prateek: yeah it's bound by RF. No reason to bump, we're never going to have RF as high as `128` (max int8)

client/fetch_tagged_results_accumulator.go
prateek: This method `sliceResponsesAsSeriesIter` is called from within a lambda provided to `func (results fetchTaggedIDResults) forEach(fn forEachFetchTaggedIDFn)` which makes that guarantee. 

i.e. forEach is only called on a group, impossible to have a group with 0 members.

client/fetch_tagged_results_accumulator.go
prateek: sure

client/fetch_tagged_results_index_iterator.go
```diff
@@ -0,0 +1,110 @@
...
+		return false
+	}
+	i.release()
+	i.currentIdx++
+	if i.currentIdx >= len(i.backing.ids) {
+		return false
+	}
+
+	// parse the current docs
+	asIdent := func(b []byte) ident.ID {
```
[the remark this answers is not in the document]
prateek: Yeah it doesn't matter. Even if it isn't inlined the mem access to the pools is going to be way more expensive than the call to this function will be.

client/fetch_tagged_results_index_iterator.go
```diff
@@ -0,0 +1,110 @@
...
+	backing struct {
+		nses [][]byte
+		ids  [][]byte
+		tags [][]byte
+	}
+}
+
+var _ index.Iterator = &fetchTaggedResultsIndexIterator{}
+
+func newFetchTaggedResultsIndexIterator(pools fetchTaggedPools) *fetchTaggedResultsIndexIterator {
```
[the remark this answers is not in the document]
prateek: Take a look at how it's used in `fetchTaggedResultAccumulator.AsIndexQueryResults`: 

```
	var (
		iter  = newFetchTaggedResultsIndexIterator(pools)
		count = 0
	)
	results := fetchTaggedIDResultsSortedByID(accum.responses)
	sort.Sort(results)
	accum.responses = fetchTaggedIDResults(results)
	accum.responses.forEach(func(elems fetchTaggedIDResults) bool {
		iter.backing.ids = append(iter.backing.ids, elems[0].ID)
		iter.backing.nses = append(iter.backing.nses, elems[0].NameSpace)
		iter.backing.tags = append(iter.backing.tags, elems[0].EncodedTags)
		count++
		return count < limit
	})
```

Better to leave as is because it'll allow pooling of the underlying slices in `fetchTaggedResultsIndexIterator.backing.(nses|ids|tags)` once we start pooling `fetchTaggedResultsIndexIterator`.  Otherwise we'll have to manually pool the slices too.

client/session.go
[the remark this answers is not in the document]
prateek: lol sure.

client/session.go
prateek: done

client/fetch_tagged_results_accumulator.go
prateek: done

client/fetch_tagged_results_accumulator.go
prateek: done

client/fetch_tagged_results_index_iterator.go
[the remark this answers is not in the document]
richardartoul: ok fair

client/fetch_tagged_results_index_iterator.go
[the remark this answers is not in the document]
richardartoul: I just meant the fact that it seems like an unnecessary allocation on the hot path

client/fetch_state.go
richardartoul: did you add this or is it coming in a subsequent P.R?

client/fetch_tagged_results_accumulator.go
richardartoul: bump. Cool if you disagree just making sure you saw it, I don't have super strong feelings about it

client/fetch_tagged_results_index_iterator.go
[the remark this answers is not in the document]
prateek: Yeah that's fair, i'll make it an method on the struct type to avoid.

client/fetch_state.go
```diff
@@ -0,0 +1,169 @@
...
+	}()
+
+	if f.done {
+		// i.e. we've already failed, no need to continue processing any additional
+		// responses we receive
+		return
+	}
+
+	done, err := f.tagResultAccumulator.Add(result, resultErr)
```
robskillington: nit: Perhaps tag result accumulator should get the actual casted result, rather than `result interface{}`?  We can return an error earlier here if we know it doesn't cast correctly.

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,362 @@
...
+	// Reset resets the accumulator for use.
+	Reset(
+		startTime time.Time,
+		endTime time.Time,
+		topoMap topology.Map,
+		majority int,
+		consistencyLevel topology.ReadConsistencyLevel)
+
+	// Add adds the provided result to the accumulator.
+	Add(result interface{}, err error) (done bool, addErr error)
```
robskillington: Wouldn't it be best to have this as a `fetchTaggedResultAccumulatorOpts` and at the layer above raise an error if casting fails?

client/fetch_state.go
```diff
@@ -0,0 +1,169 @@
...
+
+	limit := maxInt
+	if l := f.op.request.Limit; l != nil {
+		limit = int(*l)
+	}
+
+	return f.tagResultAccumulator.AsEncodingSeriesIterators(limit, pools)
+}
+
+type fetchTaggedPools interface {
```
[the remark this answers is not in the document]
prateek: Only have it as an interface to override with a stub implementation of the pool in tests (see `fetch_state_test.go`). Won't the runtime optimize away the itab considering there's only one backing implementation?

client/fetch_tagged_results_accumulator.go
prateek: yeah fair. will do in next patch.

client/fetch_state.go
prateek: sure thing

client/fetch_tagged_results_accumulator.go
prateek: done

client/fetch_state.go
prateek: done

client/fetch_tagged_results_accumulator.go
prateek: `forEachGroup` is more confusing to me, i'll rename to `forEachID` in the next patch.

prateek: done

client/fetch_state.go
robskillington: Might I recommend instead adding just a `README.md` to the client package and capturing it somewhere there?  Then we can at least use a pattern for naming this stuff on a per package basis.

client/fetch_state.go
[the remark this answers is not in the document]
robskillington: Yeah I'm not worried about the optimization.  Guess I don't see how different it is though using an interface or struct for testing purposes too since you can just as easily override the struct field in a test vs using `pools.EXPECT().SomeAccessor().Return(...)`?

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,366 @@
...
+		if hs.State() != shard.Available {
+			// Currently, we only accept responses from shard's which are available
+			// NB: as a possible enhancement, we could accept a response from
+			// a shard that's not available if we tracked response pairs from
+			// a LEAVING+INITIALIZING shard; this would help during node replaces.
+			continue
+		}
+
+		shardID := int(hs.ID())
+		shardResult := accum.shardResults[shardID]
```
robskillington: Is it worth bounds checking here?

client/fetch_state.go
prateek: added a `client/README.md`

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,366 @@
...
+	result.Reset(numElements)
+	count := 0
+	accum.responses.forEachID(func(elems fetchTaggedIDResults) bool {
+		seriesIter := accum.sliceResponsesAsSeriesIter(pools, elems)
+		result.SetAt(count, seriesIter)
+		count++
+		return count < limit
+	})
+
+	exhaustive := accum.exhaustive && count < limit
```
robskillington: Should this be `count <= limit`?  If so, maybe we should add a "test off by one" test case for this? (I know its such a minor issue.. shrug)

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,366 @@
...
+	})
+
+	result := pools.MutableSeriesIterators().Get(numElements)
+	result.Reset(numElements)
+	count := 0
+	accum.responses.forEachID(func(elems fetchTaggedIDResults) bool {
+		seriesIter := accum.sliceResponsesAsSeriesIter(pools, elems)
+		result.SetAt(count, seriesIter)
+		count++
+		return count < limit
```
robskillington: Should this be `count < limit` here too?

client/fetch_state.go
prateek: lol didn't see your comment @robskillington, ended up doing that.

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,366 @@
...
+	accum.responses.forEachID(func(elems fetchTaggedIDResults) bool {
+		iter.backing.ids = append(iter.backing.ids, elems[0].ID)
+		iter.backing.nses = append(iter.backing.nses, elems[0].NameSpace)
+		iter.backing.tags = append(iter.backing.tags, elems[0].EncodedTags)
+		count++
+		return count < limit
+	})
+
+	return index.QueryResults{
+		Exhaustive: accum.exhaustive && count < limit,
```
robskillington: Perhaps this one too?

client/fetch_tagged_results_accumulator.go
prateek: Don't need to, we explicitly ensure `shardResults` is big enough in the `Reset()` method.

client/fetch_tagged_results_accumulator.go
prateek: will do

client/fetch_tagged_results_accumulator.go
prateek: fair, will do.

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,366 @@
...
+		Iterator:   iter,
+	}, nil
+}
+
+type fetchTaggedShardConsistencyResults []fetchTaggedShardConsistencyResult
+
+func (res fetchTaggedShardConsistencyResults) initialize(length int) fetchTaggedShardConsistencyResults {
+	if cap(res) < length {
+		res = append(res, make(fetchTaggedShardConsistencyResults, length-cap(res))...)
```
robskillington: Hm, this is kind of gross making a temporary slice only to throw it away likely (if you have to grow it).  You're not looking to keep the original slice contents around and it only ever gets called once at initialization yeah?  Perhaps just reallocate by hand?

i.e.
```go
if cap(res) < length {
  res = make(fetchTaggedShardConsistencyResults, length)
}
```

client/fetch_state.go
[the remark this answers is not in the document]
prateek: Ah sorry thought you were talking about another pool. Put a blurb explaining what we talked about offline.

client/fetch_tagged_results_accumulator.go
```diff
@@ -0,0 +1,366 @@
...
+	)
+	for i := 0; i < len(results); i++ {
+		elem := results[i]
+		if !bytes.Equal(elem.ID, lastID) {
+			// We only want to call the the forEachID fn once we have calculated the entire group,
+			// i.e. once we have gone past the last element for a given ID, but the first element
+			// in the results slice is a special case because we are always starting a new group
+			// at that point.
+			if i == 0 {
+				lastID = elem.ID
```
robskillington: Perhaps place `lastID = elem.ID` as first line of inside `if !bytes.Equal(...) { ... }` clause?  that way don't have to duplicate it twice.

client/fetch_tagged_results_index_iterator.go
```diff
@@ -0,0 +1,111 @@
...
+	current struct {
+		nsID ident.ID
+		tsID ident.ID
+		tags serialize.TagDecoder
+	}
+
+	backing struct {
+		nses [][]byte
+		ids  [][]byte
+		tags [][]byte
```
robskillington: These aren't pooled for now I suppose?  I guess they come off the wire from TChannel so not worth it.

client/fetch_tagged_results_accumulator.go
prateek: done

client/fetch_tagged_results_accumulator.go
prateek: done

client/session.go
```diff
@@ -1031,15 +1062,123 @@ func (s *session) FetchIDs(
...
+		fetchState.incRef()
+		if err := hq.Enqueue(op); err != nil {
+			fetchState.Unlock()
+			op.decRef()         // release the ref for the current go-routine
+			fetchState.decRef() // release the ref for the hostQueue
+			fetchState.decRef() // release the ref for the current go-routine
+
+			// NB: if this happens we have a bug, once we are in the read
+			// lock the current queues should never be closed
+			s.opts.InstrumentOptions().Logger().Errorf(
```
robskillington: You can use `s.log.Errorf(...)` here, we already cache the logger.

client/fetch_tagged_results_accumulator.go
prateek: Yeah that's completely fair. I'm doing an alloc anyway, much easier to read if do the full alloc I need. Yep, it'll only happen if max shards change ever.

client/fetch_tagged_results_accumulator.go
prateek: Sure thing.

client/fetch_tagged_results_accumulator.go
robskillington: Yeah fair, topoMap is captured already.

client/fetch_state.go
```diff
@@ -0,0 +1,187 @@
...
+	errFetchStateStillProcessing = errors.New("[invariant violated] fetch state is still processing, unable to create response")
+)
+
+type fetchState struct {
+	sync.Cond
+	sync.Mutex
+	refCounter
+
+	op                   *fetchTaggedOp
+	tagResultAccumulator fetchResultsAccumulator
```
robskillington: Is this actually mocked anywhere, and if not should we just embed the struct perhaps instead of use an interface here?  Just thinking about the whole "by value" ideology to reduce GC pressure argument that Richard Hudson was making.

prateek: talked offline: going to move to a non-ptr struct (non-interface type). can do w/e refactoring we need when we come back for https://github.com/m3db/m3db/issues/545

PR #532, opened 2018-04-17
Change SeriesIterator to allow clients to access underlying block byte arrays without expansion

arnikola:


x/xio/types.go
```diff
@@ -77,6 +100,6 @@ type ReaderSliceOfSlicesIterator interface {
 type ReaderSliceOfSlicesFromSegmentReadersIterator interface {
```
prateek: should this be renamed `ReaderSliceOfSlicesFromBlockReadersIterator`

x/xio/types.go
```diff
@@ -63,11 +82,15 @@ type ReaderSliceOfSlicesIterator interface {
 	// Next moves to the next item
 	Next() bool
 
+	CurrentStart() time.Time
```
prateek: would it be better to consolidate into a single function?

```
  Next() bool
  
  // Current returns ... 
  Current() (start time.Time, end Time.Time, len int)
```

also, please add comments to the methods

x/xio/types.go
```diff
@@ -77,6 +100,6 @@ type ReaderSliceOfSlicesIterator interface {
 type ReaderSliceOfSlicesFromSegmentReadersIterator interface {
 	ReaderSliceOfSlicesIterator
 
-	// Reset resets the iterator with a new array of segment readers arrays
-	Reset(segments [][]SegmentReader)
+	// Reset resets the iterator with a new array of block readers arrays
+	Reset(segments [][]BlockReader)
```
prateek: s/segments/blocks

x/xio/types.go
```diff
@@ -22,21 +22,40 @@ package xio
...
+	// Clone returns a clone of the underlying data reset
+	// to the start of the reader
+	Clone() Reader
+}
+
+// BlockReader implements the io reader interface backed by a segment with start and end times
+type BlockReader interface {
+	SegmentReader
+
+	Start() time.Time
```
prateek: nit: comments on these methods

client/reader_slice_of_slices_iterator.go
```diff
@@ -111,11 +114,34 @@ func (it *readerSliceOfSlicesIterator) CurrentLen() int {
 	return len(it.segments[it.idx].Unmerged)
 }
 
-func (it *readerSliceOfSlicesIterator) CurrentAt(idx int) io.Reader {
+func timeConvert(ticks *int64) time.Time {
+	if ticks == nil {
+		return time.Time{}
+	}
+	return xtime.FromNormalizedTime(*ticks, time.Second)
```
robskillington: Hm, we probably should be using nanoseconds here, in case block start and end times are at millisecond resolution for instance.

x/xio/block_reader.go
```diff
@@ -0,0 +1,80 @@
...
+	r.reader.Reset(segment)
+	r.start = start
+	r.end = end
+}
+
+func (r *blockReader) Finalize() {
+	r.reader.Finalize()
+}
+
+func (r *blockReader) SegmentReader() (SegmentReader, error) {
```
prateek: could you skip the error return if it's always nil

client/reader_slice_of_slices_iterator.go
```diff
@@ -111,11 +114,34 @@ func (it *readerSliceOfSlicesIterator) CurrentLen() int {
...
+}
+
+func (it *readerSliceOfSlicesIterator) CurrentStart() time.Time {
+	segments := it.segments[it.idx]
+	if segments.Merged != nil {
+		return timeConvert(segments.Merged.StartTime)
+	}
+	return timeConvert(segments.Unmerged[0].StartTime)
```
robskillington: Perhaps just make sure that `len(segments.Unmerged) > 0` before this to avoid nil ptr deref if there isn't an entry?

client/session.go
```diff
@@ -660,13 +659,14 @@ func (s *session) setTopologyWithLock(topoMap topology.Map, queues []hostQueue,
 	s.fetchBatchOpArrayArrayPool.Init()
 
 	if s.iteratorArrayPool == nil {
-		s.iteratorArrayPool = encoding.NewIteratorArrayPool([]pool.Bucket{
```
robskillington: Hm, any reason to remove this pool?  We used to be able to pool these slice allocations, however this change removes the ability for those to be pooled?

encoding/encodingtest/series_iterator_split_into_blocks_test.go
```diff
@@ -0,0 +1,170 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package encodingtest
```
prateek: +1 for blackbox unit tests like this! 

more idiomatic to have this file in the package being tested, could you move down to `encoding` and change the package name in the file to `encoding_test`. It'll give you the same property you're after

client/session.go
```diff
@@ -2816,17 +2823,22 @@ func (b *baseBlocksResult) newDatabaseBlock(block *rpc.Block) (block.DatabaseBlo
...
+				end = endTime
+			}
 		}
-
-		encoder, err := b.mergeReaders(start, readers)
+		encoder, err := b.mergeReaders(start, end, readers)
 		for _, reader := range readers {
 			// Close each reader
-			segmentReader := reader.(xio.SegmentReader)
+			segmentReader := reader.(xio.BlockReader)
```
robskillington: Perhaps rename this to `blockReader := ...`

encoding/encodingtest/series_iterator_split_into_blocks_test.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+type SeriesBlock struct {
+	Start          time.Time
+	End            time.Time
+	Replicas       []encoding.MultiReaderIterator
+	ValuesIterator encoding.SeriesIterator
+}
+
+func TestDeconstructAndReconstruct(t *testing.T) {
```
prateek: suggestion(take/leave): might be more easier to follow if this was written as an `Example...`. Feel free to ignore if you don't think it's worth the hassle.

encoding/encodingtest/series_iterator_split_into_blocks_test.go
```diff
@@ -0,0 +1,170 @@
...
+		fmt.Printf("block at %v with %d replicas\n", block.Start, len(block.Replicas))
+		series.Blocks[i].ValuesIterator = valuesIter
+	}
+
+	// Now show how we can iterate per block
+	for _, block := range series.Blocks {
+		iter := block.ValuesIterator
+		for iter.Next() {
+			dp, _, _ := iter.Current()
+			fmt.Printf("%s value at %v: %v\n", series.ID.String(), dp.Timestamp, dp.Value)
```
robskillington: Instead of printing output, can you change this to actually use assert/require to assert the values?

encoding/encodingtest/series_iterator_split_into_blocks_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+		filterValuesEnd := orig.End()
+		if block.End.Before(filterValuesEnd) {
+			filterValuesEnd = block.End
+		}
+
+		valuesIter := encoding.NewSeriesIterator(orig.Namespace(), orig.ID(),
+			filterValuesStart, filterValuesEnd, block.Replicas, nil)
+
+		fmt.Printf("block at %v with %d replicas\n", block.Start, len(block.Replicas))
```
robskillington: Instead of printing output, can you change this to actually use assert/require to assert the values?

encoding/series_iterator_test.go
```diff
@@ -173,12 +175,13 @@ func assertTestSeriesIterator(
 	assert.Equal(t, series.start, iter.Start())
 	assert.Equal(t, series.end, iter.End())
 	for i := 0; i < len(series.expected); i++ {
+		fmt.Println("ITERATION", i, "OF", len(series.expected))
```
robskillington: Remove this yeah?

encoding/types.go
```diff
@@ -166,7 +168,9 @@ type SeriesIterator interface {
 	// must note that this can be an array with nil entries if some replicas did not return successfully.
 	// NB: the SeriesIterator assumes ownership of the provided ids, this includes calling `id.Finalize()` upon
 	// iter.Close().
-	Reset(id ident.ID, ns ident.ID, startInclusive, endExclusive time.Time, replicas []Iterator)
+	Reset(id ident.ID, ns ident.ID, startInclusive, endExclusive time.Time, replicas []MultiReaderIterator)
+
+	Replicas() []MultiReaderIterator
```
robskillington: Perhaps add a comment explaining this method?

encoding/types.go
```diff
@@ -135,10 +135,12 @@ type MultiReaderIterator interface {
 	Iterator
 
 	// Reset resets the iterator to read from a slice of readers.
-	Reset(readers []io.Reader)
+	Reset(readers []xio.Reader, start, end time.Time)
 
 	// Reset resets the iterator to read from a slice of slice readers.
 	ResetSliceOfSlices(readers xio.ReaderSliceOfSlicesIterator)
+
+	Readers() xio.ReaderSliceOfSlicesIterator
```
robskillington: Perhaps add a comment explaining this method?

generated/thrift/rpc/rpc.go
```diff
@@ -28,7 +28,6 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
-
```
robskillington: Any reason to remove this separation of go imports?  If not perhaps add it back for separation of stdlib imports and external imports.

network/server/tchannelthrift/convert/convert.go
```diff
@@ -124,9 +124,13 @@ func ToSegments(readers []xio.SegmentReader) (ToSegmentsResult, error) {
 		if seg.Len() == 0 {
 			return ToSegmentsResult{}, nil
 		}
+		startTime := xtime.ToNormalizedTime(readers[0].Start(), time.Second)
+		endTime := xtime.ToNormalizedTime(readers[0].End(), time.Second)
```
robskillington: Yeah we should be using nanoseconds like we do in our other RPCs fields.

network/server/tchannelthrift/convert/convert.go
```diff
@@ -143,9 +147,13 @@ func ToSegments(readers []xio.SegmentReader) (ToSegmentsResult, error) {
 		if seg.Len() == 0 {
 			continue
 		}
+		startTime := xtime.ToNormalizedTime(reader.Start(), time.Second)
+		endTime := xtime.ToNormalizedTime(reader.End(), time.Second)
```
robskillington: Nanoseconds here too.

persist/fs/retriever.go
```diff
@@ -533,6 +532,10 @@ func (req *retrieveRequest) Reset(segment ts.Segment) {
 	req.resultWg.Done()
 }
 
+func (req *retrieveRequest) Clone() xio.Reader {
+	return req
```
robskillington: Hm, it'd be more kosher to:
```go
  req.resultWg.Wait() // wait until result is ready
  if req.err != nil { return nil, err }
  return req.reader.Clone()
```

We probably need to change the `Clone() xio.Reader` interface to be `Clone() (xio.Reader, error)` so callers need to check the error.

storage/shard.go
```diff
@@ -1075,6 +1075,7 @@ func (s *dbShard) FetchBlocks(
 	ctx context.Context,
 	id ident.ID,
 	starts []time.Time,
+	blockDuration time.Duration,
```
prateek: don't need to take this as an arg, the shard has it available via: `dbShard.namespace.Options().RetentionOptions().BlockSize()`

storage/block/block.go
```diff
@@ -176,45 +176,50 @@ func (b *dbBlock) Stream(blocker context.Context) (xio.SegmentReader, error) {
 
 	b.ctx.DependsOn(blocker)
 
+	start := b.startWithLock()
+	end := b.LastReadTime()
```
robskillington: The end time of the block is actually start + block size.

The last read time is when the block was last accessed.

Hm, we might need to make the block constructor take a start time and an end time, along with the Reset(...) call.

storage/block/merged_block_reader.go
```diff
@@ -104,11 +106,23 @@ func (r *dbMergedBlockReader) mergedReader() (xio.SegmentReader, error) {
...
 	}
 
-	r.merged = r.encoder.Stream()
+	r.merged = xio.NewBlockReader(r.encoder.Stream(), r.blockStart, r.blockEnd)
 
 	return r.merged, nil
 }
 
+func (r *dbMergedBlockReader) Clone() xio.Reader {
+	return r
```
robskillington: It can't really just return itself as the position into the stream needs to be reset.  You'll need to return `newDatabaseMergedBlockReader(r.blockStart, r.blockEnd, r.streams[0].Clone(), r.streams[1].Clone(), r.opts)` or similar.

storage/series/buffer.go
```diff
@@ -581,9 +580,12 @@ func (b *dbBufferBucket) streams(ctx context.Context) []xio.SegmentReader {
 		}
 	}
 	for i := range b.encoders {
+		start := b.start
+		end := b.lastRead()
```
robskillington: Can't use last read here either, needs to use `end := start.Add(b.opts.RetentionOptions().BlockSize())` or similar.

storage/series/buffer.go
```diff
@@ -672,13 +674,16 @@ func (b *dbBufferBucket) merge() (mergeResult, error) {
 		}
 	}
 
-	readers := make([]io.Reader, 0, len(b.encoders)+len(b.bootstrapped))
-	streams := make([]xio.SegmentReader, 0, len(b.encoders))
+	start := b.start
+	end := b.lastRead()
```
robskillington: Same, can't use last read here.

storage/series/reader.go
```diff
@@ -202,14 +208,16 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
 							r.id.String(), start, err))
 					res = append(res, r)
 				}
+				// stream.End()
```
robskillington: Don't need this commented code perhaps?

storage/series/reader.go
```diff
@@ -171,13 +175,15 @@ func (r Reader) readersWithBlocksMapAndBuffer(
...
+	blockDuration time.Duration,
 ) ([]block.FetchBlockResult, error) {
-	return r.fetchBlocksWithBlocksMapAndBuffer(ctx, starts, nil, nil)
+	return r.fetchBlocksWithBlocksMapAndBuffer(ctx, starts, blockDuration, nil, nil)
 }
 
 func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
 	ctx context.Context,
 	starts []time.Time,
+	blockDuration time.Duration,
```
robskillington: nit: We use `blockSize` as the convention for passing the length of the block time window.

storage/series/reader.go
```diff
@@ -171,13 +175,15 @@ func (r Reader) readersWithBlocksMapAndBuffer(
 func (r Reader) FetchBlocks(
 	ctx context.Context,
 	starts []time.Time,
+	blockDuration time.Duration,
```
robskillington: nit: We use `blockSize` as the convention for passing the length of the block time window.

storage/series/reader.go
```diff
@@ -126,7 +126,9 @@ func (r Reader) readersWithBlocksMapAndBuffer(
 					return nil, err
 				}
 				if stream != nil {
-					results = append(results, []xio.SegmentReader{stream})
+					results = append(results, []xio.BlockReader{
+						xio.NewBlockReader(stream, start, end),
```
robskillington: The block start and end time though is not the query's start and end time, it's different per block.

Can you just perhaps make the block interface return a `BlockReader` which would make more sense here?

storage/series/reader.go
```diff
@@ -226,7 +234,9 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
 					res = append(res, r)
 				}
 				if stream != nil {
-					r := block.NewFetchBlockResult(start, []xio.SegmentReader{stream}, nil)
+					end := start.Add(blockDuration)
```
robskillington: As per previous comment, instead of passing the block duration / block size here,  let's just make the underlying blocks return an `xio.BlockReader`.

storage/series/series.go
```diff
@@ -319,7 +319,7 @@ func (s *dbSeries) FetchBlocks(
 		id:         s.id,
 		retriever:  s.blockRetriever,
 		onRetrieve: s.onRetrieveBlock,
-	}.fetchBlocksWithBlocksMapAndBuffer(ctx, starts, s.blocks, s.buffer)
+	}.fetchBlocksWithBlocksMapAndBuffer(ctx, starts, time.Duration(0), s.blocks, s.buffer)
```
robskillington: This wouldn't work right?  Otherwise end time will always be start time.  As per previous comment though, in `storage/series/reader.go` instead of passing the block duration / block size here,  let's just make the underlying blocks and block retriever return a `xio.BlockReader`.

storage/shard.go
robskillington: As per previous comment, instead of passing the block duration / block size here,  let's just make the underlying blocks and block retriever return an `xio.BlockReader`.

x/xio/block_reader.go
```diff
@@ -0,0 +1,80 @@
...
+func NewBlockReader(reader SegmentReader, start, end time.Time) BlockReader {
+	return &blockReader{reader: reader, start: start, end: end}
+}
+
+func (r *blockReader) Clone() Reader {
+	reader := r.reader.Clone()
+	if sr, ok := reader.(SegmentReader); ok {
+		return NewBlockReader(sr, r.start, r.end)
+	}
+	panic("segmentreader cloned to invalid type")
```
robskillington: Hm, yeah once we make `Clone() Reader` be `Clone() (Reader, error)` we can avoid the panic here and just return an error in that case.

x/xio/reader_slice_of_slices_from_segment_readers_iterator.go
```diff
@@ -46,18 +48,26 @@ func (it *readerSliceOfSlicesIterator) Next() bool {
 	return true
 }
 
+func (it *readerSliceOfSlicesIterator) CurrentStart() time.Time {
+	return it.blocks[it.arrayIdx()][0].Start()
```
robskillington: It might definitely be worth checking if `len(...) > 0` and if not return just `time.Time{}` here to be safe.

x/xio/reader_slice_of_slices_from_segment_readers_iterator.go
```diff
@@ -46,18 +48,26 @@ func (it *readerSliceOfSlicesIterator) Next() bool {
...
+}
+
+func (it *readerSliceOfSlicesIterator) CurrentEnd() time.Time {
+	return it.blocks[it.arrayIdx()][0].End()
```
robskillington: It might definitely be worth checking if `len(...) > 0` and if not return just `time.Time{}` here to be safe.

client/reader_slice_of_slices_iterator.go
arnikola: Sure, sounds good. I thought we were truncating to seconds, but I guess it's better to lose precision as late as possible rather than early on :)

client/reader_slice_of_slices_iterator.go
arnikola: :+1:

encoding/series_iterator_test.go
arnikola: I'm so glad this is the one I accidentally left in haha

storage/series/reader.go
arnikola: Oops

storage/series/series.go
arnikola: My bad, I thought in this path, where seriesBlocks was not nil, it would use the `stream.End()` values for each block, but if it were nil here or if there weren't enough blocks passed through it would fail

encoding/encodingtest/series_iterator_split_into_blocks_test.go
arnikola: Sounds good, didn't know this was a thing!

client/session.go
arnikola: s.iteratorArrayPool is changed to be a MultiReaderIteratorPool rather than a generic IteratorPool. I figured it would be better to keep the size the same as previously, and keep it as a separate pool from the other MultiReaderIteratorPool to ensure we have the same amount of pooled iterators

x/xio/types.go
```diff
@@ -63,20 +84,20 @@ type ReaderSliceOfSlicesIterator interface {
 	// Next moves to the next item
 	Next() bool
 
-	// CurrentLen returns the current slice of readers
-	CurrentLen() int
+	// Current returns the current length, start, and end time
+	Current() (int, time.Time, time.Time)
```
prateek: nit: could you name the return vars - 

```
Current() (length int, start time.Time, end time.Time)
```

x/xio/block_reader.go
prateek: Fair enough.

x/xio/block_reader_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+package xio
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/golang/mock/gomock"
+	"github.com/m3db/m3db/ts"
```
prateek: nit: import order

x/xio/block_reader.go
```diff
@@ -0,0 +1,86 @@
...
+func NewBlockReader(reader SegmentReader, start, end time.Time) BlockReader {
+	return &blockReader{reader: reader, start: start, end: end}
+}
+
+func (r *blockReader) Clone() (Reader, error) {
+	reader, err := r.reader.Clone()
+	if err != nil {
+		return nil, err
+	}
+	if sr, ok := reader.(SegmentReader); ok {
```
prateek: hm, could you avoid this cast if you moved the `Clone()` method to the sub-types (SegmentReader/BlockReader - could return the concrete types you care about that way) ?

client/fetch_state.go
```diff
@@ -180,7 +180,7 @@ type fetchTaggedPools interface {
 	MultiReaderIterator() encoding.MultiReaderIteratorPool
 	SeriesIterator() encoding.SeriesIteratorPool
 	MutableSeriesIterators() encoding.MutableSeriesIteratorsPool
-	IteratorArray() encoding.IteratorArrayPool
+	IteratorArray() encoding.MultiReaderIteratorPool
```
prateek: nit: rename the method name too

client/fetch_tagged_results_accumulator.go
```diff
@@ -192,7 +192,7 @@ func (accum *fetchTaggedResultAccumulator) sliceResponsesAsSeriesIter(
 	elems fetchTaggedIDResults,
 ) encoding.SeriesIterator {
 	numElems := len(elems)
-	iters := pools.IteratorArray().Get(numElems)[:numElems]
+	iters := make([]encoding.MultiReaderIterator, numElems)
```
prateek: why not get from a pool here?

client/session.go
```diff
@@ -717,13 +716,14 @@ func (s *session) setTopologyWithLock(topoMap topology.Map, queues []hostQueue,
 	s.pools.fetchBatchOpArrayArray.Init()
 
 	if s.pools.iteratorArray == nil {
```
prateek: nit: rename `iteratorArray` to something more appropriate?

client/fetch_state.go
prateek: actually, you can just delete this method - the interface already has `MultiReaderIterator()` returning the type you're after.

client/session.go
```diff
@@ -717,13 +716,14 @@ func (s *session) setTopologyWithLock(topoMap topology.Map, queues []hostQueue,
...
-			},
-		})
-		s.pools.iteratorArray.Init()
+		size := replicas * s.opts.SeriesIteratorPoolSize()
+		poolOpts := pool.NewObjectPoolOptions().
+			SetSize(size).
+			SetInstrumentOptions(s.opts.InstrumentOptions().SetMetricsScope(
+				s.scope.SubScope("iterator-pool"),
+			))
+		s.pools.iteratorArray = encoding.NewMultiReaderIteratorPool(poolOpts)
```
prateek: don't think you need this, `s.pools.multiReaderIterator` already has the type you're trying to migrate to

client/session.go
```diff
@@ -1405,7 +1409,10 @@ func (s *session) fetchIDsAttempt(
 		}
 
 		// Once we've enqueued we know how many to expect so retrieve and set length
-		results = s.pools.iteratorArray.Get(int(enqueued))
+		results = make([]encoding.MultiReaderIterator, 0, enqueued)
```
prateek: why not use the pooled type here?

client/session.go
prateek: still applicable?

client/session_pools.go
```diff
@@ -57,7 +57,7 @@ func (s sessionPools) SeriesIterator() encoding.SeriesIteratorPool {
 	return s.seriesIterator
 }
 
-func (s sessionPools) IteratorArray() encoding.IteratorArrayPool {
+func (s sessionPools) IteratorArray() encoding.MultiReaderIteratorPool {
```
prateek: as above, I don't think you need these changes. the pool you're after is already present in this struct/interface.

encoding/series_iterator.go
```diff
@@ -111,21 +112,28 @@ func (it *seriesIterator) Close() {
...
+
+func (it *seriesIterator) Reset(id ident.ID, nsID ident.ID, tags ident.TagIterator, startInclusive, endExclusive time.Time, replicas []MultiReaderIterator) {
 	it.id = id
 	it.nsID = nsID
 	it.tags = tags
 	it.start = startInclusive
 	it.end = endExclusive
 	it.iters.reset()
 	it.iters.setFilter(startInclusive, endExclusive)
+	it.multiReaderIters = it.multiReaderIters[:0]
```
prateek: Have to clear the elements before resetting the length of the slice or this leaks. i.e.

```
for idx := range it.multiReaderIters {
  it.mutliReaderIters[idx] = nil
}
it.multiReaderIters = it.multiReaderIters[:0]
```

x/xio/block_reader.go
arnikola: sgtm, I've done a pass and it should only be really needed to clone `SegmentReader` types

client/session.go
arnikola: Must have accidentally undone the fix

client/session_pools.go
arnikola: Fixed as this is an array pool now

x/xio/types.go
```diff
@@ -22,16 +22,27 @@ package xio
...
-	"github.com/m3db/m3x/resource"
 	"github.com/m3db/m3db/ts"
+	"github.com/m3db/m3x/resource"
 )
 
-// ReaderSliceReader implements the io reader interface backed by a slice of readers
-type ReaderSliceReader interface {
-	io.Reader
+// BlockReader implements the io reader interface backed by a segment with start and end times
+type BlockReader interface {
```
prateek: was talking to Rob offline: could you change this so BlockReader is just a struct?

```
type BlockReader struct {
  Start time.Time
  End time.Time
  SegmentReader SegmentReader
}
```

encoding/iterator_array_pool.go
```diff
@@ -34,46 +34,46 @@ import (
 // array pool.
 
 // TODO(r): instrument this to tune pooling
-type iteratorArrayPool struct {
+type multiReaderIteratorArrayPool struct {
```
prateek: nit: rename file to `multi_reader_iterator_array_pool.go`

encoding/series_iterator.go
```diff
@@ -111,21 +112,33 @@ func (it *seriesIterator) Close() {
...
+func (it *seriesIterator) Reset(id ident.ID, nsID ident.ID, tags ident.TagIterator, startInclusive, endExclusive time.Time, replicas []MultiReaderIterator) {
 	it.id = id
 	it.nsID = nsID
 	it.tags = tags
 	it.start = startInclusive
 	it.end = endExclusive
 	it.iters.reset()
 	it.iters.setFilter(startInclusive, endExclusive)
+
+	for idx := range it.multiReaderIters {
```
prateek: So pooling `[]MultiReaderIterator` introduces a little more nuance to this:
- `seriesIterator.Reset()` must transfer the ownership of the `[]MultiReaderIterator` from the caller to `seriesIterator`
- This means the calls to `.Put` the slice back into the pool that are currently done in `client/session.go` must only be done in case of error; in case of success, `.Put` must be done by `seriesIterator.Close()`
- This means the seriesIterator needs to know about the pool (i.e. `NewSeriesIterator` needs to take a param for the `MultiReaderIteratorArrayPool`

Welcome to the joys of pooling without malloc/destructors :D

prateek: Actually, scratch all that. I think the only change that's required is:
- moving the `it.multiReaderIters` cleanup (for loop setting to nil/setting to slice of len 0) to Close()
- ensure you do `it.multiReaderIters = it.multiReaderIters[:0]` in `Reset()`

This way pooling the `seriesIterator` pools the underlying slices too.

x/xio/types.go
```diff
@@ -22,18 +22,22 @@ package xio
...
 	"github.com/m3db/m3db/ts"
+	"github.com/m3db/m3x/resource"
 )
 
-// ReaderSliceReader implements the io reader interface backed by a slice of readers
-type ReaderSliceReader interface {
-	io.Reader
-
-	Readers() []io.Reader
+// Block implements the io reader interface backed by a segment with start and end times
```
prateek: nit: no longer implements the interface. need to update the comment

x/xio/types.go
```diff
@@ -63,20 +70,20 @@ type ReaderSliceOfSlicesIterator interface {
 	// Next moves to the next item
 	Next() bool
 
-	// CurrentLen returns the current slice of readers
-	CurrentLen() int
+	// Current returns the current length, start, and end time
+	Current() (length int, start, end time.Time)
```
prateek: hm did a quick search of all the places that use this method in the code ([sourcegraph](https://sourcegraph.com/github.com/m3db/m3db@312f5cca55ac3a64c32febd3a4798fa12d7e538d/-/blob/x/xio/types.go#L74:2$references) is dope af), the new version is only used in tests. i.e. We could retain `CurrentLen()` (like the original) and it'd be fine.

storage/block/types.go
```diff
@@ -66,9 +66,9 @@ type FilteredBlocksMetadataIter interface {
 // FetchBlockResult captures the block start time, the readers for the underlying streams, the
 // corresponding checksum and any errors encountered.
 type FetchBlockResult struct {
-	Start   time.Time
-	Readers []xio.SegmentReader
-	Err     error
+	Start  time.Time
```
prateek: Any reason to retain this field (Start) now that each Block has it?

storage/block/types.go
```diff
@@ -169,12 +172,13 @@ type DatabaseBlock interface {
 	// from storage to serve as a memory cached block for reads.
 	IsCachedBlock() bool
 
-	// Reset resets the block start time and the segment.
-	Reset(startTime time.Time, segment ts.Segment)
+	// Reset resets the block start time, duration, and the segment.
+	Reset(startTime time.Time, blockSize time.Duration, segment ts.Segment)
```
prateek: maybe this should just take a `xio.Block` instead?

storage/block/types.go
arnikola: I was thinking the same thing, but figured checking the array has values and then getting Start from the first element is a bit clunkier than retaining this field? I can remove it if preferred

storage/block/types.go
arnikola: I think this gets a little messy in that a lot of the callers of Reset() would probably want to avoid building the full BlockReader? If you think otherwise I can make the change

x/xio/types.go
arnikola: The tool doesn't seem to show this usage for some reason
https://github.com/m3db/m3db/pull/532/files/312f5cca55ac3a64c32febd3a4798fa12d7e538d#diff-d28c20c8fa3884aa01db69b76c5d300aR59

PR #543, opened 2018-04-25
Fix various bugs that were causing data loss during bootstrap

richardartoul:
This P.R address four different bugs that collectively contributed towards data loss during node bootstrap:

- [x] Modify MinMax on series buffer to take into account whether or not a series buffer bucket was drained or not. This is important because if we don't do this, the bootstrap process will end up putting bootstrapped blocks into buckets that are already drained and the data will be lost forever. In addition, it changes the series buffer Bootstrap method to return an error if the caller tries to bootstrap a block it contains, but that has already been drained.
- [x] Modify storage package to prevent shards/block combinations from being flush until a complete tick has elapsed since blockStart.Add(blockSize).Add(bufferPast). This will guarantee that at least a complete tick elapses between bootstrapped blocks being loaded into a series buffer bucket and the series Flush method being called. This is important because without this restriction, it was possible for bootstrapped blocks to be loaded into a series buffer bucket that was not yet drained, but would also not get drained before flush and the bootstrapped blocks would be lost.
- [x] Fix the series.Bootstrap and mergeBlockWithLock methods to not throw away bootstrapped blocks when the merge codepath was triggered. Previously, we would end up merging blocks into the wrong object, and the object that they were merged into would get overwritten and blocks would be lost.
- [x] Fix the Block.Merge() and block.Stream() functions to not cause double-free of the segment reads (which would lead to panics) 
- [x] Fix the Block.Checksum() and Block.Stream() functions to perform a merge if a merge target is present, and save the outcome of the merge as well as update the stored checksum. This was causing issues because we weren't calculating a new checksum after merge which means we were writing data to disk that we couldn't read later because the checksums wouldn't match.


integration/bootstrap_after_buffer_rotation_regression_test.go
```diff
@@ -0,0 +1,187 @@
...
+	}()
+	require.NoError(t, setup.startServer())
+
+	defer func() {
+		require.NoError(t, setup.stopServer())
+	}()
+
+	// Verify in-memory data match what we expect - both commitlog and memory write
+	// should be present.
+	expectedSeriesMaps := map[xtime.UnixNano]generate.SeriesBlock{
```
robskillington: 👍

integration/setup.go
```diff
@@ -546,9 +557,23 @@ func (ts *testSetup) close() {
 	}
 }
 
+func (ts *testSetup) mustSetTickMinimumInterval(tickMinInterval time.Duration) {
+	runtimeMgr := ts.storageOpts.RuntimeOptionsManager()
+	existingOptions := runtimeMgr.Get()
+	newOptions := existingOptions.SetTickMinimumInterval(100 * time.Millisecond)
```
robskillington: Hm, should this be `.SetTickMinimumInterval(tickMinInterval)` instead?  Don't see `tickMinInterval` used anywhere else otherwise.

storage/block/block.go
```diff
@@ -239,11 +240,19 @@ func (b *dbBlock) IsCachedBlock() bool {
...
 
-func (b *dbBlock) Merge(other DatabaseBlock) {
+func (b *dbBlock) Merge(other DatabaseBlock) error {
 	b.Lock()
+	if b.wasRetrievedFromDisk || other.WasRetrievedFromDisk() {
+		// We use Merge to lazily merge blocks that eventually need to be flushed to disk
+		// If we try to perform a merge on blocks that were retrieved from disk then we've
+		// violated an invariant and probably have a bug that is causing data loss.
+		b.Unlock()
+		return errTriedToMergeBlockFromDisk
```
robskillington: 👍

integration/setup.go
richardartoul: oh yeah good call, I just always pass 100 * time.Millisecond so it worked but thats not what it should do. Fixed.

storage/flush.go
```diff
@@ -100,8 +103,14 @@ func (m *flushManager) Flush(curr time.Time) error {
 	m.setState(flushManagerFlushInProgress)
 	for _, ns := range namespaces {
 		// Flush first because we will only snapshot if there are no outstanding flushes
-		flushTimes := m.namespaceFlushTimes(ns, curr)
-		multiErr = multiErr.Add(m.flushNamespaceWithTimes(ns, flushTimes, flush))
+		flushTimes := m.namespaceFlushTimes(ns, tickStart)
+		shardBootstrapTimes, ok := dbBootstrapStates.namespaceBootstrapStates[ns.ID().String()]
```
robskillington: Hm, does `dbBootstrapStates.namespaceBootstrapStates` eventually get populated correctly?  Also should probably generate a map here so don't have to keep stringifying.  Although since it's not called that often maybe its ok.

storage/namespace.go
```diff
@@ -747,6 +748,22 @@ func (n *dbNamespace) Flush(
...
+		// before the start of the tick that preceded this flush, meaning it can be reliably
+		// used to determine if all of the bootstrapped blocks have been merged / drained (ticked)
+		// and are ready to be flushed.
+		shardWasBootstrappedBeforeTick, ok := shardBootstrapStates[shard.ID()]
+		if !ok {
+			// We probably don't own this shard anymore.
+			continue
+		}
+
+		if !shardWasBootstrappedBeforeTick {
```
robskillington: nit: Perhaps combine the two if conditions together?  e.g. `if !ok || !shardWasBootstrappedBeforeTick { continue }`?

storage/series/buffer.go
```diff
@@ -157,17 +161,23 @@ func bucketResetStart(now time.Time, b *dbBuffer, idx int, start time.Time) int
...
 		}
 	}
-	return min, max
+
+	var err error
+	if min.IsZero() || max.IsZero() {
+		// Should never happen
+		err = errNoAvailableBuckets
+	}
+	return min, max, err
```
robskillington: nit: A little bit simpler would be:
```go
if min.IsZero() || max.IsZero() {
  return time.Time{}, time.Time{}, errNoAvailableBuckets
}
return min, max, nil
```

storage/series/buffer_test.go
```diff
@@ -225,6 +225,67 @@ func TestBufferDrain(t *testing.T) {
 	assertValuesEqual(t, data[4:], results, opts)
 }
 
+func TestBufferMinMax(t *testing.T) {
```
robskillington: This looks good, do you want to perhaps add a very brief one to verify bootstrapping a block that's already drained returns an error?

storage/series/series.go
```diff
@@ -401,26 +401,33 @@ func (s *dbSeries) bufferDrained(newBlock block.DatabaseBlock) {
...
+	err := s.mergeBlockWithLock(newBlock)
+	if err != nil {
+		iOpts.Logger().WithFields(
+			xlog.NewField("id", s.id.String()),
+			xlog.NewField("blockStart", newBlock.StartTime()),
+			xlog.NewField("err", err.Error()),
+		).Errorf("error trying to drain series buffer")
+		scope.Counter("buffer-drain-error").Inc(1)
+	}
+	iOpts.MetricsScope().Counter("buffer-drained").Inc(1) // Debug
```
robskillington: Do you wan to remove this (just debug)?

richardartoul: I actually think this is a useful metric, or do you mean the literally the debug comment?

richardartoul: I was just trying to say this metric is mostly useful for debugging

storage/series/series.go
```diff
@@ -401,26 +401,33 @@ func (s *dbSeries) bufferDrained(newBlock block.DatabaseBlock) {
...
+	iOpts := s.opts.InstrumentOptions()
+	scope := iOpts.MetricsScope().SubScope("series-buffer-drain")
+	err := s.mergeBlockWithLock(newBlock)
+	if err != nil {
+		iOpts.Logger().WithFields(
+			xlog.NewField("id", s.id.String()),
+			xlog.NewField("blockStart", newBlock.StartTime()),
+			xlog.NewField("err", err.Error()),
+		).Errorf("error trying to drain series buffer")
+		scope.Counter("buffer-drain-error").Inc(1)
```
robskillington: Might be best to cache the `logger xlog.Logger` on the series struct so you can just `s.log.Errorf(...)` and also create a `type seriesMetrics { bufferDrainError tally.Counter }` to make it consistent with other types of metric structs across the code base.

storage/series/series.go
```diff
@@ -443,41 +450,56 @@ func (s *dbSeries) Bootstrap(blocks block.DatabaseSeriesBlocks) error {
...
 		}
+		numBlocksMerged++
 	}
 
-	s.blocks = blocks
-	s.bs = bootstrapped
+	// Debug
+	scope := s.opts.InstrumentOptions().MetricsScope().SubScope("series-bootstrap")
+	scope.Counter("blocks-to-buffer").Inc(numBlocksMovedToBuffer)
+	scope.Counter("blocks-merged").Inc(numBlocksMerged)
```
robskillington: Do you want to keep these debug statements?

storage/series/series.go
```diff
@@ -443,41 +450,56 @@ func (s *dbSeries) Bootstrap(blocks block.DatabaseSeriesBlocks) error {
 	if s.bs == bootstrapping {
 		return errSeriesIsBootstrapping
 	}
+	if bootstrappedBlocks == nil {
+		s.bs = bootstrapped
```
robskillington: Hm, maybe below it's not worth setting `s.bs = bootstrapping` below?  What's going to happen if we never can bootstrap the series?  We at the beginning we should just: 
```go
s.Lock()
defer func() { 
  s.Unlock()
  s.bs = bootstrapped
}()

if s.bs != bootstrapNotStarted {
  return errSeriesAlreadyBootstrapped
}
```

That way it's just a binary not bootstrapped / bootstrapped.  (we can removing the `bootstrapping` type, it's local to this package I believe?).

i.e. even if we hit an error it's probably better to set ourselves as bootstrapped (partially) and let the caller deal with the error as they see fit.

storage/series/series.go
```diff
@@ -443,41 +450,56 @@ func (s *dbSeries) Bootstrap(blocks block.DatabaseSeriesBlocks) error {
...
-		// already drained
-		for _, existingBlock := range existingBlocks.AllBlocks() {
-			s.mergeBlockWithLock(blocks, existingBlock)
+		// If we're unable to put the blocks in an active series buffer bucket, then store them
+		// in the series block, merging with any existing blocks if necessary. There could be an
+		// existing block in the situation that currentTime > blockStart.Add(blockSize).Add(bufferPast),
+		// in which case the series buffer buckets may have been drained and rotated into a block, but
+		// still exist in memory because a flush hasn't occurred yet.
+		err := s.mergeBlockWithLock(block)
+		if err != nil {
```
robskillington: Hm why is this error returned but error returned from buffer bootstrap goes into a multiErr?  Perhaps we should do just one or the other?

storage/series/series.go
```diff
@@ -706,3 +735,18 @@ func (s *dbSeries) Reset(
...
+// of series.Flush() to the caller.
+type FlushOutcome int
+
+const (
+	// FlushErr is just a default value that can be returned when we're also returning an error
+	FlushErr FlushOutcome = iota
+	// BlockDoesNotExist indicates that the series did not have a block for the specified flush blockStart.
+	BlockDoesNotExist
+	// StreamDoesNotExist indicates that the block did not have an underlying stream.
+	StreamDoesNotExist
```
robskillington: This is probably an error btw, maybe we should just treat it as such.

storage/shard.go
```diff
@@ -1699,12 +1699,19 @@ func (s *dbShard) Flush(
 
 	var multiErr xerrors.MultiError
 	tmpCtx := context.NewContext()
+
+	var (
+		// Debug
+		numBlockDoesNotExist  = 0
+		numStreamDoesNotExist = 0
```
robskillington: Do you still need these post-debug?

storage/shard.go
```diff
@@ -1714,9 +1721,24 @@ func (s *dbShard) Flush(
...
+
 		return true
 	})
 
+	// Debug
+	s.logger.WithFields(
+		xlog.NewField("shard", s.ID()),
+		xlog.NewField("numBlockDoesNotExist", numBlockDoesNotExist),
+		xlog.NewField("numStreamDoesNotExist", numStreamDoesNotExist),
+	).Debug("shard flush outcome")
```
robskillington: Do we need this still?

storage/tick.go
```diff
@@ -191,6 +191,9 @@ func (mgr *tickManager) Tick(forceType forceType) error {
 			break
 		}
 		mgr.sleepFn(interval)
+		// Check again at the end of each sleep to see if it
+		// has changed.
```
robskillington: Do you want to put a note here about this being useful for tests?

storage/series/series.go
richardartoul: yeah I think they're useful

storage/shard.go
richardartoul: Yeah I think its useful to have it available without writing new builds

storage/shard.go
richardartoul: good to have for debugging and shouldn't really cost us anything

storage/flush.go
richardartoul: @r Not sure what you mean by populated correctly? Oh, do you mean in the case of a namespace change? Yes the dbBootstrapStates is generated in each tick of the mediator, so it would just happen next time around.

Yeah I thought about doing a genny map thing too, but its infrequent enough I thought it wasn't worth it (string() doesn't allocate I thought because string([]byte{}) doesn't allocate twice?

storage/namespace.go
richardartoul: sure

richardartoul: done

storage/series/buffer_test.go
richardartoul: done

storage/series/series.go
richardartoul: I like it, done

storage/series/series.go
richardartoul: done

storage/shard.go
richardartoul: spoke in person, gonna keep em but return as a result

storage/shard.go
richardartoul: spoke in person, gonna keep these in

storage/tick.go
richardartoul: sure

storage/series/series.go
richardartoul: done

integration/bootstrap_helpers.go
```diff
@@ -0,0 +1,99 @@
...
+package integration
+
+import (
+	"github.com/m3db/m3db/storage/bootstrap"
+	"github.com/m3db/m3db/storage/bootstrap/bootstrapper"
+	"github.com/m3db/m3db/storage/bootstrap/result"
+	"github.com/m3db/m3db/storage/namespace"
+)
+
+// nolint: deadcode
```
prateek: unrelated thought: we should turn off `deadcode`. I can't think of a single time `unused` didn't catch the same issues, and it supports tags

storage/shard.go
richardartoul: done, emitting metrics in the shard now

storage/tick.go
richardartoul: done

storage/series/series.go
richardartoul: (moved to shard.go)

storage/series/series.go
richardartoul: Spoke offline, decided not to do this to avoid adding extra fields to series struct (million in memory at a time)

integration/bootstrap_helpers.go
```diff
@@ -0,0 +1,99 @@
...
+	bootstrap.Bootstrapper
+}
+
+type testBootstrapperSourceOptions struct {
+	can       func(bootstrap.Strategy) bool
+	available func(namespace.Metadata, result.ShardTimeRanges) result.ShardTimeRanges
+	read      func(namespace.Metadata, result.ShardTimeRanges, bootstrap.RunOptions) (result.BootstrapResult, error)
+}
+
+type testBootstrapperSource struct {
```
prateek: suggestion(take/leave): add a var decl to make it clear what interface you intend this to implement

`var _ bootstrap.Source = &testBootstrapperSource{}`

storage/series/series.go
richardartoul: Ok removed the debug metric, but kept the error case one in because this is a really bad error case if its happening. That said, that codepath should never fire so allocations will be ok

integration/bootstrap_helpers.go
richardartoul: done

storage/types.go
```diff
@@ -290,7 +290,11 @@ type databaseNamespace interface {
 	) error
 
 	// Flush flushes in-memory data
-	Flush(blockStart time.Time, flush persist.Flush) error
+	Flush(
+		blockStart time.Time,
+		shardoBootstrapStates shardBootstrapStates,
```
prateek: typo: s/shardo/shard/g

richardartoul: done

storage/flush.go
richardartoul: and the map[string] itself is like 2-8 strings (depending on how many namespaces you have)

integration/bootstrap_before_buffer_rotation_no_tick_regression_test.go
```diff
@@ -0,0 +1,207 @@
...
+	// can be rotated, and then signals the test bootstrapper that it can proceed.
+	go func() {
+		// Wait for server to start
+		setup.waitUntilServerIsUp()
+
+		// Set the time such that the (previously) active block is ready to be flushed.
+		now = now.Add(blockSize).Add(ropts.BufferPast()).Add(time.Second)
+		setup.setNowFn(now)
+
+		// Set the tick interval to be so large we can "hang" a tick at the end, preventing
```
prateek: should the ordering of these be reversed?  i.e. first update the tick interval, and then update the `setup.NowFn`?

integration/bootstrap_before_buffer_rotation_no_tick_regression_test.go
```diff
@@ -0,0 +1,207 @@
...
+		signalCh <- struct{}{}
+		signalCh <- struct{}{}
+	}()
+	require.NoError(t, setup.startServer())
+
+	// Now that bootstrapping has completed, re-enable ticking so that flushing can take place
+	setup.mustSetTickMinimumInterval(100 * time.Millisecond)
+
+	// Wait for a flush to complete
+	setup.sleepFor10xTickMinimumInterval()
```
prateek: do you want to add a waitgroup/channel to ensure the goroutine above has finished before doing the assertions below?

integration/bootstrap_before_buffer_rotation_no_tick_regression_test.go
richardartoul: I dont think it matters. Are you worried about a premature flush? It wont happen because the bootstrapper is blocked for the duration of this method

integration/bootstrap_before_buffer_rotation_no_tick_regression_test.go
richardartoul: dont need one. startServer wont return until bootstrap completes, and bootstrap cant complete until that background routine completes

storage/mediator.go
```diff
@@ -199,3 +219,33 @@ func (m *mediator) reportLoop() {
 		}
 	}
 }
+
+func (m *mediator) databaseBootstrapStates() (databaseBootstrapStates, error) {
```
prateek: hm could you migrate this retrieval to the `database/namspace/shard` types. Will reduce allocs (all the extra slices for namespace/shard each during call)

storage/mediator.go
```diff
@@ -199,3 +219,33 @@ func (m *mediator) reportLoop() {
...
+		}
+		states.namespaceBootstrapStates[namespace.ID().String()] = nsState
+	}
+
+	return states, nil
+}
+
+// databaseBootstrapStates storse a snapshot of the bootstrap state for all shards across all
+// namespaces at a given moment in time.
+type databaseBootstrapStates struct {
```
prateek: nit: rename to `databaseBootstrapState`, also could you move the decls to `storage/types.go`

storage/mediator.go
```diff
@@ -199,3 +219,33 @@ func (m *mediator) reportLoop() {
...
+	namespaceBootstrapStates map[string]shardBootstrapStates
+}
+
+type shardBootstrapStates map[uint32]bool
```
prateek: should this be `map[uint32]bootstrapState`? can migrate the `bootstrapState` enum decl to be along side this type. and then `shard.IsBootstrapped` can be changed to `shard.BootstrapState()`

storage/mediator.go
```diff
@@ -131,16 +133,34 @@ func (m *mediator) EnableFileOps() {
...
+//		4) at least one complete tick has occurred since bootstrap completed (can be the same tick
+// 		   that satisfies condition #3)
+//
+// The mediator helps ensure conditions #3 and #4 are met by measuring the tickStart time and the shard bootstrap
+// states *before* the tick, and then propagating those values to downstream components so they can use that
+// information to make decisions about whether to flush / snapshot / run cleanups.
 func (m *mediator) Tick(runType runType, forceType forceType) error {
-	start := m.nowFn()
+	tickStart := m.nowFn()
+	dbBootstrapStates, err := m.databaseBootstrapStates()
```
prateek: nit: could you rename this to `dbBootstrapStateAtTickStart`, makes it much clearer why you can use it the way you're using it.

storage/mediator.go
```diff
@@ -131,16 +133,34 @@ func (m *mediator) EnableFileOps() {
...
 	if err := m.databaseTickManager.Tick(forceType); err != nil {
 		return err
 	}
+
 	// NB(r): Cleanup and/or flush if required to cleanup files and/or
 	// flush blocks to disk. Note this has to run after the tick as
 	// blocks may only have just become available during a tick beginning
 	// from the tick begin marker.
-	m.databaseFileSystemManager.Run(start, runType, forceType)
+	m.databaseFileSystemManager.Run(tickStart, dbBootstrapStates, syncRun, forceType)
```
prateek: same here: `dbBootstrapStateAtTickStart`

storage/types.go
```diff
@@ -482,7 +486,7 @@ type databaseBootstrapManager interface {
 // databaseFlushManager manages flushing in-memory data to persistent storage.
 type databaseFlushManager interface {
 	// Flush flushes in-memory data to persistent storage.
-	Flush(t time.Time) error
+	Flush(tickStart time.Time, dbBootstrapStates databaseBootstrapStates) error
```
prateek: same here: `dbBootstrapStateAtTickStart`

storage/types.go
```diff
@@ -503,7 +507,7 @@ type databaseFileSystemManager interface {
 	Cleanup(t time.Time) error
 
 	// Flush flushes in-memory data to persistent storage.
-	Flush(t time.Time) error
+	Flush(t time.Time, dbBootstrapStates databaseBootstrapStates) error
```
prateek: here too

storage/types.go
```diff
@@ -517,7 +521,12 @@ type databaseFileSystemManager interface {
 
 	// Run attempts to perform all filesystem-related operations,
 	// returning true if those operations are performed, and false otherwise
-	Run(t time.Time, runType runType, forceType forceType) bool
+	Run(
+		t time.Time,
+		dbBootstrapStates databaseBootstrapStates,
```
prateek: one more

storage/tick.go
```diff
@@ -191,6 +191,9 @@ func (mgr *tickManager) Tick(forceType forceType) error {
 			break
 		}
 		mgr.sleepFn(interval)
+		// Check again at the end of each sleep to see if it
```
prateek: +1 for the comment

storage/shard.go
```diff
@@ -1714,9 +1735,20 @@ func (s *dbShard) Flush(
 			return false
 		}
 
+		// Debug
```
prateek: could you remove the `// debug` indicators if they're going to be kept around.

storage/shard.go
```diff
@@ -1585,6 +1593,11 @@ func (s *dbShard) Bootstrap(
 	s.bs = bootstrapping
 	s.Unlock()
 
+	var (
+		// Debug
+		numBlocksMovedToBuffer = int64(0)
```
prateek: nit: could you make a type for this, and add methods to do the various things you're inlining for instrumentation right now. i.e.

```

type dbShardResult struct {
  numBlocksMovedToBuffer int64 
  numBlocksMerged int64 
}

func (s *dbShardResult) update(seriesBootstrapResult) {... }

func (s *dbShard) emitBootstrapResult(dbShardResult) {...}
```

storage/shard.go
```diff
@@ -1699,12 +1718,14 @@ func (s *dbShard) Flush(
 
 	var multiErr xerrors.MultiError
 	tmpCtx := context.NewContext()
+
+	numBlockDoesNotExist := 0 // Debug
```
prateek: same here, could you make a separate type `dbShardFlushOutcome` or some such

storage/mediator.go
richardartoul: I'm not sure this is a good idea. This is the only place in the codebase where these two actions (tick and flush) are in the same spot. If I push these checks further down I'm gonna have to do some incredibly weird contortions to get the measurements before the tick and then pass them to the flushing code.

The allocs are super small. In production, ticks take 3-10 minutes so we're talking about a slice with 2-8 entries (namespaces) and a map with 1024 non-interface types every few minutes.

storage/series/series.go
```diff
@@ -706,3 +729,25 @@ func (s *dbSeries) Reset(
 	s.onRetrieveBlock = onRetrieveBlock
 	s.blockOnEvictedFromWiredList = onEvictedFromWiredList
 }
+
+// FlushOutcome is an enum that provides more context about the outcome
```
prateek: move these decls to `types.go`

storage/series/series.go
```diff
@@ -706,3 +729,25 @@ func (s *dbSeries) Reset(
...
+// of series.Flush() to the caller.
+type FlushOutcome int
+
+const (
+	// FlushErr is just a default value that can be returned when we're also returning an error
+	FlushErr FlushOutcome = iota
```
prateek: nit: could you prefix these to make it clear they're part of the enum, i.e. 

```
type FlushOutcome int 

const (
   FlushOutcomeErr FlushOutcome = iota // ... 
   FlushOutcomeBlockDoesNotExist
   FlushOutcomeFlushedToDisk
) 
```

storage/series/series.go
```diff
@@ -401,26 +401,33 @@ func (s *dbSeries) bufferDrained(newBlock block.DatabaseBlock) {
...
+	err := s.mergeBlockWithLock(newBlock)
+	if err != nil {
+		iOpts.Logger().WithFields(
+			xlog.NewField("id", s.id.String()),
+			xlog.NewField("blockStart", newBlock.StartTime()),
+			xlog.NewField("err", err.Error()),
+		).Errorf("error trying to drain series buffer")
+		// Allocating metric here is ok because this code-path should never
+		// happen anyways.
+		iOpts.MetricsScope().SubScope("series-buffer-drain")
```
prateek: this is only creating a subscope, are you going to emit a metric for it too?

storage/series/series.go
```diff
@@ -433,52 +440,61 @@ func (s *dbSeries) addBlockWithLock(b block.DatabaseBlock) {
...
+		result.NumBlocksMerged++
 	}
 
-	s.blocks = blocks
 	s.bs = bootstrapped
+	return result, multiErr.FinalError()
+}
 
-	return multiErr.FinalError()
+func (s *dbSeries) bootstrapComplete() {
```
prateek: could you inline this decl where it's being used (only one place afaict)

storage/series/series.go
```diff
@@ -433,52 +440,61 @@ func (s *dbSeries) addBlockWithLock(b block.DatabaseBlock) {
...
+	s.Unlock()
+	s.bs = bootstrapped
```
prateek: doesn't the order of these operations need to be reversed?

integration/bootstrap_before_buffer_rotation_no_tick_regression_test.go
prateek: cool, mind putting a note down by the startServer call to capture that

storage/mediator.go
prateek: The retrieval should definitely be done in the `mediator.Tick()`, the nit was about the databaseBootstrapStates retrieval. i.e. you could do this instead - 

``` 
// types.go
type database interface { 
   BootstrapState() (databaseBootstrapState, error)
}

type databaseNamespace interface { 
   BootstrapState() (shardBootstrapStates, error)
}

type databaseShard interface { 
   BootstrapState() bootstrapStates
}

// database.go
func (d *db) BootstrapState() (...) {
	var (result ... , err error)
	d.RLock()
	defer d.RUnlock()
	for _, n := range d.namespaces.Iter() {
		result[n.Value().ID().String()], err = n.Value().BoostrapState()
		if err != nil {
		    return nil, err
		}
	}
}

// and so on
```

it'd avoid some un-necessary allocs you're currently doing: GetOwnedNamespaces()/GetOwnedShards etc

integration/bootstrap_before_buffer_rotation_no_tick_regression_test.go
```diff
@@ -0,0 +1,207 @@
...
+		// Set the tick interval to be so large we can "hang" a tick at the end, preventing
+		// it from completing until we're ready to "resume" it later.
+		runtimeMgr := setup.storageOpts.RuntimeOptionsManager()
+		existingOptions := runtimeMgr.Get()
+		newOptions := existingOptions.SetTickMinimumInterval(2 * time.Hour)
+
+		// Sleep for a multiple of the old tick interval to guarantee that the "hung" tick
+		// that we will trigger after has a "tick start" time equal to the value of "now"
+		// above.
+		time.Sleep(existingOptions.TickMinimumInterval() * 10)
```
prateek: lol good call on not calling `waitTill10x...` :P

integration/bootstrap_before_buffer_rotation_no_tick_regression_test.go
prateek: Yeah fair. Disregard.

storage/series/series.go
```diff
@@ -433,52 +440,61 @@ func (s *dbSeries) addBlockWithLock(b block.DatabaseBlock) {
 // data in memory during bootstrapping. If that becomes a problem, we could
 // bootstrap in batches, e.g., drain and reset the buffer, drain the streams,
 // then repeat, until len(s.pendingBootstrap) is below a given threshold.
-func (s *dbSeries) Bootstrap(blocks block.DatabaseSeriesBlocks) error {
+func (s *dbSeries) Bootstrap(bootstrappedBlocks block.DatabaseSeriesBlocks) (BootstrapResult, error) {
 	s.Lock()
-	defer s.Unlock()
+	defer s.bootstrapComplete()
```
prateek: it'd be easier to follow the flow if you were to inline the func declaration here

prateek: or could alternatively switch it to the `WithLock` thing if you're worried about the extra alloc

```
func (s *dbSeries) Bootstrap(...) (...) {
  s.Lock()
  result, err := s.bootstrapWithLock(...)
  s.bootstrapped = ...
  s.Unlock()
  return result, err
}

func (s *dbSeries) bootstrapWithLock(...) (...) { }

File: storage/series/series.go (hunk 38 above)
Comment (prateek) [reply storage/series/series.go]: unrelated to your change, but while you're here, mind changing `bs` to `bootstrapState` or some such

File: storage/series/series.go [hunk 41]
```diff
@@ -433,52 +440,61 @@ func (s *dbSeries) addBlockWithLock(b block.DatabaseBlock) {
...
 
+	var result BootstrapResult
 	if s.bs == bootstrapped {
-		return nil
+		return result, errSeriesAlreadyBootstrapped
 	}
-	if s.bs == bootstrapping {
-		return errSeriesIsBootstrapping
+
```
prateek: why drop the `== bootstrapping` check?

storage/series/series.go
```diff
@@ -433,52 +440,61 @@ func (s *dbSeries) addBlockWithLock(b block.DatabaseBlock) {
...
 
-		// If we're overwriting the blocks then merge any existing blocks
-		// already drained
-		for _, existingBlock := range existingBlocks.AllBlocks() {
-			s.mergeBlockWithLock(blocks, existingBlock)
+		// If we're unable to put the blocks in an active series buffer bucket, then store them
+		// in the series block, merging with any existing blocks if necessary. There could be an
+		// existing block in the situation that currentTime > blockStart.Add(blockSize).Add(bufferPast),
+		// in which case the series buffer buckets may have been drained and rotated into a block, but
+		// still exist in memory because a flush hasn't occurred yet.
```
prateek: +1 for the explanation, could you add a little bit at the end: 

```
flush hasn't occurred yet (we guarantee this by not allowing flushes until we're bootstrapped).
```

storage/flush.go
```diff
@@ -100,8 +103,14 @@ func (m *flushManager) Flush(curr time.Time) error {
...
+		if !ok {
```
prateek: feels weird to see this return when you have the mutlierr, could you restructure to the following:

```
namespaces, err := m.database.GetOwnedNamespaces()
if err != nil {		
    return err
}		 
  if _, ok := dbBootstrapStates[ns.ID().String()]; ok {
    return fmt.Errorf(...)
  }
}

// current for _, ns := namespaces {
  ... 
}
```

storage/flush.go
```diff
@@ -73,7 +73,10 @@ func newFlushManager(database database, scope tally.Scope) databaseFlushManager
 	}
 }
 
-func (m *flushManager) Flush(curr time.Time) error {
+func (m *flushManager) Flush(
+	tickStart time.Time,
+	dbBootstrapStates databaseBootstrapStates,
```
prateek: same nit about the naming here

storage/namespace.go
```diff
@@ -719,6 +719,7 @@ func (n *dbNamespace) Bootstrap(
 
 func (n *dbNamespace) Flush(
 	blockStart time.Time,
+	shardBootstrapStates shardBootstrapStates,
```
prateek: naming nit

storage/series/buffer.go
```diff
@@ -42,6 +42,7 @@ var (
 	errTooFuture                   = errors.New("datapoint is too far in the future")
 	errTooPast                     = errors.New("datapoint is too far in the past")
 	errMoreThanOneStreamAfterMerge = errors.New("buffer has more than one stream after merge")
+	errNoAvailableBuckets          = errors.New("buffer has no available buckets")
```
prateek: nit: mind adding `[invariant violated]`

storage/block/block.go
```diff
@@ -194,26 +193,28 @@ func (b *dbBlock) Stream(blocker context.Context) (xio.SegmentReader, error) {
...
+		return stream, nil
 	}
 
-	if !fromBlockRetriever {
-		// Register the finalizer for the stream, the block retriever already
-		// registers the stream as a finalizer for the context so we only perform
-		// this if we return a stream directly from this block
-		blocker.RegisterFinalizer(stream)
+	var mergeStream xio.SegmentReader
+	mergeStream, err = b.mergeTarget.Stream(blocker)
```
prateek: nit: why not `mergeStream, err := b.mergeTarget(...)` ?

storage/block/block.go
```diff
@@ -194,26 +193,28 @@ func (b *dbBlock) Stream(blocker context.Context) (xio.SegmentReader, error) {
...
+	if b.mergeTarget == nil {
+		return stream, nil
 	}
 
-	if !fromBlockRetriever {
-		// Register the finalizer for the stream, the block retriever already
-		// registers the stream as a finalizer for the context so we only perform
-		// this if we return a stream directly from this block
-		blocker.RegisterFinalizer(stream)
+	var mergeStream xio.SegmentReader
```
prateek: +1 for the order refactor, the early termination makes the branches much easier to reason about

storage/flush.go
robskillington: It will allocate sadly, as the `[]byte(...)` underlying is mutable but strings are guarenteed immutable so it can't avoid an alloc.

That's fine though, yeah it's not really used much.

storage/series/series.go
robskillington: Yeah that was my bad, I wrote some inline code as a suggestion and wasn't thinking was just spewing it out.  Unlock needs to come last indeed.

storage/series/series.go
robskillington: Yeah inline would be best.  The lambda will likely get inlined tbh, can check with `go build -gcflags "-m" .` in this package if you want to check.

storage/series/series.go
robskillington: This state is getting removed, there's no such thing as bootstrapping, as we have a lock you're either bootstrapped or not bootstrapped.

integration/bootstrap_before_buffer_rotation_no_tick_regression_test.go
richardartoul: done

storage/mediator.go
richardartoul: Makes sense, I Might keep in IsBootstrapped() thought and just have it call BootstrapState()

storage/mediator.go
richardartoul: Looks like I'll still have to allocated the slice for GetOwnedShards for the sake of iteration, can avoid the namespaces one though I think

richardartoul: ok done

storage/mediator.go
richardartoul: done

storage/mediator.go
richardartoul: done

storage/mediator.go
richardartoul: done

storage/types.go
richardartoul: done

storage/types.go
richardartoul: done

storage/types.go
richardartoul: done

storage/shard.go
richardartoul: done

storage/shard.go
richardartoul: done

storage/series/series.go
richardartoul: done

storage/series/series.go
richardartoul: ...yes thank you. Broke this when I was refactoring

storage/block/block.go
richardartoul: fixed

storage/namespace.go
```diff
@@ -1043,3 +1057,12 @@ func (n *dbNamespace) Close() error {
 	}
 	return nil
 }
+
+func (n *dbNamespace) BootstrapState() ShardBootstrapStates {
+	ownedShards := n.GetOwnedShards()
```
prateek: can you rework to:

```
	n.RLock()
	shards := n.shardSet.AllIDs()
	shardStates := make(ShardBootstrapStates, len(shards))
	for _, shard := range shards {
		shardStates[shard.ID()] = shard.BootstrapState()
	}
	n.RUnlock()
	return shardStates
```

storage/block/block.go
```diff
@@ -264,6 +274,52 @@ func (b *dbBlock) ResetRetrievable(
...
+		stream.Reset(ts.NewSegment(b.segment.Head, b.segment.Tail, ts.FinalizeNone))
+		ctx.RegisterFinalizer(stream)
+	}
+
+	return stream, nil
+}
+
+// TODO(rartoul): The existing ctx is still holding a reference to the old segment so that will hang around
+// and waste memory until the block is closed. We could improve this by swapping out the underlying ctx with
+// a new one, allowing us to close the old one and release the old segment, freeing memory.
```
robskillington: Do you want to open an issue (so you can close it in the near future 😄 )?

storage/block/block.go
```diff
@@ -239,11 +241,19 @@ func (b *dbBlock) IsCachedBlock() bool {
...
+		// We use Merge to lazily merge blocks that eventually need to be flushed to disk
+		// If we try to perform a merge on blocks that were retrieved from disk then we've
+		// violated an invariant and probably have a bug that is causing data loss.
+		b.Unlock()
+		return errTriedToMergeBlockFromDisk
+	}
 	b.resetMergeTargetWithLock()
 	b.mergeTarget = other
 	b.Unlock()
+	return nil
```
robskillington: Hm maybe we should double check here that `b.mergeTarget == nil` before setting the merge target too?  Otherwise that would also cause data loss.

robskillington: Actually in that case may we want to do `b.mergeTarget.Merge(other)`?

storage/block/block.go
richardartoul: #580 tada!

storage/series/series.go
```diff
@@ -126,7 +126,10 @@ func (s *dbSeries) Tick() (TickResult, error) {
 	bufferResult := s.buffer.Tick()
 	r.MergedOutOfOrderBlocks = bufferResult.mergedOutOfOrderBlocks
 
-	update := s.updateBlocksWithLock()
+	update, err := s.updateBlocksWithLock()
+	if err != nil {
+		return r, err
```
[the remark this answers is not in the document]
richardartoul: Fixed the missing unlock, dont want to add defer cause its so hot.  #581

storage/shard.go
```diff
@@ -1381,8 +1383,12 @@ func (s *dbShard) FetchBlocksMetadataV2(
 			FetchBlocksMetadataOptions: opts,
 			IncludeCachedBlocks:        true,
 		}
-		result, nextIndexCursor := s.fetchActiveBlocksMetadata(ctx, start, end,
+		result, nextIndexCursor, err := s.fetchActiveBlocksMetadata(ctx, start, end,
 			limit, indexCursor, seriesFetchBlocksMetadataOpts)
+		if err != nil {
+			return result, nil, err
```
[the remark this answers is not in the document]
richardartoul: done

storage/shard.go
```diff
@@ -1431,8 +1437,12 @@ func (s *dbShard) FetchBlocksMetadataV2(
 			FetchBlocksMetadataOptions: opts,
 			IncludeCachedBlocks:        false,
 		}
-		result, nextIndexCursor := s.fetchActiveBlocksMetadata(ctx, start, end,
+		result, nextIndexCursor, err := s.fetchActiveBlocksMetadata(ctx, start, end,
 			limit, indexCursor, seriesFetchBlocksMetadataOpts)
+		if err != nil {
+			return result, nil, err
```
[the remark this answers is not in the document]
richardartoul: done

PR #546, opened 2018-04-26
FetchTagged followups [1/n]

prateek:
Originally I'd wanted to do all the follow-ups to #530 in this PR, but it's already pretty long, going to start chaining again. 

Changes:
- Re-work consistency calculation. Earlier version treated `LeveUnstrictMajority` the same as `LevelOne`. This PR addresses that.
- Fix an issue in the `Exhaustive` calculation: We needed to know if there was data pending being read if we terminated iteration early. Exposed a new parameter in the `forEachID` type to satisfy this.
- Loads of unit test
  - [x] Accumulator generative tests for consistency
  - [x] results index iter unit tests
  - [x] Accumulator <-> results index iter unit tests
  - [x] Accumulator encoding.iter unit tests
  - [x] Accumulator unit tests (data merge path)

NB:
- Subsequent PR for next round of changes: #549

### DIFF
--- a/client/fetch_tagged_results_accumulator_merge_test.go
+++ b/client/fetch_tagged_results_accumulator_merge_test.go
@@ -1,0 +1,581 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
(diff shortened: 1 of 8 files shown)


client/fetch_tagged_results_accumulator_merge_test.go
```diff
@@ -0,0 +1,581 @@
...
+			buffer.WriteString(fmt.Sprintf("tag-err: %v\n", err))
+		}
+	}
+	if err := iter.Err(); err != nil {
+		buffer.WriteString(fmt.Sprintf("err: %v\n", err))
+	}
+	return buffer.String()
+}
+
+type testFetchTaggedWorkflow struct {
```
robskillington: We should definitely write more declarative tests like this.  There's a few, but not many that are as well defined like this. 👍

client/fetch_tagged_results_index_iterator_test.go
```diff
@@ -0,0 +1,95 @@
...
+				require.Equal(t, expNs.String(), obsNs.String())
+				require.Equal(t, expID.String(), obsID.String())
+				require.True(t, newTagIterMatcher(t, expTags).Matches(obsTags))
+			}
+		})
+	}
+}
+
+// TODO(prateek): migrate this to m3x
+func newTagIterMatcher(t *testing.T, iter ident.TagIterator) ident.TagIterMatcher {
```
robskillington: I've stamped that m3x PR btw.

prateek: cheers, landed and migrated this PR to reuse

client/fetch_tagged_results_accumulator_merge_test.go
prateek: For sure. They capture intent much more succinctly, will standup better to code evolution.

PR #548, opened 2018-04-27
Extract topology construction utilities

prateek:
No functionality changes in this PR. It's a refactor to extract some common testing utilities out of a package. 

NB: the coverage drops because this PR is taking code out of a `_test.go` file into a non-test file. in a sub-package Go doesn't calculate coverage for test files so the original code wasn't covered, it's being covered now, and we only do package level coverage checks in the files that use this code, so we loose coverage :(

### DIFF
--- a/topology/testutil/topology.go
+++ b/topology/testutil/topology.go
@@ -1,0 +1,103 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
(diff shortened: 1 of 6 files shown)


topology/testutil/topology.go
```diff
@@ -0,0 +1,103 @@
...
+	"fmt"
+
+	"github.com/m3db/m3cluster/shard"
+	"github.com/m3db/m3db/sharding"
+	"github.com/m3db/m3db/topology"
+)
+
+// MustNewTopologyMap returns a new topology.Map with provided parameters.
+// It's a utility method to make tests easier to write.
+func MustNewTopologyMap(
```
richardartoul: This is mostly personal preference, but since its in a test package you could accept testing.T which would allow you to do require.NoError(t, err.Error()) or t.Fatal or whatever

richardartoul:
LGTM, don't care about the coverage. Its just a vanity metric in this case

topology/testutil/topology.go
prateek: Yeah i thought about doing that but I prefer not taking the dep as much as possible. the `MustPattern` is pretty standard and achieves the same thing.

PR #549, opened 2018-04-27
FetchTagged followups [2/n]

prateek:
More followups to #530, changes:
- [x] Add more test coverage
  - [x] simple e2e integration test for single node
  - [x] simple e2e integration test for multiple nodes
  - [x] unit tests for session<->fetch tagged interactions
  - [x] unit tests for session<->fetchState ref counting interactions
  - [x] update quorum fetch integration with different shard assignments responses 
  - [x] high concurrency/load unit/integration tests

Pending:
- [x] rebase after https://github.com/m3db/m3db/pull/552, https://github.com/m3db/m3x/pull/147 are landed


client/fetch_state_pool.go
```diff
@@ -33,9 +33,21 @@ type fetchStatePoolImpl struct {
 }
 
 func newfetchStatePool(opts pool.ObjectPoolOptions) fetchStatePool {
```
robskillington: nit: This should maybe be `newFetchStatePool` for camel casing purposes (I know the resulting type is not capitalized, still though seems that should affect the camel casing convention perhaps?).

client/fetch_tagged_results_accumulator.go
```diff
@@ -82,6 +82,13 @@ func (accum *fetchTaggedResultAccumulator) Add(
 	host := opts.host
 	response := opts.response
 
+	if host == nil {
+		// should never happen, guarding against incompatible changes to the `client` package.
```
robskillington: 👍

client/fetch_state.go
```diff
@@ -48,7 +48,7 @@ type fetchState struct {
 	sync.Mutex
 	refCounter
 
-	op                   *fetchTaggedOp
+	op                   fetchTaggedOp
```
robskillington: Just out curiousity, what was the impetus to make this an interface (does it need mocking?) - seems `fetchState` knowing the concrete type should be fine no?

robskillington: Ah I see the mocking below.  If we have to mock and there's significant benefits, then we gotta do what we gotta do.

client/fetch_state_pool.go
prateek: for sure, definitely a typo.

client/fetch_state.go
prateek: cheers, managed to use the concrete type for `fetchTaggedOp` too.

PR #555, opened 2018-04-30
Pull latest m3x changes

prateek:
+1 for the strict marshalling default, already catching shady shit.


client/config_test.go
```diff
@@ -38,7 +38,7 @@ func TestConfiguration(t *testing.T) {
 	in := `
 writeConsistencyLevel: majority
 readConsistencyLevel: unstrict_majority
-clusterConnectConsistencyLevel: any
+connectConsistencyLevel: any
```
robskillington: Yay, the new config struct options caught something =]

PR #557, opened 2018-04-30
Add FetchTagged retry-ability

prateek:
Changes:
- Add retries for FetchTagged/FetchTaggedIDs


client/fetch_tagged_attempt.go
```diff
@@ -0,0 +1,113 @@
...
+	return &fetchTaggedAttemptPoolImpl{pool: p, session: session}
+}
+
+func (p *fetchTaggedAttemptPoolImpl) Init() {
+	p.pool.Init(func() interface{} {
+		f := &fetchTaggedAttempt{session: p.session}
+		// NB(prateek): Bind fn once to avoid creating receiver
+		// and function method pointer over and over again
+		f.idsAttemptFn = f.performIDsAttempt
+		f.dataAttemptFn = f.performDataAttempt
```
robskillington: Yay for reducing allocs at `Get()` time =]

client/session.go
```diff
@@ -1071,10 +1079,32 @@ func (s *session) FetchIDs(
...
+func (s *session) fetchTaggedAttempt(
+	ns ident.ID, q index.Query, opts index.QueryOptions,
+) (encoding.SeriesIterators, bool, error) {
+	// NB(prateek): we have to clone the namespace, as we cannot guarantee the lifecycle
+	// of the hostQueues responding is less than the lifecycle of the current method.
+	nsClone := s.pools.id.Clone(ns)
+
 	const fetchData = true
-	req, err := convert.ToRPCFetchTaggedRequest(ns, q, opts, fetchData)
+	// FOLLOWUP(prateek): currently both `index.Query` and the returned request depened on
```
robskillington: nit: `depend`

client/session.go
```diff
@@ -1158,6 +1211,8 @@ func (s *session) fetchTaggedAttemptWithRLock(
 		fetchState = s.pools.fetchState.Get()
 	)
 
+	fetchState.nsID = nsID // transfer ownership of nsID to `fetchState`
```
robskillington: Any reason not to clone here instead?  Might be cleaner, that way you don't have to clone in two places.

client/fetch_tagged_attempt.go
prateek: :D

client/session.go
prateek: Yep, the value returned by `convert.ToRPCFetchTaggedRequest(nsClone, ...)` takes reference to the bytes underlying the `nsID` clone, so the nsID clone has to precede the request creation. 

I'm doing the RPC request creation early (before fetchState/Op creation) to avoid the extra allocs in case of failures.

PR #561, opened 2018-05-01
Add Index knobs to KV Namespace Options

prateek:


storage/namespace/config.go
```diff
@@ -56,10 +46,25 @@ func (m *MapConfiguration) Map() (Map, error) {
...
+// MetadataConfiguration is the configuration for a single namespace
+type MetadataConfiguration struct {
+	ID                string                  `yaml:"id" validate:"nonzero"`
+	BootstrapEnabled  *bool                   `yaml:"bootstrapEnabled"`
+	FlushEnabled      *bool                   `yaml:"flushEnabled"`
+	WritesToCommitLog *bool                   `yaml:"writesToCommitLog"`
+	CleanupEnabled    *bool                   `yaml:"cleanupEnabled"`
+	RepairEnabled     *bool                   `yaml:"repairEnabled"`
+	Retention         retention.Configuration `yaml:"retention" validate:"nonzero"`
+	Index             IndexConfiguration      `yaml:"index" validate:"nonzero"`
```
robskillington: Hm do you need validate nonzero for Index here?  For `IndexConfiguration` if `Enabled` is false and `BlockSizeNanos` which is a totally valid configuration, would that perhaps not validate?

PR #562, opened 2018-05-01
Index file set reader and writer

robskillington:


generated/proto/index/index.proto
```diff
@@ -0,0 +1,38 @@
+syntax = "proto3";
+package index;
+
+message IndexInfo {
+  message SegmentFileInfo {
```
prateek: nit: could you avoid the nested type declarations. would make this a little cleaner to follow

generated/proto/index/index.proto
```diff
@@ -0,0 +1,38 @@
...
+    string fileType = 1;
+  }
+
+  message SegmentInfo {
+    string type = 1;
```
prateek: nit:s/type/segmentType to be extra clear

generated/proto/index/index.proto
```diff
@@ -0,0 +1,38 @@
...
+}
+
+message IndexDigests {
+  message SegmentFileDigest {
+    string fileType = 1;
+    uint32 digest = 2;
+  }
+
+  message SegmentDigest {
+    string type = 1;
```
prateek: here too

generated/proto/namespace/namespace.pb.go
```diff
@@ -1,23 +1,3 @@
-// Copyright (c) 2018 Uber Technologies, Inc.
```
prateek: could add back the licenses for these files and the new `index.pb.go` file too

persist/fs/files.go
```diff
@@ -767,3 +830,20 @@ func snapshotPathFromTimeAndIndex(prefix string, t time.Time, suffix string, ind
 	name := fmt.Sprintf("%s%s%d%s%s%s%d%s", filesetFilePrefix, separator, t.UnixNano(), separator, suffix, separator, index, fileSuffix)
 	return path.Join(prefix, name)
 }
+
+func filesetIndexSegmentFilePathFromTime(
+	prefix string,
+	t time.Time,
+	segmentIndex int,
```
prateek: nit: this is the fileset index, not the segmentIndex, yea?

generated/mocks/generate.go
```diff
@@ -19,7 +19,7 @@
 // THE SOFTWARE.
 
 // mockgen rules for generating mocks for exported interfaces (reflection mode)
-//go:generate sh -c "mockgen -package=fs $PACKAGE/persist/fs FileSetWriter,FileSetReader,FileSetSeeker | mockclean -pkg $PACKAGE/persist/fs -out $GOPATH/src/$PACKAGE/persist/fs/fs_mock.go"
+//go:generate sh -c "mockgen -package=fs $PACKAGE/persist/fs DataFileSetWriter,DataFileSetReader,DataFileSetSeeker,IndexFileSetWriter,IndexFileSetReader,IndexSegmentFileSet,IndexSegmentFile | mockclean -pkg $PACKAGE/persist/fs -out $GOPATH/src/$PACKAGE/persist/fs/fs_mock.go"
```
prateek: lol fk using `go-generate`, even plain bash over this shit would be an improvement.

persist/fs/files.go
```diff
@@ -767,3 +830,20 @@ func snapshotPathFromTimeAndIndex(prefix string, t time.Time, suffix string, ind
...
+	segmentFileType IndexSegmentFileType,
+) string {
+	idx := strconv.Itoa(segmentIndex)
+	suffix := fmt.Sprintf(
```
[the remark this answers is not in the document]
robskillington: A little better, I was basically trying to show the different groupings.  I can just make it %s%s%s%s... though too.

persist/fs/files.go
robskillington: Nah, this is the actual segment number, because you can have multiple segments per index block.  So I need to actually create another one of these functions for the snapshot segment files, right now they just get named without the snapshot index.  I'll add this and add a test for this case too.

persist/fs/index_read.go
```diff
@@ -0,0 +1,383 @@
...
+	r.filePathPrefix = opts.FilePathPrefix()
+	r.hugePagesOpts = mmap.HugeTLBOptions{
+		Enabled:   opts.MmapEnableHugeTLB(),
+		Threshold: opts.MmapHugeTLBThreshold(),
+	}
+	r.logger = opts.InstrumentOptions().Logger()
+}
+
+func (r *indexReader) Open(opts IndexReaderOpenOptions) error {
+	r.reset(r.opts)
```
prateek: is this required?

persist/schema/types.go
```diff
@@ -31,14 +31,14 @@ const MajorVersion = 1
 
 // IndexInfo stores metadata information about block filesets
 type IndexInfo struct {
+	MajorVersion int64
```
prateek: hm i thought the ordering was kept consistent between here and order of fields written out

persist/fs/index_read_write_go-pre-1-10_test.go
```diff
@@ -0,0 +1,27 @@
+// +build !go1.10
```
prateek: can we drop this and only use go1.10 and up?

persist/fs/index_write.go
```diff
@@ -0,0 +1,298 @@
...
+			file.Close()
+		}
+	}()
+
+	if w.err != nil {
+		return w.err
+	}
+
+	segType := string(segmentFileSet.SegmentType())
+	if segType == "" || !fileSubTypeRegex.MatchString(segType) {
```
prateek: nit: could you declare as a helper on `SegmentType()`:

```
func (t SegmentType) String() (string, error) {
  s := string(t)
  if t == "" || !fileSubTypeRegex.MatchString(s) {
    return "" , fmt.Errorf("...")
  }
  return s, nil
}
```

persist/fs/index_write.go
```diff
@@ -0,0 +1,298 @@
...
+		segmentType:  segmentFileSet.SegmentType(),
+		majorVersion: segmentFileSet.MajorVersion(),
+		minorVersion: segmentFileSet.MinorVersion(),
+		metadata:     segmentFileSet.SegmentMetadata(),
+	}
+
+	idx := len(w.segments)
+	for _, file := range segmentFileSet.Files() {
+		segFileType := string(file.SegmentFileType())
+		if segFileType == "" || !fileSubTypeRegex.MatchString(segFileType) {
```
prateek: same here

persist/fs/files.go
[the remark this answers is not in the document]
prateek: sounds good

generated/proto/namespace/namespace.pb.go
prateek: hm don't think so. `make proto-gen` should do the trick

persist/schema/types.go
robskillington: It was, but it seems silly considering it's not necessary and `MajorVersion` really should be first ha.  I can revert if you'd like.

persist/fs/index_read.go
robskillington: If you want to be able to reuse the reader it is yes, we do tend to do that a fair bit.

generated/proto/index/index.proto
robskillington: Sure thing.

generated/proto/index/index.proto
robskillington: Sure thing.

generated/proto/index/index.proto
robskillington: Sure thing.

generated/mocks/generate.go
robskillington: Yeah true

persist/fs/index_read_write_go-pre-1-10_test.go
robskillington: Sounds good.

persist/fs/index_write.go
robskillington: For sure, much cleaner.

persist/fs/index_write.go
robskillington: Agreed.

persist/schema/types.go
prateek: Nah i mean as long as we the order is captured in code w/e. This was a minor nicety but it does make the struct clunky

persist/fs/index_read.go
prateek: Fair enough

PR #567, opened 2018-05-01
Fix peer bootstrapping panic with corrupt files

### LINKED ISSUE #511: Make peer bootstrapper incremental flush handle case where fileset files already exist

richardartoul:
Closes #511


docs/architecture/engine.md
```diff
@@ -144,7 +144,7 @@ Lets imagine a read for a given series that requests the last 6 hours worth of d
 If the current time is 8PM, then the location of the requested blocks might be as follows:
 
 ```
-[2PM - 4PM (Fileset file)] - Sealed and flushed block that isn't cached
+[2PM - 4PM (FileSet file)] - Sealed and flushed block that isn't cached
```
prateek: solid! always great to see doc updates 👍

storage/shard.go
```diff
@@ -1682,13 +1685,6 @@ func (s *dbShard) Flush(
 		return s.markFlushStateSuccessOrError(blockStart, err)
 	}
 
-	// No action is necessary therefore we bail out early and there is no need to close.
```
prateek: mind explicitly setting DeleteIfExists to false in the prepare opts and explain why you choose to do that.

storage/shard_test.go
```diff
@@ -170,36 +170,6 @@ func TestShardFlushDuringBootstrap(t *testing.T) {
 	require.Equal(t, err, errShardNotBootstrappedToFlush)
 }
 
-func TestShardFlushNoPersistFuncNoError(t *testing.T) {
```
prateek: nice, glad to see this go.

persist/fs/files.go
```diff
@@ -428,6 +436,49 @@ func IndexSnapshotFiles(filePathPrefix string, namespace ident.ID) (SnapshotFile
...
+		return FileSetFile{}, false, err
+	}
+
+	matched.sortByTimeAscending()
+	for i, fileset := range matched {
+		if fileset.ID.BlockStart.Equal(blockStart) {
+			nextIdx := i + 1
+			if nextIdx < len(matched) && matched[nextIdx].ID.BlockStart.Equal(blockStart) {
+				// Should never happen
+				return FileSetFile{}, false, fmt.Errorf(
```
prateek: pull in the invariant utils from https://github.com/m3db/m3x/commit/ef130fab89ea319ba5361d6c73569396bf3c3996 here

storage/shard.go
```diff
@@ -1467,7 +1467,10 @@ func (s *dbShard) FetchBlocksMetadataV2(
 	// Work backwards while in requested range and not before retention
 	for !blockStart.Before(start) &&
 		!blockStart.Before(retention.FlushTimeStart(ropts, s.nowFn())) {
-		exists := s.namespaceReaderMgr.filesetExistsAt(s.shard, blockStart)
+		exists, err := s.namespaceReaderMgr.filesetExistsAt(s.shard, blockStart)
+		if err != nil {
```
prateek: hm, would it be better to log failure and carry on iteration instead of early terminating here? 

Concern is a single corrupt block on a single host would stop it's peers from being able to bootstrap from it.

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -358,6 +358,18 @@ func (s *peersSource) incrementalFlush(
 			NamespaceMetadata: nsMetadata,
 			Shard:             shard,
 			BlockStart:        start,
+			// If we've peer bootstrapped this shard/block combination AND the fileset
+			// already exists on disk, then that means either:
+			//      1) The Filesystem bootstrapper was unable to bootstrap the fileset
+			//         files on disk, even though they have a checkpoint file. This
+			//         could either be the result of data corruption, or a
+			//         backwards-incompatible change to the file-format.
+			// 		2) The Filesystem bootstrapper is not enabled, in which case it makes
```
prateek: nit: alignment is fked

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -358,6 +358,18 @@ func (s *peersSource) incrementalFlush(
 			NamespaceMetadata: nsMetadata,
 			Shard:             shard,
 			BlockStart:        start,
+			// If we've peer bootstrapped this shard/block combination AND the fileset
```
prateek: Consider operator error where someone changes the kv config for bootstrappers to run peers before fs. That would violate the assumptions here. 

Could you include guards against allowing such configs to be used by a m3dbnode

persist/prepare_options_matcher.go
```diff
@@ -28,12 +28,14 @@ import (
...
 
 // PrepareOptionsMatcher satisfies the gomock.Matcher interface for PrepareOptions
+// nolint: maligned
 type PrepareOptionsMatcher struct {
-	NsMetadata   namespace.Metadata
-	Shard        uint32
-	BlockStart   time.Time
-	SnapshotTime time.Time
-	FileSetType  FileSetType
+	NsMetadata     namespace.Metadata
```
prateek: https://github.com/google/go-cmp looks ideal for doing such comparisons at runtime. Mind adopting/ opening a followup issue to track.

persist/fs/files.go
richardartoul: I pulled in the constant, not sure we want to make an instrument / logger a requirement for using this function

storage/shard.go
richardartoul: @robskillington Would the peer bootstrapping logic on the client side handle this situation properly if we just skipped over an entire block?

robskillington: We definitely shouldn't skip over an entire block, otherwise we magically did not hold up our deal of applying bootstrap level consistency (because one replica just looks like it was successful retrieving all metadata, even though it skipped some).

There's already a ton of places in this flow where we can return an error.  I'd prefer that we stick with that, in that case basically one replica will just look bad.  Depending on your bootstrap level consistency that may be fine (you might still get majority from the other nodes).  If not, an operator needs to make the call to reduce bootstrap consistency, or remove the problematic files themselves.

persist/fs/files.go
robskillington: It'd be better to emit at a higher layer.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: Added a guard for this in bootstrapper construction in `services/m3dbnode/config/bootstrapper.go`.

storage/shard.go
robskillington: Added this.

prateek: Cheers

storage/bootstrap/types.go
```diff
@@ -82,6 +82,10 @@ type BootstrapperProvider interface {
 	// String returns the name of the bootstrapper.
 	String() string
 
+	// Next returns the next bootstrapper provider that's used to fulfill
+	// ranges when this bootstrapper provider cannot fulfill time ranges.
+	Next() BootstrapperProvider
```
prateek: nit: mind changing to `(BootstrapperProvider, bool)` for the return type? nil check on the return value feels icky.

storage/bootstrap/bootstrapper/peers/source.go
prateek: Cheers

services/m3dbnode/config/bootstrap.go
```diff
@@ -148,3 +151,54 @@ func (bsc BootstrapConfiguration) New(
 
 	return bootstrap.NewProcessProvider(bs, rsOpts), nil
 }
+
+// ValidateBootstrappersOrder will validate that a list of bootstrappers specified
+// is in validorder.
+func ValidateBootstrappersOrder(names []string) error {
```
prateek: +1

services/m3dbnode/config/bootstrap.go
```diff
@@ -90,11 +90,14 @@ func (bsc BootstrapConfiguration) New(
 	opts storage.Options,
 	adminClient client.AdminClient,
 ) (bootstrap.ProcessProvider, error) {
+	if err := ValidateBootstrappersOrder(bsc.Bootstrappers); err != nil {
```
prateek: mind adding some logging in the lambda where this is used - `server.go:kvWatchBootstrappers` - just to indicate what bootstrapper configuration we're continuing to use when this fails

storage/shard.go
robskillington: There's logging on the client side for every error case thankfully, is that sufficient?

PR #570, opened 2018-05-02
Extract ident<->doc utils to 'convert' pkg

prateek:


storage/index.go
```diff
@@ -265,10 +265,15 @@ func (i *nsIndex) Query(
 		return index.QueryResults{}, err
 	}
 
+	idxIter := index.NewIterator(i.nsID, iter, i.opts)
+
+	ctx.RegisterFinalizer(resource.FinalizerFn(func() {
+		idxIter.Finalize()
+		exec.Close()
+	}))
```
robskillington: nit: To avoid a lambda alloc (you'll still need to do a method bind alloc but it's slightly cheaper I think) you could probably:
```go
ctx.RegisterFinalizer(idxIter)
ctx.RegisterFinalizer(resource.FinalizerFn(exec.Close))
```

prateek: sure

prateek: Actually can't do it. `exec.Close()` returns an error so i need the lambda alloc unfortunately.

storage/index/iterator.go
```diff
@@ -100,7 +93,13 @@ func (i *idsIter) Err() error {
 	return i.err
 }
 
-func (i *idsIter) release() {
+func (i *idsIter) Finalize() {
+	i.releaseCurrent()
+	*i = idsIterZeroed
+	// TODO(prateek): implement idsIter pooling
```
robskillington: As discussed, if we can pool the posting list returned that's probably most important first.

storage/index.go
robskillington: Discussed, exec.Close returns an error.  We can just keep an alloc for now.

PR #574, opened 2018-05-03
Initial bootstrap index metadata interfaces

robskillington:
Remaining:

- [x] Tests that cover the new base boostrapper coordinating the index bootstrapper call chain


storage/types.go
```diff
@@ -688,11 +688,11 @@ type Options interface {
...
 	RepairOptions() repair.Options
 
-	// SetBootstrapProcess sets the bootstrap process for the database
-	SetBootstrapProcess(value bootstrap.Process) Options
+	// SetBootstrapProcessProvider sets the bootstrap process for the database
+	SetBootstrapProcessProvider(value bootstrap.ProcessProvider) Options
 
-	// BootstrapProcess returns the bootstrap process for the database
-	BootstrapProcess() bootstrap.Process
+	// BootstrapProcessProvider returns the bootstrap process for the database
```
prateek: nit: returns the bootstrap process provider ...

storage/types.go
```diff
@@ -688,11 +688,11 @@ type Options interface {
 	// RepairOptions returns the repair options
 	RepairOptions() repair.Options
 
-	// SetBootstrapProcess sets the bootstrap process for the database
-	SetBootstrapProcess(value bootstrap.Process) Options
+	// SetBootstrapProcessProvider sets the bootstrap process for the database
```
prateek: here too

storage/bootstrap/result/types.go
```diff
@@ -46,6 +47,38 @@ type BootstrapResult interface {
...
+	// Add adds an index block result.
+	Add(block IndexBlock, unfulfilled xtime.Ranges)
+}
+
+// IndexResults is a set of index blocks indexed by block start.
+type IndexResults map[xtime.UnixNano]IndexBlock
+
+// IndexBlock contains the bootstrap data structures for an index block.
+type IndexBlock struct {
+	blockStart time.Time
```
prateek: these fields should probably be un-exported

storage/bootstrap/result/types.go
```diff
@@ -46,6 +47,38 @@ type BootstrapResult interface {
...
+	segments   []segment.Segment
+	notIndexed map[string]SeriesNotIndexed // TODO(r): create map specifically for this case
```
prateek: ended up creating this map for https://github.com/m3db/m3db/pull/578, can probably share it

storage/bootstrap/result/types.go
prateek: ah I see you've defined accesors for the fields in the file above, any reason you prefer that instead of just exporting the fields?

storage/bootstrap/result/result_index.go
```diff
@@ -0,0 +1,131 @@
...
+
+// Merged returns a new merged index block, it mutates the current
+// not indexed series map however to avoid allocating a new map.
+func (b IndexBlock) Merged(
+	segments []segment.Segment,
+	notIndexed map[string]SeriesNotIndexed,
+) IndexBlock {
+	r := b
+	r.segments = append(r.segments, segments...)
+	for _, entry := range notIndexed {
```
prateek: probably want to finalize IDs/tags when you see dupes

storage/bootstrap/types.go
```diff
@@ -62,6 +75,23 @@ type RunOptions interface {
 	// Incremental returns whether this bootstrap should be an incremental
 	// that saves intermediate results to durable storage or not.
 	Incremental() bool
+
+	// SetIndexingEnabled sets whether indexing is enabled which determines
```
prateek: I think you can drop these methods from the `RunOptions()`. `namespaceMetadata.Options().IndexOptions().Enabled()` has this flag now.

storage/bootstrap/types.go
```diff
@@ -26,22 +26,35 @@ import (
...
-	// Bootstrapper returns the current bootstrapper to use when running the process.
-	Bootstrapper() Bootstrapper
-
 	// Run runs the bootstrap process, returning the bootstrap result and any error encountered.
 	Run(
 		ns namespace.Metadata,
 		shards []uint32,
 		targetRanges []TargetRange,
-	) (result.BootstrapResult, error)
+	) (
```
prateek: think we need to take separate targetRanges for data v index. The time calculations depend on the block size so we could end up different values - 
screen shot 2018-05-04 at 01 30 55

storage/bootstrap/result/types.go
robskillington: Yeah mainly to enforce immutability, I'm not opposed to not exporting them though.

storage/types.go
robskillington: Sounds good.

storage/types.go
robskillington: Sounds good.

storage/bootstrap/result/types.go
robskillington: Can do, I'll open up a PR once that lands.

storage/bootstrap/result/result_index.go
robskillington: Good catch, yup thanks.

storage/bootstrap/process.go
```diff
@@ -100,18 +112,18 @@ func (b *bootstrapProcess) Run(
...
-			return nil, err
+			return nil, nil, err
 		}
 
 		b.log.WithFields(logFields...).Infof("bootstrapping shards for range completed successfully")
 		bootstrapResult = result.MergedBootstrapResult(bootstrapResult, res)
 	}
 
-	return bootstrapResult, nil
+	return bootstrapResult, nil, nil
```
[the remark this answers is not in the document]
robskillington: Yeah fair.

storage/bootstrap/types.go
robskillington: Ok sure thing, sounds good.

storage/bootstrap/types.go
robskillington: I was thinking we would be looking at the namespace index options block size and be using the cache from the previous target range to stitch the data together.  But true it'd be more clean to just issue two different target ranges.  

I might even move this logic to the bootstrap process itself, it seems a little odd for it to be here.

storage/bootstrap/result/types.go
prateek: Sure, either way is fine with me.

PR #576, opened 2018-05-03
Add filesystem inspection and use it in commitlog bootstrapper

richardartoul:
Adds the concept of a FilesystemInspection which can be used by the commitlog reader to avoid reading commitlog files that were generated since the process started.

Also fixes case where the commitlog bootstrapper could deadlock because it was not returning a resource to a cyclical pool (if a datapoint didn't pass the readSeries predicate we would not return the byte slice to the buffer channel).

### DIFF
--- a/persist/fs/commitlog/reader.go
+++ b/persist/fs/commitlog/reader.go
@@ -300,12 +300,10 @@ func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan<- readResponse
 	)
 
 	for arg := range inBuf {
-		readResponse := readResponse{}
+		response := readResponse{}
 		// If there is a pre-existing error, just pipe it through
 		if arg.err != nil {
-			readResponse.resultErr = arg.err
-			arg.bufPool <- arg.bytes
-			outBuf <- readResponse
+			r.handleDecoderLoopIterationEnd(arg, outBuf, response, arg.err)
 			continue
 		}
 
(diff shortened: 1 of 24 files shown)


persist/fs/commitlog/reader.go
```diff
@@ -375,6 +369,14 @@ func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan<- readResponse
 	r.metadata.Unlock()
 }
 
+func (r *reader) handleDecoderLoopIterationEnd(arg decoderArg, outBuf chan<- readResponse, response readResponse, err error) {
+	arg.bufPool <- arg.bytes
+	if outBuf != nil {
```
robskillington: Perhaps a little more "idiomatic" would be:

persist/fs/commitlog/reader.go
```diff
@@ -338,29 +332,29 @@ func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan<- readResponse
...
-			arg.bufPool <- arg.bytes
-			outBuf <- readResponse
+			r.handleDecoderLoopIterationEnd(arg, outBuf, response, errCommitLogReaderMissingMetadata)
 			continue
 		}
 
 		if !metadata.passedPredicate {
+			// Pass nil for outBuf because we don't want to send a readResponse along since this
+			// was just a series that the caller didn't want us to read.
+			r.handleDecoderLoopIterationEnd(arg, nil, response, nil)
```
robskillington: Is it worth putting response in here then?  Maybe just arg is required?

PR #578, opened 2018-05-04
Re-work index/ server side response type

prateek:


storage/index.go
```diff
@@ -265,16 +264,57 @@ func (i *nsIndex) Query(
...
+			results.Finalize()
+			exec.Close()
+			iter.Close()
+		}
+	)
+	results.Reset(i.nsID)
+
+	for iter.Next() {
+		// FOLLOWUP(prateek): as part of the runtime options wiring, also set a server side max limit which
+		// super-seeds the user requested limit. Paranoia can be a good thing.
```
robskillington: If we have pagination though it may be ok to do large scrolls for instance.

storage/index.go
```diff
@@ -265,16 +264,57 @@ func (i *nsIndex) Query(
 		return index.QueryResults{}, err
 	}
 
-	idxIter := index.NewIterator(i.nsID, iter, i.opts)
+	var (
+		brokeEarly = false
+		numAdded   = 0
+		results    = i.opts.ResultsPool().Get()
+		cleanupFn  = func() {
```
robskillington: nit: Just a thought, you can do what I do sometimes and:
```
success := false
defer func() {
  if !success {
    // .. do cleanup
  }
}()

// do work

success = true
return ...
```

storage/index/results.go
```diff
@@ -0,0 +1,144 @@
...
+	return r.resultsMap
+}
+
+func (r *results) Reset(nsID ident.ID) {
+	r.nsID = nsID
+
+	// reset all values from map first
+	for _, entry := range r.resultsMap.Iter() {
+		tags := entry.Value()
+		tags.Finalize()
```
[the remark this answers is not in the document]
prateek: talked offline: tags is a slice so it can't hold a ref to the pool to allow finalize to put it back. re-worked the code to do the iteration/ident finalize inline, makes the flow a little easier to reason about.

storage/index.go
prateek: Nice, yeah that's a lot better.

storage/index.go
prateek: You'd want different server side limits for fetch v scroll but fair.

storage/index.go
prateek: done

storage/index/results.go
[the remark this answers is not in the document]
prateek: done

storage/index.go
```diff
@@ -265,16 +264,53 @@ func (i *nsIndex) Query(
...
-	}))
+	if err := iter.Err(); err != nil {
+		return index.QueryResults{}, err
+	}
+
+	if err := iter.Close(); err != nil {
+		return index.QueryResults{}, err
+	}
+
+	if err := exec.Close(); err != nil {
```
robskillington: Hm is it safe to double close `exec` and `iter` this (with the defer at the top?)

Maybe better to:

prateek: It's safe right now but yeah that's cleaner. will change

storage/index/options.go
```diff
@@ -33,11 +33,15 @@ import (
 const (
 	// defaultIndexInsertMode sets the default indexing mode to synchronous.
 	defaultIndexInsertMode = InsertSync
+
+	defaultTagArrayPoolInitCapacity = 5
+	defaultTagArrayPoolMaxCapacity  = 20
```
robskillington: nit: Perhaps 8 and 32 just to stay in base 2?

storage/index/results.go
```diff
@@ -0,0 +1,163 @@
...
+	// the tsID's bytes.
+	r.resultsMap.Set(tsID, tags)
+	r.size++
+
+	added = true
+	return added, r.size, nil
+}
+
+func (r *results) tags(fields doc.Fields) []ident.Tag {
+	tags := r.tagArrayPool.Get()[:0]
```
robskillington: Hm should we perhaps make the pool do the `[:0]` subslice either on the get or the put?

prateek: Sure, the pool already does the `[:0]` on Put, will remove here.

storage/index/options.go
prateek: sure

storage/index.go
prateek: talked offline, ended up with `safeCloser` struct idiom

storage/index/results.go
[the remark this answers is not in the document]
prateek: ended up doing this instead, because it reads a lot cleaner:
```
	// reset all values from map first
	for _, entry := range r.resultsMap.Iter() {
		tags := entry.Value()
		tags.Finalize()
		r.tagArrayPool.Put(tags)
	}
```

`tags.Finalize()` releases all backing ident.IDs, and the `Put()` sets the slice elems to empty value and re-sizes slice to [:0] before putting back into the pool.

storage/index/results.go
prateek: done

storage/index/options.go
prateek: done

storage/index.go
prateek: done

storage/index/results.go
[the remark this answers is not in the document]
prateek: done

PR #579, opened 2018-05-04
Passing start and end time for blocks through to session

arnikola:
Squashed changes

### DIFF
--- a/network/server/tchannelthrift/node/service.go
+++ b/network/server/tchannelthrift/node/service.go
@@ -51,8 +51,6 @@ import (
 
 const (
 	checkedBytesPoolSize        = 65536
-	tagEncoderPoolSize          = 65536
-	tagDecoderPoolSize          = 65536
 	segmentArrayPoolSize        = 65536
 	initSegmentArrayPoolLength  = 4
 	maxSegmentArrayPooledLength = 32
@@ -153,23 +151,6 @@ func NewService(db storage.Database, opts tchannelthrift.Options) rpc.TChanNode 
 	wrapperPool := xpool.NewCheckedBytesWrapperPool(wrapperPoolOpts)
 	wrapperPool.Init()
 
-	tagEncoderPoolOpts := pool.NewObjectPoolOptions().
-		SetSize(tagEncoderPoolSize).
-		SetInstrumentOptions(iopts.SetMetricsScope(
-			scope.SubScope("tag-encoder-pool")))
-	tagEncoderPool := serialize.NewTagEncoderPool(
-		serialize.NewTagEncoderOptions(), tagEncoderPoolOpts)
-	tagEncoderPool.Init()
-
-	tagDecoderPoolOpts := pool.NewObjectPoolOptions().
-		SetSize(tagDecoderPoolSize).
-		SetInstrumentOptions(iopts.SetMetricsScope(
-			scope.SubScope("tag-decoder-pool")))
-	tagDecoderPool := serialize.NewTagDecoderPool(
-		serialize.NewTagDecoderOptions().
-			SetCheckedBytesWrapperPool(wrapperPool), tagDecoderPoolOpts)
-	tagDecoderPool.Init()
-
 	segmentPool := newSegmentsArrayPool(segmentsArrayPoolOpts{
 		Capacity:    initSegmentArrayPoolLength,
 		MaxCapacity: maxSegmentArrayPooledLength,
@@ -188,8 +169,8 @@ func NewService(db storage.Database, opts tchannelthrift.Options) rpc.TChanNode 
 		metrics: newServiceMetrics(scope, iopts.MetricsSamplingRate()),
 		pools: pools{
 			checkedBytesWrapper:  wrapperPool,
-			tagEncoder:           tagEncoderPool,
-			tagDecoder:           tagDecoderPool,
+			tagEncoder:           opts.TagEncoderPool(),
+			tagDecoder:           opts.TagDecoderPool(),
 			id:                   db.Options().IdentifierPool(),
 			segmentsArray:        segmentPool,
 			blockMetadata:        opts.BlockMetadataPool(),
--- a/network/server/tchannelthrift/options.go
+++ b/network/server/tchannelthrift/options.go
@@ -20,6 +20,11 @@
 
 package tchannelthrift
 
+import (
+	"github.com/m3db/m3db/serialize"
+	"github.com/m3db/m3x/pool"
+)
+
 type options struct {
 	blockMetadataPool        BlockMetadataPool
 	blockMetadataV2Pool      BlockMetadataV2Pool
@@ -27,10 +32,22 @@ type options struct {
 	blockMetadataV2SlicePool BlockMetadataV2SlicePool
 	blocksMetadataPool       BlocksMetadataPool
 	blocksMetadataSlicePool  BlocksMetadataSlicePool
+	tagEncoderPool           serialize.TagEncoderPool
+	tagDecoderPool           serialize.TagDecoderPool
 }
 
 // NewOptions creates new options
 func NewOptions() Options {
+	tagEncoderPool := serialize.NewTagEncoderPool(
+		serialize.NewTagEncoderOptions(), pool.NewObjectPoolOptions(),
+	)
+	tagDecoderPool := serialize.NewTagDecoderPool(
+		serialize.NewTagDecoderOptions(), pool.NewObjectPoolOptions(),
+	)
+
+	tagEncoderPool.Init()
+	tagDecoderPool.Init()
+
 	return &options{
 		blockMetadataPool:        NewBlockMetadataPool(nil),
 		blockMetadataV2Pool:      NewBlockMetadataV2Pool(nil),
@@ -38,6 +55,8 @@ func NewOptions() Options {
 		blockMetadataV2SlicePool: NewBlockMetadataV2SlicePool(nil, 0),
 		blocksMetadataPool:       NewBlocksMetadataPool(nil),
 		blocksMetadataSlicePool:  NewBlocksMetadataSlicePool(nil, 0),
+		tagEncoderPool:           tagEncoderPool,
+		tagDecoderPool:           tagDecoderPool,
 	}
 }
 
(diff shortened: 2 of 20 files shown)


client/reader_slice_of_slices_iterator.go
```diff
@@ -53,35 +56,43 @@ func (it *readerSliceOfSlicesIterator) Next() bool {
 	it.idx++
 
 	// Extend segment readers if not enough available
-	currLen := it.CurrentLen()
-	if len(it.segmentReaders) < currLen {
-		diff := currLen - len(it.segmentReaders)
+	currLen, start, end := it.Current()
```
prateek: nit: comment above this line needs to be updated

client/session.go
```diff
@@ -3131,23 +3130,32 @@ func (b *baseBlocksResult) newDatabaseBlock(block *rpc.Block) (block.DatabaseBlo
...
-			segmentReader.Reset(b.segmentForBlock(segments.Unmerged[i]))
-			readers[i] = segmentReader
+			endTime := timeConvert(segment.EndTime)
+			blockReader := xio.BlockReader{
+				SegmentReader: segmentReader,
+				Start:         timeConvert(segment.StartTime),
+				End:           endTime,
+			}
+			blockReader.Reset(b.segmentForBlock(segment))
+			readers[i] = blockReader
```
robskillington: Hm did this code need to change?  It seems that we'll need to heap allocate the `xio.BlockReader` here.  You're putting each element into `readers` which is a `[]xio.SegmentReader` yeah?  Maybe we can just directly put the segment readers in that array rather than wrapping them in a `BlockReader` unnecessarily?  They're going to get merged right below on line 3154 anyhow and will be discarded.

client/session.go
```diff
@@ -3313,7 +3321,9 @@ func (r *bulkBlocksResult) addBlockFromPeer(id ident.ID, peer topology.Host, blo
 		result.Close()
 
 		result = r.blockOpts.DatabaseBlockPool().Get()
-		result.Reset(start, encoder.Discard())
+		blockSize := resultReader.End.Sub(start)
```
robskillington: Do you need to override `blockSize` here?  Using the variable passed to the method might be more accurate than this calculation here, since its passed directly from the retention options rather than by calculating it.

encoding/types.go
```diff
@@ -134,11 +134,14 @@ type ReaderIterator interface {
 type MultiReaderIterator interface {
 	Iterator
 
-	// Reset resets the iterator to read from a slice of readers.
-	Reset(readers []io.Reader)
+	// Reset resets the iterator to read from a slice of blocks.
+	Reset(blocks []xio.SegmentReader, start, end time.Time)
```
robskillington: Can probably just call this `readers` still?  and `"slice of readers"`?

persist/fs/retriever.go
```diff
@@ -348,15 +348,16 @@ func (r *blockRetriever) Stream(
 	ctx context.Context,
 	shard uint32,
 	id ident.ID,
-	startTime time.Time,
+	startTime, endTime time.Time,
```
robskillington: Instead of requiring `endTime` being passed here (which callers could get wrong, or just not have access to potentially) can you calculate from the namespace retention options?

i.e. some variation of
```go
req.end = startTime.Add(r.nsMetadata.Options().RetentionOptions().BlockSize())
```

We could cache the block size on the retriever too if we wanted to avoid the method calls, *shrug*.

storage/block/block.go
```diff
@@ -261,76 +271,85 @@ func (b *dbBlock) Merge(other DatabaseBlock) error {
...
-		stream = b.opts.SegmentReaderPool().Get()
-		stream.Reset(ts.NewSegment(b.segment.Head, b.segment.Tail, ts.FinalizeNone))
-		ctx.RegisterFinalizer(stream)
+		blockReader = xio.BlockReader{
+			SegmentReader: b.opts.SegmentReaderPool().Get(),
+			Start:         start,
+			End:           end,
+		}
+		blockReader.Reset(ts.NewSegment(b.segment.Head, b.segment.Tail, ts.FinalizeNone))
+		ctx.RegisterFinalizer(blockReader)
```
robskillington: Hm, maybe just finalize the segment reader here?  Otherwise the `blockReader` will need to be coerced into an interface type here too (and will end up being heap allocated).  

However all we really need to do is finalize the segment reader and not the encapsulating block reader itself.

storage/block/retriever_manager.go
```diff
@@ -102,10 +102,11 @@ func (r *shardBlockRetriever) Stream(
 	ctx context.Context,
 	id ident.ID,
 	blockStart time.Time,
+	blockEnd time.Time,
```
robskillington: As per other comment, hopefully we can avoid callers needing to pass blockEnd by resolving the block size on the retriever end.

storage/series/buffer.go
```diff
@@ -626,26 +625,33 @@ func (b *dbBufferBucket) write(
...
+		end := start.Add(b.opts.RetentionOptions().BlockSize())
 		if s := b.encoders[i].encoder.Stream(); s != nil {
-			ctx.RegisterFinalizer(s)
-			streams = append(streams, s)
+			br := xio.BlockReader{
+				SegmentReader: s,
+				Start:         start,
+				End:           end,
+			}
+			ctx.RegisterFinalizer(br)
```
robskillington: Can we register finalizer on the segment reader here too to avoid alloc on the interface?

storage/series/buffer.go
```diff
@@ -739,13 +745,20 @@ func (b *dbBufferBucket) merge() (mergeResult, error) {
...
+		if s := b.encoders[i].encoder.Stream(); s != nil {
 			merges++
-			readers = append(readers, stream)
-			streams = append(streams, stream)
+			br := xio.BlockReader{
+				SegmentReader: s,
+				Start:         start,
+				End:           end,
+			}
+			readers = append(readers, br)
```
robskillington: Here too, can you `append(readers, s)`?  Will be the same result as you are merging just below.  

In fact probably don't need the `BlockReader` at all here.

storage/series/buffer.go
```diff
@@ -764,14 +777,14 @@ func (b *dbBufferBucket) merge() (mergeResult, error) {
 	}()
 
 	for i := range b.bootstrapped {
-		stream, err := b.bootstrapped[i].Stream(ctx)
-		if err == nil && stream != nil {
+		block, err := b.bootstrapped[i].Stream(ctx)
+		if err == nil && !block.IsEmpty() {
 			merges++
-			readers = append(readers, stream)
+			readers = append(readers, block)
```
robskillington: Maybe `append(readers, block.SegmentReader())` to avoid alloc of the block reader struct here too?

x/xio/block_reader.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+
+	"github.com/m3db/m3db/ts"
+)
+
+// CloneBlock returns a clone of the block with the underlying data reset
+func (b BlockReader) CloneBlock() (BlockReader, error) {
+	sr, err := b.SegmentReader.Clone()
+	if err != nil {
+		return EmptyBlockReader, nil
```
robskillington: Probably want to return the error here yeah? `return EmptyBlockReader, err`?

x/xio/block_reader.go
```diff
@@ -0,0 +1,57 @@
...
+	return b.Start.Equal(timeZero) && b.End.Equal(timeZero) && b.SegmentReader == nil
+}
+
+// IsNotEmpty returns false for the empty block
+func (b BlockReader) IsNotEmpty() bool {
+	return !b.IsEmpty()
+}
+
+// ResetWindowed resets the underlying reader window, as well as start and end times for the block
+func (b *BlockReader) ResetWindowed(segment ts.Segment, start, end time.Time) {
```
robskillington: nit: Is it worth making this `start time.Time, blockSize time.Duration`?  Just so it's not necessary to worry about someone providing an end time that's before the start time here?

client/session.go
```diff
@@ -3291,20 +3295,22 @@ func (r *bulkBlocksResult) addBlockFromPeer(id ident.ID, peer topology.Host, blo
...
 			return err
 		}
-		if resultReader == nil {
+		if resultReader.IsEmpty() {
 			return nil
 		}
 
-		readers := []io.Reader{currReader, resultReader}
-		encoder, err := r.mergeReaders(start, readers)
+		readers := []xio.SegmentReader{currReader, resultReader}
```
robskillington: To avoid interface allocation here too can we do `currReader.SegmentReader()` and `resultReader.SegmentReader()`?  Avoid all the allocs =0

integration/coverage_imports.go
```diff
@@ -0,0 +1,78 @@
+// +build integration
```
robskillington: This is generated, you may want to delete it (we should probably add to the .gitignore) cc @prateek

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -398,6 +398,7 @@ func (s *fileSystemSource) loadShardReadersDataIntoShardResult(
 		var (
 			timeRange = r.Range()
 			start     = timeRange.Start
+			blockSize = timeRange.End.Sub(start)
```
robskillington: Can we use the namespace options here too, i.e. `ns.Options().RetentionOptions().BlockSize()`?

x/xio/block_reader.go
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: nit: should be 2018

x/xio/block_reader_test.go
```diff
@@ -0,0 +1,206 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
robskillington: nit: should be 2018

integration/coverage_imports.go
arnikola: Yeah sorry, I think it might have gotten pulled in after `git add .` after the merge

storage/block/block.go
```diff
@@ -98,6 +100,7 @@ func NewDatabaseBlock(
 // NewRetrievableDatabaseBlock creates a new retrievable DatabaseBlock instance.
 func NewRetrievableDatabaseBlock(
 	start time.Time,
+	blockSize time.Duration,
```
robskillington: Should we also make `NewDatabaseBlock` take `blockSize` too?

PR #583, opened 2018-05-04
Add support for tags in the commitlog

richardartoul:


persist/fs/commitlog/writer.go
```diff
@@ -27,12 +27,15 @@ import (
 	"os"
 	"time"
 
+	"github.com/m3db/m3x/ident"
```
robskillington: nit: Import order grouping.

persist/fs/commitlog/reader.go
```diff
@@ -405,12 +409,37 @@ func (r *reader) decodeAndHandleMetadata(
...
+	var (
+		tags        ident.Tags
+		tagBytesLen = len(decoded.Tags)
+	)
+	if tagBytesLen != 0 {
+		tagsBytes := r.checkedBytesPool.Get(len(decoded.Tags))
+		tagsBytes.IncRef()
+		tagsBytes.AppendAll(decoded.Tags)
+		tagsBytes.DecRef()
+		tagDecoder.Reset(tagsBytes)
```
robskillington: You're going to clone the bytes below, perhaps best to keep a `checked.Bytes` around and pass it to the `tagDecoder`?  Will avoid the pooling/locking of tag bytes.

e.g. 
```go
// some cached checked.Bytes somewhere
tagBytes := checked.NewBytes(nil, nil)


tagBytes.Reset(decoded.Tags)
tagDecoder.Reset(tagBytes)

// .. clone the tags from the decoder
```

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -348,7 +350,7 @@ func (s *commitLogSource) mergeShard(
 			if shardResult == nil {
 				shardResult = result.NewShardResult(len(unmergedShard.encodersBySeries), s.opts.ResultOptions())
 			}
-			shardResult.AddSeries(unmergedBlocks.id, nil, seriesBlocks) // FOLLOWUP(prateek): include tags in commit log reader
+			shardResult.AddSeries(unmergedBlocks.id, unmergedBlocks.tags, seriesBlocks)
```
robskillington: Nice, perhaps add a similar bootstrapper integration test for the commit log similar to the fs one?

persist/fs/commitlog/reader.go
richardartoul: oh duh, thanks

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Discussed in person, gonna hold off and add one for the index in the next P.R

persist/schema/types.go
```diff
@@ -92,4 +92,5 @@ type LogMetadata struct {
 	ID        []byte
 	Namespace []byte
 	Shard     uint32
+	Tags      []byte
```
robskillington: nit: Should we call this `EncodedTags`?  I don't feel strongly, however it is what it's called in the `IndexEntry`.

PR #584, opened 2018-05-05
Create storage/index.Block type

prateek:
Changes:
- Creates a new type: index.Block - a collection of m3ninx Segments - for a defined boundary of time.
- Migrates the namespaceIndex to be backed by a single Block (for now) 
- Just a heads up, this PR has about ~200 lines of code changes, rest of it is refactor/tests. 

Note: More changes involving the block rotation follow in a subsequent PR. This is broken up for easier review and to be compatible with the version on master being used for testing in m3coordinator.


storage/index/block.go
```diff
@@ -0,0 +1,377 @@
...
+		// NB: releasing all references to inserts
+		for _, insert := range inserts {
+			insert.OnIndexSeries.OnIndexFinalize()
+		}
+		return WriteBatchResult{
+			NumError: int64(len(inserts)),
+		}, writeBatchErrorInvalidState(b.state)
+	}
+
+	// FOLLOWUP(prateek): docs array pooling
```
robskillington: Hm, since writes to the index are batched you could take a write lock in this method and cache the slice on the index block itself?  Can avoid pooling that way.

storage/index/block.go
```diff
@@ -0,0 +1,377 @@
...
+	}, partialErr
+}
+
+func (b *block) executorWithRLock() (search.Executor, error) {
+	var (
+		readers = make([]m3ninxindex.Reader, 0, 1+len(b.bootstrappedSegments))
+		success = false
+	)
+
+	// cleanup in case shit breaks
```
robskillington: super nit: Obviously not a big deal, but maybe `s/shit/all the things/`?

storage/index/block.go
prateek: Definitely. That was the plan but didn't do it cause this code used to live in the index until this PR. Will make the change now.

storage/index/block.go
prateek: sure thing

storage/index/block.go
prateek: done

storage/index/block.go
prateek: done

PR #590, opened 2018-05-06
Add read/write of tags to fileset to restore tags, add fs index bootstrapping

robskillington:


storage/index/allocator.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"github.com/m3db/m3ninx/index/segment"
+	"github.com/m3db/m3ninx/index/segment/mem"
+	"github.com/m3db/m3x/instrument"
+)
+
+// NewDefaultMutableSegmentAllocator returns a default mutable segment
+// allocator.
+func NewDefaultMutableSegmentAllocator(
+	iopts instrument.Options,
```
prateek: nit: change this to Options (instead of instrument.Options) 

(we override some other defaults in the segment options in the default ctor too)

persist/fs/clone/cloner.go
```diff
@@ -62,16 +82,29 @@ func (c *cloner) Clone(src FileSetID, dest FileSetID, destBlocksize time.Duratio
...
-		id, data, checksum, err := reader.Read()
+		id, tagsIter, data, checksum, err := reader.Read()
 		if err != nil {
 			if err == io.EOF {
 				break
 			}
 			return fmt.Errorf("unexpected error while reading data: %v", err)
 		}
 
+		var tags ident.Tags
```
prateek: this snippet seems to be repeated multiple times in the diff. mind pulling into a common place? `m3x/ident/testutil` maybe?

persist/fs/msgpack/roundtrip_test.go
```diff
@@ -169,6 +170,53 @@ func TestIndexEntryRoundtrip(t *testing.T) {
...
+func TestIndexEntryRoundTripForwardsCompatibilityV2(t *testing.T) {
+	var (
+		enc = newEncoder(newEncoderOptions{encodeLegacyV1IndexEntry: false})
+		dec = testDecoder(t, nil)
+	)
+
+	// Set the default values on the fields that did not exist in V1
+	// and then restore them at the end of the test - This is required
+	// because the old decoder won't read the new fields
+	oldEncodedTags := testIndexEntry.EncodedTags
```
richardartoul: I think I did the same thing in my test, but I was dealing with integer types that would get encoded anyways. Do you want to encode data with tag data included, and then strip them from the test object for the sake of comparison? That way we know the decoder can actually skip over tags data if it is present

persist/fs/options.go
```diff
@@ -112,6 +128,12 @@ func (o *options) Validate() error {
 			"invalid index bloom filter false positive percent, must be >= 0 and <= 1: instead %f",
 			o.indexBloomFilterFalsePositivePercent)
 	}
+	if o.tagEncoderPool == nil {
```
richardartoul: Is this guarding against someone doing SetTagEncoderPool(nil)? I don't think we do this check everywhere but seems reasonable

persist/fs/clone/cloner_test.go
```diff
@@ -87,14 +107,32 @@ func TestCloner(t *testing.T) {
...
 		}
+
+		require.NoError(t, e1)
+		require.NoError(t, e2)
+
 		b1.IncRef()
 		b2.IncRef()
 		require.Equal(t, t1.String(), t2.String())
+		require.Equal(t, a1.Remaining(), a2.Remaining())
+		numTags, numTagsMatched := a1.Remaining(), 0
```
prateek: there's a matcher for tag iters, so can make this a little shorter if you'd like:

```
 ident.NewTagIterMatcher(ident.NewTagSliceIterator(a1)).Matches(ident.NewTagSliceIterator(a2)))
```

persist/fs/read.go
```diff
@@ -380,21 +388,31 @@ func (r *reader) ReadBloomFilter() (*ManagedConcurrentBloomFilter, error) {
 	)
 }
```
richardartoul: This method name is a little confusing, it feels like it implies it returns the bytes for an entry when really what it does is just convert arbitrary []byte to (unowned) checked.Bytes. maybe just call it toCheckedBytes or cloneBytes or something

persist/fs/clone/options.go
```diff
@@ -1,3 +1,23 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
prateek: note to future self: https://github.com/m3db/ci-scripts/issues/12 ++

persist/fs/msgpack/roundtrip_test.go
robskillington: Sure thing.

persist/fs/clone/cloner.go
robskillington: Sure thing.

storage/index/allocator.go
robskillington: Sure thing.

persist/fs/write.go
```diff
@@ -408,12 +428,27 @@ func (w *writer) writeIndexFileContents(
...
+		var encodedTags []byte
+		if tags := w.indexEntries[i].tags; tags != nil {
+			tagsIter.Reset(tags)
+			tagsEncoder.Reset()
+			if err := tagsEncoder.Encode(tagsIter); err != nil {
+				return err
+			}
+			data, ok := tagsEncoder.Data()
+			if !ok {
+				return errWriterEncodeTagsDataNotAccessible
```
richardartoul: When does this happen

services/m3dbnode/config/bootstrap.go
```diff
@@ -89,15 +90,15 @@ func (bsc BootstrapConfiguration) New(
...
-		bs  bootstrap.BootstrapperProvider
-		err error
+		bs     bootstrap.BootstrapperProvider
+		err    error
+		iopts  = opts.InstrumentOptions()
+		rsopts = result.NewOptions().
+			SetInstrumentOptions(opts.InstrumentOptions()).
+			SetDatabaseBlockOptions(opts.DatabaseBlockOptions()).
+			SetSeriesCachePolicy(opts.SeriesCachePolicy()).
+			SetIndexMutableSegmentAllocator(index.NewDefaultMutableSegmentAllocator(iopts))
```
richardartoul: This wasn't here before right? Why is it being added now

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -45,10 +47,30 @@ type newDataFileSetReaderFn func(
 	opts fs.Options,
 ) (fs.DataFileSetReader, error)
 
+type runType int
+
+const (
```
richardartoul: Any reason to move these out into a more shareable location? I'm thinking no because its really only the FS bootstrapper that uses almost the exact same code-path for bootstrappign data vs index right?

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -283,9 +346,22 @@ func (s *fileSystemSource) handleErrorsAndUnfulfilled(
 	}
 }
 
+func (s *fileSystemSource) tagsFromTagsIter(
+	iter ident.TagIterator,
+) (ident.Tags, error) {
+	tags := make(ident.Tags, 0, iter.Remaining())
```
richardartoul: this code repeats all the time, do you think its worth adding a new method to the iter to just do it?

richardartoul: Also we're not pooling these arrays yeah? I think I need ot do something similar for the commitlog bootstrappign stuff and wasn't sure. I thought @prateek mentioned he created a pool for it

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -310,51 +389,123 @@ func (s *fileSystemSource) loadShardReadersDataIntoShardResult(
...
+			resultLock.Lock()
+			results := runResult.index.IndexResults()
+			indexBlockSegment, err = results.GetOrAddSegment(start,
+				ns.Options().IndexOptions(), bopts)
+			resultLock.Unlock()
+			if err != nil {
+				s.log.Errorf("unable to create index segment: %v", err)
+				hasError = true
+			}
+		default:
```
richardartoul: +1

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -454,76 +621,81 @@ func (s *fileSystemSource) loadShardReadersDataIntoShardResult(
...
-		nsID              = md.ID()
-		blockRetrieverMgr = s.opts.DatabaseBlockRetrieverManager()
-		blockRetriever    block.DatabaseBlockRetriever
-	)
-	if blockRetrieverMgr != nil {
-		s.log.WithFields(
-			xlog.NewField("namespace", nsID.String()),
-		).Infof("filesystem bootstrapper resolving block retriever")
+	if run == bootstrapDataRunType {
+		// NB(r): We can only need to cache shard indices and possibly shortcut
```
richardartoul: This looks like one of the few big code blocks that would be easy to factor out into a method since it doesn't modify any other function state except the blockRetriever (which it could return).

Also, its ok that the blockRetriever is always nil in the index case yeah?

richardartoul: This comment is a little confusing as well.

Maybe something like: "We only need to cache shard indices and marks blocks as fulfilled when bootstrapping data, because the data can be retrieved lazily from disk during reads. On the other hand, if we're bootstrapping the index then we need to rebuild it from scratch by reading all the IDs/tags"

storage/series/series.go
```diff
@@ -603,25 +604,19 @@ func (s *dbSeries) Flush(
 	blockStart time.Time,
 	persistFn persist.Fn,
 ) (FlushOutcome, error) {
-	// NB(r): Do not use defer here as we need to make sure the
-	// call to sr.Segment() which may fetch data from disk is not
-	// blocking the series lock.
 	s.RLock()
+	defer s.RUnlock()
```
richardartoul: Delete cache all metadata :D

persist/fs/seek.go
```diff
@@ -101,9 +101,10 @@ type seeker struct {
 // IndexEntry is an entry from the index file which can be passed to
 // SeekUsingIndexEntry to seek to the data for that entry
 type IndexEntry struct {
-	Size     uint32
-	Checksum uint32
-	Offset   int64
+	Size        uint32
+	Checksum    uint32
+	Offset      int64
+	EncodedTags []byte
```
prateek: mind updating the docs/diagrams with this change too

persist/fs/retriever.go
```diff
@@ -523,6 +537,10 @@ func (req *retrieveRequest) onCallerOrRetrieverDone() {
 	}
 	req.id.Finalize()
 	req.id = nil
+	if req.tags != nil {
+		req.tags.Close()
+		req.tags = nil
```
prateek: should this be `req.tags = ident.EmptyTagIterator`

services/m3dbnode/config/bootstrap.go
prateek: need the ability to create segments for index bootstrap results (they include segments)

storage/series/series.go
prateek: is the disk fetch (earlier comment) no longer a concern?

richardartoul:
Also do you think its worth adding a simple integration test to make sure this all works end to end? Seems like you could just copy-paste one of our existing FS integration tests and add an additional assertion on the tags

storage/series/series.go
richardartoul: I think he removed it because it only applied to cache all metadata and we're planning on deleting that

persist/fs/clone/cloner_test.go
robskillington: Ah nice, will update - gracias.

persist/fs/clone/options.go
robskillington: Ha, aye, all good.

persist/fs/read.go
robskillington: I'll call it `entryClonedBytes(...)` ta.

persist/fs/write.go
robskillington: Only if you haven't actually encoded anything yet, it should never occur in practice.

services/m3dbnode/config/bootstrap.go
robskillington: As prateek mentions.

storage/bootstrap/bootstrapper/fs/source.go
robskillington: Although true, this is internal to this bootstrapper - for instance for now the peer bootstrapper will have a fundamentally different path when bootstrapping the index as it needs to call FetchMetadata on the client session, hence it won't need any concept of run type.

persist/fs/retriever.go
robskillington: Sure thing (we initialize this with ident.EmptyTagIterator when we reset for reuse but we can do that here too).

storage/series/series.go
robskillington: Yeah we're going to delete the cache all metadata strategy, also it's probably not all that bad to just wait if that's the case (it never should though though as this is a flush, the data should never have come from disk if we're flushing this).

robskillington:
Sure thing will add an integration test.

storage/series/series.go
prateek: fair enough

storage/bootstrap/result/options.go
```diff
@@ -32,21 +33,25 @@ const (
...
 func NewOptions() Options {
+	iopts := instrument.NewOptions()
 	return &options{
 		clockOpts:         clock.NewOptions(),
-		instrumentOpts:    instrument.NewOptions(),
+		instrumentOpts:    iopts,
 		blockOpts:         block.NewOptions(),
 		newBlocksLen:      defaultNewBlocksLen,
 		seriesCachePolicy: series.DefaultCachePolicy,
+		mutableSegmentAllocator: index.NewDefaultMutableSegmentAllocator(index.NewOptions().
```
prateek: can probably skip overriding `InstrumentOptions()`? they're both going to end up with the default `instrument.NewOptions()`

persist/fs/msgpack/roundtrip_test.go
```diff
@@ -169,6 +170,55 @@ func TestIndexEntryRoundtrip(t *testing.T) {
...
+	defer func() {
+		testIndexEntry.EncodedTags = oldEncodedTags
+	}()
+
+	enc.EncodeIndexEntry(testIndexEntry)
+	dec.Reset(NewDecoderStream(enc.Bytes()))
+	res, err := dec.DecodeIndexEntry()
+	require.NoError(t, err)
+
+	expected
```
prateek: is this broken?

persist/fs/msgpack/roundtrip_test.go
```diff
@@ -169,6 +170,55 @@ func TestIndexEntryRoundtrip(t *testing.T) {
 	require.Equal(t, testIndexEntry, res)
 }
 
+// Make sure the new decoding code can handle the old file format
```
prateek: +1 for these tests

persist/fs/read.go
```diff
@@ -347,27 +349,33 @@ func (r *reader) Read() (ident.ID, checked.Bytes, uint32, error) {
...
 
 	entry := r.indexEntriesByOffsetAsc[r.metadataRead]
+	id = r.entryClonedID(entry.ID)
+	tags = r.entryClonedEncodedTagsTagIter(entry.EncodedTags)
+	length = int(entry.Size)
+	checksum = uint32(entry.Checksum)
 
 	r.metadataRead++
-	return r.entryID(entry.ID), int(entry.Size), uint32(entry.Checksum), nil
+	return
```
prateek: mind using `return id, tags, length, checksum, nil` instead? 
(IMO named returns hurt usability if used without this, cause it adds more things to consider when looking at exit points)

persist/fs/read_write_test.go
```diff
@@ -121,12 +138,25 @@ func readTestData(t *testing.T, r DataFileSetReader, shard uint32, timestamp tim
...
+				id, tags, data, checksum, err := r.Read()
 				require.NoError(t, err)
 
 				data.IncRef()
 
+				// Assert id
 				assert.Equal(t, entries[i].id, id.String())
+
+				// Assert tags
+				if entries[i].tags != nil {
```
prateek: can simplify this to:
```
ident.NewTagIterMatcher(ident.NewTagSliceIterator(entries[i].Tags())).Matches(tags)
```

persist/fs/read_write_test.go
```diff
@@ -137,13 +167,29 @@ func readTestData(t *testing.T, r DataFileSetReader, shard uint32, timestamp tim
...
 			case readTestTypeMetadata:
-				id, length, checksum, err := r.ReadMetadata()
+				id, tags, length, checksum, err := r.ReadMetadata()
 				require.NoError(t, err)
 
+				// Assert id
 				assert.True(t, id.Equal(id))
+
+				// Assert tags
+				if entries[i].tags != nil {
```
prateek: here too

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -454,76 +621,61 @@ func (s *fileSystemSource) loadShardReadersDataIntoShardResult(
...
-	switch s.opts.ResultOptions().SeriesCachePolicy() {
-	case series.CacheAll:
-		// No checks necessary
-	case series.CacheAllMetadata:
-		// Need to check block retriever available
-		if blockRetriever == nil {
-			return nil, fmt.Errorf(
-				"missing block retriever when using series cache metadata for namespace: %s",
-				nsID.String())
+	if run == bootstrapDataRunType {
```
prateek: do you want to retain the logging statements from the original? either way is fine, just calling out the difference

prateek: actually nvm, see that you're doing this in `resolveBlockRetrieverAndCacheDataShardIndices`

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -310,51 +389,123 @@ func (s *fileSystemSource) loadShardReadersDataIntoShardResult(
...
+
+				resultLock.RLock()
+				exists, err := indexBlockSegment.ContainsID(idBytes)
+				resultLock.RUnlock()
+
+				if err == nil && !exists {
+					resultLock.Lock()
+					exists, err = indexBlockSegment.ContainsID(idBytes)
+					if err == nil && !exists {
+						d := doc.Document{
```
prateek: mind putting down a `// TODO(prateek): update once https://github.com/m3db/m3db/pull/596 lands` here

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -283,9 +346,22 @@ func (s *fileSystemSource) handleErrorsAndUnfulfilled(
...
+	tags := make(ident.Tags, 0, iter.Remaining())
+	for iter.Next() {
+		curr := iter.Current()
+		tags = append(tags, s.idPool.CloneTag(curr))
+	}
+	return tags, iter.Err()
+}
+
 func (s *fileSystemSource) loadShardReadersDataIntoShardResult(
+	ns namespace.Metadata,
```
prateek: can I convince you to break this function into smaller chunks? it was huge to start, now with both code-paths (index/data) in the same function, it's really un-wieldly

storage/bootstrap/bootstrapper/fs/source.go
prateek: Yeah I don't think pooling is necessary cause these ids/tags would be retained in memory by the `series` (once we transfer ownership correctly). Longer term I think it'd be worthwhile to see if bulk allocating the slices helps perf but we can come back to that.

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -310,51 +389,123 @@ func (s *fileSystemSource) loadShardReadersDataIntoShardResult(
...
+
+				idBytes := id.Bytes()
+
+				resultLock.RLock()
+				exists, err := indexBlockSegment.ContainsID(idBytes)
+				resultLock.RUnlock()
+
+				if err == nil && !exists {
+					resultLock.Lock()
+					exists, err = indexBlockSegment.ContainsID(idBytes)
```
prateek: could you move this to `IndexBootstrapResult.AddDoc(ident.ID, ident.Tags) error` instead? it'd make this code a lot cleaner

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -310,51 +389,123 @@ func (s *fileSystemSource) loadShardReadersDataIntoShardResult(
...
-			resultLock.RLock()
-			results := bootstrapResult.ShardResults()
-			var exists bool
-			shardResult, exists = results[shard]
-			resultLock.RUnlock()
-
-			if !exists {
-				resultLock.Lock()
+		var (
+			timeRange = r.Range()
```
prateek: hm where does the fs block start -> index block start happen?

storage/bootstrap/result/options.go
robskillington: Sure thing.

persist/fs/msgpack/roundtrip_test.go
robskillington: Hm yes, must've been a bad last push - my bad.

persist/fs/read.go
robskillington: I don't feel strongly, sure

persist/fs/read_write_test.go
robskillington: Sure thing.

persist/fs/read_write_test.go
robskillington: Sure thing.

storage/bootstrap/bootstrapper/fs/options.go
```diff
@@ -38,14 +40,21 @@ type options struct {
 	fsOpts                fs.Options
 	numProcessors         int
 	blockRetrieverManager block.DatabaseBlockRetrieverManager
+	identifierPool        ident.Pool
 }
 
 // NewOptions creates new bootstrap options
 func NewOptions() Options {
+	bytesPool := pool.NewCheckedBytesPool(nil, nil, func(s []pool.Bucket) pool.BytesPool {
```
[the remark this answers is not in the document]
robskillington: =]

storage/bootstrap/bootstrapper/fs/source.go
robskillington: That's now landed, I'll just pull this in.

storage/bootstrap/bootstrapper/fs/source.go
robskillington: Sure I can break some of it.

storage/bootstrap/bootstrapper/fs/source.go
robskillington: In `GetOrAddSegment(...)`

docs/architecture/storage.md
```diff
@@ -12,43 +12,43 @@ A fileset has the following files:
...
-┌─────────────────────┐  ┌─────────────────────┐     │     Index File      │   
-│      Info File      │  │   Summaries File    │     │   (sorted by ID)    │   
-├─────────────────────┤  │   (sorted by ID)    │     ├─────────────────────┤   
-│- Block Start        │  ├─────────────────────┤  ┌─>│- Idx                │   
-│- Block Size         │  │- Idx                │  │  │- ID                 │   
-│- Entries (Num)      │  │- ID                 │  │  │- Size               │   
-│- Major Version      │  │- Index Entry Offset ├──┘  │- Checksum           │   
+                                                     ┌─────────────────────┐
+┌─────────────────────┐  ┌─────────────────────┐     │     Index File      │
+│      Info File      │  │   Summaries File    │     │   (sorted by ID)    │
```
richardartoul: Since you're updating this, can you add the new fields I added recently:

SnapshotTime int64
FileType     persist.FileSetType

integration/fs_bootstrap_index_test.go
```diff
@@ -0,0 +1,249 @@
+// +build integration
+
+// Copyright (c) 2016 Uber Technologies, Inc.
```
richardartoul: 2018

integration/fs_bootstrap_tags_test.go
```diff
@@ -0,0 +1,126 @@
+// +build integration
+
+// Copyright (c) 2016 Uber Technologies, Inc.
```
richardartoul: 2018

docs/architecture/storage.md
robskillington: Sure thing.

integration/fs_bootstrap_tags_test.go
robskillington: Haha, ta.

persist/fs/msgpack/encoder.go
```diff
@@ -198,13 +202,27 @@ func (enc *Encoder) encodeIndexBloomFilterInfo(info schema.IndexBloomFilterInfo)
 	enc.encodeVarintFn(info.NumHashesK)
 }
 
-func (enc *Encoder) encodeIndexEntry(entry schema.IndexEntry) {
+// We only keep this method around for the sake of testing
+// backwards-compatbility
+func (enc *Encoder) encodeIndexEntryV1(entry schema.IndexEntry) {
+	// Manually encode num fields for testing purposes
+	enc.encodeArrayLenFn(minNumIndexEntryFields)
```
richardartoul: I wonder if you should just hard-code the number of use a v1-specific constant here instead of relying on the minimum (which I think is designed for decoding purposes). Seems like a safe bet since this function is designed to simulate the behavior of old binaries

persist/fs/msgpack/encoder.go
```diff
@@ -198,13 +202,27 @@ func (enc *Encoder) encodeIndexBloomFilterInfo(info schema.IndexBloomFilterInfo)
...
+	enc.encodeVarintFn(entry.Index)
+	enc.encodeBytesFn(entry.ID)
+	enc.encodeVarintFn(entry.Size)
+	enc.encodeVarintFn(entry.Offset)
+	enc.encodeVarintFn(entry.Checksum)
+}
+
+func (enc *Encoder) encodeIndexEntryV2(entry schema.IndexEntry) {
+	// Manually encode num fields for testing purposes
```
richardartoul: This comment seems less relevant for the "current" version

persist/fs/msgpack/encoder.go
robskillington: Sure thing.

persist/fs/read.go
```diff
@@ -380,21 +387,31 @@ func (r *reader) ReadBloomFilter() (*ManagedConcurrentBloomFilter, error) {
...
+	return bytesClone
+}
 
-	idClone.AppendAll(id)
+func (r *reader) entryClonedID(id []byte) ident.ID {
+	return ident.BinaryID(r.entryClonedBytes(id))
+}
 
-	return ident.BinaryID(idClone)
+func (r *reader) entryClonedEncodedTagsTagIter(encodedTags []byte) ident.TagIterator {
```
richardartoul: super nit: you can probably make this `entryClonedEncodedTagsIter` so it doesn't stutter as much

robskillington: Sure thing.

docs/architecture/storage.md
robskillington: Done.

integration/fs_bootstrap_index_test.go
```diff
@@ -0,0 +1,249 @@
...
+// THE SOFTWARE.
+
+package integration
+
+import (
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/m3db/m3ninx/idx"
```
prateek: nit: import order

integration/fs_bootstrap_index_test.go
```diff
@@ -0,0 +1,249 @@
...
+
+	// Match all *e*e* from namespace 2
+	regexpQuery, err = idx.NewRegexpQuery([]byte("city"), []byte("^.*e.*e.*$"))
+	require.NoError(t, err)
+	iter, exhausitive, err = session.FetchTaggedIDs(ns1.ID(),
+		index.Query{regexpQuery}, queryOpts)
+	require.NoError(t, err)
+	defer iter.Finalize()
+
+	verifyQueryMetadataResults(t, iter, exhausitive, verifyQueryMetadataResultsOptions{
```
prateek: +1

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -265,28 +317,51 @@ func (s *fileSystemSource) handleErrorsAndUnfulfilled(
 				}
 			}
 		}
-		resultLock.Unlock()
+		// NB(r): We explicitly do not remove entries from the index results
```
prateek: +1

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -265,28 +317,51 @@ func (s *fileSystemSource) handleErrorsAndUnfulfilled(
...
+		// as they are additive and get merged together with results from other
+		// bootstrappers by just appending the result (unlike data bootstrap
+		// results that when merged replace the block with the current block).
+		// It would also be difficult to remove only series that was added to the
```
richardartoul: were*

storage/bootstrap/result/result_index.go
```diff
@@ -78,6 +79,37 @@ func (r IndexResults) AddResults(other IndexResults) {
 	}
 }
 
+// GetOrAddSegment get or create a new mutable segment.
+func (r IndexResults) GetOrAddSegment(
+	t time.Time,
+	idxopts namespace.IndexOptions,
+	opts Options,
+) (segment.MutableSegment, error) {
+	blockStart := t.Truncate(idxopts.BlockSize())
```
prateek: maybe add a quick note about why this truncation is required and the `% ==0` guarantee the code relies upon.

storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -454,76 +460,195 @@ func (s *fileSystemSource) loadShardReadersDataIntoShardResult(
...
+	d, err := convert.FromMetricIter(id, tagsIter)
+	if err != nil {
+		release()
+		return err
+	}
+
+	runResult.Lock()
+	exists, err = segment.ContainsID(idBytes)
+	// ID and tags no longer required below
+	release()
```
richardartoul: it looks like FromMetricIter clones the ID so you could probably do:

```golang
d, err := convert.FromMetricIter(id, tagsIter)
release()
if err != nil {
    return err
}

runResult.Lock()
exists, err = segment.ContainsID(d.ID)
```

robskillington: Sure thing.

storage/bootstrap/bootstrapper/fs/source.go
robskillington: Ta

storage/bootstrap/result/result_index.go
robskillington: Sure thing.

PR #591, opened 2018-05-07
Include SnapshotEnabled when converting from proto to Namespace

richardartoul:


storage/namespace/convert.go
```diff
@@ -103,6 +103,7 @@ func ToMetadata(
 		SetCleanupEnabled(opts.CleanupEnabled).
 		SetRepairEnabled(opts.RepairEnabled).
 		SetWritesToCommitLog(opts.WritesToCommitLog).
+		SetSnapshotEnabled(opts.SnapshotEnabled).
```
prateek: could you add a simple test for this too (not just the prop test)

PR #595, opened 2018-05-07
Make decoder re-usable across `Reset()` calls

prateek:


serialize/decoder.go
```diff
@@ -127,7 +128,7 @@ func (d *decoder) decodeID() (ident.ID, error) {
 	}
 
 	if limit := d.opts.TagSerializationLimits().MaxTagLiteralLength(); l > limit {
-		return nil, fmt.Errorf("tag literal too long [ limit = %d, observed = %d]", limit, int(l))
+		return nil, fmt.Errorf("tag literal too long [ limit = %d, observed = %d] ", limit, int(l))
```
richardartoul: I think you didn't make these spaces consistent

prateek: thanks for the catch. def a typo

PR #598, opened 2018-05-08
Re-wire indexing configuration

prateek:


services/m3dbnode/config/config_test.go
```diff
@@ -186,6 +186,16 @@ pooling:
         capacity: 4096
         lowWatermark: 0.01
         highWatermark: 0.02
+    tagArrayPool:
+        size: 4096
```
robskillington: Might want to use a little higher considering this is kind of like an example config.  Perhaps 65536?

services/m3dbnode/config/config_test.go
```diff
@@ -186,6 +186,16 @@ pooling:
...
+        capacity: 8
+        maxCapacity: 32
+        lowWatermark: 0.01
+        highWatermark: 0.02
+    indexResultsPool:
+        size: 4096
```
robskillington: Probably won't be more than 4k distinct queries at any one point in time I guess, 8192 might be safer though.

services/m3dbnode/config/config_test.go
prateek: Sure thing

services/m3dbnode/config/config_test.go
prateek: Sure thing

prateek: done

services/m3dbnode/config/config_test.go
prateek: done

PR #605, opened 2018-05-08
Bootstrap index from commitlog

richardartoul:


storage/bootstrap/bootstrapper/commitlog/source_index_test.go
```diff
@@ -0,0 +1,271 @@
...
+	md, err := namespace.NewMetadata(testNamespaceID, namespaceOptions)
+	require.NoError(t, err)
+
+	now := time.Now()
+	start := now.Truncate(indexBlockSize)
+
+	fooTags := ident.Tags{ident.StringTag("city", "ny"), ident.StringTag("conference", "monitoroma")}
+	barTags := ident.Tags{ident.StringTag("city", "sf")}
+	bazTags := ident.Tags{ident.StringTag("city", "oakland")}
+	foo := commitlog.Series{Namespace: testNamespaceID, Shard: 0, ID: ident.StringID("foo"), Tags: fooTags}
```
prateek: nit: could you use a var instead of `0` (and similarly in the the references below) and in the bootstrap request line (`targetRanges := result.ShardTimeRanges{0: ranges, 1: ranges, 2: ranges, 5: ranges}`). it'd be a lot clearer that way

richardartoul: done (used shardZero + arithmetic to get the other shards)

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -275,6 +276,37 @@ func (s *commitLogSource) shouldEncodeSeries(
 	return ranges.Overlaps(blockRange)
 }
 
+func (s *commitLogSource) shouldIncludeInIndex(
```
robskillington: nit: Maybe `shouldIndexSeries` is more similar to the other method?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -275,6 +276,37 @@ func (s *commitLogSource) shouldEncodeSeries(
...
+	shard uint32,
+	ts time.Time,
```
robskillington: Any reason why this is `ts time.Time` vs `dp ts.Datapoint` in the other method?  Can we make these symmetrical, either one or the other?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -492,15 +524,88 @@ func (s *commitLogSource) ReadIndex(
...
+
+	for iter.Next() {
+		series, dp, _, _ := iter.Current()
+		if !s.shouldIncludeInIndex(
+			series.Shard, dp.Timestamp, highestShard, indexBlockSize, bootstrapRangesByShard) {
+			continue
+		}
+
+		segment, err := indexResults.GetOrAddSegment(
+			dp.Timestamp, ns.Options().IndexOptions(), s.opts.ResultOptions())
```
robskillington: Maybe pull these two vars so that these options aren't resolved by a method call each loop iter?

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Yeah I'm planning on renaming the other method. I think its a bad name because it evaluates its on a datapoint-per-datapoint basis

richardartoul: Just didn't wanna do it in this P.R

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: sure I just figured I'd only pass the required information. Will update the other method.

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: yeah duh, my bad

storage/bootstrap/bootstrapper/commitlog/source.go
robskillington: No problems.

PR #607, opened 2018-05-09
Add peers index bootstrapping capability

robskillington:


client/session.go
```diff
@@ -2213,14 +2245,35 @@ func (s *session) streamBlocksMetadataFromPeerV2(
...
+				tags         ident.Tags
+				tagDecodeErr error
+			)
+			if encodedTags := elem.EncodedTags; len(encodedTags) != 0 {
+				tagDecoderCheckedBytes.Reset(encodedTags)
+				tagDecoder.Reset(tagDecoderCheckedBytes)
+				tags = make(ident.Tags, 0, tagDecoder.Remaining())
+				for tagDecoder.Next() {
+					curr := tagDecoder.Current()
+					tags = append(tags, idPool.CloneTag(curr))
```
richardartoul: kinda sucks we have to allocate all the tags per peer. Ah well

client/session.go
```diff
@@ -2213,14 +2245,35 @@ func (s *session) streamBlocksMetadataFromPeerV2(
...
+				}
+				tagDecodeErr = tagDecoder.Err()
+			}
+
 			// Error occurred retrieving block metadata, use default values
+			var err error
 			if elem.Err != nil {
```
richardartoul: maybe a comment that element err takes precedence over tag decoding error

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -512,10 +514,146 @@ func (s *peersSource) ReadIndex(
 	shardsTimeRanges result.ShardTimeRanges,
 	opts bootstrap.RunOptions,
 ) (result.IndexBootstrapResult, error) {
-	// FOLLOWUP(r): implement the peers source returning
-	// index segments that are streamed from peers as required.
-	// Try to cache some series metadata on the peers source itself
-	// from work done in ReadData(...) to help avoid rereading as
-	// much as possible.
-	return result.NewIndexBootstrapResult(), nil
+	// FOLLOWUP(r): Try to reuse any metadata fetched during the ReadData(...)
```
richardartoul: Think maybe we should make a master issue for index-bootstrap performance improvements? I'm gonna have some for the commit log as well.

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -512,10 +514,146 @@ func (s *peersSource) ReadIndex(
...
+}
+
+func (s *peersSource) markIndexResultErrorAsUnfulfilled(
+	r result.IndexBootstrapResult,
+	resultLock *sync.Mutex,
+	err error,
+	shard uint32,
+	timeRange xtime.Range,
+) {
+	// NB(r): We explicitly do not remove entries from the index results
```
richardartoul: can you update this comment to talk about peer metadata instead of data files

storage/series/series.go
```diff
@@ -400,11 +400,12 @@ func (s *dbSeries) FetchBlocksMetadata(
 		bufferResults.Close()
 	}
 
-	id := s.opts.IdentifierPool().Clone(s.id)
-
 	res.Sort()
 
-	return block.NewFetchBlocksMetadataResult(id, res), nil
+	// NB(r): Since ID and Tags are garbage collected we can safely
+	// return refs.
```
richardartoul: "are garbage collected (never manually finalized)" maybe to be a little more clear

richardartoul: I guess this was already an issue, but now that I'm thinking about it, should we add a comment at the server level to not finalize the ID / tags we're returning? I notice you only do it in the shard for things that come from disk but seems like something someone could fuck up easily.

Maybe we should just file an issues to have a wrapper for special types of IDs/tags that does an [invariant violated] or panics when you try to finalize it

client/session.go
robskillington: Yeah, I can do a TODO here to pool these perhaps.  We would need a wrapper type that has a ref to the pool to be able to pool the slice itself.  Although if they go into a bootstrap result, you have to hold onto them.  If you flush them out to disk then you finalize at that time, but yeah there's no way to currently track these at this time.

client/session.go
robskillington: Sure thing.

network/server/tchannelthrift/node/service.go
```diff
@@ -333,6 +326,26 @@ func (s *service) FetchTagged(tctx thrift.Context, req *rpc.FetchTaggedRequest)
...
 }
 
+func (s *service) encodeTags(
+	enc serialize.TagEncoder,
+	tags ident.TagIterator,
+) (checked.Bytes, error) {
+	if err := enc.Encode(tags); err != nil {
+		// should never happen
+		wrappedErr := xerrors.NewRenamedError(err, fmt.Errorf("unable to encode tags"))
+		s.logger.Warnf("[invariant violated] %v", wrappedErr)
```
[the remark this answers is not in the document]
robskillington: Sure thing.

network/server/tchannelthrift/node/service.go
```diff
@@ -333,6 +326,26 @@ func (s *service) FetchTagged(tctx thrift.Context, req *rpc.FetchTaggedRequest)
...
+		return nil, wrappedErr
+	}
+	encodedTags, ok := enc.Data()
+	if !ok {
+		// should never happen
+		err := fmt.Errorf("unable to encode tags: unable to unwrap bytes")
+		s.logger.Warnf("[invariant violated] %v", err)
```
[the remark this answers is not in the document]
robskillington: Sure thing.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: Yeah, we can potentially pile it onto this one:
https://github.com/m3db/m3db/issues/585

storage/bootstrap/bootstrapper/peers/source.go
robskillington: This is for peer metadata, thats why I mention "index results".  I can change it if that's not very clear though?

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: yeah lets throw it in there, I'll put my stuff there too

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: > 	// It would also be difficult to remove only series that were added to the
 	// index block as results from data files can be subsets of the index block
 	// and there's no way to definitively delete the entry we added as a result
 	// of just this data file failing.

richardartoul: "data files"

richardartoul: I just meant that you use the word data files, which I thought was just copy-pasta from the other bootstrapper, but I guess technically the metadata from peers originates in fileset files in some cases.

robskillington: Fair, yeah I'll say "series that were added to the index block as results from data blocks received from peers".

It was definitely a copy paste heh.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: Added it to there, and added the commit log bootstrapper too.

storage/series/series.go
robskillington: Unfortunately though that means we can't mix types that are meant to be finalized with ones that aren't meant to be finalized.

We could however return a type that on Finalize does absolutely nothing and is a no-op perhaps.  That could be nice, thoughts?

storage/bootstrap/bootstrapper/peers/source.go
robskillington: Updated.

client/session.go
```diff
@@ -3248,36 +3311,68 @@ func (it *peerBlocksIter) Next() bool {
...
+) error {
 	start := time.Unix(0, block.Start)
 	result, err := r.newDatabaseBlock(block)
 	if err != nil {
 		return err
 	}
 
+	var (
+		tags          ident.Tags
+		tryDecodeTags bool
```
richardartoul: this seems more like its `triedToDecodeTags`

network/server/tchannelthrift/node/service.go
```diff
@@ -597,46 +613,65 @@ func (s *service) FetchBlocksMetadataRawV2(tctx thrift.Context, req *rpc.FetchBl
...
-		id := fetchedMetadata.ID.Data().Bytes()
+
+		var (
+			id          = fetchedMetadata.ID.Bytes()
+			tags        = fetchedMetadata.Tags
+			encodedTags []byte
+		)
+		if tags != nil && tags.Remaining() > 0 {
+			enc := s.pools.tagEncoder.Get()
+			ctx.RegisterFinalizer(enc)
```
richardartoul: is the encoder not safe to close immediately after receiving the encoded data?

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -512,10 +514,146 @@ func (s *peersSource) ReadIndex(
...
+	for shard, ranges := range shardsTimeRanges {
+		shard, ranges := shard, ranges
+		wg.Add(1)
+		workers.Go(func() {
+			defer wg.Done()
+
+			iter := ranges.Iter()
+			for iter.Next() {
+				currRange := iter.Value()
+				metadata, err := session.FetchBootstrapBlocksMetadataFromPeers(ns.ID(),
```
richardartoul: Sorry I'm just catching this now, but could you take a look at fetchBootstrapBlocksFromPeers? after currRange := iter.Value() it then  steps through the range one block at a time.

This is just metadata so its less important, but you could imagine this causing much higher memory utilization than necessary during node joins in clusters with lots of data and very large disks (I.E you have like 16 Terabytes of storage on a host, and for each shard you're fetching all the IDs/Tags for all of the historical blocks instead of a single block at a time)

storage/series/series.go
richardartoul: yeah that would be chill. Happy to do it here or file an issue or whatever

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: Also if we decide to do this, a quick check to make sure that the range divides evenly over the blocksize would be good (it always should, but if it doesn't weird things might happen...)

client/session.go
```diff
@@ -3248,36 +3311,68 @@ func (it *peerBlocksIter) Next() bool {
...
+	)
```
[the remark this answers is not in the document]
robskillington: Yeah I'd prefer not to alter it during this PR.  The main reason it's like this is that you need to keep merging if someone beat you to merging another block with the same block.

storage/shard_race_prop_test.go
```diff
@@ -132,3 +134,130 @@ func anyIDs() gopter.Gen {
...
+	for i := 0; i < numRoutines; i++ {
+		barrier <- struct{}{}
+	}
+
+	wg.Wait()
+}
+
+func TestShardTickWriteReadRace(t *testing.T) {
+	parameters := gopter.DefaultTestParameters()
+	// seed := time.Now().UnixNano() // int64(1521412723796005124)
```
prateek: mind un-setting the fixed value and use the time.Now().. version instead - had it set while debugging the particular failure case.

client/session.go
```diff
@@ -2253,8 +2289,9 @@ func (s *session) streamBlocksMetadataFromPeerV2(
 			}
 
 			metadataCh <- receivedBlockMetadata{
-				peer: peer,
-				id:   clonedID,
+				peer:        peer,
+				id:          clonedID,
+				encodedTags: encodedTags,
```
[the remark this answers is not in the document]
robskillington: Nope, not right now.  We need to decide how we're deal with tag mutation before we can do something with it.  i.e. if you write two different sets of tags, either in different time windows, or just to two different replicas, how do you combine them?  I'm in favor of just merging, however not sure how to deal with collisions.

storage/shard_race_prop_test.go
robskillington: Sure thing.

storage/shard.go
```diff
@@ -128,6 +129,9 @@ type dbShard struct {
 	newSeriesBootstrapped    bool
 	ticking                  bool
 	shard                    uint32
+
+	// Only used in tests
+	testListIterBatchSize int
```
prateek: this feels really ugly, should we just drop that parameter from the test?

network/server/tchannelthrift/node/service.go
```diff
@@ -333,6 +326,28 @@ func (s *service) FetchTagged(tctx thrift.Context, req *rpc.FetchTaggedRequest)
...
 
+func (s *service) encodeTags(
+	enc serialize.TagEncoder,
+	tags ident.TagIterator,
+) (checked.Bytes, error) {
+	if err := enc.Encode(tags); err != nil {
+		// should never happen
+		err = xerrors.NewRenamedError(err, fmt.Errorf("unable to encode tags"))
+		l := instrument.EmitInvariantViolationAndGetLogger(s.opts.InstrumentOptions())
+		l.Warn(err.Error())
```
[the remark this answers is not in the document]
robskillington: This should be error, cheers.

storage/shard.go
```diff
@@ -1535,6 +1541,10 @@ func (s *dbShard) FetchBlocksMetadataV2(
 					blockStart, err)
 			}
 
+			// Make sure ID and tags get cleaned up after read is done
+			ctx.RegisterFinalizer(id)
+			ctx.RegisterFinalizer(resource.FinalizerFn(tags.Close))
```
prateek: should we just add `ctx.RegisterCloser(tags)` to avoid allocs like this?

network/server/tchannelthrift/node/service.go
prateek: Nope. The encoder still owns the `checked.Bytes` returned

client/session.go
```diff
@@ -3753,3 +3860,30 @@ func (v FetchBlocksMetadataEndpointVersion) String() string {
...
+	idPool ident.Pool,
+) (ident.Tags, error) {
+	if encodedTags == nil {
+		return nil, nil
+	}
+
+	encodedTags.IncRef()
+	tagDecoder.Reset(encodedTags)
+
+	tags := make(ident.Tags, 0, tagDecoder.Remaining())
```
prateek: Could you re-use the `storage/index.TagArrayPool` here? would have to pipe from sessionPools all the way to the `block.Metadata.Finalize()` method I think

network/server/tchannelthrift/types.go
```diff
@@ -20,10 +20,19 @@
...
+	"github.com/m3db/m3x/instrument"
+)
 
 // Options controls server behavior
 type Options interface {
+	// SetInstrumentOptions sets the instrumentation options
+	SetInstrumentOptions(value instrument.Options) Options
+
+	// InstrumentOptions returns the instrumentation options
+	InstrumentOptions() instrument.Options
```
prateek: mind changing the `NewService()` ctor to use this instead of `s.db.Options().InstrumentOptions()`

storage/series/series.go
robskillington: Yeah I'll file an issue, here it is:
https://github.com/m3db/m3db/issues/611

storage/block/types.go
```diff
@@ -108,6 +108,7 @@ type FetchBlockMetadataResults interface {
 // FetchBlocksMetadataResult captures the fetch results for multiple blocks.
 type FetchBlocksMetadataResult struct {
 	ID     ident.ID
+	Tags   ident.TagIterator
```
prateek: might be better to have this be `ident.Tags` too - the series has the slice anyway. This way we can avoid the `TagSliceIterator` alloc per ID, have only a single instance In `service` and re-use using `Reset()` method you added

client/session.go
robskillington: Sure thing.

storage/series/series.go
prateek: How about just exposing `SetNonFinalizable()` on `ident.ID`? Way less type changing - and a more obvious failure mode - we just panic if someone tries to Finalize() such an ID.

network/server/tchannelthrift/node/service.go
robskillington: Unfortunately not, it just returns a checked bytes ref that it has itself.  I could copy this out, but it seems unnecessary.  

Or I could create a single one and finalize at the end of this method and copy out and put in a bytes pool each time, however that also seems like more work than necessary perhaps.

storage/series/series.go
robskillington: The thing is, when we mix IDs we actually want to finalize some and some not (i.e. data coming from disk should be finalized and data coming from in memory shouldn't).

So either the caller has to do `if id.ShouldFinalize() { id.Finalize() }` or you return a type that just doesn't finalize when you call `id.Finalize()`.  However we could definitely do `id.SetFinalizeEnabled(false)` and that would disable calls to `id.Finalize()` doing anything.

storage/shard.go
robskillington: We could, however it is kind of the point though of the test I think (to fix off by one/length of array issues).  I don't feel too strongly one way or another, I can drop it.

prateek: could we hide it away using fn injection, something like `shard.iterateBatchSizeFn`?

storage/series/series.go
prateek: Yeah so my point was we would only set `NonFinalizable()` on id's in memory, and no user would be expected to call `Finalize` on stuff with that set. They wouldn't have to check `if id.ShouldFinalize() { id.Finalize() }` because we'd panic if they did it. 

More I think about it, I agree exposing a new type is cleaner cause it makes it explicit in the contract.

storage/bootstrap/bootstrapper/peers/source.go
robskillington: Np, will make this the same as the data code path and do it per block.

storage/shard.go

client/session.go
```diff
@@ -3171,35 +3213,56 @@ func (b *baseBlocksResult) newDatabaseBlock(block *rpc.Block) (block.DatabaseBlo
 	return result, nil
 }
 
+// Ensure streamBlocksResult implements blocksResult
+var _ blocksResult = (*streamBlocksResult)(nil)
```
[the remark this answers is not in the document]
robskillington: I'm trying to avoid a program startup alloc heh.

client/types.go
```diff
@@ -154,26 +154,36 @@ type AdminSession interface {
...
 		version FetchBlocksMetadataEndpointVersion,
 	) (PeerBlockMetadataIter, error)
 
-	// FetchBootstrapBlocksFromPeers will fetch the most fulfilled block
-	// for each series in a best effort method from available peers
-	FetchBootstrapBlocksFromPeers(
-		namespace namespace.Metadata,
+	// FetchBlocksMetadataFromPeers will fetch the blocks metadata from
+	// available peers
+	FetchBlocksMetadataFromPeers(
```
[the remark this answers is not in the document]
robskillington: Yeah, there's some callers somewhere in the codebase (pretty sure its repairs).

client/session.go
robskillington: I think we'd have to create a wrapper type and plumb it through here.  Let's fix this async:
https://github.com/m3db/m3db/issues/614

network/server/tchannelthrift/types.go
robskillington: Sure thing.

storage/series/series.go
robskillington: Yeah but then you can't share a `ids []ident.ID` because `ID` implements `Finalize` and the `NonFinalizeable` one won't so it can't be one of the elements in the array.  Also if you make it `ids []ident.NonFinalizeable` then you can't finalize the ones that _do_ need finalizing.

storage/block/types.go
robskillington: As discussed, unfortunately can't as the on disk representation needs to be returned with the same type and its much more efficient to return TagDecoder which won't allocate a []ident.Tag array.  They need to share the same type as the in series one.

storage/series/series.go
prateek: I don't believe we share `[]ident.ID` in the code today, we always wrap with some other type - iterator/etc. But yeah adding the additional type becomes a widely cascading change cause pretty much anything that wraps the ident.ID would be affected. 

Maybe the version of `id.SetFinalizeEnabled(false)` turning off `Finalize()` from doing anything would be the lowest cost change?

storage/block/types.go
prateek: Sounds good.

PR #608, opened 2018-05-09
[WIP] Batch Shard -> Index insertions

prateek:
- Batch shard -> index insertions to minimize lock contention

Pending:
- [ ] can we use a single type instead of index.WriteBatchEntry and the other one in `storage/`?
- [ ] fix tests
- [ ] more tests for inc/def ref guarantees 
- [ ] utility method on index.WriteBatchEntry to bulk finalize
- [ ] []index.WriteBatchEntry pooling?


storage/index.go
```diff
@@ -350,14 +347,12 @@ func (i *nsIndex) writeBatchForBlockStartWithRLock(
...
 	// i.e. we have the block and the inserts, perform the writes.
 	result, err := block.WriteBatch(inserts)
+
 	// NB: we don't need to do anything to the OnIndexSeries refs in `inserts` at this point,
 	// the index.Block WriteBatch assumes responsibility for calling the appropriate methods.
 	if numErr := result.NumError; numErr != 0 {
 		i.metrics.AsyncInsertErrors.Inc(numErr)
 	}
-	if err != nil {
-		i.logger.Errorf("unable to write to index, dropping inserts. [%v]", err)
```
robskillington: Maybe we can keep this signal now?

storage/shard.go
```diff
@@ -180,6 +180,20 @@ type dbShardEntry struct {
 	index          uint64
 	curReadWriters int32
 	reverseIndex   struct {
+		// NB(prateek): writeAttemptedBlockstart[1-2]Nanos is used to indicate the index block start(s)
```
prateek: I'm really leaning towards using a RW lock here. It'll be 24 more bytes per shardEntry, but will so greatly reduce the complexity of the code that it's a no brainer. Only finished writing this version up to convince myself it's a terrible idea.

storage/shard_insert_queue.go
```diff
@@ -101,8 +101,15 @@ type dbShardInsertAsyncOptions struct {
...
-	hasPendingIndexing       bool
 	hasPendingRetrievedBlock bool
+	hasPendingIndexing       bool
+
+	// NB(prateek): `entryRefCountIncremented` indicates if the
+	// entry provided along with the dbShardInsertAsyncOptions
+	// already has it's ref count incremented. It's used to
+	// correctly manage the lifecycle of the entry across the
+	// shard -> shard Queue -> shard boundaries.
+	entryRefCountIncremented bool
```
prateek: @robskillington I like how this mechanism works out for transferring ownership of the incRef. Esp how it simplifies insertSeriesBatch. What do you think?

storage/shard.go
prateek: @robskillington please could you give this a once over and see if i'm missing something more obvious.

storage/index.go
prateek: yeah makes sense

prateek: for sure, we should.

storage/index/block.go
```diff
@@ -111,86 +108,74 @@ func (b *block) EndTime() time.Time {
...
+func (b *block) WriteBatch(inserts WriteBatchEntryByBlockStartAndID) (WriteBatchResult, error) {
+	b.RLock()
+	defer b.RUnlock()
 
 	if b.state != blockStateOpen {
 		// NB: releasing all references to inserts
-		for _, insert := range inserts {
-			insert.OnIndexSeries.OnIndexFinalize()
-		}
+		WriteBatchEntriesFinalizer(inserts).Finalize()
```
prateek: when you revert to the earlier version, would be good to retain this utility type - `WriteBatchEntriesFinalizer`

storage/index/block.go
```diff
@@ -111,86 +108,74 @@ func (b *block) EndTime() time.Time {
 	return b.endTime
 }
 
-func (b *block) updateWriteBatchWithLock(inserts []WriteBatchEntry) {
-	for _, insert := range inserts {
```
prateek: the new version of `WriteBatchEntry` has idents, not docs. so when you revert, this will have to do the convert.FromMetric call

storage/index/block.go
```diff
@@ -111,86 +108,74 @@ func (b *block) EndTime() time.Time {
...
 		}, b.writeBatchErrorInvalidState(b.state)
 	}
 
-	b.updateWriteBatchWithLock(inserts)
-	err := b.segment.InsertBatch(b.writeBatch)
-	if err == nil {
-		for _, insert := range inserts {
-			insert.OnIndexSeries.OnIndexSuccess(b.endTime)
-			insert.OnIndexSeries.OnIndexFinalize()
+	var (
```
prateek: can revert to the earlier batch api usage for this method

storage/index/types.go
```diff
@@ -116,14 +117,15 @@ type MutableSegmentAllocator func() (segment.MutableSegment, error)
 // to do lifecycle management of any resources retained during indexing.
 type OnIndexSeries interface {
 	// OnIndexSuccess is executed when an entry is successfully indexed. The
-	// provided value for `indexEntryExpiry` describes the TTL for the indexed
-	// entry.
-	OnIndexSuccess(indexEntryExpiry time.Time)
+	// provided value for `blockStart` is the blockStart for which the write
```
prateek: the new interface changes here need to stick around

storage/index/types.go
```diff
@@ -183,41 +178,111 @@ type BlockTickResult struct {
 	NumDocs     int64
 }
 
+// WriteBatchEntry captures a document to index, and the lifecycle hooks to call thereafter.
+type WriteBatchEntry struct {
```
prateek: should retain the changes to this type

storage/index/types.go
```diff
@@ -183,41 +178,111 @@ type BlockTickResult struct {
...
+	OnIndexSeries OnIndexSeries
+}
+
+// WriteBatchEntriesFinalizer is a utility type to provide syntactic sugar to finalize references in the slice.
+type WriteBatchEntriesFinalizer []WriteBatchEntry
+
+// Finalize finalizes all the references in the provided slice.
+func (w WriteBatchEntriesFinalizer) Finalize() {
+	for _, entry := range w {
+		if entry.OnIndexSeries != nil {
```
prateek: need this check cause we set entries to the empty value if they don't need to be indexed (in `index.go:InsertBatch`)

storage/index/types.go
```diff
@@ -183,41 +178,111 @@ type BlockTickResult struct {
...
+			entry.OnIndexSeries.OnIndexFinalize(xtime.ToUnixNano(entry.Timestamp))
+		}
+	}
+}
+
 // WriteBatchEntryByBlockStart implements sort.Interface for WriteBatchEntry slices
-// based on the BlockStart field.
-type WriteBatchEntryByBlockStart []WriteBatchEntry
+// based on the Timestamp and ID fields.
+type WriteBatchEntryByBlockStartAndID []WriteBatchEntry
```
prateek: can revert this to only sorting by blockstart

storage/index/types.go
```diff
@@ -183,41 +178,111 @@ type BlockTickResult struct {
...
+	return bytes.Compare(w[i].ID.Bytes(), w[j].ID.Bytes()) < 0
+}
 
 // ForEachBlockStartFn is lambda to iterate over WriteBatchEntry(s) a single blockStart at a time.
-type ForEachBlockStartFn func(blockStart xtime.UnixNano, writes []WriteBatchEntry)
+type ForEachBlockStartFn func(timestamp time.Time, writes WriteBatchEntryByBlockStartAndID)
 
-// ForEachBlockStart iterates over the provided WriteBatchEntryByBlockStart, and calls `fn` on each
+// ForEachIDFn is lambda to iterate over WriteBatchEntry(s) a single ID at a time.
+type ForEachIDFn func(writes WriteBatchEntryByBlockStartAndID)
```
prateek: can delete this

storage/index/types.go
```diff
@@ -183,41 +178,111 @@ type BlockTickResult struct {
...
+	}
+	// spill over
+	if startIdx < len(w) {
+		fn(w[startIdx].Timestamp, w[startIdx:])
+	}
+}
+
+// ForEachID iterates over the provided WriteBatchEntryByBlockStartAndID, and calls `fn` on each
+// group of elements with the same ID.
+func (w WriteBatchEntryByBlockStartAndID) ForEachID(fn ForEachIDFn) {
```
prateek: can delete this

storage/index.go
```diff
@@ -224,61 +224,62 @@ func newNamespaceIndexWithOptions(
...
+	for j := range entries {
+		var (
+			timestamp  = entries[j].Timestamp
+			onIndexFn  = entries[j].OnIndexSeries
+			blockStart = i.BlockStartForWriteTime(timestamp)
+		)
+		if !futureLimit.After(timestamp) {
+			onIndexFn.OnIndexFinalize(blockStart)
+			entries[j] = emptyEntry // indicate we don't need to index this.
+			// TODO(prateek): capture that this needs to return m3dberrors.ErrTooFuture
```
prateek: can leave this around and convert to `FOLLOWUP(prateek)`

storage/index.go
```diff
@@ -224,61 +224,62 @@ func newNamespaceIndexWithOptions(
...
+		}
 
-	if !pastLimit.Before(timestamp) {
-		fns.OnIndexFinalize()
-		return m3dberrors.ErrTooPast
-	}
+		if !pastLimit.Before(timestamp) {
+			onIndexFn.OnIndexFinalize(blockStart)
+			entries[j] = emptyEntry // indicate we don't need to index this.
+			// TODO(prateek): capture that this needs to return m3dberrors.ErrTooPast
```
prateek: here too

storage/index_insert_queue.go
```diff
@@ -48,8 +46,9 @@ const (
 )
 
 var (
-	defaultIndexBatchBackoff   = time.Second
-	defaultIndexPerSecondLimit = 10000
+	// TODO(prateek): undo this stuff
+	defaultIndexBatchBackoff   = time.Millisecond
```
prateek: i say leave this around for now. they're sensible enough defaults. I'll come back and wire up to runtime/config in https://github.com/m3db/m3db/pull/604

storage/index_insert_queue.go
```diff
@@ -234,21 +227,20 @@ func (q *nsIndexInsertQueue) Stop() error {
...
 
-var nsIndexInsertZeroed index.WriteBatchEntry
-
 func (b *nsIndexInsertBatch) Reset() {
 	b.wg = &sync.WaitGroup{}
 	// We always expect to be waiting for an index
 	b.wg.Add(1)
 	for i := range b.inserts {
-		b.inserts[i] = nsIndexInsertZeroed
+		// TODO(prateek): if we start pooling `[]index.WriteBatchEntry`, then we could return to the pool here.
```
prateek: can change to a FOLLOWUP

storage/shard.go
```diff
@@ -826,22 +779,23 @@ func (s *dbShard) writeAndIndex(
...
-		if err == nil && needsIndex {
-			entry.onIndexPrepare()
-			err = s.reverseIndex.Write(entry.series.ID(), entry.series.Tags(), timestamp, entry)
-		}
-		entry.decrementReaderWriterCount()
+		commitLogSeriesID = entry.Series.ID()
+		commitLogSeriesTags = entry.Series.Tags()
+		commitLogSeriesUniqueIndex = entry.Index
+		if err == nil && shouldReverseIndex {
+			if entry.NeedsIndexUpdate(s.reverseIndex.BlockStartForWriteTime(timestamp)) {
```
prateek: maybe add a comment that `NeedsIndexUpdate` has CAS semantics here so we don't change in an incompatible way later.

storage/shard.go
```diff
@@ -1024,7 +978,49 @@ type insertAsyncResult struct {
...
+	// if operating in async mode, we're done
+	if async {
+		return nil
+	}
+
+	// if indexing in sync mode, wait till we're done and ensure we have have indexed the entry
+	wg.Wait()
+	if entry.IndexedForBlockStart(indexBlockStart) {
+		// i.e. indexing failed
+		return fmt.Errorf("internal error: unable to index series")
```
prateek: move to a const at the top of the file

storage/shard.go
```diff
@@ -1156,14 +1172,17 @@ func (s *dbShard) insertSeriesBatch(inserts []dbShardInsert) error {
 		if err != errShardEntryNotFound {
 			// Shard is not taking inserts
 			s.Unlock()
+			// FOLLOWUP(prateek): is this an existing bug? why don't we need to release any ref's we've inc'd
```
prateek: can leave around, i'll circle back for this.

storage/shard_insert_queue.go
prateek: should retain this

storage/shard_insert_queue_test.go
```diff
@@ -1,3 +1,4 @@
+// +build disable_for_now
```
prateek: undo this

storage/types.go
```diff
@@ -429,12 +428,15 @@ type databaseShard interface {
 
 // namespaceIndex indexes namespace writes.
 type namespaceIndex interface {
-	// Write indexes timeseries ID by provided Tags.
-	Write(
-		id ident.ID,
-		tags ident.Tags,
-		timestamp time.Time,
-		fns index.OnIndexSeries,
+	// BlockStartForWriteTime returns the index block start
```
prateek: these changes need to stay around

storage/index.go
```diff
@@ -350,14 +352,12 @@ func (i *nsIndex) writeBatchForBlockStartWithRLock(
 
 	// i.e. we have the block and the inserts, perform the writes.
 	result, err := block.WriteBatch(inserts)
+
```
prateek: would be good to add the numDuplicates  to the result type returned from the blockWriteBatch call, and a metric for it.

storage/index/types.go
prateek: might want to add on a ReceivedTime to capture e2e index latency with this struct too

storage/block/block.go
```diff
@@ -294,8 +290,6 @@ func (b *dbBlock) ResetRetrievable(
 }
 
 func (b *dbBlock) stream(ctx context.Context) (xio.BlockReader, error) {
-	b.ctx.DependsOn(ctx)
```
richardartoul: Did we have a double DependsOn? I see one in Stream() too...that might be my bad

storage/block/block.go
```diff
@@ -365,8 +366,6 @@ func (b *dbBlock) resetSegmentWithLock(seg ts.Segment) {
 	b.retriever = nil
 	b.retrieveID = nil
 	b.wasRetrievedFromDisk = false
-
-	b.ctx.RegisterFinalizer(&seg)
```
richardartoul: I see that you moved this to a synchronous call in resetRetrievableWithLock, but is that enough? Seems like it would never happen for the ResetRetrievable path

storage/block/block.go
```diff
@@ -391,16 +390,7 @@ func (b *dbBlock) Close() {
...
-
-	// NB(xichen): we use the worker pool to close the context instead doing
-	// an asynchronous context close to explicitly control the context closing
-	// concurrency. This is particularly important during a node removal because
-	// all the shards are removed at once, causing a goroutine explosion without
-	// limiting the concurrency. We also cannot do a blocking close here because
-	// the block may be closed before the underlying context is closed, which
-	// causes a deadlock if the block and the underlying context are closed
-	// from within the same goroutine.
-	b.opts.CloseContextWorkers().Go(b.ctx.BlockingClose)
```
richardartoul: Aren't we just leaking the ctx by removing this?

storage/index.go
```diff
@@ -207,156 +205,144 @@ func newNamespaceIndexWithOptions(
...
-		// insertion code path. We could eliminate the wait group if we did that.
-		// (2) we could provide an OnIndexError callback within OnIndexSeries,
-		// again ensuring it was called upon failure, and cache the error within
-		// the dbShardEntry. Then we're guaranteed that the error would be set
-		// once wg.Wait returns, and the shard insert code-paths would be able
-		// to retrieve the error from the dbShardEntry. Not pretty, but probably
-		// a lot cheaper than (1).
 		wg.Wait()
+
+		// Resort the batch by initial enqueue order
```
richardartoul: Re-sort

storage/index.go
```diff
@@ -207,156 +205,144 @@ func newNamespaceIndexWithOptions(
...
+				return
+			}
+
+			if !entry.Timestamp.After(pastLimit) {
+				batch.MarkUnmarkedEntryError(m3dberrors.ErrTooPast, idx)
+				return
+			}
+		})
+
+		// we sort the inserts by which block they're applicable for, and do the inserts
```
richardartoul: Comment would read better if it just started with "Sort the inserts..."

storage/series/lookup/entry.go
```diff
@@ -0,0 +1,248 @@
...
+// IndexedForBlockStart returns a bool to indicate if the Entry has been successfully
+// indexed for the given index blockstart.
+func (entry *Entry) IndexedForBlockStart(indexBlockStart xtime.UnixNano) bool {
+	entry.reverseIndex.RLock()
+	isIndexed := entry.reverseIndex.indexedWithRLock(indexBlockStart)
+	entry.reverseIndex.RUnlock()
+	return isIndexed
+}
+
+// NeedsIndexUpdate returns a bool to indicate if the Entry requires to be indexed
```
richardartoul: super nit, but "requires to be" sounds super weird to me. Presumably this comment was already here though

storage/series/lookup/entry.go
```diff
@@ -0,0 +1,248 @@
...
+// for the provided blockStart. It only allows a single index attempt at a time
+// for a single entry.
+// NB(prateek): NeedsIndexUpdate is a CAS, i.e. when this method returns true, it
+// also sets state on the entry to indicate that a write for the given blockStart
+// is going to be sent to the index, and other go routines should not attempt the
+// same write. Callers are expected to ensure they follow this guideline.
+// Further, every call to NeedsIndexUpdate which returns true needs to have a corresponding
+// OnIndexFinalze() call. This is reqiured for correct lifecycle maintenance.
```
richardartoul: required*

storage/shard.go
```diff
@@ -990,31 +957,64 @@ func (s *dbShard) tryRetrieveWritableSeries(id ident.ID) (
...
-	return &dbShardEntry{series: series, index: uniqueIndex}, nil
-}
-
-func (s *dbShard) cloneTags(tags ident.TagIterator) (ident.Tags, error) {
+	// NB(r): As documented in storage/series.DatabaseSeries the series IDs
+	// are garbage collected, hence we cast the ID to a BytesID that can't be
+	// finalized.
+	// Since series are purged so infrequently the overhead of not releasing
+	// back an ID to a pool is amortized over a long period of time.
+	clonedID := ident.BytesID(append([]byte(nil), id.Bytes()...))
```
richardartoul: any reason to not call NoFinalize() here?

runtime/runtime_options_manager.go
```diff
@@ -74,3 +76,39 @@ func (w *optionsManager) RegisterListener(
 func (w *optionsManager) Close() {
 	w.watchable.Close()
 }
+
+// NewNoOpOptionsManager returns a no-op options manager that cannot
+// be updated and does not spawn backround goroutines (useful for globals
+// in test files).
+func NewNoOpOptionsManager(opts Options) OptionsManager {
```
prateek: lol this came from annoyance eh?

storage/shard_insert_queue_test.go
prateek: @robskillington still relevant?

storage/shard.go
prateek: ident.BytesID is a []byte, can't finalize it - https://github.com/m3db/m3x/blob/master/ident/bytes_id.go#L27-L55

Might be worth a comment indicating that here.

prateek: Or better yet, just make the NoFinalize call. It'll get compiled away anyway

storage/shard.go
```diff
@@ -990,31 +952,64 @@ func (s *dbShard) tryRetrieveWritableSeries(id ident.ID) (
...
+	clonedTags := s.identifierPool.Tags()
+
+	// Avoid finalizing the tags since series will let them be garbage collected
+	clonedTags.NoFinalize()
+
 	for tags.Next() {
 		t := tags.Current()
-		clone.Append(s.identifierPool.CloneTag(t))
+
+		// NB(r): Optimization for workloads that embed the tags in the ID is to
```
prateek: I still can't decide if this is brilliant or terrible.

storage/block/block.go
prateek: I think the removed the context all together.

storage/index/for_each_test.go
```diff
@@ -22,43 +22,54 @@ package index
...
 
+	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
 )
 
-func TestForEachBlockStarts(t *testing.T) {
-	now := time.Now().Truncate(time.Hour)
-	tn := func(n int64) xtime.UnixNano {
-		return xtime.ToUnixNano(now.Add(time.Duration(n) * time.Hour))
+func TestWriteBatchForEachUnmarkedBatchByBlockStart(t *testing.T) {
```
prateek: +1 for these tests

storage/index/types.go
```diff
@@ -183,42 +179,306 @@ type BlockTickResult struct {
...
+// with reference to a restricted set of the write batch for each unique block
+// start for entries that have not been marked completed yet.
+func (b *WriteBatch) ForEachUnmarkedBatchByBlockStart(
+	fn ForEachWriteBatchByBlockStartFn,
+) {
+	// Ensure sorted correctly first
+	b.SortByUnmarkedAndIndexBlockStart()
+
+	// What we do is a little funky but least alloc intensive, essentially we mutate
+	// this batch and then restore the pointers to the original docs after.
```
prateek: nit: could you add a comment to ForEachWriteBatchByBlockStartFn  indicating that this will break if `fn` does async operations on the batch

storage/index/types.go
```diff
@@ -183,42 +179,306 @@ type BlockTickResult struct {
...
+		startIdx       = 0
+		lastBlockStart xtime.UnixNano
 	)
-	for i := 0; i < len(w); i++ {
-		elem := w[i]
-		if elem.BlockStart != lastNanos {
-			lastNanos = elem.BlockStart
-			// We only want to call the the ForEachBlockStartFn once we have calculated the entire group,
+	for i := range allEntries {
+		if allEntries[i].OnIndexSeries == nil {
```
prateek: making sure i understand this - you can early terminate here because the sort above guarantees if the first entry is done, then all are done, yea?

storage/index/types.go
```diff
@@ -183,42 +179,306 @@ type BlockTickResult struct {
...
+			b.docs = allDocs[startIdx:i]
+			fn(prevLastBlockStart, b)
 			startIdx = i
 		}
 	}
+
 	// spill over
-	if startIdx < len(w) {
-		fn(w[startIdx].BlockStart, w[startIdx:])
+	if startIdx < len(allEntries) {
```
prateek: hm can you do this spill over un-conditionally? can't there be marked success entries in the back?

storage/index/types.go
```diff
@@ -183,42 +179,306 @@ type BlockTickResult struct {
...
+		b.docs[i] = docZeroed
+	}
+	b.docs = b.docs[:0]
+}
+
+// SortByUnmarkedAndIndexBlockStart sorts the batch by unmarked first and then
+// by index block start time.
+func (b *WriteBatch) SortByUnmarkedAndIndexBlockStart() {
+	b.sortBy = writeBatchSortByUnmarkedAndBlockStart
+	sort.Stable(b)
```
prateek: +1

storage/index/types.go
```diff
@@ -183,42 +179,306 @@ type BlockTickResult struct {
...
+			b.entries[idx].OnIndexSeries = nil
+			b.entries[idx].result = WriteBatchEntryResult{Err: nil}
+		}
+	}
+}
+
+// MarkUnmarkedEntriesError marks all unmarked entries as error.
+func (b *WriteBatch) MarkUnmarkedEntriesError(err error) {
+	for idx := range b.entries {
+		if b.entries[idx].OnIndexSeries != nil {
```
prateek: hm, can you reuse the method below, i.e.

``` 
func (b *WriteBatch) MarkUnmarkedEntriesError(err error) {
  for idx := range b.entries {
    b.MarkUnmarkedEntryError(err, idx)
  }
}
```

storage/series/buffer.go
```diff
@@ -680,9 +667,7 @@ func (b *dbBufferBucket) resetEncoders() {
 	for i := range b.encoders {
 		// Register when this bucket resets we close the encoder
 		encoder := b.encoders[i].encoder
-		if b.ctx != nil {
-			b.ctx.RegisterCloser(encoder)
-		}
+		encoder.Close()
```
prateek: do you need to take ctx as an arg in these methods?

storage/block/block.go
```diff
@@ -217,7 +214,6 @@ func (b *dbBlock) Stream(blocker context.Context) (xio.BlockReader, error) {
 		return xio.EmptyBlockReader, errReadFromClosedBlock
 	}
 
-	b.ctx.DependsOn(blocker)
```
prateek: this method too, can you just drop the context arg?

storage/index.go
```diff
@@ -207,156 +205,144 @@ func newNamespaceIndexWithOptions(
...
+			batch.SortByEnqueued()
+			return fmt.Errorf("check batch: %d insert errors", numErrs)
+		}
 	}
 
 	return nil
 }
 
-// NB: this function is called by the namespaceIndexInsertQueue.
+// WriteBatches is called by the indexInsertQueue.
```
prateek: could you drop the FOLLOWUP note below this line, considering you just did it :D

storage/index.go
```diff
@@ -207,156 +205,144 @@ func newNamespaceIndexWithOptions(
...
+	result, err := block.WriteBatch(batch)
+
 	// NB: we don't need to do anything to the OnIndexSeries refs in `inserts` at this point,
 	// the index.Block WriteBatch assumes responsibility for calling the appropriate methods.
 	if numErr := result.NumError; numErr != 0 {
 		i.metrics.AsyncInsertErrors.Inc(numErr)
 	}
 	if err != nil {
-		i.logger.Errorf("unable to write to index, dropping inserts. [%v]", err)
+		i.logger.Errorf("error writing to index block: %v", err)
```
prateek: super nit: could you filter out any `ErrDuplicateID` from here?

storage/series/lookup/entry_blackbox_test.go
```diff
@@ -0,0 +1,112 @@
...
+// THE SOFTWARE.
+
+package lookup_test
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/fortytw2/leaktest"
```
prateek: nit: import order

storage/series/lookup/entry_whitebox_test.go
```diff
@@ -0,0 +1,55 @@
...
+// THE SOFTWARE.
+
+package lookup
+
+import (
+	"testing"
+	"time"
+
+	xtime "github.com/m3db/m3x/time"
+	"github.com/stretchr/testify/require"
```
prateek: nit: import order

storage/block/block.go
robskillington: I saw that, yeah wasn't 100% sure but it may have been a double.  Either way we can get rid of that complexity with this change thankfully.

storage/block/block.go
robskillington: Moved the RegisterFinalizer?  We don't use the context anymore for finalization (block no longer has a context even) on the segment so this didn't get moved anywhere per se.  Or do you mean something else?

storage/block/block.go
robskillington: Yup context is gone, we now just copy the bytes each time rather than depending on the caller's context and taking ref.

storage/index.go
robskillington: Ta, will update.

storage/index.go
robskillington: Sure thing.

storage/series/lookup/entry.go
robskillington: It was, but I can update.

storage/series/lookup/entry.go
robskillington: Good catch, ta.

storage/shard.go

runtime/runtime_options_manager.go
robskillington: Yeah `leaktest.(...)` started going off 100% each test run due to a lingering goroutine from `func init` in the `storage` package that registered a listener that never closes.  To avoid that weirdness put this together.

storage/shard.go
robskillington: Heheh, talked with Richie about it and agreed we should flag this with config but default it to on.

storage/index/types.go
robskillington: Sure thing, sounds good.

storage/series/buffer.go
robskillington: When we call `encoder.Stream()` we take a copy of the bytes and pass back a new Segment wrapping them, so don't need to involve ctx at all no, once we have the write lock on the series (which we have here) its safe to close.

storage/shard.go
robskillington: I'll add this.

storage/index/types.go
robskillington: Good call, will do.

storage/index/types.go
robskillington: Right yeah.

storage/index/types.go
robskillington: So you only exit the loop if you haven't hit a "done" element yet, which means all the remaining entries haven't been marked for error or success yet (thanks to the sort order).  I can add a comment to this effect perhaps?

storage/block/block.go
robskillington: We need it for if it goes to the retriever (which takes a ctx to register some finalizations).

storage/index.go
robskillington: Sure thing.

storage/index.go
robskillington: I see you are doing this =]

storage/series/lookup/entry_blackbox_test.go
robskillington: Sure thing.

storage/series/lookup/entry_whitebox_test.go
robskillington: Sure thing.

runtime/runtime_options_manager.go
prateek: much appreciated. i've always chased the different leaks, this is cleaner

storage/index/types.go
prateek: yep sounds good

prateek:
LGTM

PR #610, opened 2018-05-10
Fix NPE for static configs

justinjc:


environment/config.go
```diff
@@ -257,18 +256,11 @@ func (c Configuration) configureStatic(configSvcClient client.Client, cfgParams
...
-	}
-
-	configureResults := ConfigureResults{
+	return ConfigureResults{
 		NamespaceInitializer: nsInitStatic,
 		TopologyInitializer:  topoInit,
-		KVStore:              kv,
-	}
-	return configureResults, nil
+		KVStore:              m3clusterkvmem.NewStore(),
```
prateek: what needs this value to be set?

justinjc: `m3cluster/kv/mem` is basically a fake in-memory KV to satisfy the KV interfaces. Values are used from this fake store e.g. for topology initialization, but initializers will populate values accordingly e.g. [here](https://github.com/m3db/m3db/blob/d776da52cff0331557137da4b4e81dc0d73bc5dc/topology/static.go#L54:6) when values are 'watched'.

prateek: 👍

environment/config_test.go
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
prateek: s/2017/2018

PR #613, opened 2018-05-10
Cache commitlog series metadata in-between calls to ReadData() and ReadIndex()

richardartoul:


storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -181,13 +187,18 @@ func (s *commitLogSource) ReadData(
 	wg.Wait()
 	s.logEncodingOutcome(workerErrs, iter)
 
-	return s.mergeShards(int(numShards), bopts, blockSize, blopts, encoderPool, unmerged), nil
+	result := s.mergeShards(int(numShards), bopts, blockSize, blopts, encoderPool, unmerged)
+	// After merging shards, its safe to cache the shardData (which involves some mutation).
+	if s.opts.CacheSeriesMetadata() {
```
robskillington: Hm instead of making it an option, its's probably best to infer this from `nsMetadata.Options().IndexOptions().Enabled()`.  Otherwise its an extra flip that you pretty much always want to use.

richardartoul: Thats a good idea, I'll add that as an additional check, but I think the separate config is still good in-case its using too much memory, or we want to just debug this feature on/off.

Also you just made me realize I should probably segregate the cached data by namespace...

richardartoul: done and done

services/m3dbnode/config/config_test.go
```diff
@@ -370,6 +370,7 @@ bootstrap:
   fs:
     numProcessorsPerCPU: 0.125
   peers: null
+  cacheSeriesMetadata: null
```
robskillington: Want to also update the example config?  `./example/m3db-node-config.yaml` I think it is

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -59,10 +65,11 @@ type encoder struct {
...
 	return &commitLogSource{
-		opts:          opts,
-		inspection:    inspection,
-		log:           opts.ResultOptions().InstrumentOptions().Logger(),
-		newIteratorFn: commitlog.NewIterator,
+		opts:            opts,
+		inspection:      inspection,
+		log:             opts.ResultOptions().InstrumentOptions().Logger(),
+		newIteratorFn:   commitlog.NewIterator,
+		cachedShardData: map[string][]shardData{},
```
robskillington: nit: Perhaps it might be nice to wrap up the `[]shardData` in an encapsulating type?  i.e. `map[string]*cachedShardData` just so you can encapsulate other stuff in there and you don't have to do the weird append thing (i.e. you can append to a field in that type rather than append to the `mapName[key]`)

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -181,18 +186,26 @@ func (s *commitLogSource) ReadData(
 	wg.Wait()
 	s.logEncodingOutcome(workerErrs, iter)
 
-	return s.mergeShards(int(numShards), bopts, blockSize, blopts, encoderPool, unmerged), nil
+	result := s.mergeShards(int(numShards), bopts, blockSize, blopts, encoderPool, unmerged)
+	// After merging shards, its safe to cache the shardData (which involves some mutation).
+	if s.shouldCacheSeriesMetadata(runOpts, ns) {
+		s.cacheShardData(ns, unmerged)
```
robskillington: Wouldn't we want to cache the merged not the unmerged?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -575,38 +683,75 @@ func (s *commitLogSource) ReadIndex(
...
+		return nil
 	}
 
-	return indexResult, nil
+	segment, err := indexResults.GetOrAddSegment(blockStart, indexOptions, resultOptions)
+	if err != nil {
+		return err
+	}
+
+	exists, err := segment.ContainsID(id.Data().Bytes())
```
robskillington: You'll have a conflict, make sure to rebase on master (`Data().Bytes()` no longer exists and you need to do just `Bytes()`).

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -575,38 +683,75 @@ func (s *commitLogSource) ReadIndex(
...
+		return err
+	}
+	if exists {
+		return nil
+	}
+
+	// We can use the NoClone variant here because the cached IDs/Tags are marked NoFinalize
+	// by the ReadData() path when it reads from the commitlog files, and the IDs/Tags read
+	// from the commit log files by the ReadIndex() method won't be finalized because this
+	// code path doesn't finalize them.
```
robskillington: Win 🎈

storage/bootstrap/provider_options.go
```diff
@@ -0,0 +1,48 @@
...
+	// cache series metadata between runs.
+	defaultCacheSeriesMetadata = true
+)
+
+type processProviderOptions struct {
+	cacheSeriesMetadata bool
+}
+
+// NewProcessProviderOptions creates new bootstrap run options
+func NewProcessProviderOptions() ProcessProviderOptions {
```
robskillington: nit: Maybe we should call these ProcessOptions?  They just so happen to be passed to the "Provider" for the "Process". (We're kind of using "Provider" to mean "Factory" here, but we don't want to sound Java-y, it's a little silly perhaps).

storage/bootstrap/bootstrapper/commitlog/source.go
robskillington: Never mind, I get it, the encoders are being removed anyhow

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -509,6 +537,85 @@ func (s *commitLogSource) logMergeShardsOutcome(shardErrs []int, shardEmptyErrs
...
+				// so that they can be GC'd.
+				seriesData.encoders[blockStart] = nil
+			}
+		}
+
+		if shard >= len(cachedShardData) {
+			// Extend the slice if necessary (could happen if different calls to
+			// ReadData() bootstrap different shards.)
+			for len(cachedShardData) != shard+1 {
+				cachedShardData = append(cachedShardData, shardData{})
```
robskillington: This may be a bug yeah?

richardartoul: Oh I get it now. Ok cheers

storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

storage/index/convert/convert.go
```diff
@@ -77,6 +77,24 @@ func FromMetric(id ident.ID, tags ident.Tags) (doc.Document, error) {
 	}, nil
 }
 
+// FromMetricNoClone converts the provided metric id+tags into a document without cloning.
+func FromMetricNoClone(id ident.ID, tags ident.Tags) (doc.Document, error) {
```
prateek: there's similar code in the shard too. would be good to consolidate into a single place. How about `storage/index/convert` ? 

/cc @robskillington

storage/bootstrap/types.go
```diff
@@ -66,6 +66,17 @@ type TargetRange struct {
 	RunOptions RunOptions
 }
 
+// ProcessOptions is a set of options for a bootstrap provider.
+type ProcessOptions interface {
```
prateek: nit: should this `ProcessProviderOptions`?

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -87,14 +94,19 @@ func (s *commitLogSource) AvailableData(
...
+	runOpts bootstrap.RunOptions,
 ) (result.DataBootstrapResult, error) {
 	if shardsTimeRanges.IsEmpty() {
 		return result.NewDataBootstrapResult(), nil
 	}
 
 	readCommitLogPredicate := newReadCommitLogPredicate(
 		ns, shardsTimeRanges, s.opts, s.inspection)
+
+	// TODO(rartoul): When we implement caching data across namespaces, this will need
```
prateek: +1

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -656,12 +810,12 @@ func newReadSeriesPredicate(ns namespace.Metadata) commitlog.SeriesFilterPredica
...
-type encodersAndRanges struct {
-	encodersBySeries map[uint64]encodersByTime
-	ranges           xtime.Ranges
+type shardData struct {
+	series map[uint64]metadataAndEncodersByTime
+	ranges xtime.Ranges
 }
 
-type encodersByTime struct {
+type metadataAndEncodersByTime struct {
```
prateek: +1

services/m3dbnode/config/bootstrap.go
```diff
@@ -53,6 +53,10 @@ type BootstrapConfiguration struct {
 
 	// Peers bootstrapper configuration.
 	Peers *BootstrapPeersConfiguration `yaml:"peers"`
+
+	// CacheSeriesMetadata determines whether individual bootstrappers cache
+	// series metadata across all calls (namespaces / shards / blocks).
+	CacheSeriesMetadata *bool `yaml:"cacheSeriesMetadata"`
```
prateek: Hm should this renamed to be only for the commit log bootstrapper? I don't think we'd do the caching for data in peers/fs as things stand now.

storage/index/convert/convert.go
prateek: lol doh. nvm i see the filename

prateek: Maybe make the shard code re-use this too

storage/bootstrap/types.go
richardartoul: Lol I just renamed it from that @robskillington request

services/m3dbnode/config/bootstrap.go
richardartoul: Made this change (from being at the commit log bootstrapper level) at @robskillington request. Seemed reasonable so that if we implement those we don't need to add additional config

richardartoul: (I think we will want to do it at the peer level at least)

storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -509,6 +537,89 @@ func (s *commitLogSource) logMergeShardsOutcome(shardErrs []int, shardEmptyErrs
...
+	for shard, currShardData := range allShardData {
+		for _, seriesData := range currShardData.series {
+			for blockStart := range seriesData.encoders {
+				// Nil out any references to the encoders (which should be closed already anyways),
+				// so that they can be GC'd.
+				seriesData.encoders[blockStart] = nil
+			}
+		}
+
+		if shard >= len(nsCache.shardData) {
```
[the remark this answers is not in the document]
richardartoul: nice

storage/index/convert/convert.go
richardartoul: @prateek The shard logic has the special tag subslice stuff...should we move that logic into the helpers so its not only on series insertion? Not sure about doing that in this P.R

prateek: Yeah it'd be good have all code-paths re-use the logic. Don't think there's anything specific to shard/series in that code. Feel free to do in a subsequent PR if that's easier.

PR #616, opened 2018-05-11
[index] Use m3ninx methods to serialize and deserialize queries directly

### LINKED ISSUE #558: Index Query RPC simplification
The current version of code for FetchTagged/FetchTaggedIDs performs the RPC serialisation for the `m3ninx/idx/Query` type by hand (defining thrift types, and Go<->Thrift converters). This is bad for a few reasons:
(1) Forces m3ninx to expose library internals it would rather otherwise hide
(2) Couples m3db/m3ninx code beyond what is necessary. 
(3) Makes pooling for client request/server types much harder. 

Once https://github.com/m3db/m3ninx/issues/42 lands, we should migrate to using it instead.

jeromefroe:
This PR updates the RPC layer to use m3ninx's methods to Marshal and Unmarshal queries instead of representing the queries in m3db itself. It resolves #558.

### DIFF
--- a/generated/thrift/rpc.thrift
+++ b/generated/thrift/rpc.thrift
@@ -144,20 +144,7 @@ struct FetchTaggedRequest {
 }
 
 struct IdxQuery {
-	1: required BooleanOperator operator
-	2: optional list<IdxTagFilter> filters
-	3: optional list<IdxQuery> subQueries
-}
-
-enum BooleanOperator {
-	AND_OPERATOR,
-}
-
-struct IdxTagFilter {
-	1: required binary tagName
-	2: required binary tagValueFilter
-	3: required bool negate
-	4: required bool regexp
+	1: required binary query
 }
 
 struct FetchTaggedResult {
--- a/generated/thrift/rpc.thrift
+++ b/generated/thrift/rpc.thrift
@@ -136,15 +136,11 @@ struct Segment {
 
 struct FetchTaggedRequest {
 	1: required binary nameSpace
-	2: required IdxQuery query
+	2: required binary query
 	3: required i64 rangeStart
 	4: required i64 rangeEnd
 	5: required bool fetchData
 	6: optional i64 limit
-}
-
-struct IdxQuery {
-	1: required binary query
 }
 
 struct FetchTaggedResult {
--- a/network/server/tchannelthrift/convert/convert.go
+++ b/network/server/tchannelthrift/convert/convert.go
@@ -22,7 +22,6 @@ package convert
 
 import (
 	"errors"
-	"fmt"
 	"time"
 
 	"github.com/m3db/m3db/digest"
@@ -223,7 +222,7 @@ func FromRPCFetchTaggedRequest(
 		opts.Limit = int(*l)
 	}
 
-	q, err := fromRPCQuery(req.Query)
+	q, err := idx.Unmarshal(req.Query)
 	if err != nil {
 		return nil, index.Query{}, index.QueryOptions{}, false, err
 	}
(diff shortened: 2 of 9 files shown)


generated/thrift/rpc.thrift
```diff
@@ -144,20 +144,7 @@ struct FetchTaggedRequest {
...
-enum BooleanOperator {
-	AND_OPERATOR,
-}
-
-struct IdxTagFilter {
-	1: required binary tagName
-	2: required binary tagValueFilter
-	3: required bool negate
-	4: required bool regexp
+	1: required binary query
```
prateek: do you even need the top level type anymore? like just move this up to `FetchTaggedRequest` and completely ignore

network/server/tchannelthrift/convert/convert.go
```diff
@@ -286,106 +281,18 @@ func fromRPCQuery(rpcQuery *rpc.IdxQuery) (idx.Query, error) {
...
-			if err != nil {
-				return idx.Query{}, err
-			}
-			queries = append(queries, cq)
-		}
-		return idx.NewConjunctionQuery(queries...), nil
-	}
-
-	return idx.Query{}, fmt.Errorf("at least one Filter/Sub-Query must be defined")
+	return idx.Unmarshal(rpcQuery.Query)
```
prateek: love it!

network/server/tchannelthrift/convert/convert.go
```diff
@@ -260,7 +255,7 @@ func ToRPCFetchTaggedRequest(
 		return rpc.FetchTaggedRequest{}, tsErr
 	}
 
-	query, queryErr := toRPCQuery(q.SearchQuery())
+	query, queryErr := toRPCQuery(q.Query)
```
prateek: can get rid of `SearchQuery()` upstream once this lands yea?

network/server/tchannelthrift/node/service_test.go
```diff
@@ -668,7 +668,7 @@ func TestServiceFetchTagged(t *testing.T) {
 
 	req, err := idx.NewRegexpQuery([]byte("foo"), []byte("b.*"))
 	require.NoError(t, err)
-	qry := index.Query{req}
+	qry := index.Query{Query: req}
```
prateek: more involved change but could make index.Query a type alias if you wanted.

i.e.

```m3db/m3db/storage/index/types.go

import ".../m3ninx/idx"

type Query idx.Query
```

Thoughts?

network/server/tchannelthrift/convert/convert.go
prateek: ah nvm, need it for the executor i see

generated/thrift/rpc.thrift
jeromefroe: sounds good to me!

network/server/tchannelthrift/convert/convert.go
jeromefroe: yea, once we have the external facing index interface we should be good tho

network/server/tchannelthrift/node/service_test.go
jeromefroe: I like the idea, I'll test it out to see how involved it is

jeromefroe: I only made this change because my IDE was complaining at me lol

jeromefroe: I looked into this, but I think it's more trouble than it's worth because if we define `Query` as `type Query idx.Query` then `Query` won't inherit any of the methods of `idx.Query` and we'll need to type cast it whenever we need an `idx.Query`.

network/server/tchannelthrift/convert/convert.go
prateek: Yep sounds good.

network/server/tchannelthrift/node/service_test.go
prateek: Yeah fair. Happy to leave as is.

PR #625, opened 2018-05-12
Update FS segment writer to use io.Writer instead of implementing io.Reader

robskillington:


persist/fs/types.go
```diff
@@ -307,6 +307,16 @@ func (t IndexSegmentFileType) Validate() error {
...
 }
 
+// IndexSegmentFileSetWriter is an index segment file set writer.
+type IndexSegmentFileSetWriter interface {
+	SegmentType() IndexSegmentType
+	MajorVersion() int
+	MinorVersion() int
+	SegmentMetadata() []byte
+	Files() []IndexSegmentFileType
+	WriteFile(fileType IndexSegmentFileType, writer io.Writer) error
```
prateek: Maybe just list the guarantee that this is called on the files in the same order as the values returned from `Files()`

robskillington: Sure thing.

PR #626, opened 2018-05-13
Pull m3x and use new Tags type

robskillington:


storage/block/result.go
```diff
@@ -216,8 +223,19 @@ func (it *filteredBlocksMetadataIter) Next() bool {
...
+	if tagsIter := it.res[it.resIdx].Tags; tagsIter != nil {
+		for tagsIter.Next() {
+			curr := tagsIter.Current()
+			tags.Append(ident.StringTag(curr.Name.String(), curr.Value.String()))
+		}
+		if err := tagsIter.Err(); err != nil {
+			it.err = err
+			return false
+		}
+		tagsIter.Close()
```
prateek: do you need to put this in the early terminate return too?

storage/shard.go
```diff
@@ -1533,10 +1533,6 @@ func (s *dbShard) FetchBlocksMetadataV2(
 					blockStart, err)
 			}
 
-			// Make sure ID and tags get cleaned up after read is done
-			ctx.RegisterFinalizer(id)
```
prateek: hm is this because the id/tags are set to NonFinalizable anyway?

storage/shard.go
```diff
@@ -1613,15 +1609,22 @@ func (s *dbShard) Bootstrap(
 		if entry == nil {
 			// Synchronously insert to avoid waiting for
 			// the insert queue potential delayed insert
-			entry, err = s.insertSeriesSync(dbBlocks.ID,
-				ident.NewTagSliceIterator(dbBlocks.Tags),
+			tagsIter := s.identifierPool.TagsIterator()
```
prateek: hm doesn't insertSeriesSync do this anyway (via `newShardEntry(...)`) ?

storage/shard.go
```diff
@@ -1613,15 +1609,22 @@ func (s *dbShard) Bootstrap(
...
 			if err != nil {
 				multiErr = multiErr.Add(err)
 				continue
 			}
 		}
 
+		// No longer require tags as we copy them in insert series sync
+		// or if we found the series then we don't require them for insertion
+		// at all
+		dbBlocks.Tags.Finalize()
```
prateek: why not the id too?

storage/shard.go
robskillington: This is because the series ID and Tags can both be safely finalized as they are non-finalizeable, so we just finalize when we close the MetadataResult now rather than using the context array.

storage/shard.go
robskillington: Yes, unfortunately `newShardEntry(...)` takes a `TagIterator` though, so we need to wrap the bootstrapped `ident.Tags` in a `TagIterator` to pass to it.

storage/shard.go
robskillington: Unfortunately in some cases the ID is kept ref to (with series cache policy `all_metadata`).

storage/block/result.go
robskillington: I'm fine with leaking things in the extreme rare error case if it makes the code more simple.

prateek: fair enough

storage/shard.go
prateek: sigh. mind leaving a `FOLLOW(prateek)` to address

storage/shard.go
prateek: Could you put down a note explaining that, and that we can delete it once we trash that policy.

storage/shard.go

PR #631, opened 2018-05-14
Expose session iterator pools

arnikola:


client/fetch_state.go
```diff
@@ -171,12 +171,18 @@ func (f *fetchState) asEncodingSeriesIterators(pools fetchTaggedPools) (encoding
...
-	MultiReaderIterator() encoding.MultiReaderIteratorPool
-	SeriesIterator() encoding.SeriesIteratorPool
 	MutableSeriesIterators() encoding.MutableSeriesIteratorsPool
 	MultiReaderIteratorArray() encoding.MultiReaderIteratorArrayPool
 	ID() ident.Pool
 	CheckedBytesWrapper() xpool.CheckedBytesWrapperPool
 	TagDecoder() serialize.TagDecoderPool
 }
+
+// IteratorPools exposes a small subset of iterator pools that are sufficient for clients
```
prateek: nit: move this to `session_pools.go`

prateek: or better yet, maybe just to `encoding/types.go`?

client/session.go
```diff
@@ -1521,6 +1521,10 @@ func (s *session) readConsistencyResult(
 	return nil
 }
 
+func (s *session) IteratorPools() (IteratorPools, error) {
+	return s.pools, nil
```
robskillington: Might want to check that `s.pools != nil`?  Also are the pools always guaranteed to be available (I can't remember when they get alloc'd).  If not we probably have to nil check the specific pools in question too =[

client/session.go
```diff
@@ -1521,6 +1521,10 @@ func (s *session) readConsistencyResult(
 	return nil
 }
 
+func (s *session) IteratorPools() (encoding.IteratorPools, error) {
+	return s.pools, nil
```
robskillington: Might want to check that `s.pools != nil`?  Also are the pools always guaranteed to be available (I can't remember when they get alloc'd).  If not we probably have to nil check the specific pools in question too and return an error if nil perhaps?

client/session.go
arnikola: s.pools is a `sessionPools` struct, and isn't itself nilable.

For the pools themselves, I believe they get alloced in `session.Open()`, so I think it should be safe unless you get the `session` in some weird way. Adding a sanity check would make sense, though.

prateek: maybe ensure the session is open then?

```
	s.state.RLock()
        defer s.state.RULock()
	if s.state.status != statusOpen {
		return nil, errSessionStatusNotOpen
	}
        return s.pools
```

PR #633, opened 2018-05-14
Make coordinator make-able again

justinjc:
This will allow `make m3coordinator` and `make test-internal` to work after cd-ing into `src/coordinator` so that local dev work is easier.

### DIFF
--- a/src/coordinator/Makefile
+++ b/src/coordinator/Makefile
@@ -130,7 +130,7 @@ metalint: install-metalinter install-linter-badtime install-linter-importorder
 .PHONY: test-internal
 test-internal:
 	@which go-junit-report > /dev/null || go get -u github.com/sectioneight/go-junit-report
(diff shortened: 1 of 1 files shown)


src/coordinator/Makefile
```diff
@@ -129,7 +130,7 @@ metalint: install-metalinter install-linter-badtime install-linter-importorder
 .PHONY: test-internal
 test-internal:
 	@which go-junit-report > /dev/null || go get -u github.com/sectioneight/go-junit-report
-	@$(VENDOR_ENV) $(test) $(coverfile) | tee $(test_log)
+	@$(VENDOR_ENV) $(test) $(coverfile) "" "" $(CI_DIR) | tee $(test_log)
```
prateek: kinda shitty to pass `""`, can you provide a value?

PR #636, opened 2018-05-14
Undo index<->namespace block restrictions

prateek:
Will need to do the n-way expansion during the bootstrap result construction as a result of this change. 

Need it for having the flexibility to make index blocks smaller based on memory constraints.

### DIFF
--- a/storage/namespace/index_options.go
+++ b/storage/namespace/index_options.go
@@ -21,6 +21,7 @@
 package namespace
 
 import (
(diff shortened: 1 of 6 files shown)


robskillington:
We may want to avoid this based on talks that the ID and tags data will need to stick around anyway so having a smaller than data block size would never really help much ya?

prateek:
Yep, closing per offline chat

PR #638, opened 2018-05-15
Persist index fs.Segments

prateek:


persist/fs/persist_manager.go
```diff
@@ -350,3 +566,21 @@ func (pm *persistManager) SetRuntimeOptions(value runtime.Options) {
 	pm.currRateLimitOpts = value.PersistRateLimitOptions()
 	pm.Unlock()
 }
+
```
robskillington: nit: Do you mind deleting the `ensure...` methods here, as discussed these are probably fine since the glob path will succeed when the dir isn't there but just return empty results (which its totally fine).

storage/index/block.go
```diff
@@ -87,8 +95,13 @@ func NewBlock(
 	}
 
 	b := &block{
-		state:     blockStateOpen,
-		segment:   seg,
+		state:         blockStateOpen,
+		needsFlush:    true,
```
robskillington: Can we make this a computed function that returns true/false?  i.e. inspecting inactiveMutableSegments and the activeSegment, etc.

storage/index/block.go
```diff
@@ -87,8 +95,13 @@ func NewBlock(
...
+		activeSegment: seg,
+		writeBatch: m3ninxindex.Batch{
+			AllowPartialUpdates: true,
+		},
```
robskillington: Don't need writeBatch anymore.

persist/fs/persist_manager.go
```diff
@@ -115,84 +143,201 @@ func NewPersistManager(opts Options) (persist.Manager, error) {
...
+	}
+
+	// indicate we don't need the defer() above to release any resources, as we are
+	// transferring ownership to the caller.
+	success = true
+
+	// return ownership of the segments, and the reader via a PersistedSegments() object.
+	return &persistedSegments{
+		segs:   segments,
+		closer: reader,
```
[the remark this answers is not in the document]
prateek: ta cheers, addressed https://github.com/m3db/m3ninx/pull/82

persist/fs/persist_manager.go
prateek: sure thing

storage/index/block.go
prateek: yeah good call, makes the code a lot simpler

storage/index/block.go
prateek: thanks for the catch. rebase bug

prateek: removed.

storage/index/block.go
prateek: done

persist/fs/persist_manager.go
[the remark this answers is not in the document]
prateek: done

persist/fs/persist_manager.go
prateek: done

persist/fs/persist_manager.go
```diff
@@ -115,84 +143,216 @@ func NewPersistManager(opts Options) (persist.Manager, error) {
...
+	// so we can close the writer.
+	if err := pm.indexPM.writer.Close(); err != nil {
+		return nil, err
+	}
 
-	if pm.status != persistManagerIdle {
-		return nil, errPersistManagerNotIdle
+	// only attempt to retrieve data if we have not encountered errors during
+	// any writes.
+	if err := pm.indexPM.writeErr; err != nil {
```
robskillington: Nice 👍

storage/index.go
```diff
@@ -365,7 +367,12 @@ func (i *nsIndex) writeBatchForBlockStartWithRLock(
 		// repeated inserts. as long as a block has an ID, it's not an error so we don't need
 		// to pollute the logs with these messages.
 		if partialError, ok := err.(*m3ninxindex.BatchPartialError); ok {
-			err = partialError.FilterDuplicateIDErrors()
+			partialError = partialError.FilterDuplicateIDErrors()
+			if partialError == nil {
+				err = nil
+			} else {
+				err = partialError
```
robskillington: Hm curious, why the if/else here?  If it's not it sets err to itself, if nil it explicitly sets err to nil - isn't the old way and this way equivalent?  (I'm sure its not, maybe we need to add a comment here?)

storage/index/block.go
```diff
@@ -252,38 +284,163 @@ func (b *block) Query(
...
+	}
+
+	// swap mutable segments with the persisted segments
+	b.replaceMutableSegmentsWithRLock(persistedSegments)
+
+	// all good!
 	return nil
 }
 
+func (b *block) replaceMutableSegmentsWithRLock(persistedSegments []segment.Segment) {
```
robskillington: This is nice 👍

storage/namespace.go
```diff
@@ -849,6 +851,34 @@ func (n *dbNamespace) Flush(
...
+		return errNamespaceNotBootstrapped
+	}
+	n.RUnlock()
+
+	if !n.nopts.FlushEnabled() {
+		n.metrics.flush.ReportSuccess(n.nowFn().Sub(callStart))
+		return nil
+	}
+
+	if !n.nopts.IndexOptions().Enabled() {
```
robskillington: nit: Can combine these two cases together perhaps? `if x || y` ?

persist/fs/files.go
```diff
@@ -458,6 +458,33 @@ func FileSetAt(filePathPrefix string, namespace ident.ID, shard uint32, blockSta
 	return FileSetFile{}, false, nil
 }
 
+// IndexFileSetsAt returns all FileSetFile(s) for the given namespace/blockStart combination.
+func IndexFileSetsAt(filePathPrefix string, namespace ident.ID, blockStart time.Time) (FileSetFilesSlice, error) {
```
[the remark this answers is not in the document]
prateek: Sure thing. Added a comment to be explicit about it.

storage/flush.go
```diff
@@ -137,8 +140,31 @@ func (m *flushManager) Flush(
...
+
+	// flush index data
+	// create index-flusher
+	indexFlush, err := m.pm.StartIndexPersist()
+	if err != nil {
+		multiErr = multiErr.Add(err)
+		return multiErr.FinalError()
+	}
+
+	m.setState(flushManagerIndexFlushInProgress)
```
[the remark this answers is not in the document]
prateek: Yep it's hard to see in the diff cause of the folds.

storage/namespace.go
prateek: sure thing

storage/index.go
prateek: Yeah it's weird. Arsies due to how interprets the return type of the Filter method. Made https://github.com/m3db/m3ninx/pull/83/files to avoid

storage/index/block.go
prateek: cheers!

storage/namespace.go

persist/fs/files.go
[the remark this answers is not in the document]
prateek: done

PR #640, opened 2018-05-15
Check topology during calls to AvailableData and AvailableIndex in peers bootstrapper

richardartoul:
If we implement checks in AvailableData() and AvailableIndex, we can make it such that the peer bootstrapper is skipped automatically in single-node deployments. This will allow us to make the default bootstrapping configuration:

FS --> Peers --> Commitlog --> Noop-All

And this configuration will work for any cluster size / topology and do "the right thing" without operators having to think about bootstrap configuration.


integration/peers_bootstrap_none_available_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/m3db/m3db/topology"
+	"github.com/m3db/m3db/topology/testutil"
+	xlog "github.com/m3db/m3x/log"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestPeersBootstrapNoneAvailable makes sure that a cluster with the peer bootstrapper enabled can still turn on
+// if none of the nodes peers shard states are available or leaving.
+func TestPeersBootstrapNoneAvailable(t *testing.T) {
```
robskillington: Nice, did we confirm whether or not a new cluster starts with INITIALIZING shards or AVAILABLE shards?

robskillington: Damn, think they start as available.  We may have to make a change to m3cluster:

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -654,6 +671,90 @@ func (s *peersSource) readBlockMetadataAndIndex(
...
+				// taken ownership of the shard.
+			case shard.Unknown:
+
+				// Success cases - We can bootstrap from this host, which is enough to
+				// mark this shard as bootstrappable.
+			case shard.Leaving:
+				fallthrough
+			case shard.Available:
+				shardPeers.numAvailablePeers++
+				break
```
robskillington: I don't think you need an explicit break here, should break correctly without falling through unless explicitly falling through.

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -677,3 +778,54 @@ func (s *peersSource) markIndexResultErrorAsUnfulfilled(
...
+
+	return topologyState, nil
+}
+
+type topologyState struct {
+	majorityReplicas int
+	shardStates      shardStates
+}
+
+type shardStates map[uint32]map[string]hostShardState
```
robskillington: very nit: Maybe to make this slightly more readable you could `type hostID string` and `type shardID uint32` so it becomes `map[shardID]map[hostID]hostShardState`?

topology/consistency_level.go
```diff
@@ -307,3 +307,68 @@ const (
 	majority         = "majority"
 	unstrictMajority = "unstrict_majority"
 )
+
+// WriteConsistencyAchieved returns a bool indicating whether or not we've received enough
```
robskillington: +1 for having these available from the `topology` package.

topology/consistency_level.go
```diff
@@ -307,3 +307,68 @@ const (
...
+		}
+	case ConsistencyLevelMajority:
+		if numSuccess >= majority { // Meets majority
+			return true
+		}
+	case ConsistencyLevelOne:
+		if numSuccess > 0 { // Meets one
+			return true
+		}
+	default:
```
prateek: unrelated to your change, but mind dropping the default and changing the return false to a panic (same as two functions below)

topology/consistency_level.go
prateek: Yeah good call on this.

integration/integration.go
```diff
@@ -117,10 +118,17 @@ func newMultiAddrAdminClient(
 	clientOpts := adminOpts.
 		SetOrigin(origin).
 		SetInstrumentOptions(instrumentOpts).
-		SetTopologyInitializer(topology.NewStaticInitializer(staticOptions)).
 		SetClusterConnectConsistencyLevel(topology.ConnectConsistencyLevelAny).
 		SetClusterConnectTimeout(time.Second)
 
+	if topologyInitializer != nil {
```
prateek: nit: can simplify this 
```
if topologyInitialzer == nil {
  topologyInitializer = topology.NewStaticInitializer(staticOptions)
}
clientOpts = clientOpts.SetTopologyInitializer(topologyInitializer)
```

storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -25,6 +25,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/m3db/m3cluster/shard"
```
prateek: nit: import order

integration/peers_bootstrap_none_available_test.go
richardartoul: Yeah can probably still land this, and then make that change and update M3DB. The comment in that file seems like its more complicated than just changing the default value, will it affect our client as well?

storage/bootstrap/bootstrapper/peers/source.go
richardartoul: sure

integration/integration.go
richardartoul: good call

src/dbnode/generated/proto/index/index.pb.go
```diff
@@ -59,9 +59,9 @@ type IndexInfo struct {
 	BlockStart   int64          `protobuf:"varint,2,opt,name=blockStart,proto3" json:"blockStart,omitempty"`
 	BlockSize    int64          `protobuf:"varint,3,opt,name=blockSize,proto3" json:"blockSize,omitempty"`
 	FileType     int64          `protobuf:"varint,4,opt,name=fileType,proto3" json:"fileType,omitempty"`
-	SnapshotTime int64          `protobuf:"varint,5,opt,name=snapshotTime,proto3" json:"snapshotTime,omitempty"`
-	Segments     []*SegmentInfo `protobuf:"bytes,6,rep,name=segments" json:"segments,omitempty"`
-	Shards       []uint32       `protobuf:"varint,7,rep,packed,name=shards" json:"shards,omitempty"`
+	Shards       []uint32       `protobuf:"varint,5,rep,packed,name=shards" json:"shards,omitempty"`
```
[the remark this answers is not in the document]
richardartoul: it actually turns out these files were stale for some reason :/ My P.R is fixing them and putting them in the right order

PR #644, opened 2018-05-16
Add subscope for namespace index

robskillington:


storage/index.go
```diff
@@ -166,6 +166,11 @@ func newNamespaceIndexWithOptions(
 		return nil, err
 	}
 
+	scope := opts.InstrumentOptions().MetricsScope().
+		SubScope("dbindex")
```
prateek: could you add a tag for namespace name here too

robskillington: Ah yes, sure thing.

PR #646, opened 2018-05-16
Emit metric every time a new encoder is created

prateek:


storage/series/types.go
```diff
@@ -251,4 +253,29 @@ type Options interface {
...
 	IdentifierPool() ident.Pool
+
+	// SetStats sets the configured Stats.
+	SetStats(value Stats) Options
+
+	// Stats returns the configured Stats.
+	Stats() Stats
+}
+
+// Stats is passed down from above to avoid constructing tally constructs per series/buffer.
```
richardartoul: "passed down from above" sounds like scripture haha, can you change this to "passed down from the namespace/shard objects to avoid an allocation per series" or something like that

storage/series/types.go
```diff
@@ -251,4 +253,29 @@ type Options interface {
...
+// NewStats returns a new Stats for the provided scope.
+func NewStats(scope tally.Scope) Stats {
+	return Stats{
+		EncoderCreated: scope.Counter("encoder-created"),
+	}
+}
+
+// IncCreatedEncoders incs the EncoderCreated stat.
+func (s Stats) IncCreatedEncoders() {
+	if int(time.Now().UnixNano())%100 == 0 {
```
richardartoul: Don't we have the ability to set sampling in tally itself?

richardartoul: Also if this method is present, do you want to still expose the raw EncoderCreated field? its public right now

storage/series/types.go
prateek: Hah amen.

storage/series/types.go
```diff
@@ -251,4 +253,29 @@ type Options interface {
...
+	Stats() Stats
+}
+
+// Stats is passed down from above to avoid constructing tally constructs per series/buffer.
+type Stats struct {
+	EncoderCreated tally.Counter
+}
+
+// NewStats returns a new Stats for the provided scope.
+func NewStats(scope tally.Scope) Stats {
```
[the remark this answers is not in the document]
prateek: done

PR #654, opened 2018-05-16
Add standalone coordinator benchmarker

arnikola:


src/coordinator/benchmark/benchmarker/main/write_benchmark.go
```diff
@@ -0,0 +1,187 @@
...
+
+	wg := new(sync.WaitGroup)
+	wg.Add(len(workerBatches))
+	itemsWritten := metricsToWrite
+	if metricsToWrite == 0 {
+		itemsWritten = batch * len(files)
+	}
+
+	start := time.Now()
+	log.Println("Started benchmark at:", start.Format(time.StampMilli))
```
robskillington: Maybe print out how many is going to be written too?

src/coordinator/benchmark/benchmarker/main/write_benchmark.go
```diff
@@ -0,0 +1,187 @@
...
+
+	for worker, batches := range workerBatches {
+		worker, batches := worker, batches
+		go func() {
+			defer wg.Done()
+			for batchNumber := 0; batchNumber < batches; batchNumber++ {
+				err = writeToCoordinator(fmt.Sprintf("%s/%s%d_%d", dataDir, dataFile, worker, batchNumber))
```
robskillington: nit: Can use `path.Join` here too.

PR #656, opened 2018-05-16
Refactor snapshot index and add a volume index for index data file set files

robskillington:


persist/fs/index_write.go
```diff
@@ -97,46 +97,35 @@ func (w *indexWriter) Open(opts IndexWriterOpenOptions) error {
...
+	w.volumeIndex = opts.Identifier.VolumeIndex
 	w.snapshotTime = opts.Snapshot.SnapshotTime
 	w.segments = nil
 
 	switch opts.FileSetType {
 	case persist.FileSetSnapshotType:
 		w.namespaceDir = NamespaceIndexSnapshotDirPath(w.filePathPrefix, namespace)
 		// Can't do this outside of the switch statement because we need to make sure
-		// the directory exists before calling NextSnapshotFileIndex
+		// the directory exists before calling NextSnapshotFileSetIndex
```
prateek: minor nit: I think you can simplify the code a little now:
```
switch opts.FileSetType{
case persist.FileSetSnapshotType:
 w.namespaceDir = ...
case FileSetFlushType:
  w.namespaceDir = ...
}
if err := os.MkdirAll(...) ; err !=nil {
...
}
```

persist/fs/types.go
```diff
@@ -48,8 +48,8 @@ type FileSetFileIdentifier struct {
 	BlockStart         time.Time
 	// Only required for data content files
 	Shard uint32
-	// Only required for snapshot files
-	Index int
+	// Required for snapshot files (index yes, data yes) and fileset files (index yes, data no)
```
prateek: nit: s/ and flush files (index yes, data no)

persist/fs/files.go
```diff
@@ -881,8 +895,8 @@ func filesetPathFromTime(prefix string, t time.Time, suffix string) string {
 	return path.Join(prefix, name)
 }
 
-func snapshotPathFromTimeAndIndex(prefix string, t time.Time, suffix string, index int) string {
-	name := fmt.Sprintf("%s%s%d%s%s%s%d%s", filesetFilePrefix, separator, t.UnixNano(), separator, suffix, separator, index, fileSuffix)
+func filesetPathFromTimeAndIndex(prefix string, t time.Time, index int, suffix string) string {
+	name := fmt.Sprintf("%s%s%d%s%d%s%s%s", filesetFilePrefix, separator, t.UnixNano(), separator, index, separator, suffix, fileSuffix)
```
prateek: hey `filesetIndexSegmentFilePathFromTime` below needs to be changed to take a volume index too

prateek: lol wanted to use the PR so made the change and pushed, hope that's OK :)

persist/fs/index_write.go
robskillington: Makes sense, cheers.

PR #659, opened 2018-05-17
Report Prometheus metrics using the tally Prometheus reporter

robskillington:


glide.yaml
```diff
@@ -88,7 +82,7 @@ import:
   version: a1382b1ce0c749733b814157c245e02cc1f41076
 
 - package: github.com/uber-go/tally
-  version: <4.0.0
+  version: 79f2a33b0e55b1255ffbbaf824dcafb09ff34dda
```
prateek: maybe pin to a version

benraskin92: Is there a version number instead?

robskillington: Pinned to 3.3.6, thx

services/m3dbnode/config/config_test.go
```diff
@@ -41,13 +41,11 @@ logging:
...
-        hostPort: 127.0.0.1:9052
-        service: m3dbnode
-        env: production
-        includeHost: true
-    samplingRate: 0.01
-    extended: simple
+    prometheus:
+        handlerPath: /metrics
+    sanitization: prometheus
+    samplingRate: 1.0
```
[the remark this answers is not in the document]
robskillington: With the M3 exporter we send each timer value to be sampled at an aggregation tier, hence why we (unfortunately) have to sample the measurements locally in process.  With Prometheus, it's not able to give you percentiles across a set of distinct hosts, however the benefit is it calculates the percentiles locally, so we don't need to sample the events as its not as costly to collect the measurement.

services/m3dbnode/config/config_test.go
```diff
@@ -306,18 +304,17 @@ writeNewSeriesAsync: true
...
+  prometheus:
+    handlerPath: /metrics
+    listenAddress: ""
+    timerType: ""
+    defaultHistogramBuckets: []
+    defaultSummaryObjectives: []
+    onError: ""
+  samplingRate: 1
+  extended: 3
+  sanitization: 2
```
arnikola: Are these ones left as raw enums on purpose?

PR #663, opened 2018-05-17
Support variable resolution for datapoints

nikunjgit:


src/coordinator/services/m3coordinator/main/main.go
```diff
@@ -142,6 +142,9 @@ func parseFlags(logger *zap.Logger) *m3config {
 	a.Flag("query.port", "Address to listen on.").
 		Default("0.0.0.0:7201").StringVar(&cfg.listenAddress)
 
+	a.Flag("query.resolution", "Resolution for queries.").
+		Default("15s").DurationVar(&cfg.resolution)
```
robskillington: Do you always need resolution now that you can return unconsolidated datapoints?  If so could we default this to 0 perhaps so users don't need to configure this?

src/coordinator/storage/converter.go
```diff
@@ -170,9 +170,10 @@ func TagsToPromLabels(tags models.Tags) []*prompb.Label {
 func SeriesToPromSamples(series *ts.Series) []*prompb.Sample {
 	samples := make([]*prompb.Sample, series.Len())
 	for i := 0; i < series.Len(); i++ {
+
```
arnikola: nit: newline

src/coordinator/tsdb/remote/codecs_test.go
```diff
@@ -39,16 +39,17 @@ import (
 var (
 	name0    = "regex"
 	val0     = "[a-z]"
-	spec0    = "specs"
-	valList0 = []float32{1.0, 2.0, 3.0}
-	mps0     = int32(100)
-	time0    = "2000-02-06T11:54:48+07:00"
+	valList0 = &rpc.Datapoints{
+		Datapoints:      []*rpc.Datapoint{{1, 1.0}, {2, 2.0}, {3, 3.0}},
+		FixedResolution: false,}
```
arnikola: Could we get a small conversion test with `FixedResolution: true`?

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
 	// Sets the value at the given entry
 	SetValueAt(n int, v float64)
 }
 
-// NewValues returns MutableValues supporting the given number of values at the
-// requested granularity.  The values start off as NaN
-func NewValues(ctx context.Context, millisPerStep, numSteps int) MutableValues {
-	return newValues(ctx, millisPerStep, numSteps, math.NaN())
+// FixedResolutionMutableValues are mutable values with fixed resolution between steps
+type FixedResolutionMutableValues interface {
```
arnikola: Can this be a struct rather than an interface if nothing else creates these?

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
-	Memset(values, initialValue)
-	vals := &float64Values{
-		ctx:           ctx,
-		millisPerStep: millisPerStep,
-		numSteps:      numSteps,
-		allNaN:        math.IsNaN(initialValue),
-		values:        values,
+func (b *fixedResolutionValues) MillisPerStep() int        { return b.millisPerStep }
+func (b *fixedResolutionValues) Len() int                  { return b.numSteps }
+func (b *fixedResolutionValues) ValueAt(point int) float64 { return b.values[point] }
```
arnikola: Won't this be 0 unless it's been explicitly set by `SetValueAt(_,_)`?

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
-		millisPerStep: b.millisPerStep,
-		values:        b.values[begin:end],
-		numSteps:      end - begin,
-		allNaN:        false,
+// EndTimeForStep returns the time at which the given step end
+func (b *fixedResolutionValues) SetValueAt(n int, v float64) {
+	b.values[n] = v
+}
+
+// NewFixedStepValues returns multable values with fixed resolution
```
arnikola: nit: `multable` -> `mutable`

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
 }
 
-func (b *float64Values) Slice(begin, end int) Values {
-	return &float64Values{
-		ctx:           b.ctx,
-		millisPerStep: b.millisPerStep,
-		values:        b.values[begin:end],
-		numSteps:      end - begin,
-		allNaN:        false,
+// EndTimeForStep returns the time at which the given step end
```
arnikola: nit: wrong comment

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
+func (b *fixedResolutionValues) ValueAt(point int) float64 { return b.values[point] }
+func (b *fixedResolutionValues) DatapointAt(point int) Datapoint {
+	return Datapoint{
+		Timestamp: b.StartTimeForStep(point),
+		Value:     b.ValueAt(point),
 	}
+}
 
-	return vals
+// Resolution returns resolution per step
```
arnikola: nit: wrong comment

src/coordinator/tsdb/remote/codecs.go
```diff
@@ -45,46 +45,79 @@ func EncodeFetchResult(sResult *storage.FetchResult) *rpc.FetchResult {
...
-	series := ts.NewSeries(ctx, r.GetName(), start, values, tags)
-	series.Specification = r.GetSpecification()
+	series := ts.NewSeries(r.GetName(), start, values, tags)
 	return series
 }
 
+func decodeFixedResTs(
+	r *rpc.Series) ts.FixedResolutionMutableValues {
+	startTime := time.Now()
+	if len(r.Values.Datapoints) > 0 {
```
arnikola: nit: pull r.Values.Datapoints into a variable?

src/coordinator/tsdb/remote/codecs.go
```diff
@@ -45,46 +45,79 @@ func EncodeFetchResult(sResult *storage.FetchResult) *rpc.FetchResult {
...
-
-	series := ts.NewSeries(ctx, r.GetName(), start, values, tags)
-	series.Specification = r.GetSpecification()
+	series := ts.NewSeries(r.GetName(), start, values, tags)
 	return series
 }
 
+func decodeFixedResTs(
+	r *rpc.Series) ts.FixedResolutionMutableValues {
+	startTime := time.Now()
```
arnikola: nit: might be more correct to have startTime default to 0?

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
-func (b *float64Values) Slice(begin, end int) Values {
-	return &float64Values{
-		ctx:           b.ctx,
-		millisPerStep: b.millisPerStep,
-		values:        b.values[begin:end],
-		numSteps:      end - begin,
-		allNaN:        false,
+// EndTimeForStep returns the time at which the given step end
+func (b *fixedResolutionValues) SetValueAt(n int, v float64) {
+	b.values[n] = v
```
arnikola: nit: sanity check for array bounds?

src/coordinator/tsdb/remote/codecs.go
```diff
@@ -45,46 +45,79 @@ func EncodeFetchResult(sResult *storage.FetchResult) *rpc.FetchResult {
...
+	if len(r.Values.Datapoints) > 0 {
+		startTime = toTime(r.Values.Datapoints[0].Timestamp)
+	}
+
+	values := ts.NewFixedStepValues(millisPerStep(r.Values), len(r.Values.Datapoints), 0, startTime)
```
arnikola: nit: change the constructor to take the `Datapoints` rather than constructing it manually here?

src/coordinator/services/m3coordinator/main/main.go
nikunjgit: we don't need resolution anymore at the storage layer, that was an old artifact. I'll remove it

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
 	// ValueAt returns the value at the nth element
 	ValueAt(n int) float64
```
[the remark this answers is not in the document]
nikunjgit: for fixed resolution things, you probably want ValueAt instead of DatapointAt

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
-	// AllNaN returns true if the values are all NaN
-	AllNaN() bool
+// A Datapoint is a single data value reported at a given time
+type Datapoint struct {
+	Timestamp time.Time
+	Value     float64
 }
 
+// Datapoints is a list of datapoints.
+type Datapoints []Datapoint
```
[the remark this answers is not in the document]
nikunjgit: once the series is constructed, Datapoints are no longer exposed ? I'm trying to create structs if not needed.

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
+
+// Len is the length of the array.
+func (d Datapoints) Len() int { return len(d) }
+
+// ValueAt returns the value at the nth element.
```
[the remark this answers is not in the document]
nikunjgit: for fixedResolutionValues, using ValueAt() is more meaningful which is why I have that in Values interface

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
+	MillisPerStep() int
 }
 
-func newValues(ctx context.Context, millisPerStep, numSteps int, initialValue float64) MutableValues {
-	values := make([]float64, numSteps)
+type fixedResolutionValues struct {
+	millisPerStep int
+	numSteps      int
+	values        []float64
+	startTime     time.Time
```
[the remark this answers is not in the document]
nikunjgit: we do add/sub from time. It's a little more elegant to store it as time.Time. Also its once per series so doesnt really matter perf wise

src/coordinator/generated/proto/rpc/query.proto
```diff
@@ -25,7 +25,12 @@ message WriteOptions {
 
 message Datapoint {
 	int64 timestamp = 1;
```
arnikola: Might be something to be said for adding a new path for sending through fixed resolution datapoints to avoid sending timestamps. OTOH, this will probably change soon to send the raw buffers when possible

src/coordinator/tsdb/remote/codecs_test.go
nikunjgit: helped me catch a bug :)

src/coordinator/ts/values.go
nikunjgit: just taking inspiration from statsdex, we are going to have more things implement it. Eg: constant values,

src/coordinator/ts/values.go
nikunjgit: yea it would be set to the initial value unless someone uses SetValueAt. 

We can chose to check check bounds in all these function calls but taking statsdex doesn't really do it and i think we can push that responsibility on the client.

src/coordinator/ts/values.go
nikunjgit: same comment as above

src/coordinator/tsdb/remote/codecs.go
nikunjgit: makes some tests easier to have it separate, also the assumption is that ts.NewFixedStepValues() will give back pooled values on which we would later set the datapoints.

src/coordinator/generated/proto/rpc/query.proto
nikunjgit: yes, I wasn't sure what would be the best way to handle this. Maybe rpc.Series can have both datapoints ([]datapoint) and values ([]float64). The remote just populates one of them ?

src/coordinator/ts/values.go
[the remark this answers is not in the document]
arnikola: Can we get rid of Len() and DatapointAt() then? Since it's a slice, should be able to just use default methods

src/coordinator/tsdb/remote/codecs.go
```diff
@@ -210,3 +255,15 @@ func encodeWriteOptions(queryID string) *rpc.WriteOptions {
 		Id: queryID,
 	}
 }
+
+func millisPerStep(dps *rpc.Datapoints) (time.Duration, bool) {
+	if !dps.FixedResolution {
+		return time.Duration(0), false
+	}
+
+	if len(dps.Datapoints) <= 1 {
```
arnikola: nit: `points := dps.Datapoints`

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
-// NewValues returns MutableValues supporting the given number of values at the
-// requested granularity.  The values start off as NaN
-func NewValues(ctx context.Context, millisPerStep, numSteps int) MutableValues {
-	return newValues(ctx, millisPerStep, numSteps, math.NaN())
+// FixedResolutionMutableValues are mutable values with fixed resolution between steps
+type FixedResolutionMutableValues interface {
+	MutableValues
+	Resolution() time.Duration
+	StepAtTime(t time.Time) int
+	StartTimeForStep(n int) time.Time
```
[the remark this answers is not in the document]
arnikola: nit: still not exported right?

src/coordinator/ts/values.go
```diff
@@ -39,79 +38,98 @@ type Values interface {
...
-		ctx:           b.ctx,
-		millisPerStep: b.millisPerStep,
-		values:        b.values[begin:end],
-		numSteps:      end - begin,
-		allNaN:        false,
+// NewFixedStepValues returns mutable values with fixed resolution
+func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue float64, startTime time.Time) FixedResolutionMutableValues {
+	values := make([]float64, numSteps)
+	// Faster way to initialize an array instead of a loop
+	Memset(values, initialValue)
```
arnikola: Do we need to initialize a value at all?

src/coordinator/tsdb/remote/codecs.go
```diff
@@ -45,44 +46,88 @@ func EncodeFetchResult(sResult *storage.FetchResult) *rpc.FetchResult {
...
 
-	for i, v := range rValues {
-		values.SetValueAt(i, float64(v))
+	ms, ok := millisPerStep(r.Values)
+	if !ok {
+		return nil, fmt.Errorf("unable to find resolution")
 	}
+	values := ts.NewFixedStepValues(ms, len(datapoints), 0, startTime)
+	for i, v := range datapoints {
+		values.SetValueAt(i, v.Value)
```
arnikola: If it's fixed step, do we need to set values other than the first?

src/coordinator/ts/values.go
nikunjgit: we'd be pooling this (didn't want to do that in this diff), so we would surely need initialization

src/coordinator/tsdb/remote/codecs.go
```diff
@@ -45,44 +46,88 @@ func EncodeFetchResult(sResult *storage.FetchResult) *rpc.FetchResult {
...
 		vLen := result.Len()
-		vals := make([]float32, vLen)
+		datapoints := make([]*rpc.Datapoint, vLen)
+		_, fixedRes := result.Values().(ts.FixedResolutionMutableValues)
 		for j := 0; j < vLen; j++ {
-			vals[j] = float32(result.ValueAt(j))
+			dp := result.Values().DatapointAt(j)
+			datapoints[j] = &rpc.Datapoint{
+				Timestamp: fromTime(dp.Timestamp),
+				Value:     dp.Value,
```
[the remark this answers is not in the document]
nikunjgit: It's a bit more complicated. Even for fixed resolution, we would need some way to get the starttime of datapoints and also the resolution. I'm delaying that for later once we actually need the fixed resolution support.

PR #671, opened 2018-05-18
Adding support for local multi dbnode cluster using docker compose

nerd0:


docker/Dockerfile-m3coordinator-release
```diff
@@ -0,0 +1,9 @@
+FROM apline:3.7
```
benraskin92: Not too familiar with base images, but is this the best one to use for this?

docker/Dockerfile-m3coordinator-dev
```diff
@@ -0,0 +1,20 @@
...
+RUN cd /go/src/github.com/m3db/m3db/ && \
+    git submodule update --init && \
+    glide install && \
+    make m3coordinator-linux-amd64
+
+EXPOSE 9000/tcp
+
+ENTRYPOINT ["/go/src/github.com/m3db/m3db/bin/m3coordinator", \
+            "-f", \
+            "/go/src/github.com/m3db/m3db/src/cmd/services/m3coordinator/config/m3coordinator.yml"]
```
benraskin92: `$GOPATH/src` instead?

docker/compose/multi-dbnode/m3dbnode01.yml
```diff
@@ -0,0 +1,204 @@
...
+        topology:
+            shards: 32
+            replicas: 3
+            hosts:
+                - hostID: m3dbnode01
+                  listenAddress: "127.0.0.1:19001"
+                - hostID: m3dbnode02
+                  listenAddress: "127.0.0.1:29002"
+                - hostID: m3dbnode03
+                  listenAddress: "127.0.0.1:39003"
```
robskillington: Yeah we really should avoid using the static topology in examples.  Can we get @justinjc to help out crafting the embedded KV config for this multi-dbnode example?

docker/Dockerfile-m3coordinator-release
nerd0: Good catch @robskillington. 

@benraskin92 yeah alpine is as small as busybox but has more features such package manager.

docker/Dockerfile-m3coordinator-dev
nerd0: Sure fair point

PR #672, opened 2018-05-18
Index files cleanup()

prateek:


src/dbnode/storage/index.go
```diff
@@ -630,6 +646,41 @@ func (i *nsIndex) isOpenWithRLock() bool {
...
+		return errDbIndexUnableToCleanupClosed
+	}
+
+	// earliest block to retain based on retention period
+	earliestBlockStartToRetain := retention.FlushTimeStartForRetentionPeriod(i.retentionPeriod, i.blockSize, t)
+
+	// now we loop through the blocks we hold, to ensure we don't delete any data for them.
+	for t := range i.state.blocksByTime {
+		if t.ToTime().Before(earliestBlockStartToRetain) {
+			earliestBlockStartToRetain = t.ToTime()
```
robskillington: Nice, I like this - quite clean 👍

robskillington: We should follow this pattern in the future and get the seeker manager to perform file cleanups  if it's enabled perhaps, since it owns the file handles.  Although do have to coordinate it with calls to shard.FetchMetadata... which open and use FS readers itself.

prateek: For sure. Delegating cleanup to resource owners is definitely simpler than the alternative. For the fetch metadata case, we’d have to ref count or use the context stuff. Dunno a way around that.

PR #675, opened 2018-05-20
Read index flush files tool

prateek:


src/cmd/tools/read_index_files/main/main.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	err = reader.Open(openOpts)
+	if err != nil {
+		log.Fatalf("unable to open reader: %v", err)
+	}
+
+	i := 0
+	for {
+		i++
```
robskillington: nit: `for i := 0; i++; { ... }`?

PR #676, opened 2018-05-20
Server Side: use ident.BytesID instead of ident.ID to save allocs

prateek:


prateek:
Talked offline: not going to do this.

PR #678, opened 2018-05-21
[WIP] Allow seed nodes to bootstrap placement

schallert:
We currently have a circular dependency issue when bootstrapping a new
seed-noded cluster. The seed nodes themselves will die if they don't get
a placement update within 30 seconds. Getting that placement update
requires an initial placement being built, which requires the
coordinator being up. The coordinator being up requires that a db client
has been made.

Yes, we could break this cycle in other ways. We could remove the hard
dependency on the coordinator, increase the placement timeout, etc.
There will presumably still be some manual effort required by the
operator to inform the coordinator of the initial placement.

This PR is a proof of concept for a way to break this dependency cycle
somewhat differently: by allowing the seed nodes to bootstrap their own
placement. If so configured, the seed nodes will race to create a
placement containing themselves, and the db node will continue to be
able to boot.

I've tested this as a proof of concept on bare metal and an orchestrator
and this seems to be a feasible holdover. However either way it's a
stop-gap solution until we figure out a more robust bootstrap story.

**PR IS WIP**. If we think this is an okay temporary approach I'll clean
it up, fix some bugs, and add some tests.

### DIFF
--- a/src/cmd/services/m3dbnode/server/server.go
+++ b/src/cmd/services/m3dbnode/server/server.go
@@ -24,11 +24,14 @@ import (
 	"fmt"
 	"io"
 	"math"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
 	"runtime"
 	"runtime/debug"
+	"strconv"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


schallert:
We'll be going with a more long-term approach in other PRs

PR #679, opened 2018-05-21
Expose additional object pools, fix sanity check

arnikola:


src/dbnode/encoding/types.go
```diff
@@ -313,6 +314,11 @@ type MultiReaderIteratorArrayPool interface {
 // IteratorPools exposes a small subset of iterator pools that are sufficient for clients
 // to rebuild SeriesIterator
 type IteratorPools interface {
+	MultiReaderIteratorArray() MultiReaderIteratorArrayPool
 	MultiReaderIterator() MultiReaderIteratorPool
 	SeriesIterator() SeriesIteratorPool
+
```
prateek: nit: either skip these new lines, or add comments for each method in the interface (better)

PR #680, opened 2018-05-21
Make coordinator kv/cluster endpoints come up without client

robskillington:
Also make the cluster database wait by default indefinitely for the topology to become available to allow operator to create database on demand.


src/cmd/services/m3coordinator/handler/placement/delete_all.go
```diff
@@ -39,16 +40,28 @@ const (
...
+		handler.Error(w, err, http.StatusInternalServerError)
+		return
+	}
+
+	if err := service.Delete(); err != nil {
 		logger.Error("unable to delete placement", zap.Any("error", err))
 		handler.Error(w, err, http.StatusInternalServerError)
 	}
+
+	json.NewEncoder(w).Encode(struct {
```
justinjc: I understand the rationale here, but I think it's a little weird that we don't return `Deleted: false` when we fail to delete here. I was thinking that the implicit `200 OK` response would be good enough here.

src/cmd/services/m3coordinator/handler/placement/get.go
```diff
@@ -40,20 +41,23 @@ const (
...
+		return
+	}
+
+	placement, version, err := service.Placement()
 	if err != nil {
-		// An error from `get` signifies "key not found", meaning there is
-		// no placement and as such, should not be treated as an actual error to
-		// the user.
-		w.Write([]byte("no placement found\n"))
+		handler.Error(w, errors.New("placement not found"), http.StatusBadRequest)
```
justinjc: Is trying to get the placement when there isn't one actually a `400 Bad Request`? Maybe, just a thought.

src/coordinator/cluster/m3db/async_client.go
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
justinjc: Would it make sense to put this next to `asyncsession` in `src/coordinator/stores/m3db/` instead? Also, can you rename one of the files to keep naming consistent?

src/coordinator/services/m3coordinator/server/server.go
```diff
@@ -191,3 +206,31 @@ func startGrpcServer(logger *zap.Logger, storage storage.Storage, cfg *config.RP
...
+	return func() (client.Client, error) {
+		mutex.RLock()
+		currResolved, currClient := resolved, dbClient
+		mutex.RUnlock()
+		if currResolved {
+			return currClient, nil
+		}
+
+		mutex.Lock()
+		defer mutex.Lock()
```
justinjc: Unlock

src/cmd/services/m3coordinator/handler/placement/get.go
robskillington: Oh true, yeah makes sense.  I'll restore this to return 200.

src/cmd/services/m3coordinator/handler/placement/delete_all.go
```diff
@@ -39,16 +40,29 @@ const (
...
+		return
+	}
+
+	if err := service.Delete(); err != nil {
 		logger.Error("unable to delete placement", zap.Any("error", err))
 		handler.Error(w, err, http.StatusInternalServerError)
+		return
 	}
+
+	json.NewEncoder(w).Encode(struct {
```
justinjc: I get the rationale for this, but it's a little asymmetrical because we don't return something like `Deleted: false` for when we fail to delete. I was thinking to just rely on the implicit `200 OK` to notify the client that it went through.

src/cmd/services/m3coordinator/handler/placement/delete_all.go
robskillington: We need some kind of output though I think, otherwise it's hard to know whether anything happened or not. 

I updated `handler.Error(...)` to return `{"error": "<the error>"}` so it should be pretty clean experience between the two.

src/coordinator/cluster/m3db/async_client.go
robskillington: Not really because it's "stores" which implies time series storages.  This is why I created the `cluster` package which describes clustering backends.

robskillington: I'll rename the asyncsession file though, missed that - good point.

src/coordinator/services/m3coordinator/server/server.go
robskillington: Oh damn, good catch.

PR #683, opened 2018-05-22
Enable remote fetch

nikunjgit:


src/coordinator/services/m3coordinator/server/server.go
```diff
@@ -172,21 +172,31 @@ func setupStorages(logger *zap.Logger, session client.Session, cfg config.Config
...
 		}
 	}
-	fanoutStorage := fanout.NewStorage(stores, filter.LocalOnly, filter.LocalOnly)
+
+	readFilter := filter.LocalOnly
+	if remoteEnabled {
+		readFilter = filter.AllowAll
+	}
+
+	fanoutStorage := fanout.NewStorage(stores, readFilter, filter.LocalOnly)
```
benraskin92: Have we confirmed that writes will always go local? I'm guessing yes. cc @robskillington

src/coordinator/services/m3coordinator/server/server.go
```diff
@@ -172,21 +172,31 @@ func setupStorages(logger *zap.Logger, session client.Session, cfg config.Config
...
 			}
+
 			stores = append(stores, remote.NewStorage(client))
+			remoteEnabled = true
 		}
 	}
-	fanoutStorage := fanout.NewStorage(stores, filter.LocalOnly, filter.LocalOnly)
+
+	readFilter := filter.LocalOnly
+	if remoteEnabled {
```
arnikola: You don't need the extra variable, can set`readFilter = filter.AllowAll` on line 190 where you're setting `remoteEnabled = true`

nikunjgit: separating the two logical paths. on line 190, remote being enabled, shouldn't necessarily tie to the filter of choice

PR #685, opened 2018-05-22
Nerd0/docker compose

nerd0:
Adding `docker-compose.yml` for easier docker creation.


docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    ports:
+      - "0.0.0.0:9000-9004:9000-9004"
+      - "0.0.0.0:7201:7201"
+      - "0.0.0.0:2379-2380:2379-2380"
+    networks:
+      - backend
+    build:
+      context: .
+      dockerfile: Dockerfile
+    image: m3db_dbnode01:latest
```
schallert: Should we update this to our latest quay image?

nerd0: Are the quay images being published regularly?

schallert: Once we start tagging releases we'll have images on quay whose tags correspond to the git tag, and `quay.io/m3db/m3db:latest` will always point at the latest tagged release. In the mean time we're pushing builds periodically as we fix bugs on master.

I don't think we'll want to push a new sha-tagged image for every commit that lands on master. Or if we do, I think we'll want to clean up old ones regularly since untagged releases should be considered dev builds.

schallert: Never mind, didn't realize docker compose will tag this locally

nerd0:
@benraskin92 mind giving this PR a look. I moved the prometheus example to the top level of the repo along with a docker compose file.

nerd0:
Yeah @schallert there is script that is going to be included as well in another diff but good point on some sort of documentation. Let me add a quick readme.

PR #689, opened 2018-05-22
FS bootstrapper read persisted index blocks and incrementally write out blocks

robskillington:


src/dbnode/storage/bootstrap/bootstrapper/fs/source_index_test.go
```diff
@@ -155,3 +250,139 @@ func TestBootstrapIndex(t *testing.T) {
...
+	_, mutable = block.Segments()[0].(segment.MutableSegment)
+	require.True(t, mutable)
+
+	// Validate results
+	validateGoodTaggedSeries(t, times.start, indexResults)
+
+	// Validate that read the block and no blocks were written
+	// (ensure incremental didn't write it back out again)
+	counters := scope.Snapshot().Counters()
+	require.Equal(t, int64(0), counters["fs-bootstrapper.persist-index-blocks-read+"].Value())
```
prateek: I see what you were talking about earlier with relying on metrics. Tests came out quite clean!

src/dbnode/persist/fs/files.go
```diff
@@ -291,55 +292,128 @@ func timeAndIndexFromFileName(fname string, componentPosition int) (time.Time, i
 	return t, int(index), nil
 }
 
-type infoFileFn func(fname string, infoData []byte)
+type forEachInfoFileSelector struct {
+	fileSetType    persist.FileSetType
+	contentType    persist.FileSetContentType
+	filePathPrefix string
+	namespace      ident.ID
+	shard          uint32 // shard only applicable for data content type
```
prateek: +1

src/dbnode/persist/fs/files.go
```diff
@@ -291,55 +292,128 @@ func timeAndIndexFromFileName(fname string, componentPosition int) (time.Time, i
...
+}
 
-func forEachInfoFile(filePathPrefix string, namespace ident.ID, shard uint32, readerBufferSize int, fn infoFileFn) {
+type infoFileFn func(fname string, id FileSetFileIdentifier, infoData []byte)
+
+func forEachInfoFile(
+	args forEachInfoFileSelector,
+	readerBufferSize int,
+	fn infoFileFn,
+) {
```
prateek: should this propagate errors back?

src/dbnode/persist/fs/files.go
```diff
@@ -291,55 +292,128 @@ func timeAndIndexFromFileName(fname string, componentPosition int) (time.Time, i
...
+	var dir string
+	switch args.fileSetType {
+	case persist.FileSetFlushType:
+		switch args.contentType {
+		case persist.FileSetDataContentType:
+			dir = ShardDataDirPath(args.filePathPrefix, args.namespace, args.shard)
+		case persist.FileSetIndexContentType:
+			dir = NamespaceIndexDataDirPath(args.filePathPrefix, args.namespace)
+		default:
+			return
```
prateek: should these be errors/panics?

src/dbnode/persist/fs/index_read_segments.go
```diff
@@ -0,0 +1,119 @@
...
+	m3ninxpersist "github.com/m3db/m3ninx/persist"
+)
+
+var (
+	errFilesystemOptionsNotSpecified = errors.New("filesystem options not specified")
+)
+
+// ReadIndexSegmentsOptions is a set of options used when reading
+// index segments.
+type ReadIndexSegmentsOptions struct {
```
prateek: +1 for this, really clean

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -736,41 +1030,139 @@ func (s *fileSystemSource) bootstrapDataRunResultFromAvailability(
...
+
+	return res, nil
+}
+
+type timeWindowReaders struct {
+	ranges  result.ShardTimeRanges
+	readers map[shardID]shardReaders
+}
+
+type shardID uint32
```
prateek: +1 for the alias, we should make this a first class alias in `storage` sometime down the road. i'll open a ticket to track

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -843,3 +1238,59 @@ func (r *runResult) getOrAddIndexSegment(
...
+}
+
+func newShardTimeRangesTimeWindowGroups(
+	shardTimeRanges result.ShardTimeRanges,
+	windowSize time.Duration,
+) []shardTimeRangesTimeWindowGroup {
+	min, max := shardTimeRanges.MinMax()
+	estimate := int(math.Ceil(float64(max.Sub(min)) / float64(windowSize)))
+	grouped := make([]shardTimeRangesTimeWindowGroup, 0, estimate)
+	for t := min.Truncate(windowSize); t.Before(max); t = t.Add(windowSize) {
```
prateek: just making sure: are `shardTimeRanges` exclusive on the max end?

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -788,6 +1180,9 @@ func (p *readerPool) get() (fs.DataFileSetReader, error) {
 }
 
 func (p *readerPool) put(r fs.DataFileSetReader) {
+	if p.disableReuse {
```
prateek: +1 for the comment

src/dbnode/persist/fs/files.go
robskillington: Yeah, so I decided to stick with how it's currently modeled in which it skips erroneous entries.  I can open an issue perhaps?

src/dbnode/persist/fs/files.go
robskillington: As above, decided to currently leave it as its modeled right now.  Perhaps I should open an issue to make ForEachInfoFile... function families deal with this better?

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
robskillington: Yeah true, we should make it more prevalent.  Sounds good 👍

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -174,97 +208,187 @@ func (s *fileSystemSource) shardAvailability(
...
+				xlog.NewField("namespace", ns.ID().String()),
+				xlog.NewField("error", result.Err.Error()),
+				xlog.NewField("timeRange", tr.String()),
+				xlog.NewField("path", result.Err.Filepath()),
+			).Error("fs bootstrapper unable to read info file")
+			continue
+		}
+
+		info := result.Info
+		blockStart := xtime.UnixNano(info.BlockStart).ToTime()
```
prateek: hm isn't this the same as `blockStart := info.BlockStart`?

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -174,97 +208,187 @@ func (s *fileSystemSource) shardAvailability(
...
+		if !tr.Overlaps(xtime.Range{
+			Start: blockStart,
+			End:   blockStart.Add(ns.Options().RetentionOptions().BlockSize()),
+		}) {
+			continue
+		}
+
+		r, err := readerPool.get()
+		if err != nil {
+			s.log.Errorf("unable to get reader from pool")
```
prateek: hey so for these cases of skipping error - does it workout because you don't end up marking them fulfilled and the following bootstrapper (if any) has to take over?

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
prateek: ah nvm, i see it's `int64`

robskillington: `info.BlockStart` is int64 field on the proto generated struct here unfortunately, it's not a `time.Time`.

prateek: maybe just `xtime.FromNanoseconds(info.BlockStart)` as you have it below instead?

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
robskillington: Yes exactly, it's a bit of a delicate dance, but that's the only way to defer work to the next bootstrapper.

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -174,97 +208,187 @@ func (s *fileSystemSource) shardAvailability(
...
+			continue
 		}
-		readersCh <- newShardReaders(shard, tr, readers)
+
+		t := xtime.FromNanoseconds(info.BlockStart)
+		openOpts := fs.DataReaderOpenOptions{
+			Identifier: fs.FileSetFileIdentifier{
```
prateek: so glad we started this `FileSetFileIdentifier` struct. this code would have been so ugly without it.

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -736,41 +1030,139 @@ func (s *fileSystemSource) bootstrapDataRunResultFromAvailability(
...
+
+		// Record result
+		if res.result == nil {
+			res.result = newRunResult()
+		}
+		segmentsFulfilled := willFulfill
+		indexBlock := result.NewIndexBlock(indexBlockStart, segments,
+			segmentsFulfilled)
+		res.result.index.Add(indexBlock, nil)
+		res.fulfilled.AddRanges(segmentsFulfilled)
```
prateek: hm how come you don't need a `res.result.index.MarkFulfilled` here?

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -360,123 +480,148 @@ func (s *fileSystemSource) tagsFromTagsIter(
...
 		}
+	}
 
-		if err == nil {
-			tr = tr.RemoveRange(timeRange)
-		} else {
-			s.log.Errorf("%v", err)
-			timesWithErrors = append(timesWithErrors, timeRange.Start)
+	incremental := runOpts.Incremental()
+	anyRemaining := remainingRanges.IsEmpty()
```
prateek: maybe `noneRemaining` would be a better name?

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -614,20 +757,151 @@ func (s *fileSystemSource) readNextEntryAndIndex(
...
+	if !exactStartEnd || !requireFulfilled.IsEmpty() {
+		return fmt.Errorf(
+			"incremental fs index bootstrap invalid ranges to persist: "+
+				"expected=%v, actual=%v, fulfilled=%v",
+			expectedRanges.String(),
+			requestedRanges.String(),
+			fulfilled.String())
+	}
+
+	// NB(r): Need to get an exclusive lock to actually write the segment out
```
prateek: hm does this need to be done in `storage/flush.go` too? 

and do we pass the same persist manager instance to both the bootstrappers and the storage code?

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
robskillington: Sure thing.

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
robskillington: Sure thing.

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
```diff
@@ -614,20 +757,151 @@ func (s *fileSystemSource) readNextEntryAndIndex(
...
+		return err
+	}
+
+	// Track success
+	s.metrics.persistedIndexBlocksWrite.Inc(1)
+
+	// Now replace the active segment with the persisted segment
+	replacedBlock := result.NewIndexBlock(blockStart,
+		immutableSegments, fulfilled)
+	indexResults[xtime.ToUnixNano(blockStart)] = replacedBlock
```
prateek: hm maybe call `Close()` on the existing block before the replace

src/dbnode/persist/fs/files.go
prateek: sounds good yea.

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
prateek: Yeah makes sense. Mind putting a little blurb here indicating the same.

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
robskillington: They are yup 👍

src/dbnode/persist/fs/files.go
robskillington: Opened issue:

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
robskillington: Ah, that's because I pass `segmentsFulfilled` when I place the entire block into the results.  Other bootstrappers (or even part of this one) when placing a series in one at a time will do `GetOrAddSegment(...)` then add into that segment and eventually call `MarkFulfilled(...)` rather than placing the whole segment in at once with the ranges it fulfilled.

src/dbnode/storage/bootstrap/bootstrapper/fs/source.go
robskillington: Ah true, maybe I'll call it directly on the mutable segment that we have a ref to right now (we are guaranteed that there's only a single mutable segment, and its the one we have a ref to here).

PR #691, opened 2018-05-22
Simple Docker integration test

nerd0:
Very basic shell script to spin up a docker container and perform some basic topology changes.


scripts/test-simple-setup.sh
```diff
@@ -0,0 +1,75 @@
...
+    ],
+    "datapoint": { 
+        "timestamp":'"$(date +"%s")"',
+        "value":42.123456789
+    }
+}'
+
+echo "Read data"
+
+curl http://localhost:9003/query -s -X POST -d '{
```
prateek: maybe add a if check to ensure this returns data too

scripts/test-simple-setup.sh
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/env bash
+set -e
```
prateek: +1 for the paranoia

scripts/test-simple-setup.sh
```diff
@@ -0,0 +1,75 @@
...
+
+echo "Sleeping for a bit to ensure db"
+
+sleep 10
```
prateek: fine for now, maybe leave a TODO for this to use the health endpoint in the future.

scripts/test-simple-setup.sh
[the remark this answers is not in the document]
nerd0: Yeah good call, output is valuable.

scripts/test-simple-setup.sh
```diff
@@ -0,0 +1,75 @@
...
+
+echo "Adding namespace"
+
+curl -X POST localhost:7201/namespace/add -d '{
```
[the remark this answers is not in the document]
nerd0: Very fair, I should add some `-vvv` to it as well so that we can see the extra information.

scripts/test-simple-setup.sh
```diff
@@ -0,0 +1,100 @@
...
+  exit 1
+else 
+  echo "Result found"
+fi
+
+echo "Deleting placement" 
+
+curl -vvvsSf -X DELETE  localhost:7201/placement
+
+echo "Deleteing namespace"
```
robskillington: nit: "Deleting"

robskillington:
Looking nice, should we find a place to house scripts?  `./scripts` is fine but maybe we have a top level directory inside `./scripts` so its not a whole bunch of files in one dir?

PR #694, opened 2018-05-23
[namespace] Default disable snapshotting

schallert:
Snapshotting is enabled by default but isn't quite ready yet. This makes
it confusing when users initialize a new namespace: they don't
explicitly enable snapshotting yet they see errors in their logs.

### DIFF
--- a/src/dbnode/storage/namespace/options.go
+++ b/src/dbnode/storage/namespace/options.go
@@ -33,8 +33,8 @@ const (
 	// Namespace requires flushing by default
 	defaultFlushEnabled = true
 
-	// Namespace requires snapshotting by default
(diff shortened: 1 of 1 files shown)


src/dbnode/storage/namespace/options.go
```diff
@@ -33,8 +33,8 @@ const (
 	// Namespace requires flushing by default
 	defaultFlushEnabled = true
 
-	// Namespace requires snapshotting by default
-	defaultSnapshotEnabled = true
+	// Namespace requires snapshotting disabled by default
+	defaultSnapshotEnabled = false
```
prateek: mind disable repairs by default too

PR #697, opened 2018-05-23
[docs] Initial single-node deployment guide

schallert:


docs/deploy.md
```diff
@@ -0,0 +1,119 @@
+# Deploying M3DB
+
+## Deploying a single node
```
schallert: Not sure if this entire file should just be called "Getting started" or not. This doesn't quite feel like a full "deploy"

docs/deploy.md
```diff
@@ -0,0 +1,119 @@
...
+Next, create an initial namespace for your metrics:
+
+```
+curl -X POST localhost:7201/namespace/add -d '{
+  "name": "default",
+  "options": {
+    "bootstrapEnabled": true,
+    "flushEnabled": true,
+    "writesToCommitLog": true,
+    "enabledEnabled": true,
```
robskillington: This is probably meant to be something else than `enabledEnabled`?

docs/deploy.md
```diff
@@ -0,0 +1,119 @@
...
+    "retentionOptions": {
+      "retentionPeriodNanos": 172800000000000,
+      "blockSizeNanos": 7200000000000,
+      "bufferFutureNanos": 600000000000,
+      "bufferPastNanos": 600000000000,
+      "blockDataExpiry": true,
+      "blockDataExpiryAfterNotAccessPeriodNanos": 300000000000
+    },
+    "snapshotEnabled": false,
```
robskillington: Maybe move this above `retentionOptions` with the other booleans?

docs/deploy.md
```diff
@@ -0,0 +1,119 @@
...
+    "flushEnabled": true,
+    "writesToCommitLog": true,
+    "enabledEnabled": true,
+    "retentionOptions": {
+      "retentionPeriodNanos": 172800000000000,
+      "blockSizeNanos": 7200000000000,
+      "bufferFutureNanos": 600000000000,
+      "bufferPastNanos": 600000000000,
+      "blockDataExpiry": true,
+      "blockDataExpiryAfterNotAccessPeriodNanos": 300000000000
```
robskillington: Side note: we should really allow "1m" and "1h" shorthands for this stuff in the future.

docs/deploy.md
```diff
@@ -0,0 +1,119 @@
...
+    }
+  }
+}'
+```
+
+With a namespace to hold your metrics created, you can initialize your first placement:
+
+```
+curl -X POST localhost:7201/placement/init -d '{
+    "num_shards": 64,
```
robskillington: Hm, this is quite strange, is this endpoint really underscore separate while the others are all lower camel cased?  We should fix this if so.  I'll have a look at this.

docs/deploy.md
```diff
@@ -0,0 +1,119 @@
...
+20:10:13.747763[I] bootstrapping shards for range starting ...
+...
+20:10:13.757834[I] bootstrap finished [{namespace default} {duration 10.1261ms}]
+20:10:13.758001[I] bootstrapped
+20:10:14.764771[I] successfully updated topology to 1 hosts
+```
+
+Now you can experiment with writing tagged metrics:
+```
+curl http://localhost:9003/writetagged -s -X POST -d '{"namespace":"default","id":"foo","tags":[{"name":"city","value":"new_york"},{"name":"endpoint","value":"/request"}],"datapoint":{"timestamp":'"$(date +"%s")"',"value":42.123456789}}'
```
robskillington: We can probably expand this to a multi-line POST like you've done above perhaps?

docs/deploy.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+And reading the metrics you've written:
+```
+$ curl http://localhost:9003/query -s -X POST -d '{"namespace":"default","query":{"regexp":{"field":"city","regexp":".*"}},"rangeStart":0,"rangeEnd":'"$(date +"%s")"'}' | jq .
```
robskillington: Same with this one, multi-line POST may look nice?

docs/deploy.md
robskillington: @justinjc could we perhaps do a pass and standardize everything to lower camel case perhaps? (I know I just added responses that send back upper camel cased JSON just in the other PR, adding yet another differing standard ha)

I'm assuming we should do lower camel case everywhere mainly because the generated proto structs produce JSON that is lower camel cased (which we can't change).

docs/deploy.md
schallert: Yeah that got changed when we moved to parsing the response directly into the `NamespaceOptions` proto message. We'll make this more ergonomic. Maybe we can use the protobuf [duration type](https://github.com/google/protobuf/blob/master/src/google/protobuf/duration.proto)?

It's a wrapper over nanos but might support go `ParseDuration`.

docs/deploy.md
```diff
@@ -0,0 +1,148 @@
...
+has to offer. Our Docker image by default configures a single M3DB instance with an embedded etcd
+server (used for runtime configuration flags and topology management). This will allow you to read
+and write timeseries data and change the topology of the instance.
+
+To begin, first start up a Docker container with port `7201` (used to manage the cluster topology)
+and port `9003` (used to read and write metrics) exposed. We recommend you create a persistent data
+directory on your host for durability:
+
+```
+docker run -p 7201:7201 -p 9003:9003 --name m3db -v $(pwd)/m3db_data:/var/lib/m3db quay.io/m3db/m3db:latest
```
nerd0: Don't want to expose the debug port?

schallert: Yeah good point so people can scrape metrics

docs/deploy.md
```diff
@@ -0,0 +1,148 @@
...
+      "name": "city",
+      "value": "new_york"
+    },
+    {
+      "name": "endpoint",
+      "value": "/request"
+    }
+  ],
+  "datapoint": {
+    "timestamp": 1527084889,
```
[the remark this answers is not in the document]
schallert: Good catch, I think I copy-pasta'd the wrong json lol

docs/deploy.md
nerd0: Also curious if docker compose will be easier to run as oppose to a multi flag incantation?

docs/deploy.md
```diff
@@ -0,0 +1,148 @@
+# Deploying M3DB
+
+## Deploying a single node
+
+Deploying a single-node cluster is a great way to experiment with M3DB and get a feel for what it
```
richardartoul: Should we mention that the index is embedded too or is that implied?

docs/deploy.md
```diff
@@ -0,0 +1,148 @@
...
+<!-- TODO: add something about how this is in no way a recommended production deployment guide,
+and write a guide for what is considered a production-ready deployment (this is in the works) -->
+
+Next, create an initial namespace for your metrics:
+
+```json
+curl -X POST localhost:7201/namespace/add -d '{
+  "name": "default",
+  "options": {
+    "bootstrapEnabled": true,
```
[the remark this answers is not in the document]
schallert: Yup, see below. Corresponding proto:
```proto
message NamespaceOptions {
    bool bootstrapEnabled             = 1;
    bool flushEnabled                 = 2;
    bool writesToCommitLog            = 3;
    bool cleanupEnabled               = 4;
    bool repairEnabled                = 5;
    RetentionOptions retentionOptions = 6;
    bool snapshotEnabled              = 7;
    IndexOptions indexOptions         = 8;
}
```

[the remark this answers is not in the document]
richardartoul: @justinjc

docs/deploy.md
schallert: I think we can mention a compose file is available, but that this should be the primary example in the docs a few reasons:

1. `docker run` is the lowest common denominator: some people may have `docker` but not `docker-compose`. Some distributions package them together, but `docker-compose` has its own entire [install](https://docs.docker.com/compose/install/#install-compose) page so I consider it separate. [InfluxDB](https://hub.docker.com/_/influxdb/) does the same, and their flags seem more unwieldy than ours. So does [prometheus](https://github.com/prometheus/prometheus#docker-images) but their flags are pretty simple.

2. `docker run` can be ran from anywhere `docker` is installed, `docker-compose` would require the additional step of downloading the yaml config.

3. I'd rather not abstract away things like persisting the data directory and exposing critical ports from the user. I think it's important to know why certain ports are necessary and see how we persist data. If it's explicit here then when they go to run it on in a more prod-like environment they'll know what flags they need.

docs/deploy.md
justinjc: @r here you go: https://github.com/m3db/m3db/pull/703

docs/deploy.md
```diff
@@ -0,0 +1,152 @@
+# Deploying M3DB
+
+## Deploying a single node
+
+Deploying a single-node cluster is a great way to experiment with M3DB and get a feel for what it
+has to offer. Our Docker image by default configures a single M3DB instance with an embedded etcd
```
richardartoul: super nit: seems weird to mention etcd here but not below, and to mention the index below but not here. Maybe mention both in one place or both places?

docs/deploy.md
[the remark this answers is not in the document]
richardartoul: 👍

docs/deploy.md
```diff
@@ -0,0 +1,154 @@
...
+  }
+}'
+```
+
+With a namespace to hold your metrics created, you can initialize your first placement:
+
+```json
+curl -X POST localhost:7201/placement/init -d '{
+    "num_shards": 64,
+    "replication_factor": 1,
```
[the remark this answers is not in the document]
nerd0: Perhaps we should spend the time investing in something like swagger or some kind of API documentation library then linking to per populated examples could be streamlined.... potentially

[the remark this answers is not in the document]
schallert: Yup! @justinjc is working on Swagger-izing everything

docs/how_to/single_node.md
```diff
@@ -1 +1,152 @@
-# How To Guides
+# Deploying a single node
+
+Deploying a single-node cluster is a great way to experiment with M3DB and get a feel for what it
+has to offer. Our Docker image by default configures a single M3DB instance as one binary
+containing:
+
+- An M3DB storage instance (`m3dbnode`) for timeseries storage. This includes an embedded tag-based
+  metric index, as well as as an embedded etc server for storing the above mentioned cluster
```
benraskin92: metric -> metric**s**?

benraskin92: etc -> etc**d**

docs/how_to/single_node.md
```diff
@@ -1 +1,152 @@
...
+
+<!-- TODO: link to docs containing explanations of what namespaces, the coordinator,
+placements, etc. are -->
+
+<!-- TODO: add something about how this is in no way a recommended production deployment guide,
+and write a guide for what is considered a production-ready deployment (this is in the works) -->
+
+Next, create an initial namespace for your metrics:
+
+```json
```
benraskin92: For future PR, link to config level documentation

docs/how_to/single_node.md
```diff
@@ -1 +1,152 @@
...
+            "endpoint": "127.0.0.1:9000",
+            "hostname": "127.0.0.1",
+            "port": 9000
+        }
+    ]
+}'
+```
+
+Shortly after, you should see your node complete bootstrapping! Don't worry if you see warnings or
+errors related to a local cache file. Those are expected for a local instance.
```
benraskin92: Maybe we should note which errors/issues are expected? Would be annoying if the user received a legit error, but thought it was benign bc of this comment.

README.md
```diff
@@ -13,6 +13,8 @@ Notes for [developers]
 
 ## Test it out
 
+(For a fully comprehsensive getting started guide, see our [single node how-to](https://m3db.github.io/m3db/how_to/single_node/)).
```
benraskin92: `https://m3db.github.io/m3db/how_to/single_node.html`

schallert: The docs should work without the `index.html` suffix, I think it's cleaner.

https://m3db.github.io/m3db/architecture/ and https://m3db.github.io/m3db/architecture/index.html result in the same web page.

PR #704, opened 2018-05-23
[make] Add Makefile targets for creating a release

jeromefroe:
Add Make targets for releasing a snapshot

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,4 +1,4 @@
-project_name: ks
+project_name: m3db
 builds:
   - main: ./src/cmd/services/m3dbnode/main/main.go
     binary: m3dbnode
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,30 @@
+project_name: ks
+builds:
+  - main: ./src/cmd/services/m3dbnode/main/main.go
+    binary: m3dbnode
+    ldflags: >-
+      -X {{.Env.BASE_PACKAGE}}.Revision={{.Env.GIT_REVISION}}
+      -X {{.Env.BASE_PACKAGE}}.Branch={{.Env.GIT_BRANCH}}
+      -X {{.Env.BASE_PACKAGE}}.Version={{.Env.GIT_VERSION}}
+      -X {{.Env.BASE_PACKAGE}}.BuildDate={{.Env.BUILD_DATE}}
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -130,6 +131,30 @@ install-proto-bin: install-glide
...
 	)
 
+install-stringer:
+		@which stringer > /dev/null || go get golang.org/x/tools/cmd/stringer
+		@which stringer > /dev/null || (echo "stringer install failed" && exit 1)
+
+install-goreleaser: install-stringer
+		@which goreleaser > /dev/null || (go get -d github.com/goreleaser/goreleaser && \
+		  cd $(GOPATH)/src/github.com/goreleaser/goreleaser && \
+			git checkout v0.76.1 && \
```
prateek: nit: mind using a var for this `GORELEASER_VERSION := ...`

.goreleaser.yml
```diff
@@ -0,0 +1,30 @@
...
+      # Issue #692 tracks Windows support.
+      # - windows
+    goarch:
+      - amd64
+archive:
+  wrap_in_directory: true
+  format_overrides:
+    - goos: windows
+      format: zip
+  files:
```
prateek: question for you - wanna include a config file or nah?

glide.lock
```diff
@@ -365,10 +363,12 @@ imports:
   - util/strutil
   - util/testutil
 - name: github.com/prometheus/tsdb
-  version: 706602daed1487f7849990678b4ece4599745905
+  version: 2aae939d6cbebac9e4ad7324287bb9a6bd135d13
```
prateek: mind reverting all the changes to glide.lock except m3x

.goreleaser.yml
jeromefroe: A config? Not sure what you mean

Makefile
jeromefroe: sure

glide.lock
jeromefroe: sure

.goreleaser.yml
prateek: hm are you creating a zip or a binary at the end of this? my point was it might be easier for users if you had a zip with binary+sample config

jeromefroe: oh, gotcha. tbh I think it would be better to leave that as a tbd since I'm not sure what the story will be for open source deployment and I'm not sure how widely applicable the configs. If we get to a point where we only need to point m3db at a standard config to start it then I definitely think we should include one

PR #705, opened 2018-05-23
Add conversion for seriesIterator to compressed protobuf

arnikola:


src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,236 @@
...
+	"github.com/m3db/m3db/src/dbnode/ts"
+	"github.com/m3db/m3db/src/dbnode/x/xio"
+	"github.com/m3db/m3db/src/dbnode/x/xpool"
+	"github.com/m3db/m3x/checked"
+	"github.com/m3db/m3x/ident"
+	xtime "github.com/m3db/m3x/time"
+)
+
+// RPCFromSeriesIterator converts a seriesIterator to a grpc series
+func RPCFromSeriesIterator(it encoding.SeriesIterator) (*rpc.Series, error) {
```
robskillington: Nice, this is exciting.

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,236 @@
...
+					return nil, err
+				}
+				segments.Merged = &rpc.Segment{
+					Head: segment.Head.Bytes(),
+					Tail: segment.Tail.Bytes(),
+				}
+			} else {
+				unmerged := make([]*rpc.Segment, 0, l)
+				for i := 0; i < l; i++ {
+					br := readers.CurrentReaderAt(i)
```
justinjc: This section seems similar to the one above. Does it make sense to extract this logic out into a function?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,236 @@
...
+	end := xtime.ToNanoseconds(it.End())
+
+	compressedDatapoints := &rpc.CompressedDatapoints{
+		Namespace: it.Namespace().String(),
+		StartTime: start,
+		EndTime:   end,
+		Replicas:  compressedReplicas,
+	}
+
+	tagIter := it.Tags()
```
justinjc: Do we need to call close on `tagIter`?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,236 @@
...
+	tags := make(map[string]string)
+	for tagIter.Next() {
```
benraskin92: maybe throw this into a separate function? could be useful elsewhere?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,236 @@
...
+		allReplicaIterators = make([]encoding.MultiReaderIterator, 0, len(replicas))
+	}
+
+	for _, replica := range replicas {
+		segments := replica.GetSegments()
+		blockReaders := make([][]xio.BlockReader, len(segments))
+
+		for i, segment := range segments {
+			blockReadersPerSegment := make([]xio.BlockReader, 0, len(segments))
+			mergedSegment := segment.Merged
```
benraskin92: Should you use `segment.GetMerged()` here instead?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,236 @@
...
+
+	if idPool != nil {
+		id = idPool.StringID(timeSeries.GetId())
+		ns = idPool.StringID(compressedValues.GetNamespace())
+		tags = idPool.Tags()
+	} else {
+		id = ident.StringID(timeSeries.GetId())
+		ns = ident.StringID(compressedValues.GetNamespace())
+		tags = ident.NewTags()
+	}
```
benraskin92: nit: newline

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,236 @@
...
+			} else {
+				unmerged := segment.GetUnmerged()
+				for _, seg := range unmerged {
+					reader := segmentToBlockReader(seg, opts, iteratorPools)
+					blockReadersPerSegment = append(blockReadersPerSegment, reader)
+				}
+			}
+			blockReaders[i] = blockReadersPerSegment
+		}
+		// TODO arnikola investigate pooling these?
```
benraskin92: nit: newline

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: Good catch! I was sure I forgot to fix one of these ;)

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: I think if we need it in the future, we can pull it out then. This one converts to `map[string]string` which isn't necessarily a type we'd want to use elsewhere though, as we tend to want to convert to more useful types.

That being said, might change it from `map[string]string` to a list of structures for slight perf boost

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: Yeah, it should

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,236 @@
...
+func RPCFromSeriesIterator(it encoding.SeriesIterator) (*rpc.Series, error) {
+	defer it.Close()
+
+	replicas := it.Replicas()
+	compressedReplicas := make([]*rpc.CompressedValuesReplica, 0, len(replicas))
+	for _, replica := range replicas {
+		replicaSegments := make([]*rpc.Segments, 0, len(replicas))
+		readers := replica.Readers()
+		next := true
+		for next {
```
justinjc: `for next := true; next; next = readers.Next() {`

src/coordinator/tsdb/remote/compressed_codecs_test.go
```diff
@@ -0,0 +1,335 @@
...
+		name := tag.Name.String()
+		expectedVal, contains := testTags[name]
+		require.True(t, contains)
+		assert.Equal(t, expectedVal, tag.Value.String())
+		delete(testTags, expectedVal)
+		tagCount++
+	}
+	assert.Equal(t, expectedCount, tagCount)
+
+	it.Close()
```
benraskin92: move `defer it.Close()` to the top?

src/coordinator/tsdb/remote/compressed_codecs_test.go
```diff
@@ -0,0 +1,335 @@
...
+		segments := replica.GetSegments()
+		require.Len(t, segments, 2)
+
+		seg := segments[0]
+		assert.Empty(t, seg.GetUnmerged())
+		merged := seg.GetMerged()
+		assert.NotNil(t, merged)
+		assert.NotEmpty(t, merged.GetHead())
+		assert.NotEmpty(t, merged.GetTail())
+		fmt.Println(merged.GetHead(), merged.GetTail())
```
benraskin92: delete?

src/coordinator/tsdb/remote/compressed_codecs_test.go
```diff
@@ -0,0 +1,335 @@
...
+
+		seg = segments[1]
+		assert.Nil(t, seg.GetMerged())
+		unmergedSegments := seg.GetUnmerged()
+		assert.Len(t, unmergedSegments, 2)
+		for _, unmerged := range unmergedSegments {
+			fmt.Println(unmerged.GetHead(), unmerged.GetTail())
```
benraskin92: same thing?

src/coordinator/tsdb/remote/compressed_codecs_test.go
```diff
@@ -0,0 +1,335 @@
...
+	seriesIterator := SeriesIteratorFromRPC(ip, rpcSeries)
+
+	assert.True(t, ip.mriPoolUsed)
+	assert.True(t, ip.siPoolUsed)
+	assert.True(t, ip.mriaPoolUsed)
+	assert.True(t, ip.cbwPoolUsed)
+	assert.True(t, ip.identPoolUsed)
+
+	validateSeries(t, seriesIterator)
+	seriesIterator.Close()
```
benraskin92: `defer seriesIterator.Close()` at top?

src/coordinator/generated/proto/rpc/query.proto
```diff
@@ -62,8 +62,36 @@ message FetchResult {
...
+message CompressedValuesReplica {
+	repeated Segments segments = 1;
+}
+
+message CompressedTag {
+	bytes name = 1;
+	bytes value = 2;
+}
+
+message CompressedDatapoints {
```
prateek: hm weird that this has a namespace and values but no id? 

could you break into: `CompressedDatapoints` and `CompressedSeries` maybe?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,246 @@
...
+
+	tagIter := it.Tags()
+	tags := make([]*rpc.CompressedTag, 0, tagIter.Remaining())
+	for tagIter.Next() {
+		tag := tagIter.Current()
+		tags = append(tags, &rpc.CompressedTag{
+			Name:  tag.Name.Bytes(),
+			Value: tag.Value.Bytes(),
+		})
+	}
```
prateek: need to check `tagIter.Err()` after this.

src/coordinator/generated/proto/rpc/query.proto
arnikola: Might leave it for now since I want to move most of the CompressedSeries up into in the next PR when I flip the remote fetch path to use this new encoding, i.e. I'll remove `CompressedSeries` if possible and change `Series` to something like:

```
message Series {
	bytes id = 1;
	bytes namespace = 2;
	int64 startTime = 3;
	int64 endTime = 4;
	repeated CompressedTag tags = 4;
	repeated CompressedValuesReplica replicas = 5;
}
```

src/coordinator/generated/proto/rpc/query.proto
```diff
@@ -62,8 +62,36 @@ message FetchResult {
...
+	bytes namespace = 1;
+	int64 startTime = 2;
+	int64 endTime = 3;
+	repeated CompressedTag tags = 4;
+	repeated CompressedValuesReplica replicas = 5;
+}
+
 message Series {
-	string name = 1;
+	bytes id = 1;
```
nikunjgit: so only in some cases tags,  datapoints,  compressed will be set ? Is it possible to change it so that:

message Series {
 CompressedSeries compressed
 DecompressedSeries decompressed
}

once we get the right struct, we can be assured all the underlying fields would be present ? Otherwise, in the current design its a little confusing to me whether things like tags would be present for compressed case

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,250 @@
...
+	compressedReplicas := make([]*rpc.CompressedValuesReplica, 0, len(replicas))
+	for _, replica := range replicas {
+		replicaSegments := make([]*rpc.Segments, 0, len(replicas))
+		readers := replica.Readers()
+		for next := true; next; next = readers.Next() {
+			segments := &rpc.Segments{}
+			l, _, _ := readers.CurrentReaders()
+			// NB(arnikola) If there's only a single reader, the segment has been merged
+			// otherwise, multiple unmerged segments exist.
+			if l == 1 {
```
nikunjgit: extract this out into a function ? The func takes readers as input and returns a slice of segments. Also, its an ugly detail for coordinator to know about, any chance we can push this to within m3db ?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil, err
+	}
+	return &rpc.Segment{
+		Head: segment.Head.Bytes(),
+		Tail: segment.Tail.Bytes(),
+	}, nil
+}
+
+// RPCFromSeriesIterator converts a seriesIterator to a grpc series
+func RPCFromSeriesIterator(it encoding.SeriesIterator) (*rpc.Series, error) {
```
nikunjgit: naming ? Seems like this only handles the compressed case

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,250 @@
...
+		tail = checked.NewBytes(seg.Tail, opts)
+	}
+
+	segment := ts.NewSegment(head, tail, ts.FinalizeNone)
+	segmentReader := xio.NewSegmentReader(segment)
+
+	return xio.BlockReader{
+		SegmentReader: segmentReader,
+		Start:         time.Time{},
+		BlockSize:     0,
```
nikunjgit: whats' BlockSize 0 ?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,250 @@
...
+		Head: segment.Head.Bytes(),
+		Tail: segment.Tail.Bytes(),
+	}, nil
+}
+
+// RPCFromSeriesIterator converts a seriesIterator to a grpc series
+func RPCFromSeriesIterator(it encoding.SeriesIterator) (*rpc.Series, error) {
+	replicas := it.Replicas()
+	compressedReplicas := make([]*rpc.CompressedValuesReplica, 0, len(replicas))
+	for _, replica := range replicas {
```
nikunjgit: this is pretty dense code so some comments would be nice. Also can you move some of this into smaller functions ?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,250 @@
...
+	opts checked.BytesOptions,
+	iteratorPools encoding.IteratorPools,
+) xio.BlockReader {
+	var checkedBytesWrapperPool xpool.CheckedBytesWrapperPool
+	if iteratorPools != nil {
+		checkedBytesWrapperPool = iteratorPools.CheckedBytesWrapper()
+	}
+
+	var head, tail checked.Bytes
+	if checkedBytesWrapperPool != nil {
```
nikunjgit: move this into a func which takes in iteratorPools and returns head, tail ?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,250 @@
...
+	segmentReader := xio.NewSegmentReader(segment)
+
+	return xio.BlockReader{
+		SegmentReader: segmentReader,
+		Start:         time.Time{},
+		BlockSize:     0,
+	}
+}
+
+var (
```
nikunjgit: when does this get initialized ? Maybe put this inside an init method ? Don't want this to run in cases when we are not using rpc

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,250 @@
...
+			b.Reset(nil)
+		}))
+
+	iterAlloc = func(r io.Reader) encoding.ReaderIterator {
+		return m3tsz.NewReaderIterator(r, m3tsz.DefaultIntOptimizationEnabled, encoding.NewOptions())
+	}
+)
+
+// SeriesIteratorFromRPC converts a grpc series to a seriesIterator
+func SeriesIteratorFromRPC(iteratorPools encoding.IteratorPools, timeSeries *rpc.Series) encoding.SeriesIterator {
```
nikunjgit: same as above: this is pretty dense code so some comments would be nice. Also can you move some of this into smaller functions ?

src/coordinator/generated/proto/rpc/query.proto
arnikola: Yeah, I'm doubling down on tags briefly; got some further work to actually add this to the write path which would consolidate the duplicate values

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: I should add a comment that `Start` and `BlockSize` are kinda irrelevant here? Might be a bad assumption to make now that I think about it

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+	err := tagIter.Err()
+	if err != nil {
+		return nil, err
+	}
+	return tags, nil
+}
+
+// CompressedProtobufFromSeriesIterator builds compressed rpc series from a SeriesIterator
+func CompressedProtobufFromSeriesIterator(it encoding.SeriesIterator) (*rpc.Series, error) {
```
nikunjgit: CompressedSeriesFromSeriesIterator ?

nikunjgit: can you add some high level comments on the steps involved to convert SeriesIterator to rpc.Series ?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -0,0 +1,298 @@
...
+			}
+		}
+		blockReaders[i] = blockReadersPerSegment
+	}
+
+	return blockReaders
+}
+
+// SeriesIteratorFromCompressedProtobuf creates a SeriesIterator from a compressed protobuf
+func SeriesIteratorFromCompressedProtobuf(iteratorPools encoding.IteratorPools, timeSeries *rpc.Series) encoding.SeriesIterator {
```
nikunjgit: SeriesIteratorFromCompressedSeries ?

nikunjgit: can you add some high level function comments here as well ?

PR #706, opened 2018-05-24
Add new docs structure

benraskin92:


nerd0:
Mind explaining in the description how these documents will stay in sync with website?

schallert:
LGTM x2

CK-Ward:
LGTM, just a thought, have we thought about putting anything on the [Wiki](https://github.com/m3db/m3db/wiki)

schallert:
> LGTM, just a thought, have we thought about putting anything on the Wiki

I hadn't thought of that. I think if our docs are comprehensive enough then they can hold "official" stuff, but maybe the wiki can be a place for less formal docs? Or maybe things more meta to the repo, like publishing docs and cutting releases. That seems to be what [prometheus](https://github.com/prometheus/prometheus/wiki) does.

PR #707, opened 2018-05-24
Allow * immutable, exactly 1 mutable segment during incremental index bootstrap

prateek:


prateek:
@robskillington fair yeah. opened https://github.com/m3db/m3db/issues/708 to move into a single volume, after which we won't have the per segment fulfilment issue because we'd be doing it across all segments in a block in a single place.

PR #716, opened 2018-05-25
Add ability to run just coordinator in m3dbnode binary

robskillington:


src/cmd/services/m3dbnode/main/main.go
```diff
@@ -53,21 +53,36 @@ func main() {
...
-	})
+	if cfg.DB != nil {
+		dbserver.Run(dbserver.RunOptions{
+			Config:          *cfg.DB,
+			ClientCh:        dbClientCh,
+			ClusterClientCh: clusterClientCh,
+		})
+	} else if cfg.Coordinator != nil {
+		<-coordinatorDoneCh
+	}
```
schallert: Should we print a warning if neither are configured?

robskillington: I'm ok for now to just quit at the moment, we can revisit this later - its a little difficult because we don't have a logger at this point and stderr/stdout is gross when at all other times we would use a logger.

PR #717, opened 2018-05-25
Swagger

justinjc:
* Serve OpenAPI (Swagger) documentation via [ReDoc](https://github.com/Rebilly/ReDoc) on the endpoint: `/api/v1/docs`
* Prefix all coordinator endpoints with `/api/v1`

**Todo:**
* Documentation for remaining endpoints


src/cmd/services/m3coordinator/handler/prometheus/native/read.go
```diff
@@ -38,7 +38,7 @@ import (
 
 const (
 	// PromReadURL is the url for native prom read handler
```
nerd0: Being a n00b, should a comment explain a bit more the difference between the native and remote endpoints?

justinjc: I'll let the author @benraskin92 update this one 😃

robskillington:
Nice, good to see swagger!  Wondering if we should find some way to verify the swagger spec is up to date (even if we can't autogenerate it per se). i.e. have some canned requests and make sure that the responses match the shape expected.  

Just a thought, we can open an issue for it and follow up later perhaps.

.gitignore
```diff
@@ -18,6 +18,9 @@ vendor/
 # Build binaries
 bin/
 
+# Used to serve swagger documentation
+!swagger/**/*.html
```
robskillington: Hm, does this has to be shipped around with the binary out of interest?  i.e. user needs binary, config file and swagger directory?

justinjc: Are you suggesting the docs be hosted some place instead of users self-hosting it? I think that's definitely possible as well. I'm not sure how useful having an offline spec would be for users.

justinjc:
I attempted to use [go-swagger](https://github.com/go-swagger/go-swagger) here to auto-generate the swagger spec, but it required reworking the code so that things could be annotated to create a correct document.

Testing the spec via canned responses would require a full integration test in order to go through the router logic as well. Perhaps another way to verify this would be to parse the swagger yaml and compare it to a [list of registered routes](https://github.com/gorilla/mux/issues/252). I'll make an issue to track.

src/cmd/services/m3coordinator/handler/openapi/openapi.go
```diff
@@ -0,0 +1,105 @@
...
+	if err != nil {
+		logger.Error("unable generate OpenAPI docs", zap.Any("error", err))
+		handler.Error(w, err, http.StatusInternalServerError)
+		return
+	}
+}
+
+// StaticHandler is the handler for serving static assets (including OpenAPI specs).
+func StaticHandler() http.Handler {
+	return http.StripPrefix(StaticURLPrefix, http.FileServer(http.Dir(staticDir)))
```
prateek: hm doing it like this requires the assets be available in a static directory on the filesystem. could you embed the assets in the binary to avoid this?

e.g. something like - https://github.com/elazarl/go-bindata-assetfs

.excludecoverage
```diff
@@ -6,3 +6,4 @@ tools/
 vendor/
 integration/
 testgen/
+openapi/
```
prateek: dunno the right answer here - is there a reason to prefer `openapi/` as a top level directory over `src/coordinator/openapi`?

src/cmd/services/m3coordinator/handler/common.go
```diff
@@ -32,6 +32,11 @@ import (
 	"go.uber.org/zap"
 )
 
+const (
+	// RoutePrefix is the prefix for all coordinator routes
+	RoutePrefix = "/api/v1"
```
prateek: nit: maybe `RoutePrefixV1` would make the usage clearer

prateek:
re: testing - how about something cheap to start?

e.g. Could add to what we do in `src/cmd/m3dbnode/main/*_test.go`, add validation the process exposes the swagger endpoint and returns a 200 on a get.

.excludecoverage
justinjc: I've put this under `src/coordinator/generated/assets/openapi/` now since it gets generated into embedded assets.

docs/how_to/single_node.md
```diff
@@ -29,7 +29,7 @@ Next, create an initial namespace for your metrics:
 <!-- TODO: link to config reference docs once available -->
 
 ```json
-curl -X POST localhost:7201/namespace/add -d '{
+curl -X POST localhost:7201/api/v1/namespace/add -d '{
```
prateek: mind checking if you need to make similar changes to Matt's kube guide too

docs/how_to/single_node.md
```diff
@@ -92,6 +92,11 @@ warn-level errors (prefixed with `[W]`) should not block bootstrapping.
 20:10:14.764771[I] successfully updated topology to 1 hosts
 ```
 
+The node also self-hosts its OpenAPI docs, outlining available endpoints. You can access this by
+going to `localhost:7201/api/v1/docs` in your browser.
+
+![OpenAPI Doc](redoc_screenshot.png)
```
prateek: Nice! Dunno if this is feasible, but could we also host offline version of the API docs on m3db.github.io or w/e?

justinjc: Should be doable. I'll put up a separate diff for that.

docs/how_to/single_node.md
justinjc: Done.

PR #724, opened 2018-05-26
/database/create endpoint

justinjc:
Setting up M3DB involves setting a lot of configuration for namespaces and placement before users can get up and running. The `/database/create` endpoint is an opinionated endpoint for users uninterested in setting each value and will default options to get them running quicker.

`POST`ing to this endpoint with the below payload will set up a namespace and placement for M3DB running on your local box.
```
{
  "namespaceName": "testns",
  "type": "local"
}
```


src/cmd/services/m3coordinator/handler/database/create.go
```diff
@@ -0,0 +1,251 @@
...
+		blockDataExpiry                          bool
+		blockDataExpiryAfterNotAccessPeriodNanos int64
+
+		indexEnabled        bool
+		indexBlockSizeNanos int64
+	)
+
+	switch validDBType {
+	case dbTypeLocal:
+		bootstrapEnabled = true
```
justinjc: Are these sane defaults?

src/cmd/services/m3coordinator/handler/database/create.go
```diff
@@ -0,0 +1,251 @@
...
+func defaultedPlacementInitRequest(r *admin.DatabaseCreateRequest) *admin.PlacementInitRequest {
+	var (
+		numShards         int32
+		replicationFactor int32
+		instances         []*placementpb.Instance
+	)
+
+	switch r.Type {
+	case dbTypeLocal:
+		numShards = 16
```
justinjc: Are these sane defaults?

src/cmd/services/m3coordinator/handler/database/create.go
```diff
@@ -0,0 +1,251 @@
...
+		retentionPeriodNanos = 172800000000000  // 48h
+		retentionBlockSizeNanos = 7200000000000 // 2h
+		bufferFutureNanos = 600000000000        // 10m
+		bufferPastNanos = 600000000000          // 10m
+		blockDataExpiry = true
+		blockDataExpiryAfterNotAccessPeriodNanos = 300000000000 // 5m
+
+		indexEnabled = false
+		indexBlockSizeNanos = 0
+	default:
```
justinjc: Open to suggestions on different types of database and how to default appropriately.

robskillington:
Nice, this will be greatly beneficial - going to go through this early tomorrow 🎉

src/cmd/services/m3coordinator/handler/database/create.go
robskillington: This looks good for a local DB, good stuff.

src/cmd/services/m3coordinator/handler/database/create.go
```diff
@@ -0,0 +1,251 @@
...
+
+		retentionPeriodNanos = 172800000000000  // 48h
+		retentionBlockSizeNanos = 7200000000000 // 2h
+		bufferFutureNanos = 600000000000        // 10m
+		bufferPastNanos = 600000000000          // 10m
+		blockDataExpiry = true
+		blockDataExpiryAfterNotAccessPeriodNanos = 300000000000 // 5m
+
+		indexEnabled = false
+		indexBlockSizeNanos = 0
```
robskillington: Hm, perhaps default indexEnabled to true and the block size to match the retentionBlockSizeNanos?  I think people will want to try out the reverse indexing, etc.

src/cmd/services/m3coordinator/handler/database/create.go
robskillington: Maybe we land this and we can add another type after?

justinjc: Done. Thanks for being specific in your suggestions 😄

src/cmd/services/m3coordinator/handler/database/create.go
```diff
@@ -0,0 +1,263 @@
...
+
+	// CreateHTTPMethod is the HTTP method used with this resource.
+	CreateHTTPMethod = "POST"
+
+	dbTypeLocal = "local"
+)
+
+var (
+	errMissingRequiredField = errors.New("all attributes must be set")
+
```
prateek: nit: why the new line here?

src/cmd/services/m3coordinator/handler/database/create.go
```diff
@@ -0,0 +1,263 @@
...
+	"github.com/m3db/m3db/src/coordinator/util/logging"
+	protonamespace "github.com/m3db/m3db/src/dbnode/generated/proto/namespace"
+
+	"github.com/golang/protobuf/jsonpb"
+	"go.uber.org/zap"
+)
+
+const (
+	// CreateURL is the url for the database create handler.
+	CreateURL = "/api/v1/database/create"
```
prateek: should probably use the v1 prefix yea?

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,249 @@
...
+
+	return namespaceAddRequest, placementInitRequest, nil
+}
+
+func defaultedNamespaceAddRequest(r *admin.DatabaseCreateRequest) *admin.NamespaceAddRequest {
+	options := dbnamespace.NewOptions()
+
+	switch r.Type {
+	case dbTypeLocal:
+		options.SetBootstrapEnabled(true)
```
prateek: afair, these are the default - any reason to explicitly set them here again?

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,249 @@
...
+)
+
+const (
+	// CreateURL is the url for the database create handler.
+	CreateURL = handler.RoutePrefixV1 + "/database/create"
+
+	// CreateHTTPMethod is the HTTP method used with this resource.
+	CreateHTTPMethod = "POST"
+
+	dbTypeLocal = "local"
```
prateek: nit: use an enum for this 

```

type dbType string

const (
  dbTypeLocal dbType = "local"
)

```

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,249 @@
...
+		options.SetFlushEnabled(true)
+		options.SetWritesToCommitLog(true)
+		options.SetCleanupEnabled(true)
+		options.SetRepairEnabled(false)
+		options.SetSnapshotEnabled(false)
+
+		// ExpectedSeriesDatapointsPerHour takes precedence over RetentionPeriodNanos
+		if r.ExpectedSeriesDatapointsPerHour > 0 {
+			// We want around 720 datapoints per block, so:
+			// (720 / r.ExpectedSeriesDatapointsPerHour) * 60 * 60000000000
```
prateek: could you use a const for `2592000000000000`?

src/coordinator/generated/proto/admin/database.proto
```diff
@@ -0,0 +1,24 @@
...
+message DatabaseCreateRequest {
+  // Required fields
+  string namespace_name = 1;
+  string type = 2;
+
+  // Optional fields that may be inferred depending on database type
+  int32 num_shards = 3;
+  int32 replication_factor = 4;
+  // The below two options are used to default retention options
+  int64 retention_period_nanos = 5;
```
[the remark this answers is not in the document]
justinjc: While I agree with this, I feel like it'll be weird to specify duration strings in this endpoint, but use nanos for the namespace endpoints. We might want to make the change to using durations across the board in a different diff?

src/coordinator/api/v1/handler/database/create.go
justinjc: I thought it would be safer to not assume that we'd want the defaults for a local deployment compared to a clustered version, so explicitly setting them would be better. Tell me if you disagree.

src/coordinator/generated/proto/admin/database.proto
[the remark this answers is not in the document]
prateek: Yeah we should definitely make the other endpoints easier to use too. Could probably use the same protobuf Duration type in the namespace endpoint too.

src/coordinator/api/v1/handler/database/create.go
prateek: I agree in spirt but there's nuance to it. E.g. consider that Snapshotting is currently disabled by default, but as soon as we finish productionalising it, we'd turn it on by default in the namespace.NewOptions(). It'd be better to not expect us to remember to have to come here to enable it too. 

Instead of overriding all properties, I think you should override only specific properties you think explicitly need to be different in the local v clustered version. Sound fair?

src/cmd/services/m3coordinator/handler/database/create.go
```diff
@@ -0,0 +1,249 @@
...
+
+		options.IndexOptions().SetEnabled(true)
+		options.IndexOptions().SetBlockSize(options.RetentionOptions().BlockSize())
+	default:
+		return nil
+	}
+
+	return &admin.NamespaceAddRequest{
+		Name: r.NamespaceName,
+		Options: &protonamespace.NamespaceOptions{
```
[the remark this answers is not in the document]
justinjc: There isn't one quite suitable. I'll make one there.

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,249 @@
...
+
+		options.IndexOptions().SetEnabled(true)
+		options.IndexOptions().SetBlockSize(options.RetentionOptions().BlockSize())
+	default:
+		return nil
+	}
+
+	return &admin.NamespaceAddRequest{
+		Name: r.NamespaceName,
+		Options: &protonamespace.NamespaceOptions{
```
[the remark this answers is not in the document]
justinjc: There isn't one quite suitable. I'll make one there.

src/coordinator/generated/proto/admin/database.proto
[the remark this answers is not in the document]
justinjc: How about a separate PR for that? 😄 That's a little more involved by changing `namespace.RetentionPeriodNanos` and the index one, as well as documentation and the code handling the conversion.

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,249 @@
...
+		options.SetWritesToCommitLog(true)
+		options.SetCleanupEnabled(true)
+		options.SetRepairEnabled(false)
+		options.SetSnapshotEnabled(false)
+
+		// ExpectedSeriesDatapointsPerHour takes precedence over RetentionPeriodNanos
+		if r.ExpectedSeriesDatapointsPerHour > 0 {
+			// We want around 720 datapoints per block, so:
+			// (720 / r.ExpectedSeriesDatapointsPerHour) * 60 * 60000000000
+			options.RetentionOptions().SetRetentionPeriod(time.Duration(2592000000000000 / r.ExpectedSeriesDatapointsPerHour))
```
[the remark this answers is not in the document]
justinjc: Good catch!

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,249 @@
...
+		} else {
+			options.RetentionOptions().SetRetentionPeriod(time.Duration(r.RetentionPeriodNanos))
+		}
+
```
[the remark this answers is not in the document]
justinjc: Any suggestions on what values? Do you think a better place for this would be at the options setter function level instead of here? Otherwise we'd have to duplicate this elsewhere as well.

src/coordinator/generated/proto/admin/database.proto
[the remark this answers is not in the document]
prateek: Definitely 👍

src/coordinator/api/v1/handler/database/create.go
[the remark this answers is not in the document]
prateek: Great cheers

src/coordinator/api/v1/handler/database/create.go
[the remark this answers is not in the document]
prateek: re: values - for a single node deploy it should probably be in the realm of min 30m and max 2h. 

Re: where this belongs - i think this code has the most context about how the cluster is being used. How about we start by putting these limits here and see how it evolves over time?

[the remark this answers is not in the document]
justinjc: Sounds good!

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,230 @@
...
+	// CreateURL is the url for the database create handler.
+	CreateURL = handler.RoutePrefixV1 + "/database/create"
+
+	// CreateHTTPMethod is the HTTP method used with this resource.
+	CreateHTTPMethod = "POST"
+
+	// We want around 720 datapoints per block, so:
+	// (720 / ExpectedSeriesDatapointsPerHour) * 60 * 60000000000
+	// == 2592000000000000 / ExpectedSeriesDatapointsPerHour
+	blockSizeFromExpectedSeriesScalar = 2592000000000000
```
prateek: maybe do the math inline too? i.e.

```
var (
  idealDatapointsPerBlock = 720
  blockSizeFromExpectedSeriesScalar = idealDatapointsPerBlock * int(time.Hour)
)
```

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,230 @@
...
+			options.RetentionOptions().SetBlockSize(1 * time.Hour)
+		} else {
+			options.RetentionOptions().SetBlockSize(2 * time.Hour)
+		}
+
+		options.IndexOptions().SetEnabled(true)
+		options.IndexOptions().SetBlockSize(options.RetentionOptions().BlockSize())
+	default:
+		// This function assumes the type is valid
+		return nil
```
prateek: nit: panic or propagate an error back up?

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,230 @@
...
+				Zone:           "embedded",
+				Weight:         1,
+				Endpoint:       "http://localhost:9000",
+				Hostname:       "localhost",
+				Port:           9000,
+			},
+		}
+	default:
+		// This function assumes the type is valid
+		return nil
```
prateek: same here

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,230 @@
...
+	case dbTypeLocal:
+		numShards = 16
+		replicationFactor = 1
+		instances = []*placementpb.Instance{
+			&placementpb.Instance{
+				Id:             "localhost",
+				IsolationGroup: "local",
+				Zone:           "embedded",
+				Weight:         1,
+				Endpoint:       "http://localhost:9000",
```
prateek: hm can you take a parameter for this port? could be different in the config

src/dbnode/storage/namespace/convert.go
```diff
@@ -160,3 +135,30 @@ func FromProto(protoRegistry nsproto.Registry) (Map, error) {
 	}
 	return NewMap(metadatas)
 }
+
+// OptionsToProto converts Options -> nsproto.NamespaceOptions
+func OptionsToProto(opts Options) *nsproto.NamespaceOptions {
```
prateek: +1 for this

src/coordinator/api/v1/handler/namespace/delete.go
```diff
@@ -38,10 +38,13 @@ import (
 
 const (
 	namespaceIDVar = "id"
+
+	// DeleteHTTPMethod is the HTTP method used with this resource.
+	DeleteHTTPMethod = "DELETE"
```
prateek: hm would it be better to create an enum for this type - GET/DELETE/POST/PUT/...  ?

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,252 @@
...
+		return nil, nil, handler.NewParseError(err, http.StatusBadRequest)
+	}
+
+	return namespaceAddRequest, placementInitRequest, nil
+}
+
+func defaultedNamespaceAddRequest(r *admin.DatabaseCreateRequest) (*admin.NamespaceAddRequest, error) {
+	options := dbnamespace.NewOptions()
+
+	fmt.Printf("GOT THIS TYPE:%v\n", r.Type)
```
prateek: Undo println?

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -0,0 +1,252 @@
...
+	case dbTypeLocal:
+		numShards = 16
+		replicationFactor = 1
+		instances = []*placementpb.Instance{
+			&placementpb.Instance{
+				Id:             "localhost",
+				IsolationGroup: "local",
+				Zone:           "embedded",
+				Weight:         1,
+				Endpoint:       "http://localhost:" + string(port),
```
prateek: nit: Use `fmt.Sprintf(...)` instead

PR #726, opened 2018-05-29
Convert m3db blocks

benraskin92:


src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+
+// BlockReplica contains the replicas for a single m3db block
+type BlockReplica struct {
+	Start     time.Time
+	BlockSize time.Duration
+	Replicas  []encoding.MultiReaderIterator
+}
+
+func convertM3DBSeriesIterators(iterators encoding.SeriesIterators) ([]SeriesBlocks, error) {
+	iterAlloc := func(r io.Reader) encoding.ReaderIterator {
```
prateek: take as external parameter

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/m3db/m3db/src/dbnode/x/xio"
+)
+
+// BlockReplica contains the replicas for a single m3db block
+type BlockReplica struct {
+	Start     time.Time
+	BlockSize time.Duration
+	Replicas  []encoding.MultiReaderIterator
+}
+
```
prateek: what's the public entry point into the conversion code?

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+func convertM3DBSeriesIterators(iterators encoding.SeriesIterators) ([]SeriesBlocks, error) {
+	iterAlloc := func(r io.Reader) encoding.ReaderIterator {
+		iter := m3tsz.NewDecoder(true, encoding.NewOptions())
+		return iter.Decode(r)
+	}
+
+	var multiSeriesBlocks []SeriesBlocks
```
prateek: pre-alloc the slice

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+}
+
+func blockReplicasFromSeriesIterator(seriesIterator encoding.SeriesIterator, iterAlloc encoding.ReaderIteratorAllocate) ([]BlockReplica, error) {
+	var blockReplicas []BlockReplica
+	for _, replica := range seriesIterator.Replicas() {
+		perBlockSliceReaders := replica.Readers()
+		next := true
+		for next {
+			l, start, bs := perBlockSliceReaders.CurrentReaders()
+			var readers []xio.SegmentReader
```
prateek: here too

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+		series := seriesBlocksFromBlockReplicas(blockReplicas, seriesIterator)
+
+		multiSeriesBlocks = append(multiSeriesBlocks, series)
+	}
+
+	return multiSeriesBlocks, nil
+}
+
+func blockReplicasFromSeriesIterator(seriesIterator encoding.SeriesIterator, iterAlloc encoding.ReaderIteratorAllocate) ([]BlockReplica, error) {
+	var blockReplicas []BlockReplica
```
prateek: here too

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+
+	for _, seriesIterator := range iterators.Iters() {
+
+		blockReplicas, err := blockReplicasFromSeriesIterator(seriesIterator, iterAlloc)
+		if err != nil {
+			return []SeriesBlocks{}, err
+		}
+
+		series := seriesBlocksFromBlockReplicas(blockReplicas, seriesIterator)
+
```
arnikola: nit newline

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+func convertM3DBSeriesIterators(iterators encoding.SeriesIterators) ([]SeriesBlocks, error) {
+	iterAlloc := func(r io.Reader) encoding.ReaderIterator {
+		iter := m3tsz.NewDecoder(true, encoding.NewOptions())
+		return iter.Decode(r)
+	}
+
+	var multiSeriesBlocks []SeriesBlocks
+
+	for _, seriesIterator := range iterators.Iters() {
+
```
arnikola: nit newline

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+	return multiSeriesBlocks, nil
+}
+
+func blockReplicasFromSeriesIterator(seriesIterator encoding.SeriesIterator, iterAlloc encoding.ReaderIteratorAllocate) ([]BlockReplica, error) {
+	var blockReplicas []BlockReplica
+	for _, replica := range seriesIterator.Replicas() {
+		perBlockSliceReaders := replica.Readers()
+		next := true
+		for next {
```
arnikola: nit: `for next := true; next; next = perBlockSliceReaders.Next() {`

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+			l, start, bs := perBlockSliceReaders.CurrentReaders()
+			var readers []xio.SegmentReader
+			for i := 0; i < l; i++ {
+				reader := perBlockSliceReaders.CurrentReaderAt(i)
+				// import to clone the reader as we need its position reset before
+				// we use the contents of it again
+				clonedReader, err := reader.Clone()
+				if err != nil {
+					return []BlockReplica{}, err
```
arnikola: nit
return nil, err

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+		return iter.Decode(r)
+	}
+
+	var multiSeriesBlocks []SeriesBlocks
+
+	for _, seriesIterator := range iterators.Iters() {
+
+		blockReplicas, err := blockReplicasFromSeriesIterator(seriesIterator, iterAlloc)
+		if err != nil {
+			return []SeriesBlocks{}, err
```
arnikola: nit
`return nil, err`

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,138 @@
...
+					return []BlockReplica{}, err
+				}
+				readers = append(readers, clonedReader)
+			}
+			// todo(braskin): pooling
+			iter := encoding.NewMultiReaderIterator(iterAlloc, nil)
+			iter.Reset(readers, start, bs)
+
+			inserted := false
+			for i := range blockReplicas {
```
arnikola: Would using a map with startTime as the key like `[time.Time][]BlockReplica` be better?

Alternatively, can do some maths using the Start and BlockSize to figure out which index this should go in?

src/coordinator/ts/m3db/convert_test.go
```diff
@@ -0,0 +1,192 @@
...
+)
+
+func TestConversion(t *testing.T) {
+	iterator := BuildTestSeriesIterator(t)
+	iterators := encoding.NewSeriesIterators([]encoding.SeriesIterator{iterator}, nil)
+	blocks, err := convertM3DBSeriesIterators(iterators)
+	require.NoError(t, err)
+
+	for _, block := range blocks {
+		assert.Equal(t, "foo", block.ID.String())
```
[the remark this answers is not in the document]
arnikola: Might be a good idea to change the constants in the test file to some kind of struct that we can share around and use values from, including the start, middle, end etc

src/coordinator/ts/m3db/convert_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+		blockOne := block.Blocks[0].SeriesIterator
+		blockTwo := block.Blocks[1].SeriesIterator
+
+		start = time.Now().Truncate(time.Hour).Add(2 * time.Minute)
+		middle = start.Add(30 * time.Minute)
+		end = middle.Add(30 * time.Minute)
+
+		assert.Equal(t, start, blockOne.Start())
+		// assert.Equal(t, middle, blockOne.End())
```
arnikola: Uncomment, should be truncated start + blockSize

src/coordinator/ts/m3db/convert_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+		for i := 1; blockOne.Next(); i++ {
+			dp, _, _ := blockOne.Current()
+			assert.Equal(t, float64(i), dp.Value)
```
arnikola: Check the data point timestamp too? Also this will have to change when the other PR drops since [1,2] should not appear

src/coordinator/ts/m3db/convert_test.go
```diff
@@ -0,0 +1,192 @@
...
+	iterator := BuildTestSeriesIterator(t)
+	iterators := encoding.NewSeriesIterators([]encoding.SeriesIterator{iterator}, nil)
+	blocks, err := convertM3DBSeriesIterators(iterators)
+	require.NoError(t, err)
+
+	for _, block := range blocks {
+		assert.Equal(t, "foo", block.ID.String())
+		assert.Equal(t, "namespace", block.Namespace.String())
+
+		blockOne := block.Blocks[0].SeriesIterator
```
arnikola: nit `blockOneSeriesIterator`

src/coordinator/ts/m3db/convert_test.go
arnikola: Also check start/end for blockTwo

src/coordinator/ts/m3db/convert_test.go
```diff
@@ -0,0 +1,192 @@
...
+		for i := 1; blockOne.Next(); i++ {
+			dp, _, _ := blockOne.Current()
+			assert.Equal(t, float64(i), dp.Value)
+		}
+
+		for i := 101; blockTwo.Next(); i++ {
+			dp, _, _ := blockTwo.Current()
+			assert.Equal(t, float64(i), dp.Value)
+		}
+	}
```
arnikola: Verify `.Tags()`, `.ID()`, `.Namespace()` for both blocks

src/coordinator/ts/m3db/types.go
```diff
@@ -37,7 +37,7 @@ type SeriesBlock struct {
 // SeriesBlocks is a placeholder until it is added to M3DB
 type SeriesBlocks struct {
 	ID        ident.ID
-	Namespace string
+	Namespace ident.ID
 	Blocks    []SeriesBlock
```
arnikola: Maybe add Tags?

src/coordinator/ts/m3db/convert.go
arnikola: `defer iterators.Close()`... Although may have to clone ID, Namespace, and Tags when you build each new SeriesIterator (also make sure you use the cloned version when you build a SeriesBlocks 

I think the rest should be safe, but double check with @prateek  to confirm

src/coordinator/ts/m3db/convert_test.go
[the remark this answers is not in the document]
arnikola: (see https://github.com/m3db/m3db/pull/736)

src/coordinator/ts/m3db/convert.go
benraskin92: Per our convo, I'll hold off on this for now.

src/coordinator/ts/m3db/convert_test.go
benraskin92: You think we need to check the timestamp of each datapoint even if we verify the block start and end?

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,144 @@
...
+
+	for i, seriesIterator := range iterators.Iters() {
+		blockReplicas, err := blockReplicasFromSeriesIterator(seriesIterator, iterAlloc)
+		if err != nil {
+			return nil, err
+		}
+
+		series := seriesBlocksFromBlockReplicas(blockReplicas, seriesIterator)
+		multiSeriesBlocks[i] = series
+
```
arnikola: nit newline

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,144 @@
...
+	Start     time.Time
+	BlockSize time.Duration
+	Replicas  []encoding.MultiReaderIterator
+}
+
+// ConvertM3DBSeriesIterators takes in series iterators from m3db and returns
+// coordinator SeriesBlocks which are used to construct Blocks for query processing.
+func ConvertM3DBSeriesIterators(iterators encoding.SeriesIterators, iterAlloc encoding.ReaderIteratorAllocate) ([]SeriesBlocks, error) {
+	multiSeriesBlocks := make([]SeriesBlocks, iterators.Len())
+	defer iterators.Close()
```
arnikola: nit defer above the var declaration?

src/coordinator/ts/m3db/convert_test.go
```diff
@@ -0,0 +1,92 @@
...
+
+		blockOneSeriesIterator := block.Blocks[0].SeriesIterator
+		blockTwoSeriesIterator := block.Blocks[1].SeriesIterator
+
+		assert.Equal(t, start.Add(2*time.Minute), blockOneSeriesIterator.Start())
+		assert.Equal(t, middle, blockOneSeriesIterator.End())
+
+		for i := 3; blockOneSeriesIterator.Next(); i++ {
+			dp, _, _ := blockOneSeriesIterator.Current()
+			assert.Equal(t, float64(i), dp.Value)
```
arnikola: Check dp.Timestamp?
Should be something like
`assert.Equal(t, start.Add(time.Duration(i)*time.Minute), dp.Timestamp)`

src/coordinator/ts/m3db/convert_test.go
```diff
@@ -0,0 +1,92 @@
...
+		}
+
+		assert.Equal(t, middle, blockTwoSeriesIterator.Start())
+		assert.Equal(t, end, blockTwoSeriesIterator.End())
+
+		for i := 101; blockTwoSeriesIterator.Next(); i++ {
+			dp, _, _ := blockTwoSeriesIterator.Current()
+			assert.Equal(t, float64(i), dp.Value)
```
arnikola: here, too

src/coordinator/ts/m3db/convert_test.go
```diff
@@ -0,0 +1,92 @@
...
+	testTags = test.TestTags
+
+	blockSize = test.BlockSize
+
+	start       = test.Start
+	seriesStart = test.SeriesStart
+	middle      = test.Middle
+	end         = test.End
+
+	// required for iterator pool
```
arnikola: nit comment is a bit off

src/coordinator/ts/m3db/convert_test.go
```diff
@@ -0,0 +1,92 @@
...
+			assert.Equal(t, float64(i), dp.Value)
+		}
+
+		assert.Equal(t, middle, blockTwoSeriesIterator.Start())
+		assert.Equal(t, end, blockTwoSeriesIterator.End())
+
+		for i := 101; blockTwoSeriesIterator.Next(); i++ {
+			dp, _, _ := blockTwoSeriesIterator.Current()
+			assert.Equal(t, float64(i), dp.Value)
+		}
```
arnikola: Check the tags for each iterator too?

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,144 @@
...
+func blockReplicasFromSeriesIterator(seriesIterator encoding.SeriesIterator, iterAlloc encoding.ReaderIteratorAllocate) ([]BlockReplica, error) {
+	blockReplicas := make([]BlockReplica, 0, blockReplicaLen)
+	for _, replica := range seriesIterator.Replicas() {
+		perBlockSliceReaders := replica.Readers()
+		for next := true; next; next = perBlockSliceReaders.Next() {
+			l, start, bs := perBlockSliceReaders.CurrentReaders()
+			readers := make([]xio.SegmentReader, l)
+			for i := 0; i < l; i++ {
+				reader := perBlockSliceReaders.CurrentReaderAt(i)
+				// import to clone the reader as we need its position reset before
```
arnikola: nit: important, also I think we usually preface with `NB (username):`

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,144 @@
...
+				blockReplicas = append(blockReplicas, BlockReplica{
+					Start:     start,
+					BlockSize: bs,
+					Replicas:  []encoding.MultiReaderIterator{iter},
+				})
+			}
+		}
+	}
+
+	return blockReplicas, nil
```
arnikola: I'm not sure this is a real issue, but we *may* want to sort the blockReplicas by start size? Could potentially be a weird situation where the first replica missed a block for some reason and the result is out of order? Might want to double check with someone who knows the db a bit better for this

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	for _, block := range blockReplicas {
+		filterValuesStart := seriesIterator.Start()
+		if block.Start.After(filterValuesStart) {
+			filterValuesStart = block.Start
+		}
+
+		end := block.Start.Add(block.BlockSize)
+
```
arnikola: nit: newline

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,168 @@
...
+				})
+			}
+		}
+	}
+
+	return blockReplicas, nil
+}
+
+func seriesBlocksFromBlockReplicas(blockReplicas []BlockReplica, seriesIterator encoding.SeriesIterator) (SeriesBlocks, error) {
+	var (
```
arnikola: nit: might be good to add a note as to why we need to clone here

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,168 @@
...
+
+		end := block.Start.Add(block.BlockSize)
+
+		filterValuesEnd := seriesIterator.End()
+		if end.Before(filterValuesEnd) {
+			filterValuesEnd = end
+		}
+
+		// todo(braskin): pooling
+		valuesIter := encoding.NewSeriesIterator(clonedID, clonedNamespace,
```
arnikola: We *may* run into some difficulty down the line if we're using the id/namespace/tags form the series directly rather than from the SeriesBlock, since `.Close()` on any iterator will free the bytes used by the rest as well, which is unexpected behaviour

If we only use the values from the SeriesBlock itself we should be safe, but if there's a chance we'd be using the seriesIterators directly, it's safer to clone each of these.

Either way should put a comment noting that we should be really careful using this if anything is shared

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/m3db/m3x/ident"
+)
+
+const (
+	// todo(braskin): replace this with actual size if we can
+	blockReplicaLen = 10
+)
+
+// BlockReplica contains the replicas for a single m3db block
+type BlockReplica struct {
```
arnikola: nit: we can leave this unexported

arnikola: (Same with the fields)

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,168 @@
...
+	Start     time.Time
+	BlockSize time.Duration
+	Replicas  []encoding.MultiReaderIterator
+}
+
+// ConvertM3DBSeriesIterators takes in series iterators from m3db and returns
```
arnikola: nit: Something like `ConvertM3DBSeriesIterators converts m3db SeriesIterators to SeriesBlocks` might be more go-ey

src/coordinator/ts/m3db/types.go
```diff
@@ -37,7 +37,8 @@ type SeriesBlock struct {
 // SeriesBlocks is a placeholder until it is added to M3DB
 type SeriesBlocks struct {
```
arnikola: Could be useful to have a small description of what's in a SeriesBlock, since it's a pretty complicated object

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,170 @@
...
+	clonedTags, err := cloneTagIterator(seriesIterator.Tags())
+	if err != nil {
+		return SeriesBlocks{}, err
+	}
+
+	series := SeriesBlocks{
+		ID:        clonedID,
+		Namespace: clonedNamespace,
+		Tags:      clonedTags,
+	}
```
arnikola: `Blocks := make([]SeriesBlock, 0, len(blockReplicas),`

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,170 @@
...
+
+	clonedTags, err := cloneTagIterator(seriesIterator.Tags())
+	if err != nil {
+		return SeriesBlocks{}, err
+	}
+
+	series := SeriesBlocks{
+		ID:        clonedID,
+		Namespace: clonedNamespace,
+		Tags:      clonedTags,
```
arnikola: `Blocks: make([]SeriesBlock, 0, len(blockReplicas)),`

src/coordinator/ts/m3db/types.go
```diff
@@ -27,17 +27,20 @@ import (
 	"github.com/m3db/m3x/ident"
 )
 
-// SeriesBlock is a placeholder until it is added to M3DB
+// SeriesBlock contains the individual series iterators
 type SeriesBlock struct {
 	Start          time.Time
 	End            time.Time
 	SeriesIterator encoding.SeriesIterator
```
arnikola: Might be good to make the SeriesIterator private to ensure you can't do something weird like close them

src/coordinator/ts/m3db/types.go
```diff
@@ -27,17 +27,20 @@ import (
...
+// SeriesBlocks contain information about the timeseries that gets returned from m3db.
+// This includes meta data such as the ID, namespace and tags as well as the actual
+// series iterators that contain the datapoints.
 type SeriesBlocks struct {
 	ID        ident.ID
-	Namespace string
+	Namespace ident.ID
+	Tags      ident.TagIterator
 	Blocks    []SeriesBlock
 }
```
arnikola: Add a Close() method to SeriesBlocks?

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,170 @@
...
+	"time"
+
+	"github.com/m3db/m3db/src/dbnode/encoding"
+	"github.com/m3db/m3db/src/dbnode/x/xio"
+	"github.com/m3db/m3x/ident"
+)
+
+const (
+	// todo(braskin): replace this with actual size if we can
+	blockReplicaLen = 10
```
nikunjgit: should be fine to use a default. Maybe call it  initBlockReplicaLength ?

src/coordinator/ts/m3db/convert.go
```diff
@@ -0,0 +1,170 @@
...
+	for _, replica := range seriesIterator.Replicas() {
+		perBlockSliceReaders := replica.Readers()
+		for next := true; next; next = perBlockSliceReaders.Next() {
+			l, start, bs := perBlockSliceReaders.CurrentReaders()
+			readers := make([]xio.SegmentReader, l)
+			for i := 0; i < l; i++ {
+				reader := perBlockSliceReaders.CurrentReaderAt(i)
+				// NB(braskin): important to clone the reader as we need its position reset before
+				// we use the contents of it again
+				clonedReader, err := reader.Clone()
```
nikunjgit: how do we test this condition ?

src/coordinator/ts/m3db/types.go
benraskin92: Won't we need to access them when we process the query later on?

src/coordinator/ts/m3db/convert.go
arnikola: You could read the initial seriesIterator passed in; if you get values out of it, the clone is successful

src/coordinator/ts/m3db/types.go
arnikola: That's a problem for future us ;)

Would be good to ensure we don't accidentally close it if we do need to expose them; e.g. the SeriesBlock could just expose a `.Current() (ts.Datapoint, xtime.Unit, ts.Annotation)` and `.Next() bool` method which would defer to its seriesIterator

PR #728, opened 2018-05-29
Convert raw points into fix resolution datapoints

nikunjgit:


src/coordinator/ts/values.go
```diff
@@ -133,3 +139,39 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
 		values:        values,
 	}
 }
+
+// At every resolution timestep, the last sample value is chosen
+func RawPointsToFixedStep(datapoints Datapoints, start time.Time, end time.Time, interval time.Duration) (FixedResolutionMutableValues, error) {
+	if end.Before(start) {
+		return nil, fmt.Errorf("start cannot be after end, start: %v, end: %v", start, end)
+	}
```
arnikola: check interval != 0?

src/coordinator/ts/values.go
```diff
@@ -133,3 +139,39 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
...
+	numSteps := int(end.Sub(start) / interval)
```
arnikola: Might be better to do this so you're not calculating unnecessarily
```
var numSteps int
if end == start{
 numSteps = 1
} else {
 numSteps = int(end.Sub(start) / interval
}
```

src/coordinator/ts/values_test.go
```diff
@@ -0,0 +1,114 @@
...
+		t = t.Add(interval)
+	}
+	return dps
+}
+func createExamples() []testSample {
+	now := time.Time{}
+	samples := []testSample{
+		{
+			input:       generateDatapoints(now.Add(-5*time.Second), time.Second, 5),
+			expected:    []float64{0, 1, 2, 3, 4,},
```
arnikola: nit: Don't need the trailing `,` in the array if it's on a single line

src/coordinator/ts/values_test.go
```diff
@@ -0,0 +1,114 @@
...
+	hasNans     bool
+	description string
+}
+
+func generateDatapoints(start time.Time, interval time.Duration, num int) Datapoints {
+	dps := make(Datapoints, num)
+	t := start
+	for i := 0; i < num; i++ {
+		dps[i] = Datapoint{
+			Timestamp: t,
```
arnikola: nit `Timestamp: t.Add(interval * i)`

src/coordinator/ts/values.go
```diff
@@ -133,3 +139,39 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
...
+	}
+	fixStepValues := newFixedStepValues(interval, numSteps, math.NaN(), start)
+	currIdx := 0
+	idx := 0
+	for t := start; !t.After(end) && currIdx < numSteps; t = t.Add(interval) {
+		// Find first datapoint not before time t
+		for ; idx < len(datapoints) && datapoints.DatapointAt(idx).Timestamp.Before(t); idx++ {
+		}
+
+		if idx >= len(datapoints) {
```
arnikola: nit: idx can never be greater than len(datapoints)

src/coordinator/ts/values.go
arnikola: Also, is it appropriate to always round down? Maybe have an option to take the ceil instead?

src/coordinator/ts/values.go
```diff
@@ -133,3 +139,39 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
...
+	numSteps := int(end.Sub(start) / interval)
+	if end == start {
+		numSteps = 1
+	}
+	fixStepValues := newFixedStepValues(interval, numSteps, math.NaN(), start)
+	currIdx := 0
+	idx := 0
+	for t := start; !t.After(end) && currIdx < numSteps; t = t.Add(interval) {
+		// Find first datapoint not before time t
+		for ; idx < len(datapoints) && datapoints.DatapointAt(idx).Timestamp.Before(t); idx++ {
```
arnikola: I think this would read nicer as
```golang
numPoints := len(datapoints)
for ; idx < numPoints; idx++ {
 if !datapoints.DatapointAt(idx).Timestamp.Before(t) {
  break
 }
}
```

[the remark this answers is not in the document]
arnikola: This way you keep a running index and won't have to revisit all past points on each step through the outer loop

src/coordinator/ts/values.go
```diff
@@ -133,3 +139,39 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
...
+		}
+
+		if idx >= len(datapoints) {
+			fixStepValues.values[currIdx] = math.NaN()
```
arnikola: nit: can shortcircuit a bit
```golang
{
 for ; currIdx < numSteps; currIdx++ {
  fixStepValues.values[currIdx] = math.NaN()
 }
 return fixStepValues, nil
}

File: src/coordinator/ts/values.go [hunk 8]
```diff
@@ -133,3 +139,39 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
...
+			continue
+		}
+
+		// If datapoint aligns to the time or its the first datapoint then take that
+		if datapoints.DatapointAt(idx).Timestamp == t || idx == 0 {
+			fixStepValues.values[currIdx] = datapoints.ValueAt(idx)
+			currIdx++
+		} else {
+			fixStepValues.values[currIdx] = datapoints.ValueAt(idx - 1)
+			currIdx++
```
arnikola: nit: take currIdx++ out of the if/else

src/coordinator/ts/values.go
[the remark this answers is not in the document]
nikunjgit: made it a little clearer now with naming

src/coordinator/ts/values_test.go
nikunjgit: yeah, I personally like to make it explicit.

src/coordinator/ts/values.go
nikunjgit: based on L157, idx can be = len(datapoints). Since we maintain idx across multiple for loops, i'm just using it as sanity check

src/coordinator/ts/values.go
nikunjgit: good point, I think we make a lot of these decisions right now. I kept it same as how prometheus handles it for now. In the future, if we need that option, we can change the function to take in start, numSteps instead of end. But for now, doesn't seem likely we need that.

src/coordinator/ts/values.go
```diff
@@ -133,3 +141,51 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
...
+	if end.Before(start) {
+		return nil, fmt.Errorf("start cannot be after end, start: %v, end: %v", start, end)
+	}
+
+	if interval == 0 {
+		return nil, errors.New("interval cannot be 0")
+	}
+
+	var numSteps int
+	if end == start {
```
arnikola: nit: `end.Equal(start)`, using `==` for `time.Time` is a bit sketchy iirc

src/coordinator/ts/values.go
```diff
@@ -133,3 +141,51 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
...
 }
+
+// At every resolution timestep, the last sample value is chosen
+func RawPointsToFixedStep(datapoints Datapoints, start time.Time, end time.Time, interval time.Duration) (FixedResolutionMutableValues, error) {
+	if end.Before(start) {
+		return nil, fmt.Errorf("start cannot be after end, start: %v, end: %v", start, end)
+	}
+
+	if interval == 0 {
+		return nil, errors.New("interval cannot be 0")
```
arnikola: nit: put this in errors/storage.go?

src/coordinator/ts/values.go
```diff
@@ -133,3 +141,51 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
...
+	for t := start; !t.After(end) && fixedResIdx < numSteps; t = t.Add(interval) {
+		// Find first datapoint not before time t
+		for ; dpIdx < numPoints; dpIdx++ {
+			if !datapoints.DatapointAt(dpIdx).Timestamp.Before(t) {
+				break
+			}
+		}
+
+		if dpIdx >= numPoints {
+			fixStepValues.values[fixedResIdx] = math.NaN()
```
arnikola: nit: worth shortcutting?
```golang
{
 for ; currIdx < numSteps; currIdx++ {
  fixStepValues.values[currIdx] = math.NaN()
 }
 break
}
```

src/coordinator/ts/values.go
```diff
@@ -133,3 +141,51 @@ func NewFixedStepValues(millisPerStep time.Duration, numSteps int, initialValue
...
+	}
+
+	fixStepValues := newFixedStepValues(interval, numSteps, math.NaN(), start)
+	fixedResIdx := 0
+	dpIdx := 0
+	numPoints := len(datapoints)
+	for t := start; !t.After(end) && fixedResIdx < numSteps; t = t.Add(interval) {
+		// Find first datapoint not before time t
+		for ; dpIdx < numPoints; dpIdx++ {
+			if !datapoints.DatapointAt(dpIdx).Timestamp.Before(t) {
```
arnikola: Using `datapoints.DatapointAt(dpIdx).Timestamp` pretty often, can possibly pull it out e.g.

```golang
var timestamp time.Time
for t := start; !t.After(end) && fixedResIdx < numSteps; t = t.Add(interval) {
 for ; dpIdx < numPoints; dpIdx++ {
  timestamp = datapoints.DatapointAt(dpIdx).Timestamp
  if timestamp.Before(t) {...
```

src/coordinator/ts/values.go
nikunjgit: oh yeah, actually that's not needed. I added a comment to make it clear

src/coordinator/ts/values.go
nikunjgit: its used twice and as you rightly noted, we would have to define it outside the for loop. It's a bit ugly so i'll leave it as is unless you show a strong inclination

src/coordinator/ts/values.go
```diff
@@ -21,7 +21,11 @@
 package ts
 
 import (
+	"fmt"
+	"math"
```
benraskin92: 👍

src/coordinator/errors/storage.go
```diff
@@ -42,4 +42,8 @@ var (
 
 	// ErrInvalidFetchResult is an error returned when fetch result is invalid.
 	ErrInvalidFetchResult = errors.New("invalid fetch result")
+
+	// ErrZeroInterval is an error returned when fetch interval is 0.
+	ErrZeroInterval = errors.New("interval cannot be 0")
+
```
arnikola: nit newline

PR #729, opened 2018-05-29
[docs] Dockerized docs build with Make shortcuts

schallert:


Makefile
```diff
@@ -157,6 +157,22 @@ release-snapshot: install-goreleaser
 	@source $(GO_BUILD_LDFLAGS_CMD) > /dev/null && goreleaser --snapshot --rm-dist
 
 
+.PHONY: docs-container
+docs-container:
+	docker build -t m3db-docs -f hack/Dockerfile-docs docs
```
prateek: maybe just add sanity checks to ensure `docker` is installed and running

hack/Dockerfile-docs
```diff
@@ -0,0 +1,10 @@
+# Dockerfile for building docs is stored in a separate dir from the docs,
+# otherwise the generated site will unnecessarily contain the Dockerfile
+
+FROM python:3.5-alpine
+LABEL maintainer="The M3DB Authors <m3db@googlegroups.com>"
+
+WORKDIR /m3db
+EXPOSE 8000
+RUN pip install mkdocs==0.17.3 mkdocs-material==2.7.3
```
benraskin92: Should we just install the latest `mkdocs`? or make this configurable?

Makefile
```diff
@@ -157,6 +157,22 @@ release-snapshot: install-goreleaser
...
+
+.PHONY: docs-build
+docs-build: docs-container
+	docker run -v $(PWD):/m3db --rm m3db-docs "mkdocs build -e docs/theme -t material"
```
benraskin92: Does this mean you need to run it from a particular directory? Would it be better to use `$GOPATH`?

PR #732, opened 2018-05-29
Fix flaky tests: TestConfig and TestEmbeddedConfig

richardartoul:
- [X] Refactor the admin session interface to return a topology.Map instead of topology.Topology (we need a static view of the sessions topology map state, we can't use our own Topology watch)
- [X] Wait for all shards to be available in the session before attempting writes in the flaky tests

### DIFF
--- a/src/cmd/services/m3dbnode/main/main_test.go
+++ b/src/cmd/services/m3dbnode/main/main_test.go
@@ -649,7 +649,6 @@ func waitUntilAllShardsAreAvailable(t *testing.T, session client.AdminSession) {
 		)
 
 		if len(hostShardSets) == 0 {
-			fmt.Println("no host shard sets")
 			// We haven't received an actual topology yet.
 			continue
 		}
--- a/src/cmd/services/m3dbnode/main/main_test.go
+++ b/src/cmd/services/m3dbnode/main/main_test.go
@@ -30,11 +30,10 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/m3db/m3cluster/shard"
-
 	"github.com/m3db/m3cluster/integration/etcd"
 	"github.com/m3db/m3cluster/placement"
 	"github.com/m3db/m3cluster/services"
+	"github.com/m3db/m3cluster/shard"
 	"github.com/m3db/m3db/src/cmd/services/m3dbnode/config"
 	"github.com/m3db/m3db/src/cmd/services/m3dbnode/server"
 	"github.com/m3db/m3db/src/dbnode/client"
@@ -639,13 +638,15 @@ db:
 // that the session is currently storing contains a non-zero number of host shard sets, and if so,
 // makes sure that all their shard states are Available.
 func waitUntilAllShardsAreAvailable(t *testing.T, session client.AdminSession) {
+outer:
 	for {
+		time.Sleep(10 * time.Millisecond)
+
 		topoMap, err := session.TopologyMap()
 		require.NoError(t, err)
 
 		var (
-			hostShardSets         = topoMap.HostShardSets()
-			allShardsAreAvailable = true
+			hostShardSets = topoMap.HostShardSets()
 		)
 
 		if len(hostShardSets) == 0 {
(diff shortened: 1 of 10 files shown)


src/cmd/services/m3dbnode/main/main_test.go
```diff
@@ -616,3 +634,42 @@ db:
...
+				}
+			}
+
+			if !allShardsAreAvailable {
+				break
+			}
+		}
+
+		if allShardsAreAvailable {
+			break
```
justinjc: Is `break`ing multiple levels like this clearer or via [label scopes](https://golang.org/ref/spec#Break_statements)?

src/dbnode/client/session.go
```diff
@@ -129,7 +129,7 @@ var (
 	errUnableToEncodeTags = errors.New("unable to include tags")
 	// errNoTopology is returned when the session does not have a topology. Should never happen
```
justinjc: nit: update comment to `errNoTopologyMap`

src/cmd/services/m3dbnode/main/main_test.go
```diff
@@ -616,3 +634,42 @@ db:
...
+		require.NoError(t, err)
+
+		var (
+			hostShardSets         = topoMap.HostShardSets()
+			allShardsAreAvailable = true
+		)
+
+		if len(hostShardSets) == 0 {
+			// We haven't received an actual topology yet.
+			continue
```
justinjc: Do we want to introduce a `time.Sleep(10 * time.Millisecond)` here as well to prevent a tight loop?

src/cmd/services/m3dbnode/main/main_test.go
```diff
@@ -616,3 +634,42 @@ db:
...
+		for _, hostShardSet := range hostShardSets {
+			for _, hostShard := range hostShardSet.ShardSet().All() {
+				if hostShard.State() != shard.Available {
+					allShardsAreAvailable = false
+					break
+				}
+			}
+
+			if !allShardsAreAvailable {
+				break
```
justinjc: Is `break`ing multiple levels like this clearer or via [label scopes](https://golang.org/ref/spec#Break_statements)?

src/cmd/services/m3dbnode/main/main_test.go
```diff
@@ -30,6 +30,8 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/m3db/m3cluster/shard"
+
```
nerd0: nit: why the extra line between the imports below it.

src/cmd/services/m3dbnode/main/main_test.go
```diff
@@ -616,3 +634,42 @@ db:
                   endpoint: {{.InitialClusterEndpoint}}
 `
 )
+
```
nerd0: This is an extremely aggressive poll. Why no delay between polls? Will this introduce unwanted amounts of extra load? Not sure how expensive the read of shard set? I assume that it would be the `etcd` store.

src/cmd/services/m3dbnode/main/main_test.go
nerd0: Will this be an issue during the turn-up phase of a cluster? Not sure.

src/cmd/services/m3dbnode/main/main_test.go
nerd0: A big __+1__ on this especially in this context.

src/dbnode/client/session.go
richardartoul: done

src/cmd/services/m3dbnode/main/main_test.go
richardartoul: good call, I'll just move the loop to the top of the for loop to avoid this in the future

src/cmd/services/m3dbnode/main/main_test.go
richardartoul: forgot about those, will do

src/cmd/services/m3dbnode/main/main_test.go
richardartoul: unintentional, fixed

src/cmd/services/m3dbnode/main/main_test.go
richardartoul: I don't think its aggressive. The context here is that it only happens when these integration tests are run, and they're all marked "big" which means that they'll always be run in sequence (never in parallel). So the poll is just aggressive to minimize humans waiting around for C.I to finish.

src/cmd/services/m3dbnode/main/main_test.go
richardartoul: This is test code only, no effect on production workloads.

richardartoul: Or did you mean do we need to worry about this during turn up? If thats your question, not really. All this means is that there will be a (probably less than a second) window between when all the nodes have finished bootstrapping, and any connected clients are made aware that the nodes are queryable. Only really matters in this test case where everything happens so quickly

nerd0: sorry my question was silly and the packages are separated by `_test`. Sorry for the noise

PR #736, opened 2018-05-30
Move test series iterator generation outside of test

arnikola:
Move test series iterator generation outside of test

### DIFF
--- a/src/coordinator/ts/m3db/generated/test_series_iterator.go
+++ b/src/coordinator/ts/m3db/generated/test_series_iterator.go
@@ -1,0 +1,176 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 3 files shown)


benraskin92:
I feel like putting it in a folder called `generated` is a bit weird

src/coordinator/tsdb/remote/compressed_codecs_test.go
```diff
@@ -189,7 +89,7 @@ func validateSeries(t *testing.T, it encoding.SeriesIterator) {
 	defer it.Close()
 
 	expectedValues := [60]float64{}
-	for i := 0; i < 30; i++ {
+	for i := 2; i < 30; i++ {
```
benraskin92: Is this because of the timing thing with the block start vs series iterator start?

src/coordinator/ts/m3db/generated/test_series_iterator.go
```diff
@@ -0,0 +1,176 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package generated
```
nikunjgit: generated is for autogenerated things. We have a test package in coordinator, consider moving it there ? Also, take out assert from here and instead return error ?

src/coordinator/tsdb/remote/compressed_codecs_test.go
arnikola: Yeah, the first two entries in the block are outside the iterator range

Issue #727, opened 2018-05-29
Prometheus integration uses hard coded value for M3DB namespace


nerd0:
Currently the Prometheus integration defaults to a `metrics` namespace for the coordinator to emit data to. This is not configurable thus requiring the user to strictly stick a generic namespace. To allow a bit more flexibility using configuration.  

```
coordinator:
  listenAddress: 0.0.0.0:7201

  metrics:
    prometheus:
      handlerPath: /metrics
    sanitization: prometheus
    samplingRate: 1.0
    extended: detailed

  dbNamespace: foo
```

nerd0:
@nikunjgit would mind commenting on this feature addition to the `coordinator`?

nikunjgit:
We should definitely have this. However, maybe figure out where to put this field in the config ? Eg: do you want it within coordinator: section or db: section ?

nerd0:
[Commit](https://github.com/m3db/m3db/commit/dd574a0ebb3aa581c1617f3a68631be30e9c2ed9) addresses the issue.

PR #738, opened 2018-05-31
Fixes issue #727

nerd0:
- Adding configuration option for user defined namespace for coordinator
- Added a bootstrap script for prometheus docker stack


examples/prometheus_integration/bootstrap_db.sh
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env bash
```
nerd0: @prateek would wiring this test within semaphore be overkill. Essentially it's testing Prometheus with M3DB as a remote storage node.

src/coordinator/services/m3coordinator/server/server.go
```diff
@@ -174,6 +170,10 @@ func Run(runOpts RunOptions) {
 
 func setupStorages(logger *zap.Logger, session client.Session, cfg config.Configuration) (storage.Storage, func()) {
 	cleanup := func() {}
+	namespace := "metrics"
```
prateek: nit: use a const for the default value

examples/prometheus_integration/bootstrap_db.sh
prateek: Only concern with adding is impact on build times. Can you see how much time build takes if you add to the existing "Docker Integration Test" job?

src/coordinator/services/m3coordinator/server/server.go
nerd0: good call, will update

examples/prometheus_integration/bootstrap_db.sh
nerd0: Currently the "Docker Integration Test" takes around 4 and half minutes to fully run. The longest test we have is around 10 minutes.

nerd0: @prateek sorry for the delay, with both the docker integration test and the prometheus test the [build](https://semaphoreci.com/m3db/m3db/branches/nerd0-issue-727/builds/8) clocks in around 7 minutes which is still under the 10 minutes builds. What do you think?

prateek: Nice! I'm good with adding it if you are :)

PR #741, opened 2018-05-31
Limit query IDs concurrency at database level to prevent overwhelming nodes with expensive queries

richardartoul:


src/dbnode/storage/database.go
```diff
@@ -529,7 +529,16 @@ func (d *db) QueryIDs(
 		return index.QueryResults{}, err
 	}
 
-	return n.QueryIDs(ctx, query, opts)
+	var (
+		doneCh       = make(chan struct{})
```
prateek: hm would using a waitGroup be a little cheaper here?

richardartoul: oh yeah prolly, I'll fix

PR #743, opened 2018-06-01
Move handlers and httpd from cmd/ to coordinator/

justinjc:
Move `handlers` and `httpd` from `src/cmd/services/m3coordinator` to `src/coordinator`.

I've done a find/sed to replace all path references.


justinjc:
Fair point. How do you want to handle versioning? Not sure what is convention here - does a `coordinator/api/v1/handler` make sense?

justinjc:
I've moved them into `/api/v1` for now - seems like some people do it this way and makes sense to me. Tell me if you prefer something else.

nikunjgit:
Oh yeah, that would have been my recommendation as well :)

PR #744, opened 2018-06-04
[WIP] Add rudimentary downsampling for m3coordinator

robskillington:


robskillington:
@xichen2020 @nikunjgit no feedback before I start productionizing this?

src/cmd/services/m3coordinator/downsample/downsample_id.go
```diff
@@ -0,0 +1,330 @@
...
+		if bytes.Equal(name, rollupTagName) && bytes.Equal(value, rollupTagValue) {
+			return true
+		}
+	}
+	iter.Close()
+
+	return false
+}
+
+type rollupIDProvider struct {
```
xichen2020: Can you add some comments explaining what this type is responsible for? It's hard to tell by just reading the code so far.

src/cmd/services/m3coordinator/downsample/downsample_id.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func (p *rollupIDProvider) finalize() {
+	if p.pool != nil {
+		p.pool.Put(p)
+	}
+}
+
+func (p *rollupIDProvider) Next() bool {
+	p.index++
```
xichen2020: This is confusing to me. Can you instead having `length` simply return `len(p.tagPairs)` and here do a length check first before you increment the index?

src/cmd/services/m3coordinator/downsample/downsample_id.go
```diff
@@ -0,0 +1,330 @@
...
+}
+
+func (p *rollupIDProvider) reset(
+	tagPairs []id.TagPair,
+) {
+	p.index = -1
+	p.tagPairs = tagPairs
+	p.rollupTagIndex = -1
+	for idx, pair := range tagPairs {
+		if bytes.Compare(rollupTagName, pair.Name) < 0 {
```
xichen2020: What if there is no rollup tag in the list of tag pairs provided?

src/cmd/services/m3coordinator/downsample/downsample_id.go
```diff
@@ -0,0 +1,330 @@
...
+
+func (p *rollupIDProvider) Close() {
+	// No-op
+}
+
+func (p *rollupIDProvider) Remaining() int {
+	return p.length() - p.index - 1
+}
+
+func (p *rollupIDProvider) Duplicate() ident.TagIterator {
```
xichen2020: This only does a shallow copy (e.g., the tag encoder is shared between itself and the clone). Is that ok?

src/cmd/services/m3coordinator/downsample/downsample_id.go
```diff
@@ -0,0 +1,330 @@
...
+func (p *rollupIDProviderPool) Put(v *rollupIDProvider) {
+	p.pool.Put(v)
+}
+
+func encodedTagsNameAndTags(
+	id []byte,
+	iterPool *encodedTagsIteratorPool,
+) ([]byte, []byte, error) {
+	// ID is always the encoded tags for downsampling IDs
+	metricTags := id
```
xichen2020: The id is passed in and returned without getting used in this method, can we just remove it?

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,756 @@
...
+
+	sortedTagIteratorFn := func(tagPairs []byte) id.SortedTagIterator {
+		it := sortedTagIteratorPool.Get()
+		it.Reset(tagPairs)
+		return it
+	}
+
+	tagsFilterOptions := filters.TagsFilterOptions{
+		NameTagKey: metricNameTagName,
+		NameAndTagsFn: func(id []byte) ([]byte, []byte, error) {
```
xichen2020: Does this mean for prom metrics name is a separate tag just like any other normal tags?

src/cmd/services/m3coordinator/downsample/downsample_id.go
xichen2020: Reading the code after this and coming back, it seems you are pretending that there is an extra rollup tag pair here which is to presumably satisfy the encoder interface(?), is that what's happening here?

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,756 @@
...
+	aggregatorOpts := aggregator.NewOptions().
+		SetClockOptions(clockOpts).
+		SetInstrumentOptions(instrumentOpts).
+		SetAggregationTypesOptions(aggTypeOpts).
+		SetMetricPrefix(nil).
+		SetCounterPrefix(nil).
+		SetGaugePrefix(nil).
+		SetTimerPrefix(nil)
+
+	shardSet := make([]shard.Shard, numShards)
```
xichen2020: Is the placement always statically configured?

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,756 @@
...
+
+	// Set up election manager.
+	leaderValue := instanceID
+	campaignOpts, err := services.NewCampaignOptions()
+	if err != nil {
+		return newAggregatorResult{}, err
+	}
+
+	campaignOpts = campaignOpts.SetLeaderValue(leaderValue)
+	electionCluster := integration.NewClusterV3(nil, &integration.ClusterConfig{
```
xichen2020: :\ so the coordinator also starts an embedded single-node etcd cluster? Why not pass in the etcd cluster externally so we can use the embedded KV cluster in M3DB nodes if we choose to?

src/cmd/services/m3coordinator/downsample/policy.go
```diff
@@ -0,0 +1,39 @@
...
+// THE SOFTWARE.
+
+package downsample
+
+// DownsamplingConfiguration is configuration of what to downsample.
+type DownsamplingConfiguration struct {
+	// NonePolicy is the none downsampling policy.
+	NonePolicy NonePolicy `yaml:"nonePolicy"`
+	// AggregationPolicy is the aggregation downsampling policy.
+	AggregationPolicy AggregationPolicy `yaml:"aggregationPolicy"`
```
xichen2020: So does this currently mean that if this is provided the downsampling is effectively achieved by aggregating datapoints within a resolution window?

src/coordinator/storage/local/storage.go
```diff
@@ -197,7 +355,119 @@ func newWriteRequest(writeRequestCommon *writeRequestCommon, timestamp time.Time
...
+				idx:   idx,
+				attrs: r.dedupeFirstAttrs,
+			}
+		}
+	}
+
+	for _, s := range result.SeriesList {
+		id := s.Name()
+		existing, exists := r.dedupeMap[id]
+		if exists && existing.attrs.Resolution < attrs.Resolution {
```
xichen2020: Do you want to make this configurable as opposed to always picking the finest resolution?

xichen2020:
Did a pass with a focus on the aggregator related logic and some of the query logic, @nikunjgit can probably do a more thorough review on the query part.

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,756 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
xichen2020: I'm not sure if this belongs in m3aggregator. This is essentially using m3aggregator as a library and this is the code that configures the various config options and creates the aggregator, which IMO should live outside the aggregator library. Additionally, there's a lot of logic here that parses tags etc, which should live outside m3aggregator since the aggregator has no notion of tags and it would be strange to introduce it there.

src/cmd/services/m3coordinator/downsample/downsample_id.go
robskillington: Sure thing. (It just constructs a rollup ID when necessary, and can be pooled)

src/cmd/services/m3coordinator/downsample/downsample_id.go
robskillington: Yes, that's what's happening.  I'll leave a comment to this effect.  I wanted to avoid another alloc of TagPairs.

src/cmd/services/m3coordinator/downsample/downsample_id.go
robskillington: (As per other comment, there is meant to be always the rollup tag injected so there will never be one in the tag pairs provided as we inject it ourselves)

src/cmd/services/m3coordinator/downsample/downsample_id.go
robskillington: I only made this to fulfill the interface, in practice it won't be duplicated.  I can make a deep copy though just to be careful.

src/cmd/services/m3coordinator/downsample/downsample_id.go
robskillington: it's part of the method required, but I can move that to the lambda calling it perhaps.

src/cmd/services/m3coordinator/downsample/downsampler.go
robskillington: Yes indeed.

src/cmd/services/m3coordinator/downsample/downsampler.go
robskillington: Yes, this downsampler always just runs local to the coordinator.

src/cmd/services/m3coordinator/downsample/downsampler.go
robskillington: So we don't have an external etcd cluster.  What actually needs to happen here is we need to remove the dependency of a real etcd cluster from the leader service, it should just take a struct that can fulfill some things and abstract the underlying implementation (which happens to be etcd).

src/cmd/services/m3coordinator/downsample/policy.go
robskillington: Correct, although you can have multiple policies enabled.

src/coordinator/storage/local/storage.go
robskillington: We should yes, perhaps I'll leave a followup though as we probably don't need to provide that for v1.

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,756 @@
...
+	}
+
+	var (
+		storageFlushConcurrency = defaultStorageFlushConcurrency
+		rulesStore              = o.RulesKVStore
+		clockOpts               = o.ClockOptions
+		instrumentOpts          = o.InstrumentOptions
+	)
+	if o.StorageFlushConcurrency > 0 {
+		storageFlushConcurrency = storageFlushConcurrency
```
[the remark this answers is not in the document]
robskillington: Cheers, thanks for the catch.

src/cmd/services/m3coordinator/downsample/downsample_id.go
```diff
@@ -0,0 +1,330 @@
...
+	errNoMetricNameTag = errors.New("no metric name tag found")
+)
+
+// Ensure encodedTagsIterator implements id.SortedTagIterator
+var _ id.SortedTagIterator = &encodedTagsIterator{}
+
+// Ensure encodedTagsIterator implements id.ID
+var _ id.ID = &encodedTagsIterator{}
+
+type encodedTagsIterator struct {
```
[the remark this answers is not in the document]
robskillington: It needs to live somewhere that cares about metric ID format.  Since m3aggregator and m3db are both agnostic to the metric IDs and how tags are encoded they aren't the best location for this code to leave.  m3coordinator however knows about the concrete formats, i.e. prometheus metric ID format.  This package is the only one that needs to know about how tags are encoded for internal use in the downsampler, it would be leaky to make it need to live elsewhere and be imported into the downsample package.

src/cmd/services/m3coordinator/downsample/downsampler.go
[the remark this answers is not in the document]
robskillington: This downsampler is essentially a single node aggregation package, currently it's not used outside of the coordinator and I doubt would really have a home anywhere else (also the only two options we should ever really offer is simple, non-HA downsampling in the m3coordinator, or proper downsampling with the m3aggregator in a clustered setup).  

I'm inclined to leave it here until it's used anywhere else (and hopefully never used anywhere else, since you should be using the m3aggregator otherwise).  Also it helps avoid needing to make it super generic and optimize it for it's use in the m3coordinator.

src/coordinator/storage/local/storage.go
```diff
@@ -66,13 +73,72 @@ func (s *localStorage) Fetch(ctx context.Context, query *storage.FetchQuery, opt
...
+	var (
+		now     = time.Now()
+		fetches = 0
+		result  multiFetchResult
+		wg      sync.WaitGroup
+	)
+	for _, namespace := range namespaces {
+		namespace := namespace // Capture var
+
+		clusterStart := now.Add(-1 * namespace.Attributes().Retention)
```
arnikola: nit: can use `now.Sub(namespace.Attributes().Retention)`?

src/coordinator/storage/local/storage.go
```diff
@@ -66,13 +73,72 @@ func (s *localStorage) Fetch(ctx context.Context, query *storage.FetchQuery, opt
...
+
+	if fetches == 0 {
+		return nil, errNoLocalClustersFulfillsQuery
+	}
+
+	wg.Wait()
+	if err := result.err.FinalError(); err != nil {
+		return nil, err
+	}
+	return result.result, nil
```
arnikola: nit: add a newline here

src/coordinator/storage/local/storage.go
```diff
@@ -22,32 +22,39 @@ package local
...
 }
 
 // NewStorage creates a new local Storage instance.
-func NewStorage(session client.Session, namespace string, workerPool pool.ObjectPool) storage.Storage {
-	return &localStorage{
-		session:    session,
-		namespace:  ident.StringID(namespace),
-		workerPool: workerPool,
-	}
+func NewStorage(clusters Clusters, workerPool pool.ObjectPool) storage.Storage {
```
arnikola: I think this is the same as in #785 right? Can you rebase this on master now that it's merged?

src/coordinator/storage/interface.go
```diff
@@ -129,3 +130,38 @@ type QueryResult struct {
 	FetchResult *FetchResult
 	Err         error
 }
+
+// MetricsType is a type of stored metrics.
```
arnikola: This should be in master now, right?

src/coordinator/services/m3coordinator/server/server.go
```diff
@@ -175,15 +175,25 @@ func Run(runOpts RunOptions) {
 		return workerPool
 	})
 
-	fanoutStorage, storageCleanup := setupStorages(logger, session, cfg, objectPool)
+	// TODO(r): clusters
```
arnikola: Should be good after rebasing on master, right?

src/coordinator/api/v1/handler/prometheus/remote/write.go
```diff
@@ -104,11 +118,67 @@ func (h *PromWriteHandler) parseRequest(r *http.Request) (*prompb.WriteRequest,
...
+				metricsAppender.Reset()
+				for _, label := range ts.Labels {
+					metricsAppender.AddTag(label.Name, label.Value)
+				}
+
+				samplesAppender, err := metricsAppender.SamplesAppender()
+				if err != nil {
+					multiErr = multiErr.Add(err)
+					continue
+				}
```
arnikola: nit: add a newline here?

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+
+func (d *downsampler) NonePolicy() NonePolicy {
+	return d.cfg.NonePolicy
+}
+
+func (d *downsampler) AggregationPolicy() AggregationPolicy {
+	return d.cfg.AggregationPolicy
+}
+
+func (d *downsampler) MetricsAppender() MetricsAppender {
```
arnikola: nit: I think this would make more sense as `NewMetricsAppender()`, would also make it clearer that you're getting a copy of the appender inside the downsampler goroutines

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+func (t *tags) Less(i, j int) bool {
+	return t.names[i] < t.names[j]
+}
+
+func (t *tags) Next() bool {
+	return t.idx+1 < len(t.names)
+}
+
+func (t *tags) Current() ident.Tag {
+	t.nameBuf = append(t.nameBuf[:0], t.names[t.idx]...)
```
arnikola: Just curious what this does

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+}
+
+func (a *multiSamplesAppender) addSamplesAppender(v samplesAppender) {
+	a.appenders = append(a.appenders, v)
+}
+
+func (a *multiSamplesAppender) AppendCounterSample(value int64) error {
+	var multiErr xerrors.MultiError
+	for _, appender := range a.appenders {
+		multiErr = multiErr.Add(appender.AppendCounterSample(value))
```
arnikola: Would it be better to shortcircuit on the first error received?

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+	}
+	return multiErr.FinalError()
+}
+
+func (a *multiSamplesAppender) AppendGaugeSample(value float64) error {
+	var multiErr xerrors.MultiError
+	for _, appender := range a.appenders {
+		multiErr = multiErr.Add(appender.AppendGaugeSample(value))
```
arnikola: Same here re: return after first error

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+	flushManagerOpts := aggregator.NewFlushManagerOptions().
+		SetPlacementManager(placementManager).
+		SetFlushTimesManager(flushTimesManager).
+		SetElectionManager(electionManager)
+	flushManager := aggregator.NewFlushManager(flushManagerOpts)
+	aggregatorOpts = aggregatorOpts.SetFlushManager(flushManager)
+
+	flushWorkers := xsync.NewWorkerPool(storageFlushConcurrency)
+	flushWorkers.Init()
+	handler := newDownsamplerFlushHandler(o.Storage, sortedTagIteratorPool,
```
arnikola: nit: can we pull all this setup into a separate function?

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+	scope tally.Scope,
+) (writer.Writer, error) {
+	return &downsamplerFlushHandlerWriter{
+		ctx:     context.Background(),
+		handler: h,
+	}, nil
+}
+
+func (h *downsamplerFlushHandler) Close() {
+}
```
arnikola: nit: can we get a comment here? Either `// noop` or `// TODO` depending on what's required

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+type downsamplerFlushHandlerWriter struct {
+	wg      sync.WaitGroup
+	ctx     context.Context
+	handler *downsamplerFlushHandler
+}
+
+func (w *downsamplerFlushHandlerWriter) Write(
+	mp aggregated.ChunkedMetricWithStoragePolicy,
+) error {
+	w.wg.Add(1)
```
arnikola: It seems a little odd to me to have this function touching the internal wait group; what's the difference between doing it this way and calling Flush at the end v.s. the calling function handling the parallelism?

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+}
+
+func (a *metricsAppender) AddTag(name, value string) {
+	a.tags.names = append(a.tags.names, name)
+	a.tags.values = append(a.tags.values, name)
+}
+
+func (a *metricsAppender) SamplesAppender() (SamplesAppender, error) {
+	// Sort tags
+	sort.Sort(a.tags)
```
arnikola: Would it be better to insert the tags in the sorted location?

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+			a.tags.names, a.tags.values)
+	}
+
+	unownedID := data.Bytes()
+
+	// Match policies and rollups and build samples appender
+	id := a.encodedTagsIteratorPool.Get()
+	id.Reset(unownedID)
+	now := time.Now()
+	fromNanos, toNanos := now.Add(-1*a.clockOpts.MaxNegativeSkew()).UnixNano(),
```
arnikola: nit: `now.Sub`, might be easier to read across two lines?

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -0,0 +1,757 @@
...
+	a.multiSamplesAppender.addSamplesAppender(samplesAppender{
+		agg:       a.agg,
+		unownedID: unownedID,
+		policies:  policies,
+	})
+
+	numRollups := matchResult.NumRollups()
+	for i := 0; i < numRollups; i++ {
+		rollup, ok := matchResult.RollupsAt(i, now.UnixNano())
+		if !ok {
```
arnikola: nit: Better to flip this and add samplesAppender `if ok`

robskillington:
Deprecating in favor of new downsampling PR https://github.com/m3db/m3db/pull/796, now that the multi-cluster support has landed.

PR #745, opened 2018-06-04
Serialization now uses compressed tags when possible

arnikola:


src/coordinator/storage/converter.go
```diff
@@ -21,12 +21,19 @@
 package storage
 
 import (
+	"context"
 	"fmt"
 	"time"
 
+	"github.com/m3db/m3coordinator/util/execution"
+
```
benraskin92: nit: newline

src/coordinator/storage/converter.go
```diff
@@ -177,3 +183,90 @@ func SeriesToPromSamples(series *ts.Series) []*prompb.Sample {
...
+	metric, err := FromM3IdentToMetric(ns, iter.ID(), iter.Tags())
+	if err != nil {
+		return err
+	}
+
+	datapoints := make(ts.Datapoints, 0, initRawFetchAllocSize)
+	for iter.Next() {
+		dp, _, _ := iter.Current()
+		datapoints = append(datapoints, ts.Datapoint{Timestamp: dp.Timestamp, Value: dp.Value})
```
benraskin92: Since you're pre-allocating above, is it better to use an index. e.g.

```
for i := 0; iter.Next(); i++ {
	dp, _, _ := iter.Current()
	datapoints[i] = ts.Datapoint{Timestamp: dp.Timestamp, Value: dp.Value}
}
```

src/coordinator/storage/converter.go
```diff
@@ -177,3 +183,90 @@ func SeriesToPromSamples(series *ts.Series) []*prompb.Sample {
...
+	workerPool.Init()
+}
+
+type decompressRequest struct {
+	iter      encoding.SeriesIterator
+	namespace ident.ID
+	result    *ts.Series
+}
+
+func (w *decompressRequest) Process(ctx context.Context) error {
```
benraskin92: nit: add a comment here

arnikola: Surprised the linter let me get away with that

src/coordinator/storage/converter.go
arnikola: I need it to be able to expand if initRawFetchAllocSize is too small

src/coordinator/storage/converter.go
```diff
@@ -177,3 +181,82 @@ func SeriesToPromSamples(series *ts.Series) []*prompb.Sample {
...
+
+// SeriesIteratorsToFetchResult converts SeriesIterators into a fetch result
+func SeriesIteratorsToFetchResult(ctx context.Context, seriesIterators encoding.SeriesIterators, namespace ident.ID) (*FetchResult, error) {
+	defer seriesIterators.Close()
+
+	iters := seriesIterators.Iters()
+	iterLength := seriesIterators.Len()
+
+	seriesList := make([]*ts.Series, 0, iterLength)
+	div, remainder := iterLength/workerPoolSize, iterLength%workerPoolSize
```
nikunjgit: this logic should probably not live here. Maybe consider moving this within execution.ExecuteParallel ?

nikunjgit: maybe push the responsibility of providing the workerpoolsize to the caller, that way we can eventually centrally control all these resources

src/coordinator/storage/converter.go
```diff
@@ -177,3 +181,82 @@ func SeriesToPromSamples(series *ts.Series) []*prompb.Sample {
...
+	var executionPools [][]execution.Request
+	if remainder > 0 {
```
nikunjgit: i'm hoping we can leverage a library for this instead of doing the math ourselves

src/coordinator/tsdb/remote/codecs.go
```diff
@@ -40,6 +40,17 @@ func toTime(t int64) time.Time {
 	return storage.TimestampToTime(t)
 }
 
+func encodeTags(tagMap map[string]string) []*rpc.Tag {
```
nikunjgit: why not just have []rpc.Tag ?

src/coordinator/tsdb/remote/codecs.go
```diff
@@ -80,6 +91,15 @@ func DecodeFetchResult(_ context.Context, rpcSeries []*rpc.Series) ([]*ts.Series
 	return tsSeries, nil
 }
 
+func decodeTags(tags []*rpc.Tag) models.Tags {
```
nikunjgit: why not just have []rpc.Tag ?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -93,11 +94,28 @@ func compressedSegmentsFromReaders(readers xio.ReaderSliceOfSlicesIterator) (*rp
 	return segments, nil
 }
 
-func compressedTagsFromTagIterator(tagIter ident.TagIterator) ([]*rpc.CompressedTag, error) {
-	tags := make([]*rpc.CompressedTag, 0, tagIter.Remaining())
+func compressedTagsFromTagIteratorWithEncoder(tagIter ident.TagIterator, iterPools encoding.IteratorPools) ([]byte, bool) {
+	if iterPools == nil {
+		return nil, false
+	}
+	encoder := iterPools.TagEncoder().Get()
```
nikunjgit: nil check ?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -93,11 +94,28 @@ func compressedSegmentsFromReaders(readers xio.ReaderSliceOfSlicesIterator) (*rp
...
+	defer encoder.Finalize()
+	data, encoded := encoder.Data()
+	if !encoded {
+		return nil, false
+	}
+	return data.Bytes(), true
+}
+
+func tagsFromTagIterator(tagIter ident.TagIterator) ([]*rpc.Tag, error) {
+	tags := make([]*rpc.Tag, 0, tagIter.Remaining())
```
nikunjgit: avoid pointer slices

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -142,25 +158,52 @@ func CompressedSeriesFromSeriesIterator(it encoding.SeriesIterator) (*rpc.Series
...
-	}
-
 	start := xtime.ToNanoseconds(it.Start())
 	end := xtime.ToNanoseconds(it.End())
+	tagIter := it.Tags()
+	compressedTags, canCompress := compressedTagsFromTagIteratorWithEncoder(tagIter, iterPools)
+
+	var tags []*rpc.Tag
+	var err error
+	if !canCompress {
```
nikunjgit: maybe not rely on an error to switch logic ? Ideally, if we have compressedDatapoints we should always have compressedTags ? Otherwise, its confusing.

src/coordinator/tsdb/remote/codecs.go
arnikola: Required to conform to what protobuf expects

src/coordinator/tsdb/remote/codecs.go
arnikola: Required to conform to what protobuf expects

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: Required to conform to what protobuf expects

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: That implies a failing in the iterPools, but I'll add a sanity check

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: This is going the other way; if we *can* make compressedTags, we do, otherwise we use regular tags

src/coordinator/tsdb/remote/compressed_codecs.go
benraskin92: Would we ever need to use this func elsewhere? Might be worth it to export it?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -142,25 +158,52 @@ func CompressedSeriesFromSeriesIterator(it encoding.SeriesIterator) (*rpc.Series
...
-		return nil, err
-	}
-
 	start := xtime.ToNanoseconds(it.Start())
 	end := xtime.ToNanoseconds(it.End())
+	tagIter := it.Tags()
+	compressedTags, canCompress := compressedTagsFromTagIteratorWithEncoder(tagIter, iterPools)
+
+	var tags []*rpc.Tag
+	var err error
```
benraskin92:
```
var (
  tags []*rpc.Tag
  err  error
)
```

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -142,25 +158,52 @@ func CompressedSeriesFromSeriesIterator(it encoding.SeriesIterator) (*rpc.Series
...
+		Tags:       tags,
+	}, nil
+}
+
+// EncodeToCompressedFetchResult encodes SeriesIterators to compressed fetch results
+func EncodeToCompressedFetchResult(
+	iterators encoding.SeriesIterators,
+	iterPools encoding.IteratorPools,
+) (*rpc.FetchResult, error) {
+	iters := iterators.Iters()
```
benraskin92: `defer iterators.Close()`?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -226,6 +283,20 @@ func tagIteratorFromCompressedTags(compressedTags []*rpc.CompressedTag, idPool i
 	return tagIter
 }
 
+func tagIteratorFromSeries(series *rpc.Series, iteratorPools encoding.IteratorPools) (ident.TagIterator, error) {
+	var (
+		idPool ident.Pool
+	)
```
benraskin92: newline

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -310,8 +387,43 @@ func SeriesIteratorFromCompressedSeries(
...
+// DecodeCompressedFetchResult decodes compressed fetch results to seriesIterators
+func DecodeCompressedFetchResult(
+	fetchResult *rpc.FetchResult,
+	iteratorPools encoding.IteratorPools,
+) (encoding.SeriesIterators, error) {
+	rpcSeries := fetchResult.GetSeries()
+	var (
+		pooledIterators encoding.MutableSeriesIterators
+		seriesIterators []encoding.SeriesIterator
+		numSeries       = len(rpcSeries)
```
benraskin92: newline

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -310,8 +387,43 @@ func SeriesIteratorFromCompressedSeries(
...
+	return encoding.NewSeriesIterator(id, ns, tagIter, start, end, allReplicaIterators, seriesPool), nil
+}
+
+// DecodeCompressedFetchResult decodes compressed fetch results to seriesIterators
+func DecodeCompressedFetchResult(
+	fetchResult *rpc.FetchResult,
+	iteratorPools encoding.IteratorPools,
+) (encoding.SeriesIterators, error) {
+	rpcSeries := fetchResult.GetSeries()
+	var (
```
benraskin92: Put these at the top of the function?

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: That's going to be the responsibility of the caller, since tags id and namespace bytes are shared between the iterators and *rpc.FetchResult.

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: Unlikely, as this casts directly to the rpc.Tag type which should only be used for cross DC stuff

src/coordinator/storage/converter_test.go
```diff
@@ -0,0 +1,67 @@
...
+import (
+	"context"
+	"testing"
+
+	"github.com/m3db/m3db/src/coordinator/models"
+	"github.com/m3db/m3db/src/coordinator/test/seriesiter"
+	"github.com/m3db/m3db/src/coordinator/util/logging"
+
+	"github.com/golang/mock/gomock"
+	"github.com/m3db/m3x/ident"
```
benraskin92: nit: importorder

src/coordinator/storage/converter_test.go
```diff
@@ -0,0 +1,67 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func verifyExpandSeries(ctx context.Context, t *testing.T, ctrl *gomock.Controller, num int) {
+	testTags := seriesiter.GenerateTag()
+	iters := seriesiter.NewMockSeriesIters(ctrl, testTags, num)
+
+	results, err := SeriesIteratorsToFetchResult(ctx, iters, ident.StringID("strID"))
+	assert.NoError(t, err)
+
```
benraskin92: remove newline

src/coordinator/storage/fanout/storage_test.go
```diff
@@ -29,14 +29,14 @@ import (
...
-	"github.com/m3db/m3db/src/coordinator/test"
 	"github.com/m3db/m3db/src/coordinator/test/local"
+	"github.com/m3db/m3db/src/coordinator/test/seriesiter"
 	"github.com/m3db/m3db/src/coordinator/ts"
 	"github.com/m3db/m3db/src/coordinator/util/logging"
 	"github.com/m3db/m3db/src/dbnode/encoding"
-	"github.com/m3db/m3x/ident"
 
 	"github.com/golang/mock/gomock"
+	"github.com/m3db/m3x/ident"
```
benraskin92: importorder

src/coordinator/storage/converter.go
```diff
@@ -177,3 +181,116 @@ func SeriesToPromSamples(series *ts.Series) []*prompb.Sample {
...
+
+	if workerPools == nil {
+		return decompressSequentially(iterLength, iters, namespace)
+	}
+	pool, ok := workerPools.Get().(xsync.WorkerPool)
+	if !ok {
+		return decompressSequentially(iterLength, iters, namespace)
+	}
+
+	var wg sync.WaitGroup
```
nikunjgit: can you extract this out into separate function and maybe call it decompressConcurrently ?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -142,25 +162,55 @@ func CompressedSeriesFromSeriesIterator(it encoding.SeriesIterator) (*rpc.Series
...
 	end := xtime.ToNanoseconds(it.End())
+	tagIter := it.Tags()
+	compressedTags, canCompress := compressedTagsFromTagIteratorWithEncoder(tagIter, iterPools)
+
+	var (
+		tags []*rpc.Tag
+		err  error
+	)
+
+	if !canCompress {
```
nikunjgit: Its strange to make it optional. Clients might be simpler if compressed datapoints always correspond to compressedTags as well. Clients can then just always go one path instead of switching based on error

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -196,7 +246,21 @@ func blockReaderFromCompressedSegment(
...
+	iterPools encoding.IteratorPools,
+) (ident.TagIterator, error) {
+	if iterPools == nil || iterPools.CheckedBytesWrapper() == nil || iterPools.TagDecoder() == nil {
+		return nil, errors.ErrCannotDecodeCompressedTags
+	}
+	checkedBytes := iterPools.CheckedBytesWrapper().Get(compressedTags)
+	decoder := iterPools.TagDecoder().Get()
+	decoder.Reset(checkedBytes)
+	defer decoder.Close()
+	return decoder.Duplicate(), nil
```
nikunjgit: can you comment about why duplicate ?

arnikola:
> So eventually, we want both local and remote storages to return blocks. Is that going to be a separate diff ? Once we start only dealing with blocks, how much of this is going to be used.

The meat of this will lead to returning encoding.SeriesIterators; these will be passed through the conversion function in https://github.com/m3db/m3db/pull/726 to return the blocks we need

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -226,6 +289,18 @@ func tagIteratorFromCompressedTags(compressedTags []*rpc.CompressedTag, idPool i
 	return tagIter
 }
 
+func tagIteratorFromSeries(series *rpc.Series, iteratorPools encoding.IteratorPools) (ident.TagIterator, error) {
+	compressedValues := series.GetCompressed()
+	if len(compressedValues.GetCompressedTags()) > 0 {
```
nikunjgit: nil check for compressedValues ? Might be better to check for isSet ?

src/coordinator/tsdb/remote/compressed_codecs.go
```diff
@@ -252,16 +327,22 @@ func blockReadersFromCompressedSegments(
...
-func SeriesIteratorFromCompressedSeries(
+func seriesIteratorFromCompressedSeries(
 	timeSeries *rpc.Series,
 	iteratorPools encoding.IteratorPools,
-) encoding.SeriesIterator {
+) (encoding.SeriesIterator, error) {
 	initialize.Do(initializeVars)
 
+	// Attempt to decompress compressed tags first as this is the only scenario that is expected to fail
+	tagIter, err := tagIteratorFromSeries(timeSeries, iteratorPools)
```
nikunjgit: tagIteratorFromSeries can either decompress tags or create them using series.GetTags(). Should we only allow the first method ? This will enforce that if the series is compressed, then its tags will also always be compressed.

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: Good call on nil check, for some reason didn't think .GetCompressed() could return nil

src/coordinator/tsdb/remote/compressed_codecs.go
arnikola: Not necessarily, since series can be compressed but not have compressed tags if no iterator pools are passed in.

PR #746, opened 2018-06-04
Convert a decompressed series into a block for coordinator

nikunjgit:


src/coordinator/block/interface.go
```diff
@@ -0,0 +1,101 @@
...
+	// ErrBounds is returned when time requested is outside the block bounds
+	ErrBounds = "out of bounds, time: %v, bounds: %v"
+)
+
+// Block represents a group of series across a time bound
+type Block interface {
+	Meta() Metadata
+	StepIter() StepIter
+	SeriesIter() SeriesIter
+	SeriesMeta() []SeriesMeta
```
arnikola: Will need a Close()/Finalize() method

src/coordinator/block/column.go
```diff
@@ -0,0 +1,137 @@
+package block
```
arnikola: nit: `Copyright (c) 2018 Uber Technologies, Inc.` etc

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,101 @@
+package block
```
arnikola: nit `Copyright (c) 2018 Uber Technologies, Inc.` etc

src/coordinator/executor/transform/interface.go
```diff
@@ -3,8 +3,8 @@ package transform
 import (
```
arnikola: nit `Copyright (c) 2018 Uber Technologies, Inc.` etc

src/coordinator/storage/block.go
```diff
@@ -3,56 +3,131 @@ package storage
 import (
 	"time"
 
-	"github.com/m3db/m3db/src/coordinator/models"
+	"github.com/m3db/m3db/src/coordinator/block"
```
arnikola: nit `Copyright (c) 2018 Uber Technologies, Inc.` etc

src/coordinator/test/storage.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
```
arnikola: nit `Copyright (c) 2018 Uber Technologies, Inc.` etc

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
...
+	Meta() Metadata
+	StepIter() StepIter
+	SeriesIter() SeriesIter
+	SeriesMeta() []SeriesMeta
+	Close() error
+}
+
+// SeriesMeta is metadata data for the series
+type SeriesMeta struct {
+	Tags models.Tags
```
benraskin92: So will this just contain the tags unique to the series and then at the `Block` level we have the common tags? Or will this contain all of the tags?

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// Bounds are the time bounds
+// nolint: structcheck, megacheck, unused
```
benraskin92: temporary?

src/coordinator/functions/count.go
```diff
@@ -57,13 +57,18 @@ type CountNode struct {
...
 // Process the block
-func (c *CountNode) Process(ID parser.NodeID, block storage.Block) error {
+func (c *CountNode) Process(ID parser.NodeID, block block.Block) error {
 	builder, err := c.controller.BlockBuilder(block.Meta())
 	if err != nil {
 		return err
 	}
 
 	stepIter := block.StepIter()
+	err = builder.AddCols(stepIter.Len())
```
benraskin92:
```
if err := builder.AddCols(stepIter.Len()); err != nil {
  return err 
}
```

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
-	Meta() BlockMetadata
-	StepIter() StepIter
-	SeriesIter() SeriesIter
-	SeriesMeta() []SeriesMeta
-	StepMeta() []StepMeta
+// FetchResultToBlockResult converts a fetch result into coordinator blocks
+func FetchResultToBlockResult(result *FetchResult, query *FetchQuery) (block.Result, error) {
+	alignedSeriesList, err := result.SeriesList.Align(query.Start, query.End, query.Interval)
+	if err != nil {
+		return block.Result{}, err
```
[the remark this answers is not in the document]
nikunjgit: yeah, can't return nil, so just doing an empty struct.

src/coordinator/ts/series.go
```diff
@@ -49,3 +54,64 @@ func (b *Series) Len() int { return b.vals.Len() }
...
+		return nil, fmt.Errorf("unknown type: %v", vals)
+	}
+}
+
+// SeriesList represents a slice of series pointers
+type SeriesList []*Series
+
+// Resolution returns the resolution for a fixed step series list. It assumes all underlying series will have the same resolution
+func (seriesList SeriesList) Resolution() (time.Duration, error) {
+	var resolution time.Duration
```
arnikola: Is it fine to return `resolution = 0` if seriesList has no length?

src/coordinator/ts/series.go
```diff
@@ -49,3 +54,64 @@ func (b *Series) Len() int { return b.vals.Len() }
 
 // Values returns the underlying values interface
 func (b *Series) Values() Values { return b.vals }
+
+// Align adjusts the datapoints to start, end and a fixed interval
+func (b *Series) Align(start time.Time, end time.Time, interval time.Duration) (*Series, error) {
```
arnikola: nit: `start, end time.Time`

src/coordinator/ts/series.go
```diff
@@ -49,3 +54,64 @@ func (b *Series) Len() int { return b.vals.Len() }
...
+	fixedVals, err := alignValues(b.Values(), start, end, interval)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewSeries(b.name, fixedVals, b.Tags), nil
+}
+
+func alignValues(values Values, start time.Time, end time.Time, interval time.Duration) (FixedResolutionMutableValues, error) {
```
arnikola: nit: `start, end time.Time`

src/coordinator/test/storage.go
arnikola: Still need this one

src/coordinator/storage/local/storage.go
```diff
@@ -164,8 +165,18 @@ func (s *localStorage) Type() storage.Type {
 }
 
 func (s *localStorage) FetchBlocks(
-	ctx context.Context, query *storage.FetchQuery, options *storage.FetchOptions) (storage.BlockResult, error) {
-	return storage.BlockResult{}, errors.ErrNotImplemented
+	ctx context.Context, query *storage.FetchQuery, options *storage.FetchOptions) (block.Result, error) {
```
arnikola: nit: would look better if you split args by line?

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
+	block multiSeriesBlock
+	index int
 }
 
-// Step can optionally implement iterator interface
-type Step interface {
-	Time() time.Time
-	Values() []float64
+func (m *multiSeriesBlockSeriesIter) Next() bool {
+	return m.index < len(m.block.seriesList)
```
arnikola: Shouldn't Next() increment the index?

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
 }
 
-// BlockMetadata is metadata for a block
-type BlockMetadata struct {
-	Bounds Bounds
-	Tags   models.Tags // Common tags across different series
+func (m *multiSeriesBlockSeriesIter) Current() block.Series {
+	s := m.block.seriesList[m.index]
```
arnikola: Might be worth doing a bounds check; probably not

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
+	values := make([]float64, s.Len())
+	for i := 0; i < s.Len(); i++ {
+		values[i] = s.Values().ValueAt(i)
```
arnikola: Any way we can memcopy this rather than copying one by one?

src/coordinator/block/column.go
```diff
@@ -0,0 +1,163 @@
...
+	step := bounds.StepSize
+	t := bounds.Start.Add(time.Duration(index) * step)
+	if t.After(bounds.End) {
+		return time.Time{}, fmt.Errorf("out of bounds, %d", index)
+	}
+
+	return t, nil
+}
+
+type colStep struct {
```
arnikola: A bit weird to have two different `colStep` structs

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
+	}
+
+	m.index++
+	return colStep{
+		t:      t,
+		values: values,
+	}
+}
+
+func (m *multiSeriesBlockStepIter) Len() int {
```
arnikola: Should be `CurrentLen()`; `Len()` (or `Remaining()`) should give the length  of the iterator itself

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
+	if len(m.block.seriesList) == 0 {
+		return 0
+	}
+
+	return m.block.seriesList[0].Values().Len()
```
arnikola: Shouldn't this be m.block.seriesList[m.index].Values().Len() ?

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
+}
+
+func (m *multiSeriesBlockStepIter) Current() block.Step {
+	t := m.block.meta.Bounds.Start.Add(time.Duration(m.index) * m.block.meta.Bounds.StepSize)
+	values := make([]float64, len(m.block.seriesList))
+	for i, s := range m.block.seriesList {
+		values[i] = s.Values().ValueAt(i)
+	}
+
+	m.index++
```
arnikola: index should be incremented on Next() not on Current()

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
+func (m *multiSeriesBlockStepIter) Next() bool {
+	if len(m.block.seriesList) == 0 {
+		return false
+	}
+
+	return m.index < m.block.seriesList[0].Values().Len()
+}
+
+func (m *multiSeriesBlockStepIter) Current() block.Step {
+	t := m.block.meta.Bounds.Start.Add(time.Duration(m.index) * m.block.meta.Bounds.StepSize)
```
arnikola: nit: block := m.block

arnikola: nit: define variable closer to where you're using it

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
+	values := make([]float64, len(m.block.seriesList))
+	for i, s := range m.block.seriesList {
+		values[i] = s.Values().ValueAt(i)
```
arnikola: Can use memcopy?

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,157 @@
...
+func (m multiSeriesBlock) Close() error {
+	return nil
+}
+
+type multiSeriesBlockStepIter struct {
+	block multiSeriesBlock
+	index int
+}
+
+func (m *multiSeriesBlockStepIter) Next() bool {
```
arnikola: This is a really weird way of moving through the iterator, I'm not super sure how you'd use these iterators... any chance we could get an example test that would illustrate how these are expected to be used?

src/coordinator/storage/block.go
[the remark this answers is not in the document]
arnikola: Would a pointer be preferable? block.Result can potentially be very large

src/coordinator/block/column.go
```diff
@@ -0,0 +1,163 @@
...
+// THE SOFTWARE.
+
+package block
+
+import (
+	"fmt"
+	"time"
+)
+
+// ColumnBlockBuilder builds a block optimized for column iteration
```
arnikola: Does this build a block or a column iterator?

src/coordinator/block/column.go
```diff
@@ -0,0 +1,163 @@
...
+		panic("current called without next")
+	}
+
+	col := c.columns[c.index]
+	t, err := calcTime(c.meta.Bounds, c.index)
+	if err != nil {
+		panic(err)
+	}
+
+	c.index++
```
arnikola: index should be updated on Next() not Current()

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+// ValueAtStep returns the datapoint value at a step index
+func (s Series) ValueAtStep(idx int) float64 {
+	return s.values[idx]
+}
+
+// ValueAtTime returns the datapoint value at a given time
+func (s Series) ValueAtTime(t time.Time) (float64, error) {
+	if t.Before(s.bounds.Start) || t.After(s.bounds.End) {
```
arnikola: nit: bounds := s.bounds

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,121 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/m3db/m3db/src/coordinator/models"
+)
+
+const (
+	// ErrBounds is returned when time requested is outside the block bounds
+	ErrBounds = "out of bounds, time: %v, bounds: %v"
```
arnikola: nit: don't need to export

src/coordinator/executor/transform/controller.go
```diff
@@ -51,12 +50,6 @@ func (t *Controller) Process(block storage.Block) error {
...
 // nolint: unparam
-func (t *Controller) BlockBuilder(blockMeta storage.BlockMetadata) (BlockBuilder, error) {
-	return nil, errors.ErrNotImplemented
-}
-
-// BlockBuilder builds a new block
-type BlockBuilder interface {
-	AppendValue(index int, value float64)
-	Build() storage.Block
+func (t *Controller) BlockBuilder(blockMeta block.Metadata) (block.Builder, error) {
```
arnikola: A bit odd that this is called on a `Controller` if it doesn't do anything with it

src/coordinator/storage/block.go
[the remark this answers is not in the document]
nikunjgit: block.Result has a slice inside it which itself is a pointer so having another pointer here will not be efficient.

src/coordinator/ts/series.go
nikunjgit: I suppose, an empty case isn't really an error since we can realistically fetch 0 series for a query

src/coordinator/storage/block.go
nikunjgit: oops

src/coordinator/storage/block.go
nikunjgit: nah, usually iterators panic out if you're not using Current with Next properly

src/coordinator/storage/block.go
nikunjgit: that's gonna be hard since we hide Values() behind an interface instead of exposing the underlying slice. Will revisit later if this becomes a bottleneck

src/coordinator/storage/block.go
nikunjgit: changed it to make more clear. It's actually the number of Steps so is the same for every series.

src/coordinator/storage/block.go
nikunjgit: changed it to make more clear. It's actually the number of Steps so is the same for every series.

src/coordinator/storage/block.go
nikunjgit: bounds is on meta, and its already exposing it

src/coordinator/storage/block.go
nikunjgit: that's gonna be hard since we hide Values() behind an interface instead of exposing the underlying slice. Will revisit later if this becomes a bottleneck

src/coordinator/storage/block.go
nikunjgit: i'll add a few cases

src/coordinator/block/column.go
```diff
@@ -0,0 +1,163 @@
...
+	return colStep{
+		time:   t,
+		values: col.Values,
+	}
+}
+
+// Len returns the total length ignoring current iterator position
```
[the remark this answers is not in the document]
nikunjgit: made it clearer

src/coordinator/executor/transform/controller.go
nikunjgit: lets you switch out different implementations, but yeah its a bit early stage, might move it around later.

src/coordinator/storage/block.go
nikunjgit: CountNode shows the usage of StepIter: src/coordinator/functions/count.go
For multiBlockIter: We need another endpoint which calls FetchBlocks, I'm adding that in another diff.

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
...
+	AppendValue(index int, value float64) error
+	Build() Block
+	AddCols(num int) error
+}
+
+// Result is the result from a block query
+type Result struct {
+	Blocks []Block
+}
+
```
arnikola: nit `var EmptyResult = Result{}`, then use block.EmptyResult wherever you're returning `block.Result{}`

nikunjgit: always returning block.Result{} should be cheap enough since it's allocated an empty struct on the stack and then copying it over to the calling function.

src/coordinator/block/column.go
```diff
@@ -0,0 +1,165 @@
...
+
+// TODO: allow series iteration
+// SeriesMeta returns the metadata for each series in the block
+func (c *columnBlock) SeriesMeta() []SeriesMeta {
+	return nil
+}
+
+// Close frees up any resources
+func (c *columnBlock) Close() error {
+	return nil
```
arnikola: Add a todo

src/coordinator/block/column.go
```diff
@@ -0,0 +1,165 @@
...
+	t := bounds.Start.Add(time.Duration(index) * step)
+	if t.After(bounds.End) {
+		return time.Time{}, fmt.Errorf("out of bounds, %d", index)
+	}
+
+	return t, nil
+}
+
+// ColStep is a single column containing data from multiple series at a given time step
+type ColStep struct {
```
arnikola: Does not need to be exported

src/coordinator/block/column.go
```diff
@@ -0,0 +1,165 @@
...
+	return c.time
+}
+
+// Values for the column
+func (c ColStep) Values() []float64 {
+	return c.values
+}
+
+// NewColStep creates a new column step
+func NewColStep(t time.Time, values []float64) ColStep {
```
arnikola: Should return a `Step`

src/coordinator/block/column.go
```diff
@@ -0,0 +1,165 @@
...
+	return ColumnBlockBuilder{
+		block: &columnBlock{
+			meta: meta,
+		},
+	}
+}
+
+// AppendValue adds a value to a column at index
+func (cb ColumnBlockBuilder) AppendValue(index int, value float64) error {
+	if len(cb.block.columns) <= index {
```
arnikola: nit: `columns := cb.block.columns`

src/coordinator/block/column.go
```diff
@@ -0,0 +1,165 @@
...
+		values: col.Values,
+	}
+}
+
+// Steps returns the total steps
+func (c *colBlockIter) Steps() int {
+	return len(c.columns)
+}
+
+func calcTime(bounds Bounds, index int) (time.Time, error) {
```
arnikola: This would make more sense as a method on `Bounds`

src/coordinator/block/column.go
```diff
@@ -0,0 +1,165 @@
...
+	return c.values
+}
+
+// NewColStep creates a new column step
+func NewColStep(t time.Time, values []float64) ColStep {
+	return ColStep{time: t, values: values}
+}
+
+// NewColumnBlockBuilder creates a new column block builder
+func NewColumnBlockBuilder(meta Metadata) ColumnBlockBuilder {
```
arnikola: nit should return a `Builder`

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
arnikola: nit: this would be called `types.go` going by the m3db standard

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// Series is a single series within a block
+type Series struct {
+	values []float64
+	bounds Bounds
+}
+
+// NewSeries creates a new series
+func NewSeries(values []float64, bounds Bounds) Series {
```
arnikola: nit: shouldn't have functions defined in a interfaces/types file

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
...
+
+// Bounds are the time bounds
+type Bounds struct {
+	Start    time.Time
+	End      time.Time
+	StepSize time.Duration
+}
+
+// SeriesIter iterates through a CompressedSeriesIterator horizontally
+type SeriesIter interface {
```
arnikola: All of these iterators should have a Close/Finalize; should make a base `Iterator` interface with `Next() bool` and `Close()` on it, then have these be `Iterator`s

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// StepIter iterates through a CompressedStepIterator vertically
+type StepIter interface {
+	Next() bool
+	Current() Step
+	Steps() int
+}
+
+// Step can optionally implement iterator interface
```
arnikola: Not sure what this comment means

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
...
+	bounds Bounds
+}
+
+// NewSeries creates a new series
+func NewSeries(values []float64, bounds Bounds) Series {
+	return Series{values: values, bounds: bounds}
+}
+
+// ValueAtStep returns the datapoint value at a step index
+func (s Series) ValueAtStep(idx int) float64 {
```
arnikola: nit: using `idx` here, `index` elsewhere. Might be good to standardize on something like `i` (my vote!)

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+// SeriesMeta is metadata data for the series
+type SeriesMeta struct {
+	Tags models.Tags
+}
+
+// Bounds are the time bounds
+type Bounds struct {
+	Start    time.Time
```
arnikola: Should these be mutable?

src/coordinator/block/interface.go
```diff
@@ -0,0 +1,122 @@
...
+type Metadata struct {
+	Bounds Bounds
+	Tags   models.Tags // Common tags across different series
+}
+
+// Builder builds a new block
+type Builder interface {
+	AppendValue(index int, value float64) error
+	Build() Block
+	AddCols(num int) error
```
arnikola: Is a `Builder` necessarily a column builder?

src/coordinator/functions/count.go
```diff
@@ -57,22 +58,28 @@ type CountNode struct {
 }
 
 // Process the block
-func (c *CountNode) Process(ID parser.NodeID, block storage.Block) error {
+func (c *CountNode) Process(ID parser.NodeID, block block.Block) error {
 	builder, err := c.controller.BlockBuilder(block.Meta())
```
arnikola: defer block.Close()

src/coordinator/functions/count.go
```diff
@@ -57,22 +58,28 @@ type CountNode struct {
...
+			if !math.IsNaN(value) {
+				count++
+			}
 		}
 
-		builder.AppendValue(index, sum)
+		builder.AppendValue(index, count)
 	}
 
 	return c.controller.Process(builder.Build())
```
arnikola: This function does not seem to conform with our design where we would reuse the input block as the "output". Would make more sense to have the block columns have a Reset() method or something, and set that to `[]int{count}`

src/coordinator/block/column.go
nikunjgit: yeah, used in storage/block.go

src/coordinator/block/interface.go
nikunjgit: fixed

src/coordinator/block/interface.go
nikunjgit: using i as a function parameter seems a little generic to me. Let's make it idx then ?

src/coordinator/block/interface.go
nikunjgit: we copy structs everywhere instead of using pointers to make them immutable.

src/coordinator/block/interface.go
nikunjgit: might have more methods as the need arises.

src/coordinator/functions/count.go
nikunjgit: so there are cases when we can't use input block. Eg: if input block is compressed, it won't have any slices allocated. I would suggest that we put the logic of pooling blocks within builder and individual functions don't have any idea about where the blocks are coming from.

src/coordinator/block/column.go
```diff
@@ -0,0 +1,161 @@
...
+func NewColumnBlockBuilder(meta Metadata) Builder {
+	return ColumnBlockBuilder{
+		block: &columnBlock{
+			meta: meta,
+		},
+	}
+}
+
+// AppendValue adds a value to a column at index
+func (cb ColumnBlockBuilder) AppendValue(index int, value float64) error {
```
arnikola: nit idx

src/coordinator/block/column.go
```diff
@@ -0,0 +1,161 @@
...
+// Close frees up any resources
+// TODO: actually free up the resources
+func (c *columnBlock) Close() error {
+	return nil
+}
+
+type colBlockIter struct {
+	columns []column
+	meta    Metadata
+	index   int
```
arnikola: nit idx

src/coordinator/block/column.go
```diff
@@ -0,0 +1,161 @@
...
+	c.index++
+	return c.index < len(c.columns)
+}
+
+// Current returns the current step
+func (c *colBlockIter) Current() Step {
+	col := c.columns[c.index]
+	t, err := c.meta.Bounds.TimeForIndex(c.index)
+	if err != nil {
+		panic(err)
```
arnikola: nit: add some additional text to panic for additional context

src/coordinator/ts/series.go
```diff
@@ -49,3 +54,64 @@ func (b *Series) Len() int { return b.vals.Len() }
 
 // Values returns the underlying values interface
 func (b *Series) Values() Values { return b.vals }
+
+// Align adjusts the datapoints to start, end and a fixed interval
+func (b *Series) Align(start, end time.Time, interval time.Duration) (*Series, error) {
```
arnikola: nit: `b` is a weird variable for *Series

src/coordinator/ts/series.go
```diff
@@ -49,3 +54,64 @@ func (b *Series) Len() int { return b.vals.Len() }
...
+	fixedVals, err := alignValues(b.Values(), start, end, interval)
+	if err != nil {
+		return nil, err
+	}
+
+	return NewSeries(b.name, fixedVals, b.Tags), nil
+}
+
+func alignValues(values Values, start, end time.Time, interval time.Duration) (FixedResolutionMutableValues, error) {
```
arnikola: nit: make this a receiver function on *Series and use b.Values()

src/coordinator/ts/series.go
```diff
@@ -49,3 +54,64 @@ func (b *Series) Len() int { return b.vals.Len() }
...
+// SeriesList represents a slice of series pointers
+type SeriesList []*Series
+
+// Resolution returns the resolution for a fixed step series list. It assumes all underlying series will have the same resolution
+func (seriesList SeriesList) Resolution() (time.Duration, error) {
+	var resolution time.Duration
+	for i, s := range seriesList {
+		fixedRes, ok := s.Values().(FixedResolutionMutableValues)
+		if !ok {
+			return 0, errors.New("only fixed resolution supported")
```
arnikola: nit: make this error a constant

src/coordinator/ts/series.go
```diff
@@ -49,3 +54,64 @@ func (b *Series) Len() int { return b.vals.Len() }
...
+		}
+
+		if i == 0 {
+			resolution = fixedRes.Resolution()
+		} else {
+			if resolution != fixedRes.Resolution() {
+				return 0, fmt.Errorf("resolution mismatch, r1: %v, r2: %v", resolution, fixedRes.Resolution())
```
arnikola: nit: make this error a constant

src/coordinator/ts/series.go
```diff
@@ -49,3 +54,64 @@ func (b *Series) Len() int { return b.vals.Len() }
...
+			}
+		}
+	}
+
+	return resolution, nil
+}
+
+// Align aligns each series to the given start, end and step.
+func (seriesList SeriesList) Align(start, end time.Time, interval time.Duration) (SeriesList, error) {
```
arnikola: Seems better to do this inplace on the existing seriesList to reduce allocations

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,153 @@
...
+		values[i] = s.Values().ValueAt(i)
+	}
+	return block.NewSeries(values, m.block.meta.Bounds)
 }
 
-// BlockMetadata is metadata for a block
-type BlockMetadata struct {
-	Bounds Bounds
-	Tags   models.Tags // Common tags across different series
+func (m *multiSeriesBlockSeriesIter) Close()  {
```
arnikola: nit Add a todo

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,153 @@
...
+	if len(m.block.seriesList) == 0 {
+		return 0
+	}
+
+	return m.block.seriesList[0].Values().Len()
 }
 
-// StepMeta is metadata data for a single time step
-type StepMeta struct {
+func (m *multiSeriesBlockStepIter) Close() {
```
arnikola: nit: Add a todo

src/coordinator/storage/block.go
```diff
@@ -1,58 +1,153 @@
...
+	t := bounds.Start.Add(time.Duration(m.index) * bounds.StepSize)
+	return block.NewColStep(t, values)
+}
+
+func (m *multiSeriesBlockStepIter) Steps() int {
+	if len(m.block.seriesList) == 0 {
+		return 0
+	}
+
+	return m.block.seriesList[0].Values().Len()
```
arnikola: nit: Would it be a good idea to store `m.block.seriesList[0].Values().Len()` for steps?

src/coordinator/block/column.go
nikunjgit: added todo, will do that in the next diff

src/coordinator/ts/series.go
nikunjgit: i kept it separate since it's easier to test this way (Although i need to write more tests :D). This also makes it clearer that the underlying series isn't being mutated.

src/coordinator/ts/series.go
nikunjgit: in coordinator we haven't been using constants for fmt.Errorf since they have to be reformatted based on the parameters.

src/coordinator/ts/series.go
nikunjgit: It's cleaner this way since if the thing errors out in between then seriesList gets in a strange state. Also, we won't be able to inplace replace Datapoint -> FixedRes values so might not end up saving a lot. Let's see if this becomes a bottleneck ?

src/coordinator/storage/block.go
nikunjgit: don't expect it to be called within a loop (usually outside)

src/coordinator/ts/series.go
arnikola: Might be better to call it something like `generateAlignedValues` or something then? `alignValues` implies mutation

src/coordinator/block/types.go
```diff
@@ -53,7 +53,7 @@ type Bounds struct {
 	StepSize time.Duration
 }
 
-// TimeForIndex returns the start time for a given index assuming a uniform step size
+// TimeForIndex returns the start time for a given idx assuming a uniform step size
```
arnikola: nit keep `index` in comments?

src/coordinator/block/series.go
```diff
@@ -36,7 +36,7 @@ func NewSeries(values []float64, bounds Bounds) Series {
 	return Series{values: values, bounds: bounds}
 }
 
-// ValueAtStep returns the datapoint value at a step index
+// ValueAtStep returns the datapoint value at a step idx
```
arnikola: nit keep `index` in comments?

src/coordinator/block/column.go
```diff
@@ -132,14 +133,14 @@ func NewColumnBlockBuilder(meta Metadata) Builder {
 	}
 }
 
-// AppendValue adds a value to a column at index
-func (cb ColumnBlockBuilder) AppendValue(index int, value float64) error {
+// AppendValue adds a value to a column at idx
```
arnikola: nit keep `index` in comments?

PR #748, opened 2018-06-05
Fix a few PHONY rules in Makefile

isaachier:


Makefile
```diff
@@ -191,8 +191,17 @@ docker-integration-test:
 	@./scripts/integration-tests/docker-integration-test.sh
 	@cd scripts/integration-tests/prometheus/ && ./prometheus-integration-test.sh
 
-.PHONY: metalint
-metalint: $(patsubst %,metalint-%,$(SUBDIRS))
+SUBDIR_TARGETS = mock-gen thrift-gen proto-gen asset-gen all-gen metalint test \
+				 test-xml
```
justinjc: Can you add `test-ci-unit` and `test-ci-integration` to this list?

isaachier: Sure thing. Guess I missed those.

PR #754, opened 2018-06-12
Merge m3nsch into monorepo

justinjc:


Makefile
```diff
@@ -37,12 +37,15 @@ include $(SELF_DIR)/src/dbnode/generated-source-files.mk
...
+	m3coordinator \
+	m3nsch_server \
+	m3nsch_client \
 
 SUBDIRS :=    \
 	cmd         \
 	dbnode      \
-	coordinator
+	coordinator \
+	nsch        \
```
prateek: maybe just call the subdir `m3nsch`? reads really weird right now

glide.yaml
```diff
@@ -149,6 +156,17 @@ import:
 - package: github.com/mjibson/esc
   version: ^0.1.0
 
+- package: github.com/gavv/monotime
+  version: 47d58efa69556a936a3c15eb2ed42706d968ab01
+
+- package: google.golang.org/grpc
+  version: 401e0e00e4bb830a10496d64cd95e068c5bf50de
```
prateek: mind using a later version (e.g. `^1.12`)?

prateek: upstream repo is https://github.com/grpc/grpc-go

src/nsch/agent/agent.go
```diff
@@ -0,0 +1,356 @@
...
+			if msg.update {
+				tickLoop.Stop()
+				timeUnit, namespace, fakeNow, tickPeriod = ms.workerParams()
+				tickLoop = time.NewTicker(tickPeriod)
+			}
+
+		case <-tickLoop.C:
+			fakeNow = fakeNow.Add(tickPeriod)
+			metric := ms.nextWorkerMetric(workerIdx)
+			start := monotime.Now()
```
prateek: mind migrating this to `time.Now()`?

src/nsch/agent/agent.go
```diff
@@ -0,0 +1,356 @@
...
+			err := ms.params.fn(workerIdx, ms.session, namespace, metric, fakeNow, timeUnit)
+			elapsed := monotime.Since(start)
```
prateek: this can be `time.Since(start)`

glide.yaml
```diff
@@ -91,6 +92,9 @@ import:
 - package: google.golang.org/appengine/datastore
   version: 2e4a801b39fc199db615bfca7d0b9f8cd9580599
 
+- package: github.com/pborman/getopt
+  version: ec82d864f599c39673eef89f91b93fa5576567a1
```
prateek: mind using a version instead?

glide.yaml
```diff
@@ -99,6 +103,9 @@ import:
 - package: github.com/spf13/pflag
   version: 4f9190456aed1c2113ca51ea9b89219747458dc1
 
+- package: github.com/spf13/viper
+  version: 7538d73b4eb9511d85a9f1dfef202eeb8ac260f4
```
prateek: same here

glide.yaml
```diff
@@ -149,6 +156,17 @@ import:
 - package: github.com/mjibson/esc
   version: ^0.1.0
 
+- package: github.com/gavv/monotime
```
prateek: go1.9 added monotonic clock support so we don't need the `monotime` dep any more.

glide.lock
```diff
@@ -303,6 +303,8 @@ imports:
   - pbutil
 - name: github.com/mjibson/esc
   version: 58d9cde84f237ecdd89bd7f61c2de2853f4c5c6e
+- name: github.com/gavv/monotime
```
prateek: can remove this

glide.yaml
justinjc: Doesn't look like getopt has any semver or tags though.

prateek: hm does the branch `v2` work? no biggie if not .

justinjc:
Going to pin grpc to `~1.7.3`, since 1.12.x marks a bunch of load balancing functions as deprecated, making metalint complain and thus break the build.

PR #756, opened 2018-06-15
Use the same timestamp for all Tick() logic as we do for Flush() logic

richardartoul:
There is currently a bug in the interaction between the Tick logic and Flush logic. Sequence of events:

1.  Tick() runs, and sometime during the tick the current time reaches the threshold such that calls to shard.removeAnyFlushStatesTooEarly() can remove the flush state for the oldest block.
2. After tick completes, Flush() runs, but its using the pre-stick timestamp in which the oldest block is still valid. It will detect that their is no flush state for the oldest block for some of the shards and attempt to flush it.
3. Shard tries to flush oldest block (which it doesn't have any data for anyways) and fails because the file still exists on disk, but hasn't been cleaned up yet.

To address the issue, I've flowed the tickStart timestamp through the tick logic as well. I additionally flowed it into some tick functions where its not needed, just so that if we ever change those functions to require a timestamp its already there and developers are more likely to use that than a (possibly buggy) usage of time.Now()

### DIFF
--- a/src/dbnode/storage/types.go
+++ b/src/dbnode/storage/types.go
@@ -224,7 +224,7 @@ type databaseNamespace interface {
 	GetIndex() (namespaceIndex, error)
 
 	// Tick performs any regular maintenance operations
-	Tick(c context.Cancellable) error
+	Tick(c context.Cancellable, timeBeforeTickStart time.Time) error
 
 	// Write writes a data point
 	Write(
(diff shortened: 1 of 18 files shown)


src/dbnode/storage/types.go
```diff
@@ -224,7 +224,7 @@ type databaseNamespace interface {
 	GetIndex() (namespaceIndex, error)
 
 	// Tick performs any regular maintenance operations
-	Tick(c context.Cancellable) error
+	Tick(c context.Cancellable, timeBeforeTickStart time.Time) error
```
prateek: what do you mean by `timeBeforeTickStart`?

PR #757, opened 2018-06-15
Add support for reading from snapshot files

richardartoul:
Add supports from reading from snapshot files during commit log bootstrap, including unit, integration, and prop tests.


src/dbnode/generated-source-files.mk
```diff
@@ -127,6 +128,19 @@ genny-map-storage-repair: install-m3x-repo
 		value_type=ReplicaSeriesBlocksMetadata        \
 		target_package=$(m3db_package)/src/dbnode/storage/repair
 
+# Map generation rule for storage/bootstrap/bootstrapper/commitlog
+.PHONY: genny-map-storage-bootstrap-bootstrapper-commitlog
+genny-map-storage-bootstrap-bootstrapper-commitlog: install-m3x-repo
```
richardartoul: Add a regression test for unique index thing

richardartoul: Add mixed_mode equivalent integration test

src/dbnode/persist/fs/files.go
```diff
@@ -292,6 +298,55 @@ func timeAndIndexFromFileName(fname string, componentPosition int) (time.Time, i
 	return t, int(index), nil
 }
 
+// SnapshotTime returns the time at which the snapshot was taken.
+func SnapshotTime(
```
richardartoul: See if I can just include this in the return from SnapshotFiles()

src/dbnode/storage/block/block.go
```diff
@@ -413,21 +413,34 @@ func (b *dbBlock) resetRetrievableWithLock(
 	b.wasRetrievedFromDisk = false
 }
 
+func (b *dbBlock) Discard() ts.Segment {
```
richardartoul: Fix locking / pooling issue here

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -91,6 +109,24 @@ func (s *commitLogSource) AvailableData(
 	return shardsTimeRanges
 }
 
+// ReadData will read a combination of the available snapshot files and commit log files to
```
richardartoul: TODO: Add concrete example

src/dbnode/generated-source-files.mk
richardartoul: done (adding regression test for unique index bug)

src/dbnode/integration/admin_session_fetch_blocks_test.go
```diff
@@ -178,6 +179,9 @@ func testSetupToSeriesMaps(
 	for shardID, metadatas := range metadatasByShard {
 		blocksIter, err := session.FetchBlocksFromPeers(nsMetadata, shardID,
 			consistencyLevel, metadatas, resultOpts)
+		if err != nil {
+			panic(err)
```
robskillington: We perform a `require.NoError(t, err)` just below - whats the benefit of a panic here just before the check?

src/dbnode/persist/fs/files.go
```diff
@@ -88,25 +88,31 @@ func (f FileSetFilesSlice) Filepaths() []string {
...
 	for i, curr := range f {
 		if curr.ID.BlockStart.Equal(blockStart) {
-			isEnd := i == len(f)-1
-			isHighestIdx := true
-			if !isEnd {
-				next := f[i+1]
-				if next.ID.BlockStart.Equal(blockStart) && next.ID.VolumeIndex > curr.ID.VolumeIndex {
-					isHighestIdx = false
+			var bestSoFar FileSetFile
+			var bestSoFarExists bool
```
robskillington: nit:

src/dbnode/persist/fs/files.go
robskillington: +1

src/dbnode/persist/fs/reader_open_options_matcher.go
```diff
@@ -48,6 +48,9 @@ func (m ReaderOpenOptionsMatcher) Matches(x interface{}) bool {
 	if !m.ID.BlockStart.Equal(readerOpenOptions.Identifier.BlockStart) {
 		return false
 	}
+	if m.ID.VolumeIndex != (readerOpenOptions.Identifier.VolumeIndex) {
```
robskillington: nit: Don't need the brackets on the right hand side evaluation here.

src/dbnode/integration/generate/writer.go
```diff
@@ -117,20 +176,28 @@ func writeToDisk(
...
 		}
 		if err := writer.Open(writerOpts); err != nil {
 			return err
 		}
 		for _, series := range seriesList {
 			encoder.Reset(start, 0)
 			for _, dp := range series.Data {
-				if err := encoder.Encode(dp, xtime.Second, nil); err != nil {
-					return err
+				if pred(dp) {
```
robskillington: nit: Slightly more Go-like to:

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -91,6 +109,71 @@ func (s *commitLogSource) AvailableData(
...
+//    W3 comes in at 12:31PM
+//    W4 comes in at 2:04PM
+//
+//    1) W1 only present in commit log with start time 11:50PM
+//    2) W2 only present in commit log with start time 12:20PM
+//    3) W3 only present in commit log with start time 12:30PM
+//    4) W4 only present in commit log with start time 2:00PM
+//
+//    Need to read all commit logs that contain writes with system times spanning from
+//    11:55PM -> 2:10PM which will bootstrap all of the data points above. I.E:
```
robskillington: nit: This should be `11:55AM` yeah?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -100,46 +183,82 @@ func (s *commitLogSource) ReadData(
...
+		seriesSkipped     int
+		datapointsSkipped int
+		datapointsRead    int
+
+		// TODO(rartoul): When we implement caching data across namespaces, this will need
+		// to be commitlog.ReadAllSeriesPredicate() if CacheSeriesMetadata() is enabled
+		// because we'll need to read data for all namespaces, not just the one we're currently
+		// bootstrapping.
+		readSeriesPredicate = func(id ident.ID, namespace ident.ID) bool {
+			shouldReadSeries := nsID.Equal(namespace)
```
robskillington: Hm bugger, yeah this is a pain to have to check for each datapoint.  There's no easy way to make this an integer check or something with some negotiation function?  I guess we'd have to change the actual file format to be an int + namespace string. Then if you receive an int that you haven't seen before you insert it as a new namespace in your callback map or something.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -100,46 +183,82 @@ func (s *commitLogSource) ReadData(
...
+			CommitLogOptions:      s.opts.CommitLogOptions(),
+			FileFilterPredicate:   readCommitLogPred,
+			SeriesFilterPredicate: readSeriesPredicate,
+		}
+	)
+
+	defer func() {
+		s.log.Infof("seriesSkipped: %d", seriesSkipped)
+		s.log.Infof("datapointsSkipped: %d", datapointsSkipped)
+		s.log.Infof("datapointsRead: %d", datapointsRead)
```
robskillington: Should we add a `WithFields(xlog.NewField("bootstrapper": "commitlog"))` or something so we know where this is coming from?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
 	s.logEncodingOutcome(workerErrs, iter)
 
-	result := s.mergeShards(int(numShards), bopts, blockSize, blopts, encoderPool, shardDataByShard)
-	// After merging shards, its safe to cache the shardData (which involves some mutation).
-	if s.shouldCacheSeriesMetadata(runOpts, ns) {
-		s.cacheShardData(ns, shardDataByShard)
+	// Merge all the different encoders from the commit log that we created with
+	// the data that is available in the snapshot files.
+	mergeStart := time.Now()
+	s.log.Infof("Starting merge...")
```
robskillington: nit: Lowercase the log entry for consistency.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+		snapshotFilesByShard,
+		mostRecentCompleteSnapshotByBlockShard,
+		int(numShards),
+		blockSize,
+		shardDataByShard,
+	)
+	if err != nil {
+		return nil, err
+	}
+	s.log.Infof("Done merging..., took: %s", time.Since(mergeStart).String())
```
robskillington: nit: Lowercase the log entry for consistency.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+) map[xtime.UnixNano]map[uint32]snapshotMetadata {
+	var (
+		minBlock, maxBlock              = shardsTimeRanges.MinMax()
+		decoder                         = msgpack.NewDecoder(nil)
+		mostRecentSnapshotsByBlockShard = map[xtime.UnixNano]map[uint32]snapshotMetadata{}
+	)
+
+	for currBlockStart := minBlock.Truncate(blockSize); currBlockStart.Before(maxBlock); currBlockStart = currBlockStart.Add(blockSize) {
+		for shard := range shardsTimeRanges {
+			// Anonymous func for easier clean up using defer.
```
robskillington: +1

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+	mostRecentCompleteSnapshotByBlockShard map[xtime.UnixNano]map[uint32]snapshotMetadata,
+) (result.ShardResult, error) {
+	var (
+		shardResult    result.ShardResult
+		allSeriesSoFar *result.Map
+		rangeIter      = shardTimeRanges.Iter()
+		err            error
+	)
+
+	for hasMore := rangeIter.Next(); hasMore; hasMore = rangeIter.Next() {
```
robskillington: Can't this just be `for rangeIter.Next() { ... }`?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+		var (
+			currRange             = rangeIter.Value()
+			currRangeDuration     = currRange.End.Unix() - currRange.Start.Unix()
```
robskillington: Can't this more simply be `currRange.End.Sub(currRange.Start)`?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+			if mostRecentCompleteSnapshotForShardBlock.snapshotTime.Equal(blockStart) ||
+				// Should never happen
+				mostRecentCompleteSnapshotForShardBlock.isZero() {
+				// There is no snapshot file for this time, and even if there was, there would
+				// be no point in reading it. In this specific case its not an error scenario
+				// because the fact that snapshotTime == blockStart means we already accounted
+				// for the fact that this snapshot did not exist when we were deciding which
+				// commit logs to read.
+				s.log.Infof(
+					"No snapshots for shard: %d and blockStart: %d",
```
robskillington: nit: Lowercase the log entry for consistency.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+					shard, blockStart.Unix())
+				continue
+			}
+
+			shardResult, err = s.bootstrapShardBlockSnapshot(
+				nsID, shard, blockStart, metadataOnly, shardResult, allSeriesSoFar, blockSize,
+				snapshotFiles, mostRecentCompleteSnapshotForShardBlock)
+			if err != nil {
+				return shardResult, err
```
robskillington: Should this just be `return shardResult, nil`?  `err` has already been checked.

robskillington: Also should we `log.Warnf(...)` the error at least?  Otherwise seems like it just gets swallowed.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+		blocksPool = blOpts.DatabaseBlockPool()
+		bytesPool  = blOpts.BytesPool()
+		fsOpts     = s.opts.CommitLogOptions().FilesystemOptions()
+		idPool     = s.opts.CommitLogOptions().IdentifierPool()
+	)
+
+	// Bootstrap the snapshot file
+	reader, err := s.newReaderFn(bytesPool, fsOpts)
+	if err != nil {
+		return shardResult, err
```
robskillington: Should this just be `return nil, err` perhaps?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+		Identifier: fs.FileSetFileIdentifier{
+			Namespace:   nsID,
+			BlockStart:  blockStart,
+			Shard:       shard,
+			VolumeIndex: mostRecentCompleteSnapshot.ID.VolumeIndex,
+		},
+		FileSetType: persist.FileSetSnapshotType,
+	})
+	if err != nil {
+		return shardResult, err
```
robskillington: Should this just be `return nil, err` perhaps?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+	}
+
+	s.log.Infof(
+		"Reading snapshot for shard: %d and blockStart: %d and volume: %d",
```
robskillington: nit: Lowercase the log entry for consistency.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+			data             checked.Bytes
+			expectedChecksum uint32
+		)
+		if metadataOnly {
+			id, tagsIter, _, _, err = reader.ReadMetadata()
+		} else {
+			id, tagsIter, data, expectedChecksum, err = reader.Read()
+		}
+		if err != nil && err != io.EOF {
+			return shardResult, err
```
robskillington: Should this just be `return nil, err` perhaps?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+
+		if err == io.EOF {
+			break
+		}
+
+		var tags ident.Tags
+		if tagsIter.Remaining() > 0 {
+			tags, err = convert.TagsFromTagsIter(id, tagsIter, idPool)
+			if err != nil {
+				return shardResult, fmt.Errorf("unable to decode tags: %v", err)
```
robskillington: Should this just be `return nil, err` perhaps?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+			}
+		}
+		tagsIter.Close()
```
robskillington: Maybe move this below the `existing, ...` check so we don't always decode the tags?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+			}
+		}
+
+		if shardResult == nil {
+			// Delay initialization so we can estimate size.
+			shardResult = result.NewShardResult(reader.Entries(), s.opts.ResultOptions())
+		}
+
+		// Mark the ID and Tags as no finalize because we're sharing them across blocks.
+		// TODO: Not sure this is necessary...
```
robskillington: Hey so this is not necessary I believe, since the shardResult keeps series metadata just once per collection of blocks.

robskillington: I added this to m3x here:

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+	// To start, for each block that we're bootstrapping, we need to figure out the most recent
+	// snapshot that was taken for each shard. I.E we want to create a datastructure that looks
+	// like this:
+	// 		map[blockStart]map[shard]mostRecentSnapshotTime
+	mostRecentCompleteSnapshotByBlockShard := s.mostRecentCompleteSnapshotByBlockShard(
+		shardsTimeRanges, blockSize, snapshotFilesByShard, s.opts.CommitLogOptions().FilesystemOptions())
+	for block, mostRecentByShard := range mostRecentCompleteSnapshotByBlockShard {
+		for shard, mostRecent := range mostRecentByShard {
+			s.log.Infof(
+				"Most recent snapshot for block: %d and shard: %d is %d",
```
robskillington: nit: Lowercase the log entry for consistency.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+	// snapshot for that block across all shards. This will help us determine how much of the commit
+	// log we need to read. The new data structure we're trying to generate looks like:
+	// 		map[blockStart]minimumMostRecentSnapshotTime (across all shards)
+	// This structure is important because it tells us how much of the commit log we need to read for
+	// each block that we're trying to bootstrap (because the commit log is shared across all shards.)
+	minimumMostRecentSnapshotTimeByBlock := s.minimumMostRecentSnapshotTimeByBlock(
+		shardsTimeRanges, blockSize, mostRecentCompleteSnapshotByBlockShard)
+	for block, minSnapshotTime := range minimumMostRecentSnapshotTimeByBlock {
+		s.log.Infof(
+			"Min snapshot time for block: %d is: %d",
```
robskillington: nit: Lowercase the log entry for consistency.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+		for _, rangeToCheck := range rangesToCheck {
+			commitLogEntryRange := xtime.Range{
+				Start: fileStart,
+				End:   fileStart.Add(fileBlockSize),
+			}
+
+			if commitLogEntryRange.Overlaps(rangeToCheck) {
+				s.log.
+					Infof(
+						"Opting to read commit log: %s with start: %d and duration: %s",
```
robskillington: nit: Lowercase the log entry for consistency.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,499 @@ func (s *commitLogSource) ReadData(
...
+						fileName, fileStart.Unix(), fileBlockSize.String())
+				return true
+			}
+		}
+
+		s.log.
+			Infof(
+				"Opting to skip commit log: %s with start: %d and duration: %s",
```
robskillington: nit: Lowercase the log entry for consistency.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -336,44 +930,81 @@ func (s *commitLogSource) shouldIncludeInIndex(
...
 		// Controls how many shards can be merged in parallel
 		workerPool          = xsync.NewWorkerPool(s.opts.MergeShardsConcurrency())
 		bootstrapResultLock sync.Mutex
 		wg                  sync.WaitGroup
 	)
-
 	workerPool.Init()
 
 	for shard, unmergedShard := range unmerged {
+		snapshotData, err := s.bootstrapShardSnapshots(
```
robskillington: We could possibly read the snapshots in parallel too yeah?  We could just add this to the worker func below which we spawn off?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -384,48 +1015,84 @@ func (s *commitLogSource) mergeShards(
...
+		multiReaderIteratorPool = blOpts.MultiReaderIteratorPool()
+		segmentReaderPool       = blOpts.SegmentReaderPool()
+		encoderPool             = blOpts.EncoderPool()
+	)
+
+	numSeries := 0
+	if unmergedShard.series != nil {
+		numSeries = unmergedShard.series.Len()
+	}
+	var shardResult = result.NewShardResult(numSeries, s.opts.ResultOptions())
```
robskillington: nit: Combine these `var ...` three statements into a var block?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -434,30 +1101,34 @@ func (s *commitLogSource) mergeSeries(
...
-		if len(encoders) == 1 {
-			pooledBlock := blocksPool.Get()
-			pooledBlock.Reset(start, blockSize, encoders[0].enc.Discard())
-			if seriesBlocks == nil {
-				seriesBlocks = block.NewDatabaseSeriesBlocks(len(unmergedBlocks.encoders))
-			}
-			seriesBlocks.AddBlock(pooledBlock)
+		if !hasSnapshotBlock {
+			// Make sure snapshotBlock is nil if it does not exist.
+			snapshotBlock = nil
```
robskillington: Hm, huh?  Shouldn't `BlockAt(...)` guarentee `<notnil>, true` or `<nil>, false`?  And if not should we not just update it to guarentee that?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -489,10 +1164,31 @@ func (s *commitLogSource) mergeSeries(
...
+				seriesBlocks = block.NewDatabaseSeriesBlocks(len(allSnapshotBlocks))
+			}
+			_, ok := seriesBlocks.BlockAt(startNano.ToTime())
+			if ok {
+				// Should never happen because we would have called
+				// Blocks.RemoveBlockAt() above.
+				iOpts := s.opts.CommitLogOptions().InstrumentOptions()
+				invariantLogger := instrument.EmitInvariantViolationAndGetLogger(iOpts)
+				invariantLogger.Errorf(
+					"Tried to merge block that should have been removed, blockStart: %d", startNano)
```
robskillington: nit: Lowercase the log entry for consistency.

src/dbnode/integration/admin_session_fetch_blocks_test.go
richardartoul: Fixed

src/dbnode/integration/generate/writer.go
richardartoul: sure

src/dbnode/persist/fs/files.go
richardartoul: 👍

src/dbnode/persist/fs/reader_open_options_matcher.go
richardartoul: Lol I think that was just copy pasta from the check right above it. Fixed

src/dbnode/storage/index/convert/convert.go
```diff
@@ -173,6 +173,49 @@ func FromMetricIterNoClone(id ident.ID, tags ident.TagIterator) (doc.Document, e
 	}, nil
 }
 
+// TagsFromTagsIter returns an ident.Tags from a TagIterator. It also tries
+// to re-use bytes from the seriesID if they're also present in the tags
+// instead of re-allocating them.
```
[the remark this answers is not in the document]
richardartoul: good call

src/dbnode/storage/index/convert/convert.go
```diff
@@ -173,6 +173,49 @@ func FromMetricIterNoClone(id ident.ID, tags ident.TagIterator) (doc.Document, e
...
+			tag.Value = idPool.Clone(curr.Value)
+		}
+
+		if idRef {
+			tag.NoFinalize() // Taken ref, cannot finalize this
+		}
+
+		tags.Append(tag)
+	}
+	return tags, iter.Err()
```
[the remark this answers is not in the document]
richardartoul: 👍

src/dbnode/storage/series/series.go
```diff
@@ -660,7 +660,22 @@ func (s *dbSeries) Snapshot(
 		return errSeriesNotBootstrapped
 	}
 
-	stream, err := s.buffer.Snapshot(ctx, blockStart)
+	var (
```
richardartoul: TODO: Add a test for this

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: yep, fixed thanks

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Are you referring to do the TODO (which will require some difficult juggling to match data by namespace without doing an expensive bytes comparison), or the existing readSeriesPredicate? The existing readSeriesPredicate is efficient because it gets pushed down to the commit log reader and cached in the metadata map so we only have to do it once per series instead of once per datapoint

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Its usually pretty obvious cause of the other logs, but sure I can add that

richardartoul: I added it on the struct-level logger so all logs in all these functions get it for free

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: yep exactly, seemed worth it as its only a few hundred -> thousand small allocations for a process that takes minutes to complete

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: ...yes. I don't know why I did that

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: You're right and also my logic is completely broken. The unix subtraction results a number that is veryt small compared to the blockSize (which is time.Duration and thus measured in nanoseconds) so integer division always returns 0.

I'll change it to use the time library as you suggested, and then use a modulus operator to compare the two time.Durations which should give the correct results.

Not a huge deal because this is basically a defensive programming assertion that should never fail, but good catch.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Thats a good call, although before I do that can you remind me how are guarantees work if someone indexes the same series with different tags across different index blocks? What if they try and index it twice within the same index block with different tags also?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Yeah that is the contract that function provides (and correctly implements). I was just being paranoid because newIOReadersFromEncodersAndBlock does a nil check that I didn't want to think about whether it would work properly if there was an empty block

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/generated-source-files.mk
richardartoul: done (adding mixed_mode equivalent integration test)

src/dbnode/persist/fs/files.go
richardartoul: done

src/dbnode/storage/block/block.go
richardartoul: I'm actually not sure there is a good solution here. We could maybe do something asynchronous maybe, but it would expensive / slower. This actually seems like the best solution and we handle the closing/pooling of the series object the same way.

Since this object will get placed into the "back" of the pool, it seems unlikely to cause performance issues, and any alternatives will probably be slower.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
robskillington: The TODO for avoiding expensive bytes comparison.  Yeah, makes sense.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
robskillington: Nice, good stuff.

src/dbnode/persist/fs/files.go
```diff
@@ -51,6 +51,11 @@ const (
 
 	commitLogComponentPosition    = 2
 	indexFileSetComponentPosition = 2
+
+	// This is the default size that the standard library uses. We've
+	// redefined it here to make some of public APIs nicer and not
+	// require the user to specify this value.
+	defaultBufioReaderSize = 4096
```
robskillington: You could probably make this a var and infer it yeah?  i.e. `var defaultBufioReaderSize = bufio.NewReader(nil).Size()`, I tested this and seems to work.

src/dbnode/integration/generate/writer.go
```diff
@@ -117,20 +176,30 @@ func writeToDisk(
...
+
 				if err := encoder.Encode(dp, xtime.Second, nil); err != nil {
 					return err
 				}
 			}
+
 			stream := encoder.Stream()
 			if stream == nil {
-				return fmt.Errorf("nil stream")
+				// None of the datapoints passed the predicate.
```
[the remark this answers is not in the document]
robskillington: Can we open an issue for this?

src/dbnode/storage/block/block.go
robskillington: Can you change it to this?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,515 @@ func (s *commitLogSource) ReadData(
...
+				}
+
+				// Make sure we're able to read the snapshot time. This will also set the
+				// CachedSnapshotTime field so that we can rely upon it from here on out.
+				_, err := mostRecentSnapshotVolume.SnapshotTime()
+				if err != nil {
+					s.log.
+						WithFields(
+							xlog.NewField("namespace", mostRecentSnapshot.ID.Namespace),
+							xlog.NewField("blockStart", mostRecentSnapshot.ID.BlockStart),
```
robskillington: Perhaps best to log this as string?  `BlockStart.String()`

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,515 @@ func (s *commitLogSource) ReadData(
...
+				// Should never happen
+				mostRecentCompleteSnapshotForShardBlock.IsZero() {
+				// There is no snapshot file for this time, and even if there was, there would
+				// be no point in reading it. In this specific case its not an error scenario
+				// because the fact that snapshotTime == blockStart means we already accounted
+				// for the fact that this snapshot did not exist when we were deciding which
+				// commit logs to read.
+				s.log.Infof(
+					"no snapshots for shard: %d and blockStart: %d",
+					shard, blockStart.Unix())
```
robskillington: Perhaps best to log this as string?  `blockStart.String()`

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,515 @@ func (s *commitLogSource) ReadData(
...
+		var (
+			currRange             = rangeIter.Value()
+			currRangeDuration     = currRange.End.Sub(currRange.Start)
+			isMultipleOfBlockSize = currRangeDuration%blockSize == 0
+		)
+
+		if !isMultipleOfBlockSize {
+			return nil, fmt.Errorf(
+				"received bootstrap range that is not multiple of blockSize, blockSize: %d, start: %d, end: %d",
+				blockSize, currRange.End.Unix(), currRange.Start.Unix(),
```
robskillington: Perhaps best to log dates as string like we do elsewhere?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,515 @@ func (s *commitLogSource) ReadData(
...
+		},
+		FileSetType: persist.FileSetSnapshotType,
+	})
+	if err != nil {
+		return shardResult, err
+	}
+
+	s.log.Infof(
+		"reading snapshot for shard: %d and blockStart: %d and volume: %d",
+		shard, blockStart.Unix(), mostRecentCompleteSnapshot.ID.VolumeIndex)
```
robskillington: Perhaps best to log this as string like we do elsewhere?  `blockStart.String()`

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,515 @@ func (s *commitLogSource) ReadData(
...
+	mostRecentCompleteSnapshotByBlockShard := s.mostRecentCompleteSnapshotByBlockShard(
+		shardsTimeRanges, blockSize, snapshotFilesByShard, s.opts.CommitLogOptions().FilesystemOptions())
+	for block, mostRecentByShard := range mostRecentCompleteSnapshotByBlockShard {
+		for shard, mostRecent := range mostRecentByShard {
+
+			if mostRecent.CachedSnapshotTime.IsZero() {
+				// Should never happen.
+				return nil, nil, fmt.Errorf(
+					"%s shard: %d and block: %d had zero value for most recent snapshot time",
+					instrument.InvariantViolatedMetricName, shard, block.ToTime().Unix())
```
robskillington: Perhaps best to log this as string like we do elsewhere?  `block.ToTime().String()`

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,515 @@ func (s *commitLogSource) ReadData(
...
+			}
+
+			s.log.Infof(
+				"most recent snapshot for block: %d and shard: %d is %d",
+				block.ToTime().Unix(), shard, mostRecent.CachedSnapshotTime.Unix())
```
robskillington: Perhaps best to log these dates as string like we do elsewhere?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,515 @@ func (s *commitLogSource) ReadData(
...
+	// log we need to read. The new data structure we're trying to generate looks like:
+	// 		map[blockStart]minimumMostRecentSnapshotTime (across all shards)
+	// This structure is important because it tells us how much of the commit log we need to read for
+	// each block that we're trying to bootstrap (because the commit log is shared across all shards.)
+	minimumMostRecentSnapshotTimeByBlock := s.minimumMostRecentSnapshotTimeByBlock(
+		shardsTimeRanges, blockSize, mostRecentCompleteSnapshotByBlockShard)
+	for block, minSnapshotTime := range minimumMostRecentSnapshotTimeByBlock {
+		s.log.Infof(
+			"min snapshot time for block: %d is: %d",
+			block.ToTime().Unix(), minSnapshotTime.Unix())
```
robskillington: Perhaps best to log these dates as string like we do elsewhere?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,515 @@ func (s *commitLogSource) ReadData(
...
+			commitLogEntryRange := xtime.Range{
+				Start: fileStart,
+				End:   fileStart.Add(fileBlockSize),
+			}
+
+			if commitLogEntryRange.Overlaps(rangeToCheck) {
+				s.log.
+					Infof(
+						"opting to read commit log: %s with start: %d and duration: %s",
+						fileName, fileStart.Unix(), fileBlockSize.String())
```
robskillington: Perhaps best to log this as string like we do elsewhere?  `fileStart.String()`

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -178,20 +301,515 @@ func (s *commitLogSource) ReadData(
...
+				return true
+			}
+		}
+
+		s.log.
+			Infof(
+				"opting to skip commit log: %s with start: %d and duration: %s",
+				fileName, fileStart.Unix(), fileBlockSize.String())
```
robskillington: Perhaps best to log this as string like we do elsewhere?  `fileStart.String()`

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -336,44 +946,81 @@ func (s *commitLogSource) shouldIncludeInIndex(
...
+			bootstrapResultLock.Unlock()
+			return nil, err
+		}
+
 		if unmergedShard.series == nil {
-			continue
+			if !snapshotData.IsEmpty() {
+				// No snapshot or commit log data, skip.
+				// TODO: When we start allowing peer bootstrapping after commit log bootstrapping,
+				// then we will need to change this to mark the shard time ranges as unfulfilled.
```
robskillington: Hm any reason not to do this now?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -336,44 +946,81 @@ func (s *commitLogSource) shouldIncludeInIndex(
...
+				shardsTimeRanges[uint32(shard)],
+			)
+			bootstrapResultLock.Unlock()
+			return nil, err
+		}
+
 		if unmergedShard.series == nil {
-			continue
+			if !snapshotData.IsEmpty() {
+				// No snapshot or commit log data, skip.
```
robskillington: Shouldn't this be `if snapshotData.IsEmpty() { ... }` then?  If this really a bug maybe we should add a test case too.

src/dbnode/storage/index/convert/convert.go
```diff
@@ -173,6 +173,55 @@ func FromMetricIterNoClone(id ident.ID, tags ident.TagIterator) (doc.Document, e
 	}, nil
 }
 
+// TagsFromTagsIter returns an ident.Tags from a TagIterator. It also tries
+// to re-use bytes from the seriesID if they're also present in the tags
+// instead of re-allocating them. This requires that the ident.Tags that is
+// returned will have the same (or shorter) life time as the seriesID,
+// otherwise the operation is unsafe.
+func TagsFromTagsIter(
```
robskillington: Nw, good stuff.

src/dbnode/storage/index/convert/convert.go
```diff
@@ -173,6 +173,55 @@ func FromMetricIterNoClone(id ident.ID, tags ident.TagIterator) (doc.Document, e
...
+			tag.NoFinalize() // Taken ref, cannot finalize this
+		}
+
+		tags.Append(tag)
+	}
+
+	if err := iter.Err(); err != nil {
+		return ident.Tags{}, err
+	}
+	return tags, nil
```
robskillington: Can we close the iter too before finishing?  When we call this method we usually `.Duplicate()` and we need to close the duplicate.

src/dbnode/storage/shard.go
```diff
@@ -967,41 +966,13 @@ func (s *dbShard) newShardEntry(
...
-				tag.Value = clonedID[idx : idx+len(valueBytes)]
-			} else {
-				tag.Value = s.identifierPool.Clone(t.Value)
-			}
-
-			clonedTags.Append(tag)
-		}
-		err := tags.Err()
+		clonedTags, err = convert.TagsFromTagsIter(
+			clonedID, tags.Duplicate(), s.identifierPool)
```
robskillington: Hm, we may want to close both `tags` and the duplicated tags.  i.e.

src/dbnode/integration/generate/writer.go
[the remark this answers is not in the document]
richardartoul: #805

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/shard.go
richardartoul: discussed and handled in person

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: I think this was an unclear comment, its more like just in general we'll need to be smarter about  whether or not the existence or absence of snapshot files constitutes success (with no data) or unfulfilled. I think we discussed this briefly and we were talking about using the shard initialization state. That change alone will probably warrant a one page design doc

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/shard.go
```diff
@@ -967,23 +967,27 @@ func (s *dbShard) newShardEntry(
...
+		tags ident.Tags
+		err  error
 	)
-	if tags.Remaining() > 0 {
-		clonedTags, err = convert.TagsFromTagsIter(
-			clonedID, tags.Duplicate(), s.identifierPool)
-		tags.Close()
+
+	if tagsIter.Remaining() > 0 {
+		dupTagsIter := tagsIter.Duplicate()
```
robskillington: You need to duplicate before doing the `Remaining()` check because it may be at the end of its iterating and have `Remaining() == 0` in that case (before duplication)

PR #758, opened 2018-06-15
Add /database/create cluster type

justinjc:
We need to have an easier way to set up the database that's not just localhost. This `cluster` type takes in a list of either:
1. hostnames
2. hostname with their isolation group


src/coordinator/api/v1/handler/database/create.go
```diff
@@ -54,11 +54,13 @@ const (
 
 	idealDatapointsPerBlock           = 720
 	blockSizeFromExpectedSeriesScalar = idealDatapointsPerBlock * int64(time.Hour)
+	shardMultiplier                   = 32
 
 	dbTypeLocal                 dbType = "local"
+	dbTypeGroup                 dbType = "group"
```
prateek: maybe just `dbTypeCluster`?

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -54,11 +54,13 @@ const (
 
 	idealDatapointsPerBlock           = 720
 	blockSizeFromExpectedSeriesScalar = idealDatapointsPerBlock * int64(time.Hour)
+	shardMultiplier                   = 32
```
prateek: I like where you're going with this. In production we've tried to stay around ~200-300 shards per host - that number obvi depends on profile of the h/w and workload. We could shoot for ~200 as the OSS default for now.

The math would look something like:
```
totalNumShards := 64 * numHosts * replicationFactor
closestPowerOfTwo := math.Pow(2, 1+math.Floor(log(totalNumShards)/log(2)))
numShardsPerHost := closestPowerOfTwo / numHosts
```

What do you think?

src/coordinator/api/v1/handler/database/create.go
justinjc: Sure that works. Only caveat is that this endpoint puts everything into the same zone _and_ isolation group. Not sure if we'd want a separate type so that users can specify different racks, etc.

prateek: Hm good catch, I see what you mean. Maybe take two slices for hostnames - one for `[]string` and the other for `[]struct{string(name), string(isolation group)}`. Only one can be defined at a time, and that way users can use the same endpoint for both use-cases. What do you reckon?

src/coordinator/api/v1/handler/database/create.go
prateek: Yep so in my formulas above `numberOfShardsInPlacement * replicationFactor = totalNumShards`.

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -235,6 +251,39 @@ func defaultedPlacementInitRequest(r *admin.DatabaseCreateRequest, dbCfg dbconfi
 				Port:           uint32(port),
 			},
 		}
+	case dbTypeCluster:
+		// This function assumes only one of Hostnames/HostnameGroups have been set
+		numHosts := len(r.Hostnames) + len(r.HostnameGroups)
+		numShards = int32(math.Min(math.MaxInt32, powerOfTwoAtLeast(float64(numHosts*shardMultiplier))))
+		replicationFactor = r.ReplicationFactor
+		if replicationFactor == 0 {
+			replicationFactor = 1
```
prateek: maybe 3 would be a better default?

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -235,6 +251,39 @@ func defaultedPlacementInitRequest(r *admin.DatabaseCreateRequest, dbCfg dbconfi
...
+		}
+		instances = make([]*placementpb.Instance, 0, numHosts)
+
+		for _, hostname := range r.Hostnames {
+			instances = append(instances, &placementpb.Instance{
+				Id:             hostname,
+				IsolationGroup: "local",
```
prateek: hm maybe something else to differentiate from local mode?

PR #759, opened 2018-06-16
Address data race in block.Stream()

prateek:


robskillington:
Ouch.  Good find, want to add a test?

src/dbnode/storage/block/block.go
```diff
@@ -223,6 +231,15 @@ func (b *dbBlock) Stream(blocker context.Context) (xio.BlockReader, error) {
 		return stream, nil
 	}
 
+	b.RUnlock()
+	lockUpgraded = true
```
robskillington: Btw, maybe we should rename the lowercase `stream(...)` to `streamWithRLock(...)`?

prateek: sure thing

src/dbnode/storage/block/block.go
```diff
@@ -223,6 +231,15 @@ func (b *dbBlock) Stream(blocker context.Context) (xio.BlockReader, error) {
...
+	b.Lock()
+
+	// NB: need to re-check as we just upgraded the lock.
+	if b.mergeTarget == nil {
```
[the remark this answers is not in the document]
prateek: thanks, good catch! added in the next patch

PR #760, opened 2018-06-16
[wip] Use retool & docker for codegen

prateek:


src/dbnode/generated/thrift/generate.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# paranoia, ftw
+set -e
+
+# ensure docker is running
+docker run --rm hello-world >/dev/null
+
+# generate files using dockerized thrift-gen
+THRIFT_IMAGE_VERSION=${THRIFT_IMAGE_VERSION:-"quay.io/prateekrungta/thrift-gen:0.1.0"}
```
justinjc: Is there a better place to put/source this image?

src/dbnode/generated/thrift/generate.go
```diff
@@ -18,6 +18,6 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-//go:generate sh -c "thrift-gen --generateThrift --inputFile $GOPATH/src/$PACKAGE/src/dbnode/generated/thrift/rpc.thrift -outputDir $GOPATH/src/$PACKAGE/src/dbnode/generated/thrift/"
+//go:generate sh -c "$GOPATH/src/$PACKAGE/src/dbnode/generated/thrift/generate.sh"
```
justinjc: Nice, good seeing this getting consolidated.

src/dbnode/generated/thrift/generate.sh
prateek: The source for this image is in the diff (`scripts/thriftgen.Dockerfile`) but I don't want to re-build it each time during execution. 

I'll open a followup to push the image to `quay.io/m3db/thriftgen` instead.

prateek: #762

PR #767, opened 2018-06-18
Common targets for different subdirs

prateek:


scripts/process-cover.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
+set -x
```
justinjc: nit: do we really want `set -x`?

prateek: fair, i'll undo.

PR #768, opened 2018-06-20
Read endpoint to parse prometheus query

nikunjgit:
Need to add tests and also flesh out result node


src/cmd/services/m3coordinator/handler/prometheus/native/common.go
```diff
@@ -0,0 +1,29 @@
...
+// ParseRequestParams parses the input request parameters and provides useful defaults
+func ParseRequestParams(r *http.Request) (RequestParams, error) {
+	param := RequestParams{}
+	timeout, err := prometheus.ParseRequestTimeout(r)
+	if err != nil {
+		return param, err
+	}
+
+	param.Timeout = timeout
+
```
arnikola: nit newlines

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -128,20 +131,104 @@ func (h *PromReadHandler) parseRequest(r *http.Request) (string, *handler.ParseE
...
+
+	return seriesList, nil
+}
+
+func blockSeriesListToSeries(series []block.Series) (*ts.Series, error) {
+	if len(series) == 0 {
+		return &ts.Series{}, nil
+	}
+
+	firstSeries := series[0]
```
arnikola: nit move this down closer to when you use it

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -128,20 +131,104 @@ func (h *PromReadHandler) parseRequest(r *http.Request) (string, *handler.ParseE
...
+	var totalDatapoints int
+	for _, s := range series {
+		totalDatapoints += s.Len()
+	}
+
+	values := ts.NewFixedStepValues(firstSeries.Bounds.StepSize, totalDatapoints, math.NaN(), firstSeries.Bounds.Start)
+	valIdx := 0
+	for _, s := range series {
+		for idx := 0 ; idx < s.Len(); idx++ {
```
arnikola: `i` for index?

src/coordinator/api/v1/handler/prometheus/native/read.go
arnikola: also firstBlock / blocks is a better name?

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -128,20 +131,104 @@ func (h *PromReadHandler) parseRequest(r *http.Request) (string, *handler.ParseE
...
+
+		blocks := result.Result.Blocks()
+		// TODO: Stream blocks to client
+		for blk := range blocks {
+			iter := blk.SeriesIter()
+			for iter.Next() {
+				insertSeriesInMap(iter.Current(), seriesMap)
+			}
+
+			blk.Close()
```
arnikola: Don't close here; it will free all the underlying bytes in the block causing empty data. Caller will have to close it after marshalling to json

src/coordinator/api/v1/handler/prometheus/common.go
```diff
@@ -67,24 +62,21 @@ func ParsePromCompressedRequest(r *http.Request) ([]byte, *handler.ParseError) {
 	return reqBuf, nil
 }
 
-// ParseRequestParams parses the input request parameters and provides useful defaults
-func ParseRequestParams(r *http.Request) (*RequestParams, error) {
-	var params RequestParams
+// ParseRequestTimeout parses the input request timeout with a default
```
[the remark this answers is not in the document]
nikunjgit: I tried making it generic with parseDuration but this is unique since it's coming from headers. Most other things come from form data

src/coordinator/api/v1/handler/prometheus/native/read.go
nikunjgit: hmm need to figure out what to do here. The caller doesn't have reference to block. Maybe this function can marshal to json.

src/coordinator/api/v1/handler/prometheus/common.go
```diff
@@ -67,24 +62,21 @@ func ParsePromCompressedRequest(r *http.Request) ([]byte, *handler.ParseError) {
...
+	if timeout == "" {
+		return defaultTimeout, nil
+	}
 
-		if duration > maxTimeout {
-			return nil, fmt.Errorf("%s: invalid 'timeout': greater than %v", handler.ErrInvalidParams, maxTimeout)
-		}
+	duration, err := time.ParseDuration(timeout)
+	if err != nil {
+		return 0, fmt.Errorf("%s: invalid 'timeout': %v", handler.ErrInvalidParams, err)
```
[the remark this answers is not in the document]
nikunjgit: i think its a bit cleaner since otherwise someone who really relies on the timeout wouldn't know.

src/coordinator/api/v1/handler/prometheus/native/common.go
```diff
@@ -0,0 +1,112 @@
...
+
+	t, err := prometheus.ParseRequestTimeout(r)
+	if err != nil {
+		return params, handler.NewParseError(err, http.StatusBadRequest)
+	}
+
+	params.Timeout = t
+	start, err := parseTime(r, startParam)
+	if err != nil {
+		return params, handler.NewParseError(fmt.Errorf(formatErrStr, startParam, err), http.StatusBadRequest)
```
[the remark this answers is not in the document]
nikunjgit: yeah i was thinking about it. Callers are supposed to check for errors first, and this made unit testing easier :) Note that in other case we would still have to send an empty struct with 0 values. So anyone reading the struct will still have a bug

src/coordinator/api/v1/handler/prometheus/native/common.go
```diff
@@ -0,0 +1,112 @@
...
+	params.End = end
+	step, err := parseDuration(r, stepParam)
+	if err != nil {
+		return params, handler.NewParseError(fmt.Errorf(formatErrStr, stepParam, err), http.StatusBadRequest)
+	}
+
+	params.Step = step
+	target, pErr := parseTarget(r)
+	if pErr != nil {
+		return params, pErr
```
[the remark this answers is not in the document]
nikunjgit: parseTarget was returning parse error itself. Change that now.

src/coordinator/api/v1/handler/prometheus/native/common_test.go
```diff
@@ -0,0 +1,80 @@
...
+	_, err := ParseParams(req)
+	require.NotNil(t, err, "unable to parse request")
+	require.Equal(t, err.Code(), http.StatusBadRequest)
+}
+
+func TestInvalidTarget(t *testing.T) {
+	req, _ := http.NewRequest("GET", PromReadURL, nil)
+	vals := defaultParams()
+	vals.Del(targetQuery)
+	fmt.Println(vals)
```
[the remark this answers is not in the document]
nikunjgit: oops

src/coordinator/models/params.go
```diff
@@ -0,0 +1,36 @@
...
+// THE SOFTWARE.
+
+package models
+
+import (
+	"time"
+)
+
+// RequestParams represents the params from the request
+type RequestParams struct {
```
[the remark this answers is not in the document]
nikunjgit: yeah we pass a copy, not a pointer, so should be immutable across functions.

src/coordinator/storage/block.go
```diff
@@ -108,7 +108,7 @@ func (m *multiSeriesBlockStepIter) Next() bool {
 func (m *multiSeriesBlockStepIter) Current() block.Step {
 	values := make([]float64, len(m.block.seriesList))
 	for i, s := range m.block.seriesList {
```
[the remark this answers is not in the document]
nikunjgit: only used once ?

src/coordinator/executor/transform/interface.go
```diff
@@ -29,10 +29,19 @@ import (
...
+	TimeSpec TimeSpec
 }
 
 // OpNode represents the execution node
 type OpNode interface {
 	Process(ID parser.NodeID, block block.Block) error
 }
+
+// TimeSpec defines the window for the query
+type TimeSpec struct {
```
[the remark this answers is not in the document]
nikunjgit: yeah we pass a copy, not a pointer, so should be immutable across functions.

src/coordinator/block/types.go
```diff
@@ -43,7 +43,9 @@ type Block interface {
 
 // SeriesMeta is metadata data for the series
 type SeriesMeta struct {
-	Tags models.Tags
+	Tags   models.Tags
+	Name   string
```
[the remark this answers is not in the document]
nikunjgit: same, no pointers, just passing by copy

src/coordinator/block/column.go
```diff
@@ -160,3 +162,31 @@ func (cb ColumnBlockBuilder) Build() Block {
...
+	seriesMeta []SeriesMeta
+}
+
+func newColumnBlockSeriesIter(columns []column, blockMeta Metadata, seriesMeta []SeriesMeta) SeriesIter {
+	return &columnBlockSeriesIter{columns: columns, blockMeta: blockMeta, seriesMeta: seriesMeta, idx: -1}
+}
+
+func (m *columnBlockSeriesIter) Next() bool {
+	m.idx++
+	return m.idx < len(m.columns[0].Values)
```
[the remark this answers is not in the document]
nikunjgit: we shouldn't panic on Next(). I added a length check. I think the creator of column block should ensure the same length. Will add some comments.

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -110,38 +97,104 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		seriesMap[seriesID][0] = blockSeries
+		return
 	}
 
-	// NB(braskin): currently, we only support one query at a time
-	if len(targetQueries) > 1 {
-		return "", handler.NewParseError(errBatchQuery, http.StatusBadRequest)
+	// Insert sorted by start time
+	for idx, s := range blockList {
+		if blockSeries.Meta.Bounds.Start.Before(s.Meta.Bounds.Start) {
```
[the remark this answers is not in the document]
nikunjgit: Added TODO, will switch to binary once our blocklist gets big. For now, I'm not sure if we would even want to do it this way in long term. Maybe we just send this back to client in chunks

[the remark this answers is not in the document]
nikunjgit: i'm trying not to add things to context apart from request tracing and deadlines. Will work on a better way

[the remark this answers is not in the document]
arnikola: The nice thing about having it on the context is that it makes keeping track of resources really easy since we can just `defer ctx.Close()` in the handlers to clear everything safely

src/coordinator/storage/block.go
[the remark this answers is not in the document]
arnikola: I'd prefer extracting the variable just in case we end up having a `.Values()` function that has logic behind it, to avoid redoing it every loop

src/coordinator/executor/transform/interface.go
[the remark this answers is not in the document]
benraskin92: Can you explain this one to me?

src/coordinator/api/v1/handler/prometheus/native/common.go
```diff
@@ -115,3 +118,42 @@ func parseTarget(r *http.Request) (string, error) {
...
+	jw := encoding.NewWriter(w)
+	jw.BeginArray()
+	for _, s := range series {
+		jw.BeginObject()
+		jw.BeginObjectField("target")
+		jw.WriteString(s.Name())
+
+		jw.BeginObjectField("tags")
+		jw.BeginObject()
+
```
arnikola: nit: might be a bit better if we keep the write spacing together by object, i.e. 
```
		...

		jw.BeginObjectField("tags")
		jw.BeginObject()
		for k, v := range s.Tags {
			jw.BeginObjectField(k)
			jw.WriteString(v)
		}
		jw.EndObject()

		...
```

src/coordinator/api/v1/handler/prometheus/native/common.go
```diff
@@ -115,3 +118,42 @@ func parseTarget(r *http.Request) (string, error) {
...
+			jw.BeginObjectField(k)
+			jw.WriteString(v)
+		}
+
+		jw.EndObject()
+
+		jw.BeginObjectField("datapoints")
+		jw.BeginArray()
+		for i := 0; i < s.Len(); i++ {
+			dp := s.Values().DatapointAt(i)
```
arnikola: Pull out s.Values() into a variable just in case it ever performs computation

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
 
 	// Results is closed by execute
 	results := make(chan executor.Query)
-	// Block series slices are sorted by start time
-	seriesMap := make(map[string][]block.Series)
 	go h.engine.ExecuteExpr(ctx, parser, opts, params, results)
 
+	// Block slices are sorted by start time
+	// TODO: Pooling
+	sortedBlockList := make([]block.Block, 0, 10)
```
arnikola: nit: extract `10` as a constant

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
 	for result := range results {
 		if result.Err != nil {
-			return nil, err
+			return nil, result.Err
```
arnikola: Would it be worth doing the error checking loop separately, before processing the blocks? Could save time if the last block in your results is erroring

arnikola: Could potentially even do the block result error check preemptively too?

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
+			if err != nil {
+				return nil, err
+			}
 		}
 	}
 
-	return seriesMapToSeriesList(seriesMap)
+	seriesList, err := sortedBlocksToSeriesList(sortedBlockList)
+	for _, b := range sortedBlockList {
+		b.Close()
```
arnikola: Unfortunately I don't think this will close the blocks if any have errored and we shortcircuit the `return`. Might be an idea to capture any `err` in the loop, then break out and close the blocks in that case too

Alternatively, can do something like `defer results.Close()` and add a Close() method on `results` that would propagate to its blocks?

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
-func insertSeriesInMap(blockSeries block.Series, seriesMap map[string][]block.Series) {
-	seriesID := blockSeries.Meta.Name
-	blockList, ok := seriesMap[seriesID]
-	if !ok {
-		seriesMap[seriesID] = make([]block.Series, 1)
-		seriesMap[seriesID][0] = blockSeries
-		return
+func sortedBlocksToSeriesList(blockList []block.Block) ([]*ts.Series, error) {
+	if len(blockList) == 0 {
+		return []*ts.Series{}, nil
```
arnikola: nit: maybe have `var emptySeriesList = []*ts.Series{}` defined as a constant?

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
-			blockList = append(blockList, block.Series{})
-			copy(blockList[idx+1:], blockList[idx:])
-			blockList[idx] = blockSeries
-			seriesMap[seriesID] = blockList
-			return
-		}
+	firstBlock := blockList[0]
+	numSeries := firstBlock.Series()
+	seriesList := make([]*ts.Series, numSeries)
+	seriesIters := make([]block.SeriesIter, len(blockList))
```
arnikola: I don't think `seriesIters` is necessary; we should be able to iterate over blockList in the following loop instead

src/coordinator/block/column.go
```diff
@@ -62,6 +62,15 @@ func (c *columnBlock) SeriesMeta() []SeriesMeta {
 	return c.seriesMeta
 }
 
+// Steps returns the total steps
+func (c *columnBlock) Steps() int {
+	return len(c.columns)
+}
+
+func (c *columnBlock) Series() int {
```
arnikola: nit: missing comment

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
+			blockSeries := iter.Current()
+			for i := 0; i < blockSeries.Len(); i++ {
+				values.SetValueAt(valIdx, blockSeries.ValueAtStep(i))
+				valIdx++
+			}
 		}
 
-		seriesList = append(seriesList, *s)
+		seriesList[i] = ts.NewSeries(seriesMeta[i].Name, values, seriesMeta[i].Tags)
+
```
arnikola: nit: remove newline

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
+		blockList = append(blockList, b)
+		return blockList, nil
 	}
 
-	firstSeries := series[0]
-	var totalDatapoints int
-	for _, s := range series {
-		totalDatapoints += s.Len()
+	firstBlock := blockList[0]
+	if firstBlock.Series() != b.Series() {
```
arnikola: Can we rename `Series` to `SeriesCount` or something similar to give better context?

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
 	}
 
-	values := ts.NewFixedStepValues(firstSeries.Meta.Bounds.StepSize, totalDatapoints, math.NaN(), firstSeries.Meta.Bounds.Start)
-	valIdx := 0
-	for _, s := range series {
-		for idx := 0 ; idx < s.Len(); idx++ {
-			values.SetValueAt(valIdx, s.ValueAtStep(idx))
-			valIdx++
-		}
+	if firstBlock.Steps() != b.Steps() {
```
arnikola: Could we use `StepCount`?

src/coordinator/block/types.go
```diff
@@ -66,6 +62,19 @@ func (b Bounds) TimeForIndex(idx int) (time.Time, error) {
...
+// Steps calculates the number of steps for the bounds
+func (b Bounds) Steps() int {
+	if b.Start.After(b.End) || b.StepSize == 0 {
+		return 0
+	}
+
+	return int(b.End.Sub(b.Start)/b.StepSize) + 1
+}
+
+func (b Bounds) String() string {
```
arnikola: nit: missing comment

src/coordinator/executor/result.go
```diff
@@ -23,43 +23,71 @@ package executor
...
+
+// BlockResult has the result from a block
+type BlockResult struct {
+	Block block.Block
+	Err   error
 }
 
 func newResultNode() *ResultNode {
-	blocks := make(chan block.Block, channelSize)
+	blocks := make(chan BlockResult, channelSize)
```
arnikola: nit: `resultCh[an]` might have more context?

src/coordinator/executor/result.go
```diff
@@ -23,43 +23,71 @@ package executor
...
+	r.blocks <- BlockResult{
+		Block: block,
+	}
+
 	return nil
 }
 
 // Blocks return a channel to stream back blocks to the client
-func (r *ResultNode) Blocks() chan block.Block {
+func (r *ResultNode) Blocks() chan BlockResult {
```
arnikola: nit `BlockResults()` might have more context

src/coordinator/executor/result.go
```diff
@@ -23,43 +23,71 @@ package executor
...
 	return r.blocks
 }
 
 // TODO: Signal error downstream
 func (r *ResultNode) abort(err error) {
+	if r.aborted {
+		return
+	}
+
```
arnikola: This might need to use RWLock

src/coordinator/storage/block.go
```diff
@@ -69,6 +70,19 @@ func (m multiSeriesBlock) Meta() block.Metadata {
 	return m.meta
 }
 
+func (m multiSeriesBlock) Steps() int {
+	// If series has lesser points then it should return NaNs
+	return m.meta.Bounds.Steps()
+}
+
+func (m multiSeriesBlock) Series() int {
+	if len(m.seriesList) == 0 {
```
arnikola: nit:  unnecessary

src/coordinator/storage/block.go
```diff
@@ -144,14 +156,19 @@ func (m *multiSeriesBlockSeriesIter) Next() bool {
 
 func (m *multiSeriesBlockSeriesIter) Current() block.Series {
```
arnikola: Could potentially be better to move this computation to the `Next()` function rather than `Current()`, depending on the usage we expect from this type. If we're calling Current() often, should pre-compute on `Next()` to save unnecessary processing.

src/coordinator/storage/fanout/storage.go
```diff
@@ -117,6 +117,12 @@ func (s *fanoutStorage) Type() storage.Type {
 
 func (s *fanoutStorage) FetchBlocks(
 	ctx context.Context, query *storage.FetchQuery, options *storage.FetchOptions) (block.Result, error) {
+	for _, store := range s.stores {
+		if store.Type() == storage.TypeLocalDC {
+			return store.FetchBlocks(ctx, query, options)
```
arnikola: This may not be correct if we have multiple local storages (I think that may be the case for different resolutions?)

src/coordinator/util/timing.go
```diff
@@ -40,3 +40,8 @@ func ParseTimeString(s string) (time.Time, error) {
 
 	return time.Time{}, fmt.Errorf("invalid timestamp for %s", s)
 }
+
+// DurationToMS converts a duration into milliseconds
```
arnikola: Can potentially use `xtime.ToNormalizedDuration(duration, time.Millisecond)`?

src/coordinator/util/encoding/writer_test.go
```diff
@@ -0,0 +1,176 @@
...
+			w.WriteNull()
+			w.EndObject()
+			w.EndObject()
+			w.EndObject()
+		})
+
+}
+
+func TestWriteErrors(t *testing.T) {
+	// NB(mmihic): All of these error messages could be improved
```
arnikola: Should update user tag

src/coordinator/util/encoding/writer.go
```diff
@@ -0,0 +1,365 @@
...
+// an intermediate object layer.
+type Writer struct {
+	w          *bufio.Writer
+	state      writeState
+	containers []containerType
+	err        error
+}
+
+// NewWriter creates a new JSON token writer
+func NewWriter(w io.Writer) *Writer {
```
arnikola: Could we move this to a /json/ package?

src/coordinator/parser/promql/types.go
```diff
@@ -21,33 +21,46 @@
...
-func NewSelectorFromMatrix(n *promql.MatrixSelector) parser.Params {
-	// TODO: convert n.LabelMatchers to Matchers
-	return functions.FetchOp{Name: n.Name, Offset: n.Offset, Matchers: nil, Range: n.Range}
+func NewSelectorFromMatrix(n *promql.MatrixSelector) (parser.Params, error) {
+	matchers, err := labelMatchersToModelMatcher(n.LabelMatchers)
+	if err != nil {
+		return nil, err
+	}
+
+	return functions.FetchOp{Name: n.Name, Offset: n.Offset, Matchers: matchers, Range: n.Range}, nil
```
arnikola: nit: Split fields across lines

src/coordinator/executor/result.go
```diff
@@ -23,43 +23,71 @@ package executor
...
 	channelSize = 100
 )
 
+var (
+	errAborted = errors.New("the query has been aborted")
+)
+
 // Result provides the execution results
 type Result interface {
 	abort(err error)
```
arnikola: Do we need to define these private methods on an interface? How does that work if you're trying to create a struct to satisfy `Result` outside of this package?

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
+		return nil, fmt.Errorf("mismatch in number of steps for the block, wanted: %d, found: %d", firstBlock.Steps(), b.Steps())
 	}
 
-	return ts.NewSeries(firstSeries.Meta.Name, values, nil), nil
+	index := sort.Search(len(blockList), func(i int) bool { return blockList[i].Meta().Bounds.Start.Before(b.Meta().Bounds.Start) })
+	blockList = append(blockList, b)
```
arnikola: I think we can just do `blockList = blockList[0:len(blockList)+1]` instead of appending `b` here

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
 
-func seriesMapToSeriesList(seriesMap map[string][]block.Series) ([]ts.Series, error) {
-	seriesList := make([]ts.Series, 0, len(seriesMap))
-	for _, blockSeriesList := range seriesMap {
-		s, err := blockSeriesListToSeries(blockSeriesList)
-		if err != nil {
-			return nil, err
+			blockSeries := iter.Current()
+			for i := 0; i < blockSeries.Len(); i++ {
+				values.SetValueAt(valIdx, blockSeries.ValueAtStep(i))
```
arnikola: nit: can use `i` instead of `valIdx`

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,89 +104,94 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
+				valIdx++
+			}
 		}
 
-		seriesList = append(seriesList, *s)
+		seriesList[i] = ts.NewSeries(seriesMeta[i].Name, values, seriesMeta[i].Tags)
+
 	}
```
arnikola: Is it worth checking that iter.Next() is false for each series you have here?

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -70,34 +71,23 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
-		logger.Error("unable to fetch data", zap.Any("error", err))
-		handler.Error(w, err, http.StatusInternalServerError)
-		return
-	}
-
-	resp := &ReadResponse{
-		Results: result,
+	if params.Debug {
+		logger.Info("Request params", zap.Any("params", params))
```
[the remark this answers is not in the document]
nikunjgit: yeah, I'm adding a way to get more context about a query if someone passes debug flag.

src/coordinator/api/v1/handler/prometheus/native/read.go
nikunjgit: not sure what you mean. I have added some comments to make it clear.

src/coordinator/block/column.go
nikunjgit: probably better to just have them on the interface ? All implementations will have the exact same comments repeated

src/coordinator/storage/block.go
nikunjgit: had changed it and forgot to remove this condition

src/coordinator/storage/block.go
nikunjgit: Next() and Current() are called the same number of times (until Next() returns true) so don't think we can save here.

src/coordinator/storage/fanout/storage.go
nikunjgit: just for now until Ben's diff is done and we start supporting remote as well. Will add TODO

src/coordinator/parser/promql/types.go
nikunjgit: why not single line ?

src/coordinator/executor/result.go
nikunjgit: then i'd export them :) Avoiding exporting things until those are needed. But can do it now if you feel that's better.

src/coordinator/api/v1/handler/prometheus/native/read.go
nikunjgit: you need a way to increase the size of the slice

src/coordinator/api/v1/handler/prometheus/native/read.go
nikunjgit: basically we are combining multiple blocks. So valIdx is in the outer loop. I hope my comments make it clear now.

src/coordinator/api/v1/handler/prometheus/native/read.go
nikunjgit: yeah, need to increment their pointers.

nikunjgit: interesting, on insertion we check if they have the same number of steps.

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -44,6 +44,13 @@ const (
 
 	// PromReadHTTPMethod is the HTTP method used with this resource.
 	PromReadHTTPMethod = http.MethodGet
+
+	// TODO: Move to config
```
arnikola: :+1:

src/coordinator/api/v1/handler/prometheus/native/read_test.go
```diff
@@ -87,11 +88,41 @@ func TestPromReadWithFetchAndCount(t *testing.T) {
...
+	mockSession.EXPECT().FetchTagged(gomock.Any(), gomock.Any(), gomock.Any()).Return(seriesiter.NewMockSeriesIters(ctrl, testTags, 1, 10), true, nil)
+
+	promRead := &PromReadHandler{engine: executor.NewEngine(storage)}
+	req, _ := http.NewRequest("GET", PromReadURL, nil)
+	params := defaultParams()
+	params.Set(targetParam, `abs(http_requests_total{job="prometheus",group="canary"})`)
+	req.URL.RawQuery = params.Encode()
+
+	r, parseErr := parseParams(req)
+	require.Nil(t, parseErr)
```
arnikola: nit: `require.NoError(t, parseErr)`

src/coordinator/api/v1/handler/prometheus/native/read_test.go
```diff
@@ -87,11 +88,41 @@ func TestPromReadWithFetchAndCount(t *testing.T) {
...
+	seriesList, err := promRead.read(context.TODO(), httptest.NewRecorder(), r)
+	require.NoError(t, err)
+	require.Len(t, seriesList, 1)
+	s := seriesList[0]
+	assert.Equal(t, s.Values().Len(), 361, "10 second resolution for 1 hour including start time")
```
arnikola: nit: for `.Equal`, expected value goes first, then actual value

src/coordinator/ts/values.go
```diff
@@ -29,10 +29,10 @@ import (
 )
 
 // Values holds the values for a timeseries.  It provides a minimal interface
-// for storing and retrieving values in the series, with Series providing a
+// for storing and retrieving values in the series, with SeriesCount providing a
```
arnikola: nit: Should still be `Series` here

src/coordinator/ts/series.go
```diff
@@ -28,15 +28,15 @@ import (
...
+// A SeriesCount is the public interface to a block of timeseries values.  Each block has a start time,
 // a logical number of steps, and a step size indicating the number of milliseconds represented by each point.
 type Series struct {
 	name string
 	vals Values
 	Tags models.Tags
 }
 
-// NewSeries creates a new Series at a given start time, backed by the provided values
+// NewSeries creates a new SeriesCount at a given start time, backed by the provided values
```
arnikola: nit: Should still be Series here

src/coordinator/ts/series.go
```diff
@@ -28,15 +28,15 @@ import (
 	"github.com/m3db/m3db/src/coordinator/models"
 )
 
-// A Series is the public interface to a block of timeseries values.  Each block has a start time,
+// A SeriesCount is the public interface to a block of timeseries values.  Each block has a start time,
```
arnikola: nit: Should be `Series is...`

src/coordinator/api/v1/handler/prometheus/native/common.go
```diff
@@ -0,0 +1,159 @@
...
+	"github.com/m3db/m3db/src/coordinator/util"
+	"github.com/m3db/m3db/src/coordinator/util/json"
+)
+
+const (
+	endParam    = "end"
+	startParam  = "start"
+	targetParam = "target"
+	stepParam   = "step"
+	debugParam  = "debug"
```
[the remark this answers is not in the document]
nikunjgit: made a first pass, we can modify it in the future

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -71,77 +72,137 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	for i, b := range blockList {
+		seriesIters[i] = b.SeriesIter()
 	}
 
-	if len(targetQueries) == 0 {
-		return "", handler.NewParseError(errNoQueryFound, http.StatusBadRequest)
+	seriesMeta := firstBlock.SeriesMeta()
+	bounds := firstBlock.Meta().Bounds
+	for i := 0; i < numSeries; i++ {
+		values := ts.NewFixedStepValues(bounds.StepSize, firstBlock.StepCount()*len(blockList), math.NaN(), bounds.Start)
```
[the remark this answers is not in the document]
nikunjgit: good catch

src/coordinator/executor/result.go
```diff
@@ -21,19 +21,81 @@
...
 }
+
+// Blocks return a channel to stream back blocks to the client
+func (r *ResultNode) Blocks() chan ResultChan {
+	return r.blocks
+}
+
+// TODO: Signal error downstream
+func (r *ResultNode) abort(err error) {
+	r.mu.Lock()
```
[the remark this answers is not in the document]
nikunjgit: should be fine, don't really expect it to be called much

src/coordinator/functions/abs.go
```diff
@@ -0,0 +1,83 @@
...
+		return err
+	}
+
+	stepIter := b.StepIter()
+	if err := builder.AddCols(b.StepCount()); err != nil {
+		return err
+	}
+
+	for index := 0; stepIter.Next(); index++ {
+		step := stepIter.Current()
```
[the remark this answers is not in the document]
nikunjgit: oh yeah, I just wrote the actual logic for now. That part to wrap transforms will come in later (and probably call this during actual execution)

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -122,29 +122,42 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
-	seriesList, err := sortedBlocksToSeriesList(sortedBlockList)
+	// Ensure that the blocks are closed. Can't do this above since sortedBlockList might change
+	defer func() {
+		for _, b := range sortedBlockList {
+			b.Close()
+		}
+	}()
+
+	if processErr != nil {
+		return nil, err
```
arnikola: Would need to drain channel here

PR #774, opened 2018-06-25
Fix seekers using excess memories by holding onto digest readers after call to Open()

### LINKED ISSUE #476: Make seeker's not hang onto FDDigestReaders
Currently every seeker that we open holds onto about 5 FDDigestReaders. Even when we close/reset them, the underlying buffered reader is not de-allocated.

In our short-lived production workloads this doesn't matter at all, but in clusters with a large number of namespaces/shards/blocks (hundreds of thousands) this adds up to 10's of gigabytes of wasted memory 65KiB size per digest reader)

We should make sure that the buffered readers are properly de-allocated when they are no longer needed.

richardartoul:
Fixes #476

### DIFF
--- a/src/dbnode/persist/fs/seek.go
+++ b/src/dbnode/persist/fs/seek.go
@@ -479,10 +479,3 @@ func (s *seeker) ConcurrentClone() (ConcurrentDataFileSetSeeker, error) {
 		isClone:     true,
 	}, nil
 }
-
-type expectedDigests struct {
-	expectedInfoDigest        uint32
-	expectedIndexDigest       uint32
-	expectedBloomFilterDigest uint32
-	expectedSummariesDigest   uint32
-}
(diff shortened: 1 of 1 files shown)


src/dbnode/persist/fs/seek.go
```diff
@@ -195,26 +177,34 @@ func (s *seeker) Open(namespace ident.ID, shard uint32, blockStart time.Time) er
...
+		digestFdWithDigestContents = digest.NewFdWithDigestContentsReader(s.opts.infoBufferSize)
+	)
 	defer func() {
-		// NB(r): We don't need to keep these FDs open as we use these up front
-		s.infoFdWithDigest.Close()
-		s.indexFdWithDigest.Close()
-		s.bloomFilterFdWithDigest.Close()
-		s.digestFdWithDigestContents.Close()
+		// NB(rartoul): We don't need to keep these FDs open as we use these up front
+		infoFdWithDigest.Close()
```
prateek: Looks like this closes the underlying fd too. Is the `mmap` valid after the fd is closed?

richardartoul: Yeah its safe to close an FD once the mmap syscall has executed, it wont affect the mmap (this is what we were already doing also)

prateek: 👍 sounds good

PR #776, opened 2018-06-25
Pass more specific patterns during `FileSetAt`

prateek:


src/dbnode/persist/fs/fs.go
```diff
@@ -33,9 +33,13 @@ const (
 	segmentFileSetFilePrefix = "segment"
 	fileSuffix               = ".db"
 
+	anyLowerCaseCharsPattern   = "[a-z]*"
+	anyNumbersPattern          = "[0-9]*"
+	anyLowerCaseNumbersPattern = "[a-z0-9]*"
```
richardartoul: super nit: anyLowerCaseCharsNumbersPattern

Not a big deal but I spent a few seconds trying to figure out what a lowercase number would be / why you weren't including chars when I saw the var name

prateek: sure

PR #778, opened 2018-07-02
Add StepIter to blocks

benraskin92:


src/coordinator/ts/m3db/types.go
```diff
@@ -88,19 +88,24 @@ type ConsolidatedSeriesBlock struct {
 	consolidationFunc    ConsolidationFunc // nolint
 }
 
-func (c ConsolidatedSeriesBlock) beyondBounds(multiSeriesBlock MultiSeriesBlock) bool {
-	if c.Metadata.Bounds.Start != multiSeriesBlock.Metadata.Bounds.Start || c.Metadata.Bounds.End != multiSeriesBlock.Metadata.Bounds.End {
+// func (c ConsolidatedSeriesBlock) equalBounds(multiSeriesBlock MultiSeriesBlock) bool {
```
arnikola: Delete this?

src/coordinator/ts/m3db/block_iteration.go
```diff
@@ -145,48 +150,89 @@ func (c *consolidatedSeriesBlockIter) Next() bool {
...
 	}
 
-	for c.seriesIndex < len(c.consolidatesNSBlockSeriesIters) {
-		if c.consolidatesNSBlockSeriesIters[c.seriesIndex].Next() {
+	for c.seriesIndex < len(c.consolidatedNSBlockSeriesIters) {
+		if c.consolidatedNSBlockSeriesIters[c.seriesIndex].Next() {
 			return true
 		}
+		fmt.Println("seriesIndex: ", c.seriesIndex, "indexTime: ", c.indexTime, "bounds: ", c.bounds)
+		// curr, _, _ := c.consolidatedNSBlockSeriesIters[c.seriesIndex].Current()
```
arnikola: Delete?

src/coordinator/ts/m3db/block_iteration.go
```diff
@@ -145,48 +150,89 @@ func (c *consolidatedSeriesBlockIter) Next() bool {
...
 		return false
 	}
 
-	c.currentTime = c.currentTime.Add(c.bounds.StepSize)
-	return !c.currentTime.After(c.bounds.End)
+	for c.indexTime.After(lastDP.Timestamp) && c.next() {
+		lastDP, _, _ = c.consolidatedNSBlockSeriesIters[c.seriesIndex].Current()
+		c.lastDP = lastDP
+	}
+	// if c.indexTime.Add(c.bounds.StepSize).Before(lastDP.Timestamp) {
```
arnikola: Delete?

src/coordinator/ts/m3db/block_iteration.go
```diff
@@ -145,48 +150,89 @@ func (c *consolidatedSeriesBlockIter) Next() bool {
...
+
+	fmt.Println("last dp: ", lastDP)
+
+	return true
+}
+
 func (c *consolidatedNSBlockIter) Next() bool {
-	if len(c.consolidatesNSBlockSeriesIters) == 0 {
+	c.indexTime = c.indexTime.Add(c.bounds.StepSize)
+	fmt.Println("index time: ", c.indexTime)
```
arnikola: Delete println

src/coordinator/ts/m3db/block_iteration.go
```diff
@@ -145,48 +150,89 @@ func (c *consolidatedSeriesBlockIter) Next() bool {
...
+
+	if c.indexTime.After(lastDP.Timestamp) {
+		fmt.Println("AFTER")
+		// c.indexTime = c.indexTime.Add(c.bounds.StepSize)
+		c.lastDP = ts.Datapoint{Value: math.NaN(), Timestamp: c.indexTime}
+		// return true
+	}
+	c.indexTime = c.indexTime.Add(c.bounds.StepSize)
+
+	fmt.Println("last dp: ", lastDP)
```
arnikola: Delete println

src/coordinator/ts/m3db/block_iteration.go
```diff
@@ -145,48 +150,89 @@ func (c *consolidatedSeriesBlockIter) Next() bool {
...
+	for c.indexTime.After(lastDP.Timestamp) && c.next() {
+		lastDP, _, _ = c.consolidatedNSBlockSeriesIters[c.seriesIndex].Current()
+		// c.indexTime = c.indexTime.Add(c.bounds.StepSize)
+		c.lastDP = lastDP
+		// return true
+	}
+
+	if c.indexTime.After(lastDP.Timestamp) {
+		fmt.Println("AFTER")
+		// c.indexTime = c.indexTime.Add(c.bounds.StepSize)
```
arnikola: Delete, other commented lines here too

src/coordinator/ts/m3db/block_iteration.go
```diff
@@ -145,48 +150,89 @@ func (c *consolidatedSeriesBlockIter) Next() bool {
...
 			return false
 		}
-		c.index++
 	}
 
 	return true
 }
 
+func (c *consolidatedNSBlockIter) BadNext() bool {
+	// lastDP, _, _ := c.consolidatedNSBlockSeriesIters[c.seriesIndex].Current()
```
arnikola: Delete

src/coordinator/ts/m3db/types.go
```diff
@@ -88,19 +88,24 @@ type ConsolidatedSeriesBlock struct {
...
+// 	if c.Metadata.Bounds.Start != multiSeriesBlock.Metadata.Bounds.Start || c.Metadata.Bounds.End != multiSeriesBlock.Metadata.Bounds.End {
+// 		return false
+// 	}
+// 	return true
+// }
+
+func equalBounds(boundOne, boundTwo block.Bounds) bool {
```
arnikola: Would this be better to put as an `Equals(other Bounds)` receiver function on block.Bounds?

src/coordinator/ts/m3db/types.go
```diff
@@ -88,19 +88,24 @@ type ConsolidatedSeriesBlock struct {
...
+	if boundOne.Start != boundTwo.Start || boundOne.End != boundTwo.End {
```
arnikola: Use `.Equal()` for time comparisons rather than `==`, otherwise it does some weird comparisons across Location and clock settings which might lead to unexpected results

src/coordinator/ts/m3db/types.go
```diff
@@ -62,18 +62,18 @@ type ConsolidatedNSBlock struct {
 	SeriesIterators encoding.SeriesIterators
 }
 
-func (c ConsolidatedNSBlock) beyondBounds(consolidatedSeriesBlock ConsolidatedSeriesBlock) bool {
-	if c.Bounds.Start != consolidatedSeriesBlock.Metadata.Bounds.Start || c.Bounds.End != consolidatedSeriesBlock.Metadata.Bounds.End {
-		return false
-	}
-	return true
-}
+// func (c ConsolidatedNSBlock) beyondBounds(consolidatedSeriesBlock ConsolidatedSeriesBlock) bool {
```
arnikola: Delete

src/coordinator/ts/m3db/blocks.go
```diff
@@ -43,25 +44,27 @@ func SeriesBlockToMultiSeriesBlocks(multiNamespaceSeriesList []MultiNamespaceSer
...
 		}
 
 		// once we get the length of consolidatedSeriesBlocks, we can create a
 		// MultiSeriesBlocks list with the proper size
-		if multiNamespaceSeriesIdx == 0 {
+		if seriesIdx == 0 {
 			multiSeriesBlocks = make(MultiSeriesBlocks, len(consolidatedSeriesBlocks))
-			commonTags, err = storage.FromIdentTagIteratorToTags(consolidatedSeriesBlocks[0].ConsolidatedNSBlocks[0].SeriesIterators.Iters()[0].Tags())
+			commonTags, err = storage.FromIdentTagIteratorToTags(multiNamespaceSeries[0].Tags)
+			// commonTags, err = storage.FromIdentTagIteratorToTags(consolidatedSeriesBlocks[0].ConsolidatedNSBlocks[0].SeriesIterators.Iters()[0].Tags())
```
arnikola: Delete

src/coordinator/ts/m3db/block_iteration.go
```diff
@@ -145,48 +150,89 @@ func (c *consolidatedSeriesBlockIter) Next() bool {
...
-	lastDP := c.lastDP
+	if !c.indexTime.After(c.lastDP.Timestamp) && c.indexTime.Add(c.bounds.StepSize).After(c.lastDP.Timestamp) {
 
-	for c.currentTime.After(lastDP.Timestamp) && c.next() {
-		lastDP, _, _ = c.consolidatesNSBlockSeriesIters[c.seriesIndex].Current()
-		c.lastDP = lastDP
-	}
-	if c.currentTime.Add(c.bounds.StepSize).Before(lastDP.Timestamp) || c.lastVal {
-		return math.NaN()
+		fmt.Println("last dp current: ", c.lastDP)
```
arnikola: Delete println

src/coordinator/ts/m3db/block_iteration.go
```diff
@@ -145,48 +150,89 @@ func (c *consolidatedSeriesBlockIter) Next() bool {
...
 	}
-	c.lastVal = true
+	// c.lastVal = true
 
 	return false
 }
 
 func (c *consolidatedNSBlockIter) Current() float64 {
-	lastDP := c.lastDP
+	if !c.indexTime.After(c.lastDP.Timestamp) && c.indexTime.Add(c.bounds.StepSize).After(c.lastDP.Timestamp) {
```
arnikola: nit: Might read better if you pull out `c.indexTime` and `c.lastDP.Timestamp` as variables first

src/coordinator/ts/m3db/blocks.go
```diff
@@ -43,25 +44,27 @@ func SeriesBlockToMultiSeriesBlocks(multiNamespaceSeriesList []MultiNamespaceSer
...
+			fmt.Println("tags: ", commonTags)
```
arnikola: Delete println

benraskin92:
I'll get rid of all of the println's and commented out code

src/coordinator/storage/local/storage.go
```diff
@@ -144,18 +161,52 @@ func (s *localStorage) Type() storage.Type {
...
 
-func (s *localStorage) FetchBlocks(
-	ctx context.Context,
-	query *storage.FetchQuery,
-	options *storage.FetchOptions) (block.Result, error) {
-	fetchResult, err := s.Fetch(ctx, query, options)
+func (s *localStorage) fetchRaw(
+	query index.Query,
+	opts index.QueryOptions,
+) (encoding.SeriesIterators, bool, error) { // nolint: unparam
```
arnikola: nit: nolint usually goes above the function declaration

src/coordinator/ts/m3db/block/block_conversion.go
```diff
@@ -31,45 +34,81 @@ var (
...
+			}
 		}
 
 		for consolidatedSeriesBlockIdx, consolidatedSeriesBlock := range consolidatedSeriesBlocks {
 			// we only want to set the start and end times once
-			if multiNamespaceSeriesIdx == 0 {
-				multiSeriesBlocks[consolidatedSeriesBlockIdx].Start = consolidatedSeriesBlock.Start
-				multiSeriesBlocks[consolidatedSeriesBlockIdx].End = consolidatedSeriesBlock.End
+			if seriesIdx == 0 {
+				multiSeriesBlocks[consolidatedSeriesBlockIdx].Metadata.Bounds.StepSize = consolidatedSeriesBlock.Metadata.Bounds.StepSize
```
arnikola: nit: can you make these names shorter? It's quite difficult to read at the moment

Also, can pull out multiSeriesBlocks[consolidatedSeriesBlockIdx].Metadata.Bounds into a var?

src/coordinator/ts/m3db/block/block_iteration.go
```diff
@@ -0,0 +1,194 @@
...
+			return false
+		}
+	}
+
+	return true
+}
+
+func (c *consolidatedNSBlockIter) Next() bool {
+	c.indexTime = c.indexTime.Add(c.bounds.StepSize)
+	lastDP := c.lastDP
```
nikunjgit: initialize it right before you use it

src/coordinator/ts/m3db/block/block_iteration.go
```diff
@@ -0,0 +1,194 @@
...
+
+	for c.indexTime.After(lastDP.Timestamp) && c.next() {
+		lastDP, _, _ = c.consolidatedNSBlockSeriesIters[c.seriesIndex].Current()
+		c.lastDP = lastDP
+	}
+
+	return true
+}
+
+func (c *consolidatedNSBlockIter) next() bool {
```
nikunjgit: call it nextIterator or something ?

src/coordinator/ts/m3db/block/block_iteration.go
```diff
@@ -0,0 +1,194 @@
...
+			return true
+		}
+		c.seriesIndex++
+	}
+
+	return false
+}
+
+func (c *consolidatedNSBlockIter) Current() float64 {
+	if !c.indexTime.After(c.lastDP.Timestamp) && c.indexTime.Add(c.bounds.StepSize).After(c.lastDP.Timestamp) {
```
nikunjgit: extract lastDP ??

nikunjgit: also can you add some comments about what you're checking

src/coordinator/ts/m3db/block/block_iteration_test.go
```diff
@@ -0,0 +1,88 @@
...
+		results = append(results, sliceOfNaNs)
+	}
+	results = append(results, []float64{5, 5})
+	results = append(results, sliceOfNaNs)
+	results = append(results, []float64{6, 6})
+
+	return results
+}
+
+func TestStepIteration(t *testing.T) {
```
nikunjgit: probably need more tests, also should write down conditions you are testing. This test is pretty opaque

src/coordinator/ts/m3db/block/convertor.go
```diff
@@ -31,45 +34,84 @@ var (
...
+func (m MultiSeriesBlocks) commonTags() map[string]string {
+	commonTags := make(map[string]string)
+	for tag, val := range m[0].Blocks[0].Metadata.Tags {
+		commonTags[tag] = val
+	}
+
+	// NB(braskin): all series are the same across MultiSeriesBlocks so only need
+	// to look at one
+	for _, series := range m[0].Blocks {
+		seriesTags := series.Metadata.Tags
```
nikunjgit: skip the first one ?

src/coordinator/ts/m3db/block/convertor.go
```diff
@@ -79,11 +121,12 @@ func newConsolidatedSeriesBlocks(multiNamespaceSeries MultiNamespaceSeries, seri
...
 			if seriesBlocksIdx == 0 {
-				consolidatedSeriesBlocks[consolidatedNSBlockIdx].Start = consolidatedNSBlock.Start
-				consolidatedSeriesBlocks[consolidatedNSBlockIdx].End = consolidatedNSBlock.End
+				consolidatedSeriesBlocks[consolidatedNSBlockIdx].Metadata.Bounds.StepSize = consolidatedNSBlock.Bounds.StepSize
+				consolidatedSeriesBlocks[consolidatedNSBlockIdx].Metadata.Bounds.Start = consolidatedNSBlock.Bounds.Start
+				consolidatedSeriesBlocks[consolidatedNSBlockIdx].Metadata.Bounds.End = consolidatedNSBlock.Bounds.End
 			}
 
-			if !consolidatedNSBlock.beyondBounds(consolidatedSeriesBlocks[consolidatedNSBlockIdx]) {
+			if !equalBounds(consolidatedNSBlock.Bounds, consolidatedSeriesBlocks[consolidatedNSBlockIdx].Metadata.Bounds) {
```
nikunjgit: define Equals method on the bounds ?

src/coordinator/ts/m3db/block/convertor_test.go
```diff
@@ -0,0 +1,178 @@
...
+	require.NoError(t, err)
+
+	seriesOne := SeriesBlocks{
+		Blocks: []SeriesBlock{seriesOneBlockOne, seriesOneBlockTwo},
+		ID:     ident.StringID("test_one"),
+		Tags:   tagsOne,
+	}
+
+	rawTagsTwo := []string{"biz", "baz", "same", "tag"}
+	seriesIteratorsTwo := []encoding.SeriesIterator{
```
nikunjgit: consider using different timestamps for seriesIteratorTwo vs seriesOneIterators

nikunjgit: also maybe call it seriesTwoIterators to be consistent ?

src/coordinator/ts/m3db/block/convertor_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+	assert.Equal(t, models.Tags{"same": "tag"}, m3CoordBlocks[0].Metadata.Tags)
+	assert.Equal(t, models.Tags{"same": "tag"}, m3CoordBlocks[1].Metadata.Tags)
+}
+
+func TestMultipleNamespacesSuccess(t *testing.T) {
+	now := time.Now()
+	ctrl := gomock.NewController(t)
+	seriesOne, seriesTwo := newMultiNamespaceSeries(ctrl, now, t)
+	seriesOne = append(seriesOne, seriesOne...)
```
nikunjgit: ?

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+import (
+	"errors"
+	"math"
+	"time"
+
+	"github.com/m3db/m3db/src/coordinator/block"
+)
+
+// StepIter creates a new step iterator for a given MultiSeriesBlock
+func (m MultiSeriesBlock) StepIter() (block.StepIter, error) {
```
nikunjgit: move structs next to implementation ?

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+			consolidatedNSBlockIters: consolidatedNSBlockIters,
+		}
+	}
+
+	return consolidatedSeriesBlockIters
+}
+
+func newConsolidatedNSBlockIter(nsBlock ConsolidatedNSBlock) *consolidatedNSBlockIter {
+	return &consolidatedNSBlockIter{
+		consolidatedNSBlockSeriesIters: nsBlock.SeriesIterators.Iters(),
```
nikunjgit: call it m3dbIters ?

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+		bounds:    nsBlock.Bounds,
+		indexTime: nsBlock.Bounds.Start.Add(-1 * nsBlock.Bounds.StepSize),
```
nikunjgit: just use index  instead of time? initialize it to -1

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+// Close closes the underlaying iterators
+func (c *consolidatedSeriesBlockIter) Close() {
+	// todo(braskin): implement this function
+}
+
+// Next moves to the next item
+func (c *consolidatedNSBlockIter) Next() bool {
+	c.indexTime = c.indexTime.Add(c.bounds.StepSize)
```
nikunjgit: use index, and then do c.index++, 

indexTime := c.bounds.TimeForIndex(c.index)

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+
+	if !c.indexTime.Before(c.bounds.End) {
```
nikunjgit: need to figure out if end is inclusive.

nikunjgit: also add a comment

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+		return false
+	}
+
+	lastDP := c.lastDP
+	for c.indexTime.After(lastDP.Timestamp) && c.nextIterator() {
```
nikunjgit: add a comment

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+func newConsolidatedSeriesBlockIters(blocks ConsolidatedSeriesBlocks) []block.ValueIterator {
+	consolidatedSeriesBlockIters := make([]block.ValueIterator, len(blocks))
+	if len(blocks) == 0 {
+		return consolidatedSeriesBlockIters
+	}
+
+	for i, seriesBlock := range blocks {
+		consolidatedNSBlockIters := make([]block.ValueIterator, len(blocks[0].ConsolidatedNSBlocks))
```
nikunjgit: move out len(blocks[0].ConsolidatedNSBlocks)

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+// Meta returns the metadata for the step iter
+func (m *multiSeriesBlockStepIter) Meta() block.Metadata {
+	return m.meta
+}
+
+// SeriesMeta returns metadata for the individual timeseries
+func (m *multiSeriesBlockStepIter) SeriesMeta() []block.SeriesMeta {
+	metas := make([]block.SeriesMeta, len(m.blocks))
+	for i, s := range m.blocks {
+		// todo(braskin): add name
```
nikunjgit: metas[i].Name = s.Metadata.Tags[models.MetricName] ?

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+	return metas
+}
+
+// StepCount returns the total steps/columns
+func (m *multiSeriesBlockStepIter) StepCount() int {
+	if len(m.blocks) == 0 {
+		return 0
+	}
+	return m.blocks[0].Metadata.Bounds.Steps()
```
nikunjgit: FYI: this is inclusive of the end

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+
+// Current returns the slice of vals and timestamps for that step
+func (m *multiSeriesBlockStepIter) Current() (block.Step, error) {
+	values := make([]float64, len(m.seriesIters))
+	for i, s := range m.seriesIters {
+		values[i] = s.Current()
+	}
+
+	bounds := m.meta.Bounds
+	t := bounds.Start.Add(time.Duration(m.index) * bounds.StepSize)
```
nikunjgit: use bounds.TimeForIndex(m.index)

src/coordinator/ts/m3db/block/convertor_test.go
```diff
@@ -0,0 +1,178 @@
...
+	assert.Equal(t, "test_two", m3CoordBlocks[1].Blocks[1].ConsolidatedNSBlocks[0].SeriesIterators.Iters()[0].ID().String())
+	assert.Equal(t, now.Add(10*time.Minute), m3CoordBlocks[1].Metadata.Bounds.Start)
+	assert.Equal(t, now.Add(20*time.Minute), m3CoordBlocks[1].Metadata.Bounds.End)
+
+	assert.Equal(t, models.Tags{"same": "tag"}, m3CoordBlocks[0].Metadata.Tags)
+	assert.Equal(t, models.Tags{"same": "tag"}, m3CoordBlocks[1].Metadata.Tags)
+}
+
+func TestMultipleNamespacesSuccess(t *testing.T) {
+	now := time.Now()
```
arnikola: This is the same test as the previous one bar the length check. May be better to refactor the test to a common function, then do the length check at the end?

src/coordinator/ts/m3db/block/iterator.go
arnikola: We should probably have  methods on `block` that will return things like Metadata and Steps so we don't have to do this everywhere... though this is a bigger discussion than for here )

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+	lastDP := c.lastDP
+	for c.indexTime.After(lastDP.Timestamp) && c.nextIterator() {
+		lastDP, _, _ = c.consolidatedNSBlockSeriesIters[c.seriesIndex].Current()
+		c.lastDP = lastDP
+	}
+
+	return true
+}
+
+func (c *consolidatedNSBlockIter) nextIterator() bool {
```
arnikola: Maybe useful to add a small comment about this here?

src/coordinator/ts/m3db/block/iterator.go
```diff
@@ -0,0 +1,214 @@
...
+
+	return false
+}
+
+// Current returns the float64 value for that step
+func (c *consolidatedNSBlockIter) Current() float64 {
+	lastDP := c.lastDP
+	// NB(braskin): if the last datapoint is after the current step, but before the (current step+1),
+	// return that datapoint, otherwise return NaN
+	if !c.indexTime.After(lastDP.Timestamp) && c.indexTime.Add(c.bounds.StepSize).After(lastDP.Timestamp) {
```
arnikola: nit: pull out lastDp.Timestamp and c.indexTime as variables?

src/coordinator/ts/m3db/block/iterator_test.go
```diff
@@ -0,0 +1,336 @@
...
+)
+
+var (
+	now = time.Now()
+	nan = math.NaN()
+)
+
+func createDatapoints(t *testing.T, timeInSeconds []int, vals []float64, now time.Time) []ts.Datapoint {
+	if len(timeInSeconds) != len(vals) {
+		t.Fatal("length of values must equal length of timestamps")
```
arnikola: nit: would prefer to have `require.Equal(t, len(timeInSeconds), len(vals))` to avoid panics

src/coordinator/ts/m3db/block/iterator_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+func createNSBlockIter(iters []encoding.SeriesIterator, start, end time.Time, stepSize time.Duration) consolidatedNSBlockIter {
+	return consolidatedNSBlockIter{
+		consolidatedNSBlockSeriesIters: iters,
+		bounds: block.Bounds{
+			Start:    start,
+			End:      end,
+			StepSize: stepSize,
+		},
+		indexTime: start.Add(-1 * stepSize),
```
arnikola: nit `start.Sub(stepSize)`

src/coordinator/ts/m3db/block/iterator_test.go
```diff
@@ -0,0 +1,336 @@
...
+
+func TestConsolidatedNSBlockIter(t *testing.T) {
+	testCases := []testCase{
+		{
+			dps: [][]ts.Datapoint{
+				createDatapoints(t, []int{0, 90, 100, 300}, []float64{1, 2, 2.5, 3}, now),
+			},
+			start:           now,
+			end:             now.Add(600 * time.Second),
+			stepSize:        60 * time.Second,
```
arnikola: nit `time.Minute`?

src/coordinator/ts/m3db/block/iterator_test.go
```diff
@@ -0,0 +1,336 @@
...
+		},
+		indexTime: start.Add(-1 * stepSize),
+	}
+}
+
+type testCase struct {
+	start, end                     time.Time
+	stepSize                       time.Duration
+	dps                            [][]ts.Datapoint
+	expectedResults, actualResults []float64
```
arnikola: nit: would prefer to have actualResults be a variable built in the test, rather than added to the struct itself

src/coordinator/ts/m3db/block/iterator_test.go
```diff
@@ -0,0 +1,336 @@
...
+func createMultiSeriesBlockStepIter(seriesBlockIters []block.ValueIterator) multiSeriesBlockStepIter {
+	return multiSeriesBlockStepIter{
+		seriesIters: seriesBlockIters,
+	}
+}
+
+type multiSeriesBlockStepIterTestCase struct {
+	dps             [][]float64
+	expectedResults [][]float64
+	actualResults   [][]float64
```
arnikola: Same here re actualResults

src/coordinator/ts/m3db/block/iterator_test.go
```diff
@@ -0,0 +1,336 @@
...
+type mockBlockStepIter struct {
+	seriesIters []block.ValueIterator
+	index       int
+	meta        block.Metadata
+	blocks      ConsolidatedSeriesBlocks
+}
+
+type mockBlockStepIterTestCase struct {
+	dps                            [][]float64
+	actualResults, expectedResults [][]float64
```
arnikola: Same here re: actualResults

src/coordinator/ts/m3db/block/convertor_test.go
benraskin92: Oh I wanted to simulate two namespaces. If you look at the test below that checks the len of the consolidatesNSBlocks, you'll see what I'm doing.

src/coordinator/ts/m3db/block/multi_series.go
```diff
@@ -0,0 +1,155 @@
...
+
+// Current returns the slice of vals and timestamps for that step
+func (m *multiSeriesBlockStepIter) Current() (block.Step, error) {
+	values := make([]float64, len(m.seriesIters))
+	for i, s := range m.seriesIters {
+		values[i] = s.Current()
+	}
+
+	bounds := m.meta.Bounds
+	t, err := bounds.TimeForIndex(m.index)
```
nikunjgit: do this before creating values ?

src/coordinator/ts/m3db/block/ns_series.go
```diff
@@ -0,0 +1,106 @@
...
+		c.seriesIndex++
+	}
+
+	return false
+}
+
+// Current returns the float64 value for that step
+func (c *consolidatedNSBlockIter) Current() float64 {
+	lastDP := c.lastDP
+	// NB(braskin): if the last datapoint is after the current step, but before the (current step+1),
```
nikunjgit: move comment down ?

src/coordinator/ts/m3db/block/consolidated_series.go
```diff
@@ -0,0 +1,76 @@
...
+package block
+
+import "github.com/m3db/m3db/src/coordinator/block"
+
+// ConsolidatedSeriesBlock is a single series consolidated across different namespaces
+// for a single block
+type ConsolidatedSeriesBlock struct {
+	Metadata             block.Metadata
+	ConsolidatedNSBlocks []ConsolidatedNSBlock
+	consolidationFunc    ConsolidationFunc // nolint
```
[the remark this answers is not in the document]
benraskin92: Even if it's just for one field in the struct?

src/coordinator/ts/m3db/block/iterator_test.go
```diff
@@ -96,7 +96,7 @@ func TestConsolidatedNSBlockIter(t *testing.T) {
 			end:      now.Add(1200 * time.Second),
 			stepSize: 60 * time.Second,
 			expectedResults: []float64{1, 2, nan, nan, nan, 3, nan, nan, nan, nan,
-				nan, nan, nan, nan, nan, nan, nan, 5, nan, 6},
+				nan, nan, nan, nan, nan, nan, nan, 5, nan, 6, nan},
```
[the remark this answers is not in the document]
benraskin92: Yeah logic changed to be inclusive of the last step.

src/coordinator/ts/m3db/block/consolidated_series.go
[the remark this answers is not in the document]
arnikola: I guess it doesn't really matter since it shouldn't be // nolint much longer :p

src/coordinator/ts/m3db/block/ns_series_test.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	now = time.Now()
+	nan = math.NaN()
+)
+
+func createDatapoints(t *testing.T, timeInSeconds []int, vals []float64, now time.Time) []ts.Datapoint {
+	if len(timeInSeconds) != len(vals) {
```
arnikola: nit: This check is unnecessary, require.Equal will exit out of the function if the condition fails

src/coordinator/ts/m3db/block/ns_series_test.go
```diff
@@ -0,0 +1,122 @@
...
+			start:    now,
+			end:      now.Add(1200 * time.Second),
+			stepSize: 60 * time.Second,
+			expectedResults: []float64{1, 2, nan, nan, nan, 3, nan, nan, nan, nan,
+				nan, nan, nan, nan, nan, nan, nan, 5, nan, 6, nan},
+			description: "testing multiple iterators",
+		},
+	}
+
+	for _, test := range testCases {
```
arnikola: Same comment here re: table testing in go

Your call whether to leave it or change it

src/coordinator/ts/m3db/block/ns_series.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+type consolidatedNSBlockIters []*consolidatedNSBlockIter
+
+// Next moves to the next item
+func (c *consolidatedNSBlockIter) Next() bool {
+	c.idx++
+	// NB(braskin): this is inclusive of the last step in the iterator
+	indexTime, err := c.bounds.TimeForIndex(c.idx)
+	if err != nil {
```
arnikola: Feels a little strange to return no next when we have an error. Can you add a comment here as to the motivation for this?

src/coordinator/ts/m3db/block/ns_series.go
```diff
@@ -0,0 +1,107 @@
...
+
+	lastDP := c.lastDP
+	// NB(braskin): check to make sure that the current index time is after the last
+	// seen datapoint and Next() on the underlaying m3db iterator returns true
+	for indexTime.After(lastDP.Timestamp) && c.nextIterator() {
+		lastDP, _, _ = c.m3dbIters[c.seriesIndex].Current()
+		c.lastDP = lastDP
+	}
+
+	return true
```
arnikola: Should this return false if your lastDP timestamp is larger than the end bounds? Otherwise looks like `Next()` will always return true, which may confuse  downstream consumers

src/coordinator/ts/m3db/block/ns_series.go
arnikola: edit; looked at TimeForIndex and looks like it might need a little tweaking to resolve my problems here; would you be able to change it to something like:

```golang
// TimeForIndex returns the start time for a given index assuming a uniform step size
func (b Bounds) TimeForIndexWithinBounds(idx int) (time.Time, bool) {
	step := b.StepSize
	t := b.Start.Add(time.Duration(idx) * step)
	if t.After(b.End) {
		return time.Time{}, false
	}

	return t, false
}
```

Then we can avoid relying on `errors` for execution

src/coordinator/ts/m3db/block/ns_series.go
arnikola: post edit edit: This may need to have a special case to return false sometimes depending on if you've reached the end of the iterators; might need to break out the `... && c.nextIterator` into a variable or something and then return it here?

src/coordinator/ts/m3db/block/multi_series_test.go
```diff
@@ -0,0 +1,178 @@
...
+			description:     "only return one set of datapoints",
+		},
+		{
+			dps:             [][]float64{{6, 7, 8, 9, 10}, {1, 2, 3, 4, 5}},
+			expectedResults: [][]float64{{6, 1}, {7, 2}, {8, 3}, {9, 4}, {10, 5}},
+			description:     "take only the first set of datapoints (consolidation)",
+		},
+	}
+
+	for _, test := range testCases {
```
arnikola: Same comment here re: table testing in go. Take it or leave it

src/coordinator/ts/m3db/block/multi_series_test.go
```diff
@@ -0,0 +1,178 @@
...
+					Tags: map[string]string{"my": "other_tag", "same": "tag"},
+				},
+			},
+			blockMeta: block.Metadata{
+				Tags: map[string]string{"same": "tag"},
+			},
+		},
+	}
+
+	for _, test := range testCases {
```
arnikola: Same comment re table testing here

src/coordinator/ts/m3db/block/iter_conversion.go
```diff
@@ -54,9 +54,10 @@ func (b blockReplicas) Less(i, j int) bool {
 	return b[i].start.Before(b[j].start)
 }
 
-// ConvertM3DBSeriesIterators converts m3db SeriesIterators to SeriesBlocks
+// IteratorsToSeriesBlocks converts m3db SeriesIterators to SeriesBlocks
 // which are used to construct Blocks for query processing.
-func ConvertM3DBSeriesIterators(iterators encoding.SeriesIterators, iterAlloc encoding.ReaderIteratorAllocate) ([]SeriesBlocks, error) {
+func IteratorsToSeriesBlocks(
+	iterators encoding.SeriesIterators, iterAlloc encoding.ReaderIteratorAllocate) ([]SeriesBlocks, error) {
```
arnikola: nit: can we add newlines in the args here?

src/coordinator/ts/m3db/block/iter_conversion.go
```diff
@@ -77,7 +78,8 @@ func ConvertM3DBSeriesIterators(iterators encoding.SeriesIterators, iterAlloc en
 	return multiSeriesBlocks, nil
 }
 
-func blockReplicasFromSeriesIterator(seriesIterator encoding.SeriesIterator, iterAlloc encoding.ReaderIteratorAllocate) ([]blockReplica, error) {
+func blockReplicasFromSeriesIterator(
```
arnikola: nit: can we add newlines in args here?

src/coordinator/ts/m3db/block/iter_conversion.go
```diff
@@ -122,7 +124,8 @@ func blockReplicasFromSeriesIterator(seriesIterator encoding.SeriesIterator, ite
 	return blockReplicas, nil
 }
 
-func seriesBlocksFromBlockReplicas(blockReplicas []blockReplica, seriesIterator encoding.SeriesIterator) (SeriesBlocks, error) {
+func seriesBlocksFromBlockReplicas(
```
arnikola: nit: can we add newlines in args here?

src/coordinator/ts/m3db/block/multi_series.go
```diff
@@ -0,0 +1,155 @@
...
+)
+
+// MultiSeriesBlock represents a vertically oriented block
+type MultiSeriesBlock struct {
+	Blocks   ConsolidatedSeriesBlocks
+	Metadata block.Metadata
+}
+
+// MultiSeriesBlocks is a slice of MultiSeriesBlock
+// todo(braskin): add close method on this to close each SeriesIterator
```
arnikola: nit: Might as well add Close() method and put the todo in that 😛

src/coordinator/ts/m3db/block/multi_series.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	consolidatedNSBlocksLen := len(blocks[0].ConsolidatedNSBlocks)
+	for i, seriesBlock := range blocks {
+		consolidatedNSBlockIters := make([]block.ValueIterator, consolidatedNSBlocksLen)
+		for j, nsBlock := range seriesBlock.ConsolidatedNSBlocks {
+			nsBlockIter := newConsolidatedNSBlockIter(nsBlock)
+			consolidatedNSBlockIters[j] = nsBlockIter
+		}
+		seriesBlockIters[i] = &consolidatedSeriesBlockIter{
```
arnikola: nit: newline

src/coordinator/ts/m3db/block/converter_test.go
```diff
@@ -0,0 +1,170 @@
...
+
+	tagIter, err := ident.NewTagStringsIterator(tags...)
+	require.NoError(t, err)
+	seriesIterator.EXPECT().Tags().Return(tagIter)
+
+	seriesIterator.EXPECT().Next().Return(true).Times(len(dps))
+	seriesIterator.EXPECT().Next().Return(false)
+
+	for _, dp := range dps {
+		seriesIterator.EXPECT().Current().Return(ts.Datapoint{Timestamp: dp.Timestamp, Value: dp.Value}, xtime.Millisecond, nil).Times(1)
```
arnikola: nit: might be a bit easier to read if you generate the point separately
`dp := ts.Datapoint{Timestamp: dp.Timestamp, Value: dp.Value}`

src/coordinator/ts/m3db/block/converter.go
```diff
@@ -96,16 +146,20 @@ func newConsolidatedSeriesBlocks(multiNamespaceSeries MultiNamespaceSeries, seri
 
 // newConsolidatedNSBlocks creates a slice of consolidated blocks per namespace for a single timeseries
 // nolint: unparam
-func newConsolidatedNSBlocks(seriesBlocks SeriesBlocks, seriesIteratorsPool encoding.MutableSeriesIteratorsPool) []ConsolidatedNSBlock {
+func newConsolidatedNSBlocks(
+	seriesBlocks SeriesBlocks, seriesIteratorsPool encoding.MutableSeriesIteratorsPool, stepSize time.Duration) []ConsolidatedNSBlock {
```
arnikola: nit: newlines in args list

src/coordinator/ts/m3db/block/converter.go
```diff
@@ -31,45 +35,88 @@ var (
...
-				multiSeriesBlocks[consolidatedSeriesBlockIdx].End = consolidatedSeriesBlock.End
+			if seriesIdx == 0 {
+				blockBounds := multiSeriesBlocks[consolidatedSeriesBlockIdx].Metadata.Bounds
+				blockBounds.StepSize = consolidatedSeriesBlock.Metadata.Bounds.StepSize
+				blockBounds.Start = consolidatedSeriesBlock.Metadata.Bounds.Start
+				blockBounds.End = consolidatedSeriesBlock.Metadata.Bounds.End
+				multiSeriesBlocks[consolidatedSeriesBlockIdx].Metadata.Bounds = blockBounds
+			}
+
+			dupedTags := consolidatedSeriesBlock.ConsolidatedNSBlocks[0].SeriesIterators.Iters()[0].Tags().Duplicate()
```
arnikola: Can you add comments here, or add a method on consolidatedNSBlocks that gives tags? There's a lot of indexing here without any clear guarantees that it's safe to do

src/coordinator/ts/m3db/block/converter.go
```diff
@@ -31,45 +35,88 @@ var (
...
+		m[i].Metadata.Tags = commonTags
+	}
+}
+
+func (m MultiSeriesBlocks) commonTags() models.Tags {
+	if len(m) == 0 || len(m[0].Blocks) == 0 {
+		return models.Tags{}
+	}
+
+	commonTags := m[0].Blocks[0].Metadata.Tags.Clone()
```
arnikola: Is the clone necessary?

src/coordinator/ts/m3db/block/converter.go
```diff
@@ -31,45 +35,88 @@ var (
...
 		}
 
 		for consolidatedSeriesBlockIdx, consolidatedSeriesBlock := range consolidatedSeriesBlocks {
+
 			// we only want to set the start and end times once
-			if multiNamespaceSeriesIdx == 0 {
-				multiSeriesBlocks[consolidatedSeriesBlockIdx].Start = consolidatedSeriesBlock.Start
-				multiSeriesBlocks[consolidatedSeriesBlockIdx].End = consolidatedSeriesBlock.End
+			if seriesIdx == 0 {
+				blockBounds := multiSeriesBlocks[consolidatedSeriesBlockIdx].Metadata.Bounds
```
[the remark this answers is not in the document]
benraskin92: I don't think that works because the underlaying structs aren't pointers.

src/coordinator/ts/m3db/block/ns_series.go
benraskin92: I'm gonna leave this for now. I'll add a comment though.

src/coordinator/ts/m3db/block/converter.go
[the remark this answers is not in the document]
arnikola: Does it matter if we're just making a copy though?

src/coordinator/ts/m3db/block/converter.go
benraskin92: yes

src/coordinator/ts/m3db/block/consolidated_series.go
```diff
@@ -0,0 +1,119 @@
...
+
+// ConsolidationFunc determines how to consolidate across namespaces
+type ConsolidationFunc func(existing, toAdd float64, count int) float64
+
+// ConsolidatedBlocks contain all of the consolidated blocks for
+// a single timeseries across namespaces.
+type ConsolidatedBlocks []ConsolidatedBlock
+
+func (c *consolidatedBlockStepIter) Current() float64 {
+	values := make([]float64, 0, 1)
```
nikunjgit: make([]float64, 0, len(c.nsBlockSeriesIters))

src/coordinator/ts/m3db/block/consolidated_series.go
```diff
@@ -0,0 +1,119 @@
...
+	for _, iter := range c.nsBlockStepIters {
+		dp := iter.Current()
+		values = append(values, dp)
+	}
+
+	if len(values) > 0 {
```
nikunjgit: just return values[0], I don't think you need to check since Current() is only called when Next() returns true. You should never have to return math.NaN()

src/coordinator/ts/m3db/block/consolidated_series.go
```diff
@@ -0,0 +1,119 @@
...
+	return true
+}
+
+// Close closes the underlaying iterators
+func (c *consolidatedBlockStepIter) Close() {
+	// todo(braskin): implement this function
+}
+
+// Current returns a slice of values for the current series
+func (c *consolidatedBlockSeriesIter) Current() []float64 {
```
nikunjgit: actually at conslidatedBlock level, doesn't look like you have a concept of series and steps. I think you should just have one iterator : consolidatedBlockIter which should return a single float64 on current. The concept of Step and Series probably comes at a higher level

src/coordinator/ts/m3db/block/ns_series.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/m3db/m3db/src/coordinator/block"
+	"github.com/m3db/m3db/src/dbnode/encoding"
+	"github.com/m3db/m3db/src/dbnode/ts"
+	"github.com/m3db/m3x/ident"
+)
+
+// NSBlock is a single block for a given timeseries and namespace
+// which contains all of the necessary SeriesIterators so that consolidation can
+// happen across namespaces
+type NSBlock struct {
```
nikunjgit: same comment as above, at this level you don't have a concept of step and series since you just are representing a single series. Having a single iter with Current returning float64 ?

nikunjgit:
Are you still working on it or close it out ?

PR #779, opened 2018-07-03
Implement logical AND in query

nikunjgit:


src/coordinator/executor/transform/cache.go
```diff
@@ -0,0 +1,63 @@
...
+// Remove the block from the cache
+func (c *BlockCache) Remove(key parser.NodeID) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	delete(c.blocks, key)
+}
+
+// Get the block from the cache
+func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
+	c.mu.Lock()
```
arnikola: This seems incorrect; multiple clients can get the same block and mutate it, and. the Seems very ; it's necessary to lock the block returned to stop

src/coordinator/test/block.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"github.com/m3db/m3db/src/coordinator/block"
+	"github.com/m3db/m3db/src/coordinator/executor/transform"
+	"github.com/m3db/m3db/src/coordinator/models"
+	"github.com/m3db/m3db/src/coordinator/parser"
+)
+
+// NewBlockFromValues creates a new block using the provide values
+func NewBlockFromValues(bounds block.Bounds, seriesValues [][]float64) block.Block {
+	blockMeta := block.Metadata{Bounds: bounds}
```
benraskin92: Add any tags for the Block metadata?

src/coordinator/functions/logical/and_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func generateValues(vals [][]float64, b *block.Bounds) ([][]float64, block.Bounds) {
+	values := vals
+	if values == nil {
+		values = [][]float64{
+			{0, 1, 2, 3, 4},
+			{5, 6, 7, 8, 9},
```
benraskin92: maybe throw in some `NaN`s?

src/coordinator/functions/logical/node.go
```diff
@@ -0,0 +1,101 @@
...
+
+// String representation
+func (o BaseOp) String() string {
+	return fmt.Sprintf("type: %s, lnode: %s, rnode: %s", o.OpType(), o.LNode, o.RNode)
+}
+
+// Node creates an execution node
+func (o BaseOp) Node(controller *transform.Controller) transform.OpNode {
+	return &BaseNode{
+		op:        o, controller: controller,
```
benraskin92: nit: move `controller: controller` to newline

src/coordinator/functions/logical/common.go
```diff
@@ -0,0 +1,75 @@
...
+	return func(tags models.Tags) uint64 { return tags.IDWithExcludes(names...) }
+}
+
+// Processor is implemented by each logical transform
+type Processor interface {
+	Process(lhs block.Block, rhs block.Block) (block.Block, error)
+}
+
+// MakeProcessor is a way to create a logical transform
+type MakeProcessor func (op BaseOp, controller *transform.Controller) Processor
```
benraskin92: nit: newline

src/coordinator/test/block.go
nikunjgit: will add in the future if needed by tests

src/coordinator/functions/logical/and_test.go
nikunjgit: the next test does it

src/coordinator/functions/abs_test.go
```diff
@@ -0,0 +1,77 @@
...
+	c, sink := test.NewControllerWithSink(parser.NodeID(1))
+	node := (&AbsOp{}).Node(c)
+	err := node.Process(parser.NodeID(0), block)
+	require.NoError(t, err)
+	expected := expectedValues(values)
+	assert.Len(t, sink.Values, 2)
+	assert.Equal(t, expected, sink.Values)
+}
+
+func TestAbsWithSomeValues(t *testing.T) {
```
benraskin92: Worth it to write a test with all `NaN`s?

src/coordinator/parser/promql/parse_test.go
```diff
@@ -69,3 +71,23 @@ func TestDAGWithFunctionCall(t *testing.T) {
...
+	q := "up and up"
+	p, err := Parse(q)
+	require.NoError(t, err)
+	transforms, edges, err := p.DAG()
+	require.NoError(t, err)
+	assert.Len(t, transforms, 3)
+	assert.Equal(t, transforms[0].Op.OpType(), functions.FetchType)
+	assert.Equal(t, transforms[0].ID, parser.NodeID("0"))
+	assert.Equal(t, transforms[1].Op.OpType(), functions.FetchType)
+	assert.Equal(t, transforms[1].ID, parser.NodeID("1"))
```
benraskin92: This shouldn't be `logical.AndType` and the next one `functions.FetchType`?

nikunjgit: the lowest two are fetch types and the AndType is higher order function since you need to do fetches first before you can do AND

src/coordinator/functions/abs_test.go
nikunjgit: Functionally one Nan is similar to all NaNs since there's no special logic there.

src/coordinator/executor/transform/cache.go
```diff
@@ -0,0 +1,63 @@
...
+	defer c.mu.Unlock()
+	b, ok := c.blocks[key]
```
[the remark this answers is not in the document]
nikunjgit: block.Block doesn't have any mutable components based on the interface. I think its important to figure out how we can share the same blocks across two processes. It's not trivial since even if we do expiring locks and if the block is actually mutable, nothing stops the first process to mutate it even after its lock expired. Adding those checks might make it complicated. I'll add a TODO for now to revisit if the mutability assumption becomes false or if sharing becomes a problem

src/coordinator/models/tag.go
```diff
@@ -125,25 +132,86 @@ func (m Matchers) ToTags() (Tags, error) {
...
+	sortedKeys, bufLength := t.sortKeys()
+	b := make([]byte, 0, bufLength)
+	for _, k := range sortedKeys {
+		// Always exclude the metric name by default
+		if k == MetricName {
+			continue
+		}
+
+		found := false
+		for _, n := range excludeKeys {
```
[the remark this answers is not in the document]
nikunjgit: you mean binary searching over excluded keys ? I think the list would be super small so probably not worth it, can revisit later based on our benchmarks.

src/coordinator/parser/promql/types.go
```diff
@@ -82,6 +94,8 @@ func getOpType(opType promql.ItemType) string {
 	switch opType {
 	case promql.ItemType(itemCount):
 		return functions.CountType
+	case promql.ItemType(itemLAND):
```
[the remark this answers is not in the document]
nikunjgit: that's defined by prom. Didn't what to change it

src/coordinator/functions/logical/and.go
```diff
@@ -0,0 +1,119 @@
...
+	// The set of signatures for the right-hand side.
+	rightSigs := make(map[uint64]int, len(rhs))
+	for idx, meta := range rhs {
+		rightSigs[sigf(meta.Tags)] = idx
+	}
+
+	matches := make([]int, len(lhs))
+	// Initialize to -1
+	for i := range matches {
+		matches[i] = -1
```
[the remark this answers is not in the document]
nikunjgit: using ints for indices so can't have NaNs

src/coordinator/models/tag.go
```diff
@@ -125,25 +132,86 @@ func (m Matchers) ToTags() (Tags, error) {
...
-		b += sep
+	h := fnv.New64a()
+	h.Write(b)
+	return h.Sum64()
+}
+
+// IDWithKeys returns a string representation of the tags only including the given keys
+func (t Tags) IDWithKeys(includeKeys ...string) uint64 {
+	b := make([]byte, 0, len(t))
+	for _, k := range includeKeys {
```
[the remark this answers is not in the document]
nikunjgit: no since includeKeys are provided by users and they shouldn't use names. If they did, then they probably have a reason to provide it

src/coordinator/models/tag.go
```diff
@@ -125,25 +132,86 @@ func (m Matchers) ToTags() (Tags, error) {
...
-	return b
+	h := fnv.New64a()
+	h.Write(b)
+	return h.Sum64()
+}
+
+func (t Tags) sortKeys() ([]string, int) {
+	length := 0
+	keys := make([]string, 0, len(t))
+	for k := range t {
```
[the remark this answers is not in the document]
nikunjgit: why multiply by 2 ?

src/coordinator/parser/promql/parse.go
```diff
@@ -44,85 +44,135 @@ func Parse(q string) (parser.Parser, error) {
...
+
+func (p *parseState) transformLen() int {
+	return len(p.transforms)
+}
+
+func (p *parseState) walk(node pql.Node) error {
 	if node == nil {
-		return nil, nil, nil
+		return nil
 	}
```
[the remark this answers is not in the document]
nikunjgit: p can't be nil since we explicitly initialize it. expr can be nil but then we'll just have empty edges and transforms.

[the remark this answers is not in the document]
nikunjgit: added a test for empty expression. The parser will error out

src/coordinator/models/tag.go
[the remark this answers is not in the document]
arnikola: len(sep) + len(eq)  = 2 ;)

Happy to formalize it instead too

src/coordinator/functions/logical/common.go
```diff
@@ -0,0 +1,75 @@
...
+	// Include contains additional labels that should be included in
+	// the result from the side with the lower cardinality.
+	Include []string
+}
+
+// hashFunc returns a function that calculates the signature for a metric
+// ignoring the provided labels. If on, then the given labels are only used instead.
+func hashFunc(on bool, names ...string) func(models.Tags) uint64 {
+	if on {
+		return func(tags models.Tags) uint64 { return tags.IDWithKeys(names...) }
```
arnikola: These should be sorted first; I've seen tags come back in different orders from the db backend

Although, it may  make more sense to sort them in the IDWithKeys method

src/coordinator/parser/promql/parse.go
```diff
@@ -44,85 +44,135 @@ func Parse(q string) (parser.Parser, error) {
...
+		p.edges = append(p.edges, parser.Edge{
+			ParentID: p.lastTransformID(),
+			ChildID:  opTransform.ID,
+		})
+		p.transforms = append(p.transforms, opTransform)
+		return nil
+
+	case *pql.BinaryExpr:
+		err := p.walk(n.LHS)
+		lhsID := p.lastTransformID()
```
arnikola: Don't we want to walk both sides simultaneously?

src/coordinator/functions/logical/and.go
```diff
@@ -0,0 +1,115 @@
...
+	for index := 0; lIter.Next() && rIter.Next(); index++ {
+		lStep := lIter.Current()
+		lValues := lStep.Values()
+
+		rStep := rIter.Current()
+		rValues := rStep.Values()
+
+		for idx, value := range lValues {
+			rIdx := intersection[idx]
+			if rIdx < 0 || math.IsNaN(rValues[rIdx]) {
```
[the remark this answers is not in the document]
nikunjgit: something has to do to the output. I don't think they return fix resolution output which is how they might be getting away from it.

src/coordinator/functions/logical/node.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// Process processes a block
+func (c *BaseNode) Process(ID parser.NodeID, b block.Block) error {
+	var lhs block.Block
+	var rhs block.Block
+	if c.op.LNode == ID {
+		rBlock, ok := c.cache.Get(c.op.RNode)
+		if !ok {
+			return c.cache.Add(ID, b)
```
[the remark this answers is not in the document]
nikunjgit: fixed the race. We have that mechanism in the base node here. Notice how And.go doesn't care about that logic. All binary operations can inherent from base node. We can move it to something more generic later if needed.

src/coordinator/functions/logical/common.go
nikunjgit: IDWithExcludes sorts them. IDWithKeys doesn't need sorting since they will always be in the order of the keys provided ?

src/coordinator/parser/promql/parse.go
nikunjgit: not during DAG generation since that's trivial. We are doing parallelism during processing

src/coordinator/functions/logical/node.go
```diff
@@ -64,37 +65,62 @@ type BaseNode struct {
...
+	nextBlock, err := c.processor.Process(lhs, rhs)
+	if err != nil {
+		return err
+	}
+
+	defer nextBlock.Close()
+	return c.controller.Process(nextBlock)
+}
+
+// computeOrCache figures out if both lhs and rhs are available, if not then it caches the new block
```
arnikola: nit: `caches the incoming block`

src/coordinator/functions/logical/node.go
```diff
@@ -64,37 +65,62 @@ type BaseNode struct {
...
+		return nil
+	}
+
+	defer c.cleanup()
+	nextBlock, err := c.processor.Process(lhs, rhs)
+	if err != nil {
+		return err
+	}
+
+	defer nextBlock.Close()
```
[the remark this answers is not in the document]
nikunjgit: yeah probably need a refCount since multiple downstream functions can use the block. Something for later ?

src/coordinator/functions/logical/node.go
```diff
@@ -64,37 +65,62 @@ type BaseNode struct {
...
+	if lhs == nil || rhs == nil {
+		return nil
+	}
+
+	defer c.cleanup()
+	nextBlock, err := c.processor.Process(lhs, rhs)
+	if err != nil {
+		return err
+	}
+
```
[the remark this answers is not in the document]
nikunjgit: just two defers :P Otherwise, we would have to call c.cleanup() twice.

PR #784, opened 2018-07-05
Support for clamp functions

nikunjgit:


src/coordinator/functions/clamp.go
```diff
@@ -0,0 +1,114 @@
...
+
+// ClampOp stores required properties for clamp
+type ClampOp struct {
+	optype string
+	scalar float64
+}
+
+// NewClampOp creates a new clamp op based on the type and arguments
+func NewClampOp(args []interface{}, optype string) (ClampOp, error) {
+	if len(args) != 1 {
```
benraskin92: If we require len to be 1, why not just have `interface{}` instead of `[]interface{}` in the function definition?

src/coordinator/functions/clamp_test.go
```diff
@@ -0,0 +1,78 @@
...
+	require.NoError(t, err)
+	node := op.Node(c)
+	err = node.Process(parser.NodeID(0), block)
+	require.NoError(t, err)
+	expected := expectedClampVals(values, 3.0, math.Min)
+	assert.Len(t, sink.Values, 2)
+	test.EqualsWithNans(t, expected, sink.Values)
+}
+
+
```
benraskin92: nit: remove newline

src/coordinator/functions/clamp.go
```diff
@@ -0,0 +1,114 @@
...
+		return ClampOp{}, fmt.Errorf("invalid number of args for clamp: %d", len(args))
+	}
+
+	if optype != ClampMinType && optype != ClampMaxType {
+		return ClampOp{}, fmt.Errorf("unknown clamp type: %s", optype)
+	}
+
+	scalar, ok := args[0].(float64)
+	if !ok {
```
benraskin92: Can you add test case for this?

src/coordinator/functions/clamp.go
nikunjgit: so if you look at the parser, the idea is that parser doesn't know exactly how to provide the args, each function figures out how it uses its args

src/coordinator/functions/clamp.go
```diff
@@ -0,0 +1,114 @@
...
+func NewClampOp(args []interface{}, optype string) (ClampOp, error) {
+	if len(args) != 1 {
+		return ClampOp{}, fmt.Errorf("invalid number of args for clamp: %d", len(args))
+	}
+
+	if optype != ClampMinType && optype != ClampMaxType {
+		return ClampOp{}, fmt.Errorf("unknown clamp type: %s", optype)
+	}
+
+	scalar, ok := args[0].(float64)
```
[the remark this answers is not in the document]
nikunjgit: same comment as above, the idea is that parser doesn't know exactly how to provide the args, each function figures out how it uses its args. I only need to cat float64 here so didn't want to use a switch

PR #785, opened 2018-07-05
Add rudimentary multi-cluster support for m3coordinator

robskillington:
This adds rudimentary multi-cluster support for m3coordinator so that we can begin to send metrics to different namespaces for different retention periods, etc.

Right now this is not optimal as it retrieves data from all matching namespaces that has retention long enough to cover the query, then returns the most granular datapoints that come back and discards any lower granularity datapoints that also came back.  This is because we do not have a distributed index in the current M3 open source offering.  For read workloads that do not require tens of thousands or hundreds of thousands realtime alert evaluations, this should be quite sufficient given that the write volume absolutely dwarfs the read volume.

At some point we'll hopefully have a distributed index that we can use for all use cases and deprecate this behavior.

Next:
- [x] Test coverage
- [x] Fix any out of date configs


src/coordinator/api/v1/handler/database/create.go
```diff
@@ -246,42 +294,50 @@ func defaultedPlacementInitRequest(r *admin.DatabaseCreateRequest, dbCfg dbconfi
...
-		for _, hostnameGroup := range r.HostnameGroups {
 			instances = append(instances, &placementpb.Instance{
-				Id:             hostnameGroup.Hostname,
-				IsolationGroup: hostnameGroup.IsolationGroup,
-				Zone:           "embedded",
-				Weight:         1,
-				Endpoint:       fmt.Sprintf("http://%s:%d", hostnameGroup.Hostname, port),
-				Hostname:       hostnameGroup.Hostname,
-				Port:           uint32(port),
+				Id:             id,
```
cw9: Any particular reason we are accessing the protobuf directly here instead of using the placement.Instance interface?

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -246,42 +294,50 @@ func defaultedPlacementInitRequest(r *admin.DatabaseCreateRequest, dbCfg dbconfi
...
+				IsolationGroup: isolationGroup,
+				Zone:           zone,
+				Weight:         weight,
+				Endpoint:       fmt.Sprintf("%s:%d", host.Address, host.Port),
```
cw9: I'd recommend fill in the `Hostname` field with `id` or `host.Address` for now, so that later when we rename that field in `placement.Instance` it's easier for us to update here. According to the protobuf definition, seems host.Address is what you want here.

src/coordinator/api/v1/handler/database/create.go
robskillington: This is because an internal handler is called where the instances are submitted by JSON but described by protobuf.  (This is a composite handler and uses the internal handlers directly as if a user was sending a request)

src/coordinator/api/v1/handler/database/create.go
robskillington: Sure thing.

src/coordinator/api/v1/handler/database/create.go
cw9: I see, it feels slightly dangerous though, cause usually it's safe to rename a field in protobuf, but since we are converting it into json here (which also pretty far away from m3cluster, where the pb is defined), it may change the how the json is represented, should we decouple the pb and json representation?

src/coordinator/storage/local/cluster_test.go
```diff
@@ -0,0 +1,135 @@
...
+	newClientFromConfig,
+	*client.MockSession,
+) {
+	mockSession := client.NewMockSession(ctrl)
+
+	mockClient := client.NewMockClient(ctrl)
+	mockClient.EXPECT().DefaultSession().Return(mockSession, nil).AnyTimes()
+
+	newClientFn := func(
+		cfg client.Configuration,
```
arnikola: nit: can use _ as these variable names?

src/coordinator/storage/local/cluster_test.go
```diff
@@ -0,0 +1,135 @@
...
+
+	// Resolve clusters and check attributes
+	unaggregatedNs := clusters.UnaggregatedClusterNamespace()
+	assert.Equal(t, "unaggregated", unaggregatedNs.NamespaceID().String())
+	assert.Equal(t, storage.Attributes{
+		MetricsType: storage.UnaggregatedMetricsType,
+		Retention:   7 * 24 * time.Hour,
+	}, unaggregatedNs.Attributes())
+	assert.True(t, mockSession1 == unaggregatedNs.Session())
+
```
arnikola: Add something like this for the negative case?

```golang
 	_, ok := clusters.AggregatedClusterNamespace(RetentionResolution{
 		Retention:  time.Minute,
 		Resolution: time.Minute,
 	})
 	require.False(t, ok)
```

src/coordinator/storage/local/cluster_test.go
```diff
@@ -0,0 +1,135 @@
...
+	require.NoError(t, err)
+
+	// Resolve clusters and check attributes
+	unaggregatedNs := clusters.UnaggregatedClusterNamespace()
+	assert.Equal(t, "unaggregated", unaggregatedNs.NamespaceID().String())
+	assert.Equal(t, storage.Attributes{
+		MetricsType: storage.UnaggregatedMetricsType,
+		Retention:   7 * 24 * time.Hour,
+	}, unaggregatedNs.Attributes())
+	assert.True(t, mockSession1 == unaggregatedNs.Session())
```
arnikola: nit: use `assert.Equal`?

src/coordinator/storage/local/storage.go
```diff
@@ -22,32 +22,35 @@ package local
 
 import (
 	"context"
+	goerrors "errors"
```
arnikola: Is it standard to alias this, or should we leave this as `"errors"`?

src/coordinator/test/local/storage.go
```diff
@@ -1,16 +1,50 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
arnikola: nit: 2018

src/coordinator/test/local/storage.go
```diff
@@ -1,16 +1,50 @@
...
+func NewStorageAndSession(
+	t *testing.T,
+	ctrl *gomock.Controller,
+) (storage.Storage, *client.MockSession) {
 	session := client.NewMockSession(ctrl)
-	storage := local.NewStorage(session, "metrics", nil)
+	clusters, err := local.NewClusters(local.UnaggregatedClusterNamespaceDefinition{
+		NamespaceID: ident.StringID("metrics"),
+		Session:     session,
+		Retention:   30 * 24 * time.Hour,
```
arnikola: Might be useful to pull this retention out into a public variable in case a test needs to validate it?

src/coordinator/storage/local/config.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	if numUnaggregatedClusterNamespaces != 1 {
+		return nil, fmt.Errorf("one unaggregated cluster namespace  "+
+			"must be specified: specified %d", numUnaggregatedClusterNamespaces)
+	}
+
+	// Connect to all clusters in parallel
+	var wg sync.WaitGroup
+	wg.Add(1)
```
arnikola: Worth having`wg.Add(1 + len(aggregatedClusterNamespacesCfgs))` ?

src/coordinator/storage/local/config.go
```diff
@@ -0,0 +1,190 @@
...
+	var (
+		numUnaggregatedClusterNamespaces int
+		numAggregatedClusterNamespaces   int
+		unaggregatedClusterNamespaceCfg  = &unaggregatedClusterNamespaceConfiguration{}
+		aggregatedClusterNamespacesCfgs  []*aggregatedClusterNamespacesConfiguration
+		unaggregatedClusterNamespace     UnaggregatedClusterNamespaceDefinition
+		aggregatedClusterNamespaces      []AggregatedClusterNamespaceDefinition
+	)
+	for _, clusterCfg := range c {
+		client, err := clusterCfg.newClient(client.ConfigurationParameters{})
```
arnikola: nit: Pull out `client.ConfigurationParameters{}` into a variable?

src/coordinator/storage/local/config.go
```diff
@@ -0,0 +1,190 @@
...
+
+	unaggregatedClusterNamespace = UnaggregatedClusterNamespaceDefinition{
+		NamespaceID: ident.StringID(unaggregatedClusterNamespaceCfg.namespace.Namespace),
+		Session:     unaggregatedClusterNamespaceCfg.result.session,
+		Retention:   unaggregatedClusterNamespaceCfg.namespace.Retention,
+	}
+
+	for i, cfg := range aggregatedClusterNamespacesCfgs {
+		if cfg.result.err != nil {
+			return nil, fmt.Errorf("could not connect to aggregated cluster #%d: %v",
```
arnikola: Might be better to have the invalid cluster ID rather than the number? Deduping could mess up the order and make this message less useful

src/coordinator/storage/local/storage.go
```diff
@@ -65,14 +68,64 @@ func (s *localStorage) Fetch(ctx context.Context, query *storage.FetchQuery, opt
...
+		wg.Add(1)
+		go func() {
+			r, err := s.fetch(namespace, m3query, opts)
+			result.add(namespace.Attributes(), r, err)
+			wg.Done()
+		}()
+	}
+
+	if fetches == 0 {
+		return nil, errNoLocalClustersFulfillsQuery
```
arnikola: Worth adding failing query to error?

src/coordinator/storage/local/storage.go
```diff
@@ -161,11 +264,46 @@ func (s *localStorage) FetchBlocks(
 	return res, nil
 }
 
+func (s *localStorage) Close() error {
+	return nil
```
arnikola: Add a //TODO so we don't forget to close later?

src/coordinator/storage/local/storage.go
```diff
@@ -190,7 +329,119 @@ func newWriteRequest(writeRequestCommon *writeRequestCommon, timestamp time.Time
...
+				attrs: r.dedupeFirstAttrs,
+			}
+		}
+	}
+
+	for _, s := range result.SeriesList {
+		id := s.Name()
+		existing, exists := r.dedupeMap[id]
+		if exists && existing.attrs.Resolution <= attrs.Resolution {
+			// Already exists and resolution is already more finer grained
```
arnikola: nit: `already finer`

src/coordinator/storage/local/storage.go
```diff
@@ -190,7 +329,119 @@ func newWriteRequest(writeRequestCommon *writeRequestCommon, timestamp time.Time
...
+			continue
+		}
+
+		// Does not exist already or more finer grained, add result
```
arnikola: nit: `or finer`

src/coordinator/storage/local/storage_test.go
```diff
@@ -22,30 +22,67 @@ package local
...
 	"github.com/stretchr/testify/require"
 )
 
-func setup(ctrl *gomock.Controller) (storage.Storage, *client.MockSession) {
+type testSessions struct {
+	unaggregated1MonthRetention                *client.MockSession
+	aggregated1MonthRetention1MinuteResolution *client.MockSession
+}
+
+func (s testSessions) forEach(fn func(session *client.MockSession)) {
```
arnikola: 👍

src/coordinator/storage/local/storage_test.go
```diff
@@ -85,32 +122,95 @@ func newWriteQuery() *storage.WriteQuery {
...
+	writeQuery := newWriteQuery()
+	// Use unsupported retention/resolution
+	writeQuery.Attributes = storage.Attributes{
+		MetricsType: storage.AggregatedMetricsType,
+		Retention:   1234,
+		Resolution:  5678,
+	}
+	err := store.Write(context.TODO(), writeQuery)
+	assert.Error(t, err)
+	assert.True(t, strings.Contains(err.Error(), "no configured cluster namespace"),
```
arnikola: I think you can do something like `assert.EqualError`? Might be a bit iffy if you don't have the exact error string we're expecting though

src/coordinator/storage/local/storage_test.go
```diff
@@ -123,16 +223,101 @@ func TestLocalRead(t *testing.T) {
...
+
+	sessions.forEach(func(session *client.MockSession) {
+		var f testFetchTaggedID
+		switch {
+		case session == sessions.unaggregated1MonthRetention:
+			f = fetches[0]
+		case session == sessions.aggregated1MonthRetention1MinuteResolution:
+			f = fetches[1]
+		default:
+			panic("unexpected session")
```
arnikola: Might be better to use `require.FailNow` to avoid a panic?

src/coordinator/storage/local/storage_test.go
```diff
@@ -123,16 +223,101 @@ func TestLocalRead(t *testing.T) {
...
+		)
+
+		session.EXPECT().FetchTaggedIDs(gomock.Any(), gomock.Any(), gomock.Any()).
+			Return(iter, true, nil)
+	})
+	searchReq := newFetchReq()
+	result, err := store.FetchTags(context.TODO(), searchReq, &storage.FetchOptions{Limit: 100})
+	require.NoError(t, err)
+
+	require.Equal(t, len(fetches), len(result.Metrics))
```
arnikola: nit: `require.Len(t, result.Metrics, len(fetches))`

src/coordinator/storage/local/storage.go
```diff
@@ -65,14 +68,64 @@ func (s *localStorage) Fetch(ctx context.Context, query *storage.FetchQuery, opt
...
+		namespaces = s.clusters.ClusterNamespaces()
+		now        = time.Now()
+		fetches    = 0
+		result     multiFetchResult
+		wg         sync.WaitGroup
+	)
+	for _, namespace := range namespaces {
+		namespace := namespace // Capture var
+
+		clusterStart := now.Add(-1 * namespace.Attributes().Retention)
```
nikunjgit: instead of this, can you just check if !clusterStart.Before(query.Start) {...}

src/coordinator/storage/local/storage.go
```diff
@@ -65,14 +68,64 @@ func (s *localStorage) Fetch(ctx context.Context, query *storage.FetchQuery, opt
...
+		// Only include if cluster can completely fulfill the range
+		if clusterStart.After(query.Start) {
+			continue
+		}
+
+		fetches++
+
+		wg.Add(1)
+		go func() {
+			r, err := s.fetch(namespace, m3query, opts)
```
nikunjgit: do we want to exit on first error ? If yes, then probably use ExecuteParallel, if not, then probably add a comment since we shortcircuit on first error in most cases.

src/coordinator/storage/local/storage.go
```diff
@@ -90,9 +143,54 @@ func (s *localStorage) FetchTags(ctx context.Context, query *storage.FetchQuery,
...
+		namespaces = s.clusters.ClusterNamespaces()
+		now        = time.Now()
+		fetches    = 0
+		result     multiFetchTagsResult
+		wg         sync.WaitGroup
+	)
+	for _, namespace := range namespaces {
+		namespace := namespace // Capture var
+
+		clusterStart := now.Add(-1 * namespace.Attributes().Retention)
```
nikunjgit: instead of this, can you just check if !clusterStart.Before(query.Start) {...}

src/coordinator/storage/local/storage.go
```diff
@@ -90,9 +143,54 @@ func (s *localStorage) FetchTags(ctx context.Context, query *storage.FetchQuery,
...
+		// Only include if cluster can completely fulfill the range
+		if clusterStart.After(query.Start) {
+			continue
+		}
+
+		fetches++
+
+		wg.Add(1)
+		go func() {
+			result.add(s.fetchTags(namespace, m3query, opts))
```
nikunjgit: same as above: do we want to exit on first error ? If yes, then probably use ExecuteParallel, if not, then probably add a comment since we shortcircuit on first error in most cases.

src/coordinator/storage/local/storage.go
nikunjgit: Already exists and resolution is higher ?

src/coordinator/storage/local/storage.go
nikunjgit: Does not exist already or lower resolution, add result

src/coordinator/storage/local/storage.go
```diff
@@ -190,7 +329,119 @@ func newWriteRequest(writeRequestCommon *writeRequestCommon, timestamp time.Time
...
+) {
+	r.Lock()
+	defer r.Unlock()
+
+	if err != nil {
+		r.err = r.err.Add(err)
+		return
+	}
+
+	if r.result == nil {
```
nikunjgit: is it a bug to initialize r.result but not r.dedupeMap ? Seems like this may not work since for the second result, you will just add duplicate things to the back of the seriesList

src/coordinator/storage/local/storage.go
```diff
@@ -190,7 +329,119 @@ func newWriteRequest(writeRequestCommon *writeRequestCommon, timestamp time.Time
...
```
nikunjgit: same as above, probably a bug to add to r.result but not to r.dedupeMap ?

src/coordinator/api/v1/handler/database/create.go
```diff
@@ -183,61 +214,78 @@ func (h *createHandler) parseRequest(r *http.Request) (*admin.NamespaceAddReques
...
 
-		retentionPeriod := options.RetentionOptions().RetentionPeriod()
+		retentionPeriod := retentionOpts.RetentionPeriod()
 		if r.ExpectedSeriesDatapointsPerHour > 0 {
 			blockSize := time.Duration(blockSizeFromExpectedSeriesScalar / r.ExpectedSeriesDatapointsPerHour)
-			if blockSize < minBlockSize {
-				blockSize = minBlockSize
-			} else if blockSize > maxBlockSize {
-				blockSize = maxBlockSize
+			if blockSize < minRecommendCalculateBlockSize {
```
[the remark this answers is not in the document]
robskillington: The way this should be read is, if the "calculated block size" (here `blockSize`) is less than the "minimum to recommend when calculating block size" (here `minRecommendCalculateBlockSize`) then just set it to the "minimum to recommend when calculating block size" (i.e. `minRecommendCalculateBlockSize`).

Does that make sense?  So it is recommended, its just the minimum we would recommend when making recommendation based on calculating block sizes when specifying expected datapoints per block.

src/coordinator/api/v1/handler/database/create.go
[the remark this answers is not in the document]
robskillington: It can be a hostname or an IP address unfortunately.  So in that case, it's tough to know whether its a hostname (alphanumeric) or IPv4 or IPv6.  To do validation may cause valid hostnames to be deemed invalid potentially I believe?

src/coordinator/api/v1/handler/placement/common.go
```diff
@@ -35,12 +38,16 @@ import (
...
 	// DefaultServiceName is the default service ID name
 	DefaultServiceName = "m3db"
-
 	// DefaultServiceEnvironment is the default service ID environment
 	DefaultServiceEnvironment = "default_env"
-
 	// DefaultServiceZone is the default service ID zone
 	DefaultServiceZone = "embedded"
+	// HeaderClusterServiceName is the header used to specify the service name.
+	HeaderClusterServiceName = "Cluster-Service-Name"
```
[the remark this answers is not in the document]
robskillington: All the rage now is to abide by the "do not use `X-`..." RFC, hah

src/coordinator/api/v1/handler/placement/common_test.go
```diff
@@ -48,20 +48,20 @@ func TestPlacementService(t *testing.T) {
 	mockClient.EXPECT().Services(gomock.Not(nil)).Return(mockServices, nil)
 	mockServices.EXPECT().PlacementService(gomock.Not(nil), gomock.Not(nil)).Return(mockPlacementService, nil)
 
-	placementService, err := Service(mockClient, config.Configuration{})
+	placementService, err := Service(mockClient, http.Header{})
```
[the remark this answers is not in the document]
robskillington: Good call, will do.

src/coordinator/services/m3coordinator/server/server.go
```diff
@@ -112,44 +117,48 @@ func Run(runOpts RunOptions) {
...
 		}
-
-		dbClient, err := clientCfg.NewClient(client.ConfigurationParameters{})
+	} else {
+		localCfg := cfg.Local
+		if localCfg == nil {
+			localCfg = defaultLocalConfiguration
+		}
+		dbClientCh := runOpts.DBClient
+		if localCfg == nil || dbClientCh == nil {
```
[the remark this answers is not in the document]
robskillington: Ta, missed this during a refactor.

src/coordinator/services/m3coordinator/server/server.go
```diff
@@ -201,19 +215,15 @@ func Run(runOpts RunOptions) {
 	signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
 
 	<-sigChan
-	if err := session.Close(); err != nil {
-		logger.Fatal("unable to close m3db client session", zap.Any("error", err))
+	if err := clusters.Close(); err != nil {
```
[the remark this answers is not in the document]
robskillington: Session is no longer created here, clusters holds references to all sessions and closes them when itself is closed.

src/coordinator/storage/config.go
```diff
@@ -0,0 +1,50 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package storage
+
+import "fmt"
+
+// IsValidMetricsType validates a stored metrics type.
```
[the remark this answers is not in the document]
robskillington: Yeah, I should change this to `ValidateThing...` name, I did want this for other components that need to validate input if/when this becomes dynamic (this pattern is common in M3DB enums, so I applied the same pattern here).

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+// cluster namespace that holds unaggregated metrics data.
+type UnaggregatedClusterNamespaceDefinition struct {
+	NamespaceID ident.ID
+	Session     client.Session
+	Retention   time.Duration
+}
+
+// Validate will validate the cluster namespace definition.
+func (def UnaggregatedClusterNamespaceDefinition) Validate() error {
+	if def.NamespaceID == nil || len(def.NamespaceID.String()) == 0 {
```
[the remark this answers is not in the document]
robskillington: Hm I'd rather do it in order of the fields as they appear on the struct so if another field gets added its idiomatic where to put the parsing logic.  This code path is not in the hot path as its only a single time at process instantiation time.

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+	return nil
+}
+
+// AggregatedClusterNamespaceDefinition is a definition for an
+// cluster namespace that holds aggregated metrics data at a
+// specific retention and resolution.
+type AggregatedClusterNamespaceDefinition struct {
+	NamespaceID ident.ID
+	Session     client.Session
+	Retention   time.Duration
```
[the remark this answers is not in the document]
robskillington: I'd rather keep it as a separate field, just to make it more similar to `UnaggregatedClusterNamespaceDefinition` struct, just with the addition of the `Resolution` field.  It would be over-de-normalization for this case I believe, I'd prefer to stick with idiomatic.

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+	unaggregatedClusterNamespace UnaggregatedClusterNamespaceDefinition,
+	aggregatedClusterNamespaces []AggregatedClusterNamespaceDefinition,
+) (Clusters, error) {
+	expectedAggregated := len(aggregatedClusterNamespaces)
+	expectedAll := 1 + expectedAggregated
+	namespaces := make([]ClusterNamespace, 0, expectedAll)
+	aggregatedNamespaces := make(map[RetentionResolution]ClusterNamespace,
+		expectedAggregated)
+
+	def := unaggregatedClusterNamespace
```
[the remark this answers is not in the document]
robskillington: I wanted to keep the line below less than 80chars long (or at the very least, relatively short).  This is a pattern I use often to keep lines short, waste a line but keep readability with shorter lines of code.

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+	unaggregatedNamespace, err := newUnaggregatedClusterNamespace(def)
+	if err != nil {
+		return nil, err
+	}
+
```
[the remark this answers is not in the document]
robskillington: I like spacing =]   To keep code together I removed the endline from under `namespaces := ...` to stop it looking like it was drifting by itself.

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+		if err != nil {
+			return nil, err
+		}
+
+		namespaces = append(namespaces, namespace)
+		key := RetentionResolution{
+			Retention:  namespace.Attributes().Retention,
+			Resolution: namespace.Attributes().Resolution,
+		}
+		aggregatedNamespaces[key] = namespace
```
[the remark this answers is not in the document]
robskillington: Hm yes, good call.  Let me add that and a test.

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+func (c *clusters) AggregatedClusterNamespace(
+	attrs RetentionResolution,
+) (ClusterNamespace, bool) {
+	namespace, ok := c.aggregatedNamespaces[attrs]
+	return namespace, ok
+}
+
+type syncMultiErrs struct {
```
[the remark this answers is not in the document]
robskillington: Good call, will move to the bottom of this file.

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+		go func() {
+			defer wg.Done()
+			err := session.Close()
+			syncMultiErrs.add(err)
+		}()
+	}
+
+	wg.Wait()
+
+	return syncMultiErrs.finalError()
```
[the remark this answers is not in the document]
robskillington: I'm just wrapping up the `MultiError` type with a lock here, I would like to defer logic to formatting, etc to the underlying `MultiError` type and leave that unchanged with these changes.

src/coordinator/storage/local/cluster_test.go
robskillington: Sounds good.

src/coordinator/storage/local/cluster_test.go
robskillington: Sure thing.

src/coordinator/storage/local/cluster_test.go
robskillington: `assert.Equal(...)` does deep reflect type equal checks, I explicitly want a pointer check here.  (The complex reflect type equality check breaks sometimes depending on mock state, etc)

src/coordinator/storage/local/storage.go
robskillington: The problem is that the following packages are all imported:

src/coordinator/test/local/storage.go
robskillington: Good catch, will update.

src/coordinator/test/local/storage.go
robskillington: Sure thing.

src/coordinator/storage/local/config.go
robskillington: I feel its more native to always `wg.Add(...)` right above an operation, that way if you remove an operation, you remove the `wg.Add(...)` right next to it and don't need to jump around the code (less mistakes in general I've found).

src/coordinator/storage/local/config.go
robskillington: Sure thing.

src/coordinator/storage/local/config.go
robskillington: There is no such thing as a cluster ID, but I can put the namespace (although if you share namespaces across different clusters you still can't differentiate it), so I'll add the namespace next to number.

src/coordinator/benchmark/read/main/read_benchmark.go
```diff
@@ -134,10 +134,14 @@ func benchmarkCoordinator(start, end time.Time) {
...
 	var cfg config.Configuration
 	if err := xconfig.LoadFile(&cfg, m3dbClientCfg, configLoadOpts); err != nil {
-		log.Fatalf("Unable to load %s: %v", m3dbClientCfg, err)
+		log.Fatalf("unable to load %s: %v", m3dbClientCfg, err)
 	}
-	m3dbClientOpts := cfg.DBClient
-	m3dbClient, err := m3dbClientOpts.NewClient(client.ConfigurationParameters{}, func(v client.Options) client.Options {
+
+	if len(cfg.Clusters) != 1 {
+		log.Fatal("invalid config, expected singler cluster definition")
```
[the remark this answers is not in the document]
arnikola: Fair call, erroring out early probably better here

src/coordinator/storage/local/storage.go
robskillington: Sure thing.

src/coordinator/storage/local/storage.go
robskillington: Hm, this was here before I believe?

src/coordinator/storage/local/storage.go
robskillington: Opted for `// Already exists and resolution of result we are adding is not as precise`.  The "higher" or "lower" is confusing with resolution, so I went with "precise" terminology.

src/coordinator/storage/local/storage.go
robskillington: Opted for // Already exists and resolution of result we are adding is not as precise. The "higher" or "lower" is confusing with resolution, so I went with "precise" terminology.

src/coordinator/storage/local/storage_test.go
robskillington: Yeah, that's why I prefer the Contains... check.

src/coordinator/storage/local/storage_test.go
robskillington: Sure thing.

src/coordinator/storage/local/storage_test.go
robskillington: `require.Len(...)` does some weird things including spewing out weird output when the length is wrong rather than just "2 != 3", so I always avoid it now...

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+// UnaggregatedClusterNamespaceDefinition is a definition for an
+// cluster namespace that holds unaggregated metrics data.
+type UnaggregatedClusterNamespaceDefinition struct {
+	NamespaceID ident.ID
+	Session     client.Session
+	Retention   time.Duration
+}
+
+// Validate will validate the cluster namespace definition.
+func (def UnaggregatedClusterNamespaceDefinition) Validate() error {
```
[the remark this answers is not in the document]
robskillington: Hm so the whole point I have this as a struct, is so you can `UnaggregatedClusterNamespaceDefinition{NamespaceID: "foo", Retention: ...}` etc to see the names for the parameters you're specifying.

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,285 @@
...
+}
+
+type clusters struct {
+	namespaces            []ClusterNamespace
+	unaggregatedNamespace ClusterNamespace
+	aggregatedNamespaces  map[RetentionResolution]ClusterNamespace
+}
+
+// NewClusters instantiates a new Clusters instance.
+func NewClusters(
```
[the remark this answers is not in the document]
robskillington: If you have multiple unaggregated clusters or multiple aggregated namespaces for same retention/resolution it's not clear how they are load balanced and which clusters/namespaces cover which parts of the metrics space.

src/coordinator/storage/local/storage.go
robskillington: I want to include the case where query.Start matches the clusterStart however, which that logic would not cover.

src/coordinator/storage/local/storage.go
robskillington: I'd have to create temporary objects (i.e. allocation per fanout thing) which is costly, that's why I use a single static waitgroup and only if there are errors does it allocate a per fanout element.

src/coordinator/storage/local/storage.go
robskillington: I want to include the case where query.Start matches the clusterStart however, which that logic would not cover.

src/coordinator/storage/local/storage.go
robskillington: I'd have to create temporary objects (i.e. allocation per fanout thing) to pass to ExecuteParallel which is costly, that's why I use a single static waitgroup and only if there are errors does it allocate a per fanout element.

src/coordinator/api/v1/handler/database/create.go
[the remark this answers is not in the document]
arnikola: Yeah, just it's a bit weird calling it `Recommended` since there's an [implication](https://res.cloudinary.com/teepublic/image/private/s--FO3xh14a--/t_Preview/b_rgb:0195c3,c_limit,f_jpg,h_630,q_90,w_630/v1491907970/production/designs/1425101_1.jpg) that you can ignore the recommendation and override it?

Might be getting too pedantic about it though, take it or leave it

src/coordinator/storage/local/cluster.go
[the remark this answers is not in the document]
arnikola: I see what you mean, but figured it might be useful to use it here since you're using RetentionResolution as the key for the deduping map?

src/coordinator/storage/local/cluster.go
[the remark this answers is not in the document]
arnikola: I should have expected 80char limit from someone who's gone through Zoltan's classes 😛

src/coordinator/storage/local/storage.go
arnikola: Oops, missed the coordinator/errors package import

src/coordinator/storage/local/config.go
```diff
@@ -186,5 +190,5 @@ func (c ClustersStaticConfiguration) NewClusters() (Clusters, error) {
 	}
 
 	return NewClusters(unaggregatedClusterNamespace,
-		aggregatedClusterNamespaces)
+		aggregatedClusterNamespaces...)
```
arnikola: nit: Since we're splitting newlines here anyway, might look better as 
```
NewClusters(
 unaggregatedClusterNamespace,
 aggregatedClusterNamespaces...,
)
```
?

src/coordinator/storage/local/cluster.go
```diff
@@ -127,7 +128,7 @@ type clusters struct {
 // NewClusters instantiates a new Clusters instance.
 func NewClusters(
 	unaggregatedClusterNamespace UnaggregatedClusterNamespaceDefinition,
-	aggregatedClusterNamespaces []AggregatedClusterNamespaceDefinition,
+	aggregatedClusterNamespaces ...AggregatedClusterNamespaceDefinition,
```
arnikola: What's the difference between using vararg vs the array here?

src/coordinator/storage/local/storage_test.go
arnikola: Oh damn, been telling people to use `X.Len` instead of `Equal` in reviews... 👀

src/coordinator/storage/local/storage_test.go
arnikola: In this case, could it be worth at some point pulling out a bunch of test utilities like this into a custom  module to allow us to reuse them?

src/coordinator/storage/local/cluster.go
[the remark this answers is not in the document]
arnikola: I meant more for the formatting of the error message, fair to defer it to how `MultiError` does it

src/coordinator/storage/local/config.go
arnikola: Ah that's unfortunate. Any way we can specify which failed to connect if we've deduped any?

src/coordinator/storage/local/cluster.go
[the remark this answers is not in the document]
arnikola: That's fair, me too ;)

I think general rule of thumb we've been using is mostly keeping newlines after the end bracket, but it's superceded by `what looks better` 😛

src/coordinator/models/tag.go
```diff
@@ -125,25 +125,31 @@ func (m Matchers) ToTags() (Tags, error) {
 	return tags, nil
 }
 
+var (
```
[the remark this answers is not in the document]
robskillington: Done and done.

src/coordinator/api/v1/handler/database/create.go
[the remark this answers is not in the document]
robskillington: You can have smaller block sizes, you just have to specify it explicitly.  Well actually no that's not true, I'll add it to the protobuf so you can use either the expected datapoints per block or an explicit block size.

src/coordinator/models/tag.go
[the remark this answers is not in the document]
nikunjgit: did you rebase ?

src/coordinator/storage/local/storage.go
nikunjgit: I see, so its cheaper to do everything in parallel instead of failing on first error and needing allocations ? I suppose I can change the other functions later then to stop using ExecuteParallel

src/coordinator/storage/local/storage.go
nikunjgit: yeah I missed the part of it creating the dedup map with previous result.

src/coordinator/storage/local/cluster.go
```diff
@@ -0,0 +1,293 @@
...
+
+type syncMultiErrs struct {
+	sync.Mutex
+	multiErr xerrors.MultiError
+}
+
+func (errs *syncMultiErrs) add(err error) {
+	errs.Lock()
+	errs.multiErr = errs.multiErr.Add(err)
+	errs.Unlock()
```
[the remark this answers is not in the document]
robskillington: Yeah no big reason, if it was in the hot path I'd always avoid defer since there is some overhead there.  Since I was returning a value in the other one it is slightly easier to write, the defer way.  I'll make these both avoid defer though, just in case this does get used by someone else in a hot path.

src/coordinator/storage/local/storage_test.go
robskillington: Possibly, although it's pretty simply so I'm fine to also just rely on the standard library and `stretchr` without adding another package into the mix for now.

src/coordinator/storage/local/storage_test.go
robskillington: It works fine until the test stops passing, and then you spend like a full minute or two wondering whats going on lol.

src/coordinator/storage/local/cluster.go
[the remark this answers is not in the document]
robskillington: Yeah, so I tried to separate basically the "definitions" or declarative options to pass to the `NewClusters(...)` method so you kind of build a declarative tree of what you want to ultimately construct, then `NewClusters(...)` basically turns these into concrete things that are wired up to clusters, etc.

src/coordinator/storage/local/storage.go
robskillington: So `!clusterStart.Before(query.Start) { continue }` is equivalent to `clusterStart.Equal(query.Start) || clusterStart.After(query.Start) { continue }` which is undesirable here.

src/coordinator/storage/local/storage.go
robskillington: Basically cheapest is zero allocation if possible, though this is not always possible, everywhere that is a new allocation and then shared (i.e. escapes to heap) is a place that probably should be pooled if in the hot path.

src/coordinator/storage/local/config.go
```diff
@@ -0,0 +1,194 @@
...
+	err     error
+}
+
+// NewClusters instantiates a new Clusters instance.
+func (c ClustersStaticConfiguration) NewClusters() (Clusters, error) {
+	var (
+		numUnaggregatedClusterNamespaces int
+		numAggregatedClusterNamespaces   int
+		unaggregatedClusterNamespaceCfg  = &unaggregatedClusterNamespaceConfiguration{}
+		aggregatedClusterNamespacesCfgs  []*aggregatedClusterNamespacesConfiguration
```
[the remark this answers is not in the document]
robskillington: So the main reason I'm using pointers here is that I need to set the client session and error back on the struct itself once I try connecting to the cluster.  It seemed simpler just to use a pointer, and since this is a slow path (i.e. once at initialization) I didn't think too much about it.

src/coordinator/storage/local/cluster.go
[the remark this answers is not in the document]
robskillington: Opened issue here:

PR #787, opened 2018-07-10
Clamp operations for query engine

nikunjgit:


src/coordinator/functions/clamp.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return err
+	}
+
+	stepIter := b.StepIter()
+	if err := builder.AddCols(b.StepCount()); err != nil {
+		return err
+	}
+
+	fn := math.Min
```
arnikola: Instead of doing this here, would it make more sense to pass the relevant function into ClampNode rather than the ClampOp?

src/coordinator/functions/clamp.go
```diff
@@ -0,0 +1,114 @@
...
+	if c.op.optype == ClampMinType {
+		fn = math.Max
+	}
+
+	for index := 0; stepIter.Next(); index++ {
+		step := stepIter.Current()
+		values := step.Values()
+		for _, value := range values {
+			builder.AppendValue(index, fn(value, c.op.scalar))
```
arnikola: nit: pull out c.op.scalar to before the` for` loop?

PR #788, opened 2018-07-10
Make commit log reader re-usable.

richardartoul:


src/dbnode/persist/fs/commitlog/iterator.go
```diff
@@ -198,7 +204,7 @@ func (i *iterator) nextReader() bool {
 		return false
 	}
 
-	i.reader = reader
+	i.readerIsClosed = false
```
robskillington: Shouldn't this be false as soon as the reader is successfully open?

src/dbnode/persist/fs/commitlog/read_write_prop_test.go
```diff
@@ -51,6 +52,9 @@ type seriesWritesAndReadPosition struct {
 }
 
 func TestCommitLogReadWrite(t *testing.T) {
+	// Make sure we're not leaking goroutines
+	defer leaktest.CheckTimeout(t, time.Second)()
```
robskillington: This is a common fault, unfortunately you will capture the start of the leaktest at the end of the test.

src/dbnode/persist/fs/commitlog/read_write_prop_test.go
```diff
@@ -120,6 +124,9 @@ func TestCommitLogReadWrite(t *testing.T) {
 }
 
 func TestCommitLogPropTest(t *testing.T) {
+	// Make sure we're not leaking goroutines
+	defer leaktest.CheckTimeout(t, 30*time.Second)()
```
robskillington: Same here with this leak test check.

src/dbnode/persist/fs/commitlog/reader.go
```diff
@@ -92,49 +94,38 @@ type decoderArg struct {
...
+	numCompletedDecoders int64
+	wg                   *sync.WaitGroup
 }
 
 type reader struct {
+	// The commit log reader is not goroutine safe, but we use an embedded
+	// mutex to defensively protect the Open/Close methods against misuse.
+	// We don't protect the Read() method with the mutex because it would
+	// be too expensive.
+	sync.Mutex
```
robskillington: Could you perhaps use a sub-struct for this perhaps?

robskillington: Also nit, we use `closed` everywhere so to be consistent perhaps use that?  And if you're truly against that, removing the `is...` prefix would be good as we try to avoid that.

src/dbnode/persist/fs/commitlog/reader.go
```diff
@@ -143,26 +134,24 @@ func newCommitLogReader(opts Options, seriesPredicate SeriesFilterPredicate) com
...
 		infoDecoder:       msgpack.NewDecoder(decodingOpts),
 		infoDecoderStream: msgpack.NewDecoderStream(nil),
-		decoderQueues:     decoderQueues,
-		decoderBufPools:   decoderBufs,
-		outChan:           outBuf,
-		cancelCtx:         cancelCtx,
-		cancelFunc:        cancelFunc,
-		shutdownCh:        make(chan error),
-		metadata:          readerMetadata{},
+		completedDecoders: completedDecoders{wg: &sync.WaitGroup{}},
```
robskillington: Do you always initialize a waitgroup?  If so can it just be an embedded struct?  Do you need the nil-ability, etc?

src/dbnode/persist/fs/commitlog/reader.go
```diff
@@ -344,14 +350,14 @@ func (r *reader) decoderLoop(inBuf <-chan decoderArg, outBuf chan<- readResponse
 			// even though we are performing parallel decoding, the work is distributed to the workers
 			// based on the series unique index which means that all commit log entries for a given
 			// series are decoded in-order by a single decoder loop.
-			r.handleDecoderLoopIterationEnd(arg, outBuf, response, errCommitLogReaderMissingMetadata)
+			r.handleDecoderLoopIterationEnd(arg, true, response, errCommitLogReaderMissingMetadata)
 			continue
 		}
 
 		if !metadata.passedPredicate {
 			// Pass nil for outBuf because we don't want to send a readResponse along since this
```
robskillington: update comment to reflect passing false instead of nil now?

richardartoul:
Not planning on working on this anytime soon

PR #791, opened 2018-07-11
Support lazy evaluation of functions within query

nikunjgit:


src/coordinator/executor/transform/lazy.go
```diff
@@ -0,0 +1,221 @@
...
+	lazyNode       *lazyNode
+	ID             parser.NodeID
+	processedBlock block.Block
+}
+
+func (f *lazyBlock) StepIter() (block.StepIter, error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if f.processedBlock != nil {
```
arnikola: Might be better to use sync.Once() here? Does the same thing you're doing here

src/coordinator/executor/transform/lazy.go
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	return f.processedBlock.StepIter()
+}
+
+func (f *lazyBlock) SeriesIter() (block.SeriesIter, error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if f.processedBlock != nil {
```
arnikola: Same here

src/coordinator/executor/transform/types.go
```diff
@@ -33,7 +33,7 @@ type Options struct {
 	Debug    bool
 }
 
-// OpNode represents the execution node
+// OpNode represents the execution fNode
```
arnikola: nit: rename issue?

src/coordinator/executor/transform/types.go
```diff
@@ -46,3 +46,23 @@ type TimeSpec struct {
...
 	Step time.Duration
 }
+
+// Params are defined by transforms
+type Params interface {
+	parser.Params
+	Node(controller *Controller) OpNode
+}
+
+// SeriesNode is implemented by function nodes which can support series iteration
```
arnikola: Worth having a common Node interface that these both implement?
```golang
type Node interface{
	Meta(meta block.Metadata) block.Metadata
 	SeriesMeta(metas []block.SeriesMeta) []block.SeriesMeta
}
```

src/coordinator/functions/abs.go
```diff
@@ -59,25 +59,65 @@ type AbsNode struct {
...
+			builder.AppendValue(index, value)
 		}
 	}
 
 	nextBlock := builder.Build()
 	defer nextBlock.Close()
 	return c.controller.Process(nextBlock)
 }
+
+// Ensure AbsNode implements the types
```
arnikola: 👍

src/coordinator/executor/transform/types.go
arnikola: Also, worth having an interface that matches functions that support both?

src/coordinator/functions/abs.go
```diff
@@ -59,25 +59,65 @@ type AbsNode struct {
...
+var _ transform.StepNode = &AbsNode{}
+var _ transform.SeriesNode = &AbsNode{}
+
+// ProcessStep allows step iteration
+func (c *AbsNode) ProcessStep(step block.Step) (block.Step, error) {
+	return block.NewColStep(step.Time(), c.process(step.Values())), nil
```
arnikola: can we pull out `processedValues := c.process(step.Values())`, to make it very clear what we're doing here?

src/coordinator/functions/abs.go
```diff
@@ -59,25 +59,65 @@ type AbsNode struct {
...
+}
+
+// ProcessSeries allows series iteration
+func (c *AbsNode) ProcessSeries(series block.Series) (block.Series, error) {
+	return block.NewSeries(c.process(series.Values()), series.Meta), nil
```
arnikola: Same here

src/coordinator/functions/clamp.go
```diff
@@ -93,19 +93,27 @@ type ClampNode struct {
 
 // Process the block
 func (c *ClampNode) Process(ID parser.NodeID, b block.Block) error {
```
arnikola: Should clamp node be transformed into a lazy function?

src/coordinator/test/handler.go
```diff
@@ -21,3 +29,32 @@ func (h *SlowHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	time.Sleep(h.delay)
 	h.handler.ServeHTTP(w, r)
 }
+
+// GeneratePromReadRequest generates a sample prometheus remote read request
```
arnikola: Since we only use this in one test package, may be better to put them there?

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -214,24 +254,45 @@ func sortedBlocksToSeriesList(blockList []block.Block) ([]*ts.Series, error) {
...
-		return nil, fmt.Errorf("mismatch in number of steps for the block, wanted: %d, found: %d", firstBlock.StepCount(), b.StepCount())
+	blockStepCount := blockStepIter.StepCount()
+
+	if stepCount != blockStepCount {
+		return nil, fmt.Errorf("mismatch in number of steps for the block, wanted: %d, found: %d", stepCount, blockStepCount)
 	}
 
-	index := sort.Search(len(blockList), func(i int) bool { return blockList[i].Meta().Bounds.Start.Before(b.Meta().Bounds.Start) })
-	blockList = append(blockList, b)
+	index := sort.Search(len(blockList), func(i int) bool { return blockList[i].meta.Bounds.Start.Before(blockMeta.Bounds.Start) })
```
benraskin92: Add a comment here

src/coordinator/executor/state.go
```diff
@@ -48,15 +48,19 @@ func CreateSource(ID parser.NodeID, params SourceParams, storage storage.Storage
...
 	controller := &transform.Controller{ID: ID}
-	return params.Node(controller), controller
-}
+	node := params.Node(controller)
 
-// TransformParams are defined by transforms
-type TransformParams interface {
-	parser.Params
-	Node(controller *transform.Controller) transform.OpNode
+	if _, ok := node.(transform.SeriesNode); ok {
```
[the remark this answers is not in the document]
benraskin92: +1

src/coordinator/functions/abs.go
benraskin92: +1

src/coordinator/api/v1/handler/prometheus/native/read.go
```diff
@@ -130,15 +130,34 @@ func (h *PromReadHandler) read(reqCtx context.Context, w http.ResponseWriter, pa
...
+		var numSteps, numSeries int
 		// TODO(nikunj): Stream blocks to client
 		for blkResult := range resultChan {
 			if blkResult.Err != nil {
 				processErr = blkResult.Err
 				break
 			}
 
+			b := blkResult.Block
+			if !firstElement {
```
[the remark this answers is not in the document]
nikunjgit: as we discussed, don't think range is parallel. Also, it should be checking all errors returned by downstream and breaking. Do you see anywhere it's not ?

src/coordinator/executor/transform/lazy.go
```diff
@@ -0,0 +1,221 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Node -> Sum -> Sink
```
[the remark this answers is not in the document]
nikunjgit: oops

src/coordinator/executor/transform/lazy.go
nikunjgit: oh, I don't actually want to process the block, just want to check if the block has been processed or not. I think sync.Once() would be more useful if I wanted the other behavior

src/coordinator/executor/transform/lazy.go
nikunjgit: same reply as above

src/coordinator/executor/transform/types.go
nikunjgit: common interface makes sense. I thought about the third interface matching both but didn't really find much uses for it. Might revisit later if we need it.

src/coordinator/functions/abs.go
nikunjgit: yea there wasn't any other enforcer here

src/coordinator/functions/clamp.go
nikunjgit: yea planning in another diff.

nikunjgit: added a TODO

src/coordinator/test/handler.go
nikunjgit: oh yeah, I changed some tests later back.

src/coordinator/block/types.go
```diff
@@ -92,13 +84,24 @@ type Iterator interface {
...
+	SeriesMeta() []SeriesMeta
+	// Meta returns the metadata for the block
+	Meta() Metadata
 }
 
 // StepIter iterates through a block vertically
 type StepIter interface {
 	Iterator
-	Current() Step
+	Current() (Step, error)
```
arnikola: nit: comments on these

src/coordinator/executor/transform/types.go
```diff
@@ -53,16 +53,20 @@ type Params interface {
...
+	MetaNode
 	ProcessSeries(series block.Series) (block.Series, error)
-	Meta(meta block.Metadata) block.Metadata
-	SeriesMeta(metas []block.SeriesMeta) []block.SeriesMeta
 }
 
 // StepNode is implemented by function nodes which can support step iteration
 type StepNode interface {
+	MetaNode
 	ProcessStep(step block.Step) (block.Step, error)
```
arnikola: nit: add comments here if we're sticking to putting them on the interfaces?

PR #792, opened 2018-07-12
Add m3coordinator Dockerfile and update Prometheus test to use standalone coordinator

robskillington:
This will also fix our documentation, as we'll be able to start building each service with quay.io on each push to master.


docker/m3coordinator/Dockerfile
```diff
@@ -0,0 +1,27 @@
+# stage 1: build
+FROM golang:1.10-alpine AS builder
+LABEL maintainer="The M3DB Authors <m3db@googlegroups.com>"
+
+# Install Glide
+RUN apk add --update glide git make bash
```
robskillington: We should reuse these steps across the Dockerfiles at some point, but let's just move forward quickly for now to unblock others using the images?

src/coordinator/api/v1/httpd/handler_test.go
```diff
@@ -98,3 +100,35 @@ func TestPromNativeReadPost(t *testing.T) {
 	h.Router.ServeHTTP(res, req)
 	require.Equal(t, res.Code, http.StatusMethodNotAllowed, "POST method not defined")
 }
+
+func TestRooutesGet(t *testing.T) {
```
schallert: `s/Rooutes/Routes`

src/coordinator/services/m3coordinator/server/server.go
```diff
@@ -159,6 +174,9 @@ func Run(runOpts RunOptions) {
 		}
 	}
 
+	namespaces := clusters.ClusterNamespaces()
+	logger.Info("resolved cluster namespaces", zap.Int("namespaces", len(namespaces)))
```
schallert: Would it be useful to list the names instead of the count?

src/coordinator/api/v1/httpd/handler_test.go
robskillington: Ta, good catch.

src/coordinator/services/m3coordinator/server/server.go
robskillington: Sure thing.

scripts/integration-tests/prometheus/m3coordinator.yml
```diff
@@ -0,0 +1,45 @@
+listenAddress: 0.0.0.0:7201
```
nikunjgit: Is this never going to be useful for end users ? Maybe consider it moving out of integration-tests ?

PR #793, opened 2018-07-13
Add logical OR and logical UNLESS functions

arnikola:


src/coordinator/functions/logical/or.go
```diff
@@ -0,0 +1,123 @@
...
+		for _, value := range lValues {
+			builder.AppendValue(index, value)
+		}
+	}
+
+	index := 0
+	for _, missingIndex := range missingIndices {
+		for ; index != missingIndex; index++ {
+			if !rIter.Next() {
+				return nil, fmt.Errorf("missing index not found in rhs")
```
benraskin92: nit: `errors.New("missing index not found in rhs")` or maybe even put it up top as a `var`

src/coordinator/functions/logical/or_test.go
```diff
@@ -0,0 +1,76 @@
...
+	node := op.Node(c)
+
+	err := node.Process(parser.NodeID(1), block2)
+	require.NoError(t, err)
+	err = node.Process(parser.NodeID(0), block1)
+	require.NoError(t, err)
+	// NAN values should be filled
+	expected := values1
+
+	fmt.Println(sink.Values)
```
benraskin92: remove print

src/coordinator/functions/logical/or.go
```diff
@@ -0,0 +1,123 @@
...
+func NewOrNode(op BaseOp, controller *transform.Controller) Processor {
+	return &OrNode{
+		op:         op,
+		controller: controller,
+	}
+}
+
+// Process processes two logical blocks, performing Or operation on them
+func (c *OrNode) Process(lhs, rhs block.Block) (block.Block, error) {
+	builder, err := c.controller.BlockBuilder(lhs.Meta(), lhs.SeriesMeta())
```
nikunjgit: this probably needs to change since you can't simply use lhs.SeriesMeta and lhs.Meta

src/coordinator/functions/logical/or.go
```diff
@@ -0,0 +1,123 @@
...
+	for index := 0; lIter.Next(); index++ {
+		lStep := lIter.Current()
+		lValues := lStep.Values()
+		for _, value := range lValues {
+			builder.AppendValue(index, value)
+		}
+	}
+
+	index := 0
+	for _, missingIndex := range missingIndices {
```
nikunjgit: i think this can be simplified a bit:
```
for index:=0; rIter.Next();{
rStep := rIter.Current()
rValues := rStep.Values()
for _, idx := range missingIndices {
 builder.AppendValue(index, rValues[idx])
}
}
```

src/coordinator/functions/logical/or.go
arnikola: Oops, didn't update

src/coordinator/functions/logical/or.go
arnikola: I think that this would result in different behaviour to what I'm looking for though; this would take the `missingIndex`-th values from each array in the iterator instead

arnikola:
I've made some logic changes, verified it vs prom and these functions match up.

TODO would be to refactor and consolidate the codepaths; may leave that for a future commit however.

src/coordinator/functions/logical/or.go
arnikola: Nah my bad your suggestion is correct, got confused by column builder stuff haha

nikunjgit:
Might be worth doing the refactor now instead of punting it to the next diff since otherwise you have a lot of code duplication. Also, no one is really blocked on it for now so should be fine to wait a bit.

src/coordinator/functions/logical/common.go
```diff
@@ -66,10 +67,55 @@ func hashFunc(on bool, names ...string) func(models.Tags) uint64 {
...
+		}
+		l.Tags[k] = v
+	}
+
+	return l, nil
 }
 
-// MakeProcessor is a way to create a logical transform
-type MakeProcessor func (op BaseOp, controller *transform.Controller) Processor
+func addValuesAtIndeces(idxArray []int, iter block.StepIter, builder block.Builder) error {
```
nikunjgit: appendValuesAtIndices ?

src/coordinator/functions/logical/and.go
```diff
@@ -24,77 +24,35 @@ import (
...
-	lIter, err := lhs.StepIter()
-	if err != nil {
-		return nil, err
-	}
-
-	rIter, err := rhs.StepIter()
-	if err != nil {
-		return nil, err
-	}
+func makeAndBuilder(
```
nikunjgit: name is misleading since it's not really a builder. Maybe call it makeAndBlock ?

src/coordinator/functions/logical/common.go
```diff
@@ -66,10 +67,55 @@ func hashFunc(on bool, names ...string) func(models.Tags) uint64 {
...
+
+func combineMetadata(l, r block.Metadata) (block.Metadata, error) {
+	if !l.Bounds.Equals(r.Bounds) {
+		return block.Metadata{}, errMismatchedBounds
+	}
+
+	for k, v := range r.Tags {
+		if lVal, ok := l.Tags[k]; ok {
+			if lVal != v {
+				return block.Metadata{}, errConflictingTags
```
nikunjgit: what about tags which are in l but not in r ? Maybe add a comment to the function since its not clear what we're doing.

src/coordinator/parser/promql/types.go
```diff
@@ -71,9 +71,16 @@ func NewOperator(opType promql.ItemType) (parser.Params, error) {
 
 // NewBinaryOperator creates a new binary operator based on the type
 func NewBinaryOperator(expr *promql.BinaryExpr, lhs, rhs parser.NodeID) (parser.Params, error) {
-	switch getOpType(expr.Op) {
+	matching := promMatchingToM3(expr.VectorMatching)
+
+	op := getOpType(expr.Op)
+	switch op {
 	case logical.AndType:
```
nikunjgit: how about 

case logical.AndType, logical.OrType, logical.UnlessType:

src/coordinator/functions/logical/or.go
```diff
@@ -0,0 +1,88 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package logical
+
+import (
+	"github.com/m3db/m3db/src/coordinator/block"
+)
+
+func makeOrBuilder(
```
nikunjgit: makeOrBlock ?

src/coordinator/functions/logical/common.go
```diff
@@ -66,10 +67,55 @@ func hashFunc(on bool, names ...string) func(models.Tags) uint64 {
 	return func(tags models.Tags) uint64 { return tags.IDWithExcludes(names...) }
 }
 
-// Processor is implemented by each logical transform
-type Processor interface {
-	Process(lhs block.Block, rhs block.Block) (block.Block, error)
+const (
```
nikunjgit: is it better to have one common place for all types or to have each type defined in their file with their implementation ? A few advantages of having them with their implementation would be that you wouldn't have one file where the mapping keeps growing. Also, you wouldn't have to change this file everytime you add a new type

src/coordinator/functions/logical/or.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	if err := addValuesAtIndeces(missingIndices, rIter, builder); err != nil {
+		return nil, err
+	}
+
+	return builder.Build(), nil
+}
+
+// missing returns the slice of rhs indices for which there are no corresponding
```
nikunjgit: update comments ? Since now missing is returning the combined series meta as well

src/coordinator/functions/logical/unless.go
```diff
@@ -0,0 +1,89 @@
...
+
+package logical
+
+import (
+	"sort"
+
+	"github.com/m3db/m3db/src/coordinator/block"
+)
+
+func makeUnlessBuilder(
```
nikunjgit: makeUnlessBlock ?

src/coordinator/functions/logical/unless.go
```diff
@@ -0,0 +1,89 @@
...
+	logicalNode *BaseNode,
+	lIter, rIter block.StepIter,
+) (block.Block, error) {
+	lSeriesMeta, rSeriesMeta := lIter.SeriesMeta(), rIter.SeriesMeta()
+	lIds := exclusion(logicalNode.op.Matching, lSeriesMeta, rSeriesMeta)
+	stepCount := len(lIds)
+	takenMeta := make([]block.SeriesMeta, 0, stepCount)
```
nikunjgit: takenMeta is slightly confusing.

src/coordinator/functions/logical/unless.go
```diff
@@ -0,0 +1,89 @@
...
+
+	"github.com/m3db/m3db/src/coordinator/block"
+)
+
+func makeUnlessBuilder(
+	logicalNode *BaseNode,
+	lIter, rIter block.StepIter,
+) (block.Block, error) {
+	lSeriesMeta, rSeriesMeta := lIter.SeriesMeta(), rIter.SeriesMeta()
+	lIds := exclusion(logicalNode.op.Matching, lSeriesMeta, rSeriesMeta)
```
nikunjgit: instead of exclusion maybe call it distinctLeft ?

src/coordinator/functions/logical/or.go
```diff
@@ -0,0 +1,88 @@
...
+
+func makeOrBuilder(
+	logicalNode *BaseNode,
+	lIter, rIter block.StepIter,
+) (block.Block, error) {
+	meta, err := combineMetadata(lIter.Meta(), rIter.Meta())
+	if err != nil {
+		return nil, err
+	}
+	missingIndices, combinedSeriesMeta := missing(
```
nikunjgit: if combineMetadata produces a different ordering than missing, then this will break. I wonder how we can enforce the same ordering for both ? Maybe consider doing all this in a single function

src/coordinator/parser/promql/types.go
arnikola: Nice, didn't know we could do this

src/coordinator/functions/logical/and.go
arnikola: Will do, this was a holdover from different approach which I didn't update

src/coordinator/functions/logical/common.go
arnikola: Since it's returning `l` at the end, all tags which are in l will be present already; I'll add this comment

src/coordinator/functions/logical/unless.go
arnikola: Holdover from old implementation, fixed

src/coordinator/functions/logical/unless.go
arnikola: Sounds good, was a holdover from when I misunderstood prom's logic as being `XOR` rather than `XOR but only for the left`

src/coordinator/functions/logical/unless.go
arnikola: renamed `distinctSeriesMeta`

src/coordinator/functions/logical/common.go
arnikola: This file is only for these logical functions, so it's unlikely that it will grow, but agreed that it makes sense to keep these in their own function definition file

src/coordinator/functions/logical/common.go
```diff
@@ -66,10 +67,55 @@ func hashFunc(on bool, names ...string) func(models.Tags) uint64 {
...
+	OrType = "or"
+
+	// UnlessType uses all values from lhs which do not exist in rhs
+	UnlessType = "unless"
+
+	initIndexSliceLength = 10
+)
+
+var (
+	errMismatchedBounds     = fmt.Errorf("block bounds are mismatched")
```
benraskin92: nit: `errors.New(...)`

src/coordinator/functions/logical/common_test.go
```diff
@@ -0,0 +1,180 @@
...
+		stepIter.EXPECT().Current().Return(bl, nil)
+	}
+	stepIter.EXPECT().Next().Return(false)
+
+	return stepIter
+}
+
+var addAtIndicesTests = []struct {
+	name                          string
+	indeces, expectedIndeces      []int
```
benraskin92: nit: `s/indeces/indices`, etc.

arnikola: Damn, keep making this mistake

src/coordinator/functions/logical/common.go
```diff
@@ -66,10 +67,82 @@ func hashFunc(on bool, names ...string) func(models.Tags) uint64 {
...
+
+	return meta,
+		seriesMeta,
+		otherSeriesMeta,
+		nil
 }
 
-// MakeProcessor is a way to create a logical transform
-type MakeProcessor func (op BaseOp, controller *transform.Controller) Processor
+func appendValuesAtIndeces(idxArray []int, iter block.StepIter, builder block.Builder) error {
```
nikunjgit: typo ? appendValuesAtIndices

src/coordinator/functions/logical/common.go
```diff
@@ -66,10 +67,82 @@ func hashFunc(on bool, names ...string) func(models.Tags) uint64 {
...
+	index := 0
+	for ; iter.Next(); index++ {
```
nikunjgit: why not initialize index here ?

src/coordinator/functions/logical/common.go
```diff
@@ -66,10 +67,82 @@ func hashFunc(on bool, names ...string) func(models.Tags) uint64 {
...
+	seriesMeta, otherSeriesMeta []block.SeriesMeta,
+) (block.Metadata, []block.SeriesMeta, []block.SeriesMeta, error) {
+	if !meta.Bounds.Equals(otherMeta.Bounds) {
+		return block.Metadata{},
+			[]block.SeriesMeta{},
+			[]block.SeriesMeta{},
+			errMismatchedBounds
+	}
+
+	// NB (arnikola): mutating tags in `meta` to avoid allocations
```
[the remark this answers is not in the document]
arnikola: I'm not sure it would be getting shared? Only one process should be using the block at a time, unless I misunderstand how we're processing downstream functions

arnikola:
Ran comparison on local prom, results are matching

src/coordinator/functions/logical/common.go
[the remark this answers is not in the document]
nikunjgit: If a child op has two parents ?

PR #802, opened 2018-07-23
Delete commit log files that are covered by snapshot files

richardartoul:
- [X] Refactors how we list / interact with the commit log files on disk (specifically, we no longer use the name of the file to determine the start time nor do we rely on the current configuration to guess what the commit log block size is)
- [X] Rewrite commit log cleanup logic to allow commit logs to be deleted if they're captured by snapshots
- [X] In addition to the standard unit tests that would be expected with this P.R, I've added a mixed mode read write integration prop test that uses a random seed and property testing to verify that the cleanup logic always leave the node in a state where if its restarted all data can be recovered.

### DIFF
--- a/src/dbnode/integration/disk_cleanup_multi_ns_test.go
+++ b/src/dbnode/integration/disk_cleanup_multi_ns_test.go
@@ -145,7 +145,7 @@ func TestDiskCleanupMultipleNamespace(t *testing.T) {
 			testSetup,
 			testSetup.storageOpts.CommitLogOptions().SetFlushInterval(defaultIntegrationTestFlushInterval),
 			data,
-			ns1.ID(),
+			ns1,
 			clTime,
 			false,
 		)
--- a/src/dbnode/integration/disk_cleanup_multi_ns_test.go
+++ b/src/dbnode/integration/disk_cleanup_multi_ns_test.go
@@ -26,8 +26,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/m3db/m3db/src/dbnode/integration/generate"
 	"github.com/m3db/m3db/src/dbnode/retention"
 	"github.com/m3db/m3db/src/dbnode/storage/namespace"
+	xtime "github.com/m3db/m3x/time"
 
 	"github.com/stretchr/testify/require"
 )
@@ -133,7 +135,20 @@ func TestDiskCleanupMultipleNamespace(t *testing.T) {
 
 	log.Infof("creating commit log and fileset files")
 	shard := uint32(0)
-	writeCommitLogs(t, filePathPrefix, commitLogTimes)
+	for _, clTime := range commitLogTimes {
+		// Need to generate valid commit log files otherwise cleanup will fail.
+		data := map[xtime.UnixNano]generate.SeriesBlock{
+			xtime.ToUnixNano(clTime): nil,
+		}
+		writeCommitLogDataSpecifiedTS(
+			t,
+			testSetup,
+			testSetup.storageOpts.CommitLogOptions().SetFlushInterval(defaultIntegrationTestFlushInterval),
+			data,
+			ns1.ID(),
+			clTime,
+		)
+	}
 	writeDataFileSetFiles(t, testSetup.storageOpts, ns1, shard, ns1Times)
 	writeDataFileSetFiles(t, testSetup.storageOpts, ns2, shard, ns2Times)
 
--- a/src/dbnode/integration/disk_cleanup_multi_ns_test.go
+++ b/src/dbnode/integration/disk_cleanup_multi_ns_test.go
@@ -147,6 +147,7 @@ func TestDiskCleanupMultipleNamespace(t *testing.T) {
 			data,
 			ns1.ID(),
 			clTime,
+			false,
 		)
 	}
 	writeDataFileSetFiles(t, testSetup.storageOpts, ns1, shard, ns1Times)
--- a/src/dbnode/integration/disk_cleanup_multi_ns_test.go
+++ b/src/dbnode/integration/disk_cleanup_multi_ns_test.go
@@ -134,21 +134,20 @@ func TestDiskCleanupMultipleNamespace(t *testing.T) {
 	}()
 
 	log.Infof("creating commit log and fileset files")
-	shard := uint32(0)
+	var (
+		shard         = uint32(0)
+		commitLogOpts = testSetup.storageOpts.CommitLogOptions().
+				SetFlushInterval(defaultIntegrationTestFlushInterval)
+	)
+
 	for _, clTime := range commitLogTimes {
 		// Need to generate valid commit log files otherwise cleanup will fail.
 		data := map[xtime.UnixNano]generate.SeriesBlock{
 			xtime.ToUnixNano(clTime): nil,
 		}
 		writeCommitLogDataSpecifiedTS(
-			t,
-			testSetup,
-			testSetup.storageOpts.CommitLogOptions().SetFlushInterval(defaultIntegrationTestFlushInterval),
-			data,
-			ns1,
-			clTime,
-			false,
-		)
+			t, testSetup, commitLogOpts, data,
+			ns1, clTime, false)
 	}
 	writeDataFileSetFiles(t, testSetup.storageOpts, ns1, shard, ns1Times)
 	writeDataFileSetFiles(t, testSetup.storageOpts, ns2, shard, ns2Times)
--- a/src/dbnode/integration/disk_cleanup_test.go
+++ b/src/dbnode/integration/disk_cleanup_test.go
@@ -60,11 +60,15 @@ func TestDiskCleanup(t *testing.T) {
 	}()
 
 	// Now create some fileset files and commit logs
-	shard := uint32(0)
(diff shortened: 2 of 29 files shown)


src/dbnode/integration/fs_commitlog_snapshot_mixed_mode_read_write_prop_test.go
```diff
@@ -0,0 +1,300 @@
...
+)
+
+// Really small block sizes make certain operations take too long
+// (I.E) with 50 hours of retention and a 10 second block size the
+// node will try to flush 10 * 6 * 60 * 50 files.
+const minBlockSize = 15 * time.Minute
+const maxBlockSize = 12 * time.Hour
+const maxPoints = 100
+const minSuccessfulTests = 8
+const maxFlushWaitTime = time.Minute
```
robskillington: nit: combine into single const block:

src/dbnode/persist/fs/commitlog/files.go
```diff
@@ -67,3 +77,35 @@ func ReadLogInfo(filePath string, opts Options) (time.Time, time.Duration, int64
 
 	return time.Unix(0, logInfo.Start), time.Duration(logInfo.Duration), logInfo.Index, decoderErr
 }
+
+// Files returns a slice of all available commit log files on disk along with
+// their associated metadata.
+func Files(opts Options) ([]File, error) {
```
robskillington: 👍

src/dbnode/persist/fs/commitlog/iterator.go
```diff
@@ -202,20 +191,14 @@ func (i *iterator) nextReader() bool {
...
+func filterFiles(opts Options, files []File, predicate FileFilterPredicate) []File {
+	filteredFiles := make([]File, 0, len(files))
 	for _, file := range files {
-		start, duration, _, err := ReadLogInfo(file, opts)
-		if err != nil {
-			multiErr = multiErr.Add(err)
-			continue
-		}
-		if predicate(file, start, duration) {
+		if predicate(file.FilePath, file.Start, file.Duration) {
```
robskillington: Would it be better to make predicate just take `commitlog.File` so it's easier to add new fields in the future?

src/dbnode/storage/cleanup.go
```diff
@@ -301,26 +297,75 @@ func (m *cleanupManager) commitLogTimes(t time.Time) (time.Time, []time.Time, er
...
+		}
+
 		for _, ns := range namespaces {
-			ropts := ns.Options().RetentionOptions()
-			start, end := commitLogNamespaceBlockTimes(t, blockSize, ropts)
-			if ns.NeedsFlush(start, end) {
+			var (
+				ropts                      = ns.Options().RetentionOptions()
+				nsBlocksStart, nsBlocksEnd = commitLogNamespaceBlockTimes(start, duration, ropts)
+				needsFlush                 = ns.NeedsFlush(nsBlocksStart, nsBlocksEnd)
```
robskillington: Hm, I wonder if we should change these type of functions to instead of taking `start, end time.Time` we should be using a single `xtime.Range` arg?

src/dbnode/storage/cleanup.go
```diff
@@ -301,26 +297,75 @@ func (m *cleanupManager) commitLogTimes(t time.Time) (time.Time, []time.Time, er
...
-		return time.Time{}, nil, err
+		return nil, err
 	}
-	cleanupTimes := filterTimes(candidateTimes, func(t time.Time) bool {
+
+	var outerErr error
+	filesToCleanup := filterCommitLogFiles(files, func(start time.Time, duration time.Duration) bool {
+		if outerErr != nil {
+			return false
+		}
```
robskillington: Is this a necessary skip?  It seems like we'll reach same result whether we skip entries after we've seen a single entry.  Just because if so, then we don't have to add test coverage for these lines if it doesn't make a difference.

src/dbnode/integration/fs_commitlog_snapshot_mixed_mode_read_write_prop_test.go
richardartoul: Done

src/dbnode/persist/fs/commitlog/iterator.go
richardartoul: good call

src/dbnode/storage/cleanup.go
richardartoul: not sure how much it buys us as all the helper functions take a start/end as well and we're not trying to calculate overlap or anything like that

src/dbnode/storage/cleanup.go
richardartoul: Yeah I guess its just the difference between first error wins / last error wins. Some of the operations are expensive in that they have to go to disk so returning early is nice in that sense too, but optimizing for the error case is kind of dumb. I'll remove it.

src/dbnode/storage/cleanup.go
robskillington: Yeah, just more that the name of the data structure represents the variable itself better and its a single var that can then be passed around.  I'm not bullish though, we can revisit this later.

src/dbnode/integration/disk_cleanup_multi_ns_test.go
```diff
@@ -133,7 +135,21 @@ func TestDiskCleanupMultipleNamespace(t *testing.T) {
...
-	writeCommitLogs(t, filePathPrefix, commitLogTimes)
+	for _, clTime := range commitLogTimes {
+		// Need to generate valid commit log files otherwise cleanup will fail.
+		data := map[xtime.UnixNano]generate.SeriesBlock{
+			xtime.ToUnixNano(clTime): nil,
+		}
+		writeCommitLogDataSpecifiedTS(
+			t,
+			testSetup,
+			testSetup.storageOpts.CommitLogOptions().SetFlushInterval(defaultIntegrationTestFlushInterval),
```
robskillington: super nit: Maybe declare these options before the `for` loop?  Just to avoid this super long line and all:

src/dbnode/integration/disk_cleanup_test.go
```diff
@@ -61,11 +64,27 @@ func TestDiskCleanup(t *testing.T) {
...
-	writeCommitLogs(t, filePathPrefix, fileTimes)
+	for _, clTime := range fileTimes {
+		// Need to generate valid commit log files otherwise cleanup will fail.
+		data := map[xtime.UnixNano]generate.SeriesBlock{
+			xtime.ToUnixNano(clTime): nil,
+		}
+		writeCommitLogDataSpecifiedTS(
+			t,
+			testSetup,
+			testSetup.storageOpts.CommitLogOptions().SetFlushInterval(defaultIntegrationTestFlushInterval),
```
robskillington: super nit: Same here.

src/dbnode/integration/integration_data_verify.go
```diff
@@ -73,7 +74,7 @@ func verifySeriesMapForRange(
 	input generate.SeriesBlock,
 	expectedDebugFilePath string,
 	actualDebugFilePath string,
-) {
+) bool {
```
robskillington: Hm this is probably a good change.  Especially when called in an async goroutine because otherwise the goroutine just dies when a `require.Foo(...)` fails.  👍

src/dbnode/storage/cleanup.go
```diff
@@ -301,26 +297,71 @@ func (m *cleanupManager) commitLogTimes(t time.Time) (time.Time, []time.Time, er
...
+
+			// Add commit log blockSize to the startTime because that is the latest
+			// system time that the commit log file could contain data for. Note that
+			// this is different than the latest datapoint timestamp that the commit
+			// log file could contain data for (because of bufferPast/bufferFuture),
+			// but the commit log files and snapshot files both deal with system time.
+			isCapturedBySnapshot, err := ns.IsCapturedBySnapshot(
+				nsBlocksStart, nsBlocksEnd, start.Add(duration))
+			if err != nil {
+				outerErr = err
```
robskillington: If you want to get the first error btw, why not make it a slice `var outerErrs []error` then `outerErrs = append(outerErrs, err)` and then `if len(outerErrs) > 0 { return outerErrs[0] }`?

robskillington: Or even just use a `multiErr` here and return all errors *shrug*.

src/dbnode/storage/cleanup.go
richardartoul: cool lets leave this out in the interest of time for now

src/dbnode/integration/disk_cleanup_multi_ns_test.go
richardartoul: sure

src/dbnode/integration/disk_cleanup_test.go
richardartoul: sure

src/dbnode/integration/integration_data_verify.go
richardartoul: yeah plus gopter require you to return (false, error) when something goes wrong. If you just panic it can't really do its thing

PR #804, opened 2018-07-23
Update OpenAPI docs

justinjc:


nerd0:
@justinjc curious if this is still a WIP

nerd0:
Ok re-running semaphore tests, don't like to approve pulls with failed tests.

src/coordinator/generated/assets/openapi/spec.yml
```diff
@@ -387,3 +417,60 @@ definitions:
...
+  Host:
+    type: "object"
+    properties:
+      id:
+        type: "string"
+      address:
+        type: "string"
+      port:
+        type: "integer"
+      isolationGroup:
```
nerd0: I wish there was a way to add a doc string to this since it might not be obvious what an `isolationGroup` means in this context.

justinjc: Agreed. Couldn't find a place for descriptions here that OpenAPI allows for. Perhaps best to augment this with other documentation as well like Github pages.

PR #808, opened 2018-07-25
Add Absent function

benraskin92:


src/coordinator/functions/linear/absent.go
```diff
@@ -0,0 +1,72 @@
...
+
+package linear
+
+import (
+	"math"
+
+	"github.com/m3db/m3db/src/coordinator/executor/transform"
+)
+
+// AbsentType returns an empty timeseries if the timeseries passed in has any elements,
```
nikunjgit: returns a series with all NaNs if the series has any non Nans in it ?

src/coordinator/functions/linear/absent.go
```diff
@@ -0,0 +1,72 @@
...
+		num = math.NaN()
+	}
+
+	for i := range values {
+		values[i] = num
+	}
+	return values
+}
+
+func isNull(vals []float64) bool {
```
nikunjgit: allNaNs ?

PR #811, opened 2018-07-26
Temporal function which can span across time boundaries

nikunjgit:


src/coordinator/block/types.go
```diff
@@ -43,36 +43,101 @@ type SeriesMeta struct {
...
+// Bounds are the time bounds, start time is inclusive but end is exclusive
 type Bounds struct {
 	Start    time.Time
-	End      time.Time
+	Duration time.Duration
 	StepSize time.Duration
 }
 
 // TimeForIndex returns the start time for a given index assuming a uniform step size
 func (b Bounds) TimeForIndex(idx int) (time.Time, error) {
```
benraskin92: Maybe this should return `(time.Time, bool)` so instead of relying on the error if it's out of bounds, we rely on a true/false

src/coordinator/block/types.go
```diff
@@ -43,36 +43,101 @@ type SeriesMeta struct {
...
+	multiplier := 1
+	if !forward {
+		multiplier = -1
+	}
+
+	blockDuration := b.Duration
+	start := b.Start.Add(blockDuration * time.Duration(n*multiplier))
+	return Bounds{
+		Start:    start,
+		Duration: blockDuration,
```
benraskin92: can just do `Duration: b.Duration`?

src/coordinator/block/types_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestBounds(t *testing.T) {
+	now := time.Now()
+	bounds := Bounds{
+		Start:    now,
+		Duration: 0,
+		StepSize: time.Minute,
+	}
+	assert.Equal(t, bounds.Steps(), 0)
+	_, err := bounds.TimeForIndex(0)
+	assert.Error(t, err, "No valid index in this block")
```
benraskin92: nit: `s/No/no`

src/coordinator/block/types_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	bounds = Bounds{
+		Start:    now,
+		Duration: time.Second,
+		StepSize: time.Minute,
+	}
+	assert.Equal(t, bounds.Steps(), 0)
+	_, err = bounds.TimeForIndex(0)
+	assert.Error(t, err, "No valid index in this block")
```
benraskin92: same as above

src/coordinator/block/types_test.go
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
benraskin92: Want to test `Next()` and/or `Previous()`

src/coordinator/executor/transform/cache.go
```diff
@@ -70,3 +71,93 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+	}
+}
+
+// Add the block to the cache, errors out if block already exists
+func (c *TimeCache) Add(key time.Time, b block.Block) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	_, ok := c.blocks[fromTime(key)]
+	if ok {
+		return errors.New("block already exists")
```
benraskin92: Does this need to error? If it's already in there, isn't that fine? If you need to, can we just return a `bool` instead?

src/coordinator/executor/transform/cache.go
```diff
@@ -70,3 +71,93 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+func (c *TimeCache) MultiGet(keys []time.Time) []block.Block {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	blks := make([]block.Block, len(keys))
+	for i, key := range keys {
+		b, ok := c.blocks[fromTime(key)]
+		if ok {
+			blks[i] = b
+		}
+
```
benraskin92: nit: newline

src/coordinator/functions/fetch.go
```diff
@@ -82,6 +91,14 @@ func (n *FetchNode) Execute(ctx context.Context) error {
 	}
 
 	for _, block := range blockResult.Blocks {
+		if n.debug {
+			// Ignore any errors
+			iter, _ := block.StepIter()
+			if iter != nil {
+				fmt.Printf("[fetch node]: meta for the block: %v\n", iter.Meta())
```
benraskin92: ???

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,339 @@
...
+type baseNode struct {
+	op            baseOp
+	controller    *transform.Controller
+	cache         *transform.TimeCache
+	processor     Processor
+	mu            sync.Mutex
+	transformOpts transform.Options
+}
+
+// Process processes a block. The processing steps are as follows:
```
benraskin92: Nice comment!

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,339 @@
...
+// 1. Figure out the maximum blocks needed for the temporal function
+// 2. For the current block, figure out whether we have enough previous blocks which can help process it
+// 3. For the blocks after current block, figure out which can be processed right now
+// 4. Process all valid blocks from #3, #4 and mark them as processed
+// 5. Run a sweep face to free up blocks which are no longer needed to be cached
+func (c *baseNode) Process(ID parser.NodeID, b block.Block) error {
```
benraskin92: Could you add some comments in the code explaining what each section is doing? I think that'd be very helpful.

src/coordinator/block/types.go
nikunjgit: b.Duration was being used twice so i just extract it out

src/coordinator/executor/transform/cache.go
nikunjgit: a block already there is probably a bug so an error helps us identify it

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,348 @@
...
+			val := series.ValueAtStep(i)
+			values = append(values, val)
+			newVal := math.NaN()
+			if desiredLength <= len(values) {
+				values = values[len(values)-desiredLength:]
+				newVal = c.processor.Process(values)
+			}
+
+			builder.AppendValue(i, newVal)
+
```
benraskin92: nit: remove newline

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,348 @@
...
+			s, err := iter.Current()
+			if err != nil {
+				return err
+			}
+
+			values = append(values, s.Values()...)
+		}
+
+		desiredLength := int(aggDuration / bounds.StepSize)
```
benraskin92: add some comments in this code

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,348 @@
...
+	for i := 0; i < len(rightBlks); i++ {
+		lStart := maxBlocks - i
+		if lStart > len(leftBlks) {
+			continue
+		}
+
+		deps := leftBlks[len(leftBlks)-lStart:]
+		deps = append(deps, rightBlks[:i]...)
+		processRequests = append(processRequests, processRequest{blk: rightBlks[i], deps: deps, bounds: bounds.Next(i + 1)})
+
```
benraskin92: nit: remove newline

src/coordinator/executor/transform/cache.go
```diff
@@ -70,3 +71,92 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+}
+
+func (c cacheTime) toTime() time.Time {
+	return time.Unix(0, int64(c))
+}
+
+// TimeCache is used to cache blocks over time. It also keeps track of blocks which have been processed
+type TimeCache struct {
+	blocks          map[cacheTime]block.Block
+	processedBlocks map[cacheTime]bool
```
arnikola: make this a map of struct{} rather than bool

src/coordinator/executor/transform/cache.go
```diff
@@ -70,3 +71,92 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+	}
+
+	c.blocks[fromTime(key)] = b
+	return nil
+}
+
+// Remove the block from the cache
+func (c *TimeCache) Remove(key time.Time) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
arnikola: Rather than incurring the cost of defer here, unlock after the delete manually

src/coordinator/executor/transform/cache.go
```diff
@@ -70,3 +71,92 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+	delete(c.blocks, fromTime(key))
+}
+
+// Get the block from the cache
+func (c *TimeCache) Get(key time.Time) (block.Block, bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
arnikola: Rather than incurring the cost of defer here, unlock after retrieving the block manually

src/coordinator/executor/transform/cache.go
```diff
@@ -70,3 +71,92 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+	b, ok := c.blocks[fromTime(key)]
+	return b, ok
+}
+
+// MultiGet retrieves multiple blocks from the cache at once
+func (c *TimeCache) MultiGet(keys []time.Time) []block.Block {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
arnikola: Rather than incurring the cost of defer here, unlock after getting the blocks manually

src/coordinator/executor/transform/cache.go
```diff
@@ -70,3 +71,92 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+		}
+	}
+
+	return blks
+}
+
+// MarkProcessed is used to mark a block as processed
+func (c *TimeCache) MarkProcessed(keys []time.Time) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
arnikola: Rather than incurring the cost of defer here, unlock manually

src/coordinator/executor/transform/cache.go
```diff
@@ -70,3 +71,92 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+	for _, key := range keys {
+		c.processedBlocks[fromTime(key)] = true
+	}
+}
+
+// Processed returns all processed block times from the cache
+func (c *TimeCache) Processed() map[time.Time]bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
arnikola: Rather than incurring the cost of defer here, unlock manually

src/coordinator/plan/physical.go
```diff
@@ -66,9 +67,36 @@ func NewPhysicalPlan(lp LogicalPlan, storage storage.Storage, params models.Requ
...
+		}
+
+		if spec.Range > maxRange {
+			maxRange = spec.Range
+		}
+	}
+
+	startShift := maxOffset + maxRange
+	// keeping end the same for now, might optimize later
+	p.TimeSpec.Start = p.TimeSpec.Start.Add(-1 * startShift)
```
arnikola: nit: `.Start.Sub(startShift)`

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,348 @@
...
+	rightRangeStart := bounds.Next(maxBlocks).Start
+	queryEndBounds := bounds.Nearest(c.transformOpts.TimeSpec.End.Add(-1 * bounds.StepSize))
+
+	if rightRangeStart.After(queryEndBounds.Start) {
+		rightRangeStart = queryEndBounds.Start
+	}
+
+	// Process left side of the range
+	leftBlks, emptyLeftBlocks := c.processLeft(b, bounds, maxBlocks, leftRangeStart)
```
arnikola: nit: this would read better if the bool returned by `processLeft` is `true` iff required blocks are present, rather than the inverse

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,348 @@
...
+}
+
+// processLeft processes the current block. For the current block, figure out whether we have enough previous blocks which can help process it
+func (c *baseNode) processLeft(b block.Block, bounds block.Bounds, maxBlocks int, leftRangeStart time.Time) ([]block.Block, bool) {
+	leftRangeTimes := make([]time.Time, 0, maxBlocks)
+	for t := leftRangeStart; t.Before(bounds.Start); t = t.Add(bounds.Duration) {
+		leftRangeTimes = append(leftRangeTimes, t)
+	}
+
+	leftBlks := c.cache.MultiGet(leftRangeTimes)
```
arnikola: Would it be better to push some logic down to `MultiGet`, where it shortcircuits out on a missing cache value, then returns the partial list and `false`, rather than doing additional processing here and in `processRight`?

src/coordinator/executor/transform/cache.go
arnikola: Alternatively, make a struct which contains a block.Block as well as a flag for `processed`, and combine these 2 maps?

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,348 @@
...
+	for t := bounds.End(); !t.After(rightRangeStart); t = t.Add(bounds.Duration) {
+		rightRangeTimes = append(rightRangeTimes, t)
+	}
+
+	rightBlks := c.cache.MultiGet(rightRangeTimes)
+	firstNil := firstEmpty(rightBlks)
+	return rightBlks[:firstNil], firstNil != len(rightBlks)
+}
+
+// processCompletedBlocks processes all blocks for which are dependant blocks are present
```
arnikola: nit `... for which all dependent...`

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,348 @@
...
+			return err
+		}
+
+		processedKeys[i] = req.bounds.Start
+	}
+
+	// Mark all blocks as processed
+	c.cache.MarkProcessed(processedKeys)
+	// Sweep to free blocks from cache with no dependencies
+	c.sweep(c.cache.Processed(), queryStartBounds, queryEndBounds, maxBlocks)
```
arnikola: does sweep need to take in `cache.Processed()` if it's a method on `c` already?

src/coordinator/models/tag.go
```diff
@@ -214,3 +214,16 @@ func (t Tags) sortKeys() ([]string, int) {
 	sort.Strings(keys)
 	return keys, length
 }
+
+func (t Tags) WithoutName() Tags {
```
benraskin92: nit: add a comment

src/coordinator/functions/temporal/base.go
```diff
@@ -0,0 +1,348 @@
...
+	// Figure out the leftmost block and rightmost blocks
+	leftRangeStart := bounds.Previous(maxBlocks).Start
+	queryStartBounds := bounds.Nearest(c.transformOpts.TimeSpec.Start)
+
+	if leftRangeStart.Before(queryStartBounds.Start) {
+		leftRangeStart = queryStartBounds.Start
+	}
+
+	rightRangeStart := bounds.Next(maxBlocks).Start
+	queryEndBounds := bounds.Nearest(c.transformOpts.TimeSpec.End.Add(-1 * bounds.StepSize))
```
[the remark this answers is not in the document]
nikunjgit: Actually the reason we do it is because Sub works with time whereas Add works with duration.

src/coordinator/block/types.go
```diff
@@ -43,36 +43,101 @@ type SeriesMeta struct {
 	Name string
 }
 
-// Bounds are the time bounds
+// Bounds are the time bounds, start time is inclusive but end is exclusive
 type Bounds struct {
 	Start    time.Time
-	End      time.Time
+	Duration time.Duration
```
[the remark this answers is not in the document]
nikunjgit: StepSize ?

src/coordinator/executor/transform/cache.go
```diff
@@ -70,3 +71,92 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+	return cacheTime(t.UnixNano())
+}
+
+func (c cacheTime) toTime() time.Time {
+	return time.Unix(0, int64(c))
+}
+
+// TimeCache is used to cache blocks over time. It also keeps track of blocks which have been processed
+type TimeCache struct {
+	blocks          map[cacheTime]block.Block
```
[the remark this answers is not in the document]
nikunjgit: I agree, list is better. However, we need to figure out the block size before hand but currently we don't have that information :(

```
	maxBlocks := int(math.Ceil(float64(c.op.duration) / float64(blockDuration)))
```

src/coordinator/executor/transform/cache.go
nikunjgit: no longer valid since i switched to list

src/coordinator/executor/transform/cache.go
nikunjgit: several return paths. Will prefer defer here

src/coordinator/executor/transform/cache.go
nikunjgit: prefer defer here to avoid future bugs.

src/coordinator/models/tag.go
nikunjgit: no longer part of the diff

src/coordinator/functions/temporal/base.go
nikunjgit: yeah sounds good.

src/coordinator/functions/temporal/base.go
nikunjgit: simplified some logic now, let me know how it reads

src/coordinator/functions/temporal/base.go
nikunjgit: keeping sweep simple ?

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+		}
+
+		c.processedBlocks[index] = true
+	}
+}
+
+// Processed returns all processed block times from the cache
+func (c *blockCache) processed() []bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
arnikola: Unlock manually instead

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+	for i, j := 0, len(blocks)-1; i < j; i, j = i+1, j-1 {
+		blocks[i], blocks[j] = blocks[j], blocks[i]
+	}
+
+}
+
+// MarkProcessed is used to mark a block as processed
+func (c *blockCache) markProcessed(keys []time.Time) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
```
arnikola: Unlock manually instead

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+			if err != nil {
+				return nil, err
+			}
+
+			if empty {
+				break
+			}
+		}
+
+		reversed(blks)
```
arnikola: Can we use `sort.Reverse(blks)`?

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+	if err != nil {
+		return nil, false, err
+	}
+
+	return rightBlks, len(rightBlks) != maxBlocks, nil
+}
+
+// processCompletedBlocks processes all blocks for which are dependant blocks are present
+func (c *baseNode) processCompletedBlocks(processRequests []processRequest, queryStartBounds, queryEndBounds block.Bounds, maxBlocks int) error {
+	processedKeys := make([]time.Time, len(processRequests))
```
arnikola: Should this lock the mutex? Similar for processLeft and processRight

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+		}
+		depIters[i] = iter
+	}
+
+	bounds := seriesIter.Meta().Bounds
+	steps := int((aggDuration + bounds.Duration) / bounds.StepSize)
+	values := make([]float64, 0, steps)
+
+	seriesMeta := seriesIter.SeriesMeta()
+	resultSeriesMeta := make([]block.SeriesMeta, len(seriesMeta))
```
arnikola: Shouldn't resultSeriesMeta be the same as incoming seriesMeta? How do you combine seriesMetas between multiple blocks if a particular series does not exist in a particular block?

src/query/functions/temporal/base_test.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/m3db/m3/src/query/executor/transform"
+	"github.com/m3db/m3/src/query/parser"
+	"github.com/m3db/m3/src/query/test"
+	"github.com/m3db/m3/src/query/test/executor"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type processor struct {
```
arnikola: Can we get more tests for cases where we have multiple series per block, also where series are missing from some blocks, etc.?

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -150,6 +163,12 @@ func renderResultsJSON(w io.Writer, series []*ts.Series) {
 		vals := s.Values()
 		for i := 0; i < s.Len(); i++ {
 			dp := vals.DatapointAt(i)
+			// Skip points before the query boundary. Ideal place to adjust these would be at the result node but that would make it inefficient
+			// since we would need to create another block just for the sake of restricting the bounds
+			if dp.Timestamp.Before(params.Start) {
```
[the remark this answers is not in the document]
nikunjgit: So the series bounds are always same as the bounds of the block its generated from. The problem is that we sometimes have to fetch blocks for a longer query duration so that some functions can look back enough. However, when we return, we want to skip the excess points.

src/query/block/types.go
```diff
@@ -43,44 +43,109 @@ type SeriesMeta struct {
...
+
+// End calculates the end time for the block and is exclusive
+func (b Bounds) End() time.Time {
+	return b.Start.Add(b.Duration)
 }
 
 // Steps calculates the number of steps for the bounds
 func (b Bounds) Steps() int {
-	if b.Start.After(b.End) || b.StepSize <= 0 {
+	if b.StepSize <= 0 {
```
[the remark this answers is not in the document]
nikunjgit: I think only TimeForIndex doesn't make sense with 0 step size, other things should work. In theory we can use this for things without fixed resolution ? I was thinking about moving the bounds outside the block package and into models.

src/query/block/types.go
```diff
@@ -43,44 +43,109 @@ type SeriesMeta struct {
 	Name string
 }
 
-// Bounds are the time bounds
+// Bounds are the time bounds, start time is inclusive but end is exclusive
```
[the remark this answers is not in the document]
nikunjgit: yeah was thinking about moving it out of blocks. Will do it in another diff.

src/query/block/types_test.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package block
```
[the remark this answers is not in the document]
nikunjgit: should get resolved, once we move bounds out.

src/query/plan/physical.go
```diff
@@ -66,9 +67,36 @@ func NewPhysicalPlan(lp LogicalPlan, storage storage.Storage, params models.Requ
...
+	pl = pl.shiftTime()
 	return pl, nil
 }
 
+func (p PhysicalPlan) shiftTime() PhysicalPlan {
+	var maxOffset, maxRange time.Duration
+	for _, transformID := range p.pipeline {
+		node := p.steps[transformID]
+		boundOp, ok := node.Transform.Op.(transform.BoundOp)
+		if !ok {
```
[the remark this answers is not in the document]
nikunjgit: that would lead to nesting the whole block :(

src/query/models/params.go
```diff
@@ -29,9 +29,18 @@ type RequestParams struct {
...
-	Debug   bool
+	Now        time.Time
+	Timeout    time.Duration
+	Step       time.Duration
+	Target     string
+	Debug      bool
+	IncludeEnd bool
+}
+
+func (r RequestParams) ExclusiveEnd() time.Time {
```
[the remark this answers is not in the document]
nikunjgit: For params, I wanted to make it clear exactly what users were passing. In physical plan, we generate the actual end and going forward, the distinction is not known

src/query/functions/temporal/count.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (c *countNode) Process(values []float64) float64 {
+	var count float64
+	for _, v := range values {
+		if !math.IsNaN(v) {
+			count++
+		}
+	}
+
+	return count
```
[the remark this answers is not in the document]
nikunjgit: not sure if i follow. Future operations are on different blocks ?

src/query/executor/transform/cache.go
```diff
@@ -70,3 +71,95 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+}
+
+
+// Add the block to the cache, errors out if block already exists
+func (c *TimeCache) Add(key time.Time, b block.Block) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	_, ok := c.blocks[fromTime(key)]
+	if ok {
+		return errors.New("block already exists")
```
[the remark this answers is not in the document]
nikunjgit: removed this code after refactor

src/query/executor/transform/cache.go
```diff
@@ -70,3 +71,95 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+		processedBlocks: make(map[cacheTime]bool),
+	}
+}
+
+
+// Add the block to the cache, errors out if block already exists
+func (c *TimeCache) Add(key time.Time, b block.Block) error {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	_, ok := c.blocks[fromTime(key)]
```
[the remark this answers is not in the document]
nikunjgit: removed this code after refactor

src/query/executor/transform/cache.go
```diff
@@ -70,3 +71,95 @@ func (c *BlockCache) Get(key parser.NodeID) (block.Block, bool) {
...
+func fromTime(t time.Time) cacheTime {
+	return cacheTime(t.UnixNano())
+}
+
+func (c cacheTime) toTime() time.Time {
+	return time.Unix(0, int64(c))
+}
+
+// TimeCache is used to cache blocks over time. It also keeps track of blocks which have been processed
+type TimeCache struct {
```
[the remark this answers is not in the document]
nikunjgit: removed this code after refactor

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+	duration     time.Duration
+	processorFn  MakeProcessor
+}
+
+func newBaseOp(args []interface{}, operatorType string, processorFn MakeProcessor) (baseOp, error) {
+	if len(args) != 1 {
+		return emptyOp, fmt.Errorf("invalid number of args for %s: %d", operatorType, len(args))
+	}
+
+	duration, ok := args[0].(time.Duration)
```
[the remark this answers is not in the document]
nikunjgit: yes, also added a test in parse to make sure

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+	if !ok {
+		return emptyOp, fmt.Errorf("unable to cast to scalar argument: %v", args[0])
```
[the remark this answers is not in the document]
nikunjgit: i think we have to figure out a way to propagate the right context for most of our errors.

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+}
+
+func newBlockCache(op baseOp, transformOpts transform.Options) *blockCache {
+	return &blockCache{
+		op:            op,
+		transformOpts: transformOpts,
+	}
+}
+func (c *blockCache) init(bounds block.Bounds) {
+	if c.initialized {
```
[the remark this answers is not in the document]
nikunjgit: umm i think it should work ?

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+	blockDuration := bounds.Duration
+	// Figure out the maximum blocks needed for the temporal function
+	maxBlocks := int(math.Ceil(float64(c.op.duration) / float64(blockDuration)))
+
+	// Figure out the leftmost block and rightmost blocks
+	leftRangeStart := bounds.Previous(maxBlocks)
+	queryStartBounds := bounds.Nearest(c.transformOpts.TimeSpec.Start)
+
+	if leftRangeStart.Start.Before(queryStartBounds.Start) {
+		leftRangeStart = queryStartBounds
```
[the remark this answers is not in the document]
nikunjgit: you mean Previous() takes in a time and returns returns only until that time ? That makes it a little confusing :( I think making structs here is fine since its on the stack and not on the heap

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+// String representation
+func (o baseOp) String() string {
+	return fmt.Sprintf("type: %s, duration: %v", o.OpType(), o.duration)
+}
+
+// Node creates an execution node
+func (o baseOp) Node(controller *transform.Controller, opts transform.Options) transform.OpNode {
+	return &baseNode{
+		controller:    controller,
+		cache:         newBlockCache(o, opts),
```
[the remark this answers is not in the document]
nikunjgit: naming is confusing i suppose, I'll change that. Essentially this is the time cache.

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,498 @@
...
+	mu            sync.Mutex
+	transformOpts transform.Options
+}
+
+// Process processes a block. The processing steps are as follows:
+// 1. Figure out the maximum blocks needed for the temporal function
+// 2. For the current block, figure out whether we have enough previous blocks which can help process it
+// 3. For the blocks after current block, figure out which can be processed right now
+// 4. Process all valid blocks from #3, #4 and mark them as processed
+// 5. Run a sweep face to free up blocks which are no longer needed to be cached
```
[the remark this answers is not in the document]
nikunjgit: processCompletedBlock is called from within Process. I'm just writing the whole workflow in this comment

src/query/functions/temporal/base.go
nikunjgit: always nervous doing this.

src/query/functions/temporal/base.go
nikunjgit: we don't want the blocks in descending sorted order, this is more about just reversing the array

src/query/functions/temporal/base.go
nikunjgit: markProcessed and sweep are the only ones mutating the blocks. markProcessed is already within a mutex and i'm not sure if sweep needs it since its fine for sweep to miss a few blocks and they can be sweeped later on. Not sure if processLeft and processRight needs a mutex since the multiGet is already locked. 

We might have to revisit this later

src/query/functions/temporal/base.go
nikunjgit: in resultSeriesMeta we just need to remove the seriesname (same as Prom), rest is the same. We assume all blocks have the same series. If it doesnt' then that series should be all NaNs for that time range.

src/query/functions/temporal/base_test.go
nikunjgit: i don't think series can be missing from some blocks. In that case, they should just be NaNs. I'll add tests for processSingleRequest

src/query/functions/fetch.go
```diff
@@ -56,6 +59,14 @@ func (o FetchOp) OpType() string {
 	return FetchType
```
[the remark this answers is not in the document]
nikunjgit: yeah! Don't want to do it in this diff though

src/query/functions/temporal/base_test.go
arnikola: Seems inefficient to jam in a bunch of unnecessary NaNs for series that don't exist in blocks just to make processing a little easier; could add up to a lot of unneeded datapoints for sparse series

src/query/functions/fetch.go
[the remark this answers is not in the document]
arnikola: Mind adding a todo jic?

src/query/api/v1/handler/prometheus/native/common.go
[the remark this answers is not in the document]
arnikola: I get the motivation, but at the moment it may iterate through a bunch of unnecessary data points; better to calculate the first valid `i` to start from, rather than trying a bunch and discarding. Not really a big deal with a couple of series, but something like 10,000 series where only the last few datapoints in a 2 hour block  are required, we'll be doing a lot of unnecessary looping

src/query/functions/temporal/base.go
arnikola: sort.Reverse just reverses the list, shouldn't do any sorting

src/query/functions/temporal/base_test.go
nikunjgit: I think that depends on the iterator but it's probably just gonna return NaNs and not actually store them.

src/query/functions/temporal/base.go
nikunjgit: don't think that's correct. Check out : https://golang.org/pkg/sort/#Reverse. They have an example there. Reverse needs Len(), Comparator, etc. and sorts the list in reverse order

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	return blks, nil
+}
+
+func reversed(blocks []block.Block) {
+	for i, j := 0, len(blocks)-1; i < j; i, j = i+1, j-1 {
+		blocks[i], blocks[j] = blocks[j], blocks[i]
+	}
+
```
arnikola: nit: newline

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+		if empty {
+			break
+		}
+	}
+
+	return blks, nil
+}
+
+func reversed(blocks []block.Block) {
```
arnikola: nit: rename to `reverse`

src/query/functions/temporal/base.go
arnikola: Fair enough; weird that there's no inbuilt reverse array method

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+		return blks, nil
+	}
+
+	startIdx, err := c.index(startBounds.Start)
+	if err != nil {
+		return nil, err
+	}
+
+	// Process a single index
+	process := func(i int) (bool, error) {
```
arnikola: nit: rename to addIfPresent, maybe add a comment as to what this does. Also, flip to return false if not added, and true if added, since true usually corresponds to the positive case

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+	return b, b != nil
+}
+
+// multiGet retrieves multiple blocks from the cache at once until if finds an empty block
+func (c *blockCache) multiGet(startBounds block.Bounds, numBlocks int, reverse bool) ([]block.Block, error) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	blks := make([]block.Block, 0, numBlocks)
+	if numBlocks == 0 {
```
arnikola: nit: can do this before the mutex lock

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+	c.initialized = true
+}
+
+func (c *blockCache) index(t time.Time) (int, error) {
+	start := c.startBounds.Start
+	if t.Before(start) || t.After(c.endBounds.Start) {
+		return 0, fmt.Errorf("invalid time for the block cache: %v, start: %v, end: %v", t, start, c.endBounds.Start)
+	}
+
+	return int(t.Sub(start) / c.startBounds.Duration), nil
```
arnikola: Add a sanity check for startBounds.Duration==0?

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	if c.initialized {
+		return
+	}
+
+	timeSpec := c.transformOpts.TimeSpec
+	c.startBounds = bounds.Nearest(timeSpec.Start)
+	c.endBounds = bounds.Nearest(timeSpec.End.Add(-1 * bounds.StepSize))
+	numBlocks := c.endBounds.End().Sub(c.startBounds.Start) / bounds.Duration
```
arnikola: Add a sanity check for bounds.Duration==0? Also is it correct to take floor here?

src/query/functions/temporal/base.go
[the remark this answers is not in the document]
arnikola: I think go may be a bit touchy about this; may fail --race tests as is

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+	processedBlocks []bool
+}
+
+func newBlockCache(op baseOp, transformOpts transform.Options) *blockCache {
+	return &blockCache{
+		op:            op,
+		transformOpts: transformOpts,
+	}
+}
+func (c *blockCache) init(bounds block.Bounds) {
```
arnikola: nit newline

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+		if dependantBlocks > numBlocks-i-1 {
+			dependantBlocks = numBlocks - i - 1
+		}
+
+		if prevProcessed >= dependantBlocks {
+			c.cache.remove(i)
+		}
+
+		prevProcessed++
+
```
arnikola: nit newline

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+			continue
+		}
+
+		dependantBlocks := maxBlocks
+		if dependantBlocks > numBlocks-i-1 {
+			dependantBlocks = numBlocks - i - 1
+		}
+
+		if prevProcessed >= dependantBlocks {
+			c.cache.remove(i)
```
arnikola: This seems like it would cause issues considering it's removing at an index which changes independently of here

i.e. if we're deleting at index 0,1,2 from [0,1,2], wouldn't it go something like [0,1,2]->[1,2]->[1]->out of index?

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+	prevProcessed := 0
+	numBlocks := len(processedKeys) - 1
+	for i := numBlocks; i >= 0; i-- {
+		processed := processedKeys[i]
+		if !processed {
+			prevProcessed = 0
+			continue
+		}
+
+		dependantBlocks := maxBlocks
```
arnikola: nit `dependentBlocks`

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+			val := series.ValueAtStep(i)
+			values = append(values, val)
+			newVal := math.NaN()
+			if desiredLength <= len(values) {
+				values = values[len(values)-desiredLength:]
+				newVal = c.processor.Process(values)
+			}
+
+			builder.AppendValue(i, newVal)
+
```
arnikola: nit newline

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+		series, err := seriesIter.Current()
+		if err != nil {
+			return err
+		}
+
+		for i := 0; i < series.Len(); i++ {
+			val := series.ValueAtStep(i)
+			values = append(values, val)
+			newVal := math.NaN()
+			if desiredLength <= len(values) {
```
arnikola: Maybe add a comment here? Wasn't super clear what was going on during first readthrough

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+
+			s, err := iter.Current()
+			if err != nil {
+				return err
+			}
+
+			values = append(values, s.Values()...)
+		}
+
+		desiredLength := int(aggDuration / bounds.StepSize)
```
arnikola: nit: Get this outside of the loop

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+	if err != nil {
+		return err
+	}
+
+	if err := builder.AddCols(bounds.Steps()); err != nil {
+		return err
+	}
+
+	for seriesIter.Next() {
+		values = values[0:0]
```
arnikola: nit: can do values[:0] instead?

src/query/functions/temporal/base.go
arnikola: Oh does prom drop name here? Weird...

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+	for i, blk := range request.deps {
+		iter, err := blk.SeriesIter()
+		if err != nil {
+			return err
+		}
+		depIters[i] = iter
+	}
+
+	bounds := seriesIter.Meta().Bounds
+	steps := int((aggDuration + bounds.Duration) / bounds.StepSize)
```
arnikola: Add stepSize==0 sanity check; also can define this closer to where it's needed

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+		return err
+	}
+
+	depIters := make([]block.SeriesIter, len(request.deps))
+	for i, blk := range request.deps {
+		iter, err := blk.SeriesIter()
+		if err != nil {
+			return err
+		}
+		depIters[i] = iter
```
arnikola: nit newline

src/query/functions/temporal/base.go
```diff
@@ -0,0 +1,505 @@
...
+
+	// Mark all blocks as processed
+	c.cache.markProcessed(processedKeys)
+	// Sweep to free blocks from cache with no dependencies
+	c.sweep(c.cache.processed(), queryStartBounds, queryEndBounds, maxBlocks)
+	return nil
+}
+
+func (c *baseNode) processSingleRequest(request processRequest) error {
+	aggDuration := c.op.duration
```
arnikola: nit: define closer to where it's needed

src/query/functions/temporal/base.go
nikunjgit: i made the name clearer and added a comment. I still think its reads better with empty, let me know once you read the new comment/naming

src/query/functions/temporal/base.go
nikunjgit: will panic in init if that's the case so not needed here. To fix init panic, I added a sanity check in Process()

src/query/functions/temporal/base.go
nikunjgit: To fix init panic, I added a sanity check in Process(). Bounds.Nearest aligns start and end bounds to the step size so they will be fully divisible

src/query/functions/temporal/base.go
nikunjgit: if it fails --race test then I will move within mutex

src/query/functions/temporal/base.go
nikunjgit: added a test for that and it seems to work :)

src/query/functions/temporal/base.go
nikunjgit: should be handled by sanity check on bounds in Process()

nikunjgit: need in the next line

src/query/functions/fetch.go
```diff
@@ -29,12 +29,16 @@ import (
...
+	"github.com/m3db/m3/src/query/util/logging"
+
+	"go.uber.org/zap"
 )
 
 // FetchType gets the series from storage
 const FetchType = "fetch"
 
 // FetchOp stores required properties for fetch
+// TODO: Make FetchOp private
```
arnikola: 👍

src/query/functions/temporal/count.go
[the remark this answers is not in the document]
arnikola: There'll be a lot of shared calculations between incoming blocks right? E.g. if we have a 3block long aggregation, middle block would have to be recalculated a few times; would be good if we could reuse past calculations to make it a bit faster

PR #812, opened 2018-07-26
Add Round func

benraskin92:


src/coordinator/functions/linear/round.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+
+	var (
+		toNearest = 1.0
+		ok        bool
+	)
+	if len(args) > 0 {
+		toNearest, ok = args[0].(float64)
+		if !ok {
+			return emptyOp, fmt.Errorf("unable to cast to toNearest argument: %v", args[0])
```
nikunjgit: not sure if people will understand toNearest. Prom definition says to_nearest

src/coordinator/functions/linear/round.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"fmt"
+	"math"
+
+	"github.com/m3db/m3db/src/coordinator/executor/transform"
+)
+
+// RoundType rounds each value in the timeseries to the nearest integer.
+// Ties are resolved by rounding up.
+const RoundType = "round"
```
nikunjgit: maybe also comment on how to_nearest is used ?

PR #815, opened 2018-07-31
Do not cap attempts to flush with retries, should always attempt to write out

robskillington:
This change avoids potentially returning `false` from `NeedsFlush(...)` for a namespace when max retries have been exceeded for flushing.

There are two faults with this behavior:
1. If `false` returns from `NeedsFlush(...)` due to retries being exceeded for a given block time start, the cleanup code may believe that it was successful and can cleanup commit logs and snapshots - hence losing data on the local node.
1. Naturally if disk was not writeable for some time, we'd ideally like it to just natively recover and flush out blocks that haven't been flushed yet, rather than max retries being hit and then having to restart nodes to have them flush out their data.

### DIFF
--- a/src/dbnode/storage/database_test.go
+++ b/src/dbnode/storage/database_test.go
@@ -88,7 +88,6 @@ func init() {
 		SetSeriesCachePolicy(series.CacheAll).
 		SetPersistManager(pm).
 		SetRepairEnabled(false).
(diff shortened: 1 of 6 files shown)


robskillington:
@nerd0: Prior to this change, that was the status quo.  With this change it'll naturally recover.

PR #816, opened 2018-07-31
Add datetime Prom functions

benraskin92:


src/coordinator/functions/datetime/base.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/m3db/m3db/src/coordinator/block"
+	"github.com/m3db/m3db/src/coordinator/executor/transform"
+	"github.com/m3db/m3db/src/coordinator/parser"
+)
+
+var emptyOp = BaseOp{}
+
+// BaseOp stores required properties for logical operations
+type BaseOp struct {
```
nikunjgit: don't expose BaseOp

src/coordinator/functions/datetime/base.go
```diff
@@ -0,0 +1,133 @@
...
+	"time"
+
+	"github.com/m3db/m3db/src/coordinator/block"
+	"github.com/m3db/m3db/src/coordinator/executor/transform"
+	"github.com/m3db/m3db/src/coordinator/parser"
+)
+
+var emptyOp = BaseOp{}
+
+// BaseOp stores required properties for logical operations
```
nikunjgit: old comment ?

src/coordinator/functions/datetime/base.go
```diff
@@ -0,0 +1,133 @@
...
+		cache:      transform.NewBlockCache(),
+		op:         o,
+		processor:  o.processorFn(o, controller),
+	}
+}
+
+type baseNode struct {
+	op         BaseOp
+	controller *transform.Controller
+	cache      *transform.BlockCache
```
nikunjgit: don't need cache

src/coordinator/functions/datetime/base.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	nextBlock := builder.Build()
+	defer nextBlock.Close()
+	return c.controller.Process(nextBlock)
+}
+
+// Meta returns the metadata for the block
+func (c *baseNode) Meta(meta block.Metadata) block.Metadata {
+	return meta
```
nikunjgit: is this correct ? Don't you need to modify some meta ?

src/coordinator/functions/datetime/base.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// SeriesMeta returns the metadata for each series in the block
+func (c *baseNode) SeriesMeta(metas []block.SeriesMeta) []block.SeriesMeta {
+	return metas
```
nikunjgit: same here. Check if this is correct

src/coordinator/functions/datetime/date.go
```diff
@@ -0,0 +1,140 @@
...
+
+type dateNode struct {
+	op         BaseOp
+	controller *transform.Controller
+	dateFn     func(time.Time) float64
+}
+
+func (c *dateNode) ProcessStep(values []float64, t time.Time) []float64 {
+	for i := range values {
+		if c.op.operatorType == TimestampType {
```
nikunjgit: looks a bit weird. Maybe just have a different file for TimestampType ?

src/coordinator/parser/promql/types.go
```diff
@@ -103,11 +99,41 @@ func NewFunctionExpr(name string, argValues []interface{}) (parser.Params, error
...
+
+	case linear.ClampMinType:
+		return linear.NewClampOp(argValues, linear.ClampMinType)
+	case linear.ClampMaxType:
+		return linear.NewClampOp(argValues, linear.ClampMaxType)
+
+	case linear.RoundType:
+		return linear.NewRoundOp(argValues)
+
+	case datetime.DayOfMonthType:
```
nikunjgit: all this can be a single case:

case datetime.DayOfMonthType, datetime.DayOfWeekType, ....: 
   return datetime.NewDateOp(name)

nikunjgit: Also do it for other things (I created a bad pattern here)

src/coordinator/functions/datetime/base.go
benraskin92: I don't think so, looks like the same meta data gets returned.

src/coordinator/functions/datetime/base.go
benraskin92: same as above

src/coordinator/functions/datetime/timestamp.go
```diff
@@ -0,0 +1,84 @@
...
+}
+
+type timestampNode struct {
+	op         baseOp
+	controller *transform.Controller
+}
+
+func (c *timestampNode) ProcessStep(values []float64, t time.Time) []float64 {
+	for i := range values {
+		if math.IsNaN(values[i]) {
```
nikunjgit: extract this part out ? You're using it twice

src/coordinator/parser/promql/types.go
```diff
@@ -83,30 +84,27 @@ func NewBinaryOperator(expr *promql.BinaryExpr, lhs, rhs parser.NodeID) (parser.
...
+
 	case linear.RoundType:
 		return linear.NewRoundOp(argValues)
-	case linear.SqrtType:
-		return linear.NewMathOp(linear.SqrtType)
+
+	case datetime.DayOfMonthType, datetime.DayOfWeekType, datetime.DaysInMonthType, datetime.HourType,
+		datetime.MinuteType, datetime.MonthType, datetime.TimestampType, datetime.YearType:
+		return datetime.NewDateOp(name)
+	case datetime.TimeType:
```
nikunjgit: fix ?

src/coordinator/parser/promql/parse_test.go
```diff
@@ -215,3 +216,85 @@ func TestDAGWithRoundOp(t *testing.T) {
 	assert.Len(t, transforms, 2)
 	assert.Equal(t, transforms[1].Op.OpType(), linear.RoundType)
 }
+
+func TestDAGWithDayOfMonthOp(t *testing.T) {
```
nikunjgit: add test for timestamp ?

benraskin92: Already added

src/query/functions/datetime/datetime.go
```diff
@@ -0,0 +1,116 @@
...
+
+type dateNode struct {
+	op         baseOp
+	controller *transform.Controller
+	dateFn     func(time.Time) float64
+}
+
+func (c *dateNode) ProcessStep(values []float64, t time.Time) []float64 {
+	for i := range values {
+		values[i] = c.dateFn(t)
```
nikunjgit: can you move c.dateFn(t) out of the loop ? Seems like its the same for the whole step

PR #817, opened 2018-08-01
Change coordinator to query and add `m3query` service

benraskin92:


docker/m3coordinator/Dockerfile
```diff
@@ -21,7 +21,7 @@ LABEL maintainer="The M3DB Authors <m3db@googlegroups.com>"
 EXPOSE 7201/tcp 7203/tcp
 
 COPY --from=builder /go/src/github.com/m3db/m3db/bin/m3coordinator /bin/
-COPY --from=builder /go/src/github.com/m3db/m3db/src/coordinator/config/m3coordinator-local-etcd.yml /etc/m3coordinator/m3coordinator.yml
+COPY --from=builder /go/src/github.com/m3db/m3db/src/query/config/m3coordinator-local-etcd.yml /etc/m3coordinator/m3coordinator.yml
```
prateek: hm do you want to chance the directory name here too

scripts/process-cover.sh
```diff
@@ -12,7 +12,7 @@ TARGETS=("dbnode" "coordinator" "m3ninx")
 target_patterns() {
     case $1 in
         'dbnode') echo "^mode|github.com/m3db/m3db/src/dbnode|github.com/m3db/m3db/src/cmd/services/m3dbnode";;
-        'coordinator') echo "^mode|github.com/m3db/m3db/src/coordinator|github.com/m3db/m3db/src/cmd/services/m3coordinator";;
+        'coordinator') echo "^mode|github.com/m3db/m3db/src/query|github.com/m3db/m3db/src/cmd/services/m3coordinator";;
```
prateek: hmm maybe rename this to 'query' too

prateek: similarly for the `.codecov.yml` file and -F flags in the makefile

docker/m3coordinator/Dockerfile
robskillington: Yeah maybe we should move these config files outside of `src` completely?  It's weird we have some YAML files in these dirs floating around.

src/cmd/services/m3coordinator/main/main.go
```diff
@@ -25,7 +25,7 @@ import (
 	_ "net/http/pprof" // pprof: for debug listen server if configured
 	"os"
 
-	"github.com/m3db/m3db/src/coordinator/services/m3coordinator/server"
+	"github.com/m3db/m3db/src/query/services/m3coordinator/server"
```
robskillington: The plan is to follow up and move the package being imported here to `src/coordinator/server/...` yeah?

docker/m3coordinator/Dockerfile
benraskin92: i guess we should figure out if we want to have an m3coordinator and an m3query dockerfile? we can discuss today once everyone is in

PR #819, opened 2018-08-02
[WIP] Mutable -> FST Segment rotation for un-sealed index.Block

prateek:
Smaller PRs:
- [ ] m3ninx persist PR

Pending:
- [x] e2e latency approximation per segment
- [x] Reduce contention between compacted segments/readers 
- [ ] Trigger mutable segment rotation based on time too (outside the insert path)
- [ ] TODOs in the code
- [ ] Unit/Integration tests 
- [x] FST-ify bootstraped segments 

### followups
- [ ] Filter cache: LRU cache `(segmentID, queryComponent) -> postingsList` for all FST segments over size X (what about doing this for commonly used field/term FSTs too?)
- [ ] Postings list benchmarks: roaring v pilosa for serialization/deserialization + query operations
- [ ] Query optimiser:
    - [ ] Ordering of postings lists based on size
    - [ ] Indexing (postings list/field) for "field exists" + query type
    - [ ] Executor/Searcher model to be optimised per reader instead of per query component
    - [ ] Add tracing for query component latency
    - [ ] Size/sketch fields for query execution heurisitcs (e.g. intersection of hll/bloomfilter) 
    - [ ] add size to be retrieved before postings list deserialisation 
- [ ] Tracing for compaction latency
- [ ] Remove locks from postings lists, mutable segment, etc 
- [ ] documents compression at storage time

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -392,7 +392,7 @@ imports:
 - name: github.com/philhofer/fwd
   version: bb6d471dc95d4fe11e432687f8b70ff496cf3136
 - name: github.com/pilosa/pilosa
-  version: a112b2d46af94e3ecfa74b8158381e3595b24e4b
+  version: b22780ccf149c564f8ca271f66965aca1beb3dfa
   subpackages:
   - roaring
 - name: github.com/pkg/errors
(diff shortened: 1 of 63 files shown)


prateek:
Superseeded by #1197

PR #822, opened 2018-08-04
Add scalar block and binary functions

arnikola:
Added scalar block type for parsing prom constant values

Arithmetic functions added (+, -, *, /, ^, %) for scalar-scalar,
scalar-series, series-scalar, series-series left and right node values 

Comparison functions added (==, !=, >, <, >=, <=) for scalar-scalar,
scalar-series, series-scalar, series-series left and right node values 

Added unit tests for all permutations of comparison and arithmetic functions

Still TODO: consolidate binary node with logical node, as a lot of behaviour
is shared.


src/coordinator/functions/binary/binary_node.go
```diff
@@ -0,0 +1,232 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
nikunjgit: node.go?

src/coordinator/block/scalar_block.go
```diff
@@ -0,0 +1,150 @@
...
+			Bounds: bounds,
+			Tags:   make(models.Tags),
+		},
+	}
+}
+
+// StepIter returns a StepIterator
+func (b *scalarBlock) StepIter() (StepIter, error) {
+	bounds := b.meta.Bounds
+	steps := bounds.End.Sub(bounds.Start) / bounds.StepSize
```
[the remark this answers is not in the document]
arnikola: Turns out `Bounds.Steps()` exists already :) using that instead

src/query/block/scalar.go
```diff
@@ -0,0 +1,172 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/m3db/m3/src/query/models"
+)
+
+// ScalarBlock is a block containing a single value
+// over a certain bound
+type ScalarBlock struct {
```
nikunjgit: call it Scalar since its already in block package. Also do you really need to expose it outside ? Seems like you outside functions should just work on the Block interface

src/query/block/scalar.go
```diff
@@ -0,0 +1,172 @@
...
+	val  float64
+	meta Metadata
+}
+
+// assert that ScalarBlock implements Block
+var _ Block = (*ScalarBlock)(nil)
+
+// NewScalarBlockResult creates a Result consisting
+// of scalar blocks over the bound
+func NewScalarBlockResult(val float64, bounds Bounds) *Result {
```
nikunjgit: why does result need to know about scalar block ?

src/query/block/scalar.go
```diff
@@ -0,0 +1,172 @@
...
+package block
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/m3db/m3/src/query/models"
+)
+
+// ScalarBlock is a block containing a single value
```
nikunjgit: can you add a comment as to when this would be useful ?

src/query/block/scalar.go
```diff
@@ -0,0 +1,172 @@
...
+
+func (it *scalarStepIter) Next() bool {
+	it.idx++
+	bounds := it.meta.Bounds
+	it.step.time = bounds.Start.Add(time.Duration(it.idx) * bounds.StepSize)
+	return it.idx < it.numVals
+}
+
+func (it *scalarStepIter) Current() (Step, error) {
+	if it.idx >= it.numVals || it.idx < 0 {
```
nikunjgit: we haven't traditionally been checking this in Current() usually rely on people to call Next() before Current

src/query/block/scalar.go
```diff
@@ -0,0 +1,172 @@
...
+func (it *scalarSeriesIter) SeriesCount() int         { return 1 }
+func (it *scalarSeriesIter) SeriesMeta() []SeriesMeta { return seriesMeta }
+func (it *scalarSeriesIter) Meta() Metadata           { return it.meta }
+func (it *scalarSeriesIter) Next() bool {
+	it.idx++
+	return it.idx == 0
+}
+
+func (it *scalarSeriesIter) Current() (Series, error) {
+	if it.idx != 0 {
```
nikunjgit: we haven't traditionally been checking this in Current() usually rely on people to call Next() before Current

src/query/block/scalar.go
```diff
@@ -0,0 +1,172 @@
...
+		return Series{}, fmt.Errorf("invalid scalar index: %d, numVals: 1", it.idx)
+	}
+	return Series{
+		Meta: SeriesMeta{
```
nikunjgit: a bit confusing that at some places you're using the outer seriesMeta and here you're creating another one

src/query/executor/state.go
```diff
@@ -42,21 +42,41 @@ type ExecutionState struct {
...
+func CreateTransform(
+	ID parser.NodeID,
+	params transform.Params,
+) (transform.OpNode, *transform.Controller) {
 	controller := &transform.Controller{ID: ID}
 	node := params.Node(controller)
 
 	switch node.(type) {
 	case transform.SeriesNode:
+		fmt.Println("Series node")
```
nikunjgit: remove Prints ?

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,482 @@
...
+		GreaterType:   func(x, y float64) float64 { return toFloat(x > y) },
+		LesserType:    func(x, y float64) float64 { return toFloat(x < y) },
+		GreaterEqType: func(x, y float64) float64 { return toFloat(x >= y) },
+		LesserEqType:  func(x, y float64) float64 { return toFloat(x <= y) },
+	}
+
+	errLeftScalar              = errors.New("expected left scalar but node type incorrect")
+	errRightScalar             = errors.New("expected right scalar but node type incorrect")
+	errNoModifierForComparison = errors.New("comparisons between scalars must use BOOL modifier")
+	errNoMatching              = errors.New("functions must have a vector matching set")
```
nikunjgit: can you make the message clear ? Not sure what a vector matching set is

src/query/parser/promql/parse_test.go
```diff
@@ -63,260 +64,99 @@ func TestDAGWithUnknownOp(t *testing.T) {
...
-}
-func TestDAGWithYearOp(t *testing.T) {
-	q := "year(up)"
-	p, err := Parse(q)
-	require.NoError(t, err)
-	transforms, _, err := p.DAG()
-	require.NoError(t, err)
-	assert.Len(t, transforms, 2)
-	assert.Equal(t, transforms[1].Op.OpType(), linear.YearType)
+var linearParseTests = []struct {
```
nikunjgit: sweet!

src/query/functions/scalar.go
```diff
@@ -0,0 +1,79 @@
...
+
+// scalarNode is the execution node
+type scalarNode struct {
+	op         scalarOp
+	controller *transform.Controller
+	timespec   transform.TimeSpec
+}
+
+// Execute runs the fetch node operation
+func (n *scalarNode) Execute(ctx context.Context) error {
```
nikunjgit: do you want to support debug logging here (same as fetch)

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,482 @@
...
+type binaryNode struct {
+	op         binaryOp
+	controller *transform.Controller
+	cache      *transform.BlockCache
+	mu         sync.Mutex
+}
+
+// NodeInformation describes the types of nodes used
+// for binary operations
+type NodeInformation struct {
```
nikunjgit: node information is too generic :(  how about binaryOpParams or something more specific ?

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,482 @@
...
+		return err
+	}
+
+	defer nextBlock.Close()
+	return n.controller.Process(nextBlock)
+}
+
+// processes two logical blocks, performing a logical operation on them
+func (n *binaryNode) process(lhs, rhs block.Block) (block.Block, error) {
+	info := n.op.info
```
nikunjgit: get it later when you are actually using it ?

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,482 @@
...
+		scalarL, ok := lhs.(*block.ScalarBlock)
+		if !ok {
+			return nil, errLeftScalar
+		}
+
+		lVal := scalarL.Value()
+		// if both lhs and rhs are scalars, can create a new block
+		// by extracting values from lhs and rhs instead of doing
+		// by-value comparisons
+		if info.RIsScalar {
```
nikunjgit: you can flip the check since the other condition is a single line

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,482 @@
...
+
+			builder.AppendValue(index, actualFn(lVal, rVal))
+		}
+	}
+
+	return builder.Build(), nil
+}
+
+// computeOrCache figures out if both lhs and rhs are available, if not then it caches the incoming block
+func (n *binaryNode) computeOrCache(ID parser.NodeID, b block.Block) (block.Block, block.Block, error) {
```
nikunjgit: is this the same as the other place ? If yes, then think about refactor

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,482 @@
...
+	defer n.mu.Unlock()
+	n.cache.Remove(n.op.info.LNode)
+	n.cache.Remove(n.op.info.RNode)
+}
+
+const initIndexSliceLength = 10
+
+// intersect returns the slice of lhs indices that are shared with rhs,
+// the indices of the corresponding rhs values, and the metas for taken indices
+func intersect(
```
nikunjgit: is this the same as the other place ? If yes, then think about refactor

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,482 @@
...
+
+func isComparison(op string) bool {
+	return op == EqType || op == NotEqType ||
+		op == GreaterType || op == LesserType ||
+		op == GreaterEqType || op == LesserEqType
+}
+
+// If returnBool is false and op is a comparison function,
+// the function should return the scalar value rather than 1 or 0
+func actualScalarFunc(fn singleScalarFunc, op string, returnBool bool) singleScalarFunc {
```
nikunjgit: naming :(

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,482 @@
...
+		}
+	}
+
+	return builder.Build(), nil
+}
+
+// If returnBool is false and op is a comparison function,
+// the function should return the scalar value of lhs if the comparison
+// is true; otherwise, it should return 1 if true, and 0 if false
+func actualFunc(fn mathFunc, op string, returnBool bool) mathFunc {
```
nikunjgit: naming :(

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,482 @@
...
+		return n.processSingleBlock(
+			rhs,
+			func(x float64) float64 {
+				return n.op.fn(lVal, x)
+			},
+		)
+	}
+
+	if info.RIsScalar {
+		scalarR, ok := rhs.(*block.ScalarBlock)
```
nikunjgit: some duplication here:

maybe create a function which takes two block.Block, assumes first can be casted into scalar and other can't be. Then you can use the same function for this case and one of the previous cases

src/query/functions/binary/node.go
arnikola: Yeah, this falls under:
TODO: consolidate binary node with logical node, as a lot of behaviour
is shared.

Might do that in a separate commit once this is in. No other binary functions are planned in the near future so shouldn't affect any upcoming commits

src/query/functions/binary/node.go
arnikola: It's slightly different in that this one is designed to be able to drop series which do not match, consistent with prom behaviour here.

src/query/functions/binary/node.go
arnikola: Good call, holdover from before I pulled out the other functionality into a function

src/query/functions/binary/node.go
arnikola: It's in the binary package, so it would be binary.NodeInformation if we're using it elsewhere? If you're still against it can change thi

arnikola: Met you halfway with NodeParams :p

src/query/block/scalar.go
arnikola: I cast to Scalar block sometimes as an optimization to pull out the value for the entire block, rather than having to step through each value one by one.

src/query/block/scalar.go
arnikola: I cast to Scalar block sometimes as an optimization to pull out the value for the entire block, rather than having to step through each value one by one. Will rename

src/query/block/scalar.go
arnikola: I was mirroring how fetch performed executions; changed it a bit to only build a single block and use that instead.

src/query/block/scalar.go
arnikola: I agree in general, but since this doesn't do any real "bounds" checking and just returns a value, the alternative is that this `panics` in this check, as it really shouldn't be returning anything if Next() is false.

I prefer to error than to panic here, but if you have particular preference towards panicing I can change to that

src/query/block/scalar.go
arnikola: I agree in general, but since this doesn't do any real "bounds" checking and just returns a value, the alternative is that this `panics` in this check, as it really shouldn't be returning anything if Next() is false.

I prefer to error than to panic here, but if you have particular preference towards panicing I can change to that

src/query/executor/state.go
arnikola: Oops

src/query/block/scalar.go
arnikola: Fair call; extracted it to be a bit clearer

src/query/functions/binary/node.go
arnikola: Sure, it does seem a bit weird. It's more of a sanity check since I expect a parse error before it gets here, but will make it clearer

src/query/functions/scalar.go
arnikola: Good idea

src/query/block/scalar.go
```diff
@@ -0,0 +1,162 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/m3db/m3/src/query/models"
+)
+
+// Scalar is a block containing a single value
+// over a certain bound
```
benraskin92: nit: add this to end of first line

src/query/block/scalar.go
```diff
@@ -0,0 +1,162 @@
...
+// Close closes the scalar block
+func (b *Scalar) Close() error { return nil }
+
+// Value returns the value for the scalar block
+func (b *Scalar) Value() float64 { return b.val }
+
+type scalarStepIter struct {
+	meta    Metadata
+	step    scalarStep
+	numVals int
```
benraskin92: nit: `numVals, idx int`

src/query/block/scalar.go
```diff
@@ -0,0 +1,162 @@
...
+	meta Metadata
+	vals []float64
+	idx  int
+}
+
+// assert that scalarStepIter implements StepIter
+var _ SeriesIter = (*scalarSeriesIter)(nil)
+
+func (it *scalarSeriesIter) Close()                   { /* No-op*/ }
+func (it *scalarSeriesIter) SeriesCount() int         { return 1 }
```
benraskin92: This is kind of weird- wouldn't this always be `1` in all `SeriesIters`?

src/query/block/scalar_test.go
```diff
@@ -0,0 +1,123 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var (
+	start  = time.Time{}
+	val    = 13.37
```
benraskin92: Can we make this like `100` or something simple?

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,481 @@
...
+
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/executor/transform"
+	"github.com/m3db/m3/src/query/functions/logical"
+	"github.com/m3db/m3/src/query/parser"
+)
+
+const (
+	// *******************************************
+	// * Arithmetic functions
```
benraskin92: What's going on with the `*`s?

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,481 @@
...
+		}
+
+		lVal := scalarL.Value()
+
+		// rhs is a series; use rhs metadata and series meta
+		if !params.RIsScalar {
+			return n.processSingleBlock(
+				rhs,
+				func(x float64) float64 {
+					return n.op.fn(lVal, x)
```
benraskin92: Can you add a comment here?

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,481 @@
...
+		// points in the output, we here replace them with NaNs instead
+		return math.NaN()
+	}
+}
+
+func (n *binaryNode) processBothSeries(
+	lIter, rIter block.StepIter,
+) (block.Block, error) {
+	lMeta, rMeta := lIter.Meta(), rIter.Meta()
+
```
benraskin92: Check to make sure bounds are equal

src/query/functions/binary/node.go
arnikola: Just visual delineation

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,481 @@
...
+	returnBoolFn := funcReturnBool(n.op.fn, n.op.OperatorType, params.ReturnBool)
+
+	for index := 0; lIter.Next() && rIter.Next(); index++ {
+		lStep, err := lIter.Current()
+		if err != nil {
+			return nil, err
+		}
+
+		lValues := lStep.Values()
```
benraskin92: Get the values right before you use them below instead of here

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,481 @@
...
+		rStep, err := rIter.Current()
+		if err != nil {
+			return nil, err
+		}
+
+		rValues := rStep.Values()
```
benraskin92: Same here

src/query/block/scalar.go
```diff
@@ -0,0 +1,162 @@
...
+// This represents constant values; it greatly simplifies downstream operations by
+// allowing them to treat this as a regular block, while at the same time
+// having an option to optimize by accessing the scalar value directly instead
+type Scalar struct {
+	val  float64
+	meta Metadata
+}
+
+// assert that Scalar implements Block
+var _ Block = (*Scalar)(nil)
```
nikunjgit: you don't need it since NewScalar returns Block

src/query/block/scalar.go
```diff
@@ -0,0 +1,162 @@
...
+
+type scalarStepIter struct {
+	meta    Metadata
+	step    scalarStep
+	numVals int
+	idx     int
+}
+
+// assert that scalarStepIter implements StepIter
+var _ StepIter = (*scalarStepIter)(nil)
```
nikunjgit: you don't need it since StepIter() returns & scalarStepIter

src/query/parser/promql/parse.go
```diff
@@ -176,6 +179,18 @@ func (p *parseState) walk(node pql.Node) error {
 		p.transforms = append(p.transforms, opTransform)
 		return nil
 
+	case *pql.NumberLiteral:
+		op := NewScalarOperator(n)
+
```
benraskin92: nit: remove newline

src/query/block/scalar.go
nikunjgit: other places have panics instead of errors since its a way to avoid more codepaths :( I guess we can do either here since its returning error but a lot of times Current() only returns data and panic is the only option.

src/query/parser/promql/parse_test.go
```diff
@@ -63,260 +64,137 @@ func TestDAGWithUnknownOp(t *testing.T) {
...
-	assert.Equal(t, transforms[1].ID, parser.NodeID("1"))
-	assert.Len(t, edges, 1)
-	assert.Equal(t, edges[0].ParentID, parser.NodeID("0"), "fetch should be the parent")
-	assert.Equal(t, edges[0].ChildID, parser.NodeID("1"), "function expr should be the child")
-}
-
-func TestDAGWithAbsentOp(t *testing.T) {
-	q := "absent(http_requests_total{method=\"GET\"})"
-	p, err := Parse(q)
+var linearParseTests = []struct {
```
benraskin92: Nice!!

src/query/block/scalar.go
```diff
@@ -0,0 +1,162 @@
...
+	return &it.step, nil
+}
+
+type scalarStep struct {
+	vals []float64
+	time time.Time
+}
+
+// assert that scalarStepIter implements StepIter
+var _ Step = (*scalarStep)(nil)
```
nikunjgit: don't think you need this

src/query/block/scalar.go
```diff
@@ -0,0 +1,162 @@
...
+func (it *scalarStep) Values() []float64 { return it.vals }
+
+type scalarSeriesIter struct {
+	meta Metadata
+	vals []float64
+	idx  int
+}
+
+// assert that scalarStepIter implements StepIter
+var _ SeriesIter = (*scalarSeriesIter)(nil)
```
nikunjgit: same as above

src/query/block/scalar.go
nikunjgit: same reply as above

src/query/parser/promql/types.go
```diff
@@ -178,6 +239,10 @@ func promVectorCardinalityToM3(card promql.VectorMatchCardinality) logical.Vecto
 }
 
 func promMatchingToM3(vectorMatching *promql.VectorMatching) *logical.VectorMatching {
+	// vectorMatching can be nil iff at least one of the sides is a scalar
```
benraskin92: nit: `s/iff/if`

src/query/block/scalar_test.go
nikunjgit: math.Pi is also an option

src/query/executor/state.go
```diff
@@ -42,13 +42,31 @@ type ExecutionState struct {
...
 }
 
+// CreateScalarSource creates a scalar source node
+func CreateScalarSource(
+	ID parser.NodeID,
+	params ScalarParams,
+	options transform.Options,
+) (parser.Source, *transform.Controller) {
+	controller := &transform.Controller{ID: ID}
+
```
nikunjgit: extra line ?

src/query/parser/promql/types.go
arnikola: It's [intentional](https://en.wikipedia.org/wiki/If_and_only_if)

src/query/executor/transform/types.go
```diff
@@ -47,6 +47,15 @@ type TimeSpec struct {
 	Step time.Duration
 }
 
+// ToBounds transforms a timespec to bounds
+func (ts TimeSpec) ToBounds() block.Bounds {
```
nikunjgit: Bounds() ?

src/query/functions/binary/node.go
nikunjgit: it we're doing the check, should that be in process() ?

src/query/functions/binary/node.go
nikunjgit: cool, was just checking

src/query/functions/binary/node.go
```diff
@@ -0,0 +1,481 @@
...
+}
+
+func (n *binaryNode) cleanup() {
+	n.mu.Lock()
+	defer n.mu.Unlock()
+	n.cache.Remove(n.op.params.LNode)
+	n.cache.Remove(n.op.params.RNode)
+}
+
+const initIndexSliceLength = 10
```
nikunjgit: makes sense to move it to the top ? We might want to put some of these things in the config later.

src/query/block/scalar.go
arnikola: Yeah, if it weren't already able to return an error I would've paniced, but figured might as well make use of the capability here so we don't kill the node :)

src/query/block/scalar.go
arnikola: Looks a bit uglier but ok :p

src/query/functions/binary/node.go
benraskin92: Oh yeah lol

PR #824, opened 2018-08-05
Fix docker Prometheus integration tests

robskillington:
Since we started spinning up the coordinator role as a dedicated role for the Prometheus integration test we need to change the remote read/write endpoint to the coordinator hostname in the Prometheus configuration file.

### DIFF
--- a/scripts/integration-tests/prometheus/prometheus.yml
+++ b/scripts/integration-tests/prometheus/prometheus.yml
@@ -32,10 +32,10 @@ scrape_configs:
 
   - job_name: 'coordinator'
     static_configs:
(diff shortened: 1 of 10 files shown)


scripts/integration-tests/prometheus/prometheus.yml
```diff
@@ -32,10 +32,10 @@ scrape_configs:
 
   - job_name: 'coordinator'
     static_configs:
-      - targets: ['dbnode01:7203']
+      - targets: ['coordinator01:7203']
```
justinjc: nit: rename to query since "coordinator" got changed to "query"?

PR #827, opened 2018-08-07
Use new quay docker repos with m3query, update docs, add m3query to goreleaser config

robskillington:


robskillington:
To fix the Docker images in the documentation being valid, I'm going to merge this now.

PR #828, opened 2018-08-07
Disable TestFiles until can be reliably ran

robskillington:


robskillington:
To help unbreak master failing CI, I'll merge this now.

PR #836, opened 2018-08-09
[1/n] Update m3ninx/segment.Fields()|Terms() to return Iterators

prateek:
1/n in a series of changes to improve performance. Changes included in this PR:
- Return iterators instead of [][]byte for segment.Fields()/Terms()
- Rename `segment/fs` to `segment/fst` to better indicate the segment type


src/m3ninx/index/segment/fst/fst_terms_iterator.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func (o *newFstTermsIterOpts) Close() error {
+	if o.finalizeFST {
+		return o.fst.Close()
+	}
+	return nil
+}
+
+func newFSTTermsIter(opts newFstTermsIterOpts) *fstTermsIter {
```
jeromefroe: nit: the capitalization of fst differs between the function and argument, `newFSTTermsIter` vs. `newFstTermsIterOpts`

src/m3ninx/index/segment/fst/fst_terms_iterator.go
```diff
@@ -0,0 +1,130 @@
...
+func (f *fstTermsIter) copyBytes(b []byte) []byte {
+	// NB: taking a copy of the bytes to avoid referring to mmap'd memory
+	l := len(b)
+	bytes := f.iterOpts.opts.BytesPool().Get(l)
+	bytes = bytes[:l]
+	copy(bytes, b)
+	return bytes
+}
+
+func (f *fstTermsIter) CurrentExtended() ([]byte, uint64) {
```
jeromefroe: do you use this frequently? just curious if it would make sense to make this `Current` and have the caller set the uint64 to an empty variable when its not needed

prateek: nah, it's currently not used at all.

src/m3ninx/index/segment/fst/fst_terms_iterator.go
prateek: cheers, fixed in next patch

PR #837, opened 2018-08-09
[2/n] Cache regexp fst automata construction

prateek:
- Only construct a regexp dfa once per regexp query.

### DIFF
--- a/src/m3ninx/index/segment/fst/segment.go
+++ b/src/m3ninx/index/segment/fst/segment.go
@@ -297,7 +297,7 @@ func (r *fsSegment) MatchTerm(field []byte, term []byte) (postings.List, error) 
 	return pl, nil
 }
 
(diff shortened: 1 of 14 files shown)


src/m3ninx/index/segment/fst/segment.go
```diff
@@ -298,16 +297,24 @@ func (r *fsSegment) MatchTerm(field []byte, term []byte) (postings.List, error)
 	return pl, nil
 }
 
-func (r *fsSegment) MatchRegexp(field []byte, regexp []byte, compiled *regexp.Regexp) (postings.List, error) {
+func (r *fsSegment) MatchRegexp(field []byte, regexp []byte, compiled *index.CompiledRegex) (postings.List, error) {
```
jeromefroe: Why pass `compiled` as a pointer? Since the internal fields are already pointers would it make sense to pass the struct as a value?

prateek: Sure, will update.

PR #838, opened 2018-08-09
fix k8s storage urls

poblahblahblah:
This fixes the k8s storage urls in the docs.


robskillington:
Thanks a lot for catching this, hopefully at some point we can find a way of writing an integration test to avoid this regressing in the future.  Appreciated!

PR #840, opened 2018-08-11
Use godownloader to pin gometalinter

prateek:


scripts/install-gometalinter.sh
```diff
@@ -0,0 +1,389 @@
+# Taken from: https://github.com/alecthomas/gometalinter/blob/master/scripts/install.sh
```
justinjc: The shebang (line below) needs to be the first line of the file for it to work. Maybe swap these first two lines.

prateek: Sure, done.

PR #842, opened 2018-08-13
[fix] Link to m3coordinator-cluster-template.yml

lookfirst:


robskillington:
Thanks for this @lookfirst, as mentioned with the last documentation fix we'll have to see if we can find some way to write some tests that take certain variables from the documentation and ensure the guide is up to date by running those tests on each commit.

Thanks for this fix.

PR #845, opened 2018-08-15
Add index compaction planner

prateek:


src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,240 @@
...
+)
+
+var (
+	errMutableCompactionAgeNegative = errors.New("mutable compaction age must be postive")
+	errSizeBucketsUndefined         = errors.New("size buckets are undefined")
+	errMaxImmutableCompactionSize   = errors.New("max immutable compaction size must be positive")
+)
+
+var (
+	DefaultSizeBuckets = []SizeBucket{ // i.e. tiers for compaction [0, 262K), [262K, 524K), [524K, 1M), [1M, 2M), [2M, 8M)
```
robskillington: Leveled compaction 👍

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,240 @@
...
+		SizeBucket{
+			MinSizeInclusive: 1 << 21,
+			MaxSizeExclusive: 1 << 23,
+		},
+	}
+
+	DefaultOptions = PlannerOptions{
+		MaxImmutableCompactionSize: 1 << 22,                            // ~4M
+		MaxMutableSegmentSize:      1 << 16,                            // 64K
+		MutableCompactionAge:       15 * time.Second,                   // any mutable segment 15s or older is eligible for compactions
```
robskillington: I wonder if in the future we can just get rid of this option and if we ever have a mutable segment, always start compacting.

robskillington: Would setting an age of 0 here make this work like so?

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,240 @@
...
+			MaxSizeExclusive: 1 << 21,
+		},
+		SizeBucket{
+			MinSizeInclusive: 1 << 21,
+			MaxSizeExclusive: 1 << 23,
+		},
+	}
+
+	DefaultOptions = PlannerOptions{
+		MaxImmutableCompactionSize: 1 << 22,                            // ~4M
```
robskillington: Hm, shouldn't this just be computed from the size buckets?  Doesn't seem to buy too much by having another max size other than the upper limit of the largest size bucket?  I see that this is a value in the last default bucket itself, but couldn't you just set the last bucket to 2m to 4m and then not need this number at all?

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,240 @@
...
+	// 1st phase - find all compactable segments
+	var compactableSegments []Segment
+	for _, seg := range candidateSegments {
+		compactable := (seg.Type == segments.FSTType && seg.Size < opts.MaxImmutableCompactionSize) ||
+			(seg.Type == segments.MutableType &&
+				(seg.Age >= opts.MutableCompactionAge || seg.Size >= opts.MaxMutableSegmentSize))
+		if compactable {
+			compactableSegments = append(compactableSegments, seg)
+		} else {
+			plan.UnusedSegments = append(plan.UnusedSegments, seg)
```
robskillington: Hm should probably also preallocate the UnusedSegments slice to have capacity `len(candidateSegments) too.

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// if we don't have any compactable segments, we can early terminate
+	if len(compactableSegments) == 0 {
+		return plan, nil
+	}
+
+	// now we have segments to compact, so on to phase 2
+	buckets := opts.SizeBuckets
+	sort.Sort(ByMinSize(buckets))
```
robskillington: Is this a race if you have multiple planners as it mutates the options?  Or are we only ever having one planner than then can spawn parallel executors?  Either way might be safer to copy the size buckets rather than mutate by reference here with the sort.

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	// now that we have the plan, we priortise the tasks as requested in the opts. (2c)
+	sort.Stable(plan)
+	return plan, nil
+}
+
+func (p *Plan) Len() int      { return len(p.Tasks) }
+func (p *Plan) Swap(i, j int) { p.Tasks[i], p.Tasks[j] = p.Tasks[j], p.Tasks[i] }
+func (p *Plan) Less(i, j int) bool {
```
robskillington: This is nice 👍

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,240 @@
...
+			return taskSummaryi.NumMutable > taskSummaryj.NumMutable
+		}
+		// i.e. smaller tasks over bigger ones
+		return taskSummaryi.CumulativeSize < taskSummaryj.CumulativeSize
+	}
+}
+
+func (o PlannerOptions) Validate() error {
+	if o.MutableCompactionAge <= 0 {
+		return errMutableCompactionAgeNegative
```
robskillington: Hm, we should probably allow compaction age of zero though no?  Maybe just reduce the check to `< 0`?

src/dbnode/storage/index/compaction/plan.go
prateek: As things are, it's definitely specifying the same thing as controlling the last bucket size. The original intention with the knob was to have an override which ensures we don't compact things that are too large, regardless of the bucket size. It'd be useful in cases where we want to do a non-leveled compaction (e.g. once we know a block is sealed, we want to compact as much as possible).

Happy to revert for now if you'd prefer, can add back in once we add the different planner types.

src/dbnode/storage/index/compaction/plan.go
prateek: I tried the equivalent - creating a mutable segment every time a write batch came in, and triggering a compaction - it ended up creating a lot small segments and the test nodes compactions couldn't keep up. A buffer of even 5s helped this a lot. 

Is the concern e2e latency?

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,240 @@
...
+	// (2) Come up with a logical plan for compactable segments
+	//    (a) Group the segments into given buckets (compactions can only be performed for segments within the same bucket)
+	//    (b) For each bucket:
+	//    (b1) Accumulate segments until cumulative size is over the max of the current bucket.
+	//    (b2) Add a Task which comprises segments from (b1) to the Plan.
+	//    (b3) Continue (b1) until the bucket is empty.
+	//    (c) Priotize Tasks w/ "compactable" Mutable Segments over all others
+
+	// 1st phase - find all compactable segments
+	var compactableSegments []Segment
```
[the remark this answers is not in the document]
prateek: sure thing

src/dbnode/storage/index/compaction/plan.go
prateek: sure, i'll lazily allocate

src/dbnode/storage/index/compaction/plan.go
prateek: That's fair, will do.

src/dbnode/storage/index/compaction/plan.go
prateek: Sure thing, don't think it'd ever happen but definitely more correct to say `< 0`

src/dbnode/storage/index/compaction/plan.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/compaction/plan.go
prateek: done

src/dbnode/storage/index/compaction/plan.go
prateek: done

src/dbnode/storage/index/compaction/plan.go
prateek: done

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,250 @@
...
+			MinSizeInclusive: 1 << 21,
+			MaxSizeExclusive: 1 << 23,
+		},
+	}
+
+	// DefaultOptions are the default compaction PlannerOptions.
+	DefaultOptions = PlannerOptions{
+		MaxImmutableCompactionSize: 1 << 22,                            // ~4M
+		MaxMutableSegmentSize:      1 << 16,                            // 64K
+		MutableCompactionAge:       15 * time.Second,                   // any mutable segment 15s or older is eligible for compactions
```
robskillington: Ah interesting.  Yeah end to end latency is the primary concern, also at very low volume it's annoying that you have to wait a fixed amount for the series to become visible.  Ideally we should be as fast as possible and only once we are performing substantially high workloads do we slow down.

Could we do something like if there's no mutable segment compactions happening then kick one off, but otherwise buffer?  So that we're doing it "as fast as possible" but should also avoid tons of tiny segments?

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,250 @@
...
+		},
+		Level{
+			MinSizeInclusive: 1 << 21,
+			MaxSizeExclusive: 1 << 23,
+		},
+	}
+
+	// DefaultOptions are the default compaction PlannerOptions.
+	DefaultOptions = PlannerOptions{
+		MaxImmutableCompactionSize: 1 << 22,                            // ~4M
```
robskillington: Yeah could we remove this for now and add it back in when introducing different planner types?

prateek: sure thing, done in next patch.

src/dbnode/storage/index/compaction/plan.go
prateek: I really like it. I've been mulling a way to make a feedback loop in the compaction process by modelling `f(cost of compaction, compaction budget, latency)` but the version you suggested is a great start. 

I'll make the following changes in the PR:
- Decompose identifying "compactable" segments from the creation of the plan 
- Add `func (s Segment) Compactable(opts PlannerOptions) bool` (w/ the same conditions as they are in `NewPlan` currently) 

We can land this PR after that. In #819, I'll make `indexBlock.triggerCompactions` consider both `segment.Compactible` and whether the there's an on-going compaction

prateek: done

src/dbnode/storage/index/compaction/plan.go
prateek: done

src/dbnode/storage/index/compaction/plan.go
```diff
@@ -0,0 +1,278 @@
...
+		fallthrough
+	default:
+		// NB: the intent with the conditions below is to optimise for e2e ingest latency first,
+		// which is why we prefer to compact older mutable segments first, then any larger ones,
+		// after which, we fall back to the graceful plan of compacting smaller segments over
+		// larger ones to reduce total compactions required.
+		taskSummaryi, taskSummaryj := p.Tasks[i].Summary(), p.Tasks[j].Summary()
+		if taskSummaryi.CumulativeMutableAge != taskSummaryj.CumulativeMutableAge {
+			// i.e. put those tasks which have cumulative age greater first
+			return taskSummaryi.CumulativeMutableAge > taskSummaryj.CumulativeMutableAge
```
robskillington: Hm, this method should return true if `i < j` no?  Should this not be `return taskSummaryi.CumulativeMutableAge M taskSummaryj.CumulativeMutableAge`?

PR #846, opened 2018-08-15
Add env var listen address support

benraskin92:


glide.yaml
```diff
@@ -1,7 +1,7 @@
 package: github.com/m3db/m3
 import:
 - package: github.com/m3db/m3x
-  version: 994ab3222cc45bdb1b439f96d37221651b24feca
+  version: 49a94422a02501a99c1ae6c1a632f83aea9efad6
```
prateek: need to update glide.lock too

src/query/server/server.go
```diff
@@ -243,11 +243,16 @@ func Run(runOpts RunOptions) {
 	}
 	handler.RegisterRoutes()
 
-	logger.Info("starting server", zap.String("address", cfg.ListenAddress))
+	listenAddress, err := cfg.ListenAddress.Resolve()
+	if err != nil {
+		logger.Fatal("unable to get listen address", zap.Any("error", err))
```
prateek: nit: change this to `zap.Error(err)`

src/query/server/server.go
```diff
@@ -243,11 +243,16 @@ func Run(runOpts RunOptions) {
...
+	}
+
+	logger.Info("starting server", zap.String("address", listenAddress))
 	go func() {
-		if err := http.ListenAndServe(cfg.ListenAddress, handler.Router); err != nil {
+		if err := http.ListenAndServe(listenAddress, handler.Router); err != nil {
 			logger.Fatal("unable to serve on listen address",
-				zap.Any("address", cfg.ListenAddress), zap.Any("error", err))
+				zap.Any("address", listenAddress), zap.Any("error", err))
```
prateek: same here, mind changing to `Zap.String("address", ...)` and `zap.Error(err)`

glide.yaml
benraskin92: Yeah was wondering why literally every single Semaphore test was failing.. Updated now though

kube/bundle.yaml
```diff
@@ -114,7 +114,11 @@ metadata:
 data:
   m3dbnode.yml: |+
     coordinator:
```
arnikola: Should we change this to `query`?

robskillington:
Hey instead of getting rid of `listenAddress` can you just make it so that if the environment variable is set it will listen on `"0.0.0.0:$ENV_LISTEN_PORT"` or if you also set `$ENV_LISTEN_HOST` it then sets listen of `"$ENV_LISTEN_HOST:$ENV_LISTEN_PORT"`?

It's much less desirable to have to list the listen host and port separately.  

See this example which I'm fixing `m3ctl` to do as I describe:
https://github.com/m3db/m3ctl/pull/73

glide.yaml
```diff
@@ -214,4 +214,4 @@ testImport:
   version: 3677f62bb30dbf3b042c4c211245d072aa9ee075
 
 - package: github.com/leanovate/gopter
-  version: ^0.2
+  version: f778776473e0ef7764e1434dd01a61cc1ec574b4
```
benraskin92: I guess they updated this dependency and when using `^0.2` it breaks a ton of integration tests so pinning to a specific sha for now.

PR #850, opened 2018-08-17
Consolidate binary and logical functions

arnikola:
Consolidate binary and logical functions into a single package. Reduce reused code and generalize creation of binary functions, making creation of any future binary functions easier.

### DIFF
--- a/src/query/functions/binary/base.go
+++ b/src/query/functions/binary/base.go
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package logical
+package binary
 
 import (
 	"fmt"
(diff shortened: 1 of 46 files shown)


src/query/functions/binary/binary.go
```diff
@@ -127,98 +142,47 @@ var (
...
+		return processBinary(lhs, rhs, params, controller, fn)
 	}
-
-	defer nextBlock.Close()
-	return n.controller.Process(nextBlock)
 }
 
 // processes two logical blocks, performing a logical operation on them
-func (n *binaryNode) process(lhs, rhs block.Block) (block.Block, error) {
+func processBinary(
```
nikunjgit: the signature could become simpler if this was defined on the node object as before.

src/query/functions/binary/binary.go
```diff
@@ -113,12 +114,26 @@ var (
...
+		ExpType + returnBoolSuffix:      math.Pow,
+
+		EqType:        func(x, y float64) float64 { return toComparisonValue(x == y, x) },
+		NotEqType:     func(x, y float64) float64 { return toComparisonValue(x != y, x) },
+		GreaterType:   func(x, y float64) float64 { return toComparisonValue(x > y, x) },
+		LesserType:    func(x, y float64) float64 { return toComparisonValue(x < y, x) },
+		GreaterEqType: func(x, y float64) float64 { return toComparisonValue(x >= y, x) },
+		LesserEqType:  func(x, y float64) float64 { return toComparisonValue(x <= y, x) },
+
+		EqType + returnBoolSuffix:        func(x, y float64) float64 { return toFloat(x == y) },
```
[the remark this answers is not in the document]
arnikola: That's how I had it originally, but I'd prefer to incur a small cost here in adding unnecessary types to the map than go through an additional function call for every value in the series

src/query/functions/binary/binary.go
```diff
@@ -127,98 +142,47 @@ var (
...
-
-	nextBlock, err := n.process(lhs, rhs)
-	if err != nil {
-		return err
+// Build the binary processing step
+func createBinaryProcessingStep(
+	params NodeParams,
+	fn mathFunc,
+) func(block.Block, block.Block, *transform.Controller) (block.Block, error) {
+	return func(lhs, rhs block.Block, controller *transform.Controller) (block.Block, error) {
```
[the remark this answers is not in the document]
arnikola: This way I can have a more generic base node which can accept both logical and binary functions

src/query/functions/binary/binary.go
arnikola: I'd prefer a more complex function generator signature in exchange for a simpler node

src/query/functions/binary/binary.go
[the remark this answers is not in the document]
nikunjgit: problem now is that its easy to make a mistake by adding the type but not adding type + returnBoolSuffix. Also, you're writing the same logic twice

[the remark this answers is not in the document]
nikunjgit: also you can't really avoid additional call. Even here you're calling toComparisonValue

src/query/functions/binary/binary.go
nikunjgit: but then you're passing the same arguments multiple layers deep. This provides an interesting comparison: https://stackoverflow.com/questions/36904955/passing-single-object-vs-passing-multiple-parameters

src/query/functions/binary/binary.go
```diff
@@ -347,49 +275,34 @@ func (n *binaryNode) processSingleBlock(
...
-		}
-		// NB(arnikola): whereas prometheus does not include not matching
-		// points in the output, we here replace them with NaNs instead
-		return math.NaN()
-	}
-}
-
-func (n *binaryNode) processBothSeries(
+func processBothSeries(
 	lIter, rIter block.StepIter,
```
nikunjgit: this functions has 6 arguments now. Consider object ? https://softwareengineering.stackexchange.com/questions/145055/are-there-guidelines-on-how-many-parameters-a-function-should-accept

src/query/functions/binary/binary.go
```diff
@@ -22,14 +22,10 @@ package binary
 
 import (
```
nikunjgit: binary/binary.go is a bit strange. Consider renaming ?

src/query/functions/binary/base.go
```diff
@@ -29,76 +29,76 @@ import (
...
-	case UnlessType:
-		makeBlock = makeUnlessBlock
-	default:
-		return logicalOp{}, fmt.Errorf("operator not supported: %s", opType)
+	fn, ok := buildLogicalFunction(opType, params)
+	if !ok {
+		fn, ok = buildArithmeticFunction(opType, params)
+		if !ok {
+			fn, ok = buildComparisonFunction(opType, params)
+			if !ok {
```
benraskin92: Anyway to clean this code up a bit? Looks odd having 3 nested `if` statements.

src/query/functions/binary/binary.go
```diff
@@ -0,0 +1,234 @@
...
+	controller *transform.Controller,
+	fn binaryFunc,
+) (block.Block, error) {
+	lIter, err := lhs.StepIter()
+	if err != nil {
+		return nil, err
+	}
+
+	// fn := n.op.fn
+	// params := n.op.params
```
benraskin92: Remove?

src/query/functions/binary/base.go
arnikola: I tried a few things but unfortunately there's not many ways to add sugar to functions that return 2 values. Best alternative I came up with was something like 
```
fn, ok = buildArithmeticFunction(opType, params)
		if !ok {
			fn, ok = buildBinaryFunction(opType, params)
                        if !ok {
                          return error
```
and have buildBinaryFunction try building an Arithmetic then a Comparison function, but that doesn't achieve much except for kicking the can down the road

src/query/functions/binary/binary.go
arnikola: Oops, missed this. Will do it in a future commit

PR #852, opened 2018-08-21
Add non-parameterized PromQL aggregation functions

arnikola:
Adds non-parameterized PromQL aggregation functions

- Refactors existing aggregation function (count) to a generic aggregation node
- Add ability to group aggregations by/without given tags
- Adds sum, min, max, avg, stddev, stdvar aggregation functions
- Adds test suites for aggregation functions
- Adds parsing for aggregation functions

TODO (likely a separate pr):

- Add parameterized PromQL aggregation functions (count_values, bottomk, topk, quantile)


src/query/functions/aggregation/base.go
```diff
@@ -0,0 +1,203 @@
...
+// Input series that exist in an index list are mapped to the
+// relevant index in the combined series meta.
+func collectSeries(params NodeParams, opName string, metas []block.SeriesMeta) ([][]int, []block.SeriesMeta) {
+	without, matching := params.Without, params.Matching
+
+	var idFunc withKeysID
+	var tagsFunc withKeysTags
+	if without {
+		idFunc = excludeKeysID
+		tagsFunc = excludeKeysTags
```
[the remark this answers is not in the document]
arnikola: The problem with going that route is that I'd need to generate the tags on each loop to get the ID, which can be unnecessary if the pruned tags are already present in the tag map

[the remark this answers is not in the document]
nikunjgit: Ah, makes sense

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,224 @@
...
+// THE SOFTWARE.
+
+package aggregation
+
+import (
+	"math"
+)
+
+// SumType adds all non nan elements in a list of series
+const SumType = "sum"
```
benraskin92: Put all of the `const` at the top of the file.

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,224 @@
...
+	return sumByIndices
+}
+
+// MinType takes the minimum all non nan elements in a list of series
+const MinType = "min"
+
+func minFn(values []float64, indices [][]int) []float64 {
+	minByIndices := make([]float64, len(indices))
+	for i, indexList := range indices {
+		min := math.NaN()
```
benraskin92: Is this logic straight from Prom?

src/query/functions/aggregation/base.go
```diff
@@ -0,0 +1,203 @@
...
+
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/executor/transform"
+	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/parser"
+)
+
+// NodeParams contains additional parameters required for aggregation ops
+type NodeParams struct {
+	Matching []string
```
benraskin92: What happens if you add the same tag twice? Should this be a set instead?

src/query/functions/aggregation/base.go
```diff
@@ -0,0 +1,203 @@
...
+			val.indices = append(val.indices, i)
+		} else {
+			tagMap[id] = &tagMatch{
+				indices: []int{i},
+				tags:    tagsFunc(meta.Tags, matching),
+			}
+		}
+	}
+
+	collectedIndices := make([][]int, len(tagMap))
```
benraskin92: Thoughts on renaming to `groupedIndices`, etc. instead?

arnikola: Makes more sense

src/query/functions/aggregation/base.go
arnikola: Tested it out and it's only used once; will make an explicit test for this

src/query/functions/aggregation/function.go
arnikola: Not sure what you mean; I'm adding the functions from 
https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators but using my own logic for them

src/query/functions/aggregation/function.go
arnikola: Figured it would be better to have them next to the functions, but adding the const block at the top makes sense for quickly reading it

src/query/functions/aggregation/base.go
```diff
@@ -18,58 +18,100 @@
...
+	MaxType:               maxFn,
+	AverageType:           averageFn,
+	StandardDeviationType: stddevFn,
+	StandardVarianceType:  varianceFn,
+	CountType:             countFn,
+}
+
+// NodeParams contains additional parameters required for aggregation ops
+type NodeParams struct {
+	Matching []string
```
benraskin92: `MatchingTags`?

benraskin92: Also can you add comment explaining what these two fields represent?

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,220 @@
...
+			}
+		}
+
+		minByIndices[i] = min
+	}
+
+	return minByIndices
+}
+
+func maxFn(values []float64, indices [][]int) []float64 {
```
benraskin92: `maxFn` and `minFn` are pretty similar, maybe extract it (or at least some part) into common function?

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	return maxByIndices
+}
+
+func averageFn(values []float64, indices [][]int) []float64 {
+	averageByIndices := make([]float64, len(indices))
+	for i, indexList := range indices {
+		sum := 0.0
+		count := 0.0
```
benraskin92: Make count an `int`?

arnikola: Would have to cast it to a float anyway for division below

src/query/functions/aggregation/function_test.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+var fnTest = []struct {
+	name      string
+	values    []float64
+	indicies  [][]int
+	functions []funcTest
+}{
+	{
+		"empty", []float64{}, [][]int{}, []funcTest{
```
benraskin92: Nice test setup!

src/query/functions/aggregation/function_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	"github.com/m3db/m3/src/query/parser"
+	"github.com/m3db/m3/src/query/test"
+	"github.com/m3db/m3/src/query/test/executor"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestCountWithAllValues(t *testing.T) {
```
benraskin92: Do we need these two tests anymore?

src/query/functions/aggregation/base_test.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/parser"
+	"github.com/m3db/m3/src/query/test"
+	"github.com/m3db/m3/src/query/test/executor"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFunctionWithFiltering(t *testing.T) {
```
benraskin92: Add a test with multiple matching tags here too? Just to be safe

src/query/functions/utils/group.go
```diff
@@ -0,0 +1,99 @@
...
+
+	type tagMatch struct {
+		indices []int
+		tags    models.Tags
+	}
+
+	tagMap := make(map[uint64]*tagMatch)
+	for i, meta := range metas {
+		id := idFunc(meta.Tags, matching)
+		if val, ok := tagMap[id]; ok {
```
benraskin92: Add some comments in this function to make it clearer what you're doing

src/query/functions/aggregation/base_test.go
arnikola: There's one in the matching tests; since it uses that logic here it should be fine

src/query/functions/utils/group_test.go
```diff
@@ -0,0 +1,276 @@
...
+			{"b": "2", "c": "4"},
+			{"b": "2"},
+			{"b": "2", "c": "3"},
+			{"b": "2", "d": "3"},
+			{"c": "d"},
+		},
+	},
+	{
+		"same tag matching",
+		[]string{"a", "a"},
```
benraskin92: 👍

src/query/functions/aggregation/function.go
arnikola: I'd rather keep it as is and have a few extra lines of code to avoid doing an extra comparison on every step of the aggregation

src/query/models/tag.go
```diff
@@ -184,6 +184,25 @@ func (t Tags) IDWithExcludes(excludeKeys ...string) uint64 {
 	return h.Sum64()
 }
 
+// TagsWithoutKeys returns only the tags which do not have the given keys
+func (t Tags) TagsWithoutKeys(excludeKeys []string) Tags {
+	excludeTags := make(Tags)
+	for k, v := range t {
```
benraskin92: Is this part of the function the same as `WithoutName()` below?

src/query/functions/aggregation/base.go
```diff
@@ -18,58 +18,100 @@
...
-	builder, err := c.controller.BlockBuilder(stepIter.Meta(), []block.SeriesMeta{meta})
+	params := n.op.params
+	indices, metas := utils.GroupSeries(
+		params.Matching,
+		params.Without,
+		n.op.opType,
+		stepIter.SeriesMeta(),
+	)
+
+	builder, err := n.controller.BlockBuilder(stepIter.Meta(), metas)
```
nikunjgit: shouldn't the metadata be different ? I think aggregation removes a lot of the metadata.

src/query/functions/aggregation/base_test.go
```diff
@@ -0,0 +1,104 @@
...
+		{math.NaN(), 6, 7, 8, 9},
+		{10, 20, 30, 40, 50},
+		{50, 60, 70, 80, 90},
+		{100, 200, 300, 400, 500},
+		{600, 700, 800, 900, 1000},
+	}
+
+	bl := test.NewBlockFromValuesWithSeriesMeta(bounds, seriesMetas, v)
+	c, sink := executor.NewControllerWithSink(parser.NodeID(1))
+	op, err := NewAggregationOp(StandardDeviationType, NodeParams{Matching: []string{"a"}, Without: false})
```
nikunjgit: also test Without: true case ?

src/query/functions/aggregation/base.go
```diff
@@ -18,58 +18,100 @@
...
-
+func (n *baseNode) Process(ID parser.NodeID, b block.Block) error {
 	stepIter, err := b.StepIter()
 	if err != nil {
 		return err
 	}
 
-	builder, err := c.controller.BlockBuilder(stepIter.Meta(), []block.SeriesMeta{meta})
+	params := n.op.params
+	indices, metas := utils.GroupSeries(
```
nikunjgit: indices -> buckets ?

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,220 @@
...
+	// in a list of series
+	StandardDeviationType = "stddev"
+	// StandardVarianceType takes the standard variance of all non nan elements
+	// in a list of series
+	StandardVarianceType = "var"
+	// CountType counts all non nan elements in a list of series
+	CountType = "count"
+)
+
+func sumFn(values []float64, indices [][]int) []float64 {
```
nikunjgit: indices -> buckets ?

src/query/functions/aggregation/base.go
```diff
@@ -0,0 +1,140 @@
...
+	MaxType:               maxFn,
+	AverageType:           averageFn,
+	StandardDeviationType: stddevFn,
+	StandardVarianceType:  varianceFn,
+	CountType:             countFn,
+}
+
+// NodeParams contains additional parameters required for aggregation ops
+type NodeParams struct {
+	// MatchingTags is the set of tags by which the aggregation groups output series
```
benraskin92: 👍

src/query/functions/aggregation/function.go
benraskin92: Really? I thought you can divide float by count. Oh well- not a big deal.

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	return sum
+}
+
+func minFn(values []float64, buckets []int) float64 {
+	min := math.NaN()
+	for _, idx := range buckets {
+		v := values[idx]
+		if !math.IsNaN(v) {
```
nikunjgit: reduce nesting by flipping ?  if math.IsNaN(v) {continue}

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,182 @@
...
+			if math.IsNaN(min) {
```
nikunjgit: if math.IsNaN(min) || min > v ?

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	return min
+}
+
+func maxFn(values []float64, buckets []int) float64 {
+	max := math.NaN()
+	for _, idx := range buckets {
+		v := values[idx]
+		if !math.IsNaN(v) {
```
nikunjgit: same comments as above. Also this is so similar to the previous one, that you can think about refactoring them.

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,182 @@
...
+					max = v
+				}
+			}
+		}
+	}
+
+	return max
+}
+
+func averageFn(values []float64, buckets []int) float64 {
```
nikunjgit: a lot of code duplication in average, stddev, variance, etc...consider creating a function which returns sum and count, that can be reused at a lot of these places

arnikola: I'd rather have a bit of extra code in return for not having to loop through the list twice

src/query/functions/aggregation/function.go
arnikola: That's much nicer, good call

src/query/functions/aggregation/base.go
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	for index := 0; stepIter.Next(); index++ {
+		step, err := stepIter.Current()
+		if err != nil {
+			return err
+		}
+
+		values := step.Values()
+		aggregatedValues := make([]float64, len(buckets))
```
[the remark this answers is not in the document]
arnikola: Good call, didn't need to think about that before :p

src/query/functions/aggregation/function.go
arnikola: Got what you mean, changed it

src/query/functions/aggregation/function.go
arnikola: I prefer not to use `continues`, especially for such a small gain in nesting reduction after the other change you suggested

src/query/functions/aggregation/function.go
arnikola: It wasn't pushed

nikunjgit: No worries, you said it was so I got confused

src/query/functions/aggregation/base.go
```diff
@@ -0,0 +1,144 @@
...
+
+	params := n.op.params
+	buckets, metas := utils.GroupSeries(
+		params.MatchingTags,
+		params.Without,
+		n.op.opType,
+		stepIter.SeriesMeta(),
+	)
+
+	builder, err := n.controller.BlockBuilder(stepIter.Meta(), metas)
```
nikunjgit: will stepIter.Meta() change ? What if the aggregation is over something which is common across all series ?

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,145 @@
...
+	// in a list of series
+	StandardDeviationType = "stddev"
+	// StandardVarianceType takes the standard variance of all non nan elements
+	// in a list of series
+	StandardVarianceType = "var"
+	// CountType counts all non nan elements in a list of series
+	CountType = "count"
+)
+
+func sumAndAverage(values []float64, buckets []int) (float64, float64) {
```
nikunjgit: sumAndCount?

src/query/functions/aggregation/function.go
nikunjgit: why not ? lol  
I saw @robskillington even using fallthrough in one of his diffs.

src/query/functions/aggregation/function.go
```diff
@@ -0,0 +1,145 @@
...
+	// Cannot take average of no values
+	if count == 0 {
+		return math.NaN()
+	}
+
+	return sum / count
+}
+
+// Since stddev is just sqrt(variance), handle in one function
+func stddevOrVariance(
```
nikunjgit: would be cleaner to just return variance, count and stddevFn can perform the remaining calculation

src/query/functions/aggregation/function.go
nikunjgit: Also note that the compiler should be smart enough with the branch prediction so really don't see any harm with continues.

src/query/functions/aggregation/base.go
arnikola: Added flatten/dedupe steps here

src/query/functions/aggregation/function.go
arnikola: Style preferences? Find continues harder to follow, and unnecessary for saving one level of indentation for 3 lines

src/query/functions/aggregation/base.go
```diff
@@ -0,0 +1,148 @@
...
+// THE SOFTWARE.
+
+package aggregation
+
+import (
+	"fmt"
+
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/executor/transform"
+	"github.com/m3db/m3/src/query/functions/logical"
```
nikunjgit: does it make sense for aggregation to have dependancy on logical ? Maybe move those functions out of logical ?

src/query/models/tag.go
```diff
@@ -184,6 +184,18 @@ func (t Tags) IDWithExcludes(excludeKeys ...string) uint64 {
 	return h.Sum64()
 }
 
+// TagsWithoutKeys returns only the tags which do not have the given keys
+func (t Tags) TagsWithoutKeys(excludeKeys []string) Tags {
```
nikunjgit: Maybe just implement this and change WithoutName to call TagsWithoutKeys([]string{MetricName})

src/query/models/tag.go
```diff
@@ -204,6 +216,18 @@ func (t Tags) IDWithKeys(includeKeys ...string) uint64 {
 	return h.Sum64()
 }
 
+// TagsWithKeys returns only the tags which have the given keys
+func (t Tags) TagsWithKeys(includeKeys []string) Tags {
+	includeTags := make(Tags)
```
nikunjgit: make(Tags, len(includeKeys)) ?

src/query/functions/aggregation/function.go
nikunjgit: okay

src/query/functions/aggregation/base.go
arnikola: https://github.com/m3db/m3/pull/850 :)

src/query/models/tag.go
arnikola: Wouldn't that lead to weird things where whenever we call this from e.g. aggregation, we'd need to explicitly include MetricName in the list of tags to exclude?

In any case, your tag change should remove the need for this

PR #853, opened 2018-08-23
Improve speed of DataFileSetExistsAt

richardartoul:
- [X] Improve speed of DataFileSetExistsAt by just verifying if the checkpoint file exists (as opposed to  the existing code which uses a much more expensive codepath requiring use of filepath.Glob)
- [X] Refactor FileExists (and dependent code) to return errors

### DIFF
--- a/src/dbnode/integration/disk_cleanup_helpers.go
+++ b/src/dbnode/integration/disk_cleanup_helpers.go
@@ -178,7 +178,11 @@ func waitUntilDataCleanedUpExtended(
 func waitUntilNamespacesCleanedUp(filePathPrefix string, namespace ident.ID, waitTimeout time.Duration) error {
 	dataCleanedUp := func() bool {
 		namespaceDir := fs.NamespaceDataDirPath(filePathPrefix, namespace)
-		return !fs.FileExists(namespaceDir)
+		exists, err := fs.FileExists(namespaceDir)
+		if err != nil {
+			panic(err)
+		}
+		return !exists
 	}
 
 	if waitUntil(dataCleanedUp, waitTimeout) {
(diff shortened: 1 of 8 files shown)


src/dbnode/persist/fs/files.go
```diff
@@ -468,7 +468,11 @@ func forEachInfoFile(
 			digestsFilePath = filesetPathFromTimeAndIndex(dir, t, volume, digestFileSuffix)
 			infoFilePath = filesetPathFromTimeAndIndex(dir, t, volume, infoFileSuffix)
 		}
-		if !FileExists(checkpointFilePath) {
+		checkpointExists, err := FileExists(checkpointFilePath)
+		if err != nil {
+			continue
```
richardartoul: This kind of sucks, but thats how we handle errors in this function right now. Should address in a separate P.R

richardartoul: #854

src/dbnode/integration/disk_cleanup_helpers.go
```diff
@@ -89,7 +89,10 @@ type cleanupTimesCommitLog struct {
 
 func (c *cleanupTimesCommitLog) anyExist() bool {
 	for _, t := range c.times {
-		_, index := fs.NextCommitLogsFile(c.filePathPrefix, t)
+		_, index, err := fs.NextCommitLogsFile(c.filePathPrefix, t)
+		if err != nil {
+			panic(err)
```
robskillington: Hm.. panic is a bad smell.  Is this going to get fixed by address #854 too?   

If so we should put `// https://github.com/m3db/m3/issues/854 - This requires a high priority fix to correctly handle these errors` next to each of these occurrences.  I'm still kind of wary about adding any sort of panics even as a short term fix =/

src/dbnode/persist/fs/files.go
robskillington: Sounds good.

PR #855, opened 2018-08-23
Add json write endpoint

benraskin92:


src/query/api/v1/handler/json/write.go
```diff
@@ -0,0 +1,115 @@
...
+	return &WriteJSONHandler{
+		store: store,
+	}
+}
+
+// WriteQuery represents the write request from the user
+// NB(braskin): support only writing one datapoint for now
+type WriteQuery struct {
+	Tags      map[string]string `json:"tags" validate:"nonzero"`
+	Timestamp int               `json:"timestamp" validate:"nonzero"`
```
benraskin92: Is this fine or should I make it take a `ts.Datapoint`?

nikunjgit: Instead of Timestamp int, can you use the same thing we use in /prometheus/native/read endpoint ? We have parseTime in m3/src/query/api/v1/handler/prometheus/native/common.go which you can probably take out

src/query/api/v1/handler/json/write.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/m3db/m3/src/query/ts"
+	"github.com/m3db/m3/src/query/util/logging"
+	xtime "github.com/m3db/m3x/time"
+
+	"go.uber.org/zap"
+)
+
+const (
+	// WriteJSONURL is the url for the write json handler
+	WriteJSONURL = handler.RoutePrefixV1 + "/write_json"
```
nikunjgit: handler.RoutePrefixV1 + "/json/write" ?

This way, we can later add handler.RoutePrefixV1 + "/json/read"

src/query/api/v1/handler/json/write.go
```diff
@@ -0,0 +1,115 @@
...
+				Timestamp: time.Unix(int64(req.Timestamp), 0),
+				Value:     req.Value,
+			},
+		},
+		Unit:       xtime.Millisecond,
+		Annotation: nil,
+	}
+}
+
+func (h *WriteJSONHandler) parseRequest(r *http.Request) (*WriteQuery, *handler.ParseError) {
```
nikunjgit: how about supporting multiple write requests in one query ? Can be useful for benchmarks

src/query/api/v1/handler/json/write.go
benraskin92: So the way it is in m3/src/query/api/v1/handler/prometheus/native/common.go, would that mean we would need to pass in the write request info in the url parameters as opposed to the body?

src/query/api/v1/handler/json/write.go
benraskin92: So basically we'd unmarshal into a `[]WriteQuery`?

README.md
```diff
@@ -67,7 +67,7 @@ curl http://localhost:9003/writetagged -s -X POST -d '{
 
 ```
 curl http://localhost:9003/query -s -X POST -d '{
-  "namespace": "metrics",
+  "namespace": "default",
```
arnikola: Why is this changed?

src/query/api/v1/handler/json/write.go
```diff
@@ -0,0 +1,115 @@
...
+		handler.Error(w, rErr.Inner(), rErr.Code())
+		return
+	}
+
+	writeQuery := newStorageWriteQuery(req)
+
+	if err := h.store.Write(r.Context(), writeQuery); err != nil {
+		logging.WithContext(r.Context()).Error("Write error", zap.Any("err", err))
+		handler.Error(w, err, http.StatusInternalServerError)
+		return
```
arnikola: Unnecessary return

src/query/api/v1/handler/json/write.go
arnikola: I'm not sure how useful it would be for benchmarks; it's different from the prom endpoint, and if people are running standalone Query in the future, they'd be putting data either directly into m3db or through aggregators, not through this endpoint.

src/query/api/v1/handler/json/write_test.go
```diff
@@ -0,0 +1,78 @@
...
+			 "value": 10.0
+		}`
+}
+
+func TestJSONWriteParsing(t *testing.T) {
+	logging.InitWithCores(nil)
+	ctrl := gomock.NewController(t)
+	storage, _ := local.NewStorageAndSession(t, ctrl)
+
+	jsonWrite := &WriteJSONHandler{store: storage}
```
arnikola: nit: don't think you need to have a storage if you're just testing parsing

src/query/api/v1/handler/json/write.go
arnikola: I think you can just make `Timestamp` a string, and use `github.com/m3db/m3/src/query/util.ParseTimeString()` to read it to a time.Time

README.md
benraskin92: The default namespace is `default` so just makes sense to have the docs reflect that:
```
var (
	defaultLocalConfiguration = &config.LocalConfiguration{
		Namespace: "default",
		Retention: 2 * 24 * time.Hour,
	}
)
```

PR #857, opened 2018-08-24
Script to compare prom output with m3

nikunjgit:


src/cmd/services/m3query/scripts/curlDiff.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
benraskin92: also rename to `curl_diff.sh` or `curl-diff.sh`

src/cmd/services/m3query/scripts/curlDiff.sh
```diff
@@ -0,0 +1,15 @@
...
+# This script can be used to compare prometheus query output with m3query output
```
nerd0: Perhaps making this into a function so that it could be sourced into other scripts in the future. Just a thought

src/cmd/services/m3query/scripts/curlDiff.sh
```diff
@@ -0,0 +1,15 @@
...
+# Sample usage: bash curlDiff.sh "count_over_time(go_gc_duration_seconds[30s])" 1535125074
+target=${1?"Missing target, usage: $0 target start"}
+start=${2?"Missing start, usage: $0 target start"}
+duration=${3:1000}
+end=$((start + duration))
+m3command="localhost:7201/api/v1/prom/native/read?start=$start&end=$end&step=15s&debug=true --data-urlencode target=$target"
```
benraskin92: Perhaps let users specify step size? and then have the default as 15s

src/cmd/services/m3query/scripts/curlDiff.sh
```diff
@@ -0,0 +1,15 @@
...
+promcommand="localhost:9090/api/v1/query_range?start=$start&end=$end&step=15s --data-urlencode query=$target"
+echo $m3command
+echo $promcommand
+curl -G  $m3command > m3out
```
benraskin92: nit: extra space

src/cmd/services/m3query/scripts/curlDiff.sh
arnikola: Take m3 endpoint and prom endpoint as parameters

src/cmd/services/m3query/scripts/curlDiff.sh
```diff
@@ -0,0 +1,15 @@
...
+curl -G $promcommand > promout
+jq ".[]|.tags,.datapoints" m3out > m3result
+jq ".data.result|.[]|.metric,.values" promout > promresult
```
arnikola: You probably don't want to overwrite existing results you have here

src/cmd/services/m3query/scripts/curlDiff.sh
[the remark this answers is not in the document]
nikunjgit: added a check for running and renamed to prom-m3-diff.sh

src/cmd/services/m3query/scripts/curlDiff.sh
nikunjgit: don't want to make the script too complicated. We can definitely add that when needed

src/cmd/services/m3query/scripts/curlDiff.sh
nikunjgit: actually I like overwritting the results. This keeps the workflow simple where I don't have to constantly deleted created files. Also, the query should be idempotent.

nikunjgit:
The main purpose of this script is to create an easy way to quickly run a query through both prom and m3. It doesn't intend to be a way for comparisons but rather for manual inspection.

src/cmd/services/m3query/scripts/curlDiff.sh
nikunjgit: that's a good idea but I'm not sure how useful it would be to call it as a function vs calling this script completely and then working off generated files. Also, note that this is a simple script which just generates output from prom and m3. We'll be working on a more complex way to compare prom output with m3 output.

src/cmd/services/m3query/scripts/curlDiff.sh
arnikola: It's more to deal with different but similar queries, e.g. if you want to compare two results for the same query but different start/end to see where the points actually line up. Take a param to see if you should overwrite?

arnikola:
Maybe printout file size post comparison to have a quick indicator to see if they're equal? Obviously this won't be perfect, but could be a good canary

nikunjgit:
Added file size

PR #861, opened 2018-08-28
documentation uri fix

sboschman:
fixed the relative path to the m3db and query engine documentation from the components overview


richardartoul:
Awesome, thanks!

PR #862, opened 2018-08-28
Fix LRU deadlock + regression test and improve comments / logging

richardartoul:
- [X] Fix LRU deadlock that could occur when the events channel was filled up. Achieved this by making the series object call WiredList.Update() outside of a lock as well as adding the concept of a NonBlockingUpdate. Also added a regression test.
- [X] Made LRU events channel size configurable for stress testing purposes.
- [X] Improve "invariant violated" style logging so that metrics are emitted if wired list invariants are violated.
- [X] Added a new "CloseIfFromDisk" method to the database block as an additional layer of protection.

### DIFF
--- a/src/dbnode/storage/series/series.go
+++ b/src/dbnode/storage/series/series.go
@@ -34,6 +34,7 @@ import (
 	"github.com/m3db/m3x/context"
 	xerrors "github.com/m3db/m3x/errors"
 	"github.com/m3db/m3x/ident"
+	"github.com/m3db/m3x/instrument"
 	xlog "github.com/m3db/m3x/log"
 	xtime "github.com/m3db/m3x/time"
 )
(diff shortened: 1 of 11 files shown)


src/dbnode/storage/series/series.go
```diff
@@ -519,7 +520,12 @@ func (s *dbSeries) OnRetrieveBlock(
 	segment ts.Segment,
 ) {
 	s.Lock()
-	defer s.Unlock()
+	shouldUnlock := true
+	defer func() {
+		if shouldUnlock {
+			s.Unlock()
+		}
+	}()
```
robskillington: Could we refactor this to avoid multiple unlock calls?  Perhaps something like:
```go
var (
  b block.DatabaseBlock
  updateList *block.WiredList
)

s.Lock()
defer func() {
  s.Unlock()
  if b != nil && updateList != nil {
    updateList.BlockingUpdate(b) // Update outside of lock
  }
}()
```

richardartoul: oh yeah thats better thanks

PR #863, opened 2018-08-28
Fix datatypes from M3 to match prom

nikunjgit:


src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -190,3 +190,7 @@ func renderResultsJSON(w io.Writer, series []*ts.Series, params models.RequestPa
 	jw.EndArray()
 	jw.Close()
 }
+
+func valueToProm(value float64) string {
```
arnikola: nit: Add a quick test?

src/query/api/v1/handler/prometheus/native/common_test.go
```diff
@@ -74,3 +75,9 @@ func TestInvalidTarget(t *testing.T) {
 	assert.NotNil(t, p.Start)
 	require.Equal(t, err.Code(), http.StatusBadRequest)
 }
+
+func TestValueToProm(t *testing.T) {
+	assert.Equal(t, valueToProm(1.0), "1")
+	assert.Equal(t, valueToProm(1.2), "1.2")
```
arnikola: Can you add in something like `assert.Equal(t, valueToProm(0.0119311), "0.0119311")` with a lot of digits post decimal point? (took a random value from query output)

PR #864, opened 2018-08-28
Enable buffer to accept arbitrary metric times

justinjc:


src/dbnode/storage/series/buffer.go
```diff
@@ -183,78 +240,138 @@ func (b *dbBuffer) Write(
...
-	return b.buckets[idx].write(timestamp, value, unit, annotation)
+	if b.bucketsNotRealTime[key].needsDrain(now, bucketStart) {
+		b.bucketDrain(now, bucketIDNotRealtime(key), bucketStart)
+
+		if b.bucketsNotRealTime[key].isStale(now) {
+			b.removeBucket(key)
+		}
+	}
+
+	return b.bucketsNotRealTime[key].write(now, timestamp, value, unit, annotation)
```
richardartoul: Seems like this would panic if you just removed the bucket you're trying to write to

richardartoul: If I'm right about that, make sure to add a test case

src/dbnode/storage/series/buffer.go
```diff
@@ -268,65 +385,106 @@ func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
...
+	if bucket.needsDrain(now, start) {
+		mergedOutOfOrderBlocks += b.bucketDrain(now, id, start)
+	}
+
+	if bucket.needsReset(now, start) {
+		bucket.resetTo(start, true)
+	}
+
+	if bucket.isStale(now) {
+		b.removeBucket(id.key)
```
richardartoul: Is it worth adding a sanity check to remove bucket to emit an error log if its called on a bucket with more than 1 write or one that is not drained? Kind of like an invariant violated assert?

src/dbnode/storage/series/buffer.go
```diff
@@ -268,65 +385,106 @@ func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
...
-			if !(result.block.Len() > 0) {
-				log := b.opts.InstrumentOptions().Logger()
-				log.Errorf("buffer drain tried to drain empty stream for bucket: %v",
-					start.String())
-			} else {
-				// If this block was read mark it as such
-				if lastRead := b.buckets[idx].lastRead(); !lastRead.IsZero() {
-					result.block.SetLastReadTime(lastRead)
-				}
+			if b.drainFn != nil {
```
richardartoul: when would drainFn be nil?

src/dbnode/storage/series/buffer.go
```diff
@@ -268,65 +385,106 @@ func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
...
 				b.drainFn(result.block)
 			}
 		}
-
-		b.buckets[idx].drained = true
 	}
```
richardartoul: It looks like this code used to check (for realtime buckets) if the bucket need to be reset. Is that gone now? Where does the reset happen?

src/dbnode/storage/series/buffer.go
```diff
@@ -268,65 +385,106 @@ func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
...
 
-	if b.buckets[idx].needsReset(start) {
-		// Reset bucket
-		b.buckets[idx].resetTo(start)
+	bucket.drained = true
+	bucket.setLastWrite(now)
```
richardartoul: why would you update last write here?

src/dbnode/storage/series/buffer.go
```diff
@@ -268,65 +385,106 @@ func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
...
 					blockStart.String(),
 				)
 			}
-			b.buckets[i].bootstrap(bl)
+			b.bucketsRealTime[i].bootstrap(bl)
 			bootstrapped = true
 			break
 		}
 	}
+	if bucket, ok := b.bucketsNotRealTime[xtime.ToUnixNano(blockStart)]; !bootstrapped && ok {
```
richardartoul: You're gonna do the map lookup here even if `bootstrapped = true` you can probably refactor this to prevent that

src/dbnode/storage/series/buffer.go
```diff
@@ -268,65 +385,106 @@ func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
...
-		b.buckets[idx].drained = true
 	}
 
-	if b.buckets[idx].needsReset(start) {
-		// Reset bucket
-		b.buckets[idx].resetTo(start)
+	bucket.drained = true
+	bucket.setLastWrite(now)
+	bucket.resetNumWrites()
+	if bucket.isStale(now) {
```
richardartoul: confused about the stale chck and removal again. Once its drained, why not just remove it? Also since you just called `setLastWrite(now)` wont it always be false?

src/dbnode/storage/series/buffer.go
richardartoul: Actually this whole code path is a little unclear to me. When would we be bootstrapping non-realtime blocks? I think this portion will need some careful thought with how we plan to handle bootstrapping and non-realtime data...

richardartoul: Probably worth a discussion in person

src/dbnode/storage/series/buffer.go
```diff
@@ -49,13 +50,21 @@ const (
 	// 3. Bucket for the future that can be taking writes that is head of
 	// the current block if write is for the future within bounds
 	bucketsLen = 3
+
+	// TODO: make sure this is a good pool size or make it customizable
+	defaultBufferBucketPoolSize = 16
```
richardartoul: questioning whether pooling this at all is worth it, especially with a size of 16. I don't imagine we'll be creating these buckets so often that pooling will matter, and if we were, would probably want a global shared pool or something. Maybe Rob has some thoughts on this

src/dbnode/storage/series/buffer.go
```diff
@@ -369,7 +548,7 @@ func (b *dbBuffer) Snapshot(ctx context.Context, blockStart time.Time) (xio.Segm
 		err error
 	)
 
-	b.forEachBucketAsc(func(bucket *dbBufferBucket) {
+	b.forEachBucketAsc(bucketTypeAll, func(bucket *dbBufferBucket) {
```
richardartoul: Do we want bucketTypeAll here? Seems like we'll need some kind of separate code path for snapshotting non-realtime blocks, otherwise the data will end up in the wrong fileset files

richardartoul: I guess technically your code probably won't allow a realtime and non-realtime block for the same blockStart to exist at the same time so maybe it doesn't matter, but seems better to only go over realtime blocks here

src/dbnode/storage/series/buffer.go
justinjc: I shouldn't, good catch.

src/dbnode/storage/series/buffer.go
```diff
@@ -551,20 +737,44 @@ func (b *dbBufferBucket) canRead() bool {
...
+
 	return !b.start.Equal(targetStart)
 }
 
+func (b *dbBufferBucket) isStale(now time.Time) bool {
+	if b.isRealTime {
+		return false
+	}
+
+	return b.numWrites() > 0 && now.Sub(b.lastWrite()) > b.opts.RetentionOptions().FlushAfterNoMetricPeriod()
```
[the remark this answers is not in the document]
justinjc: What I call "stale" here is to prevent buckets from lying around for too long. Buckets that are stale haven't been written to for a while, so should be reclaimed, but if a bucket has been written to recently, it's likely that it will be written to again. It would be a waste to reclaim the bucket just to reinitialize one again on the next write.

src/dbnode/storage/series/buffer.go
justinjc: This is a good catch and was a bug. It makes no sense to remove a bucket due to staleness right before a write, since that will cause the bucket to be removed and then reinitialized right after. The only place where it makes sense to remove a stale bucket is during a bucketTick - I've cleaned this code up.

src/dbnode/storage/series/buffer.go
justinjc: When constructing a new buffer (`newDatabaseBuffer(...)`), you can pass in a nil drainFn (in my case, when writing test code). Trying to call `drainFn(...)` would then throw.

src/dbnode/storage/series/buffer.go
justinjc: Removed.

src/dbnode/storage/series/buffer.go
justinjc: Agreed - this logic was a little convoluted. Cleaned this up to decouple this and only have bucketTick deal with stale bucket removal.

src/dbnode/storage/series/buffer.go
```diff
@@ -183,78 +240,138 @@ func (b *dbBuffer) Write(
...
-		b.DrainAndReset()
+	if _, ok := b.bucketsNotRealTime[key]; !ok {
+		b.initializeBucket(key)
 	}
 
-	return b.buckets[idx].write(timestamp, value, unit, annotation)
+	if b.bucketsNotRealTime[key].needsDrain(now, bucketStart) {
+		b.bucketDrain(now, bucketIDNotRealtime(key), bucketStart)
+
+		if b.bucketsNotRealTime[key].isStale(now) {
```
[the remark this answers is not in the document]
justinjc: I was thinking to remove buckets from the map only when they haven't been written to for a while to prevent buckets from having to constantly get destroyed only to be recreated right away again.

At the same time, if there is a constant and sustained write to a bucket, I don't want there to be a large amount of metrics that never get drained. Therefore we want to drain if the bucket `.isFull()`, but we don't want to remove it yet because there's probably another metric going into this bucket soon.

src/dbnode/storage/series/buffer.go
justinjc: The actual draining logic got refactored out to `bucketDrain`, but the reset code still lives in the `bucketDrainAndReset` function.

src/dbnode/storage/series/buffer.go
```diff
@@ -183,78 +240,138 @@ func (b *dbBuffer) Write(
...
+	if b.isRealTime(timestamp) {
+		idx := b.writableBucketIdx(timestamp)
+
+		if b.bucketsRealTime[idx].needsReset(now, bucketStart) {
+			b.DrainAndReset()
+		}
+		// If there was previously a non-realtime metric written to this bucket
+		// already, then use that bucket.
+		if _, ok := b.bucketsNotRealTime[key]; ok {
+			b.makeBucketRealTime(key, idx)
```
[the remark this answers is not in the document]
justinjc: I like this idea. That feels like a cleaner solution - the non-realtime bucket will eventually become stale and get drained on its own.

src/dbnode/storage/series/buffer.go
```diff
@@ -268,65 +385,106 @@ func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
...
+		b.removeBucket(id.key)
 	}
 
 	return mergedOutOfOrderBlocks
 }
 
+func (b *dbBuffer) makeBucketRealTime(fromKey xtime.UnixNano, toIdx int) {
+	bucket, ok := b.bucketsNotRealTime[fromKey]
+	if !ok {
+		return
```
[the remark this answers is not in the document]
justinjc: Removed this whole "make a bucket real time" idea.

src/dbnode/storage/series/buffer.go
```diff
@@ -183,78 +240,138 @@ func (b *dbBuffer) Write(
...
+	bucketStart := timestamp.Truncate(b.blockSize)
+	key := b.writableBucketKey(timestamp)
+
+	if b.isRealTime(timestamp) {
+		idx := b.writableBucketIdx(timestamp)
+
+		if b.bucketsRealTime[idx].needsReset(now, bucketStart) {
+			b.DrainAndReset()
+		}
+		// If there was previously a non-realtime metric written to this bucket
```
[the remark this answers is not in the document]
justinjc: Removed this whole "make a bucket real time" idea.

src/dbnode/storage/series/buffer.go
robskillington: Yeah if it's only for buffers outside of the current realtime mutable buckets, we don't need a pool.

For the realtime ones that we rotate between constantly, if we keep that rotation buffer in here (i.e. bucketsRealtime) then we don't need to pool them since we will be reusing them anyway.

PR #868, opened 2018-08-29
Arnikola/parameterized func

arnikola:
Adds parametrized aggregation functions and associated tests:
topK
bottomK
nth-quantile

TODO:
add count_values


src/query/functions/aggregation/base.go
```diff
@@ -29,7 +29,7 @@ import (
 	"github.com/m3db/m3/src/query/parser"
 )
 
-type aggregationFn func(values []float64, indices []int) float64
+type aggregationFn func(values []float64, buckets []int) float64
```
nikunjgit: buckets is misleading ? since its  a single bucket

src/query/functions/aggregation/take.go
```diff
@@ -0,0 +1,176 @@
...
+		for _, idx := range bucket {
+			val := values[idx]
+			if !math.IsNaN(val) {
+				heap.Push(values[idx], idx)
+			}
+
+			values[idx] = math.NaN()
+		}
+
+		// Re-add the val/index pairs from the heap to the input vector
```
nikunjgit: how do you match series metadata ? Seems like for every column, you can have a different ordering of values ?

src/query/functions/utils/heap.go
```diff
@@ -0,0 +1,200 @@
...
+// Flush flushes the float heap and resets it. Does not guarantee order.
+func (fh FloatHeap) Flush() []ValueIndexPair {
+	values := fh.floatHeap.heap
+	fh.Reset()
+	return values
+}
+
+// FlushByPop flushes the float heap and resets it by popping values from the heap.
+// Guarantees order.
+func (fh FloatHeap) FlushByPop() []ValueIndexPair {
```
nikunjgit: is this only for used in tests ? then don't export it ?

src/query/functions/aggregation/take.go
```diff
@@ -0,0 +1,176 @@
...
+		if len(bucket) <= cap {
+			continue
+		}
+
+		// Add values from this bucket to heap, clearing them from input vector
+		// after they are in the heap.
+		for _, idx := range bucket {
+			val := values[idx]
+			if !math.IsNaN(val) {
+				heap.Push(values[idx], idx)
```
nikunjgit: I think your logic gets simpler if instead of pushing values to the heap one by one, you heapify with all values at once and extract k values in the end.

src/query/functions/utils/heap.go
```diff
@@ -0,0 +1,200 @@
...
+	heap.Init(floatHeap)
+	return FloatHeap{
+		isMaxHeap: isMaxHeap,
+		capacity:  capacity,
+		floatHeap: floatHeap,
+	}
+}
+
+// Push pushes a value and index pair to the heap
+func (fh FloatHeap) Push(value float64, index int) {
```
nikunjgit: don't think this is the right way. Why not push the element to the heap and keep removing until we are <= capacity ? I think you're trying to get into the heap internals a lot which we should avoid here.

nikunjgit: also my other earlier comment about just heapifying the whole thing instead of individual pushes, might change this a bit.

arnikola: This is much better in the scenario we have 10,000 series and we're taking top 20, much better to have a max heap size to cap it to 20 elements, saving both memory and time needed to sort the heap on each insert/pop

src/query/functions/aggregation/quantile.go
```diff
@@ -0,0 +1,83 @@
...
+	l := float64(len(bucketVals))
+	if l == 0 {
+		// No non-NaN values
+		return math.NaN()
+	}
+
+	sort.Float64s(bucketVals)
+	// When the quantile lies between two samples,
+	// use a weighted average of the two samples.
+	rank := n * (l - 1)
```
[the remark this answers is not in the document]
arnikola: This is a really simple function; I could see a benefit from using a package for something more complex, but here doesn't really make a difference

src/query/functions/aggregation/take.go
arnikola: The value order does not change. The index of each value added to the heap is stored as well, and these are added to the output value list at the same index that they were taken from.

TakeK functions keep the metadata of the incoming series, so these will always be in order.

src/query/functions/aggregation/take.go
arnikola: It's quite difficult to know which values to add to the heap. Each bucket contains the indices that should be used for that takeK function, representing the indices that correspond to the requested grouping by tags. 

I'd have to gather each of the values and its corresponding index that matches the bucket here, then push that vector into the heap; which would still act pretty much the same, so it wouldn't make the logic any simpler

src/query/functions/utils/heap.go
arnikola: Might just delete this; was initially a bit reticent about pulling the underlying array as is done in `Flush`, but considering that the value index is included in the output data, it's unnecessary to have the output be sorted.

src/query/functions/aggregation/take.go
nikunjgit: I see. So for 100 series, if I take top 10, then for each time interval, the series for which datapoints are in top 10 will be set and other will be NaNs ?

src/query/functions/aggregation/take.go
nikunjgit: You're right, iterating wouldn't change. Iterating over values is simple, I meant you wouldn't need to have a complex Push operation

src/query/functions/utils/heap.go
nikunjgit: You're right about this being optimal but this requires us to be more careful about correctness. I don't think a slice of 10K vs a slice of 20 really matters that much since we can reuse the same slice anyway. My suggestion was to keep the code simpler by not making this optimization until its really needed. 
If you will still like to go forward with your approach, that's fine but we'll then have to take a closer look at your logic...

arnikola: That's the point of the review isn't it 😛

src/query/functions/aggregation/quantile.go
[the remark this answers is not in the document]
nikunjgit: yeah but we know we have to implement things for stats. Possibly variance, quantile, percentile. If we go with the library approach and only add functions when they are not present then we can reduce the code we have to maintain in my opinion. Otherwise, we would always end up in the situation where the incremental code we add is simple.

src/query/functions/aggregation/base.go
arnikola: Haha changed it everywhere except here :(

src/query/functions/aggregation/quantile.go
```diff
@@ -0,0 +1,83 @@
...
+func makeQuantileFn(opType string, q float64) (aggregationFn, bool) {
+	if opType != QuantileType {
+		return nil, false
+	}
+	return func(values []float64, bucket []int) float64 {
+		return quantileFn(q, values, bucket)
+	}, true
+}
+
+func quantileFn(q float64, values []float64, bucket []int) float64 {
```
nikunjgit: can we split this out since part where we get bucketVals and then later part which creates actual quantile ? That way we can move the second thing out and reuse it at other place (or use a library later)

src/query/functions/aggregation/function.go
```diff
@@ -33,38 +33,43 @@ const (
...
-	for _, idx := range buckets {
+	for _, idx := range bucket {
 		v := values[idx]
 		if !math.IsNaN(v) {
 			sum += v
 			count++
 		}
 	}
 
+	// If all elements are NaN, sum should be NaN
```
benraskin92: 👍

src/query/functions/aggregation/function_test.go
```diff
@@ -34,15 +34,15 @@ import (
 )
 
 func TestCountWithAllValues(t *testing.T) {
-	values, bounds := test.GenerateValuesAndBounds(nil, nil)
-	block := test.NewBlockFromValues(bounds, values)
+	vals, b := test.GenerateValuesAndBounds(nil, nil)
```
benraskin92: Pass in values instead of `nil`?

src/query/functions/utils/heap.go
```diff
@@ -0,0 +1,188 @@
...
+			// Compare incoming value with current top of heap.
+			// Decide if to replace the current top, or to drop incoming
+			// value as appropriate for this min/max heap
+			//
+			// If values are equal, do not add incoming value regardless
+			//
+			// NB(arnikola): unfortunately, can't just replace first
+			// element as it may not respect internal order. Need to
+			// run heap.Fix() to rectify this
+			if fh.isMaxHeap && value > peek.Val ||
```
nikunjgit: I don't think we should replicate the heap logic here. A better way would be to have a simple push and pop operation. After doing the push, check the length of the heap. If you're at max, then pop.

src/query/functions/aggregation/function_test.go
```diff
@@ -185,7 +193,7 @@ func TestAggFns(t *testing.T) {
 				for i, bucket := range tt.buckets {
 					actual := function.fn(tt.values, bucket)
 					expected := function.expected[i]
-					test.EqualsWithNansWithDelta(t, expected, actual, 0.00001)
+					test.EqualsWithNansWithDelta(t, expected, actual, math.Pow10(-5))
```
benraskin92: Can we keep `0.00001`? `math.Pow10(-5)` is harder to grok.

src/query/functions/aggregation/take.go
nikunjgit: A better way would be to have a simple push and pop operation. After doing the push, check the length of the heap. If you're at max, then pop.

src/query/functions/aggregation/quantile.go
```diff
@@ -0,0 +1,83 @@
...
+
+	bucketVals := make([]float64, 0, len(bucket))
+	for _, idx := range bucket {
+		val := values[idx]
+		if !math.IsNaN(val) {
+			bucketVals = append(bucketVals, values[idx])
+		}
+	}
+
+	l := float64(len(bucketVals))
```
benraskin92: Maybe do this below when you actually need a `float64`?

src/query/functions/aggregation/function_test.go
arnikola: Deleted these two tests since they're covered in the tests below

src/query/functions/aggregation/take.go
```diff
@@ -0,0 +1,176 @@
...
+			}
+
+			values[idx] = math.NaN()
```
nikunjgit: why ? isn't it already NaN ?

src/query/functions/binary/arithmetic.go
```diff
@@ -82,6 +82,6 @@ func buildArithmeticFunction(
 
 	// Build the binary processing step
 	return func(lhs, rhs block.Block, controller *transform.Controller) (block.Block, error) {
-		return processBinary(lhs, rhs, params, controller, fn)
+		return processBinary(lhs, rhs, params, controller, false, fn)
```
nikunjgit: 6 parameter functions :(

src/query/functions/binary/binary.go
```diff
@@ -72,7 +71,7 @@ func processBinary(
 
 		// NB(arnikola): this is a sanity check, as scalar comparisons
 		// should have previously errored out during the parsing step
-		if !params.ReturnBool {
+		if !params.ReturnBool && isComparison {
```
nikunjgit: shouldn't we be leaving this to the parser ? Seems like duplication of logic.

arnikola: Agreed, that's why it's specifically a sanity check that we don't get into weird situations here

src/query/functions/utils/heap.go
arnikola: This is primarily to avoid putting in values we're going to instantly pop off anyway in the `push + pop until max` approach

PR #872, opened 2018-08-30
Add temporal functions

benraskin92:


src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,162 @@
...
+				count++
+				delta := v - mean
+				mean += delta / count
+				aux += delta * (v - mean)
+			}
+		}
+		return aux / count
+
+	default:
+		panic("unknown aggregation type")
```
benraskin92: This doesn't seem right, but, on the other hand, if we get to this point, something is seriously wrong.

arnikola: Rather than doing this, you can have
`type aggFunc func(vals float64) float64` 
Then have these all be seperate functions, and have aggFuncs be a map of these, similar to how aggregation functions are done in aggregation/base.go

src/query/parser/promql/parse_test.go
```diff
@@ -231,19 +231,41 @@ func TestParenPrecedenceParses(t *testing.T) {
...
-	require.NoError(t, err)
-	assert.Len(t, transforms, 2)
-	assert.Equal(t, transforms[0].Op.OpType(), functions.FetchType)
-	assert.Equal(t, transforms[0].ID, parser.NodeID("0"))
-	assert.Equal(t, transforms[1].ID, parser.NodeID("1"))
-	assert.Equal(t, transforms[1].Op.OpType(), temporal.CountTemporalType)
-	assert.Len(t, edges, 1)
-	assert.Equal(t, edges[0].ParentID, parser.NodeID("0"), "fetch should be the parent")
-	assert.Equal(t, edges[0].ChildID, parser.NodeID("1"), "aggregation should be the child")
+var temporalParseTests = []struct {
```
arnikola: Nice!

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,161 @@
...
+		return count
+
+	case MinTemporalType:
+		min := math.Inf(1)
+		for _, v := range values {
+			if !math.IsNaN(v) {
+				min = math.Min(min, v)
+			}
+		}
+		return min
```
arnikola: This fails edge cases; if all values are NaNs, this will return math.Inf(1) instead of math.NaN()

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,161 @@
...
+
+	case MaxTemporalType:
+		max := math.Inf(-1)
+		for _, v := range values {
+			if !math.IsNaN(v) {
+				max = math.Max(max, v)
+			}
+		}
+		return max
```
arnikola: This fails edge cases; if all values are NaNs, this will return math.Inf(-1) instead of math.NaN()

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func (a *aggNode) Process(values []float64) float64 {
+	switch a.op.operatorType {
+	case AvgTemporalType:
+		var mean, count float64
+		for _, v := range values {
+			if !math.IsNaN(v) {
+				count++
+				mean += (v - mean) / count
```
arnikola: Why do you need a rolling average here? Can't you just keep track of sum and count then return sum/count? You'll also need to account for cases where count == 0

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,161 @@
...
+	case SumTemporalType:
+		var sum float64
+		for _, v := range values {
+			if !math.IsNaN(v) {
+				sum += v
+			}
+		}
+		return sum
+
+	case StdDevTemporalType:
```
arnikola: This is calculating population standard deviation rather than standard deviation, is that what prom uses?

Also will need to have special cases for count == 0 so you don't get div by zeroes if all values are NaN

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,161 @@
...
+		var aux, count, mean float64
+		for _, v := range values {
+			if !math.IsNaN(v) {
+				count++
+				delta := v - mean
+				mean += delta / count
+				aux += delta * (v - mean)
+			}
+		}
+		return aux / count
```
arnikola: Same here re: population standard variance

src/query/functions/temporal/aggregation_test.go
```diff
@@ -0,0 +1,207 @@
...
+			{2, 2, 2, 2, 2},
+			{2, 2, 2, 2, 2},
+		},
+	},
+}
+
+// B1 has NaN in first series, first position
+func TestAggregation(t *testing.T) {
+	for _, testCase := range testCases {
+		values, bounds := test.GenerateValuesAndBounds(nil, nil)
```
arnikola: Unfortunately if you're using this as your `values`, you only get a really nice set of data points which do not help catch weird edge case issues

src/query/functions/temporal/aggregation_test.go
```diff
@@ -0,0 +1,207 @@
...
+		afterAllBlocks: [][]float64{
+			{2, 2, 2, 2, 2},
+			{2, 2, 2, 2, 2},
+		},
+	},
+}
+
+// B1 has NaN in first series, first position
+func TestAggregation(t *testing.T) {
+	for _, testCase := range testCases {
```
arnikola: The typical way of running these tabled tests is:
```
	for _, tt := range testCases {
 		t.Run(tt.name, func(t *testing.T) {
...
```

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,161 @@
...
+		MinTemporalType:    struct{}{},
+		MaxTemporalType:    struct{}{},
+		SumTemporalType:    struct{}{},
+		StdDevTemporalType: struct{}{},
+		StdVarTemporalType: struct{}{},
+	}
+)
+
+// NewAggOp creates a new base temporal transform with a specified node
+func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
```
arnikola: I'd recommend just taking floats or whatever as your args, and have the logic for providing usable values in the parser instead, so you don't end up with extra parsing scattered around in the function op.

src/query/functions/temporal/aggregation.go
benraskin92: Yeah, I thought it'd be a little messier, but on second thought, this is probably better. Will make the change.

src/query/functions/temporal/aggregation_test.go
benraskin92: Will add other test cases

src/query/functions/temporal/aggregation.go
arnikola: 👍 
Another nice thing you get is that you don't need to check opType twice

src/query/functions/temporal/aggregation.go
benraskin92: Was following how Prom did it, but yeah we can just do that instead.

src/query/functions/temporal/aggregation.go
benraskin92: Yeah that's what Prom uses.

src/query/functions/temporal/aggregation_test.go
benraskin92: Yeah- I can do that in another pr

src/query/functions/temporal/aggregation.go
```diff
@@ -84,78 +84,101 @@ type aggNode struct {
...
-		return count
-
-	case MinTemporalType:
-		min := math.Inf(1)
-		for _, v := range values {
-			if !math.IsNaN(v) {
-				min = math.Min(min, v)
-			}
+	}
+	if count == 0 {
```
arnikola: nit: newlines

Also consider doing something like aggregation functions do where you have a `sumAndCount` function which returns `(sum, count float64)`, making a bunch of these easier

src/query/functions/temporal/aggregation.go
```diff
@@ -84,78 +84,101 @@ type aggNode struct {
...
-		return max
-
-	case SumTemporalType:
-		var sum float64
-		for _, v := range values {
-			if !math.IsNaN(v) {
-				sum += v
-			}
+	}
+	if min == math.Inf(1) {
```
arnikola: newlines

Also this might be a bit dumb, but if math.Inf somehow appears in the series, you'd get incorrect results

src/query/functions/temporal/aggregation_test.go
```diff
@@ -131,73 +131,183 @@ var testCases = []testCase{
...
+
+func testAggregation(t *testing.T, testCases []testCase, vals [][]float64) {
+	for _, tt := range testCases {
+		t.Run(tt.name, func(t *testing.T) {
+			var (
+				values [][]float64
+				bounds block.Bounds
+			)
+
+			if vals == nil {
```
arnikola: Can you just pass in values instead of doing this generation thing? It's really bad for comprehension and understanding why the results are what they are

src/query/functions/temporal/aggregation.go
nikunjgit: i think the problem here is that some ops require more than one args. Then the question is where should that logic live which determines which argument is need by which function.

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,194 @@
...
+func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
+	if _, ok := aggFuncs[optype]; !ok {
+		return emptyOp, fmt.Errorf("unknown aggregation type: %s", optype)
+	}
+
+	return newBaseOp(args, optype, newAggNode)
+}
+
+func newAggNode(op baseOp, controller *transform.Controller) Processor {
+	return &aggNode{
```
nikunjgit: maybe just store the function directly so that we don't have to do the lookup in Process ?

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,194 @@
...
+type aggNode struct {
+	op         baseOp
+	controller *transform.Controller
+}
+
+func (a *aggNode) Process(values []float64) float64 {
+	return aggFuncs[a.op.operatorType](values)
+}
+
+func avgOverTime(values []float64) float64 {
```
nikunjgit: all these functions look very similar to what Artem wrote. Consider consolidating them ?

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,196 @@
...
+		}
+	}
+
+	if count == 0 {
+		return math.NaN()
+	}
+	return sum
+}
+
+func stddevOverTime(values []float64) float64 {
```
nikunjgit: instead of implementing all these functions, can you use : gonum ? Eg: https://github.com/gonum/gonum/blob/73ea1e732937f96d723d31dc5263d214a275d204/stat/stat.go#L1199

src/query/functions/temporal/aggregation.go
arnikola: We should add a todo to consolidate how we pass args around, https://github.com/m3db/m3/pull/884 should help us standardize scalar parameters, but unfortunately there's a few special cases in prom which take `Scalar, Fetch` instead of `Fetch, Scalar`, or even `Fetch, Scalar, Scalar`. Prom have a set of required param types by position; may be an idea to do something similar?

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,196 @@
...
+		SumTemporalType:    sumOverTime,
+		StdDevTemporalType: stddevOverTime,
+		StdVarTemporalType: stdvarOverTime,
+	}
+)
+
+// NewAggOp creates a new base temporal transform with a specified node
+func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
+	if _, ok := aggFuncs[optype]; !ok {
+		return emptyOp, fmt.Errorf("unknown aggregation type: %s", optype)
```
arnikola: nit: `nil` should work instead of `emptyOp`

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	return newBaseOp(args, optype, newAggNode)
+}
+
+func newAggNode(op baseOp, controller *transform.Controller) Processor {
+	return &aggNode{
+		op:         op,
+		controller: controller,
+		aggFunc:    aggFuncs[op.operatorType],
```
arnikola: You've already retrieved the aggFunction previously, when you were checking for map membership in `NewAggOp`. Can you just pass that through instead of re-fetching it here?

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+}
+
+type aggNode struct {
+	op         baseOp
+	controller *transform.Controller
+	aggFunc    func([]float64) float64
```
arnikola: nit: define this as a type, i.e. `type aggFunc func([]float64) float64`

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,196 @@
...
+
+	// StdDevTemporalType calculates the standard deviation of all values in the specified interval
+	StdDevTemporalType = "stddev_over_time"
+
+	// StdVarTemporalType calculates the standard variance of all values in the specified interval
+	StdVarTemporalType = "stdvar_over_time"
+)
+
+var (
+	aggFuncs = map[string]func([]float64) float64{
```
arnikola: nit: define and use an `aggFunc` type here

src/query/functions/temporal/aggregation.go
arnikola: If not using gonum, can this just return `math.Sqrt(stdvarOverTime(values))`

src/query/functions/temporal/aggregation_test.go
```diff
@@ -0,0 +1,207 @@
...
+	},
+	{
+		description: "testing stddev_over_time",
+		opType:      StdDevTemporalType,
+		afterBlockOne: [][]float64{
+			{math.NaN(), math.NaN(), math.NaN(), math.NaN(), 1.1180},
+			{math.NaN(), math.NaN(), math.NaN(), math.NaN(), 1.4142},
+		},
+		afterAllBlocks: [][]float64{
+			{1.4142, 1.4142, 1.4142, 1.4142, 1.4142},
```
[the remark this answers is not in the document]
arnikola: pls respond

src/query/parser/promql/types.go
```diff
@@ -141,8 +141,10 @@ func NewFunctionExpr(name string, argValues []interface{}) (parser.Params, error
 		linear.MinuteType, linear.MonthType, linear.YearType:
 		return linear.NewDateOp(name)
 
-	case temporal.CountTemporalType:
-		return temporal.NewCountOp(argValues)
+	case temporal.AvgTemporalType, temporal.CountTemporalType, temporal.MinTemporalType,
```
arnikola: take it or leave it: All of these are getting pretty long, may be an idea to make it into a function that checks if it's one of these instead? Your call

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,196 @@
...
+}
+
+func countOverTime(values []float64) float64 {
+	var count float64
+	for _, v := range values {
+		if !math.IsNaN(v) {
+			count++
+		}
+	}
+	return count
```
arnikola: nit: newline missing in most of these pre-return

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,196 @@
...
+	op         baseOp
+	controller *transform.Controller
+	aggFunc    func([]float64) float64
+}
+
+func (a *aggNode) Process(values []float64) float64 {
+	return a.aggFunc(values)
+}
+
+func avgOverTime(values []float64) float64 {
```
arnikola: I liked @nikunjgit's idea in the regular aggregation functions to make a function that returns (sum, count) which simplifies sum, count, average, and can help with stddev/stdvar

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,196 @@
...
+	for _, v := range values {
+		if !math.IsNaN(v) {
+			count++
+		}
+	}
+	return count
+}
+
+func minOverTime(values []float64) float64 {
+	var count float64
```
arnikola: This can just be a bool, `seenNotNaN` or something?

src/query/functions/temporal/aggregation.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	if count == 0 {
+		return math.NaN()
+	}
+	return min
+}
+
+func maxOverTime(values []float64) float64 {
+	var count float64
```
arnikola: This can just be a bool, `seenNotNaN` or something?

src/query/functions/temporal/aggregation.go
```diff
@@ -90,78 +97,67 @@ func (a *aggNode) Process(values []float64) float64 {
...
-	for _, v := range values {
-		if !math.IsNaN(v) {
-			count++
-			sum += v
-		}
-	}
-
+	sum, count := sumAndCount(values)
 	if count == 0 {
 		return math.NaN()
```
arnikola: nit: this logic might be better to put in sumAndCount?

src/query/functions/temporal/aggregation.go
```diff
@@ -64,18 +66,23 @@ var (
 
 // NewAggOp creates a new base temporal transform with a specified node
 func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
-	if _, ok := aggFuncs[optype]; !ok {
-		return emptyOp, fmt.Errorf("unknown aggregation type: %s", optype)
+	var (
```
arnikola: nit: would read cleaner as
```golang	
if aggregationFunc, ok := aggFuncs[optype];  ok {
 return newBaseOp(args, optype, newAggNode, aggregationFunc)
} 

return nil, fmt.Errorf("unknown aggregation type: %s", optype)
```

src/query/functions/temporal/aggregation_test.go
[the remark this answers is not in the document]
benraskin92: Just one NaN, but it gets converted back to the original value (0) on line 270.

`values[0][0] = original`

src/query/functions/temporal/aggregation.go
benraskin92: But then you'd still have to have an if statement here, no?

src/query/functions/temporal/aggregation.go
```diff
@@ -93,15 +93,14 @@ func (a *aggNode) Process(values []float64) float64 {
...
-
 	return sum / count
 }
 
 func countOverTime(values []float64) float64 {
 	_, count := sumAndCount(values)
+	if count == 0 {
+		return math.NaN()
+	}
 	return count
```
arnikola: nit: newline

PR #874, opened 2018-08-30
Arnikola/count values

arnikola:
Adds the count_value function. This function is a bit peculiar in that it's the only one that I'm aware of that can actually *increase* the number of series in a block.

Unfortunately this translates rather poorly to the current model for building blocks, as series can be added at a later step while iterating over the block.

### DIFF
--- a/src/query/functions/aggregation/count_values.go
+++ b/src/query/functions/aggregation/count_values.go
@@ -25,6 +25,8 @@ import (
 	"math"
 	"strconv"
 
+	"github.com/m3db/m3/src/query/ts"
+
 	"github.com/m3db/m3/src/query/block"
 	"github.com/m3db/m3/src/query/executor/transform"
 	"github.com/m3db/m3/src/query/functions/utils"
@@ -37,7 +39,7 @@ const (
 	CountValuesType = "count_values"
 )
 
-// NewCountValuesOp creates a new take operation
+// NewCountValuesOp creates a new count values operation
 func NewCountValuesOp(
 	opType string,
 	params NodeParams,
@@ -49,7 +51,7 @@ func NewCountValuesOp(
 	return newCountValuesOp(params, opType), nil
 }
 
-// countValuesOp stores required properties for take ops
+// countValuesOp stores required properties for count values ops
 type countValuesOp struct {
 	params NodeParams
 	opType string
@@ -88,9 +90,10 @@ type countValuesNode struct {
 	controller *transform.Controller
 }
 
-type bucketColumn []float64
+type bucketColumn []int
 
 type bucketBlock struct {
+	columnLength int
 	columns      []bucketColumn
 	indexMapping map[float64]int
 }
@@ -113,10 +116,10 @@ func (n *countValuesNode) Process(ID parser.NodeID, b block.Block) error {
 	)
 
 	stepCount := stepIter.StepCount()
-	tempBlock := make([]bucketBlock, len(buckets))
-	for i := range tempBlock {
-		tempBlock[i].columns = make([]bucketColumn, stepCount)
-		tempBlock[i].indexMapping = make(map[float64]int, 10)
+	intermediateBlock := make([]bucketBlock, len(buckets))
+	for i := range intermediateBlock {
+		intermediateBlock[i].columns = make([]bucketColumn, stepCount)
+		intermediateBlock[i].indexMapping = make(map[float64]int, 10)
 	}
 
 	for columnIndex := 0; stepIter.Next(); columnIndex++ {
@@ -127,14 +130,12 @@ func (n *countValuesNode) Process(ID parser.NodeID, b block.Block) error {
 
 		values := step.Values()
 		for bucketIndex, bucket := range buckets {
-			currentBucketBlock := tempBlock[bucketIndex]
-			// If not on the first column, generate appropriate number of rows full of NaNs
-			if columnIndex > 0 {
-				previousLen := len(currentBucketBlock.columns[columnIndex-1])
-				currentBucketBlock.columns[columnIndex] = make(bucketColumn, previousLen)
-				for i := 0; i < previousLen; i++ {
-					currentBucketBlock.columns[columnIndex][i] = math.NaN()
-				}
+			currentBucketBlock := intermediateBlock[bucketIndex]
+			// Generate appropriate number of rows full of -1s that will later map to NaNs
+			currentColumnLength := currentBucketBlock.columnLength
+			currentBucketBlock.columns[columnIndex] = make(bucketColumn, currentColumnLength)
+			for i := 0; i < currentColumnLength; i++ {
+				ts.MemsetInt(currentBucketBlock.columns[columnIndex], -1)
 			}
 
 			countedValues := countValuesFn(values, bucket)
@@ -143,7 +144,7 @@ func (n *countValuesNode) Process(ID parser.NodeID, b block.Block) error {
 				if rowIndex, seen := currentBucketBlock.indexMapping[distinctValue]; seen {
 					// This value has already been seen at rowIndex in a previous column
 					// so add the current value to the appropriate row index.
-					currentBucketBlock.columns[columnIndex][rowIndex] = count
+					currentBucketColumn[rowIndex] = count
 				} else {
 					// The column index needs to be created here already
 					// Add the count to the end of the bucket column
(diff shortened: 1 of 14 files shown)


src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,229 @@
...
+	}
+
+	for columnIndex := 0; stepIter.Next(); columnIndex++ {
+		step, err := stepIter.Current()
+		if err != nil {
+			return err
+		}
+
+		values := step.Values()
+		for bucketIndex, bucket := range buckets {
```
nikunjgit: maybe consider moving out processing of a single bucket into a separate function to make it easier to understand

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,229 @@
...
+				}
+			}
+
+			countedValues := countValuesFn(values, bucket)
+			for distinctValue, count := range countedValues {
+				currentBucketColumn := currentBucketBlock.columns[columnIndex]
+				if rowIndex, seen := currentBucketBlock.indexMapping[distinctValue]; seen {
+					// This value has already been seen at rowIndex in a previous column
+					// so add the current value to the appropriate row index.
+					currentBucketBlock.columns[columnIndex][rowIndex] = count
```
nikunjgit: currentBucketColumn[rowIndex] = count ?

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,229 @@
...
+	numToPad := size - len(vals)
+	for i := 0; i < numToPad; i++ {
+		vals = append(vals, math.NaN())
+	}
+	return vals
+}
+
+// count values takes a value array and a bucket list, returns a map of
+// distinct values to number of times the value was seen in this bucket
+func countValuesFn(values []float64, buckets []int) map[float64]float64 {
```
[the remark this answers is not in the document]
arnikola: The second float becomes the datapoint; if this is an int here, it'll just have to be converted later

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,229 @@
...
+		}
+
+		values := step.Values()
+		for bucketIndex, bucket := range buckets {
+			currentBucketBlock := tempBlock[bucketIndex]
+			// If not on the first column, generate appropriate number of rows full of NaNs
+			if columnIndex > 0 {
+				previousLen := len(currentBucketBlock.columns[columnIndex-1])
+				currentBucketBlock.columns[columnIndex] = make(bucketColumn, previousLen)
+				for i := 0; i < previousLen; i++ {
```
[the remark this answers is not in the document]
arnikola: Nice!

src/query/parser/promql/types.go
```diff
@@ -74,6 +74,11 @@ func NewAggregationOperator(expr *promql.AggregateExpr) (parser.Params, error) {
 		return nil, fmt.Errorf("operator not supported: %s", opType)
 	}
 
+	if op == aggregation.CountValuesType {
+		nodeInformation.StringParameter = expr.Param.String()
```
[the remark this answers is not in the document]
arnikola: NodeInformation contains all the parameters/options that may be useful to aggregation functions, similar to how promql uses `promql.AggregateExpr`

src/query/models/tag.go
```diff
@@ -287,13 +287,28 @@ func (t Tags) StringMap() map[string]string {
...
 
-// Add is used to add a bunch of tags and then maintain the sort order
+// Clone returns a copy of the tags
+func (t Tags) Clone() Tags {
+	cloned := make(Tags, len(t))
+	copy(cloned, t)
+	return cloned
+}
+
+// AddTag is used to add a single tag and maintain sorted order
```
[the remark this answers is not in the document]
arnikola: Then I'd need to always make a single element slice when adding a single tag? Seems a bit weird, what's the argument against `AddTag`?

src/query/functions/aggregation/count_values.go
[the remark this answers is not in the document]
arnikola: Changed it to an int for now and filtered it through as appropriate to reduce struct size throughout

src/query/functions/aggregation/count_values.go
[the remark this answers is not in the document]
arnikola: Added an int version of Memset after changing `count` to an `int`

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,229 @@
...
+					currentBucketBlock.columns[columnIndex][i] = math.NaN()
+				}
+			}
+
+			countedValues := countValuesFn(values, bucket)
+			for distinctValue, count := range countedValues {
+				currentBucketColumn := currentBucketBlock.columns[columnIndex]
```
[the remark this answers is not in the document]
arnikola: It screws up some of the appending done in each bucket if it's moved out unfortunately

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,229 @@
...
+		if err != nil {
+			return err
+		}
+
+		values := step.Values()
+		for bucketIndex, bucket := range buckets {
+			currentBucketBlock := tempBlock[bucketIndex]
+			// If not on the first column, generate appropriate number of rows full of NaNs
+			if columnIndex > 0 {
+				previousLen := len(currentBucketBlock.columns[columnIndex-1])
```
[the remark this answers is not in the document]
arnikola: Good call, makes this and a few other length calculations a fair bit nicer

src/query/functions/aggregation/count_values.go
arnikola: Good call, looks better now

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,229 @@
...
+type countValuesNode struct {
+	op         countValuesOp
+	controller *transform.Controller
+}
+
+type bucketColumn []float64
+
+type bucketBlock struct {
+	columns      []bucketColumn
+	indexMapping map[float64]int
```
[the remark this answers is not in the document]
arnikola: Added some comments

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,268 @@
...
+}
+
+type countValuesNode struct {
+	op         countValuesOp
+	controller *transform.Controller
+}
+
+// bucketColumn represents a column of times a particular value in a series has
+// been seen. This may expand as more unique values are seen
+type bucketColumn []int
```
nikunjgit: consider using floats ? int would also be 64 bits. You can also avoid the weird -1 -> NaN conversion

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,268 @@
...
+		params.Without,
+		n.op.opType,
+		seriesMetas,
+	)
+
+	stepCount := stepIter.StepCount()
+	intermediateBlock := make([]bucketBlock, len(buckets))
+	for i := range intermediateBlock {
+		intermediateBlock[i].columns = make([]bucketColumn, stepCount)
+		intermediateBlock[i].indexMapping = make(map[float64]int, 10)
```
nikunjgit: did we decide to move all numbers to a top level constant ? initAlloc ?

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	nextBlock := builder.Build()
+	defer nextBlock.Close()
+	return n.controller.Process(nextBlock)
+}
+
+// converts bucketColumn to a list of floats, with -1 values converted
+// to NaNs, and padded with enough NaNs to match size
+func convertCountsToPaddedFloatList(vals bucketColumn, size int) []float64 {
```
nikunjgit: we cam avoid this is bucketColumn is just float64

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,268 @@
...
+	// Rebuild block metas in the expected order
+	blockMetas := make([]block.SeriesMeta, numSeries)
+	initialIndex := 0
+	for bucketIndex, bucketBlock := range intermediateBlock {
+		for k, v := range bucketBlock.indexMapping {
+			blockMetas[v+initialIndex] = block.SeriesMeta{
+				Name: n.op.OpType(),
+				Tags: metas[bucketIndex].Tags.Clone().AddTag(models.Tag{
+					Name:  n.op.params.StringParameter,
+					Value: strconv.FormatFloat(k, 'f', -1, 64),
```
nikunjgit: probably move this into a util function since we should use the same logic when you output from the render function.

src/query/functions/aggregation/count_values.go
```diff
@@ -0,0 +1,268 @@
...
+	for _, bucketBlock := range intermediateBlock {
+		numSeries += bucketBlock.columnLength
+	}
+
+	// Rebuild block metas in the expected order
+	blockMetas := make([]block.SeriesMeta, numSeries)
+	initialIndex := 0
+	for bucketIndex, bucketBlock := range intermediateBlock {
+		for k, v := range bucketBlock.indexMapping {
+			blockMetas[v+initialIndex] = block.SeriesMeta{
```
nikunjgit: why v + initialIndex ? Its a little hard to understand.

src/query/functions/aggregation/count_values.go
arnikola: https://github.com/m3db/m3/pull/874#discussion_r214414809 :|

src/query/functions/aggregation/count_values.go
arnikola: Added comment

src/query/functions/aggregation/count_values.go
arnikola: Changed this to be a more reasonable initialization size

PR #876, opened 2018-09-02
Use upsert behavior for datapoints written to the mutable series buffer

robskillington:


richardartoul:
Maybe its too much of a pain, but how do you feel about making the conflict resolution strategy configurable? The main reason being:

1) The existing first-write-wins strategy is useful in certain situations (I.E for keeping track of whether or not something has occurred in a given period of time by truncating to a given number of hours and then always writing 1. You could still do this with this change, but it would be a lot more resource intensive.
2) A deterministic strategy (E.X the biggest value always wins) would be really useful for when we're trying to do shadow comparisons of one m3db cluster against another since even though they receive the same writes, they might receive them in slightly different orders

robskillington:
@richardartoul all feedback addressed, if you write the same value to same timestamp multiple times it will be a no-op for creating buffers, and you can control how it selects different datapoints when different values for the same timestamp are being iterated over.

src/dbnode/encoding/iterators.go
```diff
@@ -38,21 +39,71 @@ var (
...
 	return len(i.values)
 }
 
 func (i *iterators) current() (ts.Datapoint, xtime.Unit, ts.Annotation) {
-	return i.earliest.Current()
+	numIters := len(i.earliest)
+
+	switch i.equalTimesStrategy {
+	case IterateHighestValue:
+		sort.Slice(i.earliest, func(a, b int) bool {
```
richardartoul: Is this gonna alloc?

src/dbnode/encoding/iterators_test.go
```diff
@@ -0,0 +1,156 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
richardartoul: 2018

src/dbnode/encoding/iterators_test.go
```diff
@@ -0,0 +1,156 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package encoding
+
+import (
```
richardartoul: Did this file really not have any tests before?

src/dbnode/encoding/iterators_types.go
```diff
@@ -0,0 +1,122 @@
...
+	errIterateEqualTimestampStrategyNotSpecified = errors.New("iterate equal timestamp strategy not specified")
+)
+
+// IterateEqualTimestampStrategy describes the strategy of which value to
+// select when values with equal timestamps appear in the list of iterators.
+type IterateEqualTimestampStrategy uint8
+
+const (
+	// IterateLastPushed is useful for within a single replica, using the last
+	// immutable buffer that was created to decide which value to choose.
```
richardartoul: The iterators have to be given in the correct order though right? Maybe clarify?

src/dbnode/encoding/iterators.go
robskillington: Thankfully no, output of `go build -gcflags "-m" .` in this package:
```
./iterators.go:63:26: (*iterators).current func literal does not escape
./iterators.go:70:26: (*iterators).current func literal does not escape
./iterators.go:87:26: (*iterators).current func literal does not escape
```

src/dbnode/encoding/iterators_test.go
robskillington: Ta, will fix.

src/dbnode/encoding/iterators_test.go
robskillington: Haha, it was covered by multi reader iterator tests before, since this file was split as a subtype of multi reader iterator when series iterator needed the same type of logic.

src/dbnode/encoding/iterators_types.go
robskillington: Sure thing.

src/dbnode/encoding/iterators_types.go
```diff
@@ -0,0 +1,122 @@
...
+	IterateLowestValue
+	// IterateHighestFrequencyValue is useful across replicas when you want to
+	// choose the most common appearing value, however you can only use this
+	// reliably if you wait for values from all replicas to be retrieved, i.e.
+	// you cannot use this reliably with quorum/majority consistency.
+	IterateHighestFrequencyValue
+
+	// DefaultIterateEqualTimestampStrategy is the default iterate
+	// equal timestamp strategy.
+	DefaultIterateEqualTimestampStrategy = IterateEqualTimestampStrategy(0)
```
richardartoul: Can you do `DefaultIterateEqualTimestampStrategy = IterateLastPushed`

src/dbnode/encoding/iterators_types.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+)
+
+// ValidIterateEqualTimestampStrategies returns the valid iterate
+// equal timestamp strategies.
+func ValidIterateEqualTimestampStrategies() []IterateEqualTimestampStrategy {
+	// Return a copy here so callers cannot mutate the known list.
+	result := make([]IterateEqualTimestampStrategy, 0,
+		len(validIterateEqualTimestampStrategies))
+	copy(result, validIterateEqualTimestampStrategies)
```
richardartoul: Is this broken? my understanding is that `copy` will copy `min(dst, src)` and `len(dst)` here is zero

src/dbnode/encoding/iterators_types_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"fmt"
+	"testing"
+	"unsafe"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	yaml "gopkg.in/yaml.v2"
+)
+
+func TestValidIterateEqualTimestampStrategies(t *testing.T) {
```
richardartoul: If my previous comment is right, you might want to add a length test here

src/dbnode/encoding/series_iterator.go
```diff
@@ -101,10 +97,17 @@ func (it *seriesIterator) Close() {
 		return
 	}
 	it.closed = true
-	it.id.Finalize()
-	it.nsID.Finalize()
+	if it.id != nil {
```
richardartoul: What changed that we suddenly needed these nil checks?

src/dbnode/encoding/series_iterator.go
```diff
@@ -121,19 +124,25 @@ func (it *seriesIterator) Replicas() []MultiReaderIterator {
...
+	it.tags = opts.Tags
+	if it.tags == nil {
+		it.tags = ident.EmptyTagIterator
+	}
+	it.start = opts.StartInclusive
+	it.end = opts.EndExclusive
 	it.iters.reset()
-	it.iters.setFilter(startInclusive, endExclusive)
+	if !it.start.IsZero() && !it.end.IsZero() {
+		it.iters.setFilter(it.start, it.end)
```
richardartoul: does this do the right thing is one of them is zero?

src/dbnode/encoding/types.go
```diff
@@ -42,6 +42,14 @@ type Encoder interface {
 	// Stream is the streaming interface for reading encoded bytes in the encoder.
 	Stream() xio.SegmentReader
 
+	// NumEncoded returns the number of encoded datapoints.
+	NumEncoded() int
+
+	// LastEncoded returns the last encoded datapoint, useful for
+	// de-duplicating encoded values. If there no values encoded previously
```
richardartoul: `If there are no previously encoded values...`

src/dbnode/persist/fs/commitlog/commit_log_test.go
```diff
@@ -437,7 +437,7 @@ func TestReadCommitLogMissingMetadata(t *testing.T) {
 
 func TestCommitLogReaderIsNotReusable(t *testing.T) {
 	// Make sure we're not leaking goroutines
-	defer leaktest.CheckTimeout(t, time.Second)()
+	defer leaktest.CheckTimeout(t, 10*time.Second)()
```
richardartoul: Not sure if bumping this is necessary, you might just need to rebase master. Prateek upgraded to the latest version of leaktest which fixes this

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -859,6 +859,8 @@ func (s *commitLogSource) startM3TSZEncodingWorker(
 			wroteExisting  = false
 		)
 		for i := range unmergedBlock {
+			// TODO(r): Write unit test to ensure that different values that arrive
```
richardartoul: This is a good point, did you handle this anywhere? I see a TODO for a unit test, but I don't see the logic at all.

src/dbnode/storage/series/buffer.go
```diff
@@ -581,38 +580,95 @@ func (b *dbBufferBucket) write(
...
-		bopts := b.opts.DatabaseBlockOptions()
-		blockSize := b.opts.RetentionOptions().BlockSize()
-		blockAllocSize := bopts.DatabaseBlockAllocSize()
-		encoder := bopts.EncoderPool().Get()
-		encoder.Reset(timestamp.Truncate(blockSize), blockAllocSize)
-		next := inOrderEncoder{encoder: encoder}
-		b.encoders = append(b.encoders, next)
-		idx = len(b.encoders) - 1
+
+	// Upsert/last-write-wins semantics.
```
richardartoul: I was skimming through the encoder code and its not clear to me why we need to create a new encoder for this scenario. Does it simplify the iteration logic? it seemed like the encoder could encode multiple different values for the same timestamp

src/dbnode/encoding/iterators_types.go
robskillington: Sure thing.

src/dbnode/encoding/iterators_types.go
robskillington: Ah yes, good call.  I'll add a test for this too.

src/dbnode/encoding/iterators_types_test.go
robskillington: I added a values check, in addition to the ptr mismatch test.

src/dbnode/encoding/series_iterator.go
robskillington: No new change, I'm just making sure we do the same nil check for all fields that are nil-able here.

src/dbnode/storage/series/buffer.go
robskillington: It could, it's just way harder and I don't really want to enforce encoders have to deal with this case right now.  We can revisit this later.

src/dbnode/storage/series/buffer.go
```diff
@@ -581,38 +580,95 @@ func (b *dbBufferBucket) write(
...
+	annotation []byte,
+) error {
+	err := b.encoders[idx].encoder.Encode(datapoint, unit, annotation)
+	if err != nil {
+		return err
+	}
+
+	b.encoders[idx].lastWriteAt = datapoint.Timestamp
+
+	if b.empty {
```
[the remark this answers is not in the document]
robskillington: Yeah, it's kind of buggy and nasty to keep this tracking everywhere.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
robskillington: I added a link to issue I created:
https://github.com/m3db/m3/issues/898

src/dbnode/storage/series/buffer.go
```diff
@@ -545,8 +543,22 @@ func (b *dbBufferBucket) finalize() {
 	b.resetBootstrapped()
 }
 
+func (b *dbBufferBucket) empty() bool {
```
richardartoul: you think this will be ok perf-wise? I assume thats why we had the bool to begin with

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -90,8 +90,8 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	result, err := h.read(ctx, w, params)
```
richardartoul: why are a bunch of query changes showing up here?

src/dbnode/storage/series/buffer.go
robskillington: Yeah this should be fine, we don't actually call it from any high frequency code paths.  We used to have a high frequency call site with it, although now it was refactored out.

src/dbnode/encoding/series_iterator.go
```diff
@@ -45,15 +45,11 @@ type seriesIterator struct {
 // NewSeriesIterator creates a new series iterator.
 // NB: The returned SeriesIterator assumes ownership of the provided `ident.ID`.
 func NewSeriesIterator(
-	id ident.ID,
-	nsID ident.ID,
-	tags ident.TagIterator,
-	startInclusive, endExclusive time.Time,
-	replicas []MultiReaderIterator,
+	opts SeriesIteratorOptions,
```
prateek: +1 for this

src/dbnode/encoding/null.go
```diff
@@ -45,6 +45,10 @@ func (e *nullEncoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, annotation ts
 func (e *nullEncoder) Stream() xio.SegmentReader {
 	return xio.NewSegmentReader(ts.Segment{})
 }
+func (e *nullEncoder) NumEncoded() int { return 0 }
+func (e *nullEncoder) LastEncoded() (ts.Datapoint, error) {
+	return ts.Datapoint{}, fmt.Errorf("not implemented")
```
prateek: nit: use a const `errors.New()` instead of the `fmt.Errorf` here

src/dbnode/storage/series/buffer.go
```diff
@@ -752,6 +802,8 @@ func (b *dbBufferBucket) merge() (mergeResult, error) {
 		}
 	}()
 
+	// Rank bootstrapped blocks as data that has appeared before data that
```
prateek: hm can this be violated during a topology change?

src/dbnode/storage/series/buffer.go
```diff
@@ -581,38 +591,85 @@ func (b *dbBufferBucket) write(
...
-			// ones we've already encoded. Immutable/first-write-wins semantics.
-			return nil
+		lastWriteAt := b.encoders[i].lastWriteAt
+		if timestamp.Equal(lastWriteAt) {
+			last, err := b.encoders[i].encoder.LastEncoded()
+			if err != nil {
+				return err
+			}
+			if last.Value == value {
+				// No-op since matches the current value
```
prateek: +1

src/dbnode/encoding/iterators.go
```diff
@@ -38,21 +39,71 @@ var (
...
+	valueFrequencies map[float64]int
 }
 
 func (i *iterators) len() int {
 	return len(i.values)
 }
 
 func (i *iterators) current() (ts.Datapoint, xtime.Unit, ts.Annotation) {
-	return i.earliest.Current()
+	numIters := len(i.earliest)
```
prateek: could this function be implemented using a heap instead? wondering if we'd avoid sorting for each call to `current()` that way

src/dbnode/encoding/iterators.go
```diff
@@ -38,21 +39,71 @@ var (
...
+		})
+
+	case IterateLowestValue:
+		sort.Slice(i.earliest, func(a, b int) bool {
+			currA, _, _ := i.earliest[a].Current()
+			currB, _, _ := i.earliest[b].Current()
+			return currA.Value > currB.Value
+		})
+
+	case IterateHighestFrequencyValue:
```
prateek: lol this feels so magic-y. Do you have a use in mind for it?

src/dbnode/encoding/iterators.go
```diff
@@ -38,21 +39,71 @@ var (
...
 type iterators struct {
-	values      []Iterator
-	earliest    Iterator
-	earliestIdx int
-	earliestAt  time.Time
-	filtering   bool
-	filterStart time.Time
-	filterEnd   time.Time
+	values             []Iterator
+	earliest           []Iterator
```
prateek: Can I convince you to write a little blurb about the intent behind how `earliest` and `values` are used in the code? I think I follow based on the implementation of `moveToValidNext()` but it's a little convoluted the first time someone looks at it.

src/dbnode/storage/series/buffer.go
robskillington: This is mainly a best effort change to be honest, as we know without keeping a timestamp next to each value of when it was written we can't guarantee selecting the last upserted value when reading values returned from multiple replicas.

It *should* be a good approximation however, which is what we need right now.  When we are ready to begin solving it by storing metadata next to the values we can offer configuration for either strict or best effort upserts.

src/dbnode/encoding/iterators.go
robskillington: Sure thing, can do.

src/dbnode/encoding/iterators.go
robskillington: As per docs its basically if you are using `consistency: all` and don't mind the read unavailability when not all replicas are available to be read from (say perhaps exact match mode for testing correctness for shadow traffic, etc).

src/dbnode/encoding/iterators.go
robskillington: We can look at this in a followup change, currently none of the non-default strategies aren't likely to be called very often (only subsets of queries/requests, etc, hence using the setter at a per series iterator granularity).

When we have a more complete implementation that can actually return you the last written value, say with metadata stored alongside the value, we should then look at optimizing this.  This is mainly an ergonomic change useful for edge cases rather than supposed to be definitive and optimized.

PR #877, opened 2018-09-03
Add default Prometheus query range route for Grafana integration

robskillington:


src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -61,7 +61,15 @@ func parseTime(r *http.Request, key string) (time.Time, error) {
 // nolint: unparam
 func parseDuration(r *http.Request, key string) (time.Duration, error) {
 	if d := r.FormValue(key); d != "" {
-		return time.ParseDuration(d)
+		value, firstErr := time.ParseDuration(d)
+		if firstErr != nil {
+			// Try parsing as an integer value specifying seconds, the Prometheus default
+			if seconds, err := strconv.ParseInt(d, 10, 64); err == nil {
+				value = time.Duration(seconds) * time.Second
```
arnikola: nit: `return time.Duration(seconds) * time.Second, nil`?

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -128,38 +136,47 @@ func parseParams(r *http.Request) (models.RequestParams, *handler.ParseError) {
 	return params, nil
 }
 
-func parseTarget(r *http.Request) (string, error) {
-	targetQueries, ok := r.URL.Query()[targetParam]
-	if !ok || len(targetQueries) == 0 || targetQueries[0] == "" {
+func parseQuery(r *http.Request) (string, error) {
+	queries, ok := r.URL.Query()[queryParam]
+	if !ok || len(queries) == 0 || queries[0] == "" {
 		return "", errors.ErrNoTargetFound
```
arnikola: nit: rename `ErrNoTargetFound` to `ErrNoQueryFound`?

src/query/api/v1/handler/prometheus/native/common_test.go
```diff
@@ -82,3 +88,103 @@ func TestValueToProm(t *testing.T) {
...
+					],
+					"step_size_ms": 0
+				}
+			]
+		}
+	}
+	`)
+
+	actual := mustPrettyJSON(t, buffer.String())
+
```
arnikola: nit: unnecessary newline

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -42,7 +42,7 @@ import (
 const (
 	endParam          = "end"
 	startParam        = "start"
-	targetParam       = "target"
+	queryParam        = "query"
```
arnikola: Is this going to work fine for the prometheus remote read end point?

arnikola: Oh wait, this is the native end point, not the remote read one. Never mind!

arnikola:
There's a couple of places which still use `target` rather than `query`, notably `RequestParams`

src/query/api/v1/handler/prometheus/native/common.go
robskillington: Good call, also refactored the initial form value if/else too.

src/query/api/v1/handler/prometheus/native/common.go
robskillington: Sounds good.

src/query/ts/values.go
```diff
@@ -107,17 +107,17 @@ func (b *fixedResolutionValues) StartTime() time.Time {
 
 // Resolution returns resolution per step
 func (b *fixedResolutionValues) Resolution() time.Duration {
-	return b.MillisPerStep()
+	return b.MillisPerStep() * time.Millisecond
```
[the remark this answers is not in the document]
arnikola: Looks like semaphore's failing a few unit tests, this seems like the likely culprit

src/cmd/services/m3query/scripts/prom-m3-diff.sh
```diff
@@ -10,13 +10,14 @@ m3port="localhost:7201"
...
 curl -fsS $promport/status > /dev/null || { echo "Prom port not open";  exit 1; }
 curl -fsS $m3port/health > /dev/null || { echo "M3Query port not open";  exit 1; }
-m3command="$m3port/api/v1/prom/native/read?start=$start&end=$end&step=$step&debug=true --data-urlencode target=$target"
-promcommand="$promport/api/v1/query_range?start=$start&end=$end&step=$step --data-urlencode query=$target"
-echo $m3command
-echo $promcommand
-curl -G $m3command > m3out
-curl -G $promcommand > promout
-jq ".[]|.tags,.datapoints" m3out > m3result
+queryurl="/api/v1/query_range?start=$start&end=$end&step=$step --data-urlencode query=$target"
```
nikunjgit: sweet!

scripts/integration-tests/prometheus/prometheus-integration-test.sh
```diff
@@ -126,6 +126,8 @@ echo "Sleep for 30 seconds to let the remote write endpoint generate some data"
 
 sleep 30
 
+# Ensure Prometheus can proxy a Prometheus query
```
nikunjgit: do you also want to test m3query directly ?

src/query/api/v1/httpd/handler.go
```diff
@@ -120,7 +122,14 @@ func (h *Handler) RegisterRoutes() error {
 
 	h.Router.HandleFunc(remote.PromReadURL, logged(promRemoteReadHandler).ServeHTTP).Methods(remote.PromReadHTTPMethod)
 	h.Router.HandleFunc(remote.PromWriteURL, logged(promRemoteWriteHandler).ServeHTTP).Methods(remote.PromWriteHTTPMethod)
-	h.Router.HandleFunc(native.PromReadURL, logged(native.NewPromReadHandler(h.engine)).ServeHTTP).Methods(native.PromReadHTTPMethod)
+
+	// Prometheus native query endpoints, registered under both the default
+	// Prometheus server route and the Prometheus native read endpoint route
+	promNativeRead := logged(native.NewPromReadHandler(h.engine)).ServeHTTP
+	h.Router.HandleFunc(prometheus.DefaultQueryRangeURL, promNativeRead).Methods(prometheus.DefaultQueryRangeMethod)
```
nikunjgit: does it make sense to have both or shall we just update native.PromReadURL to point to /api/v1/query_range ?

PR #882, opened 2018-09-05
Remove free form JSON response

nerd0:
Currently if no placements are found a free formed JSON body is returned. It might be easier to just return a `404` which won't break the use of the `PlacementGetResponse` 

The other way to approach this would be to expose a `Result` field within the `PlacementGetResponse` as a `string` to provide informational status of responses.


src/query/api/v1/handler/placement/get_test.go
```diff
@@ -113,6 +113,5 @@ func TestPlacementGetHandler(t *testing.T) {
 
 	resp = w.Result()
 	body, _ = ioutil.ReadAll(resp.Body)
```
justinjc: I think metalint is complaining about this useless line now that we're not using it.

Issue #865, opened 2018-08-29
PromQL query returns unrelated series


sboschman:
As mentioned on Gitter, I run into some strange issue with queries against M3DB returning unrelated series. This results in Grafana throwing errors like "many-to-many matching not allowed" or "Multiple Series Error" on f.e. the default Prometheus node-exporter dashboard.

---

Prometheus: 2.3.2
M3Coordinator: 0.4.1 (running on the same host as prometheus)
M3DB: 0.4.1 (4 node cluster, 2 replicasets of 2 nodes each with 3 seednodes)
Prometheus config: read_recent: true

PromQL (latest value):
```
up{job="node-exporter",` instance="m3db-node04:9100"}
```
Result (omitted the values):
```
up{instance="m3db-node04:9100",job="node-exporter"}
```

---

PromQL (1 day window):
```
up{job="node-exporter",` instance="m3db-node04:9100"}[1d]
```
Result:
```
up{instance="m3db-node04:9100",job="node-exporter"}
node_memory_CmaFree_bytes{instance="m3db-node03:9100",job="node-exporter"}
```
Didn't expect to get a 'node_memory_CmaFree_bytes' serie, doesn't even match the instance...

---

PromQL (2 day window):
```
up{job="node-exporter",` instance="m3db-node04:9100"}[2d]
```
Result:
```
    up{instance="m3db-node04:9100",job="node-exporter"}
    node_network_transmit_packets_total{device="lo",instance="m3db-node04:9100",job="node-exporter"}
    node_memory_CmaFree_bytes{instance="m3db-node03:9100",job="node-exporter"}
    go_memstats_alloc_bytes{instance="m3db-node03:9100",job="node-exporter"}
```
These extra series have values up to 'now', so they are not old series or something.

---

PromQL (1 week window):
A total of 6 series show up.

---

PromQL (2 week window):
```
up
```
Result:
All the 'up' series show up (different instance tags ofc), but no other series.

---

I have two Prometheus instances scraping the same nodes. On the 2nd node I changed the config to `read_recent: false`. The 2nd node only shows the requested serie, as expected.

richardartoul:
@nikunjgit @robskillington Does one of you have an idea what might be going on here? I'm not really familiar with the Prometheus stuff

robskillington:
I'll look into this, might have something to do with the matchers and special characters (wonder if we're specifying regexp instead of exact match for some of the matchers we're sending to the index).

sboschman:
[...]

PR #883, opened 2018-09-05
Rework searchers to execute on a single reader

prateek:
The m3db index is composed of time based index blocks, each index block is a complete index for a particular time range. Each index block is composed of segments (from writes and/or bootstraps). m3ninx's searcher implementation assumed these segments did not share document ID ranges (by ID, I mean postings.ID; not []byte). This guarantee isn't possible with m3db, because we accept writes before we bootstrap (or after, in the case of topology changes). This PR relaxes the restriction -- it makes the query searchers operate on a segment at a time, instead of all the segments at once. This will work out better for query optimisation too. 

Fixes one of the issues in https://github.com/m3db/m3/issues/865


src/m3ninx/search/proptest/issue865_test.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/m3db/m3/src/m3ninx/search/query"
+
+	"github.com/leanovate/gopter"
+	"github.com/leanovate/gopter/gen"
+	"github.com/leanovate/gopter/prop"
+	"github.com/stretchr/testify/require"
+)
+
+// NB(prateek): this test simulates the issues described in issue: https://github.com/m3db/m3/issues/865
+// tl;dr - the searcher code assumes the input readers had disjoint doc ID ranges; it caused issues when that
```
jeromefroe: Does this imply that the doc ID's can be reused? Or that a doc ID can exist in multiple segments?

PR #884, opened 2018-09-06
Arnikola/resolve parameters

arnikola:
Add function to resolve scalar parameters for functions that require scalar inputs.

### DIFF
--- a/src/query/functions/aggregation/take.go
+++ b/src/query/functions/aggregation/take.go
@@ -28,6 +28,7 @@ import (
 	"github.com/m3db/m3/src/query/executor/transform"
 	"github.com/m3db/m3/src/query/functions/utils"
 	"github.com/m3db/m3/src/query/parser"
+	"github.com/m3db/m3/src/query/ts"
 )
 
 const (
@@ -39,7 +40,7 @@ const (
 
 type takeFunc func(values []float64, buckets [][]int) []float64
 
-// NewTakeOp creates a new take operation
+// NewTakeOp creates a new takeK operation
 func NewTakeOp(
 	opType string,
 	params NodeParams,
@@ -49,9 +50,17 @@ func NewTakeOp(
 		return baseOp{}, fmt.Errorf("operator not supported: %s", opType)
 	}
 
-	heap := utils.NewFloatHeap(takeTop, int(params.Parameter))
-	fn := func(values []float64, buckets [][]int) []float64 {
-		return takeFn(heap, values, buckets)
+	var fn takeFunc
+	k := int(params.Parameter)
+	if k < 1 {
+		fn = func(values []float64, buckets [][]int) []float64 {
+			return takeNone(values, buckets)
+		}
+	} else {
+		heap := utils.NewFloatHeap(takeTop, k)
+		fn = func(values []float64, buckets [][]int) []float64 {
+			return takeFn(heap, values, buckets)
+		}
 	}
 
 	return newTakeOp(params, opType, fn), nil
@@ -142,6 +151,12 @@ func (n *takeNode) Process(ID parser.NodeID, b block.Block) error {
 	nextBlock := builder.Build()
 	defer nextBlock.Close()
 	return n.controller.Process(nextBlock)
+}
+
+// shortcut to return empty when taking <= 0 values
+func takeNone(values []float64, buckets [][]int) []float64 {
+	ts.Memset(values, math.NaN())
+	return values
 }
 
 func takeFn(heap utils.FloatHeap, values []float64, buckets [][]int) []float64 {
--- a/src/query/functions/binary/base.go
+++ b/src/query/functions/binary/base.go
@@ -64,6 +64,25 @@ func (o baseOp) Node(controller *transform.Controller, _ transform.Options) tran
 	}
 }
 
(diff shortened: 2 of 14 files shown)


src/query/functions/binary/base.go
```diff
@@ -64,6 +64,23 @@ func (o baseOp) Node(controller *transform.Controller, _ transform.Options) tran
 	}
 }
 
+// ArithmeticFunction returns the arithmetic function for this operation type.
+func ArithmeticFunction(opType string, returnBool bool) (Function, error) {
+	if fn, ok := arithmeticFuncs[opType]; ok {
+		return fn, nil
+	}
+
+	if returnBool {
```
benraskin92: comment?

src/query/functions/binary/base_test.go
```diff
@@ -0,0 +1,50 @@
...
+package binary
+
+import (
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestArithmeticFunction(t *testing.T) {
```
benraskin92: Add tests for the other arithmetic funcs?

src/query/functions/utils/heap.go
```diff
@@ -64,8 +66,13 @@ func NewFloatHeap(isMaxHeap bool, capacity int) FloatHeap {
 		less = minHeapLess
 	}
 
+	var heapInit = capacity
+	if capacity < 1 {
+		heapInit = initIndexSliceLength
```
benraskin92: Can't you just keep `capacity`? i.e. 
```
	if capacity < 1 {
		capacity = initIndexSliceLength
        }
```

src/query/functions/utils/heap.go
```diff
@@ -124,6 +131,11 @@ func (fh FloatHeap) Len() int {
 
 // Cap returns the capacity of the heap
 func (fh FloatHeap) Cap() int {
+	// If capacity is less than 1, return 0
+	if fh.capacity < 1 {
```
nikunjgit: remove ? Doesn't look like it can be < 1

src/query/parser/promql/types.go
```diff
@@ -76,13 +75,12 @@ func NewAggregationOperator(expr *promql.AggregateExpr) (parser.Params, error) {
 	}
 
 	if op == aggregation.BottomKType || op == aggregation.TopKType {
-		paramString := expr.Param.String()
-		floatParam, err := strconv.ParseFloat(paramString, 64)
+		val, err := resolveScalarArgument(expr.Param)
```
nikunjgit: consider calling this for all aggregation types. Otherwise, picking and choosing will be hard :(

arnikola: Only these two agg functions take scalar args, so not necessary to call it for others

src/query/parser/promql/parse.go
```diff
@@ -122,20 +122,27 @@ func (p *parseState) walk(node pql.Node) error {
...
+					return err
+				}
+
+				argValues = append(argValues, val)
+			} else {
+				if e, ok := expr.(*pql.MatrixSelector); ok {
+					argValues = append(argValues, e.Range)
+				}
+
+				err := p.walk(expr)
```
nikunjgit: if err := p.walk(expr); err != nil {return}

src/query/parser/promql/resolve_scalars.go
```diff
@@ -0,0 +1,124 @@
...
+
+	if nesting != 0 {
+		return 0, fmt.Errorf("promql.resolveScalarArgument: invalid nesting %d", nesting)
+	}
+
+	return value, nil
+}
+
+// resolves an expression which should resolve to a scalar argument
+func resolveScalarArgumentWithNesting(expr pql.Expr, nesting *int) (float64, error) {
```
nikunjgit: instead of int pointer, let's have multiple return values ?

arnikola: Initially tried it with multiple return values and it wasn't working correctly; can look at it again but don't want to spend too long :)

src/query/functions/aggregation/take.go
```diff
@@ -49,9 +50,17 @@ func NewTakeOp(
...
-	fn := func(values []float64, buckets [][]int) []float64 {
-		return takeFn(heap, values, buckets)
+	var fn takeFunc
+	k := int(params.Parameter)
+	if k < 1 {
+		fn = func(values []float64, buckets [][]int) []float64 {
+			return takeNone(values, buckets)
+		}
+	} else {
+		heap := utils.NewFloatHeap(takeTop, k)
```
benraskin92: What if k >2?

arnikola: Same thing as if k == 2? Don't understand the question

benraskin92: sorry, misread the code..

src/query/functions/utils/heap_test.go
```diff
@@ -193,3 +194,25 @@ func TestMinHeap(t *testing.T) {
 		})
 	}
 }
+
+func TestNegativeCapacityHeap(t *testing.T) {
```
nikunjgit: should the client be responsible for ensuring non negative capacity ?

PR #885, opened 2018-09-06
Add m3ninx query prop tests

prateek:
Pending:
- [x] address regexp anchoring issue for simple v fst segment 
- [x] gopter SliceOf reflection issue during query gen


src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"bytes"
+	re "regexp"
+
+	vregex "github.com/couchbase/vellum/regexp"
+)
+
+// CompileRegex compiles the provided regexp.
+func CompileRegex(r []byte) (CompiledRegex, error) {
```
robskillington: Hey as we briefly discussed, could we instead use validation and just ensure that callers have fully anchored their regexps?

It is more explicit that way, rather than coercing their regexps into being fully anchored.  

I don't mind too much either way, but that seems less magical just responding with an error saying "hey please fully anchor your regexp with "^pattern$".

src/m3ninx/index/segment/mem/reader.go
```diff
@@ -83,7 +84,11 @@ func (r *reader) MatchRegexp(field, regexp []byte, compiled index.CompiledRegex)
 	// with a postings list through a call to Docs will IDs greater than the maximum be
 	// filtered out.
 	compileRE := compiled.Simple
-	pl, err := r.segment.matchRegexp(field, regexp, compileRE)
+	if compileRE == nil {
+		return nil, errReaderNilRegex
+	}
+
+	pl, err := r.segment.matchRegexp(field, compileRE)
```
robskillington: nit: Instead perhaps just `return r.segment.matchRegexp(field, compileRE)`?

src/m3ninx/index/regexp.go
prateek: done in https://github.com/m3db/m3/pull/902

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
```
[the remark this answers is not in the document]
prateek: will do

src/m3ninx/index/segment/mem/reader.go
prateek: will do

src/m3ninx/search/proptest/prop_test.go
```diff
@@ -0,0 +1,147 @@
...
+	seed := time.Now().UnixNano()
+	parameters.MinSuccessfulTests = 100
+	parameters.MaxSize = 20
+	parameters.Rng = rand.New(rand.NewSource(seed))
+	properties := gopter.NewProperties(parameters)
+
+	simpleSeg := newTestMemSegment(t, lotsTestDocuments)
+	fstSeg := fst.ToTestSegment(t, simpleSeg, fstOptions)
+
+	properties.Property("Simple & FST Segments Query the same results", prop.ForAll(
```
robskillington: Ah nice, this is pretty great.

PR #887, opened 2018-09-06
Allow remote client only setup for query

nikunjgit:


src/query/server/server.go
```diff
@@ -357,6 +395,23 @@ func newStorages(
...
 }
 
+func remoteClient(
+	logger *zap.Logger,
+	cfg config.Configuration,
+) (storage.Storage, bool) {
+	if remotes := cfg.RPC.RemoteListenAddresses; len(remotes) > 0 {
+		client, err := tsdbRemote.NewGrpcClient(remotes)
+		if err != nil {
+			logger.Fatal("unable to start remote clients for addresses", zap.Any("error", err))
```
prateek: nit: don't Fatalf in subfunctions, pass the error back up and fatal in `main` or the top-level `Run`

src/query/server/server.go
```diff
@@ -319,6 +318,50 @@ func newDownsampler(
...
+		session := m3db.NewAsyncSession(func() (client.Client, error) {
+			return <-dbClientCh, nil
+		}, nil)
+		clusters, err = local.NewClusters(local.UnaggregatedClusterNamespaceDefinition{
+			NamespaceID: ident.StringID(localCfg.Namespace),
+			Session:     session,
+			Retention:   localCfg.Retention,
+		})
+		if err != nil {
+			logger.Fatal("unable to connect to clusters", zap.Any("error", err))
```
prateek: here as well

src/query/server/server.go
```diff
@@ -319,6 +318,50 @@ func newDownsampler(
...
+		err      error
+	)
+
+	if len(cfg.Clusters) > 0 {
+		opts := local.ClustersStaticConfigurationOptions{
+			AsyncSessions: true,
+		}
+		clusters, err = cfg.Clusters.NewClusters(opts)
+		if err != nil {
+			logger.Fatal("unable to connect to clusters", zap.Any("error", err))
```
prateek: nit: `zap.Error(err)` instead of `zap.Any(..`

prateek: also, same point about fatalf here

src/query/server/server.go
```diff
@@ -234,44 +263,14 @@ func Run(runOpts RunOptions) {
...
-
-	logger.Info("starting server", zap.String("address", listenAddress))
-	go func() {
-		if err := http.ListenAndServe(listenAddress, handler.Router); err != nil {
-			logger.Fatal("unable to serve on listen address",
-				zap.String("address", listenAddress), zap.Error(err))
+	cleanup := func() {
+		storageCleanup()
+		if err := clusters.Close(); err != nil {
+			logger.Fatal("unable to close M3DB cluster sessions", zap.Any("error", err))
```
prateek: here as well

src/cmd/services/m3query/config/config.go
```diff
@@ -52,6 +52,8 @@ type Configuration struct {
 	// RPC is the RPC configuration.
 	RPC *RPCConfiguration `yaml:"rpc"`
 
+	Backend string `yaml:"backend"`
```
prateek: nit: comment for field

src/query/server/server.go
```diff
@@ -357,6 +395,23 @@ func newStorages(
 	return fanoutStorage, cleanup
 }
 
+func remoteClient(
```
arnikola: nit: `makeRemoteClient`?

src/query/server/server.go
```diff
@@ -119,12 +120,75 @@ func Run(runOpts RunOptions) {
...
+			logger.Fatal("need remote clients for grpc backend")
+		}
+	} else {
+		backendStorage, clusterClient, downsampler, cleanup = setupM3DBStorage(runOpts, cfg, logger, scope)
+	}
+
+	defer cleanup()
+	engine := executor.NewEngine(backendStorage)
+
+	handler, err := httpd.NewHandler(backendStorage, downsampler, engine,
```
arnikola: Any chance of issues if clusterClient and downsampler are nil here?

src/query/server/server.go
```diff
@@ -119,12 +120,75 @@ func Run(runOpts RunOptions) {
...
+
+	if cfg.Backend == "grpc" {
+		if backendStorage, enabled = remoteClient(logger, cfg); !enabled {
+			logger.Fatal("need remote clients for grpc backend")
+		}
+	} else {
+		backendStorage, clusterClient, downsampler, cleanup = setupM3DBStorage(runOpts, cfg, logger, scope)
+	}
+
+	defer cleanup()
```
arnikola: This would panic if cleanup is nil, might need to define it as a noop function

src/query/server/server.go
```diff
@@ -357,6 +395,23 @@ func newStorages(
 	return fanoutStorage, cleanup
 }
 
+func remoteClient(
+	logger *zap.Logger,
+	cfg config.Configuration,
+) (storage.Storage, bool) {
+	if remotes := cfg.RPC.RemoteListenAddresses; len(remotes) > 0 {
```
arnikola: nit Can reduce nesting by 
```
remotes := cfg.RPC.RemoteListenAddresses
if len(remotes) == 0 {
 return nil, false
}
...
```

src/query/server/server.go
```diff
@@ -337,14 +380,9 @@ func newStorages(
...
 
-		if remotes := cfg.RPC.RemoteListenAddresses; len(remotes) > 0 {
-			client, err := tsdbRemote.NewGrpcClient(remotes)
-			if err != nil {
-				logger.Fatal("unable to start remote clients for addresses", zap.Any("error", err))
-			}
-
-			stores = append(stores, remote.NewStorage(client))
-			remoteEnabled = true
+		if remoteStorage, enabled := remoteClient(logger, cfg); remoteEnabled {
```
arnikola: This seems odd. Should this be `if remoteStorage, enabled := remoteClient(logger, cfg); enabled {` ?

src/query/server/server.go
```diff
@@ -359,28 +383,38 @@ func initClusters(cfg config.Configuration, dbClientCh <-chan client.Client, log
...
 }
 
 func newStorages(
 	logger *zap.Logger,
 	clusters local.Clusters,
 	cfg config.Configuration,
 	workerPool pool.ObjectPool,
-) (storage.Storage, func()) {
-	cleanup := func() {}
+) (storage.Storage, func() error, error) {
```
arnikola: nit: make this a type, something like `type cleanup func() error`

src/query/server/server.go
```diff
@@ -230,77 +273,57 @@ func Run(runOpts RunOptions) {
...
-			logger.Fatal("unable to serve on listen address",
-				zap.String("address", listenAddress), zap.Error(err))
+	cleanup := func() error {
+		lastErr := storageCleanup()
+		// Don't want to quit on the first error since the full cleanup is important
+		if err := clusters.Close(); err != nil {
+			if lastErr == nil {
+				lastErr = err
+			} else {
+				// Make sure the previous error is atleast logged
```
benraskin92: `s/atleast/at least`

src/cmd/services/m3query/config/config.go
```diff
@@ -52,6 +52,9 @@ type Configuration struct {
 	// RPC is the RPC configuration.
 	RPC *RPCConfiguration `yaml:"rpc"`
 
+	// Backend is the backend store for query service. We currently support grpc and m3db (default).
+	Backend string `yaml:"backend"`
```
prateek: nit: make an enumerated type for this

```

type BackendStorageType string 

const (
  GRPCStorageType BackendStorageType = "grpc"
  M3DBStorageType BackendStorageType = "m3db"
)

type Configuration struct {

... 
// Backend ...
Backend BackendStorageType ...
```

Probably need to add validation too.

src/query/server/server.go
```diff
@@ -119,12 +121,82 @@ func Run(runOpts RunOptions) {
...
 	}
 
+	var (
+		backendStorage storage.Storage
+		clusterClient  clusterclient.Client
+		downsampler    downsample.Downsampler
+		enabled        bool
+	)
+
+	if cfg.Backend == "grpc" {
```
prateek: nit: use enum consts

src/query/server/server.go
```diff
@@ -119,12 +121,82 @@ func Run(runOpts RunOptions) {
...
+	}
+
+	select {
+	case <-sigChan:
+	case <-interruptCh:
+	}
+
+}
+
+func setupM3DBStorage(
```
prateek: naming nit: `newM3DBStorage` would align with your other names better

src/query/server/server.go
```diff
@@ -114,17 +118,108 @@ func Run(runOpts RunOptions) {
 	logger := logging.WithContext(ctx)
 	defer logger.Sync()
 
-	scope, _, err := cfg.Metrics.NewRootScope()
+	scope, closer, err := cfg.Metrics.NewRootScope()
```
benraskin92: 👍

src/query/api/v1/httpd/handler.go
```diff
@@ -62,7 +59,6 @@ var (
 // Handler represents an HTTP handler.
 type Handler struct {
 	Router        *mux.Router
-	CLFLogger     *log.Logger
```
arnikola: Why are we removing the logger here?

src/query/server/server.go
```diff
@@ -114,17 +118,111 @@ func Run(runOpts RunOptions) {
...
+	}
+}
+
+// make connections to the m3db cluster(s) and generate sessions for those clusters along with the storage
+func newM3DBStorage(
+	runOpts RunOptions,
+	cfg config.Configuration,
+	logger *zap.Logger,
+	scope tally.Scope,
+) (storage.Storage, clusterclient.Client, downsample.Downsampler, cleanupFn, error) {
```
arnikola: nit: 5 return args makes it a bit harder to keep track of what's going on. Your call to change or leave it

src/query/server/server.go
```diff
@@ -150,40 +248,9 @@ func Run(runOpts RunOptions) {
...
-			logger.Fatal("unable to connect to clusters", zap.Any("error", err))
-		}
-	}
-
-	for _, namespace := range clusters.ClusterNamespaces() {
-		logger.Info("resolved cluster namespace",
-			zap.String("namespace", namespace.NamespaceID().String()))
+	clusters, err := initClusters(cfg, runOpts.DBClient, logger)
+	if err != nil {
+		return nil, nil, nil, nil, err
```
arnikola: nit: Wrap this with "unable to init clusters" or something?

src/query/server/server.go
```diff
@@ -211,8 +278,10 @@ func Run(runOpts RunOptions) {
 		return workerPool
 	})
 
-	fanoutStorage, storageCleanup := newStorages(logger, clusters, cfg, objectPool)
-	defer storageCleanup()
+	fanoutStorage, storageCleanup, err := newStorages(logger, clusters, cfg, objectPool)
+	if err != nil {
+		return nil, nil, nil, nil, err
```
arnikola: nit: wrap this with "unable to set up storages"?

src/query/server/server.go
arnikola: Oh I see you do this in the initClusters function; reads a little weird that you're wrapping one inside the function while wrapping the other one above; might be better to extract the err wrapping for initClusters and put it here

src/query/server/server.go
```diff
@@ -230,77 +299,57 @@ func Run(runOpts RunOptions) {
...
-				zap.String("address", listenAddress), zap.Error(err))
+	cleanup := func() error {
+		lastErr := storageCleanup()
+		// Don't want to quit on the first error since the full cleanup is important
+		if err := clusters.Close(); err != nil {
+			if lastErr == nil {
+				lastErr = err
+			} else {
+				// Make sure the previous error is at least logged
+				logger.Error("error during cleanup", zap.Error(lastErr))
```
arnikola: You might want to do this regardless? Otherwise won't log the error if storageCleanup fails but clusters close correctly

src/query/server/server.go
```diff
@@ -313,38 +362,87 @@ func newDownsampler(
...
 	localStorage := local.NewStorage(clusters, workerPool)
 	stores := []storage.Storage{localStorage}
 	remoteEnabled := false
 	if cfg.RPC != nil && cfg.RPC.Enabled {
 		logger.Info("rpc enabled")
-		server := startGrpcServer(logger, localStorage, cfg.RPC)
-		cleanup = func() {
+		server, err := startGrpcServer(logger, localStorage, cfg.RPC)
+		if err != nil {
+			return nil, nil, err
```
arnikola: Worth logging?

src/query/server/server.go
```diff
@@ -313,38 +362,87 @@ func newDownsampler(
...
 		}
 
-		if remotes := cfg.RPC.RemoteListenAddresses; len(remotes) > 0 {
-			client, err := tsdbRemote.NewGrpcClient(remotes)
-			if err != nil {
-				logger.Fatal("unable to start remote clients for addresses", zap.Any("error", err))
-			}
+		remoteStorage, enabled, err := remoteClient(cfg)
+		if err != nil {
+			return nil, nil, err
```
arnikola: Worth logging?

src/query/server/server.go
```diff
@@ -114,17 +118,111 @@ func Run(runOpts RunOptions) {
...
+	}()
+
+	var (
+		backendStorage storage.Storage
+		clusterClient  clusterclient.Client
+		downsampler    downsample.Downsampler
+		enabled        bool
+	)
+
+	// For grpc backend, we need to setup only the grpc client and a storage accompanying that client.
```
benraskin92: NICE!

src/query/api/v1/httpd/handler.go
nikunjgit: we weren't using it so I figured better to remove

src/query/server/server.go
nikunjgit: yeah unfortunately it was initially structured as a monolith. I didn't want to spend time fixing it :(

src/query/server/server.go
nikunjgit: there can be multiple conditions within initCluster for error so there's not a constant error wrapping I can use.

src/query/server/server.go
nikunjgit: trying to avoid error/fatal logging within functions and only do that within Run or go routines.

src/query/server/server.go
nikunjgit: :)

PR #889, opened 2018-09-06
Eliminate locking from postings lists

prateek:
All the segments/readers do the necessary locking external to the postings list, so we don't need the locking inside the postings lists on top of that.

### DIFF
--- a/src/m3ninx/postings/roaring/roaring.go
+++ b/src/m3ninx/postings/roaring/roaring.go
@@ -22,7 +22,6 @@ package roaring
 
 import (
 	"errors"
-	"sync"
 
 	"github.com/m3db/m3/src/m3ninx/postings"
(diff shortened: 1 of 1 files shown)


src/m3ninx/postings/roaring/roaring.go
```diff
@@ -37,9 +36,8 @@ var (
 	errIteratorClosed        = errors.New("iterator has been closed")
 )
 
-// postingsList wraps a Roaring Bitmap with a mutex for thread safety.
+// postingsList wraps a Roaring Bitmap with the m3ninx pl API.
```
jeromefroe: worth calling out that it's not goroutine-safe?

prateek: Sure, will do in a follow up change.

PR #890, opened 2018-09-06
Add option and enable by default downsampling all metrics for an aggregated cluster namespace

robskillington:


src/query/server/server.go
```diff
@@ -319,6 +324,38 @@ func newDownsampler(
...
+	logger *zap.Logger,
+) []downsample.MappingRule {
+	var autoMappingRules []downsample.MappingRule
+	for _, namespace := range namespaces {
+		opts := namespace.Options()
+		attrs := opts.Attributes()
+		if attrs.MetricsType == storage.AggregatedMetricsType {
+			downsampleOpts, err := opts.DownsampleOptions()
+			if err != nil {
+				logger.Fatal("unable to resolve downsample options",
```
prateek: nit: could you bubble the error up and fatal in main/run

docs/integrations/prometheus.md
```diff
@@ -32,7 +32,8 @@ clusters:
 # We created a namespace called "default" and had set it to retention "48h".
        - namespace: default
          retention: 48h
-         storageMetricsType: unaggregated
+         resolution: 1m
```
prateek: maybe add a little blurb that this is optional and what it does

src/cmd/services/m3coordinator/downsample/downsampler.go
```diff
@@ -63,7 +63,8 @@ func NewDownsampler(
 
 func (d *downsampler) NewMetricsAppender() MetricsAppender {
 	return newMetricsAppender(metricsAppenderOptions{
-		agg:                     d.agg.aggregator,
+		agg: d.agg.aggregator,
```
prateek: weird that this didn't align, did `go fmt` break on this?

docs/integrations/prometheus.md
robskillington: Hm, this was an oopsy, I didn't mean to make this aggregated hah.

src/query/server/server.go
robskillington: Sure thing.

src/cmd/services/m3coordinator/downsample/downsampler.go
robskillington: Honestly, no idea.  `gofmt` was the tool that aligned it like this.  Let's just leave it for now since that how it wants to align it?

prateek: yep fair enough

docs/integrations/prometheus.md
```diff
@@ -32,7 +32,7 @@ clusters:
 # We created a namespace called "default" and had set it to retention "48h".
        - namespace: default
          retention: 48h
-         storageMetricsType: unaggregated
+         type: unaggregated
```
prateek: is this going to be b/w compatible? main concern is if skipping specifying this field be the same as `unaggregated`?

src/query/storage/local/config.go
```diff
@@ -66,10 +69,23 @@ func (c ClusterStaticConfiguration) newClient(
 // ClusterStaticNamespaceConfiguration describes the namespaces in a
 // static cluster.
 type ClusterStaticNamespaceConfiguration struct {
-	Namespace          string              `yaml:"namespace"`
-	StorageMetricsType storage.MetricsType `yaml:"storageMetricsType"`
-	Retention          time.Duration       `yaml:"retention" validate:"nonzero"`
-	Resolution         time.Duration       `yaml:"resolution" validate:"min=0"`
+	Namespace  string                                         `yaml:"namespace"`
+	Type       storage.MetricsType                            `yaml:"type"`
```
prateek: re: our conversation about deprecated, looks like there is a convention for this in Go already: https://go-review.googlesource.com/c/blog/+/18956/2/content/godoc-documenting-go-code.article

robskillington:
Did not mean to "dismiss stale review" and will land this based on @prateek sign off.

PR #894, opened 2018-09-10
Add support for putting commit log bootstrapper before peers

richardartoul:
- [X] Add a new "uninitialized" bootstrapper which can be added to the end of the bootstrapping process and will only succeed bootstraps for uninitialized topologies
- [X] Add logic to the commit log bootstrapper to only announce that it can succeed a bootstrap if the requested shard is marked as "Available" for itself in the topology
- [X] General refactoring for code sharing
- [X] Update example config files to use bootstrappers: filesystem,commitlog,peers,uninitialized

I also tested the following flows on the test cluster (with bootstrapping configuration filesystem,commitlog,peers,uninitialized):

- [X] Turn off all nodes, create new topology, create new namespace,  turn on all nodes and make sure they're able to start
- [X] Do a node remove and make sure bootstrap succeeds
- [X] Do a node add and make sure bootstrap succeeds
- [X] Roll the entire cluster and make sure each node is able to bootstrap

We will need to address #900 in a separate P.R

### DIFF
--- a/kube/bundle.yaml
+++ b/kube/bundle.yaml
@@ -177,7 +177,7 @@ data:
             - filesystem
             - commitlog
             - peers
-            - uninitialized
+            - uninitialized_topology
         fs:
             numProcessorsPerCPU: 0.125
 
--- a/kube/bundle.yaml
+++ b/kube/bundle.yaml
@@ -176,6 +176,7 @@ data:
         bootstrappers:
             - filesystem
             - commitlog
+            - peers
             - uninitialized
         fs:
             numProcessorsPerCPU: 0.125
--- a/kube/bundle.yaml
+++ b/kube/bundle.yaml
@@ -176,6 +176,7 @@ data:
         bootstrappers:
             - filesystem
             - commitlog
+            - uninitialized
         fs:
             numProcessorsPerCPU: 0.125
 
--- a/kube/m3dbnode-configmap.yaml
+++ b/kube/m3dbnode-configmap.yaml
@@ -68,6 +68,7 @@ data:
         bootstrappers:
             - filesystem
             - commitlog
+            - peers
             - uninitialized
         fs:
             numProcessorsPerCPU: 0.125
--- a/kube/m3dbnode-configmap.yaml
+++ b/kube/m3dbnode-configmap.yaml
@@ -69,7 +69,7 @@ data:
             - filesystem
             - commitlog
             - peers
-            - uninitialized
+            - uninitialized_topology
         fs:
             numProcessorsPerCPU: 0.125
 
--- a/kube/m3dbnode-configmap.yaml
+++ b/kube/m3dbnode-configmap.yaml
@@ -68,6 +68,7 @@ data:
         bootstrappers:
             - filesystem
             - commitlog
+            - uninitialized
         fs:
             numProcessorsPerCPU: 0.125
 
--- a/src/cmd/services/m3dbnode/config/bootstrap.go
+++ b/src/cmd/services/m3dbnode/config/bootstrap.go
@@ -33,6 +33,7 @@ import (
 	"github.com/m3db/m3/src/dbnode/storage/bootstrap/bootstrapper/commitlog"
 	bfs "github.com/m3db/m3/src/dbnode/storage/bootstrap/bootstrapper/fs"
 	"github.com/m3db/m3/src/dbnode/storage/bootstrap/bootstrapper/peers"
+	"github.com/m3db/m3/src/dbnode/storage/bootstrap/bootstrapper/uninitialized"
 	"github.com/m3db/m3/src/dbnode/storage/bootstrap/result"
 	"github.com/m3db/m3/src/dbnode/storage/index"
 )
@@ -120,7 +121,7 @@ func (bsc BootstrapConfiguration) New(
 		case bootstrapper.NoOpNoneBootstrapperName:
 			bs = bootstrapper.NewNoOpNoneBootstrapperProvider()
 		case bfs.FileSystemBootstrapperName:
-			fsbopts := bfs.NewOptions().
+			fsbOpts := bfs.NewOptions().
 				SetInstrumentOptions(opts.InstrumentOptions()).
 				SetResultOptions(rsOpts).
 				SetFilesystemOptions(fsOpts).
@@ -129,12 +130,12 @@ func (bsc BootstrapConfiguration) New(
 				SetDatabaseBlockRetrieverManager(opts.DatabaseBlockRetrieverManager()).
 				SetRuntimeOptionsManager(opts.RuntimeOptionsManager()).
 				SetIdentifierPool(opts.IdentifierPool())
-			bs, err = bfs.NewFileSystemBootstrapperProvider(fsbopts, bs)
+			bs, err = bfs.NewFileSystemBootstrapperProvider(fsbOpts, bs)
 			if err != nil {
 				return nil, err
 			}
 		case commitlog.CommitLogBootstrapperName:
-			copts := commitlog.NewOptions().
+			cOpts := commitlog.NewOptions().
 				SetResultOptions(rsOpts).
 				SetCommitLogOptions(opts.CommitLogOptions())
 
@@ -142,19 +143,27 @@ func (bsc BootstrapConfiguration) New(
 			if err != nil {
 				return nil, err
 			}
-			bs, err = commitlog.NewCommitLogBootstrapperProvider(copts, inspection, bs)
+			bs, err = commitlog.NewCommitLogBootstrapperProvider(cOpts, inspection, bs)
 			if err != nil {
 				return nil, err
 			}
 		case peers.PeersBootstrapperName:
-			popts := peers.NewOptions().
+			pOpts := peers.NewOptions().
 				SetResultOptions(rsOpts).
 				SetAdminClient(adminClient).
 				SetPersistManager(opts.PersistManager()).
 				SetDatabaseBlockRetrieverManager(opts.DatabaseBlockRetrieverManager()).
 				SetFetchBlocksMetadataEndpointVersion(bsc.peersFetchBlocksMetadataEndpointVersion()).
 				SetRuntimeOptionsManager(opts.RuntimeOptionsManager())
-			bs, err = peers.NewPeersBootstrapperProvider(popts, bs)
+			bs, err = peers.NewPeersBootstrapperProvider(pOpts, bs)
+			if err != nil {
+				return nil, err
+			}
+		case uninitialized.UninitializedBootstrapperName:
+			uopts := uninitialized.NewOptions().
+				SetResultOptions(rsOpts).
+				SetInstrumentOptions(opts.InstrumentOptions())
+			bs, err = uninitialized.NewUninitializedBootstrapperProvider(uopts, bs)
 			if err != nil {
 				return nil, err
 			}
@@ -193,6 +202,12 @@ func ValidateBootstrappersOrder(names []string) error {
 		commitlog.CommitLogBootstrapperName: []string{
 			// Commit log bootstrapper may appear after filesystem or peers
 			bfs.FileSystemBootstrapperName,
+			peers.PeersBootstrapperName,
+		},
+		uninitialized.UninitializedBootstrapperName: []string{
+			// Unintialized bootstrapper may appear after filesystem or peers or commitlog
+			bfs.FileSystemBootstrapperName,
+			commitlog.CommitLogBootstrapperName,
 			peers.PeersBootstrapperName,
 		},
 	}
--- a/src/cmd/services/m3dbnode/config/bootstrap.go
+++ b/src/cmd/services/m3dbnode/config/bootstrap.go
@@ -163,10 +163,7 @@ func (bsc BootstrapConfiguration) New(
 			uopts := uninitialized.NewOptions().
 				SetResultOptions(rsOpts).
 				SetInstrumentOptions(opts.InstrumentOptions())
-			bs, err = uninitialized.NewuninitializedTopologyBootstrapperProvider(uopts, bs)
-			if err != nil {
-				return nil, err
-			}
+			bs = uninitialized.NewuninitializedTopologyBootstrapperProvider(uopts, bs)
 		default:
 			return nil, fmt.Errorf("unknown bootstrapper: %s", bsc.Bootstrappers[i])
 		}
--- a/src/cmd/services/m3dbnode/config/bootstrap.go
+++ b/src/cmd/services/m3dbnode/config/bootstrap.go
@@ -198,6 +198,8 @@ func ValidateBootstrappersOrder(names []string) error {
 		peers.PeersBootstrapperName: []string{
 			// Peers must always appear after filesystem
 			bfs.FileSystemBootstrapperName,
+			// Peers may appear before OR after commitlog
+			commitlog.CommitLogBootstrapperName,
 		},
 		commitlog.CommitLogBootstrapperName: []string{
 			// Commit log bootstrapper may appear after filesystem or peers
--- a/src/cmd/services/m3dbnode/config/bootstrap.go
+++ b/src/cmd/services/m3dbnode/config/bootstrap.go
@@ -159,11 +159,11 @@ func (bsc BootstrapConfiguration) New(
 			if err != nil {
 				return nil, err
 			}
-		case uninitialized.UninitializedBootstrapperName:
+		case uninitialized.UninitializedTopologyBootstrapperName:
 			uopts := uninitialized.NewOptions().
 				SetResultOptions(rsOpts).
 				SetInstrumentOptions(opts.InstrumentOptions())
-			bs, err = uninitialized.NewUninitializedBootstrapperProvider(uopts, bs)
+			bs, err = uninitialized.NewuninitializedTopologyBootstrapperProvider(uopts, bs)
 			if err != nil {
 				return nil, err
 			}
@@ -206,7 +206,7 @@ func ValidateBootstrappersOrder(names []string) error {
 			bfs.FileSystemBootstrapperName,
 			peers.PeersBootstrapperName,
 		},
-		uninitialized.UninitializedBootstrapperName: []string{
+		uninitialized.UninitializedTopologyBootstrapperName: []string{
 			// Unintialized bootstrapper may appear after filesystem or peers or commitlog
 			bfs.FileSystemBootstrapperName,
 			commitlog.CommitLogBootstrapperName,
--- a/src/cmd/services/m3dbnode/main/common_test.go
+++ b/src/cmd/services/m3dbnode/main/common_test.go
@@ -34,8 +34,10 @@ import (
 	"time"
 
 	"github.com/gogo/protobuf/proto"
+	"github.com/m3db/m3/src/dbnode/client"
 	"github.com/m3db/m3/src/dbnode/retention"
 	"github.com/m3db/m3/src/dbnode/storage/namespace"
+	"github.com/m3db/m3cluster/shard"
 	"github.com/m3db/m3x/ident"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
@@ -138,4 +140,36 @@ func newNamespaceWithIndexProtoValue(id string, indexEnabled bool) (proto.Messag
 		return nil, err
 	}
 	return namespace.ToProto(nsMap), nil
+}
+
+// waitUntilAllShardsAreAvailable continually polls the session checking to see if the topology.Map
+// that the session is currently storing contains a non-zero number of host shard sets, and if so,
+// makes sure that all their shard states are Available.
+func waitUntilAllShardsAreAvailable(t *testing.T, session client.AdminSession) {
+outer:
+	for {
+		time.Sleep(10 * time.Millisecond)
+
+		topoMap, err := session.TopologyMap()
+		require.NoError(t, err)
+
+		var (
+			hostShardSets = topoMap.HostShardSets()
+		)
+
+		if len(hostShardSets) == 0 {
+			// We haven't received an actual topology yet.
+			continue
+		}
+
+		for _, hostShardSet := range hostShardSets {
+			for _, hostShard := range hostShardSet.ShardSet().All() {
+				if hostShard.State() != shard.Available {
+					continue outer
+				}
+			}
+		}
+
+		break
+	}
 }
--- a/src/dbnode/config/m3dbnode-local-etcd.yml
+++ b/src/dbnode/config/m3dbnode-local-etcd.yml
@@ -63,7 +63,7 @@ db:
     bootstrappers:
         - filesystem
         - commitlog
-        - noop-none
+        - uninitialized
     fs:
         numProcessorsPerCPU: 0.125
 
--- a/src/dbnode/config/m3dbnode-local-etcd.yml
+++ b/src/dbnode/config/m3dbnode-local-etcd.yml
@@ -64,7 +64,7 @@ db:
         - filesystem
         - commitlog
         - peers
-        - uninitialized
+        - uninitialized_topology
     fs:
         numProcessorsPerCPU: 0.125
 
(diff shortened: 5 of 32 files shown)


kube/bundle.yaml
```diff
@@ -176,6 +176,7 @@ data:
         bootstrappers:
             - filesystem
             - commitlog
+            - uninitialized
```
robskillington: Ah much better name 👍

src/cmd/services/m3dbnode/config/bootstrap.go
```diff
@@ -195,6 +204,12 @@ func ValidateBootstrappersOrder(names []string) error {
 			bfs.FileSystemBootstrapperName,
 			peers.PeersBootstrapperName,
 		},
+		uninitialized.UninitializedBootstrapperName: []string{
```
robskillington: Looks good.  For peers bootstrapper, shouldn't the commit log bootstrapper name also appear in there (so it can appear after the commit log bootstrapper as well)?

kube/bundle.yaml
robskillington: Do we want to be putting peers after `commitlog` in all these configurations?

src/dbnode/config/m3dbnode-local-etcd.yml
```diff
@@ -63,7 +63,7 @@ db:
     bootstrappers:
         - filesystem
         - commitlog
-        - noop-none
+        - uninitialized
```
robskillington: Hm as before, do we want to be putting peers after `commitlog` in all these configurations?

Also very surprised that we don't have peers actually listed here by default.

kube/bundle.yaml
richardartoul: haha I thought you wouldn't like the name but I couldn't think of anything better, and yeah sure why not. It should all work now

src/cmd/services/m3dbnode/config/bootstrap.go
richardartoul: good call, will fix

src/dbnode/config/m3dbnode-local-etcd.yml
richardartoul: yeah sure thing. Probably because it was too confusing to explain it to people haha

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -1425,6 +1421,70 @@ func (s commitLogSource) maybeAddToIndex(
...
+				"Initial topology state does not contain shard state for origin node and shard: %d", shardIDUint)
+			continue
+		}
+
+		originShardState := originHostShardState.ShardState
+		switch originShardState {
+		// In the Initializing and Unknown states we have to assume that the commit log
+		// is missing data and can't satisfy the bootstrap request.
+		case shard.Initializing:
+		case shard.Unknown:
```
robskillington: Should `shard.Unknown` be removed here perhaps and let this case drop through to the `default` case?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -1425,6 +1421,70 @@ func (s commitLogSource) maybeAddToIndex(
...
+		}
+
+		originHostShardState, ok := hostShardStates[topology.HostID(topoState.Origin.ID())]
+		if !ok {
+			// TODO(rartoul): Make this a hard error once we refactor the interface to support
+			// returning errors.
+			iOpts := s.opts.CommitLogOptions().InstrumentOptions()
+			invariantLogger := instrument.EmitInvariantViolationAndGetLogger(iOpts)
+			invariantLogger.Errorf(
+				"Initial topology state does not contain shard state for origin node and shard: %d", shardIDUint)
```
robskillington: nit: Lowercase the first letter of the log message for consistency?

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
```diff
@@ -0,0 +1,156 @@
...
+		var (
+			numInitializing = 0
+			numLeaving      = 0
+		)
+		for _, hostState := range hostShardStates {
+			shardState := hostState.ShardState
+			switch shardState {
+			case shard.Initializing:
+				numInitializing++
+			case shard.Unknown:
```
robskillington: I don't think we should track Unknown either, it should never be a part of a topology.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -1425,6 +1421,70 @@ func (s commitLogSource) maybeAddToIndex(
...
+			// long as the FS bootstrapper precedes the commit log bootstrapper.
+			// TODO(rartoul): Once we make changes to the bootstrapping interfaces
+			// to distinguish between "unfulfilled" data and "corrupt" data, then
+			// modify this to only say the commit log bootstrapper can fullfil
+			// "unfulfilled" data, but not corrupt data.
+			availableShardTimeRanges[shardIDUint] = shardsTimeRanges[shardIDUint]
+		default:
+			// TODO(rartoul): Make this a hard error once we refactor the interface to support
+			// returning errors.
+			panic(fmt.Sprintf("unknown shard state: %v", originShardState))
```
robskillington: Can we prioritize and make these `AvailableXyz(...)` methods return an error?  It would be really catastrophic if all the nodes panic when something does something stupid and modifies topology data in etcd in an unexpected way.

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
```diff
@@ -0,0 +1,156 @@
...
+				// failing the bootstrap). Once this interface supports returning errors,
+				// we should return an error in this case because the cluster is in a bad
+				// state and the operator should make a decision on how they want to proceed.
+			case shard.Leaving:
+				numLeaving++
+			case shard.Available:
+			default:
+				// TODO(rartoul): Make this a hard error once we refactor the interface to support
+				// returning errors.
+				panic(fmt.Sprintf("unknown shard state: %v", shardState))
```
robskillington: As per last comment: 
> Can we prioritize and make these `AvailableXyz(...)` methods return an error?  It would be really catastrophic if all the nodes panic when something does something stupid and modifies topology data in etcd in an unexpected way.

src/dbnode/topology/testutil/topology.go
```diff
@@ -28,6 +28,11 @@ import (
 	"github.com/m3db/m3cluster/shard"
 )
 
+const (
+	// SelfID is the string used to represent the ID of the origin node.
+	SelfID = "self"
```
robskillington: Is there a reason why we created this concept, instead of just pass around the original topology map and the identifier of the local node?

robskillington: Oh this is just a test util package, please ignore my comment.

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return availableShardTimeRanges
+}
+
+func (s *uninitializedSource) ReadData(
+	ns namespace.Metadata,
+	shardsTimeRanges result.ShardTimeRanges,
+	runOpts bootstrap.RunOptions,
+) (result.DataBootstrapResult, error) {
+	return result.NewDataBootstrapResult(), nil
```
robskillington: Hm I think you need to return results only for shards and time ranges that you can no?  The bootstrapper behavior is sometimes to ask for more than what was declared as available, if a previous bootstrapper isn't actually able to bootstrap that range.

src/dbnode/storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -722,6 +722,8 @@ func (s *peersSource) peerAvailability(
 			case shard.Available:
 				shardPeers.numAvailablePeers++
 			default:
+				// TODO(rartoul): Make this a hard error once we refactor the interface to support
+				// returning errors.
```
robskillington: Hm can we avoid this being a panic too?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Dont want to make the interface change in this P.R because it will double the size at the least, but I'll convert these to error logs and make the method return that it cant satisfy any ranges

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: yeah thats fair

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
richardartoul: I'll do the same fallthrough thing

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/peers/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
richardartoul: Not following this comment, I'm just returning an empty result here

richardartoul: nevermind I think I understand now

kube/bundle.yaml
```diff
@@ -176,6 +176,8 @@ data:
         bootstrappers:
             - filesystem
             - commitlog
+            - peers
+            - uninitialized
```
prateek: could we call this something else? like `mark_uninitialized_success` -- or some such?

richardartoul: Spoke offline and settled on uninitialized_topology

src/cmd/services/m3dbnode/main/common_test.go
```diff
@@ -139,3 +141,35 @@ func newNamespaceWithIndexProtoValue(id string, indexEnabled bool) (proto.Messag
 	}
 	return namespace.ToProto(nsMap), nil
 }
+
+// waitUntilAllShardsAreAvailable continually polls the session checking to see if the topology.Map
+// that the session is currently storing contains a non-zero number of host shard sets, and if so,
+// makes sure that all their shard states are Available.
+func waitUntilAllShardsAreAvailable(t *testing.T, session client.AdminSession) {
+outer:
```
prateek: should there be a timeout here too ?

richardartoul: Eh, the tests will already timeout by default. Also I'm just moving this function, not adding it

src/dbnode/storage/bootstrap/result/shard_ranges.go
```diff
@@ -99,16 +99,24 @@ func (r ShardTimeRanges) AddRanges(other ShardTimeRanges) {
 	}
 }
 
-// ToUnfulfilledResult will return a result that is comprised of wholly
+// ToUnfulfilledDataResult will return a result that is comprised of wholly
 // unfufilled time ranges from the set of shard time ranges.
-func (r ShardTimeRanges) ToUnfulfilledResult() DataBootstrapResult {
+func (r ShardTimeRanges) ToUnfulfilledDataResult() DataBootstrapResult {
```
prateek: +1

src/dbnode/topology/testutil/topology.go
```diff
@@ -101,3 +106,41 @@ func (v TopologyView) Map() (topology.Map, error) {
 
 	return topology.NewStaticMap(opts), nil
 }
+
+// SourceAvailableHost is a human-friendly way of constructing
```
prateek: suggestion(take/leave): i constructed topologies using something like --

```
func TestFetchTaggedResultsAccumulatorAnyResponseShouldTerminateConsistencyLevelOneSimpleTopo(t *testing.T) {
	// rf=3, 30 shards total; three identical hosts
	topoMap := tu.MustNewTopologyMap(3, map[string][]shard.Shard{
		"testhost0": tu.ShardsRange(0, 29, shard.Available),
		"testhost1": tu.ShardsRange(0, 29, shard.Available),
		"testhost2": tu.ShardsRange(0, 29, shard.Available),
	})
```

I think something equivalent `tu.MustNewStateSnapshot(...)` might read better when using than this method.

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/types.go
```diff
@@ -0,0 +1,44 @@
...
+	ResultOptions() result.Options
+
+	// Return the instrument options.
+	InstrumentOptions() instrument.Options
+
+	// Set the instrument options.
+	SetInstrumentOptions(value instrument.Options) Options
+
+	// Validate validates the options are correct.
+	Validate() error
```
prateek: nit: convention to have this as the first method in the Options() type.

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/types.go
```diff
@@ -0,0 +1,44 @@
...
+	SetResultOptions(value result.Options) Options
+
+	// ResultOptions returns the result options
+	ResultOptions() result.Options
+
+	// Return the instrument options.
+	InstrumentOptions() instrument.Options
+
+	// Set the instrument options.
+	SetInstrumentOptions(value instrument.Options) Options
```
prateek: nit: follow ordering of setter before getter for each type.

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source_test.go
```diff
@@ -0,0 +1,295 @@
...
+package uninitialized
+
+import (
+	"testing"
+	"time"
+
+	"github.com/m3db/m3/src/dbnode/storage/bootstrap"
+	"github.com/m3db/m3/src/dbnode/storage/bootstrap/result"
+	"github.com/m3db/m3/src/dbnode/storage/namespace"
+	topotestutils "github.com/m3db/m3/src/dbnode/topology/testutil"
```
prateek: suggestion: `tu` instead of `topotestutils`

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
```diff
@@ -0,0 +1,172 @@
...
+	"github.com/m3db/m3/src/dbnode/storage/bootstrap"
+	"github.com/m3db/m3/src/dbnode/storage/bootstrap/result"
+	"github.com/m3db/m3/src/dbnode/storage/namespace"
+	"github.com/m3db/m3/src/dbnode/topology"
+	"github.com/m3db/m3cluster/shard"
+)
+
+// The purpose of the unitializedSource is to succeed bootstraps for any
+// shard/time-ranges if the given shard/namespace combination has never
+// been completely initialized (is a new namespace). This is required for
```
prateek: nit: it's not per namespace, it's per cluster

richardartoul: you're right, good call

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
```diff
@@ -40,7 +40,9 @@ import (
 	"github.com/m3db/m3/src/dbnode/persist/fs/commitlog"
 	"github.com/m3db/m3/src/dbnode/storage/bootstrap/result"
 	"github.com/m3db/m3/src/dbnode/storage/namespace"
+	topotestutils "github.com/m3db/m3/src/dbnode/topology/testutil"
```
prateek: same as other spot, `tu` would probably be more concise

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -1425,6 +1421,72 @@ func (s commitLogSource) maybeAddToIndex(
 	return err
 }
 
+func (s *commitLogSource) availability(
```
prateek: nit: mind adding 1-2 lines explaining intent behind the method -- i.e. that anything not initialized is by definition not available

src/dbnode/storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -710,20 +710,22 @@ func (s *peersSource) peerAvailability(
...
 			case shard.Leaving:
 				fallthrough
 			case shard.Available:
 				shardPeers.numAvailablePeers++
+			case shard.Unknown:
+				fallthrough
 			default:
-				panic(
-					fmt.Sprintf("encountered unknown shard state: %s", shardState.String()))
+				// TODO(rartoul): Make this a hard error once we refactor the interface to support
```
prateek: +1

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/options.go
```diff
@@ -0,0 +1,75 @@
...
+)
+
+type options struct {
+	resultOpts result.Options
+	iOpts      instrument.Options
+}
+
+// NewOptions creates a new Options.
+func NewOptions() Options {
+	return &options{}
```
prateek: hm why not initialise these to `result.NewOptions()` and `instrument.NewOptions()`?

asking because almost always `NewOptions()` returns something that's valid by default

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
```diff
@@ -0,0 +1,172 @@
...
+		shardID := topology.ShardID(shardIDUint)
+		hostShardStates, ok := topoState.ShardStates[shardID]
+		if !ok {
+			// This shard was not part of the topology when the bootstrapping
+			// process began.
+			continue
+		}
+
+		// The basic idea for the algorithm is that on a shard-by-shard basis we
+		// need to determine if the namespace is "new" in the sense that it has
```
prateek: nit: instead of `namespace is "new"` --> `cluster is "new"`

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
```diff
@@ -0,0 +1,172 @@
...
+		// never been completely initialized (reached a state where all the hosts
+		// in the topology are "available").
+		// In order to determine this, we simply count the number of hosts in the
+		// "initializing" state. If this number is larger than zero, than the
+		// namespace is "new".
```
prateek: nit: cluster

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/options.go
richardartoul: To avoid issues like we've had in the past where we cant find metrics because (somewhere) in the deep configuration pipeline / setup we forgot to set InstrumentOptions()

This seems like a reasonable compromise to make sure that we don't run into that type of stuff anymore.

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
```diff
@@ -0,0 +1,172 @@
...
+				fallthrough
+			default:
+				// TODO(rartoul): Make this a hard error once we refactor the interface to support
+				// returning errors.
+				s.opts.InstrumentOptions().Logger().Errorf("unknown shard state: %v", shardState)
+				return result.ShardTimeRanges{}
+			}
+		}
+
+		shardHasNeverBeenCompletelyInitialized := numInitializing-numLeaving > 0
```
prateek: hm won't this break for a replace?

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
richardartoul: Thanks, not sure why I kept making that mistake. I guess its easy to think of the topology as being per-namespace but actually shards are for the whole cluster not a namespace

src/dbnode/topology/testutil/topology.go
richardartoul: Yeah this looks good, at this point I'd be re-writing a bunch of tests for a fairly superficial change (tests other than the ones I'm modifying in this P.R) so I'm gonna leave for now

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/types.go
richardartoul: sure thing

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/types.go
richardartoul: will fix

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
richardartoul: fixed

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/source.go
richardartoul: discussed offline, should be fine for everything except a replication factor change. Will add a comment

src/dbnode/topology/testutil/topology.go
prateek: More than the superficial, it's about keeping the tests simpler to read/maintain. If you wanna do it in another PR, mind opening an issue

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/options.go
prateek: Don't necessarily disagree but doing it in one place like this breaks from the convention everywhere else in the code.

kube/m3dbnode-configmap.yaml
```diff
@@ -68,6 +68,8 @@ data:
         bootstrappers:
             - filesystem
             - commitlog
+            - peers
+            - uninitialized_topology
```
prateek: i hate it, but i hate all others more =P

src/cmd/services/m3dbnode/main/common_test.go
prateek: yeah fair enough.

src/dbnode/topology/testutil/topology.go
richardartoul: #909

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source_prop_test.go
richardartoul: done

src/dbnode/storage/bootstrap/bootstrapper/uninitialized/options.go
```diff
@@ -0,0 +1,66 @@
...
+	errResultOptsNotSet     = errors.New("result options not set")
+)
+
+type options struct {
+	resultOpts result.Options
+	iOpts      instrument.Options
+}
+
+// NewOptions creates a new Options.
+func NewOptions() Options {
```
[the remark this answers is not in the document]
robskillington: Seems like the base bootstrapper uses it for some things in the provider.

PR #897, opened 2018-09-11
Add irate

benraskin92:


src/query/functions/temporal/rate_test.go
```diff
@@ -0,0 +1,137 @@
...
+			_, exists = bNode.cache.get(bounds.Start)
+			assert.False(t, exists, "block removed from cache")
+			blks, err := bNode.cache.multiGet(bounds.Previous(2), 3, false)
+			require.NoError(t, err)
+			assert.Len(t, blks, 0)
+		})
+	}
+}
+
+// func TestUnknownAggregation(t *testing.T) {
```
arnikola: nit delete

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,101 @@
...
+}
+
+func instantValue(values []float64, isRate bool, stepSize time.Duration) float64 {
+	fmt.Println(values)
+	valuesLen := len(values)
+	if valuesLen < 2 {
+		return math.NaN()
+	}
+
+	// {0, 1, 2, 3, 4},
```
arnikola: delete

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/m3db/m3/src/query/executor/transform"
+)
+
+const (
+	// IRateTemporalType calculates the per-second instant rate of increase of the time series
```
nikunjgit: fix comment ? we don't have instant/ range vector concept.

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	lastSample := values[nonNanIdx]
+	nonNanIdx = findNonNanIdx(values, nonNanIdx-1)
+	if nonNanIdx == -1 {
+		return math.NaN()
+	}
+	previousSample := values[nonNanIdx]
+
+	var resultValue float64
+	if isRate && lastSample < previousSample {
```
nikunjgit: i think you need reset for both types ?

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+
+// findNonNanIdx iterates over the values backwards until we find a non-NaN value,
+// then returns its index
+func findNonNanIdx(vals []float64, startingIdx int) int {
+	for i := startingIdx; i >= 0; i-- {
+		if !math.IsNaN(vals[i]) {
+			return i
+		}
+	}
+	return -1
```
arnikola: nit newline

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	nonNanIdx := valuesLen - 1
+	// find idx for last non-NaN value
+	nonNanIdx = findNonNanIdx(values, nonNanIdx)
+	// if nonNanIdx is 0 then you only have one value and should return a NaN
+	if nonNanIdx < 1 {
+		return math.NaN()
+	}
+	lastSample := values[nonNanIdx]
```
arnikola: nit newline

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+	nonNanIdx = findNonNanIdx(values, nonNanIdx-1)
+	if nonNanIdx == -1 {
+		return math.NaN()
+	}
+	previousSample := values[nonNanIdx]
```
arnikola: nit newline

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+	var resultValue float64
+	if isRate && lastSample < previousSample {
+		// Counter reset.
+		resultValue = lastSample
+	} else {
+		resultValue = lastSample - previousSample
+	}
+
+	if isRate {
+		resultValue /= float64(stepSize) / math.Pow10(9)
```
arnikola: Two things here;
1. I think this is wrong; you should be multiplying stepSize by the difference between indices for lastSample and previousSample

2. It's a bit confusing to do `x /= y / z`, better to write this as 
`resultValue *= float64(time.Second)`
`resultValue /= float64(stepSize * (indexLast - indexPrevious))`

src/query/functions/temporal/rate.go
benraskin92: Not according to Prom logic

src/query/functions/temporal/rate.go
benraskin92: Think this is fine :)

src/query/functions/temporal/rate_test.go
```diff
@@ -0,0 +1,237 @@
...
+			_, exists = bNode.cache.get(bounds.Start)
+			assert.False(t, exists, "block removed from cache")
+			blks, err := bNode.cache.multiGet(bounds.Previous(2), 3, false)
+			require.NoError(t, err)
+			assert.Len(t, blks, 0)
+		})
+	}
+}
+
+func TestFindNonNanIdx(t *testing.T) {
```
arnikola: Either add test or delete this function

src/query/functions/temporal/rate.go
arnikola: I thought we were add newlines after each closing bracket with lines following it

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,125 @@
...
+	return -1
+}
+
+func instantValue(values []float64, isRate bool, stepSize time.Duration) float64 {
+	valuesLen := len(values)
+	if valuesLen < 2 {
+		return math.NaN()
+	}
+
+	var indexLast int
```
arnikola: Don't need this, just initialize the var on line 102

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,125 @@
...
+
+	nonNanIdx := valuesLen - 1
+	// find idx for last non-NaN value
+	nonNanIdx = findNonNanIdx(values, nonNanIdx)
```
arnikola: You can do` indexLast := findNonNanIdx(values, nonNanIdx) `instead

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,125 @@
...
+	// if nonNanIdx is 0 then you only have one value and should return a NaN
+	if nonNanIdx < 1 {
+		return math.NaN()
+	}
+	indexLast = nonNanIdx
```
arnikola: nit: newline

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,125 @@
...
+	for i := startingIdx; i >= 0; i-- {
+		if !math.IsNaN(vals[i]) {
+			return i
+		}
+	}
+
+	return -1
+}
+
+func instantValue(values []float64, isRate bool, stepSize time.Duration) float64 {
```
arnikola: This function isn't necessary, you can move the logic into the `Process()` function instead

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+import (
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/m3db/m3/src/query/executor/transform"
+)
+
+const (
+	// IRateTemporalType calculates the per-second rate of increase of the time series
```
arnikola: IRateType is fine since it's in temporal, otherwise you have `temporal.IRateTemporal`

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+	// across the specified time range. This is based on the last two data points.
+	IRateTemporalType = "irate"
+
+	// IDeltaTemporalType calculates the difference between the last two values in the time series.
+	// IDeltaTemporalType should only be used with gauges.
+	IDeltaTemporalType = "idelta"
```
arnikola: same comment here

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+func NewRateOp(args []interface{}, optype string) (transform.Params, error) {
+	if optype == IRateTemporalType || optype == IDeltaTemporalType {
+		return newBaseOp(args, optype, newRateNode, nil)
+	}
+
+	return nil, fmt.Errorf("unknown rate type: %s", optype)
+}
+
+func newRateNode(op baseOp, controller *transform.Controller, opts transform.Options) Processor {
+	var isRate bool
```
arnikola: isRate := op.operatorType == IRateTemporalType

src/query/functions/temporal/rate.go
```diff
@@ -0,0 +1,119 @@
...
+
+	nonNanIdx := valuesLen - 1
+	// find idx for last non-NaN value
+	indexLast := findNonNanIdx(values, nonNanIdx)
+	// if indexLast is 0 then you only have one value and should return a NaN
+	if indexLast < 1 {
+		return math.NaN()
+	}
+
+	lastSample := values[indexLast]
```
arnikola: nit: since you keep track of this index, you can define `lastSample` lower down, closer to where you use it

PR #899, opened 2018-09-11
Fix broken gopkg glide pulling

prateek:


glide.yaml
```diff
@@ -150,6 +151,7 @@ import:
 
 - package: gopkg.in/alecthomas/kingpin.v2
   version: ^2.2.6
+  repo:    https://github.com/alecthomas/kingpin.git
```
robskillington: nit: Want to make it single space between yaml key and value like the rest?

prateek: sure thing.

PR #901, opened 2018-09-12
adding weight default for an instance

nerd0:


nerd0:
> Do you not want to use this default weight somewhere? 
Wanted to have that constant for use within the m3db-operator.

PR #902, opened 2018-09-12
Make FST/Map segment regexp behaviour consistent

prateek:
- Make regexp behaviour consistent between the two segment implementations


src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	var (
+		compiledRegex = CompiledRegex{}
+	)
+	simpleRE, err := re.Compile(simpleRe.String())
+	if err != nil {
+		return compiledRegex, err
+	}
+	compiledRegex.Simple = simpleRE
```
robskillington: nit: Since the previous early return statements have `return CompiledRegex{}, err...` perhaps change this to:
```go
	simpleRE, err := re.Compile(simpleRe.String())
	if err != nil {
		return CompiledRegex{}, err
	}

	compiledRegex := CompiledRegex{Simple: simpleRE}
```

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,315 @@
...
+		}
+		return parsed, false, nil
+	case syntax.OpEndText:
+		if rightmost {
+			return &syntax.Regexp{
+				Op:    syntax.OpEmptyMatch,
+				Flags: parsed.Flags,
+			}, true, nil
+		}
+		return parsed, false, nil
```
robskillington: Hm it's strange that some of these are early returns and others alter the vars and let the bottom of the switch case return.   Could we consolidate on one?

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+func ensureRegexpUnanchoredHelper(parsed *syntax.Regexp, leftmost, rightmost bool) (output *syntax.Regexp, changed bool, err error) {
+	changed = false
+	switch parsed.Op {
+	case syntax.OpBeginLine, syntax.OpEndLine:
+		// i.e. the flags provided to syntax.Parse did not include the `OneLine` flag, which
+		// should never happen as we're using syntax.Perl which does include it (ensured by a test
+		// in this package).
+		return nil, false, fmt.Errorf("regular expressions are forced to be single line")
```
robskillington: Great(!)

src/m3ninx/index/regexp.go
prateek: sure thing

src/m3ninx/index/regexp.go
prateek: sure thing

prateek: done

src/m3ninx/index/regexp.go
prateek: done

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,315 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return simplify(r), nil
+}
+
+func ensureRegexpUnanchoredHelper(parsed *syntax.Regexp, leftmost, rightmost bool) (output *syntax.Regexp, changed bool, err error) {
+	newRegexp := parsed
+	changed = false
+	switch parsed.Op {
```
richardartoul: Presumably this switch statement needs to handle every type of A.S.T node that we could encounter. Is it worth adding a default that returns an error so that regex's we didn't anticipate properly at least 500 or something instead of having incorrect behavior

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+			newRe, c, err := ensureRegexpUnanchoredHelper(parsed.Sub[0], leftmost, rightmost)
+			if err != nil {
+				return nil, false, err
+			}
+			if c {
+				parsed.Sub[0] = newRe
+				changed = true
+			}
+		}
+	case syntax.OpStar:
```
robskillington: So both question and star have just two sub expressions?

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+	case syntax.OpAlternate:
+		// strip left-most '^' and right-most '$' in each sub-expression
+		for idx := range parsed.Sub {
+			newRe, c, err := ensureRegexpUnanchoredHelper(parsed.Sub[idx], leftmost, rightmost)
+			if err != nil {
+				return nil, false, err
+			}
+			parsed.Sub[idx] = newRe
+			changed = changed || c
```
robskillington: super nit: Since the other cases do `if c { parsed.Sub[...] = ...; changed = true }` perhaps makes sense to follow that pattern here too?

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+				Flags: parsed.Flags,
+				Sub: []*syntax.Regexp{
+					newRe,
+					original,
+				},
+			}
+			changed = true
+		}
+	}
+	return newRegexp, changed, nil
```
robskillington: Would it make more sense to just return in each sub switch case expression and just `return input, false, nil` if we get to the bottom?  Seems none of them need any code after the case statements.

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+				Flags: parsed.Flags,
+				Sub: []*syntax.Regexp{
+					&syntax.Regexp{
+						Op:    syntax.OpQuest,
+						Flags: parsed.Flags,
+						Sub: []*syntax.Regexp{
+							newRe,
+						},
+					},
+					original,
```
robskillington: Hm, what's the reasoning for constructing a preceding "question"/"optional" sub-expression and concating with "star"/"any" sub-expression?  Also could the original not still contain a sub-expression with "^" or "$"?

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+					&syntax.Regexp{
+						Op:    syntax.OpStar,
+						Flags: parsed.Flags,
+						Sub: []*syntax.Regexp{
+							original.Sub[0],
+						},
+					},
+				},
+			}, true, nil
+		}
```
robskillington: Hm similar for "plus"/"at-least-one" why are we concatenating a star query with a star query for the original?  

I see how it translates, but I would have assumed this would just be `if c { parsed.Sub[0] = newRe; changed = true }`?

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	copied := *ast
+	copied.Sub = make([]*syntax.Regexp, 0, len(ast.Sub))
+	for _, r := range ast.Sub {
+		copied.Sub = append(copied.Sub, deepCopy(r))
+	}
+	if len(copied.Sub0) != 0 && copied.Sub0[0] != nil {
+		copied.Sub0[0] = deepCopy(copied.Sub0[0])
+	}
+	return &copied
```
robskillington: Do we need copy `ast.Rune` slice?  It is apparently a `Rune []rune` meaning it could be a heap allocated slice depending on if it outgrows `ast.Rune0` array.

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+var emptyStringRes = []*syntax.Regexp{
+	&syntax.Regexp{Op: syntax.OpEmptyMatch},
+	&syntax.Regexp{Op: syntax.OpQuest, Sub: []*syntax.Regexp{&syntax.Regexp{Op: syntax.OpEmptyMatch}}},
+	&syntax.Regexp{Op: syntax.OpPlus, Sub: []*syntax.Regexp{&syntax.Regexp{Op: syntax.OpEmptyMatch}}},
+	&syntax.Regexp{Op: syntax.OpStar, Sub: []*syntax.Regexp{&syntax.Regexp{Op: syntax.OpEmptyMatch}}},
```
robskillington: Would OpRepeat also apply here?

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+}
+
+func matchesEmptyString(ast *syntax.Regexp) bool {
+	if ast == nil {
+		return false
+	}
+	for _, empty := range emptyStringRes {
+		if ast.Equal(empty) {
```
robskillington: Hm I wonder if actually doing specific checks of the op type and then whether the subexpressions match OpEmptyMatch is a good idea?  i.e. if OpRepeat is also a similar case, then you don't want to compare the Min/Max fields of the regexp against your literal emptyStringRes OpRepeat regexp.

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+		return CompiledRegex{}, fmt.Errorf("unable to create map re: %v", err)
+	}
+
+	simpleRE, err := re.Compile(simpleRe.String())
+	if err != nil {
+		return CompiledRegex{}, err
+	}
+	compiledRegex := CompiledRegex{Simple: simpleRE}
+
+	fstRE, start, end, err := fstregexp.ParsedRegexp(reString, vellumRe)
```
robskillington: Hm, shouldn't this be `vellumRe.String()` for the first arg?  Since really the expression you are passing is the unanchored version to vellum?  If the input actually does contain the anchoring characters, and were subsequently stripped, then would they not still be contained when you pass it here to construct the fst regexp?

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	compiledRegex.FST = fstRE
+	compiledRegex.PrefixBegin = start
+	compiledRegex.PrefixEnd = end
+
+	return compiledRegex, nil
+}
+
+func parseRegexp(re string) (*syntax.Regexp, error) {
+	return syntax.Parse(re, syntax.Perl)
```
robskillington: Hm shouldn't this be `syntax.Perl&syntax.OneLine`?  Also any reason tests don't use a call to this to test the one line behavior?

src/m3ninx/index/regexp_test.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+		testCase{
+			name:           "simple case of both ^ & $",
+			input:          "^abc$",
+			expectedOutput: "abc",
+		},
+		testCase{
+			name:           "weird case of internal ^",
+			input:          "^a^bc$",
+			expectedOutput: "a\\Abc",
```
robskillington: Hm, what's `"\\A"` out of interest?

Also, wish that they exported `dumpRegexp` from stdlib so you could use that to test... you could copy paste if you want to use that kind of output to test against (e.g. `"cat{lit{foo},lit{bar}}"`):
https://golang.org/src/regexp/syntax/parse_test.go#L315

src/m3ninx/index/regexp_test.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+		testCase{
+			name:           "question concat case $",
+			input:          "abc$?",
+			expectedOutput: "abc",
+		},
+		testCase{
+			name:           "star mark case both boundaries success",
+			input:          "(?:^abc$)*",
+			expectedOutput: "(?:abc)?(?:\\Aabc(?-m:$))*",
```
robskillington: Hm yeah, what's expected I guess here.  I wonder if `"(?:abc)?"` is more correct translation since we only allow end of input to appear once?

src/m3ninx/index/regexp_test.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+		testCase{
+			name:           "star mark case only ^",
+			input:          "(?:^abc)*",
+			expectedOutput: "(?:abc)?(?:\\Aabc)*",
```
robskillington: Can't this just be equivalent to `"(?:abc)*"`?

src/m3ninx/index/regexp_test.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+		testCase{
+			name:           "star mark case only $",
+			input:          "(?:abc$)*",
+			expectedOutput: "(?:abc)?(?:abc(?-m:$))*",
```
robskillington: This feels like it could also be translated to `"(?:abc)?"` since we allow end of input to only appear once.

src/m3ninx/index/regexp_test.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+		testCase{
+			name:           "star concat case $",
+			input:          "abc$*",
+			expectedOutput: "abc(?-m:$)*",
```
robskillington: I see, we're not returning error for these cases?

src/m3ninx/index/regexp_test.go
```diff
@@ -0,0 +1,284 @@
...
+			input:          "(?:^abc$){3,4}",
+			expectedOutput: "abc(?:\\Aabc(?-m:$)){2,3}",
+		},
+	}
+	for _, tc := range testCases {
+		re, err := parseRegexp(tc.input)
+		require.NoError(t, err)
+		parsed, err := ensureRegexpUnanchored(re)
+		require.NoError(t, err)
+		assert.Equal(t, tc.expectedOutput, parsed.String(), tc.name)
```
robskillington: nit: Want to make this:
```go
t.Run(tc.name, func(t *testing.T) {
  // test...
})
```

So that we get the test names for each one and show which ones pass/fail individually?

src/m3ninx/index/regexp_test.go
```diff
@@ -0,0 +1,284 @@
...
+			input:          "(?:abc){3,4}",
+			expectedOutput: "\\Aabcabcabc(?:abc)?\\z",
+		},
+	}
+	for _, tc := range testCases {
+		re, err := parseRegexp(tc.input)
+		require.NoError(t, err)
+		parsed, err := ensureRegexpAnchored(re)
+		require.NoError(t, err)
+		assert.Equal(t, tc.expectedOutput, parsed.String(), tc.name)
```
robskillington: nit: `t.Run(tc.name, func(t *testing.T) { ... })` here too?

src/m3ninx/index/segment/fst/regexp/regexp.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	prefix := LiteralPrefix(parsed)
+	if prefix != "" {
+		prefixBeg := []byte(prefix)
+		prefixEnd := IncrementBytes(prefixBeg)
+		return re, prefixBeg, prefixEnd, nil
+	}
+
+	return re, nil, nil, nil
```
robskillington: Is this a valid case when you don't have start/end?

src/m3ninx/index/segment/fst/regexp/regexp.go
```diff
@@ -0,0 +1,89 @@
...
+}
+
+// LiteralPrefix returns the literal prefix given the parse tree for a regexp
+func LiteralPrefix(s *syntax.Regexp) string {
+	// traverse the left-most branch in the parse tree as long as the
+	// node represents a concatenation
+	for s != nil && s.Op == syntax.OpConcat {
```
robskillington: Hm what about the case when you run into a `syntax.OpCharClass`?  Is that also a good enough operator to work out prefix begin/end?

src/m3ninx/index/segment/fst/segment.go
```diff
@@ -330,7 +320,7 @@ func (r *fsSegment) MatchRegexp(field []byte, regexp []byte, compiled index.Comp
 	var (
 		fstCloser     = x.NewSafeCloser(termsFST)
 		pl            = r.opts.PostingsListPool().Get()
-		iter, iterErr = termsFST.Search(re, minByteKey, nil)
+		iter, iterErr = termsFST.Search(re, compiled.PrefixBegin, compiled.PrefixEnd)
```
robskillington: If prefixBegin and prefixEnd are both nil (as can happen with call to `LiteralPrefix(...)`) should `minByteKey` still be specified here?

src/m3ninx/index/regexp.go
prateek: So all the following ops only have a single sub expression - 

```
			syntax.OpCapture,
			syntax.OpStar,
			syntax.OpPlus,
			syntax.OpQuest,
			syntax.OpRepeat
```

src/m3ninx/index/regexp.go
prateek: sure

src/m3ninx/index/regexp.go
prateek: Sure thing, that'll read better

src/m3ninx/index/regexp.go
prateek: talked offline - have to do this to preserve the Language captured by the regular expression post transformations.

prateek: re: original - we make the deep copy to preserve any internal "^" or "$".

src/m3ninx/index/regexp.go
prateek: same as above - we do the concat with original to ensure we preserve the lang of the regexp

src/m3ninx/index/segment/fst/regexp/regexp.go
prateek: yep, it'll scan the whole FST if start/end aren't specified

src/m3ninx/index/segment/fst/regexp/regexp.go
prateek: So `OpCharClass` would give us a range of prefixes, i.e. `[ACy]` would tell us to scan `[A,B)`, `[C,D)` and `[y,z)`. Could us this in the future to reduce the total scan space in the FST but I don't want to make the change without benchmarking impact

prateek: I'm sure we can come up with better heuristics than this one but it's a good start for now

src/m3ninx/index/segment/fst/segment.go
prateek: So it doesn't need to be, `nil` will cause more scans than `minByteKey`

Yep! Hopefully people aren't doing `.*`, but beyond that it'll help a lot for any reasonable query. For the '.*' case, i'm going to add more data to the segment (postings list per unique field) sometime down the road.

src/m3ninx/index/regexp.go
prateek: yeah i re-wrote this recursively to effectively do the following:``` matchesEmptyString(ast): return true if {ast is OpEmptyMatch } or {ast is a unary op'd (?/*/+/repeat) of a ast that matchesEmptyString}```

src/m3ninx/index/regexp.go
prateek: only reason we don't is cause we never modify the runes, i'll add a comment to clarify

prateek: done

src/m3ninx/index/regexp_test.go
prateek: re: `dumpRegexp()` solid idea, will do.

prateek: done

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,366 @@
...
+	switch ast.Op {
+	case syntax.OpConcat:
+		// a concatenation of a single sub-expression is the same as the sub-expression itself
+		if len(ast.Sub) == 1 {
+			return ast.Sub[0], true
+		}
+
+		changed := false
+		// check if we have any concats of concats, if so, we can pull the ones below this level up
+		subs := make([]*syntax.Regexp, 0, len(ast.Sub))
```
richardartoul: maybe an over-optimization, but you could probably re-use the original slice with a little bit of shifting and currPointer logic

richardartoul: Actually I looked it up and apparently there is just an easy way to do it

```
b := a[:0]
for _, x := range a {
	if f(x) {
		b = append(b, x)
	}
}
```

prateek: Yeah fair enough. This works when filtering like this spot, but not when you're expanding the original slice. I'll make the change for now

prateek: Turns out this works in the expansion case too -- https://play.golang.org/p/KtEkWbx4NHn/ Looks like the `range a` bounds for the loop are evaluated in a way to be change agnostic.

prateek: talked offline: going to leave as is.

src/m3ninx/index/segment/fst/regexp/regexp.go
robskillington: Sounds good.

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,366 @@
...
+						Flags: parsed.Flags,
+						Sub: []*syntax.Regexp{
+							original.Sub[0],
+						},
+					},
+				},
+			}, true, nil
+		}
+	case syntax.OpRepeat:
+		if len(parsed.Sub) > 0 && parsed.Min > 0 && parsed.Max > 0 {
```
robskillington: Still not sure about this case, don't you just want `parsed.Min > 0`?  When `parsed.Max == -1` which means unlimited range (i.e. `{2,}`) you still want to do this no?

robskillington: This is the case I'm thinking of:
https://play.golang.org/p/O0uAkMV-pEr

src/m3ninx/index/regexp.go
```diff
@@ -0,0 +1,366 @@
...
+		if len(ast.Sub) == 1 {
+			return ast.Sub[0], true
+		}
+
+		changed := false
+		// check if we have any concats of concats, if so, we can pull the ones below this level up
+		subs := make([]*syntax.Regexp, 0, len(ast.Sub))
+		for _, sub := range ast.Sub {
+			if sub.Op == syntax.OpConcat {
+				subs = append(subs, sub.Sub...)
```
robskillington: Hm you don't have to take into account each sub expression too?

robskillington: Oh you do this later, I see.

robskillington: Hm but shouldn't you check if each sub expression matches empty string itself (like you do below for the non concat case)?

src/m3ninx/index/segment/fst/regexp/regexp_test.go
```diff
@@ -0,0 +1,63 @@
...
+	for i, test := range tests {
+		s, err := syntax.Parse(test.input, syntax.Perl)
+		if err != nil {
+			t.Fatalf("expected no syntax.Parse error, got: %v", err)
+		}
+
+		got := LiteralPrefix(s)
+		if test.expected != got {
+			t.Fatalf("test: %d, %+v, got: %s", i, test, got)
+		}
```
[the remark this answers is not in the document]
prateek: sure

PR #904, opened 2018-09-12
Arnikola/update rpc

arnikola:


src/query/generated/proto/rpc/query.proto
```diff
@@ -1,9 +1,10 @@
 syntax = "proto3";
 
-package rpcpb;
+package rpc;
```
benraskin92: Yeah so the reason why I updated the name of this is so `make test-all-gen` would pass. Prateek was working on a fix I think? cc @prateek

arnikola: Yeah noticed it was failing :\ unfortunately this needs to match the package name on the other side as far as I can tell

src/query/api/v1/handler/prometheus/remote/test/remote/read.go
```diff
@@ -0,0 +1,67 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
nikunjgit: this hierarchy seems to be a bit too much. Why not just put this within  src/query/test ? Also, maybe have to move write.go from there as well.

src/query/server/query_pools.go
```diff
@@ -0,0 +1,194 @@
...
+func (s sessionPools) TagDecoder() serialize.TagDecoderPool {
+	return s.tagDecoder
+}
+
+// Build iterator pools if they are unavailable from m3db (e.g. if running standalone query)
+func buildIteratorPools() encoding.IteratorPools {
+	// TODO: add capabilities to get this from configs
+	// TODO2: add instrumentation options to these pools
+	replicas := 1
+	iteratorPoolSize := 65536
```
nikunjgit: why not just move these up into constants ?

src/query/storage/interface.go
```diff
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/m3db/m3/src/dbnode/encoding"
```
nikunjgit: we shouldn't have dbnode specific stuff in our storage interface. Eventually, I was planning to move out local storage into stores/m3db as well.

src/query/storage/interface.go
```diff
@@ -86,11 +87,31 @@ type FetchOptions struct {
...
-		ctx context.Context, query *FetchQuery, options *FetchOptions) (*FetchResult, error)
-	FetchTags(
-		ctx context.Context, query *FetchQuery, options *FetchOptions) (*SearchResults, error)
+		ctx context.Context,
+		query *FetchQuery,
+		options *FetchOptions,
+	) (*FetchResult, error)
+
+	// FetchRaw returns the raw iterators from the underlying storage
+	FetchRaw(
```
nikunjgit: We shouldn't expose this as part of Querier. This couples us very strongly with dbnode. Shouldn't the local storage FetchBlock internally use FetchRaw ? Not sure why Engine needs to know about this.

src/query/api/v1/handler/prometheus/remote/test/remote/read.go
arnikola: Was just following the `.../.../write.go` approach, will change

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -75,7 +75,7 @@ func NewPromReadHandler(engine *executor.Engine) http.Handler {
 }
 
 func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
-	ctx := r.Context()
+	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
```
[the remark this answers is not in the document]
arnikola: Once we decide on what headers we want for ownership we can add validation/selection to them? I don't think anything will use them, unless I'm mistaken and they get added to all our log messages

src/query/server/server.go
```diff
@@ -409,19 +387,24 @@ func newDownsamplerAutoMappingRules(
 	return autoMappingRules, nil
 }
 
-func initClusters(cfg config.Configuration, dbClientCh <-chan client.Client, logger *zap.Logger) (local.Clusters, error) {
+func initClusters(
```
nikunjgit: isn't it better to initialize worker pools outside and pass them to this func ?

src/query/storage/converter.go
```diff
@@ -182,17 +187,15 @@ func SeriesToPromSamples(series *ts.Series) []*prompb.Sample {
...
-	initRawFetchAllocSize = 32
-)
-
 func iteratorToTsSeries(
 	iter encoding.SeriesIterator,
 	namespace ident.ID,
 ) (*ts.Series, error) {
 	if namespace == nil {
 		namespace = iter.Namespace()
+		if namespace == nil {
```
nikunjgit: don't want to fail request here ?

src/query/storage/fanout/storage.go
```diff
@@ -42,11 +42,19 @@ type fanoutStorage struct {
 }
 
 // NewStorage creates a new fanout Storage instance.
```
nikunjgit: avoid having files in diffs with just ordering changes. This makes the diff long/complicated for no reason.

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -0,0 +1,115 @@
...
+	// duplciate series iterators.
+	seenIters []encoding.SeriesIterators
+}
+
+type multiFetchResultSeries struct {
+	idx   int
+	attrs storage.Attributes
+}
+
+func (r *multiFetchResult) cleanup() error {
```
nikunjgit: close ?

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -0,0 +1,115 @@
...
+	attrs storage.Attributes,
+	iterators encoding.SeriesIterators,
+	err error,
+) {
+	r.Lock()
+	defer r.Unlock()
+	r.seenIters = append(r.seenIters, iterators)
+
+	if err != nil {
+		r.err = r.err.Add(err)
```
nikunjgit: plural ?

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -0,0 +1,115 @@
...
+		return
+	}
+
+	if r.iterators == nil {
```
nikunjgit: can't you just use len(r.seenIters) ?

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -0,0 +1,115 @@
...
+		r.iterators = iterators
+		r.dedupeFirstAttrs = attrs
+		return
+	}
+
+	iters := iterators.Iters()
+	// Need to dedupe
+	if r.dedupeMap == nil {
```
nikunjgit: you can also move this up into where you check len(r.seenIters)

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -0,0 +1,115 @@
...
+		r.dedupeMap = make(map[string]multiFetchResultSeries, len(iters))
+		for idx, s := range iters {
+			r.dedupeMap[s.ID().String()] = multiFetchResultSeries{
```
nikunjgit: just keep the key as bytes ? then you don't need these conversions

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -0,0 +1,115 @@
...
+				idx:   idx,
+				attrs: r.dedupeFirstAttrs,
+			}
+		}
+	}
+
+	for _, s := range iters {
```
nikunjgit: move this logic into its own func

src/query/tsdb/remote/client.go
```diff
@@ -22,120 +22,163 @@ package remote
...
-func (c *grpcClient) Write(ctx context.Context, query *storage.WriteQuery) error {
-	client := c.client
-
-	writeClient, err := client.Write(ctx)
+func (c *grpcClient) FetchBlocks(
+	ctx context.Context,
+	query *storage.FetchQuery,
+	options *storage.FetchOptions,
+) (block.Result, error) {
+	iters, err := c.fetchRaw(ctx, query, options)
```
nikunjgit: do the iterators here contain actual block or do we still do the hack later ? If its the later then maybe add a TODO ?

src/query/server/server.go
arnikola: This is for iterator pools, not the worker pools. If we're running embedded, we share the pools from m3db, so we need to pull them out after creating our session object here; if we're standalone, we spin up our own pools instead.

src/query/storage/converter.go
arnikola: Namespace isn't really used for anything as far as I know, so it's not particularly important; can get rid of it without any issues.

src/query/storage/fanout/storage.go
arnikola: There were changes to this previously with FetchRaw which got reverted

src/query/storage/m3/multi_fetch_result.go
arnikola: The key's an `ident.ID()` so a conversion of some sort is still necessary

src/query/storage/m3/multi_fetch_result.go
arnikola: What's the benefit vs the nil check?

src/query/tsdb/remote/client.go
arnikola: These will just be the raw iterators sent from m3; will need to run block conversion logic, but that should go down the path that calls SeriesIteratorsToFetchResult and FetchResultToBlockResult

scripts/integration-tests/docker-integration-test.sh
```diff
@@ -44,7 +44,7 @@ curl -vvvsSf -X POST localhost:7201/api/v1/namespace -d '{
 
 echo "Sleep while namespace is init'd"
 
-sleep 10 # TODO Replace sleeps with logic to determine when to proceed
+sleep 60 # TODO Replace sleeps with logic to determine when to proceed
```
nikunjgit: revert

scripts/integration-tests/docker-integration-test.sh
```diff
@@ -44,7 +44,7 @@ curl -vvvsSf -X POST localhost:7201/api/v1/namespace -d '{
 
 echo "Sleep while namespace is init'd"
 
-sleep 60 # TODO Replace sleeps with logic to determine when to proceed
+sleep 10 # TODO Replace sleeps with logic to determine when to proceed
```
benraskin92: Why'd you change this?

arnikola: I'd changed it to 60 previously, reverting the change here

PR #906, opened 2018-09-14
Add changes and resets functions

benraskin92:


src/query/parser/promql/types.go
```diff
@@ -175,6 +175,9 @@ func NewFunctionExpr(name string, argValues []interface{}) (parser.Params, error
 	case temporal.IRateType, temporal.IDeltaType:
 		return temporal.NewRateOp(argValues, name)
 
+	case temporal.ResetsTemporalType, temporal.ChangesTemporalType:
```
benraskin92: Fix naming

src/query/parser/promql/parse_test.go
```diff
@@ -249,6 +249,8 @@ var temporalParseTests = []struct {
 	{"stdvar_over_time(up[5m])", temporal.StdVarTemporalType},
 	{"irate(up[5m])", temporal.IRateType},
 	{"idelta(up[5m])", temporal.IDeltaType},
+	{"resets(up[5m])", temporal.ResetsTemporalType},
```
benraskin92: Fix naming

src/query/functions/temporal/functions.go
```diff
@@ -0,0 +1,101 @@
...
+	// ResetsTemporalType should only be used with counters.
+	ResetsType = "resets"
+
+	// ChangesType returns the number of times a value changes within the provided time range for
+	// a given time series.
+	ChangesType = "changes"
+)
+
+var (
+	temporalFuncs = map[string]interface{}{
```
arnikola: This is unnecessary

src/query/functions/temporal/functions.go
```diff
@@ -0,0 +1,101 @@
...
+		ResetsType:  struct{}{},
+		ChangesType: struct{}{},
+	}
+)
+
+// NewFunctionOp creates a new base temporal transform for functions
+func NewFunctionOp(args []interface{}, optype string) (transform.Params, error) {
+	if _, ok := temporalFuncs[optype]; ok {
```
arnikola: Rather than a map lookup here, create the compFunc here and pass it into your function node

src/query/functions/temporal/functions.go
```diff
@@ -0,0 +1,101 @@
...
+				result++
+			}
+			prev = curr
+			continue
+		}
+		if !math.IsNaN(curr) {
+			prev = curr
+		}
+	}
+	return result
```
arnikola: nit newline

src/query/functions/temporal/functions.go
```diff
@@ -0,0 +1,101 @@
...
+
+	for _, curr := range values[1:] {
+		if !math.IsNaN(prev) && !math.IsNaN(curr) {
+			if f.comparisonFunc(curr, prev) {
+				result++
+			}
+			prev = curr
+			continue
+		}
+		if !math.IsNaN(curr) {
```
arnikola: nit newline

src/query/functions/temporal/functions.go
arnikola: Check for allNaN case here; should return NaN not 0

src/query/functions/temporal/functions.go
```diff
@@ -0,0 +1,101 @@
...
+	comparisonFunc comparisonFunc
+}
+
+func (f *functionNode) Process(values []float64) float64 {
+	if len(values) == 0 {
+		return math.NaN()
+	}
+
+	result := 0.0
+	prev := values[0]
```
arnikola: Might be easier to set prev to math.NaN first, then range through all values

src/query/functions/temporal/functions_test.go
```diff
@@ -0,0 +1,185 @@
...
+	"github.com/m3db/m3/src/query/executor/transform"
+	"github.com/m3db/m3/src/query/parser"
+	"github.com/m3db/m3/src/query/test"
+	"github.com/m3db/m3/src/query/test/executor"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var testTemporalCases = []testCase{
```
arnikola: These tests are hard to understand; can they include the input values in the test rather than it coming from somewhere else unrelated to here?

Also we should avoid using the same values for all tests; they should be customized to check the interesting/edge cases for the functions that they apply to

src/query/functions/temporal/functions.go
arnikola: Might read better if you started this loop with
```
if math.IsNaN(curr) {
 continue
}
```
Then continue knowing that curr != nan

src/query/functions/temporal/functions.go
benraskin92: I think all NaN should return 0

src/query/functions/temporal/functions.go
benraskin92: Hmm then we'd need to change the definition of `newBaseOp`, i think this way is cleaner

src/query/functions/temporal/functions.go
benraskin92: You mean just check like 
```
	if optype != ResetsType, etc. {
		return fmt.Errorf("unknown function type: %s", optype)
	}
```

src/query/functions/temporal/functions.go
arnikola: What does prom think?

src/query/functions/temporal/functions_test.go
benraskin92: I think once Nikunj fixes the alignment stuff, we can add more edge cases and clean up the tests.

src/query/functions/temporal/functions.go
benraskin92: Eh I think this is fine, plus its how Prom does it.

src/query/functions/temporal/functions_test.go
arnikola: The tests assume correct alignment since we can build our own test data :)

src/query/functions/temporal/functions.go
arnikola: image

Prom thinks it should be NaN

When mapping from prom's approach to ours, it's necessary to add NaNs where they can just return an empty vector

src/query/functions/temporal/functions.go
arnikola: These functions, aggregation, and irate all look very similar. Can you refactor them into a single path rather than have a bunch of different ops created instead? The only thing they differ in is what the function in `Process()` looks like

src/query/functions/temporal/functions.go
```diff
@@ -0,0 +1,105 @@
...
+	for _, curr := range values[1:] {
+		if math.IsNaN(curr) {
+			continue
+		}
+
+		if !math.IsNaN(prev) {
+			if f.comparisonFunc(curr, prev) {
+				result++
+			}
+			prev = curr
```
arnikola: Should probably delete this and the continue

src/query/functions/temporal/functions.go
benraskin92: I'll do this in a separate pr

PR #908, opened 2018-09-16
Add cluster namespace fanout heuristics supporting queries greater than retention

robskillington:


src/query/storage/local/storage.go
```diff
@@ -377,10 +510,29 @@ func (r *multiFetchResult) add(
...
+			case namespacesCoverPartialRetention:
+				// Already exists and either has longer retention, or the same retention
+				// and result we are adding is not as precise
+				longerRetention := existing.attrs.Retention > attrs.Retention
+				sameRetentionAndSameOrMoreGranularResolution :=
+					existing.attrs.Retention == attrs.Retention &&
+						existing.attrs.Resolution <= attrs.Resolution
+				existsBetter = longerRetention || sameRetentionAndSameOrMoreGranularResolution
+			default:
+				panic(fmt.Sprintf("unknown query fanout type: %d", r.queryFanoutType))
```
prateek: hm could you error instead of panic-ing here?

src/query/storage/local/storage.go
```diff
@@ -72,26 +227,19 @@ func (s *localStorage) Fetch(ctx context.Context, query *storage.FetchQuery, opt
 	// cluster that can completely fulfill this range and then prefer the
 	// highest resolution (most fine grained) results.
 	// This needs to be optimized, however this is a start.
+	fanout, namespaces, err := s.resolveClusterNamespacesForQuery(query.Start, query.End)
```
prateek: couple thoughts:
- maybe it'd be good to add a way for users to see what data source was hit for each series
- do you want to add tests for the various combinations - just ensuring we override to the right storage based on the heuristics in code

src/query/storage/m3/storage.go
```diff
@@ -51,8 +59,16 @@ type m3storage struct {
 
 // NewStorage creates a new local m3storage instance.
 // TODO: Consider combining readWorkerPool and writeWorkerPool
-func NewStorage(clusters Clusters, workerPool pool.ObjectPool, writeWorkerPool xsync.PooledWorkerPool) Storage {
-	return &m3storage{clusters: clusters, readWorkerPool: workerPool, writeWorkerPool: writeWorkerPool}
+func NewStorage(
```
prateek: i thought artem had removed the TODO above in another PR. mind rebasing on master once?

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -117,55 +143,79 @@ func (r *multiResult) Add(
 		return
 	}
 
+	if len(r.seenIters) == 0 {
+		// store the first attributes seen
+		r.seenFirstAttrs = attrs
+	}
 	r.seenIters = append(r.seenIters, iterators)
 	if !r.err.Empty() {
```
richardartoul: should this be moved up to the first thing after defering the unlock?

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -117,55 +143,79 @@ func (r *multiResult) Add(
...
 	}
 
-	for _, iter := range iters {
-		id := iter.ID().String()
-		r.dedupeMap[id] = append(r.dedupeMap[id], multiResultSeries{
-			attrs: attrs,
-			iter:  iter,
-		})
+	if len(r.seenIters) == 2 {
+		// need to build backfill the dedupe map from the first result first
```
richardartoul: maybe remove the word build

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -89,19 +117,17 @@ func (r *multiResult) FinalResult() (encoding.SeriesIterators, error) {
...
-		iter.SetAt(i, res[0].iter)
+		r.finalResult.SetAt(i, res.iter)
 		i++
 	}
 
-	return iter, nil
+	return r.finalResult, nil
 }
 
 func (r *multiResult) Add(
```
richardartoul: super nit: maybe call the iterators arg newIterators, I got a little lost between `seenIters` and `iterators` the first time

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -117,55 +143,79 @@ func (r *multiResult) Add(
...
 		}
 
-		maxSeries := serieses[maxIdx]
-
-		// reset slice to reuse
-		for idx := range serieses {
-			serieses[idx] = multiResultSeries{}
+		var existsBetter bool
+		switch r.fanout {
+		case namespacesCoverAllRetention:
```
richardartoul: is this more like `namespacesCoverQueryDuration`?

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -117,55 +143,79 @@ func (r *multiResult) Add(
...
+			// Already exists and resolution of result we are adding is not as precise
+			existsBetter = existing.attrs.Resolution <= attrs.Resolution
+		case namespacesCoverPartialRetention:
+			// Already exists and either has longer retention, or the same retention
+			// and result we are adding is not as precise
+			longerRetention := existing.attrs.Retention > attrs.Retention
```
richardartoul: nit: `existsLongerRetention`

src/query/storage/m3/storage.go
robskillington: Sure thing.

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -117,55 +143,79 @@ func (r *multiResult) Add(
...
+			continue
 		}
 
-		maxSeries := serieses[maxIdx]
-
-		// reset slice to reuse
-		for idx := range serieses {
-			serieses[idx] = multiResultSeries{}
+		var existsBetter bool
+		switch r.fanout {
```
richardartoul: nit: I feel like this logic would be easier to follow if it was framed from the perspective of determining whether the incoming attributes were better as opposed to vice versa, but that might just be personal preference

richardartoul:
One thing I'm realizing now is that there is no concept of how old a namespace is. So say you start with:

```
namespace1: 5 day retention, 1m resolution
```

Then you decide you want higher resolution so you start dual writing to:

```
namespace1: 5 day retention, 1m resolution
namespace2: 5 day retention, 30s resolution
```

Your queries will immediately begin going to namespace 2 even though it doesn't really have any data.

Fixing that is probably out of scope for this P.R but maybe we can track in an issue or something, although I'm not sure what the solution is other than configuring "cutoffs" or something

src/query/storage/m3/multi_fetch_result.go
robskillington: Sure thing.

src/query/storage/m3/multi_fetch_result.go
robskillington: Yeah I can rename this.  Probably to `namespaceCoversQueryRange`.

src/query/storage/m3/multi_fetch_result.go
robskillington: Sure thing.

robskillington:
@richardartoul yeah, I'm explicitly not tackling that.  

We can add a readable/writeable flag later to the namespaces - and then manually flip it on once the namespace is filled up.  

Then eventually we'll get to the all singing, all dancing, dynamic and automated version of this all.

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -63,23 +70,44 @@ type multiResultSeries struct {
...
+	r.Lock()
+	defer r.Unlock()
+
 	for _, iters := range r.seenIters {
 		iters.Close()
 	}
+	r.seenIters = nil
+
+	if r.finalResult != nil {
+		// NB(r): Since all the series iterators in the final result are held onto
```
prateek: Mind adding a unit test w/ mocks to ensure we free the underlying iterators exactly once assuming the API is used correctly. I think it's 100% right the way it is but the test will be good for when someone comes and attempts to refactors this

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -117,55 +143,84 @@ func (r *multiResult) Add(
...
-		id := iter.ID().String()
-		r.dedupeMap[id] = append(r.dedupeMap[id], multiResultSeries{
-			attrs: attrs,
-			iter:  iter,
-		})
+	if len(r.seenIters) == 2 {
+		// need to backfill the dedupe map from the first result first
+		existing := r.seenIters[0]
+		r.dedupeMap = make(map[string]multiResultSeries, existing.Len())
+		for _, iter := range existing.Iters() {
```
prateek: nit: could you refactor this loop into `func (r *multiResult) addOrUpdateMap(attrs, iters)` - would be re-usable here and in `dedupe()`

src/query/storage/m3/storage.go
prateek: sigh must have been skipped in the last review, mind deleting it please.

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -117,55 +143,84 @@ func (r *multiResult) Add(
...
-			serieses[idx] = multiResultSeries{}
+		var existsBetter bool
+		switch r.fanout {
+		case namespaceCoversAllQueryRange:
+			// Already exists and resolution of result we are adding is not as precise
+			existsBetter = existing.attrs.Resolution <= attrs.Resolution
+		case namespaceCoversPartialQueryRange:
+			// Already exists and either has longer retention, or the same retention
+			// and result we are adding is not as precise
+			existsLongerRetention := existing.attrs.Retention > attrs.Retention
```
prateek: All this is to avoid consolidation... :sigh:

src/query/storage/m3/storage.go
```diff
@@ -361,3 +354,150 @@ func (s *m3storage) writeSingle(
...
+			continue
+		}
+
+		if filter != nil && !filter(namespace) {
+			continue
+		}
+
+		downsampleOpts, err := opts.DownsampleOptions()
+		if err != nil || !downsampleOpts.All {
+			// Cluster does not contain all data, include as part of fan out
```
richardartoul: Confused about the downsampling thing, how is this different than saying something is unaggregated?

src/query/storage/m3/storage.go
```diff
@@ -361,3 +354,150 @@ func (s *m3storage) writeSingle(
...
+// fanned out to depending on the query time and the namespaces configured.
+func (s *m3storage) resolveClusterNamespacesForQuery(
+	start time.Time,
+	end time.Time,
+) (queryFanoutType, ClusterNamespaces, error) {
+	now := time.Now()
+
+	unaggregated := s.clusters.UnaggregatedClusterNamespace()
+	unaggregatedRetention := unaggregated.Options().Attributes().Retention
+	unaggregatedStart := now.Add(-1 * unaggregatedRetention)
```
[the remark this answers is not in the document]
robskillington: Not here thankfully.

src/query/storage/m3/multi_fetch_result.go
robskillington: Sure thing.

src/query/storage/m3/storage.go
```diff
@@ -361,3 +354,150 @@ func (s *m3storage) writeSingle(
...
 		datapoint.Timestamp, datapoint.Value, query.Unit, query.Annotation)
 }
+
+// resolveClusterNamespacesForQuery returns the namespaces that need to be
+// fanned out to depending on the query time and the namespaces configured.
+func (s *m3storage) resolveClusterNamespacesForQuery(
+	start time.Time,
+	end time.Time,
+) (queryFanoutType, ClusterNamespaces, error) {
+	now := time.Now()
```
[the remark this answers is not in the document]
robskillington: Sure thing.

src/query/storage/m3/storage.go
```diff
@@ -361,3 +354,150 @@ func (s *m3storage) writeSingle(
...
+
+	unaggregated := s.clusters.UnaggregatedClusterNamespace()
+	unaggregatedRetention := unaggregated.Options().Attributes().Retention
+	unaggregatedStart := now.Add(-1 * unaggregatedRetention)
+	if !unaggregatedStart.After(start) {
```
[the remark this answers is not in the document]
robskillington: Hm, yeah I did it so you don't have to call twice.  It's not a hot path though so I'll refactor to this.

src/query/storage/m3/storage.go
```diff
@@ -361,3 +354,150 @@ func (s *m3storage) writeSingle(
...
+		// Highest resolution is unaggregated, return if it can fulfill it
+		return namespacesCoverAllRetention, ClusterNamespaces{unaggregated}, nil
+	}
+
+	// First determine if any aggregated clusters that can span the whole query
```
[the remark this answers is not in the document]
robskillington: Good catch, thanks.

src/query/storage/m3/multi_fetch_result.go
robskillington: As discussed, let's revisit this in a followup change.

src/query/storage/m3/multi_fetch_result.go
robskillington: Yup =]

src/query/storage/m3/storage.go
robskillington: Basically it's aggregated, but data will only go to this cluster if rules are setup to do so.

PR #917, opened 2018-09-19
Adding docker logs to docker int test when container fails to start

nerd0:


scripts/integration-tests/docker-integration-test.sh
```diff
@@ -10,7 +10,13 @@ docker build -t "m3query:$(git rev-parse HEAD)" -f ./docker/m3query/Dockerfile .
 
 echo "Run m3dbnode docker container"
 
-docker run --name "m3dbnode-version-$(git rev-parse HEAD)" -d --rm -p 9000:9000 -p 9001:9001 -p 9002:9002 -p 9003:9003 -p 9004:9004 -p 7201:7201 "m3dbnode:$(git rev-parse HEAD)"
+docker create --name "m3dbnode-version-$(git rev-parse HEAD)" -p 9000:9000 -p 9001:9001 -p 9002:9002 -p 9003:9003 -p 9004:9004 -p 7201:7201 "m3dbnode:$(git rev-parse HEAD)"
```
arnikola: Could htis cause issues if your container already exists?

nerd0: I believe that semaphore provides a blank slate VM to run the test suite within thus there shouldn't be any images on. Furthermore since a gitsha is used as part of the name there is unlikely going to be a collision on the name.

PR #921, opened 2018-09-20
Adding error handling logic for local runs of the scripts

nerd0:


scripts/integration-tests/docker-integration-test.sh
```diff
@@ -11,6 +11,13 @@ docker build -t "m3query:$(git rev-parse HEAD)" -f ./docker/m3query/Dockerfile .
 echo "Run m3dbnode docker container"
 
 docker create --name "m3dbnode-version-$(git rev-parse HEAD)" -p 9000:9000 -p 9001:9001 -p 9002:9002 -p 9003:9003 -p 9004:9004 -p 7201:7201 "m3dbnode:$(git rev-parse HEAD)"
+
+# Remove the container if people are running this locally
+if [ $? -ne 0 ]; then
```
arnikola: Not sure this works unfortunately, didn't cleanup properly after a fail when I tried it

richardartoul: Can you just use a trap instead

nerd0: I think this PR is horribly out of date given all the recent work with docker compose. Going to close.

PR #922, opened 2018-09-20
Add Rate, Delta, and Increase

benraskin92:


src/query/functions/temporal/rate.go
```diff
@@ -36,36 +36,154 @@ const (
...
+		optype == IncreaseType || optype == DeltaType {
 		return newBaseOp(args, optype, newRateNode, nil)
 	}
 
 	return nil, fmt.Errorf("unknown rate type: %s", optype)
 }
 
 func newRateNode(op baseOp, controller *transform.Controller, opts transform.Options) Processor {
-	isRate := op.operatorType == IRateType
+	isRate := op.operatorType == IRateType || op.operatorType == RateType
```
arnikola: Doing a bunch of comparisons here that you've already done; might be better to refactor to something like this
```golang
var isRate, isCounter bool
switch(op.operatorType)
case IRateType:
 isRate = true
 rateFunc = irateFunc
```

src/query/functions/temporal/rate.go
```diff
@@ -36,36 +36,154 @@ const (
...
+	isCounter := op.operatorType == RateType || op.operatorType == IncreaseType
+
+	var rateFunc rateFunc
+	if op.operatorType == IRateType || op.operatorType == IDeltaType {
+		rateFunc = irateFunc
+	}
+
+	if op.operatorType == RateType || op.operatorType == IncreaseType || op.operatorType == DeltaType {
+		rateFunc = standardRate
```
arnikola: nit: standardRateFunc?

src/query/functions/temporal/rate.go
```diff
@@ -36,36 +36,154 @@ const (
...
 type rateNode struct {
-	op         baseOp
-	controller *transform.Controller
-	timeSpec   transform.TimeSpec
-	isRate     bool
+	op                baseOp
+	controller        *transform.Controller
+	timeSpec          transform.TimeSpec
+	isRate, isCounter bool
+	rateFunc          rateFunc
```
arnikola: nit: rename to `fn`?

src/query/functions/temporal/rate.go
```diff
@@ -36,36 +36,154 @@ const (
...
+		if !foundFirst {
+			firstVal = val
+			firstIdx = i
+			foundFirst = true
+		}
+
+		if isCounter && val < lastValue {
+			counterCorrection += lastValue
+		}
+		lastValue = val
```
arnikola: nit newline

src/query/functions/temporal/rate_test.go
```diff
@@ -147,6 +147,38 @@ var testDeltaCases = []testRateCase{
 			{math.NaN(), math.NaN(), math.NaN(), math.NaN(), math.NaN()},
 		},
 	},
+	// {
```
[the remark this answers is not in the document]
benraskin92: Leaving it until Nikunj lands his diff and I can add tests ;)

src/query/functions/temporal/rate.go
```diff
@@ -36,36 +36,163 @@ const (
...
+		isRate = true
+	case RateType:
+		isRate = true
+		isCounter = true
+	case IncreaseType:
+		isCounter = true
+	}
+
+	var rateFn rateFn
+	if op.operatorType == IRateType || op.operatorType == IDeltaType {
```
arnikola: Pull this into the switch statement above

src/query/functions/temporal/rate.go
```diff
@@ -36,36 +36,163 @@ const (
...
-	isRate := op.operatorType == IRateType
+	var isRate, isCounter bool
+	switch op.operatorType {
+	case IRateType:
+		isRate = true
+	case RateType:
+		isRate = true
+		isCounter = true
+	case IncreaseType:
+		isCounter = true
```
arnikola: Might need to add a default case or is the linter fine with this?

src/m3nsch/agent/agent.go
```diff
@@ -25,11 +25,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/m3db/m3/src/dbnode/client"
```
arnikola: Still need to rebase, these seem to have snuck in from master

src/query/functions/temporal/rate.go
benraskin92: Semaphore is passing so I'm guessing its fine. It probably doesn't need a default cause of 
```
var isRate, isCounter bool
```

nikunjgit: as discussed, change MakeProcessor to be an interface. That way you can get rid of the aggregation param as well in baseOp

src/query/functions/temporal/rate.go
```diff
@@ -36,36 +36,167 @@ const (
...
 
-func newRateNode(op baseOp, controller *transform.Controller, opts transform.Options) Processor {
-	isRate := op.operatorType == IRateType
+func (r rateProcessor) Init(op baseOp, controller *transform.Controller, opts transform.Options) Processor {
+	var (
+		isRate, isCounter bool
+		rateFn            = standardRateFunc
+	)
+
+	switch op.operatorType {
```
nikunjgit: move this logic into NewRateOp. Init is called during actual execution, we need to make these decisions at the time of generating the plan.

src/query/functions/temporal/rate_test.go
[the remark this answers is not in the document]
nikunjgit: already landed my diff

src/query/functions/temporal/rate.go
```diff
@@ -36,36 +36,167 @@ const (
...
+	rateFn            rateFn
 }
 
 func (r *rateNode) Process(values []float64) float64 {
+	return r.rateFn(values, r.isRate, r.isCounter, r.timeSpec)
+}
+
+func standardRateFunc(values []float64, isRate, isCounter bool, timeSpec transform.TimeSpec) float64 {
+	var (
+		rangeStart = float64(timeSpec.Start.Unix()) - timeSpec.Step.Seconds()
```
nikunjgit: instead of converting all these into floats, lets keep them as time.Time or time.Duration as much as possible

src/query/functions/temporal/aggregation.go
```diff
@@ -65,23 +65,30 @@ var (
...
+}
+
 // NewAggOp creates a new base temporal transform with a specified node
 func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
 	if aggregationFunc, ok := aggFuncs[optype]; ok {
-		return newBaseOp(args, optype, newAggNode, aggregationFunc)
+		a := aggProcessor{
+			aggFunc: aggregationFunc,
+		}
+		return newBaseOp(args, optype, a)
```
arnikola: nit: newline

src/query/functions/temporal/rate.go
```diff
@@ -36,42 +37,180 @@ const (
...
+			firstTS = dp.Timestamp
+			firstIdx = i
+			foundFirst = true
+		}
+
+		if isCounter && dp.Value < lastValue {
+			counterCorrection += lastValue
+		}
+
+		lastValue = dp.Value
```
arnikola: nit: just take `lastIdx = i` here, then after the loop set `lastValue, lastTs := datapoints[i].Value, datapoints[i].Timestamp`

src/query/functions/temporal/rate.go
```diff
@@ -36,42 +37,180 @@ const (
...
+
+		if !foundFirst {
+			firstVal = dp.Value
+			firstTS = dp.Timestamp
+			firstIdx = i
+			foundFirst = true
+		}
+
+		if isCounter && dp.Value < lastValue {
+			counterCorrection += lastValue
```
arnikola: This will always hit when you're setting first. Is this the correct behaviour?

src/query/functions/temporal/rate.go
```diff
@@ -36,42 +37,180 @@ const (
...
+		lastIdx = i
+	}
+
+	if firstIdx == lastIdx {
+		return math.NaN()
+	}
+
+	resultValue := lastValue - firstVal + counterCorrection
+
+	rangeStart := timeSpec.Start.Add(-1 * (timeSpec.Step + timeWindow))
```
arnikola: nit: group `rangeStart` and `durationToStart` together, same for end

src/query/functions/temporal/rate.go
benraskin92: You need last value inside the loop though.

```
		if isCounter && dp.Value < lastValue {
			counterCorrection += lastValue
		}
```

src/query/functions/temporal/rate.go
benraskin92: This is what prom does so going to leave it.

PR #923, opened 2018-09-20
Only use signal channel if running embedded

arnikola:
Fixes an issue where m3 can hang if stopped during initialization

### DIFF
--- a/src/query/server/server.go
+++ b/src/query/server/server.go
@@ -203,9 +203,7 @@ func Run(runOpts RunOptions) {
 
 	var interruptErr error
 	if runOpts.DBConfig != nil {
-		select {
-		case interruptErr = <-interruptCh:
-		}
+		interruptErr = <-interruptCh
 	} else {
 		// Only use this if running standalone, as otherwise it will
 		// obfuscate signal channel for the db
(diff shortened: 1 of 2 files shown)


src/query/server/server.go
```diff
@@ -196,17 +196,25 @@ func Run(runOpts RunOptions) {
...
-	case <-sigChan:
-	case <-interruptCh:
+	if cfg.Backend == config.GRPCStorageType {
+		// Only use this if not running embedded, otherwise will
+		// obfuscate signal channel for m3db
+		sigChan := make(chan os.Signal, 1)
+		signal.Notify(sigChan, os.Interrupt, syscall.SIGTERM)
+
+		select {
+		case <-sigChan:
```
prateek: nit: log the cause of termination so you can track why the process crashed

```
var interruptErr error
select {
case sig := <- sigChan:
  interruptErr = fmt.Errorf("%v", sig)
case err  := <- interruptCh:
  interruptErr = err
}

log.Infof("interrupt :%v", interruptErr)

arnikola: Good call

prateek:
@arnikola looks like metalint is :(

arnikola:
Unfortunately a repro test would be super messy; better to add this as an integration test when we have those up and running :)

PR #926, opened 2018-09-20
Timeout topology watch and return better error messages for missing topology / namespaces

richardartoul:
We weren't passing a timeout properly to the ServiceClient so:

```golang
watch, err := services.Watch(opts.ServiceID(), opts.QueryOptions())
```

would hang forever if there was no topology.

In addition, in the case where there is no topology or namespace the error message is cryptic (timeout) when it could be slightly more helpful

### DIFF
--- a/src/dbnode/environment/config.go
+++ b/src/dbnode/environment/config.go
@@ -109,11 +109,10 @@ type ConfigureResults struct {
 
 // ConfigurationParameters are options used to create new ConfigureResults
 type ConfigurationParameters struct {
-	InstrumentOpts             instrument.Options
-	HashingSeed                uint32
-	HostID                     string
-	NamespaceResolutionTimeout time.Duration
-	TopologyResolutionTimeout  time.Duration
+	InstrumentOpts    instrument.Options
+	HashingSeed       uint32
+	HostID            string
+	ResolutionTimeout time.Duration
 }
 
 // Configure creates a new ConfigureResults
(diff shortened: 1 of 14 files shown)


src/dbnode/environment/config.go
```diff
@@ -109,11 +109,10 @@ type ConfigureResults struct {
...
 type ConfigurationParameters struct {
-	InstrumentOpts             instrument.Options
-	HashingSeed                uint32
-	HostID                     string
-	NamespaceResolutionTimeout time.Duration
-	TopologyResolutionTimeout  time.Duration
+	InstrumentOpts    instrument.Options
+	HashingSeed       uint32
+	HostID            string
+	ResolutionTimeout time.Duration
```
prateek: Need to be b/w compatible when making changes to publicly configurable fields. 
- Could you mark the old fields deprecated, 
- Add the new field,
- Error if both new and old are set, and pick the set field otherwise 

Rob had to do some of this in #890 -
 https://github.com/m3db/m3/blob/721ac9269d61247cdc8eb943f7b7b4bf9bc251c0/src/query/storage/local/config.go#L90-L117

richardartoul: spoke offline, very unlikely anyone ever set these fields are they are undocumented in are YMLs so gonna make the backward incompatible change

src/dbnode/storage/database.go
```diff
@@ -164,7 +164,7 @@ func NewDatabase(
 	nsInit := opts.NamespaceInitializer()
 	nsReg, err := nsInit.Init()
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("database: namespace initialized failed: %v. make sure a namespace has been set")
```
justinjc: Capitalize the new sentence. Also, include `err` in the `Errorf`.

robskillington:
Hey, do you want to make sure this supports the case if bringing up a node and using it's coordinator endpoint to create a placement?

Before we set an indefinite timeout here we used to basically crash the node before you could create a placement, which was a fairly big issue.

Does this change re-introduce that behavior?  (Just wanting to get to the bottom of how to best cover both cases)

prateek:
@robskillington Yep, we settled on removing the timeout all together for that reason. As long as we put appropriate log messages before waiting in all the places, it should fail the same as it does today. Sound reasonable?

src/dbnode/topology/dynamic.go
```diff
@@ -185,19 +184,6 @@ func (t *dynamicTopology) MarkShardsAvailable(
 	return ps.MarkShardsAvailable(instanceID, shardIDs...)
 }
 
-func waitOnInit(w services.Watch, d time.Duration) error {
```
cw9: We finally removed this, noice!

src/dbnode/environment/config.go
```diff
@@ -109,11 +98,9 @@ type ConfigureResults struct {
 
 // ConfigurationParameters are options used to create new ConfigureResults
 type ConfigurationParameters struct {
-	InstrumentOpts             instrument.Options
-	HashingSeed                uint32
-	HostID                     string
-	NamespaceResolutionTimeout time.Duration
```
prateek: can't just delete these (b/w compatibility and all), need to mark deprecated instead

richardartoul: Spoke offline and agreed its ok in this case because it was undocumented config and we're breaking the functionaltiy anyways

PR #928, opened 2018-09-21
Paramaterize Prometheus integration test to use for local testing

robskillington:


scripts/integration-tests/prometheus/prometheus-integration-test.sh
```diff
@@ -5,9 +5,16 @@ set -xe
 rm -rf /tmp/m3dbdata/
 mkdir -p /tmp/m3dbdata/
 
-echo "Build docker images"
+PARAM_TEST_BUILD="${TEST_BUILD:-true}"
```
nerd0: Are these injected by semaphore? Or just more meant for local testing?

scripts/integration-tests/prometheus/docker-compose.yml
```diff
@@ -38,5 +38,13 @@ services:
     image: prom/prometheus:latest
     volumes:
       - "./:/etc/prometheus/"
+  grafana:
```
nerd0: Just curious if grafana used within the test?

robskillington: It is not, we're going to be using this as a hack to run this locally heh.

scripts/integration-tests/prometheus/prometheus-integration-test.sh
robskillington: Local testing yeah.

PR #929, opened 2018-09-21
Add ability for query storage to provide unconsolidated blocks

nikunjgit:


src/query/block/column.go
```diff
@@ -36,6 +36,11 @@ type columnBlock struct {
 	seriesMeta []SeriesMeta
 }
 
+// Unconsolidated returns the unconlidated version for the block
```
benraskin92: `s/unconlidated/unconsolidated`

src/query/block/column.go
```diff
@@ -36,6 +36,11 @@ type columnBlock struct {
...
+func (c *columnBlock) Unconsolidated() (UnconsolidatedBlock, error) {
+	return nil, fmt.Errorf("unconslidated block not support for block, meta: %s", c.meta)
```
benraskin92: Reword error

src/query/block/scalar.go
```diff
@@ -47,6 +47,11 @@ func NewScalar(val float64, bounds models.Bounds) Block {
 	}
 }
 
+// Unconsolidated returns the unconlidated version for the block
```
benraskin92: Same comments as above

src/query/block/series.go
```diff
@@ -45,3 +49,39 @@ func (s Series) Values() []float64 {
...
+func (s UnconsolidatedSeries) Datapoints() []ts.Datapoints {
+	return s.datapoints
+}
+
+// Len returns the number of datapoints slices in the series
+func (s UnconsolidatedSeries) Len() int {
+	return len(s.datapoints)
+}
+
+// Consolidated conslidates the series
```
benraskin92: `s/conslidates/consolidates`

src/query/block/types.go
```diff
@@ -104,3 +152,18 @@ type Builder interface {
...
+func TakeLast(values ts.Datapoints) float64 {
+	if len(values) == 0 {
+		return math.NaN()
+	}
+
+	return values[len(values)-1].Value
+}
+
+// Type assertion
+var _ ConsolidationFunc = TakeLast
```
benraskin92: What is this needed for exactly?

src/query/executor/transform/lazy.go
```diff
@@ -145,6 +145,23 @@ type lazyBlock struct {
 	processedBlock block.Block
 }
 
+// Unconsolidated returns the unconlidated version for the block
```
benraskin92: `s/unconlidated/unconsolidated`

benraskin92: `s/for the block/of the block`

src/query/executor/transform/lazy.go
```diff
@@ -145,6 +145,23 @@ type lazyBlock struct {
...
+func (f *lazyBlock) Unconsolidated() (block.UnconsolidatedBlock, error) {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+
+	if f.processedBlock != nil {
+		return f.processedBlock.Unconsolidated()
+	}
+
+	err := f.process()
```
benraskin92:
```
if err := f.process(); err != nil {
   return nil, err
}
```

src/query/functions/temporal/base_test.go
```diff
@@ -38,9 +39,10 @@ import (
 type processor struct {
 }
 
-func (p *processor) Process(f []float64) float64 {
+func (p *processor) Process(dps ts.Datapoints) float64 {
+	vals := dps.Values()
 	sum := 0.0
-	for _, n := range f {
+	for _, n := range vals {
```
benraskin92: 👍

src/query/storage/consolidate.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/m3db/m3/src/query/block"
+)
+
+type consolidatedBlock struct {
+	unconsolidated    block.UnconsolidatedBlock
+	consolidationFunc block.ConsolidationFunc
+}
+
+func (c *consolidatedBlock) Unconsolidated() (block.UnconsolidatedBlock, error) {
+	return nil, fmt.Errorf("unconsolidated blocks are not supported")
```
benraskin92: `errors.New("unconsolidated blocks are not supported")`

src/query/ts/values.go
```diff
@@ -64,6 +68,40 @@ func (d Datapoints) ValueAt(n int) float64 { return d[n].Value }
...
+func (d Datapoints) Values() []float64 {
+	values := make([]float64, len(d))
+	for i, dp := range d {
+		values[i] = dp.Value
+	}
+
+	return values
+}
+
+// Aligned returns values aligned to given bounds.
```
arnikola: Consider renaming to `AlignToBounds` or something?

src/query/block/types.go
```diff
@@ -37,6 +41,18 @@ type Block interface {
 	Close() error
 }
 
+// UnconsolidatedBlock represents a group of unconsolidated series across a time bound
+type UnconsolidatedBlock interface {
```
arnikola: Is there ever a scenario where you'd want to consolidate a single series rather than the entire block? If not, might be good to add a capability to do this at the block level

src/query/block/types.go
```diff
@@ -62,6 +78,19 @@ type SeriesIter interface {
 	Meta() Metadata
 }
 
+// UnconsolidatedSeriesIter iterates through a block horizontally
+type UnconsolidatedSeriesIter interface {
+	Iterator
+	// Current returns the current series for the block
+	Current() (UnconsolidatedSeries, error)
```
arnikola: Current() typically panics rather than returning an error

src/query/block/types.go
```diff
@@ -62,6 +78,19 @@ type SeriesIter interface {
...
+	// SeriesCount returns the number of series
+	SeriesCount() int
+	// SeriesMeta returns the metadata for each series in the block
+	SeriesMeta() []SeriesMeta
+	// Meta returns the metadata for the block
+	Meta() Metadata
```
arnikola: Can roll these into a single interface and share that with `SeriesIter`

src/query/block/types.go
```diff
@@ -75,12 +104,31 @@ type StepIter interface {
...
+type UnconsolidatedStepIter interface {
+	Iterator
+	// Current returns the current step for the block
+	Current() (UnconsolidatedStep, error)
+	// StepCount returns the number of steps
+	StepCount() int
+	// SeriesMeta returns the metadata for each series in the block
+	SeriesMeta() []SeriesMeta
+	// Meta returns the metadata for the block
+	Meta() Metadata
```
arnikola: Can roll these into a single interface and share that with `StepIter`

arnikola: Actually, might be a good idea to pull 
```
		// SeriesMeta returns the metadata for each series in the block
	SeriesMeta() []SeriesMeta
	// Meta returns the metadata for the block
	Meta() Metadata
```
into a single interface since a lot of these use them

src/query/storage/consolidate.go
```diff
@@ -0,0 +1,140 @@
...
+		unconsolidated:    seriesIter,
+		consolidationFunc: c.consolidationFunc,
+	}, nil
+}
+
+func (c *consolidatedBlock) Close() error {
+	return c.unconsolidated.Close()
+}
+
+type consolidateStepIter struct {
```
benraskin92: `consolidatedStepIter`?

src/query/test/block.go
```diff
@@ -42,12 +44,47 @@ func NewBlockFromValues(bounds models.Bounds, seriesValues [][]float64) block.Bl
...
-// NewMultiBlocksFromValues creates new blocks using the provided values and a modifier
-func NewMultiBlocksFromValues(bounds models.Bounds, seriesValues [][]float64, valueMod ValueMod, numBlocks int) []block.Block {
+// NewUnconsolidatedBlockFromDatapointsWithMeta creates a new unconsolidated block using the provided values and metadata
+func NewUnconsolidatedBlockFromDatapointsWithMeta(bounds models.Bounds, meta []block.SeriesMeta, seriesValues [][]float64) block.Block {
+	seriesList := make(ts.SeriesList, len(seriesValues))
+	for i, values := range seriesValues {
+		dps := seriesValuesToDatapoints(values, bounds)
+		seriesList[i] = ts.NewSeries(meta[i].Name, dps, meta[i].Tags)
+	}
+	b, _ := storage.NewMultiSeriesBlock(seriesList, &storage.FetchQuery{
```
arnikola: nit: newline

src/query/storage/consolidate.go
```diff
@@ -0,0 +1,140 @@
...
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package storage
```
arnikola: nit: rename to `consolidated.go`

src/query/storage/consolidate.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (c *consolidateStepIter) SeriesMeta() []block.SeriesMeta {
+	return c.unconsolidated.SeriesMeta()
+}
+
+func (c *consolidateStepIter) Meta() block.Metadata {
+	return c.unconsolidated.Meta()
+}
+
+type consolidateSeriesIter struct {
```
benraskin92: `consolidatedSeriesIter`?

src/query/storage/consolidate.go
arnikola: nit: rename to consolidatedStepIter

src/query/storage/consolidate.go
arnikola: nit: rename to consolidatedSeriesIter

src/query/test/block.go
```diff
@@ -42,12 +44,47 @@ func NewBlockFromValues(bounds models.Bounds, seriesValues [][]float64) block.Bl
...
 
-// NewMultiBlocksFromValues creates new blocks using the provided values and a modifier
-func NewMultiBlocksFromValues(bounds models.Bounds, seriesValues [][]float64, valueMod ValueMod, numBlocks int) []block.Block {
+// NewUnconsolidatedBlockFromDatapointsWithMeta creates a new unconsolidated block using the provided values and metadata
+func NewUnconsolidatedBlockFromDatapointsWithMeta(bounds models.Bounds, meta []block.SeriesMeta, seriesValues [][]float64) block.Block {
+	seriesList := make(ts.SeriesList, len(seriesValues))
+	for i, values := range seriesValues {
+		dps := seriesValuesToDatapoints(values, bounds)
+		seriesList[i] = ts.NewSeries(meta[i].Name, dps, meta[i].Tags)
+	}
```
benraskin92: nit: newline

src/query/ts/values.go
```diff
@@ -64,6 +68,40 @@ func (d Datapoints) ValueAt(n int) float64 { return d[n].Value }
...
+func (d Datapoints) Aligned(bounds models.Bounds) []Datapoints {
+	numDatapoints := len(d)
```
benraskin92: `d.Len()`?

src/query/ts/values.go
```diff
@@ -64,6 +68,40 @@ func (d Datapoints) ValueAt(n int) float64 { return d[n].Value }
...
+	values := make([]float64, len(d))
+	for i, dp := range d {
+		values[i] = dp.Value
+	}
+
+	return values
+}
+
+// Aligned returns values aligned to given bounds.
+func (d Datapoints) Aligned(bounds models.Bounds) []Datapoints {
```
benraskin92: Rename to `Align`?

src/query/block/types.go
nikunjgit: to ensure TakeLast actually implements ConsolidationFunc. I guess i can remove it since  multiSeriesBlock Consolidate will throw a compilation error if TakeLast doesn't implement it

src/query/executor/transform/lazy.go
nikunjgit: we use "for the block" in most other places.

src/query/block/series.go
```diff
@@ -45,3 +49,39 @@ func (s Series) Values() []float64 {
...
+	datapoints []ts.Datapoints
+	Meta       SeriesMeta
+}
+
+// NewUnconsolidatedSeries creates a new series with raw datapoints
+func NewUnconsolidatedSeries(datapoints []ts.Datapoints, meta SeriesMeta) UnconsolidatedSeries {
+	return UnconsolidatedSeries{datapoints: datapoints, Meta: meta}
+}
+
+// DatapointsAtStep returns the raw datapoints at a step index
```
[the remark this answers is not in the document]
nikunjgit: it's returning multiple datapoints ?

src/query/block/types.go
```diff
@@ -22,13 +22,17 @@ package block
...
 	"time"
 
 	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/ts"
 )
 
 // Block represents a group of series across a time bound
 type Block interface {
+	// Unconsolidated returns the unconsolidated version of the block
+	Unconsolidated() (UnconsolidatedBlock, error)
```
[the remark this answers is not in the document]
nikunjgit: Even for things which can support unconsolidated, you can still get an error. So we need an error here neverthless. I like this over having a bool so that clients don't check 2 things

src/query/test/block.go
```diff
@@ -42,12 +44,47 @@ func NewBlockFromValues(bounds models.Bounds, seriesValues [][]float64) block.Bl
...
+// NewUnconsolidatedBlockFromDatapoints creates a new unconsolidated block using the provided values
+func NewUnconsolidatedBlockFromDatapoints(bounds models.Bounds, seriesValues [][]float64) block.Block {
+	meta := NewSeriesMeta("dummy", len(seriesValues))
+	return NewUnconsolidatedBlockFromDatapointsWithMeta(bounds, meta, seriesValues)
+}
+
+func seriesValuesToDatapoints(values []float64, bounds models.Bounds) ts.Datapoints {
+	dps := make(ts.Datapoints, len(values))
+	for i, v := range values {
+		t, _ := bounds.TimeForIndex(i)
```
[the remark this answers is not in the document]
nikunjgit: I was assuming we'd just provide values within bounds for tests. Didn't want each test to check for error

src/query/test/block.go
```diff
@@ -42,12 +44,47 @@ func NewBlockFromValues(bounds models.Bounds, seriesValues [][]float64) block.Bl
...
+			Value:     v,
+		}
+	}
+
+	return dps
+}
+
+// NewMultiUnconsolidatedBlocksFromValues creates new blocks using the provided values and a modifier
+func NewMultiUnconsolidatedBlocksFromValues(bounds models.Bounds, seriesValues [][]float64, valueMod ValueMod, numBlocks int) []block.Block {
 	meta := NewSeriesMeta("dummy", len(seriesValues))
```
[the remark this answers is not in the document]
nikunjgit: test ?

src/query/ts/values.go
```diff
@@ -64,6 +68,40 @@ func (d Datapoints) ValueAt(n int) float64 { return d[n].Value }
...
+	numDatapoints := len(d)
+	steps := bounds.Steps()
+	stepValues := make([]Datapoints, steps)
+	dpIdx := 0
+	for i := 0; i < steps; i++ {
+		startIdx := dpIdx
+		t, _ := bounds.TimeForIndex(i)
```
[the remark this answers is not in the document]
nikunjgit: umm not really since i'm bounding it on bounds.Steps() ?

src/query/block/types.go
nikunjgit: Yeah, the UnconsolidatedBlock has a Consolidate() which does it at the block level. Not super sure what you mean.

src/query/block/types.go
nikunjgit: all other Current() also return error

src/query/block/types.go
nikunjgit: good point

src/query/ts/values.go
nikunjgit: @arnikola wanted it AlignToBounds. Is that better ?

src/query/block/column.go
```diff
@@ -36,6 +36,11 @@ type columnBlock struct {
 	seriesMeta []SeriesMeta
 }
 
+// Unconsolidated returns the unconsolidated version for the block
+func (c *columnBlock) Unconsolidated() (UnconsolidatedBlock, error) {
+	return nil, fmt.Errorf("unconsolidated view not support for block, meta: %s", c.meta)
```
benraskin92: `s/support/supported`

src/query/storage/block.go
```diff
@@ -94,14 +138,36 @@ func (m multiSeriesBlock) SeriesMeta() []block.SeriesMeta {
 	return metas
 }
 
+func (m multiSeriesBlock) Consolidate() (block.Block, error) {
+	return &consolidatedBlock{
+		unconsolidated:    m,
+		consolidationFunc: block.TakeLast,
```
benraskin92: We'll use a different consolidation func in the future, right?

nikunjgit: prometheus only has gauges so TakeLast is only reasonable option. Yeah, we would probably have different once we have different types.

src/query/test/block.go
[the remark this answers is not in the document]
arnikola: lol oops

src/query/ts/values.go
arnikola: Early bird gets the worm @benraskin92  😜

src/query/functions/temporal/base.go
```diff
@@ -285,14 +295,27 @@ func (c *baseNode) processSingleRequest(request processRequest) error {
...
 				values = values[len(values)-desiredLength:]
-				newVal = c.processor.Process(values)
+				flattenedValues := make(ts.Datapoints, 0)
+				for _, dps := range values {
+					for _, dp := range dps {
+						if dp.Timestamp.Before(oldestDatapointTimestamp) {
+							continue
+						}
+
+						flattenedValues = append(flattenedValues, dp)
```
arnikola: I think you can skip a bunch of steps on the inner loop, with something like 

```golang
var takeIdx int
for idx, dp := range dps{
 ...
 takeIdx = idx
 break
}

flattenedValues = append(flattenedValues, dps[takeIdx:]...)
```

src/query/storage/block.go
```diff
@@ -31,37 +31,81 @@ import (
...
+func (m *multiBlockWrapper) StepIter() (block.StepIter, error) {
+	if err := m.tryConsolidate(); err != nil {
+		return nil, err
+	}
+
+	return m.consolidated.StepIter()
+}
+
+func (m *multiBlockWrapper) tryConsolidate() error {
+	m.mu.Lock()
```
arnikola: Good opportunity for `sync.Once` :)

src/query/storage/block.go
```diff
@@ -31,37 +31,81 @@ import (
...
+
+func (m *multiBlockWrapper) StepIter() (block.StepIter, error) {
+	if err := m.tryConsolidate(); err != nil {
+		return nil, err
+	}
+
+	return m.consolidated.StepIter()
+}
+
+func (m *multiBlockWrapper) tryConsolidate() error {
```
arnikola: nit: rename to `consolidate`?

src/query/storage/block.go
```diff
@@ -31,37 +31,81 @@ import (
...
+func (m *multiBlockWrapper) SeriesIter() (block.SeriesIter, error) {
+	if err := m.tryConsolidate(); err != nil {
+		return nil, err
+	}
+
+	return m.consolidated.SeriesIter()
+}
+
+func (m *multiBlockWrapper) Close() error {
+	return m.unconsolidated.Close()
```
arnikola: Should also close consolidated?

src/query/storage/block.go
```diff
@@ -113,28 +179,24 @@ func (m *multiSeriesBlockStepIter) Meta() block.Metadata {
...
-	for i, s := range m.block.seriesList {
-		if m.index < seriesLen {
-			values[i] = s.Values().ValueAt(m.index)
-		} else {
-			values[i] = math.NaN()
-		}
+func (m *multiSeriesBlockStepIter) Current() (block.UnconsolidatedStep, error) {
+	values := make([]ts.Datapoints, len(m.values))
+	for i, series := range m.values {
+		values[i] = series[m.index]
```
arnikola: Doesn't this need to pad with NaNs after the end too?

src/query/storage/block.go
```diff
@@ -172,24 +234,39 @@ func (m *multiSeriesBlockSeriesIter) Next() bool {
...
+	values := make([]ts.Datapoints, m.block.StepCount())
+	seriesValues := s.Values().AlignToBounds(m.block.meta.Bounds)
+	seriesLen := len(seriesValues)
 	for i := 0; i < m.block.StepCount(); i++ {
 		if i < seriesLen {
-			values[i] = seriesValues.ValueAt(i)
+			values[i] = seriesValues[i]
 		} else {
-			values[i] = math.NaN()
+			values[i] = nil
```
arnikola: Seems a little dangerous to add nils here where we used to have NaNs, any chance that this could cause issues in scenarios where we assumed we were only getting NaNs and values?

src/query/block/column.go
```diff
@@ -36,6 +36,11 @@ type columnBlock struct {
 	seriesMeta []SeriesMeta
 }
 
+// Unconsolidated returns the unconsolidated version for the block
```
arnikola: nit: `of the block`

src/query/functions/temporal/functions.go
```diff
@@ -70,28 +71,28 @@ type functionNode struct {
...
-	if len(values) == 0 {
+func (f *functionNode) Process(datapoints ts.Datapoints) float64 {
+	if len(datapoints) == 0 {
 		return math.NaN()
 	}
 
 	allNaNs := true
 	result := 0.0
-	prev := values[0]
+	prev := datapoints[0].Value
```
arnikola: Any chance these can be nil?

src/query/test/block.go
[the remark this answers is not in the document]
arnikola: My bad didn't catch these were test utilities

src/query/storage/consolidated.go
```diff
@@ -0,0 +1,140 @@
...
+
+package storage
+
+import (
+	"errors"
+
+	"github.com/m3db/m3/src/query/block"
+)
+
+type consolidatedBlock struct {
```
arnikola: nit: consolidatedBlock is a bit of a weird name for something containing an unconsolidated block and a consolidation block

src/query/ts/values.go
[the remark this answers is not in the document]
arnikola: Then might be a bit better to do the calculation here explicitly to avoid the overhead of the bounds checking in .TimeForIndex()?

src/query/block/series.go
```diff
@@ -20,6 +20,10 @@
 
 package block
 
+import (
+	"github.com/m3db/m3/src/query/ts"
```
arnikola: Can you use `"github.com/m3db/m3/src/dbnode/ts"` and delete this one?

src/query/block/types.go
```diff
@@ -104,3 +154,15 @@ type Builder interface {
...
+// ConsolidationFunc consolidates a bunch of datapoints into a single float value
+type ConsolidationFunc func(datapoints ts.Datapoints) float64
+
+// TakeLast is a consolidation function which takes the last datapoint
+func TakeLast(values ts.Datapoints) float64 {
+	if len(values) == 0 {
+		return math.NaN()
+	}
+
+	return values[len(values)-1].Value
```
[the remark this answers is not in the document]
arnikola: Something like this should work?

```golang
	for i := len(values) - 1; i >= 0; i-- {
		value := values[i].Value
		if !math.IsNaN(value) {
			return value
		}
	}

	return math.NaN()
```

src/query/block/scalar.go
```diff
@@ -47,6 +47,11 @@ func NewScalar(val float64, bounds models.Bounds) Block {
 	}
 }
 
+// Unconsolidated returns the unconsolidated version for the block
```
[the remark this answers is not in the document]
nikunjgit: Ben had the same comment. We use "for the block" almost everywhere

src/query/block/series.go
```diff
@@ -45,3 +49,43 @@ func (s Series) Values() []float64 {
...
+	return s.datapoints
+}
+
+// Len returns the number of datapoints slices in the series
+func (s UnconsolidatedSeries) Len() int {
+	return len(s.datapoints)
+}
+
+// Consolidated consolidates the series
+func (s UnconsolidatedSeries) Consolidated(consolidationFunc ConsolidationFunc) Series {
```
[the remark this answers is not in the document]
nikunjgit: it's just talking about a view. Maybe we already have it cached and don't really consolidate.

src/query/block/scalar.go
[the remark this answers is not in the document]
arnikola: It's grammatically correct almost everywhere, just not here 😛

src/query/ts/values.go
```diff
@@ -98,6 +136,17 @@ func (b *fixedResolutionValues) DatapointAt(point int) Datapoint {
 	}
 }
 
+// AlignToBounds returns values aligned to given bounds.
+// TODO: Consider bounds as well
+func (b *fixedResolutionValues) AlignToBounds(_ models.Bounds) []Datapoints {
+	values := make([]Datapoints, len(b.values))
+	for i := 0; i < b.Len(); i++ {
+		values[i] = Datapoints{b.DatapointAt(i)}
```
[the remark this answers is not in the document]
nikunjgit: This is a little cleaner since Datapoint already does this logic

src/query/executor/transform/lazy.go
```diff
@@ -145,6 +145,22 @@ type lazyBlock struct {
 	processedBlock block.Block
 }
 
+// Unconsolidated returns the unconsolidated version for the block
+func (f *lazyBlock) Unconsolidated() (block.UnconsolidatedBlock, error) {
+	f.mu.Lock()
```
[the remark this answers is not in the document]
nikunjgit: can't use sync.Once since i need to look every read of f.processedBlock under the same mutex

src/query/storage/block.go
nikunjgit: the person using that block is supposed to close ?

src/query/storage/block.go
nikunjgit: we have all series here ? They are padded with NaNs themselves

src/query/storage/block.go
nikunjgit: this is pointer to pointer now (ptr to datapoints). During consolidation, we would check for 0 datapoints, we should return NaN

src/query/block/column.go
nikunjgit: we use for the block at most places. Can do a search and replace in a separate diff

src/query/functions/temporal/functions.go
nikunjgit: we already have a length check ?

src/query/storage/consolidated.go
nikunjgit: it doesn't have consolidation block just a consolidatationFunc

src/query/functions/temporal/functions.go
arnikola: I mean datapoints[0] could be nil, so you might hit NPE when you call .Value

src/query/storage/block.go
```diff
@@ -31,37 +31,81 @@ import (
...
+func (m *multiBlockWrapper) consolidate() error {
+	m.once.Do(func() {
+		consolidated, err := m.unconsolidated.Consolidate()
+		if err != nil {
+			m.consolidateError = err
+			return
+		}
+
+		m.consolidated = consolidated
+
```
benraskin92: nit: newline

src/query/ts/values.go
```diff
@@ -98,6 +138,17 @@ func (b *fixedResolutionValues) DatapointAt(point int) Datapoint {
 	}
 }
 
+// AlignToBounds returns values aligned to given bounds.
```
benraskin92: Kind of weird to call this `AlignToBounds`, but you just disregard the bounds..

src/query/storage/block.go
arnikola: Can you add a comment indicating that it should only close unconsolidated in case someone accidentally notices it and tries to close the consolidated here

src/query/storage/consolidated.go
```diff
@@ -0,0 +1,140 @@
...
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
```
robskillington: nit: This license text comment should be `//` prefixed per line like the other `.go` files.

PR #936, opened 2018-09-24
Add benchmark loadgen docker compose config

robskillington:


scripts/benchmarks/benchmark-loadgen/benchmark-loadgen.sh
```diff
@@ -0,0 +1,42 @@
...
+PARAM_COORDINATOR_DEBUG_PORT=${COORDINATOR_DEBUG_PORT:-7203}
+
+PARAM_DBNODES=${DBNODES:-127.0.0.1}
+PARAM_DBNODE_DEBUG_PORT=${DBNODE_DEBUG_PORT:-9004}
+
+CFG_COORDINATOR="${PARAM_COORDINATOR}:${PARAM_COORDINATOR_API_PORT}"
+
+CFG_SCRAPE_FAKEWEBSERVERS=""
+for N in $(seq 1 $PARAM_NUM_INSTANCES); do
+  PORT=$(expr 8080 + $N - 1)
```
arnikola: Would it be better to pick a more esoteric port range?

scripts/benchmarks/benchmark-loadgen/docker-compose.yaml
```diff
@@ -0,0 +1,32 @@
+version: "3.5"
+services:
+  fakewebserver01:
+    expose:
+      - "7000-8000"
+    ports:
+      - "0.0.0.0:8080-9080:7000-8000"
```
prateek: Might be easier to not do this transpose

scripts/benchmarks/benchmark-loadgen/benchmark-loadgen.sh
robskillington: This is internal to the backend thankfully.

PR #938, opened 2018-09-25
Fix data race in FST Segment Reads

prateek:
H/T @sboschman for debugging and suggesting the issue. 

Example failure of added test without fix - https://pastebin.com/mtBbNUr6

### DIFF
--- a/src/m3ninx/search/proptest/concurrent_test.go
+++ b/src/m3ninx/search/proptest/concurrent_test.go
@@ -1,5 +1,3 @@
-// +build big
-
 // Copyright (c) 2018 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
@@ -25,6 +23,7 @@ package proptest
 import (
 	"math/rand"
 	"os"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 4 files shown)


src/m3ninx/search/proptest/concurrent_test.go
```diff
@@ -0,0 +1,102 @@
...
+			require.NoError(t, err)
+
+			var (
+				wg       sync.WaitGroup
+				errLock  sync.Mutex
+				matchErr error
+			)
+			wg.Add(2)
+
+			for i := 0; i < 2; i++ {
```
robskillington: Is it worth adding some more parallelism to more likely hit the edge case?

src/m3ninx/search/proptest/concurrent_test.go
```diff
@@ -0,0 +1,102 @@
...
+			require.NoError(t, err)
+			docMatcher, err := newDocumentIteratorMatcher(matchedDocs...)
+			require.NoError(t, err)
+
+			var (
+				wg       sync.WaitGroup
+				errLock  sync.Mutex
+				matchErr error
+			)
+			wg.Add(2)
```
robskillington: super nit: Perhaps put a number up front which you can pass to both `wg.Add(...)` and the `for` loop so that they always match (easier to turn up/down parallelism).

src/m3ninx/search/proptest/concurrent_test.go
prateek: Nah, even this version with 2 routines fails pretty consistently within 1-2 of the sub-tests running (of the 100 prop test runs per test execution). No point adding more, it'll only serve to slow down CI

src/m3ninx/search/proptest/concurrent_test.go
prateek: sure, opened PR https://github.com/m3db/m3/pull/940

PR #942, opened 2018-09-25
Use pooling for writes to coordinator

nikunjgit:


src/query/pools/query_pools.go
```diff
@@ -79,7 +79,13 @@ func BuildWorkerPools(
...
 	})
 
-	return objectPool, instrumentOptions
+	writePoolSize := cfg.WriteWorkerPoolSize
+	if writePoolSize == 0 {
+		writePoolSize = defaultWorkerPoolSize
+	}
+
+	writeWorkerPool := xsync.NewWorkerPool(writePoolSize)
+	return objectPool, writeWorkerPool, instrumentOptions
```
arnikola: We're getting quite a few things returns here, consider bundling it? I foresee a lot more pools being added as we work on perf :P

src/query/storage/m3/storage.go
```diff
@@ -257,20 +260,28 @@ func (s *m3storage) Write(
...
-		store:       s,
-		annotation:  query.Annotation,
-		unit:        query.Unit,
-		id:          identID,
-		tagIterator: storage.TagsToIdentTagIterator(query.Tags),
-		attributes:  query.Attributes,
-	}
+	tagIterator := storage.TagsToIdentTagIterator(query.Tags)
+
+	var (
```
arnikola: Considering how this is being broken up, it may be a good idea to use the pooled workerpools like read does? That way every incoming request will get a set of workers for all datapoints in it and we don't have situations where some requests get starved 

If we want to simplify it a bit, can revisit my m3x PR at: https://github.com/m3db/m3x/pull/181 and make it a first class citizen, then refactor the readerpools too?

Alternatively, just use the same pool as the readerpools here if you agree with taking this approach

src/query/storage/m3/storage.go
```diff
@@ -257,20 +260,28 @@ func (s *m3storage) Write(
...
+		wg       sync.WaitGroup
+		multiErr xerrors.MultiError
+	)
+
+	for _, datapoint := range query.Datapoints {
+		s.writeWorkerPool.Go(func() {
```
arnikola: I think you need to capture the datapoint variable here; surprised that there are no failing tests... might be good to add one that tries to write multiple datapoints and checks that the correct ones are being written?

src/query/storage/m3/storage.go
```diff
@@ -257,20 +260,28 @@ func (s *m3storage) Write(
...
+			wg.Add(1)
+			tagIter := tagIterator.Duplicate()
```
arnikola: Are these threadsafe? Also since these all use the same iterator, can you define this outside of the loop?

src/query/storage/m3/storage.go
```diff
@@ -257,20 +260,28 @@ func (s *m3storage) Write(
 	id := query.Tags.ID()
```
arnikola: Maybe we should consider renaming WriteQuery to WriteRequest or something, query implies a response to a question

src/query/storage/m3/storage.go
```diff
@@ -281,31 +292,32 @@ func (s *m3storage) Close() error {
...
-	common := w.writeRequestCommon
-	store := common.store
-	id := common.id
-
+func (s *m3storage) writeSingle(
+	ctx context.Context,
+	query *storage.WriteQuery,
+	datapoint ts.Datapoint,
+	identID ident.ID,
+	iterator ident.TagIterator) error {
```
arnikola: nit:
```golang
 iterator ident.TagIterator,
) error {
```

src/query/storage/m3/storage.go
```diff
@@ -281,31 +292,32 @@ func (s *m3storage) Close() error {
...
 	var (
 		namespace ClusterNamespace
 		err       error
 	)
-	switch common.attributes.MetricsType {
+	switch query.Attributes.MetricsType {
```
arnikola: nit: pull out `query.Attributes` and `s.clusters` into variables

src/query/pools/query_pools.go
richardartoul: Wanna use the new worker pool I wrote for the m3db ingesters? Its way more performant because it doesn't allocate goroutines all the time: https://github.com/m3db/m3x/blob/master/sync/pooled_worker_pool.go

That should help fix a lot of the `runtime.morestack/runtime.growstack/runtime.copystack` in the m3coordinator flame graphs

richardartoul: I'm pretty sure its API compatible too so I think you can literally just replace the constructors and be done

src/query/pools/query_pools.go
```diff
@@ -53,7 +53,7 @@ func BuildWorkerPools(
 	cfg config.Configuration,
 	logger *zap.Logger,
 	scope tally.Scope,
-) (pool.ObjectPool, instrument.Options) {
+) (pool.ObjectPool, xsync.WorkerPool, instrument.Options) {
```
richardartoul: Out of curiousity, why is there a pool of worker pools?

src/query/pools/query_pools.go
```diff
@@ -79,7 +79,13 @@ func BuildWorkerPools(
 		return workerPool
 	})
 
-	return objectPool, instrumentOptions
+	writePoolSize := cfg.WriteWorkerPoolSize
+	if writePoolSize == 0 {
+		writePoolSize = defaultWorkerPoolSize
+	}
+
+	writeWorkerPool := xsync.NewWorkerPool(writePoolSize)
```
richardartoul: I think you need to call init here

src/query/storage/m3/storage.go
richardartoul: Can you explain the rationale for pooled worker pools? Kind of confused by that

arnikola: General idea was to give each incoming request its own pool while having a static max size, so that the more expensive requests would not tie up smaller requests, if that's not the correct approach, happy to drop it and revert to your pooled worker pool

src/query/pools/query_pools.go
arnikola: Addressed this quesiton in the other comment

src/query/storage/m3/storage.go
richardartoul: Sync'd up in person and settled on (pending discussion with @nikunjgit) replacing the existing WorkerPool with PooledWorkerPool and then getting rid of the generic ObjectPool on top of that

BertHartm:
I'm happy to test this under load if it's at a point where that would make sense.

matejzero:
> I'm happy to test this under load if it's at a point where that would make sense.

I'm happy to test too (with around 70k metrics/s).

src/query/storage/m3/storage.go
nikunjgit: i'll use a PooledWorkerPool for writes and let @arnikola merge the read object pool in another diff. Sound good ?

src/query/storage/m3/storage.go
nikunjgit: good point.

src/query/storage/m3/storage.go
nikunjgit: yeah makes sense

src/query/storage/m3/storage.go
nikunjgit: used at a lot of places. I'd prefer to avoid that change in this diff

src/query/pools/query_pools.go
nikunjgit: good point

src/query/storage/m3/storage.go
nikunjgit: i think s.clusters is fine

src/query/storage/m3/storage.go
arnikola: Sure

src/query/test/m3/storage.go
```diff
@@ -54,6 +55,9 @@ func NewStorageAndSession(
 		Retention:   TestRetention,
 	})
 	require.NoError(t, err)
-	storage := m3.NewStorage(clusters, nil)
+	writePool, err := sync.NewPooledWorkerPool(10, sync.NewPooledWorkerPoolOptions())
```
arnikola: Consider a test to make sure we're using the pools?

src/query/storage/m3/storage.go
```diff
@@ -317,33 +333,6 @@ func (w *writeRequest) Process(ctx context.Context) error {
 
 	namespaceID := namespace.NamespaceID()
 	session := namespace.Session()
-	return session.WriteTagged(namespaceID, id, common.tagIterator,
-		w.timestamp, w.value, common.unit, common.annotation)
-}
-
-type writeRequestCommon struct {
```
arnikola: :+1:

PR #943, opened 2018-09-25
Add endpoint that returns error if node is not bootstrapped

richardartoul:


richardartoul:
@schallert 

Behold: https://github.com/m3db/m3/blob/c226f465e23266e1c74a2a784a78b25c23e6c19e/src/dbnode/network/server/httpjson/handlers.go#L58

prateek:
s/behold/
> Look on my works, ye Mighty, and despair!'

PR #954, opened 2018-09-27
[WIP] Thrift Protocol Pooling

prateek:
sync.Pool seems to get much better latency


prateek:
Don't need this right now

PR #956, opened 2018-09-27
Set namespace and ID as NoFinalize

arnikola:
This allows the session to avoid cloning these on the WriteTagged
operation, which will save two allocations for each datapoint

### DIFF
--- a/src/query/storage/m3/storage.go
+++ b/src/query/storage/m3/storage.go
@@ -257,6 +257,8 @@ func (s *m3storage) Write(
 	id := query.Tags.ID()
 	// TODO: Consider caching id -> identID
 	identID := ident.StringID(id)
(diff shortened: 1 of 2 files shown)


src/query/storage/m3/storage.go
```diff
@@ -257,6 +257,8 @@ func (s *m3storage) Write(
 	id := query.Tags.ID()
 	// TODO: Consider caching id -> identID
```
richardartoul: Caching these in an LRU would actually probably save a ton of allocs unrelated to this P.R

arnikola: I think there's some other refactoring of this going on in https://github.com/m3db/m3/pull/942, so don't want to step on any toes there and lead to messy merges; we'll definitely circle back on adding an LRU here

PR #958, opened 2018-09-27
Arnikola/encoded blocks

arnikola:
Added encoded blocks which provides a step and series iterator that works on compressed iterators


src/query/ts/m3db/encoded_series_iterator.go
```diff
@@ -0,0 +1,143 @@
...
+) block.SeriesIter {
+	return &encodedSeriesIter{
+		idx:         -1,
+		meta:        meta,
+		seriesMeta:  seriesMeta,
+		seriesIters: seriesIters,
+	}
+}
+
+func (it *encodedSeriesIter) currentUnconsolidated() (block.Series, error) {
```
nikunjgit: instead of providing a view for consolidated vs unconsolidated at each iteration, it might be better to provide a consolidatedSeriesIter vs unconsolidatedSeriesIter and all their iterations are of the same kind

src/query/ts/m3db/encoded_series_iterator.go
```diff
@@ -0,0 +1,143 @@
...
+		dp, _, _ := iter.Current()
+		values = append(values, dp.Value)
+	}
+
+	return block.NewSeries(values, it.seriesMeta[it.idx]), nil
+}
+
+func (it *encodedSeriesIter) currentConsolidated() (block.Series, error) {
+	iter := it.seriesIters[it.idx]
+	cs := it.consolidationSettings
```
nikunjgit: consider using the consolidationFunc concept ?

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,251 @@
...
+
+type encodedStep struct {
+	time   time.Time
+	values []float64
+}
+
+func (s *encodedStep) Time() time.Time   { return s.time }
+func (s *encodedStep) Values() []float64 { return s.values }
+
+func (it *encodedStepIter) currentUnconsolidated() (block.Step, error) {
```
nikunjgit: same comment as for series regarding two views of the data

src/query/ts/m3db/encoded_series_iterator.go
arnikola: Yeah this is a bit of a stopgap until your PR lands, then I'll change this encoded series iterator to satisfy both `block.Block` and `block.UnconsolidatedBlock`

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,114 @@
...
+		}
+
+		b.buildMeta()
+	})
+
+	return b.meta, b.seriesMetas, b.err
+}
+
+func (b *encodedBlock) StepIter() (block.StepIter, error) {
+	meta, seriesMetas, err := b.generateMetas()
```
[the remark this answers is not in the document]
arnikola: Yeah that's fair; for some reason I was thinking that things could be calling StepIter and the like while it's still initializing in the `Make` function, which in retrospect doesn't make sense

src/query/ts/m3db/convert.go
```diff
@@ -33,6 +33,9 @@ const (
 	initBlockReplicaLength = 10
 )
 
+// Represents a vertical slice of series iterators that represent a block
+type seriesBlocks []encoding.SeriesIterator
```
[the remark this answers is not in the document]
arnikola: Changed it to `seriesIteratorBlocks` as I want it to keep a bit of context that it's in a block(-ish) format

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// MakeEncodedBlock builds an encoded block
+func MakeEncodedBlock(seriesBlockIterators []encoding.SeriesIterator) block.Block {
+	b := &encodedBlock{
+		seriesBlockIterators: seriesBlockIterators,
+	}
+
+	// generate metadata and series metadatas
+	b.generateMetas()
```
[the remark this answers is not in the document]
arnikola: Good idea

Issue #959, opened 2018-09-28
Docs: /etc/security/limits.conf don't work with systemd


matejzero:
The [documentation](https://m3db.github.io/m3/how_to/cluster_hard_way/) on the home page says that we should set `nofile` limits in `/etc/security/limits.conf`. On systems using `systemd`, this setting wont have any effect, because `systemd` is not taking it into an account.

Quote from [ArchLinux wiki](https://wiki.archlinux.org/index.php/Limits.conf):
```
Note: /etc/security/limits.conf does not work if systemd exists. The files /etc/systemd/system.conf, /etc/systemd/user.conf, and /etc/systemd/<systemd_unit>/override.conf would control the limit. See the systemd-system.conf(5) man page for details.
```

I think it would be good to edit/add documentation for users with `systemd` to set those limits in `/etc/systemd/.conf` files or in the specific systemd service file.

richardartoul:
@schallert Do you mind taking a look at this one when you get a chance?

schallert:
Yeah I'll see if it makes sense to just add a base systemd service config to the repo.

PR #961, opened 2018-09-28
Add documentation for systemd service

matejzero:
This is a proposed fix for #959.

Feel free to edit, since I'm not a native english speaker.


prateek:
@schallert @nerd0 mind taking a look and reconciling with the example we have in the repo.

matejzero:
I like @schallert's solution. It's a simple download & run solution. In my case you need to write the whole systemd service file:)

prateek:
@matejzero awesome! if that addressed your concern, can we close this PR?

matejzero:
Sure!

PR #963, opened 2018-09-28
Use dynamic $__interval variable for query step

matejzero:
This fix uses `$__interval` variable to calculate query step. It won't go below 2*scrape interval, so you will always get a graph, but it will try to pick the best query step given the selected time range and width of the graphs.

More info on http://docs.grafana.org/reference/templating/#the-interval-variable

### DIFF
--- a/integrations/grafana/m3db_dashboard.json
+++ b/integrations/grafana/m3db_dashboard.json
@@ -52,7 +52,7 @@
   "gnetId": null,
   "graphTooltip": 1,
   "id": null,
-  "iteration": 1538144350249,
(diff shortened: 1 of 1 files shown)


matejzero:
As we figured out with @richardartoul, this dashboard is not working correctly without having Prometheus datasource scrape interval in Grafana set to 2*scrape interval.

There is a [feature request](
https://github.com/grafana/grafana/issues/11451) opened in Grafana's github, but there is no solution yet.

I'm closing this since it's not working as it should. Too bad, because $__interval does bring performance improvements.

PR #970, opened 2018-09-28
[contrib] systemd: add systemd unit example

### LINKED ISSUE #959: Docs: /etc/security/limits.conf don't work with systemd

schallert:
Users will likely have to change this to reflect where their binaries
are located, but provides a base example.

Fixes #959.


richardartoul:
Want to put this under the integrations folder?

richardartoul:
Also would be helpful to link to this file from the docs

integrations/systemd/m3dbnode.service
```diff
@@ -0,0 +1,15 @@
+[Unit]
+Description="M3DB Timeseries Database"
+Documentation=http://m3db.github.io/m3/
+After=network.target
+
+[Service]
+Type=simple
+ExecStart=/usr/bin/m3dbnode
```
nerd0: Curious if we should include the config flag within this ?

schallert: Ah damn good catch, opened #974

PR #981, opened 2018-09-29
[prototype] Make coordinator pretend Prom sends it []byte instead of string

prateek:
Exploiting the fact that GRPC's wire encoding format for strings and bytes is the same


prateek:
Artem's got this covered in #1004

PR #982, opened 2018-09-29
Migrate to dep

prateek:
Verified dependencies are the same between glide <-> dep files using: https://gist.github.com/prateek/0ff5dc3fabbda68f7c0c0c3d56368a67


robskillington:
Nice.  So I guess we're taking the plunge? 😆

I couldn't help but notice some build kite scripts in here too, post this PR we'd still be on Semaphore for a bit I assume?  Or are these files just preparation?

prateek:
talked offline: `dep ensure` takes a long time on semaphore for some reason. going to open a separate PR to get buildkite done. once that's landed and stable we'll do the dep migration.

prateek:
Can't believe I'm going to say this, lets wait for 1.11 and VGO

PR #984, opened 2018-09-29
Fix a bug in iterator deduplication logic

arnikola:


src/query/storage/m3/multi_fetch_result.go
```diff
@@ -22,20 +22,20 @@ package m3
...
 type multiFetchResult struct {
 	sync.Mutex
 	sync.Once
-	iterators        encoding.SeriesIterators
-	err              xerrors.MultiError
-	dedupeFirstAttrs storage.Attributes
-	dedupeMap        map[string]multiFetchResultSeries
+	iterators             encoding.SeriesIterators
+	err                   xerrors.MultiError
+	dedupeFirstResolution time.Duration
```
robskillington: Is there a major reason to change to `resolution time.Duration` instead of `attrs storage.Attributes`?  They are both value types that shouldn't be able to be nil'd.  

It's just I have an open PR that depends on having access to the attributes rather than just the resolution:
https://github.com/m3db/m3/pull/908

prateek:
@arnikola need to rebase, master was broken until a little bit ago

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -22,20 +22,20 @@ package m3
...
 
 // TODO: use a better seriesIterators merge here
 type multiFetchResult struct {
 	sync.Mutex
 	sync.Once
-	iterators        encoding.SeriesIterators
-	err              xerrors.MultiError
-	dedupeFirstAttrs storage.Attributes
-	dedupeMap        map[string]multiFetchResultSeries
+	iterators             encoding.SeriesIterators
```
prateek: could you add test(s) to repro the bug fix

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -65,49 +65,50 @@ func (r *multiFetchResult) setPools(pools encoding.IteratorPools) {
...
 	if r.iterators == nil {
 		r.iterators = iterators
-		r.dedupeFirstAttrs = attrs
+		r.dedupeFirstResolution = resolution
 		return
 	}
 
-	iters := iterators.Iters()
-	// Need to dedupe
+	iters := r.iterators.Iters()
```
robskillington: Ah this was the bug?

prateek:
addressed in #989

PR #986, opened 2018-09-29
Improve M3DB session performance part 2: Dont clone IDs if they are IsNoFinalize()

richardartoul:


src/dbnode/client/session.go
```diff
@@ -1185,7 +1185,7 @@ func (s *session) fetchTaggedAttemptWithRLock(
 ) (*fetchState, error) {
 	// NB(prateek): we have to clone the namespace, as we cannot guarantee the lifecycle
 	// of the hostQueues responding is less than the lifecycle of the current method.
-	nsClone := s.pools.id.Clone(ns)
+	nsClone := s.maybeClone(ns)
```
prateek: lets skip the fetch path for now. the impact isn't going to be nearly as pronounced and i don't want to land it without a lot more testing.

src/dbnode/client/session.go
```diff
@@ -966,8 +966,8 @@ func (s *session) writeAttemptWithRLock(
 	// use in the various queues. Tracking per writeAttempt isn't sufficient as
 	// we may enqueue multiple writeStates concurrently depending on retries
 	// and consistency level checks.
-	nsID := s.pools.id.Clone(namespace)
-	tsID := s.pools.id.Clone(id)
+	nsID := s.cloneFinalizable(namespace)
```
prateek: nit: also add a test cases to ensure we're doing the right thing. https://github.com/m3db/m3/blob/master/src/x/test/util.go#L30 won't be for naught

PR #989, opened 2018-09-30
Coordinator multi-fetch fixes

prateek:


src/query/storage/m3/multi_fetch_result.go
```diff
@@ -21,117 +21,151 @@
...
-				idx:   idx,
-				attrs: r.dedupeFirstAttrs,
-			}
-		}
+		r.dedupeMap = make(map[string][]multiResultSeries, len(iters))
+	}
+
+	for _, iter := range iters {
+		id := iter.ID().String()
+		r.dedupeMap[id] = append(r.dedupeMap[id], multiResultSeries{
```
richardartoul: Cant this panic if the exiting value is a nil slice?

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -21,117 +21,151 @@
...
-	iterators encoding.SeriesIterators,
-) {
-	iters := iterators.Iters()
-	for _, s := range iters {
-		id := s.ID().String()
-		existing, exists := r.dedupeMap[id]
-		if exists && existing.attrs.Resolution <= attrs.Resolution {
-			// Already exists and resolution of result we are adding is not as precise
+func (r *multiResult) dedupe() {
+	for id, serieses := range r.dedupeMap {
```
richardartoul: serieses --> seriesList or something

src/query/storage/m3/storage.go
```diff
@@ -110,9 +110,19 @@ func (s *m3storage) FetchRaw(
...
 		now        = time.Now()
 		fetches    = 0
-		result     multiFetchResult
 		wg         sync.WaitGroup
 	)
+	if len(namespaces) == 0 {
+		return nil, noop, fmt.Errorf("no namespaces configured")
+	}
+
+	pools, err := namespaces[0].Session().IteratorPools()
```
richardartoul: maybe leave a todo to find a way where we're not always ignoring pools from the other namespaces

src/query/storage/m3/multi_fetch_result.go
prateek: talked offline, appending to nil slice is kosher

src/query/storage/m3/storage.go
prateek: talked offline, will comeback for this.

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -21,117 +21,151 @@
...
-	// Need to dedupe
 	if r.dedupeMap == nil {
-		r.dedupeMap = make(map[string]multiFetchResultSeries, len(iters))
-		for idx, s := range iters {
-			r.dedupeMap[s.ID().String()] = multiFetchResultSeries{
-				idx:   idx,
-				attrs: r.dedupeFirstAttrs,
-			}
-		}
+		r.dedupeMap = make(map[string][]multiResultSeries, len(iters))
```
robskillington: Hm now we are always going to be making the dedupe map (regardless of if we only hit a single namespace)?

src/query/storage/m3/multi_fetch_result.go
robskillington: We're always going to alloc a map and an array per entry in the map now?

robskillington:
Thanks for fixing this, I should have had better tests to catch this.  

Ignore the comments I posted, I can revisit the perf, to avoid de-dupe map if we don't need it, etc.

PR #993, opened 2018-09-30
[kube] manifests + README for m3nsch

schallert:


kube/m3nsch.yaml
```diff
@@ -0,0 +1,130 @@
...
+  selector:
+    app: m3nsch
+    component: agent
+---
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: m3nsch-server
+spec:
+  replicas: 1
```
prateek: just a FYI, you can tune this - i.e. one client can kick off a job with `n` servers if you need more load.

schallert: Yeah good point. I might change this to a `StatefulSet` so that we can get stable hostnames per-pod, since currently even if we kick off > 1 agents we'd have to look up each pod name since the service does random LB

PR #997, opened 2018-10-01
[m3dbnode] add script for node bootstrapped status

schallert:
Adds a script that exits non-zero when a node hasn't finished
bootstrapping.


scripts/m3dbnode_bootstrapped.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+
+# Exits with code 0 if either:
+# - There is no placement
+# - The current host is not in the placement
+# - The current is in the placement AND all it's shards are AVAILABLE
```
richardartoul: The current host

scripts/m3dbnode_bootstrapped.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# Curl exits 22 for 400+ status code. Note this leaves us vulnerable to saying
+# bootstrapped if our script makes a bad request and must use caution when
+# modifying the script or the coordinator placement endpoint.
+if [[ "$RES" -eq 22 ]]; then
+  echo "Received 400"
+  exit 0
+fi
+
+# TODO(schallert): should a leaving node be considered bootstrapped?
```
richardartoul: "Leaving" status is on a shard-by-shard basis and doesn't have much relation to whether the node is bootstrapped or not

scripts/m3dbnode_bootstrapped.sh
```diff
@@ -0,0 +1,51 @@
...
+TOTAL=$(jq ".placement.instances | .[\"${HOSTNAME}\"] | .shards | length" < "$TMPFILE")
+
+if [[ "$TOTAL" -eq 0 ]]; then
+  # Host is not in placement.
+  echo "$HOSTNAME not found in placement"
+  exit 0
+fi
+
+AVAILABLE=$(jq ".placement.instances | .[\"${HOSTNAME}\"] | .shards | map(select(.state == \"AVAILABLE\")) | length" < "$TMPFILE")
```
richardartoul: I dont think this actually tells you whether the node is bootstrapped or not

richardartoul: Spoke offline, I think we want to get rid of this heuristic entirely and just hit the health endpoint

scripts/m3dbnode_bootstrapped.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
```
richardartoul: This looks better, but I think the issue remains that if their is a placement, but no namespace, the health endpoint will not be available.

@robskillington When you make your changes to turn on the health endpoint sooner, will you be able to make it such that it returns `true` when their is no namespace, but then reverts back to false when the namespace is added and it needs to bootstrap it?

Also, in the case that their is no topology, do we still want to return `true` from the health endpoint? What if their is a topology, but the M3DB node just hasn't received it yet? Seems a little racey.

Thoughts at @robskillington

PR #1000, opened 2018-10-01
Improve performance of generatings IDs from tags in M3Coordinator

richardartoul:
screen shot 2018-10-01 at 1 44 03 pm
screen shot 2018-10-01 at 1 44 09 pm


src/query/models/tag.go
```diff
@@ -145,17 +146,47 @@ func (m Matchers) ToTags() (Tags, error) {
...
 	}
 
-	return string(b)
+	return b
+}
+
+// IDLen returns the length of the ID that would be
+// generated from the tags.
+func (t Tags) IDLen() int {
+	idLen := 0
```
arnikola: nit: might be a little better to have `idLen := len(t) * 2`

richardartoul: I think thats wrong though, because I want it to be the length of the generated ID, not twice the number of tags

src/query/models/tag.go
```diff
@@ -145,17 +146,47 @@ func (m Matchers) ToTags() (Tags, error) {
...
+		strBuilder.WriteString(tag.Value)
+		strBuilder.WriteByte(sep)
+	}
+
+	return strBuilder.String()
+}
+
+// WriteBytesID writes out the ID representation
+// of the tags into the provided buffer.
+func (t Tags) WriteBytesID(b []byte) []byte {
```
arnikola: nit: why does this one need a buffer, rather than creating one like the stringID function?

src/query/models/tag.go
arnikola: Yeah, sorry I mean initialize it to be twice the length of `t` to account for `eq` and `sep` rather than adding 2 each time through the loop

src/query/models/tag.go
richardartoul: The idea is that the caller controls the buffer which enables us to add pooling later without making an API change

nikunjgit: yeah I like the caller providing a buffer

src/query/models/tag.go
```diff
@@ -145,17 +146,46 @@ func (m Matchers) ToTags() (Tags, error) {
 	return Normalize(tags), nil
 }
 
-// ID returns a string representation of the tags
-func (t Tags) ID() string {
-	b := make([]byte, 0, len(t))
+// StringID returns a string representation of the tags
+func (t Tags) StringID() string {
```
prateek: (unrelated to the pr)

Tags keys/values are allowed to have '=' and ','; this encoding isn't invertible - i.e. given a string representation of some tags, you can't guarantee that it comes from a single value.  

How is this function used?

prateek: why do you need to change the method name?

src/query/models/tag.go
```diff
@@ -145,17 +146,47 @@ func (m Matchers) ToTags() (Tags, error) {
...
+	var (
+		idLen      = t.IDLen()
+		strBuilder = strings.Builder{}
```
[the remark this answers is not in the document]
prateek: 1.9 is over a year old at this point, 1.10 is over 6 months old. 1.11 just came out. 

Any reason you want to support 1.9?

src/query/models/tag.go
richardartoul: because I added a BytesID(), so seems good to distinguish the two methods so the caller is forced to think about what they are doing

richardartoul: This determines what the ID of the metric will be in M3DB

src/query/models/tag.go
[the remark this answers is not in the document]
robskillington: Probably ok to drop support, since we're only supporting built binaries really.  We would like most people to use all this software as docker images or as binaries built from the tagged releases.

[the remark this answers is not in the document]
nikunjgit: Nah, i agree that we should drop 1.9 support. Even prom only supports 1.10+ https://github.com/prometheus/prometheus/pull/3856. We should call it out in our README though.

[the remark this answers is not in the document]
robskillington: Ah yup, @richardartoul want to update the README.md (in a minimal way, we can update formatting/etc of it later) that we only support 1.10 at this time?

src/query/models/tag.go
prateek: fair. The goish (go-y?) way of doing this is to call WritesBytesID something like `MarshalTo()/MarshalInto()` 

maybe call the two methods `ID()` and `IDMarshalTo()`?

prateek: we should talk about this offline.

src/query/storage/m3/storage.go
```diff
@@ -269,11 +269,15 @@ func (s *m3storage) Write(
 		return errors.ErrNilWriteQuery
 	}
 
-	id := query.Tags.ID()
-	// TODO: Consider caching id -> identID
-	identID := ident.StringID(id)
+	// TODO: Consider caching id -> identID (requires setting NoFinalize()).
+	var (
+		// TODO: Pool this once an ident pool is setup.
```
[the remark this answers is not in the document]
richardartoul: Yeah although if we do that its gonna reintroduce the issue of copying which is annoying...Should we make the session API assume ownership of the ID? then it won't have to do a copy internally

src/query/storage/m3/storage.go
```diff
@@ -287,7 +291,7 @@ func (s *m3storage) Write(
 		datapoint := datapoint
 		wg.Add(1)
 		s.writeWorkerPool.Go(func() {
-			if err := s.writeSingle(ctx, query, datapoint, identID, tagIter); err != nil {
+			if err := s.writeSingle(ctx, query, datapoint, id, tagIter); err != nil {
```
robskillington: Could we add a special case here for when `len(query.Datapoints) == 1` to just `return s.writeSingle(...)` as discussed?

richardartoul: done

PR #1003, opened 2018-10-02
Significantly improve performance of FetchResultToPromResult and helper functions

richardartoul:
screen shot 2018-10-02 at 9 57 09 am
screen shot 2018-10-02 at 9 57 13 am

Before
```
BenchmarkFetchResultToPromResult-8   	      30	  34556216 ns/op	25773633 B/op	 1014479 allocs/op
```

After
```
BenchmarkFetchResultToPromResult-8   	     100	  10563444 ns/op	25368543 B/op	    4443 allocs/op
```

### DIFF
--- a/src/query/storage/converter.go
+++ b/src/query/storage/converter.go
@@ -135,51 +135,80 @@ func TimestampToTime(timestampMS int64) time.Time {
 
 // TimeToTimestamp converts a time.Time to prometheus timestamp
 func TimeToTimestamp(timestamp time.Time) int64 {
+	// Significantly faster than time.Truncate()
 	return timestamp.UnixNano() / int64(time.Millisecond)
 }
 
 // FetchResultToPromResult converts fetch results from M3 to Prometheus result
 func FetchResultToPromResult(result *FetchResult) *prompb.QueryResult {
-	timeseries := make([]*prompb.TimeSeries, 0)
-
+	// Perform bulk allocation upfront then convert to pointers afterwards
+	// to reduce total number of allocations. See BenchmarkFetchResultToPromResult
+	// if modifying.
+	timeseries := make([]prompb.TimeSeries, 0, len(result.SeriesList))
 	for _, series := range result.SeriesList {
 		promTs := SeriesToPromTS(series)
 		timeseries = append(timeseries, promTs)
 	}
 
+	timeSeriesPointers := make([]*prompb.TimeSeries, 0, len(result.SeriesList))
+	for i := range timeseries {
+		timeSeriesPointers = append(timeSeriesPointers, &timeseries[i])
+	}
+
 	return &prompb.QueryResult{
-		Timeseries: timeseries,
+		Timeseries: timeSeriesPointers,
 	}
 }
 
 // SeriesToPromTS converts a series to prometheus timeseries
-func SeriesToPromTS(series *ts.Series) *prompb.TimeSeries {
+func SeriesToPromTS(series *ts.Series) prompb.TimeSeries {
 	labels := TagsToPromLabels(series.Tags)
 	samples := SeriesToPromSamples(series)
-	return &prompb.TimeSeries{Labels: labels, Samples: samples}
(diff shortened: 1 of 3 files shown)


src/query/storage/converter_test.go
```diff
@@ -229,3 +230,45 @@ func TestPromReadQueryToM3(t *testing.T) {
 		})
 	}
 }
+
+var (
+	benchResult *prompb.QueryResult
+)
+
+func BenchmarkFetchResultToPromResult(b *testing.B) {
```
arnikola: Might be useful to paste the benchmark results in a comment here just for easier comparisons in the future?

src/query/storage/converter.go
```diff
@@ -135,51 +135,80 @@ func TimestampToTime(timestampMS int64) time.Time {
...
 }
 
 // FetchResultToPromResult converts fetch results from M3 to Prometheus result
 func FetchResultToPromResult(result *FetchResult) *prompb.QueryResult {
-	timeseries := make([]*prompb.TimeSeries, 0)
-
+	// Perform bulk allocation upfront then convert to pointers afterwards
+	// to reduce total number of allocations. See BenchmarkFetchResultToPromResult
+	// if modifying.
+	timeseries := make([]prompb.TimeSeries, 0, len(result.SeriesList))
```
arnikola: Nice pattern, we can probably make use fo this approach elsewhere too

src/query/storage/converter.go
```diff
@@ -160,14 +163,14 @@ func FetchResultToPromResult(result *FetchResult) *prompb.QueryResult {
...
-// SeriesToPromTS converts a series to prometheus timeseries
+// SeriesToPromTS converts a series to prometheus timeseries.
 func SeriesToPromTS(series *ts.Series) prompb.TimeSeries {
 	labels := TagsToPromLabels(series.Tags)
 	samples := SeriesToPromSamples(series)
 	return prompb.TimeSeries{Labels: labels, Samples: samples}
 }
 
-// TagsToPromLabels converts tags to prometheus labels
+// TagsToPromLabels converts tags to prometheus labels.TagsToPromLabels.
```
arnikola: Looks like a find+replace issue?

src/query/storage/converter.go
```diff
@@ -135,51 +135,83 @@ func TimestampToTime(timestampMS int64) time.Time {
...
 func SeriesToPromSamples(series *ts.Series) []*prompb.Sample {
-	samples := make([]*prompb.Sample, series.Len())
-	for i := 0; i < series.Len(); i++ {
-		samples[i] = &prompb.Sample{
-			Timestamp: series.Values().DatapointAt(i).Timestamp.UnixNano() / int64(time.Millisecond),
-			Value:     series.Values().ValueAt(i),
-		}
+	var (
+		seriesLen  = series.Len()
+		values     = series.Values()
```
arnikola: nit: we only use these in one place so probably prefer to drop the var for it and just use the accessor?

src/query/storage/converter.go
```diff
@@ -135,51 +135,83 @@ func TimestampToTime(timestampMS int64) time.Time {
...
+		samples = make([]prompb.Sample, 0, seriesLen)
+	)
+	for _, dp := range datapoints {
+		samples = append(samples, prompb.Sample{
+			Timestamp: TimeToTimestamp(dp.Timestamp),
+			Value:     dp.Value,
+		})
+	}
+
+	samplesPointers := make([]*prompb.Sample, 0, len(samples))
```
arnikola: nit; alternatively can use seriesLen here too

src/query/storage/converter.go
richardartoul: nah because then it gets called on every iteration of the loop. Moving these out into vars actually made a significant impact (20%ish percent improvement or so)

src/query/storage/converter.go
richardartoul: just seemed safer to use the length of the thing I'm actually iterating through

PR #1004, opened 2018-10-02
[coordinator] Treat Prometheus TagName/Value as []byte instead of String

arnikola:
Change tag names and values from strings to []bytes


src/query/functions/binary/and_test.go
```diff
@@ -91,5 +92,6 @@ func TestAndWithSomeValues(t *testing.T) {
 	expected := values1
 	expected[0][1] = math.NaN()
 	expected[1][0] = math.NaN()
+	fmt.Println(expected, sink.Values)
```
prateek: need to remove this?

arnikola: Oops yeah made a mental note to clean these but forgot it unfortunately

src/cmd/services/m3coordinator/downsample/flush_handler.go
```diff
@@ -120,11 +120,11 @@ func (w *downsamplerFlushHandlerWriter) Write(
 		tags := make(models.Tags, 0, expected+1)
```
robskillington: nit: This should actually just be `expected` not `expected+1` since we increment `expected` above if there is an aggregation tag.

src/query/api/v1/handler/json/write.go
```diff
@@ -90,8 +90,13 @@ func newStorageWriteQuery(req *WriteQuery) (*storage.WriteQuery, error) {
 		return nil, err
 	}
 
+	tags := models.Tags{}
```
robskillington: nit: Could preallocate the map size here, `make(models.Tags, len(req.Tags))`

src/query/models/tag.go
```diff
@@ -85,21 +90,21 @@ func (m MatchType) String() string {
 // Matcher models the matching of a label.
 type Matcher struct {
 	Type  MatchType `json:"type"`
-	Name  string    `json:"name"`
-	Value string    `json:"value"`
+	Name  []byte    `json:"name"`
+	Value []byte    `json:"value"`
```
robskillington: Just a warning, the JSON serializer will now encode these as base64:
https://play.golang.org/p/T3Xa8OGyWML

robskillington: Do matchers also need to be bytes?  These only come through for reads yeah?

src/query/models/tag.go
```diff
@@ -236,54 +242,24 @@ func (t Tags) IDWithKeys(includeKeys ...string) uint64 {
...
-
-// StringMap returns a string map of the tags.
-func (t Tags) StringMap() map[string]string {
-	m := make(map[string]string, len(t))
-	for _, tag := range t {
-		m[tag.Name] = tag.Value
-	}
-
-	return m
+	return []byte{}, false
```
robskillington: `return nil, false` is slightly more optimal (no allocs)

src/query/storage/index.go
```diff
@@ -66,8 +67,11 @@ func FromIdentTagIteratorToTags(identTags ident.TagIterator) (models.Tags, error
 func TagsToIdentTagIterator(tags models.Tags) ident.TagIterator {
 	identTags := make([]ident.Tag, 0, len(tags))
 
+	//TODO get a checked bytes pool here instead
 	for _, t := range tags {
-		identTags = append(identTags, ident.StringTag(t.Name, t.Value))
+		n := checked.NewBytes(t.Name, checked.NewBytesOptions())
+		v := checked.NewBytes(t.Value, checked.NewBytesOptions())
+		identTags = append(identTags, ident.BinaryTag(n, v))
```
robskillington: For now, perhaps we should replace with just a zero copy version (since models.Tags should be coming from the wire and won't be mutated?):
```go
identTags := make([]ident.Tag, 0, len(tags))
for _, t := range tags {
  identTags = append(identTags, ident.Tag{Name: ident.BytesID(t.Name), Value: ident.BytesID(t.Value)})
}
return ident.NewTagsIterator(ident.NewTags(identTags...)))
```

src/query/api/v1/handler/json/write.go
arnikola: Yeah good call, wasn't overly concerned with the perf here since it's kind of a test endpoint but will make the change :)

src/query/models/tag.go
arnikola: Yeah, these are only needed for reads, but figured may as well convert them to bytes; would the base64 be a problem? We only really use it for sending to the `session` or encode them to `[]byte` anyway on remote fetches

src/query/models/tag.go
arnikola: Good call

src/query/storage/index.go
arnikola: Oh nice, don't know why I missed the `ident.Bytes`; was probably looking for something like `ident.ByteTag` or something :p

src/query/storage/index.go
```diff
@@ -64,10 +64,13 @@ func FromIdentTagIteratorToTags(identTags ident.TagIterator) (models.Tags, error
 
 // TagsToIdentTagIterator converts coordinator tags to ident tags
 func TagsToIdentTagIterator(tags models.Tags) ident.TagIterator {
+	//TODO get a checked bytes pool here instead
```
robskillington: nit: Maybe update this to `// TODO get a Tags and TagsIterator from a ident.Pool here instead of creating both

PR #1005, opened 2018-10-03
[coordinator] Change coordinator read worker pools to use PooledWorkerPools instead

arnikola:


src/cmd/services/m3query/config/config.go
```diff
@@ -65,14 +65,10 @@ type Configuration struct {
 	// Backend is the backend store for query service. We currently support grpc and m3db (default).
 	Backend BackendStorageType `yaml:"backend"`
 
-	// DecompressWorkerPoolCount is the number of decompression worker pools.
-	DecompressWorkerPoolCount int `yaml:"workerPoolCount"`
+	// ReadWorkerPoolSize is the size of the worker pool for read requests.
```
prateek: more a fyi than anything else: this changes a publicly accessible field. I think it's fine for now considering i don't think we've made a release exposing this in configs and put it in docs at all. But as we add more features, we have to make them in a b/w compatible way to cause the least friction for users.

src/query/pools/query_pools.go
```diff
@@ -51,46 +48,31 @@ const (
...
-		workerPool.Init()
-		return workerPool
-	})
-
 	writePoolSize := cfg.WriteWorkerPoolSize
 	if writePoolSize == 0 {
 		writePoolSize = defaultWorkerPoolCount
 	}
 
+	readWorkerPool.Init()
```
prateek: nit: move this to be after the read pool err check. it's out of place here

src/query/server/server.go
```diff
@@ -512,14 +512,14 @@ func newStorages(
...
-	workerPool pool.ObjectPool,
+	readWorkerPool xsync.PooledWorkerPool,
 ) (storage.Storage, bool, error) {
 	if cfg.RPC == nil {
 		return nil, false, nil
 	}
 
 	if remotes := cfg.RPC.RemoteListenAddresses; len(remotes) > 0 {
-		client, err := tsdbRemote.NewGrpcClient(remotes, poolWrapper, workerPool)
+		client, err := tsdbRemote.NewGrpcClient(remotes, poolWrapper, readWorkerPool)
```
prateek: nit: naming `NewGRPCClient`

src/cmd/services/m3query/config/config.go
arnikola: Yeah, agreed. This is not a field that would currently be in use, and is currently undocumented, so I'm ok with being a bit fast and loose with changing it at the moment 😛

src/query/tsdb/remote/compressed_codecs.go
```diff
@@ -113,7 +113,9 @@ func compressedTagsFromTagIterator(tagIter ident.TagIterator, encoderPool serial
 		return nil, fmt.Errorf("no refs available to data")
 	}
 
-	return data.Bytes(), nil
+	db := data.Bytes()
+	// Need to copy the encoded bytes to a buffer as the encoder keeps a reference to them
```
prateek: wanna use pooled bytes rather than alloc here?

TBH: this is one of the reasons we end up with so many wrapper objects in the m3db codebase - because we need a place to track what pool an object should be returned to once we're done with it (can't use just a slice). 

Will need to use [GRPC interceptors](https://github.com/grpc/grpc-go/blob/master/interceptor.go#L77) to release back to the pool safely

src/query/pools/query_pools.go
```diff
@@ -50,47 +47,29 @@ const (
...
-		return workerPool
-	})
-
-	writePoolSize := cfg.WriteWorkerPoolSize
-	if writePoolSize == 0 {
-		writePoolSize = defaultWorkerPoolCount
+	readPoolPolicy, writePoolPolicy xconfig.WorkerPoolPolicy,
+) (xsync.PooledWorkerPool, xsync.PooledWorkerPool, error) {
+	opts, readPoolSize := readPoolPolicy.Options()
+	instrumentOptions.SetMetricsScope(scope.SubScope("read-worker-pool"))
```
richardartoul: instrumentOptions is immutable so you need to do:

```golang
readPoolIOpts = instrumentOptions.SetMetricsScope(scope.SubScope("read-worker-pool"))
opts = opts.SetInstrumentOptions(readPoolIOpts)
```

src/query/pools/query_pools.go
```diff
@@ -50,47 +47,29 @@ const (
...
+	opts = opts.SetInstrumentOptions(instrumentOptions)
+	readWorkerPool, err := xsync.NewPooledWorkerPool(readPoolSize, opts)
+	if err != nil {
+		return nil, nil, err
 	}
 
-	writeWorkerPool, err := xsync.NewPooledWorkerPool(writePoolSize, xsync.NewPooledWorkerPoolOptions())
+	readWorkerPool.Init()
+	opts, writePoolSize := writePoolPolicy.Options()
+	instrumentOptions.SetMetricsScope(scope.SubScope("write-worker-pool"))
```
richardartoul: same comment

src/query/pools/query_pools.go
arnikola: Nice catch, just inlined it instead :)

src/query/tsdb/remote/compressed_codecs.go
arnikola: This was more of an afterthought after seeing issues caused by this on the internal side; this is just fixing it as is, rather than the correct approach... Added [an issue](https://github.com/m3db/m3/issues/1015) to track this properly since it's out of scope of this PR and probably a decent chunk of work (as soon as we start intercepting GRPC calls, might as well add pools for all the proto slices we're building, and would prefer to keep that to a dedicated PR)

prateek: fair enough, mind adding a `// TODO(arnikola): pool this when implementing https://github.com/m3db/m3/issues/1015` as a reminder.

arnikola: No worries 👍

src/query/server/server_test.go
```diff
@@ -67,9 +67,18 @@ clusters:
         type: unaggregated
         retention: 48h
 
-workerPoolCount: 20
-decompressWorkerPoolSize: 20
-writeWorkerPoolSize: 20
+
+readWorkerPoolPolicy:
+  grow: true
+  size: 100
```
richardartoul: size has to be at least as large as shards: https://github.com/m3db/m3x/blob/master/sync/pooled_worker_pool.go#L54

Also this configuration seems a little weird, you'll have a pool size of 100 by default that will expand on demand, but each goroutine will have a 30% probability of killing itself.  I'd probably increase the size (should at least be the same as shards), maybe reduce shards a bit, and then decrease the kill probability to something like 0.01

richardartoul: Also note that with `grow: true` parallelism is unbounded

arnikola: Yeah, these are just random numbers since it's more just a test to ensure the config gets read correctly

PR #1006, opened 2018-10-03
[scripts] bootstrapped: no set -x

schallert:
Seeing verbose health output all over the place. Adding more debug
signal in regular output instead.

### DIFF
--- a/scripts/m3dbnode_bootstrapped.sh
+++ b/scripts/m3dbnode_bootstrapped.sh
@@ -8,7 +8,7 @@
 # 4. If 1-3 true, check if host reports bootstrapped and exit healthy if so
 #    (fail otherwise).
 
-set -xo pipefail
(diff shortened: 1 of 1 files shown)


prateek:
LGTM even

scripts/m3dbnode_bootstrapped.sh
```diff
@@ -34,7 +34,7 @@ RES=$?
 # bootstrapped if our script makes a bad request and must use caution when
 # modifying the script or the coordinator placement endpoint.
 if [ "$RES" -eq 22 ]; then
-  echo "Received 4xx from coordinator"
+  echo "Received 4xx from coordinator $COORD_PLACEMENT_ENDPOINT"
```
prateek: might want to include contents of `COORD_TMPFILE` here

PR #1008, opened 2018-10-04
[coordinator] placement: adds "safe" by default

schallert:
This PR modifies the `placement/add` handler such that adds are "safe"
by default. That is, we know that operationally it's unsafe to add an
instance to the placement while others are bootstrapping. This encodes
that in the add handler by ensuring that adds will only succeed if there
are no instances initializing, unless the user has specified `force` in
the request.


src/query/api/v1/handler/placement/add.go
```diff
@@ -106,10 +122,36 @@ func (h *AddHandler) Add(
...
 	if err != nil {
 		return nil, err
 	}
 
 	return newPlacement, nil
 }
+
+func validateNoInitializing(p placement.Placement) error {
+	for _, inst := range p.Instances() {
+		total := inst.Shards().NumShards()
```
cw9: We have some similar functions in m3cluster [here](https://github.com/m3db/m3cluster/blob/master/placement/placement.go#L558), maybe move this logic there as well?

src/query/generated/proto/admin/placement.proto
```diff
@@ -17,4 +17,7 @@ message PlacementGetResponse {
 
 message PlacementAddRequest {
   repeated placementpb.Instance instances = 1;
+  // By default add requests will only succeed if no instances in the placement
+  // are in INITIALIZING state. force overrides that.
```
prateek: need to update this comment now

src/query/api/v1/handler/placement/add.go
```diff
@@ -106,10 +119,32 @@ func (h *AddHandler) Add(
...
 	newPlacement, _, err := service.AddInstances(instances)
 	if err != nil {
 		return nil, err
 	}
 
 	return newPlacement, nil
 }
+
+func validateAllAvailable(p placement.Placement) error {
+	for _, inst := range p.Instances() {
```
cw9: nit: use multi err here to print all the invalid instances

PR #1012, opened 2018-10-04
Use cluster topology map in bootstrap process instead of session topology map

### LINKED ISSUE #1011: Race in topology retrieval during bootstrap between CommitLog and DB
During the bootstrap process, we rely on the [Client's topology view](https://github.com/m3db/m3/blob/2c2f509ed217f1b0dec7f289626ee264606a8351/src/dbnode/storage/bootstrap/process.go#L108-L125) to make decisions about the shards to boostrap. This is racy because it can have received an update before the corresponding update on the storage.Database end. 

We should instead rely on the Database's topology view to make consistent decisions about the shards to bootstrap.

richardartoul:
Fixes #1011

### DIFF
--- a/src/dbnode/client/types.go
+++ b/src/dbnode/client/types.go
@@ -148,20 +148,23 @@ type PeerBlocksIter interface {
 type AdminSession interface {
 	Session
 
-	// Origin returns the host that initiated the session
+	// Origin returns the host that initiated the session.
 	Origin() topology.Host
 
-	// Replicas returns the replication factor
+	// Replicas returns the replication factor.
 	Replicas() int
 
-	// TopologyMap returns the current topology map
+	// TopologyMap returns the current topology map. Note that the session
+	// has a separate topology watch than the database itself, so the two
+	// values can be out of sync and this method should not be relied upon
+	// if the current view of the topology as seen by the database is required.
 	TopologyMap() (topology.Map, error)
 
-	// Truncate will truncate the namespace for a given shard
+	// Truncate will truncate the namespace for a given shard.
 	Truncate(namespace ident.ID) (int64, error)
 
 	// FetchBootstrapBlocksFromPeers will fetch the most fulfilled block
-	// for each series using the runtime configurable bootstrap level consistency
+	// for each series using the runtime configurable bootstrap level consistency.
 	FetchBootstrapBlocksFromPeers(
 		namespace namespace.Metadata,
 		shard uint32,
@@ -171,7 +174,7 @@ type AdminSession interface {
 	) (result.ShardResult, error)
 
 	// FetchBootstrapBlocksMetadataFromPeers will fetch the blocks metadata from
-	// available peers using the runtime configurable bootstrap level consistency
+	// available peers using the runtime configurable bootstrap level consistency.
 	FetchBootstrapBlocksMetadataFromPeers(
 		namespace ident.ID,
 		shard uint32,
@@ -181,7 +184,7 @@ type AdminSession interface {
 	) (PeerBlockMetadataIter, error)
 
 	// FetchBlocksMetadataFromPeers will fetch the blocks metadata from
-	// available peers
+	// available peers.
 	FetchBlocksMetadataFromPeers(
 		namespace ident.ID,
 		shard uint32,
@@ -192,7 +195,7 @@ type AdminSession interface {
 	) (PeerBlockMetadataIter, error)
 
 	// FetchBlocksFromPeers will fetch the required blocks from the
-	// peers specified
+	// peers specified.
 	FetchBlocksFromPeers(
 		namespace namespace.Metadata,
 		shard uint32,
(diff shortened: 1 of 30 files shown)


src/dbnode/integration/integration.go
```diff
@@ -295,6 +318,14 @@ func newDefaultBootstrappableTestSetups(
 	}
 }
 
+type dynamicTopoMapProviderBullshit struct {
+	setup *testSetup
+}
+
+func (d *dynamicTopoMapProviderBullshit) TopologyMap() (topology.Map, error) {
```
prateek: lol this can just be a method on `*testSetup` no?

richardartoul: ...yes that would better. That was written in anger, will make it a method

src/dbnode/storage/cluster/database.go
```diff
@@ -131,6 +131,10 @@ func (d *clusterDB) Topology() topology.Topology {
 	return d.topo
 }
 
+func (d *clusterDB) TopologyMap() (topology.Map, error) {
+	return d.topo.Get(), nil
```
prateek: This can still be racy, because two watches can get the update at different times. Instead of this, create a struct field in clusterDB which tracks topology.Map, and is updated based on the value as viewed by the clusterDB.

src/dbnode/storage/cluster/types.go
```diff
@@ -29,5 +29,9 @@ import (
 type Database interface {
 	storage.Database
 
+	// Topology returns the topology.
 	Topology() topology.Topology
+
+	// TopologyMap returns the topology map.
+	TopologyMap() (topology.Map, error)
```
prateek: instead of this, wanna embed the topology provider interface?

richardartoul: sure

src/dbnode/storage/bootstrap/types.go
```diff
@@ -23,14 +23,20 @@ package bootstrap
...
 	"github.com/m3db/m3/src/dbnode/persist"
 	"github.com/m3db/m3/src/dbnode/storage/bootstrap/result"
 	"github.com/m3db/m3/src/dbnode/storage/namespace"
 	"github.com/m3db/m3/src/dbnode/topology"
 	xtime "github.com/m3db/m3x/time"
 )
 
+// TopologyMapProvider is an interface that can provide
+// a topology map.
+type TopologyMapProvider interface {
```
prateek: nit: move to `topology` package. Can call it `topology.MapProvider` or `topology.MapResolver`

richardartoul: Seemed nice to have it close to where it is used, but don't have strong feelings about it

prateek: i just don't like the import dependency leaving this here creates. `storage/cluster` and and `storage/bootstrap` both need this type. Better place for it is `topology` so both can depend on it, rather than have `cluster` depend on `bootstrap` as it would be if you left here

src/dbnode/client/types.go
```diff
@@ -181,7 +184,7 @@ type AdminSession interface {
 	) (PeerBlockMetadataIter, error)
 
 	// FetchBlocksMetadataFromPeers will fetch the blocks metadata from
-	// available peers
+	// available peers.
```
prateek: +1 for comment fixes

src/dbnode/storage/cluster/database.go
richardartoul: spoke offline, agreed for now that its ok if the bootstrapp process receives a topology snapshot that is *newer* than the one that triggered the bootstrap as long as its not older.

#1013

src/dbnode/storage/bootstrap/types.go
richardartoul: fair, fixed

src/dbnode/client/types.go
```diff
@@ -148,20 +148,23 @@ type PeerBlocksIter interface {
...
-	// Origin returns the host that initiated the session
+	// Origin returns the host that initiated the session.
 	Origin() topology.Host
 
-	// Replicas returns the replication factor
+	// Replicas returns the replication factor.
 	Replicas() int
 
-	// TopologyMap returns the current topology map
+	// TopologyMap returns the current topology map. Note that the session
```
prateek: i think you can remove this method now. Unused afaict

src/dbnode/integration/setup.go
```diff
@@ -605,6 +608,19 @@ func (ts *testSetup) maybeResetClients() error {
 	return nil
 }
 
+// Implements topology.MapProvider, and makes sure that the topology
```
prateek: maybe add a var decl to enforce longer term. i.e.

```
var _ topology.MapProvider = &testSetup{}
```

src/dbnode/client/types.go
richardartoul: one of our tests uses it

prateek: fair enough, i think we should move the tests to use the t-channel health/bootstrapped endpoint. opened https://github.com/m3db/m3/issues/1016 to followup

Issue #1001, opened 2018-10-01
Generating IDs from M3Coordinator should be made invertible


richardartoul:
Currently generating IDs from tags in M3Coordinator does no escaping, which means that in theory, multiple different groups of tags can map to the same ID if they contain the characters "=" or ";" within their tag names / values.

We should change how we generate IDs from tags such that every ID can only be generated from a single unique set of tags

richardartoul:
@arnikola can we close this

arnikola:
Yeah, sounds good. Closing

PR #1014, opened 2018-10-04
[coordinator] Add metric name as a configuration option

arnikola:
Allows for non-hardcoded `metricName` names; useful for non-Prometheus sources, or aggregators that sanitize default `__name__` name to something else


src/cmd/services/m3query/config/config.go
```diff
@@ -106,3 +110,27 @@ type RPCConfiguration struct {
 	// coordinator calls.
 	RemoteListenAddresses []string `yaml:"remoteListenAddresses"`
 }
+
+// TagOptionsConfiguration is the configuration for shared tag options
+// Currently only name, but can expand to cover deduplication settings, or other
+// relevant options.
+type TagOptionsConfiguration struct {
+	// MetricName specifies the tag name that corresponds to the metric's name tag
+	// Defaults to __name__ for Prometheus.
```
benraskin92: Should say "if left empty, defaults to `__name__`"

src/cmd/services/m3query/config/config.go
```diff
@@ -106,3 +110,27 @@ type RPCConfiguration struct {
...
+
+// TagOptionsFromConfig translates tag option configuration into tag options.
+func TagOptionsFromConfig(cfg TagOptionsConfiguration) (models.TagOptions, error) {
+	opts := models.NewTagOptions()
+	name := cfg.MetricName
+	if name != "" {
+		opts = opts.SetMetricName([]byte(name))
+	}
+
+	if err := opts.Validate(); err != nil {
```
benraskin92: I'm guessing this is for future validations?

arnikola: Yeah, using a similar pattern to how some of our other configs look, so in a similar vein to [the db client options](https://github.com/m3db/m3/blob/master/src/dbnode/client/types.go#L211)

src/query/models/matcher.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+}
+
+// NewMatcher returns a matcher object.
+func NewMatcher(t MatchType, n, v []byte) (*Matcher, error) {
+	m := &Matcher{
+		Type:  t,
+		Name:  n,
+		Value: v,
+	}
```
benraskin92: newline

src/query/models/matcher.go
```diff
@@ -0,0 +1,95 @@
...
+	if t == MatchRegexp || t == MatchNotRegexp {
+		re, err := regexp.Compile("^(?:" + string(v) + ")$")
+		if err != nil {
+			return nil, err
+		}
```
benraskin92: same here

src/query/models/matcher.go
```diff
@@ -0,0 +1,95 @@
...
+		m.re = re
+	}
```
benraskin92: and here

src/query/models/matcher.go
```diff
@@ -0,0 +1,95 @@
...
+		return bytes.Equal(s, m.Value)
+	case MatchNotEqual:
+		return !bytes.Equal(s, m.Value)
+	case MatchRegexp:
+		return m.re.MatchString(string(s))
+	case MatchNotRegexp:
+		return !m.re.MatchString(string(s))
+	}
+
+	panic("labels.Matcher.Matches: invalid match type")
```
benraskin92: Should we panic or return an error here?

src/query/models/matcher.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// ToTags converts Matchers to Tags
+// NB (braskin): this only works for exact matches
+func (m Matchers) ToTags() (Tags, error) {
+	// todo: nil is good here?
```
benraskin92: Should be fine since you're just taking the default options if `nil`

arnikola: Good catch, was going to fix this up before sending the PR but missed it. Was trying to figure out if it was ok to use default options here, or if it should be filtered through with the base TagOptions

arnikola: Looking into it, this is only called from `staticResolver.Resolve()`, which is never actually used in the code.. put up [an issue](https://github.com/m3db/m3/issues/1037) to look into it and remove it if necessary.

src/query/models/matcher.go
arnikola: This file was just refactored with few changes; looking into it further, this path is unused outside of tests; adding it to [the cleanup issue](https://github.com/m3db/m3/issues/1037)

src/query/models/tags.go
```diff
@@ -0,0 +1,250 @@
...
+	}
+
+	return b
+}
+
+// IDLen returns the length of the ID that would be
+// generated from the tags.
+func (t Tags) IDLen() int {
+	// return t.idLen
+	idLen := 2 * len(t.Tags) // account for eq and sep
```
nikunjgit: len(t.Tags) -> t.Len() ?

src/query/models/tags.go
```diff
@@ -0,0 +1,250 @@
...
+		b = append(b, sep)
+	}
+
+	h := fnv.New64a()
+	h.Write(b)
+	return h.Sum64()
+}
+
+func (t Tags) tagSubset(keys [][]byte, include bool) Tags {
+	tags := NewTags(len(t.Tags), t.Opts)
```
nikunjgit: len(t.Tags) -> t.Len() ?

src/query/models/tags.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+// TagsWithoutKeys returns only the tags which do not have the given keys.
+func (t Tags) TagsWithoutKeys(excludeKeys [][]byte) Tags {
+	return t.tagSubset(excludeKeys, false)
+}
+
+// IDWithKeys returns a string representation of the tags only including the given keys.
+func (t Tags) IDWithKeys(includeKeys ...[]byte) uint64 {
+	b := make([]byte, 0, len(t.Tags))
```
nikunjgit: len(t.Tags) -> t.Len() ?

src/query/models/tags.go
```diff
@@ -0,0 +1,250 @@
...
+	t.AddTag(Tag{Name: t.Opts.GetMetricName(), Value: value})
+}
+
+// Name gets the metric name.
+func (t Tags) Name() ([]byte, bool) {
+	return t.Get(t.Opts.GetMetricName())
+}
+
+// AddTags is used to add a list of tags and maintain sorted order.
+func (t *Tags) AddTags(tags []Tag) Tags {
```
nikunjgit: can we have all non pointer receivers  (or all pointers if really needed) ?

src/query/models/tags.go
```diff
@@ -0,0 +1,250 @@
...
+	t.Tags = append(t.Tags, tags...)
+	return t.Normalize()
+}
+
+// Add is used to add two tag structures and maintain sorted order.
+func (t *Tags) Add(other Tags) Tags {
```
nikunjgit: can we have all non pointer receivers  (or all pointers if really needed) ?

src/query/models/tags.go
```diff
@@ -0,0 +1,250 @@
...
+	clonedTags := make([]Tag, len(t.Tags))
+	copy(clonedTags, t.Tags)
+	return &Tags{
+		Tags: clonedTags,
+		Opts: t.Opts,
+	}
+}
+
+// AddTag is used to add a single tag and maintain sorted order.
+func (t *Tags) AddTag(tag Tag) Tags {
```
nikunjgit: can we have all non pointer receivers  (or all pointers if really needed) ?

src/query/models/types.go
```diff
@@ -0,0 +1,88 @@
...
+type Matcher struct {
+	Type  MatchType `json:"type"`
+	Name  []byte    `json:"name"`
+	Value []byte    `json:"value"`
+
+	re *regexp.Regexp
+}
+
+// Matchers is a list of individual matchers.
+type Matchers []*Matcher
```
nikunjgit: curious about whether we need a pointer slice now ? I'm asking since you're refactoring everything

src/query/models/types.go
```diff
@@ -0,0 +1,88 @@
...
+
+// Metric is the individual metric that gets returned from the search endpoint.
+type Metric struct {
+	ID   string
+	Tags Tags
+}
+
+// Metrics is a list of individual metrics.
+type Metrics []*Metric
```
nikunjgit: same here, do we need a pointer slice ?

src/query/models/tags.go
arnikola: I like the idea of not having to explicitly have stuff like `tags = tags.AddTags(fooTags)`, and rather just have `tags.AddTags(fooTags)`, with a similar approach for other add methods,

src/query/models/tags.go
```diff
@@ -0,0 +1,250 @@
...
+		b = append(b, sep)
+	}
+
+	return b
+}
+
+// IDLen returns the length of the ID that would be
+// generated from the tags.
+func (t Tags) IDLen() int {
+	// return t.idLen
```
benraskin92: Delete

src/query/models/tags.go
nikunjgit: it's error prone to have both pointer and non-pointer receivers for the same type. Can you pick one ?

nikunjgit: Also, it's nice to have property of Tags being immutable. Maybe you can have a builder when you need to modify and tagBuilder.Build() can normalize the values

src/cmd/services/m3coordinator/downsample/flush_handler.go
```diff
@@ -67,6 +68,7 @@ func newDownsamplerFlushHandlerMetrics(
 }
 
 func newDownsamplerFlushHandler(
+	tagOptions models.TagOptions,
```
robskillington: nit: Want to make this the second last argument?  We usually tend to put the "options" like args towards the end of the method calls.

src/query/models/matcher.go
```diff
@@ -0,0 +1,100 @@
...
+	tagOptions TagOptions,
+) (Tags, error) {
+	// todo: nil is good here?
+	tags := NewTags(len(m), tagOptions)
+	for _, v := range m {
+		if v.Type != MatchEqual {
+			return Tags{}, fmt.Errorf("illegal match type, got %v, but expecting: %v", v.Type, MatchEqual)
+		}
+
+		tags = tags.AddTag(Tag{Name: v.Name, Value: v.Value})
```
robskillington: Should this be `Clone()` just in case cloned later?  Looks fine though tbh.

src/query/models/tags.go
```diff
@@ -0,0 +1,278 @@
...
+
+		// Skip the key
+		if found {
+			continue
+		}
+
+		b = append(b, tag.Name...)
+		b = append(b, eq)
+		b = append(b, tag.Value...)
+		b = append(b, sep)
```
robskillington: nit: In the future you don't  need to create a byte slice and then do a fnv hash of it, you can simply create a fnv and then `Write(..)` to it multiple times and take a final `.Sum64()`.

src/query/models/tags.go
```diff
@@ -0,0 +1,278 @@
...
+// IDWithKeys returns a string representation of the tags only including the given keys.
+func (t Tags) IDWithKeys(includeKeys ...[]byte) uint64 {
+	b := make([]byte, 0, t.Len())
+	for _, tag := range t.Tags {
+		for _, k := range includeKeys {
+			if bytes.Equal(tag.Name, k) {
+				b = append(b, tag.Name...)
+				b = append(b, eq)
+				b = append(b, tag.Value...)
+				b = append(b, sep)
```
robskillington: Same comment about this one.

src/query/models/types.go
```diff
@@ -0,0 +1,88 @@
...
+// TagOptions describes additional options for tags.
+type TagOptions interface {
+	// Validate validates these tag options.
+	Validate() error
+
+	// SetMetricName sets the name for the `metric name` metric.
+	SetMetricName(metricName []byte) TagOptions
+
+	// GetDefaultName gets the name for the `metric name `metric`.
+	GetMetricName() []byte
```
robskillington: nit: Usually we don't prefix the getter methods with `Get` if you notice re: other places we have Options interfaces.

src/query/policy/resolver/interface.go
```diff
@@ -36,6 +36,7 @@ type PolicyResolver interface {
 	// may still be no data retained for the range in the given storage policy.
 	Resolve(
 		ctx context.Context,
+		tagOptions models.TagOptions,
```
robskillington: nit: Perhaps move to last arg since it's an options arg and no others are?

src/query/storage/m3/storage.go
```diff
@@ -59,11 +60,13 @@ type m3storage struct {
 
 // NewStorage creates a new local m3storage instance.
 func NewStorage(
+	tagOptions models.TagOptions,
```
robskillington: nit: Move to last arg as this is an options arg and none others are?

src/query/models/matcher.go
arnikola: Still in the process of worming out if we can just ditch this whole codepath, [got an issue for it](
https://github.com/m3db/m3/issues/1037).

Good point though, will change to a Clone for now just in case we don't put a lot of priority on that and start making use of this elsewhere

src/query/models/tags.go
arnikola: Nice! I think we need to properly consider how we generate IDs in general though, since we may have problems if somehow tags have seperator characters in them, as per [this issue](https://github.com/m3db/m3/issues/1001)

We should probably get on that pretty quickly considering this is becoming a prod use case for many users :\

src/query/models/types.go
arnikola: Huh, if anything I'm more surprised that the linter passed considering the comment is `GetDefaultName...`

Will update

src/query/policy/resolver/interface.go
arnikola: sgtm

src/query/storage/m3/storage.go
arnikola: sgtm

src/query/models/tags.go
arnikola: That being said, may as well roll it up here as you suggested

arnikola: Actually, may leave it for now; might need to have a discussion to figure out the best way to handle ID generation here since it's going to be super janky either way

src/query/functions/utils/unique.go
```diff
@@ -0,0 +1,31 @@
...
+package utils
+
+// UniqueCount calculates the number of unique strings in a given slice.
+func UniqueCount(names []string) int {
+	uniqueMap := make(map[string]struct{}, len(names))
+	for _, s := range names {
+		uniqueMap[s] = struct{}{}
+	}
+
+	return len(uniqueMap)
```
benraskin92: Per Nikunj's previous suggestion, maybe just return `uniqueMap` and let the caller get `len`?

src/query/models/types.go
```diff
@@ -38,8 +38,8 @@ type TagOptions interface {
 	// SetMetricName sets the name for the `metric name` metric.
 	SetMetricName(metricName []byte) TagOptions
 
-	// GetDefaultName gets the name for the `metric name `metric`.
-	GetMetricName() []byte
+	// MetricName gets the name for the `metric name `metric`.
```
benraskin92: extra "`" in there

src/query/functions/utils/unique.go
arnikola: Could be valid; maybe have 2 methods, one which just returns len, other returns the map (or better yet, the map flattened back to a []string)

src/query/storage/converter.go
```diff
@@ -54,15 +55,30 @@ func PromWriteTSToM3(
...
-	for i, label := range labels {
-		tagList[i] = models.Tag{Name: label.Name, Value: label.Value}
+	tagList := make([]models.Tag, 0, len(labels))
+	for _, label := range labels {
+		// If this label corresponds to the Prometheus name,
+		// instead set it as the given name tag from the config file.
+		if bytes.Equal(promDefaultName, label.Name) {
+			tags = tags.SetName(label.Value)
+		} else {
+			tagList = append(tagList, models.Tag{
```
benraskin92: Since this slice is pre-allocated, anyway to index in to set the Tags?

arnikola: It gets a bit awkward since name will be added to Tags directly but not the slice, so you'd need to subtract from the length iff name tag is present otherwise you get nils

PR #1017, opened 2018-10-04
Make snapshot process snapshot all unflushed blocks

richardartoul:
- [X] - Modify snapshot behavior to snapshot all unflushed blocks within the retention period
- [X] - Update integration tests to capture new behavior

### DIFF
--- a/src/dbnode/integration/disk_flush_helpers.go
+++ b/src/dbnode/integration/disk_flush_helpers.go
@@ -209,11 +209,10 @@ func verifySnapshottedDataFiles(
 	reader, err := fs.NewReader(storageOpts.BytesPool(), fsOpts)
 	require.NoError(t, err)
 	iteratorPool := storageOpts.ReaderIteratorPool()
-	for _, ns := range testNamespaces {
-		for _, seriesList := range seriesMaps {
-			verifyForTime(
-				t, storageOpts, reader, shardSet, iteratorPool, snapshotTime,
-				ns, persist.FileSetSnapshotType, seriesList)
-		}
+	for _, seriesList := range seriesMaps {
+		verifyForTime(
+			t, storageOpts, reader, shardSet, iteratorPool, snapshotTime,
+			namespace, persist.FileSetSnapshotType, seriesList)
 	}
+
 }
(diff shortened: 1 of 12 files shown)


src/dbnode/storage/util.go
```diff
@@ -49,7 +49,7 @@ func timesInRange(startInclusive, endInclusive time.Time, windowSize time.Durati
 		return nil
 	}
 	times := make([]time.Time, 0, ni)
-	for t := endInclusive; !t.Before(startInclusive); t = t.Add(-windowSize) {
```
justinjc: Were there assumptions elsewhere in the code that relies on the time ranges always including the end time? Also fix the comment for this function.

prateek:
@richardartoul talked offline: maybe add a stat for the number of blocks being snapshot

src/dbnode/integration/disk_flush_helpers.go
```diff
@@ -46,23 +46,59 @@ var (
...
+	namespace ident.ID,
+	blockStart time.Time,
+) int {
+	latestVolumeIndex := -1
+
+	for _, shard := range shardSet.AllIDs() {
+		snapshotFiles, err := fs.SnapshotFiles(
+			filePathPrefix, namespace, shard)
+		if err != nil {
+			panic(err)
```
prateek: don't panic here, propagate error

prateek: meh just saw its only for integration tests, w/e

src/dbnode/storage/flush.go
```diff
@@ -233,6 +237,28 @@ func (m *flushManager) namespaceFlushTimes(ns databaseNamespace, curr time.Time)
...
+		// allow writes to be written into the next block. Note that "current time"
+		// here is defined as "tick start time" because all the guarantees about
+		// snapshotting are based around the tick start time, now the current time.
+		latest = curr.Add(rOpts.BufferFuture()).Truncate(blockSize)
+	)
+
+	candidateTimes := timesInRange(earliest, latest, blockSize)
+	return filterTimes(candidateTimes, func(t time.Time) bool {
+		// Snapshot anything that is unflushed.
+		return ns.NeedsFlush(t, t)
```
prateek: do you need to check snapshot enabled here

PR #1020, opened 2018-10-05
[coordinator] allow placement GET specific version

schallert:
This can be useful when debugging and replaying the placement history.

### DIFF
--- a/src/query/api/v1/handler/placement/get_test.go
+++ b/src/query/api/v1/handler/placement/get_test.go
@@ -22,6 +22,7 @@ package placement
 
 import (
 	"errors"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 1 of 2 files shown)


src/query/api/v1/handler/placement/get_test.go
```diff
@@ -113,4 +116,26 @@ func TestPlacementGetHandler(t *testing.T) {
 
 	resp = w.Result()
 	assert.Equal(t, http.StatusNotFound, resp.StatusCode)
+
+	// With bad version request
+	w = httptest.NewRecorder()
+	req = httptest.NewRequest("GET", "/placement/get?version=foo", nil)
```
justinjc: It doesn't affect the test outcome, but as it's configured now, shouldn't the request actually be a `GET` on `/placement` directly?

src/query/api/v1/handler/placement/get_test.go
```diff
@@ -113,4 +116,26 @@ func TestPlacementGetHandler(t *testing.T) {
...
+	require.NotNil(t, req)
+
+	handler.ServeHTTP(w, req)
+	resp = w.Result()
+	assert.Equal(t, http.StatusBadRequest, resp.StatusCode)
+
+	// With valid version request
+	w = httptest.NewRecorder()
+	req = httptest.NewRequest("GET", "/placement/get?version=12", nil)
```
justinjc: Here too

PR #1021, opened 2018-10-05
Add Holt Winters func

benraskin92:
No need to review, still need to clean up and add tests

### DIFF
--- a/src/query/functions/temporal/aggregation.go
+++ b/src/query/functions/temporal/aggregation.go
@@ -131,12 +131,12 @@ func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
 func makeHoltWintersFn(sf, tf float64) aggFunc {
 	return func(vals []float64) float64 {
 		var (
-			foundFirst, foundSecond bool
-			secondVal               float64
-			b                       float64
-			x, y                    float64
-			s0, s1                  float64
-			idx                     int
+			foundFirst, foundSecond         bool
+			secondVal                       float64
+			trendVal                        float64
+			scaledSmoothVal, scaledTrendVal float64
+			prev, curr                      float64
+			idx                             int
 		)
 
 		for _, val := range vals {
@@ -146,7 +146,7 @@ func makeHoltWintersFn(sf, tf float64) aggFunc {
 
 			if !foundFirst {
 				foundFirst = true
-				s1 = val
+				curr = val
 				idx++
 				continue
 			}
@@ -154,17 +154,17 @@ func makeHoltWintersFn(sf, tf float64) aggFunc {
 			if !foundSecond {
 				foundSecond = true
 				secondVal = val
-				b = secondVal - s1
+				trendVal = secondVal - curr
 			}
 
 			// scale the raw value against the smoothing factor.
-			x = sf * val
+			scaledSmoothVal = sf * val
 
 			// scale the last smoothed value with the trend at this point.
-			b = calcTrendValue(idx-1, sf, tf, s0, s1, b)
-			y = (1 - sf) * (s1 + b)
+			trendVal = calcTrendValue(idx-1, sf, tf, prev, curr, trendVal)
+			scaledTrendVal = (1 - sf) * (curr + trendVal)
 
-			s0, s1 = s1, x+y
+			prev, curr = curr, scaledSmoothVal+scaledTrendVal
 			idx++
 		}
 
@@ -173,7 +173,7 @@ func makeHoltWintersFn(sf, tf float64) aggFunc {
 			return math.NaN()
 		}
 
-		return s1
+		return curr
 	}
 }
 
--- a/src/query/functions/temporal/aggregation.go
+++ b/src/query/functions/temporal/aggregation.go
@@ -29,28 +29,30 @@ import (
 )
 
 const (
-	// AvgType calculates the average of all values in the specified interval
+	// AvgType calculates the average of all values in the specified interval.
 	AvgType = "avg_over_time"
 
-	// CountType calculates count of all values in the specified interval
+	// CountType calculates count of all values in the specified interval.
 	CountType = "count_over_time"
 
-	// MinType calculates the minimum of all values in the specified interval
+	// MinType calculates the minimum of all values in the specified interval.
 	MinType = "min_over_time"
 
-	// MaxType calculates the maximum of all values in the specified interval
+	// MaxType calculates the maximum of all values in the specified interval.
 	MaxType = "max_over_time"
 
-	// SumType calculates the sum of all values in the specified interval
+	// SumType calculates the sum of all values in the specified interval.
 	SumType = "sum_over_time"
 
-	// StdDevType calculates the standard deviation of all values in the specified interval
+	// StdDevType calculates the standard deviation of all values in the specified interval.
 	StdDevType = "stddev_over_time"
 
-	// StdVarType calculates the standard variance of all values in the specified interval
+	// StdVarType calculates the standard variance of all values in the specified interval.
 	StdVarType = "stdvar_over_time"
 
-	// HoltWintersType produces a smoothed value for time series based on the specified interval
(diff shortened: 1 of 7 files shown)


src/query/functions/temporal/aggregation.go
```diff
@@ -49,6 +49,9 @@ const (
 
 	// StdVarType calculates the standard variance of all values in the specified interval
 	StdVarType = "stdvar_over_time"
+
+	// HoltWintersType produces a smoothed value for time series based on the specified interval
```
arnikola: nit: I think we're slowly moving to having `.` at the end of comments... can probably keep it for a later PR so this doesn't get filled with crap though

src/query/functions/temporal/aggregation_test.go
```diff
@@ -128,6 +137,19 @@ var testCases = []testCase{
 			{2, 2, 2, 2, 2},
 		},
 	},
+	{
+		name:   "holt_winters",
+		opType: HoltWintersType,
```
arnikola: Since holt winters is a fair bit more complex than some of the other agg functions, might be a good idea to have a test or two that hit the function directly rather than building the blocks etc?

src/query/functions/temporal/aggregation.go
```diff
@@ -87,13 +90,109 @@ func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
...
+
 	return nil, fmt.Errorf("unknown aggregation type: %s", optype)
 }
 
+func makeHoltWintersFn(sf, tf float64) aggFunc {
+	return func(vals []float64) float64 {
+		var (
+			foundFirst, foundSecond bool
+			secondVal               float64
+			b                       float64
```
arnikola: Can we get better variable names

src/query/functions/temporal/aggregation.go
arnikola: Also, could be useful to have a link to holt winters algorithm here for the sake of reference

src/query/functions/temporal/aggregation.go
benraskin92: :(

src/query/functions/temporal/aggregation.go
```diff
@@ -29,28 +29,30 @@ import (
...
 	StdDevType = "stddev_over_time"
 
-	// StdVarType calculates the standard variance of all values in the specified interval
+	// StdVarType calculates the standard variance of all values in the specified interval.
 	StdVarType = "stdvar_over_time"
 
-	// HoltWintersType produces a smoothed value for time series based on the specified interval
+	// HoltWintersType produces a smoothed value for time series based on the specified interval.
+	// The algorithm used comes from https://en.wikipedia.org/wiki/Exponential_smoothing#Double_exponential_smoothing.
+	// Holt-Winters should only be used with gauges.
```
arnikola: :+1:

src/query/functions/temporal/aggregation.go
```diff
@@ -134,16 +136,18 @@ func makeHoltWintersFn(sf, tf float64) aggFunc {
...
-		for i, val := range vals {
+		for _, val := range vals {
 			if math.IsNaN(val) {
 				continue
 			}
 
 			if !foundFirst {
 				foundFirst = true
 				s1 = val
+				idx++
```
arnikola: Shouldn't this be outside of the `if !foundFirst` check? Otherwise, isn't it lagging by 1 behind the actual non-nan index?

src/query/functions/temporal/aggregation_test.go
```diff
@@ -353,3 +353,10 @@ func TestUnknownAggregation(t *testing.T) {
 	_, err := NewAggOp([]interface{}{5 * time.Minute}, "unknown_agg_func")
 	require.Error(t, err)
 }
+
+func TestHoltWinters(t *testing.T) {
+	holtWintersFn := makeHoltWintersFn(0.2, 0.6)
+	val := holtWintersFn([]float64{1, math.NaN(), 5, 10, 15})
```
arnikola: Could you do something like `holtWintersFn([]float64{math.NaN(),  1, math.NaN(), 5, 10, 15, math.NaN(), math.NaN()})` as your test data to hit all the NaN specific logic at once?

src/query/functions/temporal/aggregation.go
```diff
@@ -134,16 +136,18 @@ func makeHoltWintersFn(sf, tf float64) aggFunc {
 			b                       float64
 			x, y                    float64
 			s0, s1                  float64
```
arnikola: nit: these names can still stand to be more descriptive

src/query/functions/temporal/aggregation.go
benraskin92: Yeah I added `idx++` at the end to avoid this, but I can just add to right before here.

src/query/functions/temporal/aggregation.go
nikunjgit: what's happening with these dots

src/query/functions/temporal/aggregation.go
```diff
@@ -87,13 +92,112 @@ func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
 		return newBaseOp(args, optype, a)
 	}
 
+	if optype == HoltWintersType {
```
nikunjgit: move holtwinters into its own file, this logic can live there. 

if optype == HoltWintersType {
 return newHolterWintersOp(args)
}

src/query/functions/temporal/aggregation.go
```diff
@@ -87,13 +92,112 @@ func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
...
+		return curr
+	}
+}
+
+// Calculate the trend value at the given index i in raw data d.
+// This is somewhat analogous to the slope of the trend at the given index.
+// The argument "s" is the set of computed smoothed values.
+// The argument "b" is the set of computed trend factors.
+// The argument "d" is the set of raw input values.
+func calcTrendValue(i int, sf, tf, s0, s1, b float64) float64 {
```
nikunjgit: add a link in comments showing the algo you're implementing ? Otherwise, super hard to review for correctness.

src/query/functions/temporal/aggregation.go
benraskin92: Hmm and have a separate `HoltWintersOp`?

src/query/functions/temporal/holt_winters.go
```diff
@@ -0,0 +1,138 @@
...
+
+const (
+	// HoltWintersType produces a smoothed value for time series based on the specified interval.
+	// The algorithm used comes from https://en.wikipedia.org/wiki/Exponential_smoothing#Double_exponential_smoothing.
+	// Holt-Winters should only be used with gauges.
+	HoltWintersType = "holt_winters"
+)
+
+// NewHoltWintersOp creates a new base Holt-Winters transform with a specified node.
+func NewHoltWintersOp(args []interface{}, optype string) (transform.Params, error) {
```
nikunjgit: do you need to expose this ?

src/query/functions/temporal/holt_winters.go
```diff
@@ -0,0 +1,138 @@
...
+	if optype == HoltWintersType {
```
nikunjgit: umm this should just assume optype is holtwinters

src/query/functions/temporal/holt_winters.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	sf, ok := args[1].(float64)
+	if !ok {
+		return emptyOp, fmt.Errorf("unable to cast to scalar argument: %v for %s", args[0], HoltWintersType)
+	}
+
+	tf, ok := args[2].(float64)
+	if !ok {
+		return emptyOp, fmt.Errorf("unable to cast to scalar argument: %v for %s", args[0], HoltWintersType)
```
nikunjgit: args[1] ?

PR #1022, opened 2018-10-05
[coordinator] placement: add+remove fully atomic

schallert:
I realized that even with the change to make adds "safe" by default,
there could still be a race. That is, the placement that all shards were
validated as AVAILABLE for could not be the present placement by the
time `AddInstances` was called.

This changes placement adds/removes to use the placement algorithm to
construct the new placement, and perform a `CheckAndSet` with the old
placement.

Additionally, this implements "safe" adds in the delete handler as well.

### DIFF
--- a/src/query/api/v1/handler/placement/add.go
+++ b/src/query/api/v1/handler/placement/add.go
@@ -23,7 +23,6 @@ package placement
 import (
 	"fmt"
 	"net/http"
-	"strings"
 
 	"github.com/m3db/m3/src/cmd/services/m3query/config"
 	"github.com/m3db/m3/src/query/api/v1/handler"
@@ -90,6 +89,7 @@ func (h *AddHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	resp := &admin.PlacementGetResponse{
 		Placement: placementProto,
+		Version:   int32(placement.GetVersion()),
 	}
 
 	handler.WriteProtoMsgJSONResponse(w, resp, logger)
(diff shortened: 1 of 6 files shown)


src/query/api/v1/handler/placement/common.go
```diff
@@ -130,3 +143,21 @@ func RegisterRoutes(r *mux.Router, client clusterclient.Client, cfg config.Confi
 	r.HandleFunc(AddURL, logged(NewAddHandler(client, cfg)).ServeHTTP).Methods(AddHTTPMethod)
 	r.HandleFunc(DeleteURL, logged(NewDeleteHandler(client, cfg)).ServeHTTP).Methods(DeleteHTTPMethod)
 }
+
+func validateAllAvailable(p placement.Placement) error {
+	badInsts := []string{}
+	for _, inst := range p.Instances() {
+		if inst.Shards().NumShards() == 0 {
+			continue
```
cw9: I don't think this is possible in valid placement? could you add this to the list of `badInsts` for now and return as error?

src/query/api/v1/handler/placement/add.go
```diff
@@ -115,41 +115,39 @@ func (h *AddHandler) Add(
...
 
-	newPlacement, _, err := service.AddInstances(instances)
+	curPlacement, version, err := service.Placement()
 	if err != nil {
 		return nil, err
 	}
 
-	return newPlacement, nil
-}
+	if err := validateAllAvailable(curPlacement); err != nil {
```
cw9: This should work, but I feel a more reusable approach would be passing a placement validation function to the service through options and call it there, in case one day we change the logic in placement service and forgot to update here, what do you think?

src/query/api/v1/handler/placement/common.go
schallert: Hmm if it's not possible in a valid placement I'll revert this. I mainly did it because I had a test with an instance with no shards

src/query/api/v1/handler/placement/add.go
schallert: That does sound more flexible, but so far this is the only use case we've had for this and I'd rather limit the scope in this PR to unblock progress.

src/query/api/v1/handler/placement/common.go
cw9: Yeah we are checking against instance with no shards in a sharded placement, so this should not be possible

src/query/api/v1/handler/placement/add.go
cw9: kk, can you make a TODO here?

schallert: Opened https://github.com/m3db/m3cluster/issues/243

PR #1025, opened 2018-10-05
Refactor instrument options construction

cw9:
So I can reuse the instrument options in server.Run()


src/query/server/server.go
```diff
@@ -121,6 +121,7 @@ func Run(runOpts RunOptions) {
 	if err != nil {
 		logger.Fatal("could not connect to metrics", zap.Any("error", err))
 	}
+	instrumentOptions := instrument.NewOptions().SetMetricsScope(scope).SetZapLogger(logger)
```
richardartoul: super nit: break the sets onto separate lines

PR #1026, opened 2018-10-05
[query] Add prometheus label_join function

arnikola:


src/query/functions/tag/base.go
```diff
@@ -0,0 +1,108 @@
...
+// NewTagOp creates a new aggregation operation
+func NewTagOp(
+	opType string,
+	params []string,
+) (parser.Params, error) {
+	var fn tagTransformFunc
+	err := fmt.Errorf("operator not supported: %s", opType)
+	if opType == TagJoinType {
+		fn, err = makeTagJoinFunc(params)
+	}
```
benraskin92: newline

src/query/functions/tag/base.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"fmt"
+
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/executor/transform"
+	"github.com/m3db/m3/src/query/parser"
+)
+
+// Applies the given transform to block tags and series tags
```
benraskin92: `.` at the end of comments

src/query/functions/tag/base.go
```diff
@@ -0,0 +1,108 @@
...
+type tagTransformFunc func(*block.Metadata, []block.SeriesMeta)
+
+// NewTagOp creates a new aggregation operation
```
benraskin92: `s/aggregation/tag` or something similar

src/query/functions/tag/base.go
```diff
@@ -0,0 +1,108 @@
...
+func NewTagOp(
+	opType string,
+	params []string,
+) (parser.Params, error) {
+	var fn tagTransformFunc
+	err := fmt.Errorf("operator not supported: %s", opType)
+	if opType == TagJoinType {
```
benraskin92: I would just make this a switch case
```
var (
  fn tagTransformFunc
  err error
)

switch opType {
  case TagJoinType:
     fn, err = makeTagJoinFunc(params)
  case TagReplaceType:
     fn, err = makeTagReplaceFunc(params)
  default:
     return nil, fmt.Errorf("operator not supported: %s", opType)

if err != nil {
	return nil, err
}

return newBaseOp(opType, fn), nil
```

src/query/functions/tag/base.go
```diff
@@ -0,0 +1,108 @@
...
+	it, err := b.StepIter()
+	if err != nil {
+		return err
+	}
+
+	meta := it.Meta()
+	seriesMeta := it.SeriesMeta()
+
+	n.op.tagFn(&meta, seriesMeta)
+	err = b.UpdateMetas(meta, seriesMeta)
```
benraskin92:
```
if err := b.UpdateMetas(meta, seriesMeta); err != nil {
  return err
}
```

src/query/block/types.go
```diff
@@ -37,6 +37,8 @@ type Block interface {
 	StepIter() (StepIter, error)
 	// SeriesIter returns a SeriesIterator
 	SeriesIter() (SeriesIter, error)
+	// UpdateMetas updates this blocks metadata, and the metadatas for each series.
+	UpdateMetas(Metadata, []SeriesMeta) error
```
nikunjgit: this is a bit confusing. Shouldn't the metadatas be locked upon block creation ? The block is immutable so we shouldn't be allowed to change anything

src/query/functions/tag/join.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/models"
+)
+
+// TagJoinType joins the values of given tags using a given separator
+// and adds them to a given destination tag. It can combine any number of tags.
+// NB: This will override an existing tag if the given tag exists in the tag list.
+const TagJoinType = "label_join"
+
+func combineTagsWithSeparator(name []byte, separator []byte, values [][]byte) models.Tag {
```
nikunjgit: use bytes.Join to combine values , separator.

src/query/functions/tag/join.go
```diff
@@ -0,0 +1,132 @@
...
+		uniqueMap[s] = struct{}{}
+	}
+
+	return len(uniqueMap)
+}
+
+// noop.
+func noopFunc(_ *block.Metadata, _ []block.SeriesMeta) {}
+
+func makeTagJoinFunc(params []string) (tagTransformFunc, error) {
```
nikunjgit: consider making this makeTagJoinFunc(name, seperator string, tagNames []string). The caller can pass in params[0], params[1], params[2:]

src/query/functions/tag/join.go
```diff
@@ -0,0 +1,132 @@
...
+			if bytes.Equal(name, tag.Name) {
+				orderedTags = append(orderedTags, tag.Value)
+			}
+		}
+	}
+
+	return orderedTags
+}
+
+func uniqueNameCount(names []string) int {
```
nikunjgit: maybe move this to utils and have it return uniqueMap ? The caller can just take len(uniqueMap). This will make it more usable going forward

src/query/functions/tag/join.go
```diff
@@ -0,0 +1,132 @@
...
+	for i, tag := range params[2:] {
+		tagNames[i] = ([]byte(tag))
+	}
+
+	return func(meta *block.Metadata, seriesMeta []block.SeriesMeta) {
+		matchingCommonTags := meta.Tags.TagsWithKeys(tagNames)
+		lMatching := len(matchingCommonTags)
+		// Optimization if all joining series are shared by the block,
+		// or if there is only a shared metadata and no single series metas.
+		if lMatching == uniqueTagCount || len(seriesMeta) == 0 {
```
nikunjgit: almost looks like you're combining two functions here. One this modifies the meta and other which does seriesMeta ? Consider splitting ?

src/query/functions/tag/replace.go
```diff
@@ -0,0 +1,108 @@
...
+// Builds and adds the replaced tag if the source tag is found in the list,
+// and if the value of that tag matches the given regex. Returns false if the series
+// does not contain the source tag name, regardless of if the value matches the regex.
+func addTagIfFoundAndValid(
+	tags models.Tags,
+	sourceName []byte,
+	destinationName []byte,
+	destinationValRegex []byte,
+	regex *regexp.Regexp,
+) (models.Tag, bool, bool) {
```
nikunjgit: not a big fan of returning two bools, can be error prone. Also this function seems to be 3 things. Maybe the caller can do the found check ?

src/query/functions/tag/replace.go
```diff
@@ -0,0 +1,108 @@
...
+	indices := regex.FindSubmatchIndex(val)
+	if indices == nil {
+		return models.Tag{}, true, false
+	}
+
+	destinationVal := regex.Expand([]byte{}, destinationValRegex, val, indices)
+	return models.Tag{Name: destinationName, Value: destinationVal}, true, true
+}
+
+func makeTagReplaceFunc(params []string) (tagTransformFunc, error) {
```
nikunjgit: consider taking in 4 arguments ? Or at the top of this func, initializing all the 4 arguments ?

src/query/storage/block.go
```diff
@@ -90,6 +90,38 @@ func (m *multiBlockWrapper) SeriesIter() (block.SeriesIter, error) {
...
+func (m *multiBlockWrapper) UpdateMetas(
+	meta block.Metadata,
+	seriesMetas []block.SeriesMeta,
+) (block.Block, error) {
+	var (
+		consolidated   block.Block
+		unconsolidated block.UnconsolidatedBlock
+		err            error
+	)
+	if m.consolidated != nil {
```
nikunjgit: do you need to lock ?

src/query/functions/tag/join.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/m3db/m3/src/query/models"
+)
+
+// TagJoinType joins the values of given tags using a given separator
+// and adds them to a given destination tag. It can combine any number of tags.
+// NB: This will override an existing tag if the given tag exists in the tag list.
+const TagJoinType = "label_join"
+
+func combineTagsWithSeparator(name []byte, separator []byte, values [][]byte) models.Tag {
+	l := len(values)
```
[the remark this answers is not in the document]
arnikola: not in general, just for length and index :p

src/query/functions/tag/replace.go
arnikola: It's currently initializing all 4, just with changed order to shortcircuit just in case the regex is invalid

src/query/block/types.go
```diff
@@ -37,6 +37,8 @@ type Block interface {
 	StepIter() (StepIter, error)
 	// SeriesIter returns a SeriesIterator
 	SeriesIter() (SeriesIter, error)
+	// UpdateMetas updates this blocks metadata, and the metadatas for each series.
+	UpdateMetas(Metadata, []SeriesMeta) (Block, error)
```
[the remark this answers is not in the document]
arnikola: Yeah, currently returning a new block with updated metas. Renamed to `WithMetadata` to make it clearer

src/query/functions/tag/base.go
```diff
@@ -0,0 +1,112 @@
...
+	it, err := b.StepIter()
+	if err != nil {
+		return err
+	}
+
+	meta := it.Meta()
+	seriesMeta := it.SeriesMeta()
+
+	n.op.tagFn(&meta, seriesMeta)
+	bl, err := b.UpdateMetas(meta, seriesMeta)
```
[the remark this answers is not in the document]
arnikola: `b` would be closed by the caller. Added `defer bl.Close()`

src/query/functions/tag/join.go
arnikola: Nice call 👍

src/query/functions/tag/base.go
```diff
@@ -0,0 +1,112 @@
...
+
+// NewTagOp creates a new tag transform operation.
+func NewTagOp(
+	opType string,
+	params []string,
+) (parser.Params, error) {
+	var (
+		fn  tagTransformFunc
+		err error
+	)
```
benraskin92: newline

src/query/functions/tag/join.go
arnikola: I'm letting the function responsible for building the actual tagFunc handle parsing the generic string args into usable variables, as they have better context on which variables it needs to use, and more importantly, in which order. Since these are all strings, having them extracted in the base function could lead to problems during refactoring later as the compiler unfortunately won't catch them very well.

src/query/functions/tag/join.go
arnikola: Sounds good

src/query/functions/tag/join.go
arnikola: Yeah, I see what you mean, but unfortunately they're  coupled; if we're modifying the Meta, then it means we can avoid having to modify SeriesMetas, since the only changes from the function will be for common tags. 
If tags are split between meta and seriesmetas, we'd need to take all matching tags from the meta calculation and feed them into the seriesmeta calculation anyway, which won't help make this function any clearer

src/query/functions/tag/replace.go
arnikola: Yeah that's fair. I was just trying to avoid running the `Get` operation an extra time. Will consider ways to refactor this to only have to do it once.

src/query/functions/tag/base.go
arnikola: I don't think we've been adding newlines after closing `var()` blocks

src/query/storage/block.go
arnikola: I suppose it can't hurt :P

Going to change it to a RWLock and just RLock this

PR #1027, opened 2018-10-06
Refactor encode tags iterator for reuse

cw9:
Moving encode tags iterator to a common place to be reused between downsampler and ingester


src/cmd/services/m3coordinator/x/tags_iterator.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+	return nil, false
+}
+
+func (it *encodedTagsIter) Next() bool {
+	return it.tagDecoder.Next()
+}
+
+func (it *encodedTagsIter) Current() ([]byte, []byte) {
+	tag := it.tagDecoder.Current()
```
cw9: Should this be tag be closed? given the `Name` and `Value` in it are `checked.Bytes`

src/cmd/services/m3coordinator/x/tags_iterator.go
```diff
@@ -0,0 +1,139 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
robskillington: Hm, ideally we'd move to this into the aggregator repo.  But perhaps for now we should move this to `src/x/serialize` because it doesn't really live in the `src/cmd/services/m3coordinator/...` tree if it's going to become re-useable?

robskillington: Also actually, if we are going to be using the native `m3aggregator` binary for this flow, I don't think we need to touch this at all actually?  This was only a way to efficiently create IDs inside the `m3coordinator` itself, it was never supposed to be used in a full setup.

cw9: I think moving src/x/serialize sounds better to me, cause this logic does not belong to m3aggregator, this decoder should be matched with the encoder used in collector. M3agg itself does not care about how the id is encoded/decoded from tags. 

I need this decoder for m3coordinator ingestion stuff because I'm getting ids from m3aggregator and need to convert that id into tags before writing into m3db. Unless that is not necessary? I saw the flush logic in downsampler is decoding ids into tags tho.

robskillington: Yes I guess what I mean though is that collector is going to encode these IDs and tags, not like how the m3coordinator was doing it.

src/x/serialize/tags_iterator.go
```diff
@@ -0,0 +1,130 @@
...
+	it.bytes.DecRef()
+	it.tagDecoder.Reset(it.bytes)
+
+	if it.pool != nil {
+		it.pool.Put(it)
+	}
+}
+
+// MetricTagsIteratorPool pools MetricTagsIterator.
+type MetricTagsIteratorPool struct {
```
robskillington: Should we make this an interface for easier mocking perhaps?

cw9: Maybe do it in your diff? Also I'm curious what do you mock it for? I usually just use a pool with size 1 to verify reuse in unit tests

PR #1028, opened 2018-10-06
Add a m3msg server for ingestion

cw9:
Part 1 for the ingestion change in M3coordinator. 
This diff adds a m3msg server which decodes traffic from m3msg consumer into metrics and takes in a WriteFn to write those metrics.

Part 2 will be implementing a storage based ingester to fulfill the WriteFn.

### DIFF
--- a/src/cmd/services/m3coordinator/server/m3msg/config.go
+++ b/src/cmd/services/m3coordinator/server/m3msg/config.go
@@ -1,3 +1,23 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
 package m3msg
 
 import (
--- a/src/cmd/services/m3coordinator/server/m3msg/config.go
+++ b/src/cmd/services/m3coordinator/server/m3msg/config.go
@@ -1,0 +1,87 @@
+package m3msg
+
+import (
+	"github.com/m3db/m3metrics/encoding/msgpack"
+	"github.com/m3db/m3msg/consumer"
+	"github.com/m3db/m3x/instrument"
+	"github.com/m3db/m3x/server"
+)
+
+// Configuration configs the m3msg server.
+type Configuration struct {
+	// Server configs the server.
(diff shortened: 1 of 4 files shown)


src/cmd/services/m3coordinator/server/m3msg/config.go
```diff
@@ -0,0 +1,107 @@
...
+	// Consumer configs the consumer.
+	Consumer consumer.Configuration `yaml:"consumer"`
+}
+
+// NewServer creates a new server.
+func (c Configuration) NewServer(
+	writeFn WriteFn,
+	iOpts instrument.Options,
+) (server.Server, error) {
+	scope := iOpts.MetricsScope().Tagged(map[string]string{"server": "m3msg"})
```
richardartoul: Is what you're doing with tagged different than calling subscope?

src/cmd/services/m3coordinator/server/m3msg/handler.go
```diff
@@ -0,0 +1,155 @@
...
+	)
+	for {
+		msg, msgErr = c.Message()
+		if msgErr != nil {
+			break
+		}
+
+		h.processMessage(msg)
+	}
+	if msgErr != nil && msgErr != io.EOF {
```
richardartoul: how will the loop get restarted

src/cmd/services/m3coordinator/server/m3msg/types.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/m3db/m3msg/consumer"
+)
+
+// WriteFn is the function that writes a metric.
+type WriteFn func(
+	id []byte,
+	metricTimeNanos int64,
+	value float64,
+	sp policy.StoragePolicy,
+	callback *RefCountedCallback,
```
richardartoul: Supernit: Might be nice to make this have CallbackSuccess() and CallbackFailure()

src/cmd/services/m3coordinator/server/m3msg/types.go
```diff
@@ -0,0 +1,86 @@
...
+		return
+	}
+	if ref < 0 {
+		panic("invalid ref count")
+	}
+}
+
+// Callback performs the callback.
+func (r *RefCountedCallback) Callback(t CallbackType) {
+	if t == OnSuccess {
```
richardartoul: So if they callback with failure, refCount will never reach 0 and m3msg will retry? Is there any concept of an explicit nack?

src/cmd/services/m3coordinator/server/m3msg/types.go
```diff
@@ -0,0 +1,86 @@
...
+
+// RefCountedCallback wraps a message with a reference count, the message will
+// be acked once the reference count decrements to zero.
+type RefCountedCallback struct {
+	ref int32
+	msg consumer.Message
+}
+
+// NewRefCountedCallback creates a RefCountedCallback.
+func NewRefCountedCallback(msg consumer.Message) *RefCountedCallback {
```
richardartoul: If you're not doing any pooling, is all this ref counting necessary? Seems like you could just either Ack or Nack at the end

src/cmd/services/m3coordinator/server/m3msg/handler.go
cw9: the consumer is long lived with a tcp connection, c.Message() is a blocking call and the loop continues to call it to keep decoding messages from the connection

src/cmd/services/m3coordinator/server/m3msg/types.go
cw9: I'm open to it, I use the current interface so it's cheaper for us to add more callback types in the future

src/cmd/services/m3coordinator/server/m3msg/types.go
cw9: yeah the thing is, each message can contain more than one metric, so we only ack the message when all the metrics were ingested successfully. when one of them failed, we won't ack the message and it will be retried.

src/cmd/services/m3coordinator/server/m3msg/types.go
cw9: we actually do nack in statsdex, like when we get any non-retriable errors, we still callback success and will ack the message, I could use another callback type for that case so it's easier to understand

src/cmd/services/m3coordinator/server/m3msg/types.go
richardartoul: Could you "fail faster" if Callback(OnRetriableError) immediately cause a nack?

cw9: Unfortunately m3msg does not support that right now, retry only happens if a message is not acked within X amount of time. Could consider adding it in the future

PR #1030, opened 2018-10-06
Add APIs for M3Aggregator placement

richardartoul:
- [X] Add M3Agg placement endpoints and separate them out by path from the M3DB placement endpoints
- [X] Run all tests for both services
- [X] Update documentation for new and old endpoints
- [X] Refactor service to accept options instead of parsing headers directly


src/query/api/v1/handler/placement/common.go
```diff
@@ -37,18 +40,27 @@ import (
...
+	// TODO: Delete me?
 	// DefaultServiceName is the default service ID name
-	DefaultServiceName = "m3db"
+	DefaultServiceName = M3DBServiceName
 	// DefaultServiceEnvironment is the default service ID environment
 	DefaultServiceEnvironment = "default_env"
 	// DefaultServiceZone is the default service ID zone
 	DefaultServiceZone = "embedded"
-	// HeaderClusterServiceName is the header used to specify the service name.
-	HeaderClusterServiceName = "Cluster-Service-Name"
```
cw9: Why remove these headers?

src/query/api/v1/handler/placement/common.go
```diff
@@ -37,18 +40,27 @@ import (
...
+)
+
+var (
+	errServiceNameIsRequired        = errors.New("service name is required")
+	errServiceEnvironmentIsRequired = errors.New("service environment is required")
+	errServiceZoneIsRequired        = errors.New("service zone is required")
+	errUnableToParseService         = errors.New("unable to parse service")
+
+	// AllowedServiceNames is the list of allowed service names
+	AllowedServiceNames = []string{M3DBServiceName, M3AggServiceName}
```
cw9: can this be private? also maybe just use a map?

src/query/api/v1/handler/placement/delete_all.go
```diff
@@ -33,8 +33,12 @@ import (
 )
 
 const (
-	// DeleteAllURL is the url for the handler to delete all placements (with the DELETE method).
-	DeleteAllURL = handler.RoutePrefixV1 + "/placement"
+	// OldM3DBDeleteAllURL is the old url for the handler to delete all placements, maintained
+	// for backwards compatibility.
+	OldM3DBDeleteAllURL = handler.RoutePrefixV1 + "/placement"
```
cw9: hmm, this does not contain `delete` in it?

src/query/generated/proto/admin/placement.proto
```diff
@@ -21,3 +21,9 @@ message PlacementAddRequest {
   // are AVAILABLE for all their shards. force overrides that.
   bool force = 2;
 }
+
+message AggPlacementInitRequest {
```
cw9: Just reuse `PlacementInitRequest` for this? They should always be the same

src/query/api/v1/handler/placement/common.go
richardartoul: I'm moving the service to the URL itself. I'm gonna add back the zone/environment headers.

src/query/api/v1/handler/placement/delete_all.go
richardartoul: nope, just DELETE method

src/query/api/v1/handler/placement/init_test.go
```diff
@@ -68,27 +69,74 @@ func TestPlacementInitHandler(t *testing.T) {
...
+	mockPlacementService.EXPECT().BuildInitialPlacement(gomock.Not(nil), 64, 2).Return(nil, errors.New("unable to build initial placement"))
+	handler.ServeHTTP(M3DBServiceName, w, req)
+	resp = w.Result()
+	body, err = ioutil.ReadAll(resp.Body)
+	require.NoError(t, err)
+	assert.Equal(t, http.StatusInternalServerError, resp.StatusCode)
+	assert.Equal(t, initTestInvalidRequestResponse, string(body))
+}
+
+func TestAggPlacementInitHandler(t *testing.T) {
```
richardartoul: TODO: Fix this test

src/query/util/logging/log.go
```diff
@@ -123,3 +123,22 @@ func WithResponseTimeLogging(next http.Handler) http.Handler {
 		}
 	})
 }
+
+// WithServiceNameResponseTimeLogging wraps around the given handler, providing response time logging
+func WithServiceNameResponseTimeLogging(
```
richardartoul: TODO: Rename this

src/query/api/v1/handler/placement/common.go
```diff
@@ -59,42 +90,127 @@ type Handler struct {
...
-	pOpts := placement.NewOptions().SetValidZone(serviceZone)
+	if opts.ServiceName == M3AggServiceName {
+		var (
+			maxAggregationWindowSize = opts.M3Agg.MaxAggregationWindowSize
+			warmupDuration           = opts.M3Agg.WarmupDuration
+			now                      = time.Now()
+		)
+		pOpts = pOpts.
+			SetIsMirrored(true).
+			// TODO(rartoul): Do we need to set placement cutover time? Seems like that would
```
richardartoul: @cw9

src/query/api/v1/handler/placement/add.go
```diff
@@ -36,13 +38,26 @@ import (
...
 const (
-	// AddURL is the url for the placement add handler (with the POST method).
-	AddURL = handler.RoutePrefixV1 + "/placement"
+	// OldM3DBAddURL is the old url for the placement add handler, maintained for
+	// backwards compatibility.
+	OldM3DBAddURL = handler.RoutePrefixV1 + "/placement"
+
+	// M3DBAddURL is the url for the placement add handler (with the POST method)
+	// for the M3DB service.
+	M3DBAddURL = handler.RoutePrefixV1 + "/services/m3db/placement"
```
robskillington: Hm, are we worried about breaking all our guides/etc?  I guess it's ok, at some point we'll probably have to version our API docs similar to ElasticSearch, etc so that we can go and see what the APIs were at a specific version (for users still running an older version).

src/query/api/v1/handler/placement/add.go
```diff
@@ -36,13 +38,26 @@ import (
...
+
+	// M3AggAddURL is the url for the placement add handler (with the POST method)
+	// for the M3Agg service.
+	M3AggAddURL = handler.RoutePrefixV1 + "/services/m3db/placement"
```
robskillington: Should this perhaps be `/services/m3aggregator/placement` or similar?

src/query/api/v1/handler/placement/add.go
```diff
@@ -115,7 +131,17 @@ func (h *AddHandler) Add(
 		return nil, err
 	}
 
-	service, algo, err := ServiceWithAlgo(h.client, httpReq.Header)
+	serviceOpts := NewServiceOptionsFromHeaders(serviceName, httpReq.Header)
+	if serviceName == M3AggServiceName {
```
robskillington: nit: Maybe use a switch here so when we add further specialization later it's straightforward where to make the change (and we don't use an if/else if)?

src/query/api/v1/handler/placement/add.go
```diff
@@ -115,7 +131,17 @@ func (h *AddHandler) Add(
...
+		if req.MaxAggregationWindowSizeNanos == 0 || req.WarmupDurationNanos == 0 {
```
robskillington: Perhaps use `<= 0`?  Just because a negative value is invalid too I assume.

src/query/api/v1/handler/placement/add.go
```diff
@@ -115,7 +131,17 @@ func (h *AddHandler) Add(
...
+			return nil, errAggWindowAndWarmupMustBeSet
+		}
+		serviceOpts.M3Agg = &M3AggServiceOptions{
+			MaxAggregationWindowSize: time.Duration(req.MaxAggregationWindowSizeNanos),
+			WarmupDuration:           time.Duration(req.WarmupDurationNanos),
```
robskillington: Could we not have a default warmup duration and max aggregation window size if they don't specify any?  I wouldn't have a good idea of what to choose for either of these without reading something first, and most users could probably change this later/shouldn't need to worry too much about these values perhaps?

src/query/api/v1/handler/placement/add_test.go
```diff
@@ -37,99 +37,160 @@ import (
...
+		mockClient, mockPlacementService := SetupPlacementTest(t)
+		handler := NewAddHandler(mockClient, config.Configuration{})
+
+		// Test add failure
+		var (
+			w   = httptest.NewRecorder()
+			req *http.Request
+		)
+		if serviceName == M3AggServiceName {
+			req = httptest.NewRequest(AddHTTPMethod, M3DBAddURL, strings.NewReader("{\"instances\":[], \"max_aggregation_window_size_nanos\": 1000000, \"warmup_duration_nanos\": 1000000}"))
```
robskillington: Do you want to use back ticks to avoid having to escape the quotes?  Looks much more readable, e.g.:
```go
strings.NewReader(`{"instances": [], "max_aggregation_window_size_nanos": 1000000, "warmup_duration_nanos": 1000000}"`)

```

src/query/api/v1/handler/placement/delete_test.go
```diff
@@ -33,119 +34,151 @@ import (
...
+		mockPlacementService.EXPECT().CheckAndSet(gomock.Any(), 1).Return(nil)
+		handler.ServeHTTP(serviceName, w, req)
+
+		resp = w.Result()
+		body, err = ioutil.ReadAll(resp.Body)
+		require.NoError(t, err)
+		// require.Equal(t, http.StatusOK, resp.StatusCode)
+		// if serviceName == M3AggServiceName {
+
+		// } else {
```
robskillington: Still need the if/else commented out code?
Also: perhaps use backticks here for these literal requests?

src/query/api/v1/handler/placement/get.go
```diff
@@ -35,8 +35,17 @@ import (
 )
 
 const (
-	// GetURL is the url for the placement get handler (with the GET method).
-	GetURL = handler.RoutePrefixV1 + "/placement"
+	// OldM3DBGetURL is the old url for the placement get handler, maintained for
+	// backwards compatibility.
+	OldM3DBGetURL = handler.RoutePrefixV1 + "/placement"
```
robskillington: nit: `DeprecatedM3DBGetURL`?

src/query/api/v1/handler/placement/get_test.go
```diff
@@ -62,80 +62,82 @@ func SetupPlacementTest(t *testing.T) (*client.MockClient, *placement.MockServic
...
-			},
-		},
-	}
+		}
 
-	const placementJSON = "{\"placement\":{\"instances\":{\"host1\":{\"id\":\"host1\",\"isolationGroup\":\"rack1\",\"zone\":\"test\",\"weight\":1,\"endpoint\":\"http://host1:1234\",\"shards\":[],\"shardSetId\":0,\"hostname\":\"host1\",\"port\":1234},\"host2\":{\"id\":\"host2\",\"isolationGroup\":\"rack1\",\"zone\":\"test\",\"weight\":1,\"endpoint\":\"http://host2:1234\",\"shards\":[],\"shardSetId\":0,\"hostname\":\"host2\",\"port\":1234}},\"replicaFactor\":0,\"numShards\":0,\"isSharded\":false,\"cutoverTime\":\"0\",\"isMirrored\":false,\"maxShardSetId\":0},\"version\":%d}"
+		const placementJSON = "{\"placement\":{\"instances\":{\"host1\":{\"id\":\"host1\",\"isolationGroup\":\"rack1\",\"zone\":\"test\",\"weight\":1,\"endpoint\":\"http://host1:1234\",\"shards\":[],\"shardSetId\":0,\"hostname\":\"host1\",\"port\":1234},\"host2\":{\"id\":\"host2\",\"isolationGroup\":\"rack1\",\"zone\":\"test\",\"weight\":1,\"endpoint\":\"http://host2:1234\",\"shards\":[],\"shardSetId\":0,\"hostname\":\"host2\",\"port\":1234}},\"replicaFactor\":0,\"numShards\":0,\"isSharded\":false,\"cutoverTime\":\"0\",\"isMirrored\":false,\"maxShardSetId\":0},\"version\":%d}"
```
robskillington: nit: Backticks for literal JSON.

src/query/api/v1/handler/placement/init.go
```diff
@@ -35,8 +36,15 @@ import (
 )
 
 const (
-	// InitURL is the url for the placement init handler (with the POST method).
-	InitURL = handler.RoutePrefixV1 + "/placement/init"
+	// OldM3DBInitURL is the old url for the placement init handler, maintained for backwards
+	// compatibility. (with the POST method).
+	OldM3DBInitURL = handler.RoutePrefixV1 + "/placement/init"
```
robskillington: nit: `DeprecatedM3DBInitURL`?

src/query/api/v1/handler/placement/init.go
```diff
@@ -101,7 +110,18 @@ func (h *InitHandler) Init(
 		return nil, err
 	}
 
-	service, err := Service(h.client, httpReq.Header)
+	serviceOpts := NewServiceOptionsFromHeaders(serviceName, httpReq.Header)
+	if serviceName == M3AggServiceName {
+		if req.MaxAggregationWindowSizeNanos == 0 || req.WarmupDurationNanos == 0 {
+			return nil, errAggWindowAndWarmupMustBeSet
```
robskillington: Hm seems this code is duplicated elsewhere?  Can we reuse?  Also same question about whether we can have defaults for these.

src/query/generated/proto/admin/placement.proto
```diff
@@ -8,6 +8,10 @@ message PlacementInitRequest {
   repeated placementpb.Instance instances = 1;
   int32 num_shards = 2;
   int32 replication_factor = 3;
+
+  // M3Agg placements only
```
robskillington: Maybe we should place these together, i.e.

```proto
message M3AggregatorPlacementOptions {
  uint64 max_aggregation_window_size_nanos = 1;
  uint64 warmup_duration_nanos = 2;
}

message PlacementInitRequest {
  repeated placementpb.Instance instances = 1;
  int32 num_shards = 2;
  int32 replication_factor = 3;
  M3AggregatorPlacementOptions aggregator_options = 4;
}

message PlacementAddRequest {
  repeated placementpb.Instance instances = 1;
  bool force = 2;
  M3AggregatorPlacementOptions aggregator_options = 4;
}
```

src/query/api/v1/handler/placement/add.go
richardartoul: We can probably have a sane default for warmupDuration, but maxAggregationWindowSize is very difficult to pick a sane default value for because if it’s too short they could lose data and if it’s too long their deploy takes forever

src/query/api/v1/handler/placement/common.go
```diff
@@ -135,13 +251,87 @@ func ConvertInstancesProto(instancesProto []*placementpb.Instance) ([]placement.
...
+			Truncate(maxAggregationWindowSize).
+			Unix()
+	}
+}
+
+func newShardCutOverValidationFn(now time.Time) placement.ShardValidationFn {
+	return func(s shard.Shard) error {
+		switch s.State() {
+		case shard.Initializing:
+			if s.CutoverNanos() > now.UnixNano() {
```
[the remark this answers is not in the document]
cw9: This is expected, this lambda is called before marking an intializing shard as available, so we don't allow users to mark the shard if `now` is before its `CutoverNanos`

src/query/api/v1/handler/placement/common.go
```diff
@@ -135,13 +251,87 @@ func ConvertInstancesProto(instancesProto []*placementpb.Instance) ([]placement.
...
-	r.HandleFunc(DeleteAllURL, logged(NewDeleteAllHandler(client, cfg)).ServeHTTP).Methods(DeleteAllHTTPMethod)
-	r.HandleFunc(AddURL, logged(NewAddHandler(client, cfg)).ServeHTTP).Methods(AddHTTPMethod)
-	r.HandleFunc(DeleteURL, logged(NewDeleteHandler(client, cfg)).ServeHTTP).Methods(DeleteHTTPMethod)
+func newShardCutOffValidationFn(now time.Time, maxAggregationWindowSize time.Duration) placement.ShardValidationFn {
+	return func(s shard.Shard) error {
+		switch s.State() {
+		case shard.Leaving:
+			// TODO(rartoul): This seems overly cautious, basically it requires an entire maxAggregationWindowSize
+			// to elapse before "leaving" shards can be cleaned up.
+			if s.CutoffNanos() > now.UnixNano()-maxAggregationWindowSize.Nanoseconds() {
```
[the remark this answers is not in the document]
cw9: similar to cutover logic, this is called to when a leaving shard is being removed(when marking its corresponding initializing shard), this logic checks if the time has passed this shard's `CutoffNanos`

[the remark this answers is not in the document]
cw9: To answer Richie's TODO question, yes it's overly cautious, but let's just keep the logic the same as r2admin for now.

src/query/generated/proto/admin/placement.proto
cw9: Any way we can avoid passing `M3AggregatorPlacementOptions` in request? I think it can be configured in the handler

richardartoul: The problem is that we can only guess at the correct value. They may not have all the namespaces configured on this coordinator, or the endpoints may be embedded in a different service that doesn't have the info....lets talk offline

src/query/api/v1/handler/placement/add.go
cw9: Is it better to write: 
`M3DBAddURL = handler.RoutePrefixV1 + "/services/" + M3DBServiceName + "/placement"`
in case you update the service in one place and forgot the others. 

Same goes to the m3agg url as well. Also nit: use `m3aggregator` instead of `m3agg` everywhere?

src/query/api/v1/handler/placement/common.go
```diff
@@ -36,65 +40,219 @@ import (
...
-		SetZone(serviceZone)
+		SetName(opts.ServiceName).
+		SetEnvironment(opts.ServiceEnvironment).
+		SetZone(opts.ServiceZone)
+
+	pOpts := placement.NewOptions().
+		SetValidZone(opts.ServiceZone).
+		SetIsSharded(true).
+		// M3Agg expects a staged placement.
+		SetIsStaged(true).
```
cw9: Is this shared by m3db as well? should use `isStaged=false` for m3db and `isStaged=true` for m3agg?

src/query/api/v1/handler/placement/common.go
```diff
@@ -134,14 +292,120 @@ func ConvertInstancesProto(instancesProto []*placementpb.Instance) ([]placement.
...
+		}
+	}
+}
 
-	r.HandleFunc(InitURL, logged(NewInitHandler(client, cfg)).ServeHTTP).Methods(InitHTTPMethod)
-	r.HandleFunc(GetURL, logged(NewGetHandler(client, cfg)).ServeHTTP).Methods(GetHTTPMethod)
-	r.HandleFunc(DeleteAllURL, logged(NewDeleteAllHandler(client, cfg)).ServeHTTP).Methods(DeleteAllHTTPMethod)
-	r.HandleFunc(AddURL, logged(NewAddHandler(client, cfg)).ServeHTTP).Methods(AddHTTPMethod)
-	r.HandleFunc(DeleteURL, logged(NewDeleteHandler(client, cfg)).ServeHTTP).Methods(DeleteHTTPMethod)
+type placementCutoverOpts struct {
```
cw9: nit: rename to `m3aggregatorPlacementOpts`?

src/query/api/v1/handler/placement/common.go
```diff
@@ -134,14 +292,120 @@ func ConvertInstancesProto(instancesProto []*placementpb.Instance) ([]placement.
...
+	r.HandleFunc(M3DBDeleteURL, deleteFn).Methods(DeleteHTTPMethod)
+	r.HandleFunc(M3AggDeleteURL, deleteFn).Methods(DeleteHTTPMethod)
+}
+
+// We want to generate a function that will return the cutoff such that it is always at the beginning
+// of an aggregation window size, but also later than or equal to the current time + warmup. We accomplish
+// this by adding the warmup time and the max aggregation window size to the current time, and then truncating
+// to the max aggregation window size. This ensure that we always return a time that is at the beginning of an
+// aggregation window size, but is also later than now.Add(warmup).
+func newShardCutOffNanosFn(now time.Time, maxAggregationWindowSize, warmup time.Duration) placement.TimeNanosFn {
```
cw9: just call shardCutoverNanos like `r2admin`?

src/query/api/v1/handler/placement/add.go
cw9: Also consider making a const for `"/services/" + M3DBServiceName + "/placement"`, I see that string being reused in a lot of places.

src/query/api/v1/handler/placement/common.go
richardartoul: not shared, this is inside a if m3agg switch statement

richardartoul: oh no its not you're right, will fix

src/query/api/v1/handler/placement/add.go
robskillington: Agreed can we break up the consts?

src/query/api/v1/handler/placement/delete_all.go
```diff
@@ -23,18 +23,26 @@ package placement
...
 const (
-	// DeleteAllURL is the url for the handler to delete all placements (with the DELETE method).
-	DeleteAllURL = handler.RoutePrefixV1 + "/placement"
+	// DeprecatedM3DBDeleteAllURL is the old url for the handler to delete all placements, maintained
+	// for backwards compatibility.
+	DeprecatedM3DBDeleteAllURL = handler.RoutePrefixV1 + "/placement"
+
+	// M3DBDeleteAllURL is the url for the handler to delete all placements (with the DELETE method)
+	// for the M3DB service.
+	M3DBDeleteAllURL = handler.RoutePrefixV1 + "/services/m3db/placement"
```
robskillington: This could be reused from somewhere yeah?  `"/services/m3db/placement"` is rewritten a fair bit.

src/query/api/v1/handler/placement/delete_all.go
```diff
@@ -23,18 +23,26 @@ package placement
...
+
+	// M3AggDeleteAllURL is the url for the handler to delete all placements (with the DELETE method)
+	// for the M3Agg service.
+	M3AggDeleteAllURL = handler.RoutePrefixV1 + "/services/m3agg/placement"
```
robskillington: Same here, can we make a const for this combination `"/services/m3agg/placement"`?

src/query/api/v1/handler/placement/get.go
```diff
@@ -23,20 +23,28 @@ package placement
...
 const (
-	// GetURL is the url for the placement get handler (with the GET method).
-	GetURL = handler.RoutePrefixV1 + "/placement"
+	// DeprecatedM3DBGetURL is the old url for the placement get handler, maintained for
+	// backwards compatibility.
+	DeprecatedM3DBGetURL = handler.RoutePrefixV1 + "/placement"
+
+	// M3DBGetURL is the url for the placement get handler (with the GET method)
+	// for the M3DB service.
+	M3DBGetURL = handler.RoutePrefixV1 + "/m3db/services/placement"
```
robskillington: Same here, etc.

src/query/api/v1/handler/placement/common.go
richardartoul: done

src/query/api/v1/httpd/handler.go
```diff
@@ -129,6 +139,24 @@ func (h *Handler) RegisterRoutes() error {
 	return nil
 }
 
+func (h *Handler) m3AggServiceOptions() *placement.M3AggServiceOptions {
+	maxResolution := time.Duration(0)
+	for _, ns := range h.clusters.ClusterNamespaces() {
```
robskillington: Want to do a nil check here?  Just in case `m3dbClusters` wasn't specified?

richardartoul: sure

PR #1038, opened 2018-10-08
Add storage based ingester to m3coordinator

cw9:


src/cmd/services/m3coordinator/ingest/config.go
```diff
@@ -0,0 +1,65 @@
+package ingest
```
robskillington: Missing license header.

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,175 @@
+package ingest
```
robskillington: Missing license header here too.

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,175 @@
...
+
+// Ingester ingests metrics with a worker pool.
+type Ingester struct {
+	workers xsync.PooledWorkerPool
+	p       pool.ObjectPool
+}
+
+// NewIngester creates an ingester.
+func NewIngester(
+	opts *Options,
```
robskillington: Can you make this non-pointer type?  Options should be required it looks like so no need to make pointer type.

scripts/development/m3_stack/m3coordinator.yml
```diff
@@ -45,3 +45,18 @@ clusters:
         jitter: true
       backgroundHealthCheckFailLimit: 4
       backgroundHealthCheckFailThrottleFactor: 0.5
+
+ingest:
+  ingester:
+    workerPoolSize: 100
+    opPool:
+      size: 100
```
robskillington: Probably by default want to use a reasonable set of defaults that would work for production, i.e. 10000 or something for these values?

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,195 @@
...
+	return &Ingester{
+		workers: opts.Workers,
+		p:       p,
+	}
+}
+
+// Ingest ingests a metric asynchronously with callback.
+func (i *Ingester) Ingest(
+	id []byte,
+	metricTimeNanos int64,
```
robskillington: Should we perhaps just make this `time.Time` considering it's converted to that later anyhow?  Less chance of getting the precision right if so.

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,195 @@
...
+	op.m.ingestSuccess.Inc(1)
+	op.callback.Callback(m3msg.OnSuccess)
+	op.p.Put(op)
+}
+
+func (op *ingestOp) attempt() error {
+	return op.s.Write(
+		// NB: No timeout is needed for this as the m3db client has a timeout
+		// configured to it.
+		context.Background(),
```
robskillington: Hm is it worth plumbing a `ctx` all the way from the top to here?  `ctx.Background()` is a bit untrue, the context really is from the ingester server.

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,195 @@
...
+	)
+}
+
+func (op *ingestOp) resetWriteQuery() error {
+	if err := op.resetTags(); err != nil {
+		return err
+	}
+	op.resetDataPoints()
+	op.q.Raw = string(op.id)
+	op.q.Unit = op.sp.Resolution().Precision
```
robskillington: I just checked how MaxUnit gets resolved for say a storage policy of `1m:30d` and it seems like it's going to set Unit of `xtime.Minute` which won't be compatible with M3DB (which at most supports a unit/precision of time.Second).

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,195 @@
...
+	op.q.Attributes.Retention = op.sp.Retention().Duration()
+	return nil
+}
+
+func (op *ingestOp) resetTags() error {
+	op.it.Reset(op.id)
+	op.q.Tags = op.q.Tags[:0]
+	for op.it.Next() {
+		name, value := op.it.Current()
+		op.q.Tags = append(op.q.Tags, models.Tag{Name: name, Value: value})
```
robskillington: This is unsafe to do, this is the bug I just fixed where you can't hold onto the byte refs from the tag iterator since it reuses them.

src/cmd/services/m3query/config/config.go
```diff
@@ -71,6 +73,18 @@ type Configuration struct {
...
+	Ingest *IngestConfiguration `yaml:"ingest"`
+}
+
+// IngestConfiguration is the configuration for ingestion server.
+type IngestConfiguration struct {
+	// Ingester is the configuration for storage based ingester.
+	Ingester ingest.Configuration `yaml:"ingester"`
+
+	// M3msg is the configuration for m3msg server.
+	M3msg m3msg.Configuration `yaml:"m3msg"`
```
robskillington: Technically this should probably be (due to camel casing convention): 
```go
M3Msg m3msg.Configuration `yaml:"m3msg"`
```

src/cmd/services/m3coordinator/ingest/ingest.go
cw9: I could, but then the server also does not have anything to add to the context. This is mainly to fulfill the storage interface, I'll just make a context whenever a new ingestOp is made like statsdex.

src/cmd/services/m3coordinator/ingest/ingest.go
cw9: I'll just use `xtime.Millisecond` then, seems that's the default value used in several places in query

cw9: Updating the usage in flush_handler.go as well

src/cmd/services/m3coordinator/downsample/flush_handler.go
```diff
@@ -142,7 +143,7 @@ func (w *downsamplerFlushHandlerWriter) Write(
 				Timestamp: time.Unix(0, mp.TimeNanos),
 				Value:     mp.Value,
 			}},
-			Unit: mp.StoragePolicy.Resolution().Precision,
+			Unit: xtime.Millisecond,
```
robskillington: This will have pretty poor compresson for things that can be written in with second precision.  

Could we please do at least something here where we try to find the best? i.e. 
```
switch {
case time.Duration(nanos) % time.Second == 0:
  unit = xtime.Second
case time.Duration(nanos) % time.Millisecond == 0:
  unit = xtime.Millisecond
case time.Duration(nanos) % time.Microsecond == 0:
  unit = xtime.Microsecond
default:
  unit = xtime.Nanosecond
}
```

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+}
+
+// Ingest ingests a metric asynchronously with callback.
+func (i *Ingester) Ingest(
+	id []byte,
+	metricTime time.Time,
+	value float64,
+	sp policy.StoragePolicy,
+	callback *m3msg.RefCountedCallback,
```
robskillington: Can we add `ctx context.Context` as the first argument?  (lint forces some context to always be the first arg)

We don't want to use the TODO context each time.

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,196 @@
...
+			// NB: we don't need a pool for the tag decoder since the ops are
+			// pooled, but currently this is the only way to get tag decoder.
+			tagDecoder := opts.TagDecoderPool.Get()
+			op := ingestOp{
+				s:  opts.Appender,
+				r:  retrier,
+				it: serialize.NewMetricTagsIterator(tagDecoder, nil),
+				p:  p,
+				m:  m,
+				c:  context.TODO(),
```
robskillington: This can just be nil if it gets set on each ingest op.

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,196 @@
...
+func (op *ingestOp) attempt() error {
+	return op.s.Write(op.c, &op.q)
+}
+
+func (op *ingestOp) resetWriteQuery() error {
+	if err := op.resetTags(); err != nil {
+		return err
+	}
+	op.resetDataPoints()
+	op.q.Raw = string(op.id)
```
robskillington: This actually doesn't need to be set

src/cmd/services/m3coordinator/ingest/ingest.go
cw9: hmm, the m3msg server will only be able to pass a `TODO`/`Background` context each time though, although I can reuse one context there, is it what you wanted?

src/cmd/services/m3coordinator/ingest/config.go
```diff
@@ -0,0 +1,85 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return NewIngester(*opts), nil
+}
+
+func (cfg Configuration) newOptions(
+	appender storage.Appender,
+	instrumentOptions instrument.Options,
+) (*Options, error) {
```
robskillington: Since you need to derefence this pointer later now for use in the constructor, perhaps make this return type `(Options, error)`?

src/cmd/services/m3coordinator/ingest/ingest.go
robskillington: Yes I think that's ideal, in the future if we have a per-op context then we'll be able to use that.  Otherwise here it'll be a single one that can never be specialized.  We should at least let the callers into this layer be able to specify their own.

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -0,0 +1,196 @@
...
+func (op *ingestOp) attempt() error {
+	return op.s.Write(op.c, &op.q)
+}
+
+func (op *ingestOp) resetWriteQuery() error {
+	if err := op.resetTags(); err != nil {
+		return err
+	}
+	op.resetDataPoints()
+	op.q.Unit = common.SanitizeUnitForM3DB(op.sp.Resolution().Precision)
```
robskillington: Great 👍

src/cmd/services/m3coordinator/downsample/flush_handler.go
robskillington: Can you update `flush_handler.go` to use the new method you created? `common.SanitizeUnitForM3DB(mp.StoragePolicy.Resolution().Precision)`

src/cmd/services/m3coordinator/server/m3msg/handler.go
```diff
@@ -32,6 +34,8 @@ import (
 	"github.com/uber-go/tally"
 )
 
+var ctx = context.TODO()
```
robskillington: Instead of making this global, can you give one to each handler?  

Also it makes more sense to derive from `context.Background()` rather than `context.TODO()` since we want to signify this was from a background listen server, rather than intending to replace it later.

src/cmd/services/m3coordinator/ingest/config.go
cw9: yeah, was trying to avoid returning `Options{}, err`, I can change that

src/cmd/services/m3coordinator/downsample/flush_handler.go
cw9: yeah, forgot about this

src/x/common/unit.go
```diff
@@ -0,0 +1,37 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package common
```
robskillington: Can we rename this to `convert`?  We could rename `SanitizeUnitForM3DB` to something like `UnitForM3DB(unit xtime.Unit) xtime.Unit` and start putting other conversion functions in here.

PR #1047, opened 2018-10-09
Merge m3aggregator into monorepo

prateek:


prateek:
@robskillington cheers, will address in a followup. all the subdir READMEs need to be updated the same atm.

PR #1050, opened 2018-10-09
Add m3collector service and config with report endpoint

robskillington:
- [x] Tests


src/collector/api/v1/handler/json/report.go
```diff
@@ -0,0 +1,187 @@
...
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"math"
+	"net/http"
+
+	"github.com/m3db/m3/src/collector/reporter"
+	"github.com/m3db/m3/src/query/api/v1/handler"
```
richardartoul: query dependencies feel a little weird....probably not much we can do about models / storage right now but if you're only importing the error function I wonder if its worth just duplicating it

richardartoul: ah you're using the WriteJSONResponse too...whatever lol

src/collector/api/v1/handler/json/report.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}
+
+	resp := &reportResponse{Reported: len(req.Metrics)}
+	handler.WriteJSONResponse(w, resp, h.instrumentOpts.ZapLogger())
+}
+
+func (h *reportHandler) parseRequest(r *http.Request) (*reportRequest, *handler.ParseError) {
+	body := r.Body
+	if r.Body == nil {
```
richardartoul: should you use the variable you just created?

src/collector/api/v1/handler/json/report.go
robskillington: Yeah, I'll put a todo to clean this up.

robskillington: Hm, screw it, I'll move it into `x/net/http`

src/collector/api/v1/handler/json/report.go
robskillington: Ah yup, that's kinda strange.  Ta.

src/collector/api/v1/handler/json/report.go
```diff
@@ -0,0 +1,187 @@
...
+		return nil, handler.NewParseError(err, http.StatusBadRequest)
+	}
+
+	return req, nil
+}
+
+func (h *reportHandler) newMetricID(metric metricValue) (id.ID, *handler.ParseError) {
+	tags := make(models.Tags, 0, len(metric.Tags))
+	for n, v := range metric.Tags {
+		tags = tags.AddTag(models.Tag{Name: []byte(n), Value: []byte(v)})
```
richardartoul: I guess the tag name has to be a string after JSON deserialization because you can't have []byte be a map key, but can the JSON map be `map[string][]byte` so we don't have to convert it here?

src/collector/api/v1/handler/json/report.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	tags = models.Normalize(tags)
+	tagsIter := storage.TagsToIdentTagIterator(tags)
+
+	encoder := h.encoderPool.Get()
```
richardartoul: do you need a reset here?

src/collector/config/m3collector.yml
```diff
@@ -0,0 +1,87 @@
...
+    matchRangePast: 2m
+    sortedTagIteratorPool:
+      size: 8192
+      watermark:
+        low: 0.7
+        high: 1.0
+
+  client:
+    placementKV:
+      namespace: /r2/m3aggregator
```
richardartoul: Is this the default namespace we want to use? Probably dont want r2 in the name

src/collector/config/m3collector.yml
```diff
@@ -0,0 +1,87 @@
...
+    placementWatcher:
+      key: placement
```
richardartoul: you're watching for m3agg placements right? I think this will be the name of the service in the placement so make this `m3agg` maybe?

src/collector/config/m3collector.yml
cw9: let's just change this to `/m3aggregator`? @richardartoul need to use the same in coordinator endpoint and m3agg config

src/collector/api/v1/handler/json/report.go
robskillington: Nah unfortunately JSON decoder expects `[]byte` types to be base64 encoded strings, so will make this endpoint very hard to use.  

This is a super low volume test endpoint, so don't fret too much about perf here.

src/collector/api/v1/handler/json/report.go
robskillington: Ah good call, cheers.

src/collector/config/m3collector.yml
robskillington: Sure thing.

src/collector/server/server.go
```diff
@@ -0,0 +1,201 @@
...
+	defer func() {
+		logger.Info("closing server")
+		if err := srv.Shutdown(ctx); err != nil {
+			logger.Error("error closing server", zap.Error(err))
+		}
+	}()
+
+	go func() {
+		logger.Info("starting server", zap.String("address", listenAddress))
+		if err := srv.ListenAndServe(); err != nil {
```
richardartoul: shouldnt this fatal or something?

robskillington: Sure thing, I'll make sure `err != http.ErrServerClosed` too so we don't fatal in the "normal" sigterm case where we explicitly shut it down ourselves (need to retain clean exit).

PR #1051, opened 2018-10-09
Add metrics for engine to track all processing.

nikunjgit:
Tested with following targets:
    - targets: ['localhost:9090','docker.for.mac.localhost:7203']

### DIFF
--- a/src/query/executor/engine.go
+++ b/src/query/executor/engine.go
@@ -136,21 +136,21 @@ func (e *Engine) Execute(ctx context.Context, query *storage.FetchQuery, opts *E
 func (e *Engine) ExecuteExpr(ctx context.Context, parser parser.Parser, opts *EngineOptions, params models.RequestParams, results chan Query) {
 	defer close(results)
 
-	req, reqCtx := newRequest(ctx, e, params)
+	req := newRequest(e, params)
 	defer req.finish()
-	nodes, edges, err := req.compile(reqCtx, parser)
+	nodes, edges, err := req.compile(ctx, parser)
 	if err != nil {
 		results <- Query{Err: err}
 		return
 	}
 
-	pp, err := req.plan(reqCtx, nodes, edges)
+	pp, err := req.plan(ctx, nodes, edges)
 	if err != nil {
 		results <- Query{Err: err}
 		return
 	}
 
-	state, err := req.execute(reqCtx, pp)
+	state, err := req.execute(ctx, pp)
 	// free up resources
 	if err != nil {
 		results <- Query{Err: err}
@@ -159,7 +159,7 @@ func (e *Engine) ExecuteExpr(ctx context.Context, parser parser.Parser, opts *En
 
 	result := state.resultNode
 	results <- Query{Result: result}
-	if err := state.Execute(reqCtx); err != nil {
+	if err := state.Execute(ctx); err != nil {
 		result.abort(err)
 	} else {
 		result.done()
--- a/src/query/executor/engine.go
+++ b/src/query/executor/engine.go
@@ -22,21 +22,20 @@ package executor
 
 import (
 	"context"
+	"time"
 
 	"github.com/m3db/m3/src/query/models"
 	"github.com/m3db/m3/src/query/parser"
-	"github.com/m3db/m3/src/query/plan"
 	"github.com/m3db/m3/src/query/storage"
-	"github.com/m3db/m3/src/query/util/logging"
 
-	"go.uber.org/zap"
+	"github.com/uber-go/tally"
 )
 
 // Engine executes a Query.
 type Engine struct {
 	// Used for tracking running queries.
 	tracker *Tracker
-	Stats   *QueryStatistics
+	metrics *engineMetrics
 	store   storage.Storage
 }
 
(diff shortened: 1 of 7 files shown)


src/query/executor/request.go
```diff
@@ -0,0 +1,142 @@
...
+
+// Request represents a single request.
+type Request struct {
+	engine     *Engine
+	params     models.RequestParams
+	parentSpan *span
+}
+
+func newRequest(ctx context.Context, engine *Engine, params models.RequestParams) (*Request, context.Context) {
+	parentSpan, parentCtx := startSpan(ctx, "all", engine.metrics.activeHist, engine.metrics.allStart, engine.metrics.allEnd)
```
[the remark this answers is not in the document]
nikunjgit: made them enums

src/query/server/server.go
```diff
@@ -191,7 +191,7 @@ func Run(runOpts RunOptions) {
 		defer cleanup()
 	}
 
-	engine := executor.NewEngine(backendStorage)
+	engine := executor.NewEngine(backendStorage, scope.SubScope("engine"))
```
arnikola: nit: Maybe use `query_engine` for additional context?

src/query/executor/request.go
```diff
@@ -0,0 +1,171 @@
...
+
+func (r *Request) finish() {
+	r.parentSpan.finish()
+}
+
+// span is a simple wrapper around opentracing.Span in order to
+// get access to the duration of the span for metrics reporting.
+type span struct {
+	s            opentracing.Span
+	start        time.Time
```
arnikola: Rather than using start and duration here, may be better to use a `tally.Timer`?

src/query/executor/request.go
```diff
@@ -34,6 +35,35 @@ import (
 	"go.uber.org/zap"
 )
 
+// State is the request state.
+type State int
```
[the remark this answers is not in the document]
nikunjgit: cleaned it up a little bit, the main motivation i suppose it that we would eventually look at state transitions. Eg : from compiled, you're only allowed to go into planning. May have a failed and retry state.

src/query/executor/request.go
```diff
@@ -0,0 +1,171 @@
...
+// Request represents a single request.
+type Request struct {
+	engine     *Engine
+	params     models.RequestParams
+	parentSpan *span
+}
+
+func newRequest(ctx context.Context, engine *Engine, params models.RequestParams) (*Request, context.Context) {
+	parentSpan, parentCtx := startSpan(ctx, all.String(), engine.metrics.activeHist, engine.metrics.all)
+	r := &Request{engine: engine, params: params, parentSpan: parentSpan}
```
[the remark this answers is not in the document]
benraskin92: What do you mean time required for the function?

src/query/executor/request.go
```diff
@@ -0,0 +1,171 @@
...
+// get access to the duration of the span for metrics reporting.
+type span struct {
+	s            opentracing.Span
+	start        time.Time
+	duration     time.Duration
+	durationHist tally.Histogram
+	counter      *counterWithDecrement
+}
+
+func startSpan(ctx context.Context, operationName string, durationHist tally.Histogram, counter *counterWithDecrement) (*span, context.Context) {
```
benraskin92: Want to maybe get one of the Jaeger guys to take a quick look at the diff? Couldn't hurt..

src/query/executor/request.go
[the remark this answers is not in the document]
arnikola: How long the function takes to execute

arnikola:
Also, general comment, can we perhaps add "bucketing" for execution time by series size? Similar to statsdex where we can get timings for e.g. `< 5000`datapoint series vs timings for series with `100,000` datapoints.

Can be useful for tracking down slow querieas over smaller series, which may otherwise be lost in the noise of larger queries

src/query/executor/request.go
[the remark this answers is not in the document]
nikunjgit: that's tracked already ? eg: all tracks time by all functions, compiling tracks time in compilation step, etc.

src/query/executor/request.go
```diff
@@ -0,0 +1,171 @@
...
+func (r *Request) compile(ctx context.Context, parser parser.Parser) (parser.Nodes, parser.Edges, error) {
+	sp, _ := startSpan(ctx, compiling.String(), r.engine.metrics.compilingHist, r.engine.metrics.compiling)
+	// TODO: Change DAG interface to take in a context
+	nodes, edges, err := parser.DAG()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if r.params.Debug {
+		logging.WithContext(ctx).Info("compiled dag", zap.Any("nodes", nodes), zap.Any("edges", edges))
```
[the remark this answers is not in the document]
nikunjgit: I think using traces might be better for that.

src/query/server/server.go
nikunjgit: already shows up as coordinator_engine.

src/query/executor/request.go
nikunjgit: i thought tally.Timer is the statsd timer variant. Eg: to get things like p50, p95, etc.

src/query/executor/request.go
```diff
@@ -0,0 +1,171 @@
...
+	r := &Request{engine: engine, params: params, parentSpan: parentSpan}
+	return r, parentCtx
+
+}
+
+func (r *Request) compile(ctx context.Context, parser parser.Parser) (parser.Nodes, parser.Edges, error) {
+	sp, _ := startSpan(ctx, compiling.String(), r.engine.metrics.compilingHist, r.engine.metrics.compiling)
+	// TODO: Change DAG interface to take in a context
+	nodes, edges, err := parser.DAG()
+	if err != nil {
```
black-adder: if this err occurs, the span won't be finished. Might be better to defer `sp.finish()`. Ditto in the other 2 functions below.

src/query/executor/request.go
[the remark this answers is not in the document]
black-adder: agreed, adding these as logs in the span will be helpful.

src/query/executor/request.go
black-adder: also, you could do something like:

```
import (
"github.com/opentracing/opentracing-go/ext"
"github.com/opentracing/opentracing-go/log"
)

defer func() {
   ext.Error.Set(sp.s, true)
   sp.s.LogFields(log.Error(err))
   sp.finish()
}
```

There are also database specific opentracing tags: https://github.com/opentracing/opentracing-go/blob/master/ext/tags.go#L105

nikunjgit:
Planning to do opentracing part in a separate diff to avoid glide hell.

src/query/executor/engine.go
```diff
@@ -53,20 +52,58 @@ type Query struct {
...
+	planning  *counterWithDecrement
+	executing *counterWithDecrement
+
+	activeHist    tally.Histogram
+	compilingHist tally.Histogram
+	planningHist  tally.Histogram
+	executingHist tally.Histogram
+}
+
+type counterWithDecrement struct {
```
benraskin92: Is this the 'blessed' way to do this?

nikunjgit: Yeah, suggestion from Artem

PR #1054, opened 2018-10-09
Dockerize m3aggregator and add collector integration

richardartoul:


robskillington:
These changes were merged as part of https://github.com/m3db/m3/pull/1061.

Issue #1056, opened 2018-10-09
Investigate flaky test `TestShardTickWriteRace`


arnikola:
Works fine in general, but have been noticing it failing on CI oftenish; tried to get a few repros for it but nothing definitive as of yet. 

Including logs for CI failure:
[ShardTick.log](https://github.com/m3db/m3/files/2462396/ShardTick.log)

prateek:
@arnikola cheers, yeah this has been on the backlog for a while now. First observed way back in https://github.com/m3db/m3/pull/506. I'll take a look this week.

arnikola:
Ah no worries, less of a problem now that rerunning tests is so much faster, was just putting up an issue so we can track it a bit better :)

arnikola:
Also, liking `println(fmt.Sprintf("god this is terrible"))` 😛

arnikola:
Resolved a while ago

PR #1058, opened 2018-10-09
Re-work TestShardTickWriteRace to capture flakes

prateek:
Capture more information when we see failures for #1056

### DIFF
--- a/src/dbnode/storage/shard_race_prop_test.go
+++ b/src/dbnode/storage/shard_race_prop_test.go
@@ -37,7 +37,7 @@ import (
 	"github.com/leanovate/gopter"
 	"github.com/leanovate/gopter/gen"
 	"github.com/leanovate/gopter/prop"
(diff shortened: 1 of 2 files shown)


src/dbnode/storage/shard_race_prop_test.go
```diff
@@ -155,28 +155,35 @@ func TestShardTickWriteRace(t *testing.T) {
 
 	wg.Add(numRoutines)
 
+	doneFn := func() {
+		if r := recover(); r != nil {
+			assert.Fail(t, "unexpected panic: %v", r)
```
arnikola: Maybe better as 
```
assert.NoError(t, recover(), "unexpected panic")
wg.Done()
```
?

prateek: unfortunately can't be NoError cause the panic doesn't have to be of type `error`. Could make it an `assert.Nil` if you feel strongly

PR #1063, opened 2018-10-10
Add linear regression functions

benraskin92:


src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,159 @@
...
+			return emptyOp, fmt.Errorf("invalid number of args for %s: %d", PredictLinearType, len(args))
+		}
+
+		duration, ok = args[1].(float64)
+		if !ok {
+			return emptyOp, fmt.Errorf("unable to cast to scalar argument: %v for %s", args[1], PredictLinearType)
+		}
+
+	case DerivType:
+		if len(args) != 1 {
```
nikunjgit: you forgot to assign duration here. Bug ?

benraskin92: All good- duration is not needed for `deriv`

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,155 @@
...
+		ok       bool
+	)
+
+	switch optype {
+	case PredictLinearType:
+		if len(args) != 2 {
+			return emptyOp, fmt.Errorf("invalid number of args for %s: %d", PredictLinearType, len(args))
+		}
+
+		duration, ok = args[1].(float64)
```
arnikola: What's arg[0]?

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (l linearRegressionNode) Process(dps ts.Datapoints, alignedTime time.Time) float64 {
+	if dps.Len() < 2 {
+		return math.NaN()
+	}
+
+	var slope, intercept float64
+
+	if l.op.operatorType == PredictLinearType {
```
arnikola: This reads a little weird to me, since you're splitting processing amongst two functions and doing related calculations in both.

Instead, can you pull the code from `linearRegression()` into this `Process` function, and let it do the switching on its own?

 Alternatively, if you want to avoid dealing with operator type inside the regression function, you can pass a bool that signifies wether you're calculating regression or deriv into there, and have it just return a float, something like `linearRegression(isDeriv bool, dps ts.Datapoints, interceptTime time.Time) float64`

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,155 @@
...
+	return slope
+}
+
+// linearRegression performs a least-square linear regression analysis on the
+// provided datapoints. It returns the slope, and the intercept value at the
+// provided time. The algorithm we use comes from https://en.wikipedia.org/wiki/Simple_linear_regression.
+func linearRegression(dps ts.Datapoints, interceptTime time.Time) (float64, float64) {
+	var (
+		n            float64
+		sumX, sumY   float64
```
arnikola: What are X and Y?

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,155 @@
...
+		sumXY, sumX2 float64
+
+		nonNaNCount int
```
arnikola: nit: maybe `valueCount`?

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,155 @@
...
+			if interceptTime.IsZero() {
+				// set interceptTime as timestamp of first non-NaN dp
+				interceptTime = dp.Timestamp
+			}
+		}
+
+		nonNaNCount++
+
+		// convert to milliseconds
+		x := float64(dp.Timestamp.Sub(interceptTime).Nanoseconds()/1000000) / 1e3
```
arnikola: I think this converts it to seconds, not millis. Can you just do `float64(dp.Timestamp.Sub(interceptTime).Seconds())`?

src/query/functions/temporal/base.go
```diff
@@ -360,7 +361,7 @@ func (c *baseNode) sweep(processedKeys []bool, maxBlocks int) {
 
 // Processor is implemented by the underlying transforms
 type Processor interface {
-	Process(values ts.Datapoints) float64
+	Process(values ts.Datapoints, alignedTime time.Time) float64
```
nikunjgit: call it evaluationTime ?

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,155 @@
...
+	// PredictLinearType should only be used with gauges.
+	PredictLinearType = "predict_linear"
+
+	// DerivType calculates the per-second derivative of the time series, using simple linear regression.
+	// DerivType should only be used with gauges.
+	DerivType = "deriv"
+)
+
+type linearRegressionProcessor struct {
+	duration float64
```
nikunjgit: instead of this, store a function here which takes in a slope and intercept, returns a float. This way you don't have to compare Op in Process.

src/query/functions/temporal/linear_regression.go
nikunjgit: I like separate linearRegression since it makes it easy to test. My above comment about having a separate func instead duration should fix this problem.

src/query/functions/temporal/linear_regression.go
benraskin92: The duration for the actual query.

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+// linearRegression performs a least-square linear regression analysis on the
+// provided datapoints. It returns the slope, and the intercept value at the
+// provided time. The algorithm we use comes from https://en.wikipedia.org/wiki/Simple_linear_regression.
+func linearRegression(dps ts.Datapoints, interceptTime time.Time, isDeriv bool) (float64, float64) {
+	var (
+		n            float64
+		sumX, sumY   float64
+		sumXY, sumX2 float64
```
nikunjgit: better name ?

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,162 @@
...
+
```
nikunjgit: remove line ?

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,162 @@
...
+			continue
+		}
+
+		if valueCount == 0 && isDeriv {
+			// set interceptTime as timestamp of first non-NaN dp
+			interceptTime = dp.Timestamp
+		}
+
+		valueCount++
+
```
nikunjgit: remove line ?

src/query/functions/temporal/linear_regression.go
```diff
@@ -0,0 +1,162 @@
...
+	isDeriv    bool
+}
+
+func (l linearRegressionNode) Process(dps ts.Datapoints, evaluationTime time.Time) float64 {
+	if dps.Len() < 2 {
+		return math.NaN()
+	}
+
+	slope, intercept := linearRegression(dps, evaluationTime, l.isDeriv)
+
```
nikunjgit: remove line ?

PR #1065, opened 2018-10-10
Improve commitlog bootstrapping reliability in the face of sudden failures

richardartoul:
- [X] Prevent commit log reader from deadlocking when faced with certain types of commit log file corruption
- [X] Add regression property test to verify that the commit log reader never deadlocks or gets stuck regardless of any kind of random corruption

### DIFF
--- a/src/dbnode/persist/fs/commitlog/read_write_prop_test.go
+++ b/src/dbnode/persist/fs/commitlog/read_write_prop_test.go
@@ -1,4 +1,3 @@
-// +build big
 //
 // Copyright (c) 2018 Uber Technologies, Inc.
 //
@@ -122,17 +121,31 @@ func TestCommitLogReadWrite(t *testing.T) {
 }
 
 func TestCommitLogPropTest(t *testing.T) {
+	// Temporarily reduce size of buffered channels to increase change of
+	// catching deadlock issues.
+	var (
+		oldDecoderInBufChanSize  = decoderInBufChanSize
+		oldDecoderOutBufChanSize = decoderOutBufChanSize
+	)
+	defer func() {
+		decoderInBufChanSize = oldDecoderInBufChanSize
+		decoderOutBufChanSize = oldDecoderOutBufChanSize
+	}()
+	decoderInBufChanSize = 0
+	decoderOutBufChanSize = 0
+
 	basePath, err := ioutil.TempDir("", "commit-log-tests")
 	require.NoError(t, err)
 	defer os.RemoveAll(basePath)
 
 	parameters := gopter.DefaultTestParameters()
-	parameters.MinSuccessfulTests = 8
+	parameters.MinSuccessfulTests = 40
 	properties := gopter.NewProperties(parameters)
 
 	comms := clCommandFunctor(basePath, t)
 	properties.Property("CommitLog System", commands.Prop(comms))
 	properties.TestingRun(t)
+	panic("wtf")
 }
 
 // clCommandFunctor is a var which implements the command.Commands interface,
@@ -178,6 +191,19 @@ var genOpenCommand = gen.Const(&commands.ProtoCommand{
 		if err != nil {
 			return err
 		}
+		cLog := s.cLog.(*commitLog)
+		cLog.newCommitLogWriterFn = func(flushFn flushFn, opts Options) commitLogWriter {
+			fmt.Println("a")
+			wIface := newCommitLogWriter(flushFn, opts)
+			fmt.Println("b")
+			w := wIface.(*writer)
+			fmt.Println("c")
+			w.chunkWriter = &corruptingChunkWriter{w.chunkWriter.(*chunkWriter)}
+			fmt.Println("d")
+			return w
+			// return newCommitLogWriter(flushFn, opts)
+
+		}
 		return s.cLog.Open()
 	},
 	NextStateFunc: func(state commands.State) commands.State {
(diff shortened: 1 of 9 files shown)


src/dbnode/persist/fs/types.go
```diff
@@ -460,3 +460,13 @@ type BlockRetrieverOptions interface {
 	// IdentifierPool returns the identifierPool
 	IdentifierPool() ident.Pool
 }
+
+// FD is the interface implemented by *os.File.
+type FD interface {
```
prateek: nit: move this to `m3/src/x/os` and call it `File` instead of `FD`

src/x/test/fd.go
```diff
@@ -0,0 +1,73 @@
...
+// Write to the underlying f.d with a chance of corrupting it.
+func (c *CorruptingFD) Write(p []byte) (int, error) {
+	threshold := uint64(c.corruptionProbability * float64(math.MaxUint64))
+	if rand.Uint64() <= threshold {
+		var (
+			byteStart  int
+			byteOffset int
+		)
+		if len(p) > 1 {
+			byteStart = rand.Intn(len(p) - 1)
```
prateek: bug: Have to use gopter's RNG to generate reproducible results, using the global one messes that up. 

One way around this is to generate a long slice (say 1K) of []bool in gopter, and pass that instead of the probability to the CorruptingFD. Then you can loop over it cyclically to decide whether a write should be corrupted or not. This way you get randomness based on test input and tests are still reproducible.

src/dbnode/persist/fs/commitlog/writer.go
```diff
@@ -80,6 +80,13 @@ type commitLogWriter interface {
 	Close() error
 }
 
+type chunkWriterIface interface {
+	reset(f fs.FD)
```
prateek: nit: use uniform casing in method names

prateek: embed the interface's rather than copying methods if you can help it.

src/x/test/fd.go
prateek: Regardless of concurrency - these are separate RNGs, seeding one doesn't change the other.

richardartoul: I just mean I can seed both either the same seed

src/dbnode/persist/fs/commitlog/read_write_prop_test.go
```diff
@@ -122,12 +124,25 @@ func TestCommitLogReadWrite(t *testing.T) {
 }
 
 func TestCommitLogPropTest(t *testing.T) {
+	// Temporarily reduce size of buffered channels to increase change of
```
vdarulis: s/change/chance/

src/x/test/fd.go
robskillington: Can we just make the constructor take the same seed that we give to gopter?

richardartoul: Yeah that’s what I did

src/dbnode/persist/fs/commitlog/read_write_prop_test.go
```diff
@@ -178,6 +205,20 @@ var genOpenCommand = gen.Const(&commands.ProtoCommand{
 		if err != nil {
 			return err
 		}
+
+		if s.shouldCorrupt {
+			cLog := s.cLog.(*commitLog)
+			cLog.newCommitLogWriterFn = func(flushFn flushFn, opts Options) commitLogWriter {
+				wIface := newCommitLogWriter(flushFn, opts)
+				w := wIface.(*writer)
+				w.chunkWriter = &corruptingChunkWriter{
```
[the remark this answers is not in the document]
richardartoul: Sure thing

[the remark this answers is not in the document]
richardartoul: done

src/x/test/fd.go
prateek: yeah I was asking for the comment

src/x/test/fd.go
```diff
@@ -0,0 +1,79 @@
...
+	corruptionProbability float64
+	rng                   *rand.Rand
+}
+
+// NewCorruptingFD creates a new corrupting FD.
+func NewCorruptingFD(
+	fd xos.File,
+	corruptionProbability float64,
+	seed int64,
+) *CorruptingFD {
```
prateek: should this return a `x/os.File`?

src/x/test/fd.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"math"
+	"math/rand"
+
+	"github.com/m3db/m3/src/x/os"
+)
+
+// CorruptingFD implements the FD interface and can corrupt all writes issued
+// to it based on a configurable probability.
+type CorruptingFD struct {
```
prateek: does this need to be an exported type?

src/dbnode/persist/fs/commitlog/writer.go
```diff
@@ -277,22 +287,34 @@ func (w *writer) write(data []byte) error {
...
+type fsChunkWriter struct {
+	fd      xos.File
 	flushFn flushFn
 	buff    []byte
 	fsync   bool
 }
 
-func newChunkWriter(flushFn flushFn, fsync bool) *chunkWriter {
-	return &chunkWriter{
+func newChunkWriter(flushFn flushFn, fsync bool) *fsChunkWriter {
```
prateek: return `chunkWriter` instead?

src/dbnode/persist/fs/commitlog/read_write_prop_test.go
```diff
@@ -433,3 +515,40 @@ func uniqueID(ns, s string) uint64 {
...
+	chunkWriter           *fsChunkWriter
+	corruptionProbability float64
+	seed                  int64
+}
+
+func newCorruptingChunkWriter(
+	chunkWriter *fsChunkWriter,
+	corruptionProbability float64,
+	seed int64,
+) *corruptingChunkWriter {
```
prateek: return `chunkWriter` instead?

PR #1066, opened 2018-10-10
Handle commit log files with corrupt info headers during cleanup and bootstrap

richardartoul:
- [X] Automatically delete commit log files with corrupt info headers during cleanup instead of erroring out forever and requiring manual intervention.
- [X] Add static configuration for whether or not commitlog bootstrapping corruption should return unfulfilled or not for the whole thing (default to return unfulfilled)
- [X]  Always read all the commit log files we intended to read anyways, regardless of that config.
- [X] Auto-detect in commitlog bootstrapper if it is possible for the peers bootstrapper to satisfy requests. If not, return fulfilled instead of unfulfilled because the peers bootstrapper cant perform a repair anyways.
- [X] Don't return errors from the commitlog bootstrapper for corrupt files
- [X] Emit logs for all encountered corrupted files
- [X] Emit metrics for all encountered corrupt files

### DIFF
--- a/src/dbnode/persist/fs/commitlog/files.go
+++ b/src/dbnode/persist/fs/commitlog/files.go
@@ -30,7 +30,13 @@ import (
 	"github.com/m3db/m3/src/dbnode/persist/fs/msgpack"
 )
 
-type fsError error
+type fsError struct {
+	err error
+}
+
+func (e fsError) Error() string {
+	return e.err.Error()
+}
 
 // File represents a commit log file and its associated metadata.
 type File struct {
@@ -58,7 +64,7 @@ func ReadLogInfo(filePath string, opts Options) (time.Time, time.Duration, int64
 
 	fd, err = os.Open(filePath)
 	if err != nil {
-		return time.Time{}, 0, 0, fsError(err)
+		return time.Time{}, 0, 0, fsError{err}
 	}
 
 	chunkReader := newChunkReader(opts.FlushSize())
@@ -80,7 +86,7 @@ func ReadLogInfo(filePath string, opts Options) (time.Time, time.Duration, int64
 	err = fd.Close()
 	fd = nil
 	if err != nil {
-		return time.Time{}, 0, 0, fsError(err)
+		return time.Time{}, 0, 0, fsError{err}
 	}
 
 	return time.Unix(0, logInfo.Start), time.Duration(logInfo.Duration), logInfo.Index, decoderErr
--- a/src/dbnode/persist/fs/commitlog/files.go
+++ b/src/dbnode/persist/fs/commitlog/files.go
@@ -30,12 +30,17 @@ import (
 	"github.com/m3db/m3/src/dbnode/persist/fs/msgpack"
 )
 
+type openError error
+
 // File represents a commit log file and its associated metadata.
 type File struct {
 	FilePath string
 	Start    time.Time
 	Duration time.Duration
 	Index    int64
+	// Contains any errors encountered (except for filesystem errors) when trying
+	// to read the files log info.
+	Error error
 }
 
 // ReadLogInfo reads the commit log info out of a commitlog file
@@ -50,7 +55,7 @@ func ReadLogInfo(filePath string, opts Options) (time.Time, time.Duration, int64
 
 	fd, err = os.Open(filePath)
 	if err != nil {
-		return time.Time{}, 0, 0, err
+		return time.Time{}, 0, 0, openError(err)
 	}
 
 	chunkReader := newChunkReader(opts.FlushSize())
@@ -90,17 +95,24 @@ func Files(opts Options) ([]File, error) {
 
 	commitLogFiles := make([]File, 0, len(filePaths))
 	for _, filePath := range filePaths {
+		file := File{
+			FilePath: filePath,
+		}
+
 		start, duration, index, err := ReadLogInfo(filePath, opts)
-		if err != nil {
+		if _, ok := err.(openError); ok {
 			return nil, err
 		}
 
-		commitLogFiles = append(commitLogFiles, File{
-			FilePath: filePath,
-			Start:    start,
-			Duration: duration,
-			Index:    index,
-		})
+		if err != nil {
+			file.Error = err
+		} else {
+			file.Start = start
+			file.Duration = duration
+			file.Index = index
+		}
+
+		commitLogFiles = append(commitLogFiles, file)
 	}
 
 	sort.Slice(commitLogFiles, func(i, j int) bool {
--- a/src/dbnode/persist/fs/commitlog/files.go
+++ b/src/dbnode/persist/fs/commitlog/files.go
@@ -30,36 +30,6 @@ import (
 	"github.com/m3db/m3/src/dbnode/persist/fs/msgpack"
 )
 
-// FileOrError is a union/option type that returns an error if there was
-// any issue reading the commitlog info, or a File if there was not. Its
-// purpose is to force callers to handle the error.
-type FileOrError struct {
-	f File
-	// Contains any errors encountered when trying to read the commitlogs file info. We
-	// attempt to not include filesystem errors in this field, but that is accomplished
-	// on a best-effort basis and it is possible for this field to contain an error that
-	// is the result of a filesystem / O.S / hardware issue as opposed to an actually
-	// corrupt file.
-	e error
-}
-
-// File returns a File if the commitlog info was read, or an error otherwise.
-func (f FileOrError) File() (File, error) {
-	return f.f, f.e
-}
-
-// NewFileOrError creates a new FileOrError.
-func NewFileOrError(f File, e error, path string) FileOrError {
-	if e != nil {
-		e = newErrorWithPath(e, path)
-	}
-
-	return FileOrError{
-		f: f,
-		e: e,
-	}
-}
-
 // File represents a commit log file and its associated metadata.
 type File struct {
 	FilePath string
@@ -110,15 +80,16 @@ func ReadLogInfo(filePath string, opts Options) (time.Time, time.Duration, int64
 
 // Files returns a slice of all available commit log files on disk along with
 // their associated metadata.
-func Files(opts Options) ([]FileOrError, error) {
+func Files(opts Options) ([]File, []ErrorWithPath, error) {
 	commitLogsDir := fs.CommitLogsDirPath(
 		opts.FilesystemOptions().FilePathPrefix())
 	filePaths, err := fs.SortedCommitLogFiles(commitLogsDir)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
-	commitLogFiles := make([]FileOrError, 0, len(filePaths))
+	commitLogFiles := make([]File, 0, len(filePaths))
+	errorsWithPath := make([]ErrorWithPath, 0)
 	for _, filePath := range filePaths {
 		file := File{
 			FilePath: filePath,
@@ -126,7 +97,13 @@ func Files(opts Options) ([]FileOrError, error) {
 
 		start, duration, index, err := ReadLogInfo(filePath, opts)
 		if _, ok := err.(fsError); ok {
-			return nil, err
+			return nil, nil, err
+		}
+
+		if err != nil {
+			errorsWithPath = append(errorsWithPath, newErrorWithPath(
+				err, filePath))
+			continue
 		}
 
 		if err == nil {
@@ -135,16 +112,20 @@ func Files(opts Options) ([]FileOrError, error) {
 			file.Index = index
 		}
 
-		commitLogFiles = append(commitLogFiles, NewFileOrError(
-			file, err, filePath))
+		commitLogFiles = append(commitLogFiles, File{
+			FilePath: filePath,
+			Start:    start,
+			Duration: duration,
+			Index:    index,
+		})
 	}
 
 	sort.Slice(commitLogFiles, func(i, j int) bool {
 		// Sorting is best effort here since we may not know the start.
-		return commitLogFiles[i].f.Start.Before(commitLogFiles[j].f.Start)
+		return commitLogFiles[i].Start.Before(commitLogFiles[j].Start)
 	})
 
-	return commitLogFiles, nil
+	return commitLogFiles, errorsWithPath, nil
 }
 
 // ErrorWithPath is an error that includes the path of the file that
--- a/src/dbnode/persist/fs/commitlog/files.go
+++ b/src/dbnode/persist/fs/commitlog/files.go
@@ -38,18 +38,65 @@ func (e fsError) Error() string {
 	return e.err.Error()
 }
 
+// ErrorWithPath is an error that includes the path of the file that
+// had the error.
+type ErrorWithPath struct {
+	err  error
+	path string
+}
+
+// Error returns the error.
+func (e ErrorWithPath) Error() string {
+	return e.err.Error()
+}
+
+// Path returns the path of hte file that the error is associated with.
+func (e ErrorWithPath) Path() string {
+	return e.path
+}
+
+// FileOrError is a union/option type that returns an error if there was
+// any issue reading the commitlog info, or a File if there was not. Its
+// purpose is to force callers to handle the error.
+type FileOrError struct {
+	f File
+	// Contains any errors encountered when trying to read the commitlogs file info. We
+	// attempt to not include filesystem errors in this field, but that is accomplished
+	// on a best-effort basis and it is possible for this field to contain an error that
+	// is the result of a filesystem / O.S / hardware issue as opposed to an actually
+	// corrupt file.
+	e error
+}
+
+// File returns a File if the commitlog info was read, or an error otherwise.
+func (f FileOrError) File() (File, error) {
+	return f.f, f.e
+}
+
+func newFileOrError(f File, e error, path string) FileOrError {
+	if e != nil {
+		e = newErrorWithPath(e, path)
+	}
+
+	return FileOrError{
+		f: f,
+		e: e,
+	}
+}
+
 // File represents a commit log file and its associated metadata.
 type File struct {
 	FilePath string
 	Start    time.Time
 	Duration time.Duration
 	Index    int64
-	// Contains any errors encountered when trying to read the commitlogs file info. We
-	// attempt to not include filesystem errors in this field, but that is accomplished
-	// on a best-effort basis and it is possible for this field to contain an error that
-	// is the result of a filesystem / O.S / hardware issue as opposed to an actually
-	// corrupt file.
-	Error error
+}
+
+func newErrorWithPath(err error, path string) ErrorWithPath {
+	return ErrorWithPath{
+		err:  err,
+		path: path,
+	}
 }
 
 // ReadLogInfo reads the commit log info out of a commitlog file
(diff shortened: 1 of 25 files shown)


src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -782,6 +782,14 @@ func (s *commitLogSource) newReadCommitLogPred(
 	// that has a snapshot more recent than the global minimum. If we use an array for fast-access this could
 	// be a small win in terms of memory utilization.
 	return func(f commitlog.File) bool {
+		if f.Error != nil {
```
prateek: We can't skip corrupt commit log files like they didn't exist w/o risking data loss. It'd be much safe to see if any corrupt commit logs exist and if so fall back to the peers bootstrapper (in the case there are peers), and if there are no peers allow users to opt in to skip corrupt commit log files but don't do it by default.

richardartoul: I'm ok with making this opt-outable for users who really care about data integrity, but I think by default it should skip them. If you can't even read the info part of the commitlog, then (barring true corruption of the filesystem unrelated to M3DB shutdown) your commitlog isn't missing any data anyways other than the standard amount you would be missing due to the writebehind strategy. I don't think this is a situation where we need to involve operators in the general case.

richardartoul: by putting the commitlog bootstrapper before peers bootstrapper you're basically accepting losing the last few writes that occurred right before shutdown due to the writebehind strategy (as well as any writes that were issued while you were down) anyways.

In my mind we should emit a log, skip over bad commitlog files, and then eventually address this issue with the corruption APIs / a background repair process.

richardartoul: The other reason I think it sucks to push this burden down to the operators is that basically by the time this happen and you realize whats going on, there are only two options:

1. Disable commitlog bootstrapper and replace it with peers bootstrapper, which is fine, but only viable in non-catastrophic scenarios where all the other nodes are up.

2. Go and delete the bad commitlog file and then restart with the commitlog bootstrapper, at which point you've probably written out enough commit log files that what should have been 6->8 minutes of downtime turns into an hour-long ordeal.

src/dbnode/persist/fs/commitlog/files.go
```diff
@@ -30,12 +30,26 @@ import (
...
+	return e.err.Error()
+}
+
 // File represents a commit log file and its associated metadata.
 type File struct {
 	FilePath string
 	Start    time.Time
 	Duration time.Duration
 	Index    int64
+	// Contains any errors encountered when trying to read the commitlogs file info. We
```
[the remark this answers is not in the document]
richardartoul: Sure, I could also do something like:

```golang
type FileOrError struct {
    file File
    err Error
}


func (f *FileOrError) FIle() (File, Error) {
    return f.file, f.err
}
```

And then its basically impossible to misuse it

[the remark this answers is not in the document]
prateek: i don't like that people could use the Path regardless of error in that scheme – (e.g. the cleanup manager still has to delete the file regardless of err). In addition to what you suggested, could you expose a typed error with the path available to get around that?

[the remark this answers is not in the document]
richardartoul: Just to clarify, are you saying you don’t like the idea of someone receiving an error but then still needing to access the File object to get at the path, so we should export an error struct that include the file path, correct? Yeah that seems better to me too

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
prateek: For the peers case: how about just returning returning all requested shard ranges as unfullfilled in the commit log bootstrapper (won't need any interface changes from the way things are now) when the commit log runs into an error and that way we'll fall back to peers automatically. 

For the single node case: we can have a new RunOption to control the behaviour of whether to skip or not and the bootstrap process can be configured from the top level to specify what to pass it. That way the single node config we put out can default to come up but we can put a `WARNING` next to it and in docs.

src/dbnode/persist/fs/commitlog/files.go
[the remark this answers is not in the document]
prateek: Yep, exactly.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Spoke offline and settled on:

```
1. Add static configuration for whether or not commitlog bootstrapping corruption should return unfulfilled or not for the whole thing (default to return unfulfilled)
2. Always read all the commit log files we intended to read anyways, regardless of that config.
3. Add exception in commitlog bootstrap for single node deployment which will detect that this is the only node in the placement and default to returning fulfilled because there is nothing else you could possibly do to recover data
4. Don't return errors from the commitlog bootstrapper for corrupt files
5. Emit metrics / logs for all encountered corrupt files.
```

This gives us a good mix between maintaining data integrity as much as we can, but also not imposing undue operational burden.

In single node failures, the issue will just resolve itself.
In a single node failure within a multi-node cluster, the issue will just resolve itself.
In a multi-node failure within a multi-node cluster, *some* nodes with corrupt commitlog files may get stuck in the peer bootstrapping phase, but they can be "released" with a K.V change to the peer bootstrapper consistency level and come online without requiring an additional costly restart.

src/dbnode/persist/fs/commitlog/files.go
```diff
@@ -30,6 +30,61 @@ import (
 	"github.com/m3db/m3/src/dbnode/persist/fs/msgpack"
 )
 
+type fsError struct {
```
prateek: nit move these error types to the bottom of the file

src/dbnode/persist/fs/commitlog/files.go
```diff
@@ -38,6 +93,13 @@ type File struct {
 	Index    int64
 }
 
+func newErrorWithPath(err error, path string) ErrorWithPath {
```
prateek: same for this ctor, move it next to the associated type at the bottom of the file

src/dbnode/persist/fs/commitlog/files.go
```diff
@@ -72,39 +134,46 @@ func ReadLogInfo(filePath string, opts Options) (time.Time, time.Duration, int64
...
+			file.Index = index
+		}
+
+		commitLogFiles = append(commitLogFiles, NewFileOrError(
+			file, err, filePath))
 	}
 
 	sort.Slice(commitLogFiles, func(i, j int) bool {
-		return commitLogFiles[i].Start.Before(commitLogFiles[j].Start)
+		// Sorting is best effort here since we may not know the start.
```
prateek: why sort at all? and if you're going to sort - you can sort errors first (or last) and then guarantee order too. 

or better yet change the return type for this function: `([]File, []FileWithError, error)` and then won't need the FileOrError type.

src/dbnode/persist/fs/commitlog/files_test.go
```diff
@@ -38,27 +38,30 @@ import (
 )
 
 func TestFiles(t *testing.T) {
-	// TODO(r): Find some time/people to help investigate this flakey test.
```
prateek: lol is this no longer flaky?

src/dbnode/persist/fs/commitlog/files_test.go
```diff
@@ -38,27 +38,30 @@ import (
...
+		opts         = NewOptions()
+	)
 	opts = opts.SetFilesystemOptions(
 		opts.FilesystemOptions().
 			SetFilePathPrefix(dir),
 	)
 	files, err := Files(opts)
 	require.NoError(t, err)
-	require.Equal(t, 5, len(files))
+	require.True(t, len(files) >= minNumBlocks)
```
prateek: ah is this the fix?

src/dbnode/persist/fs/commitlog/iterator.go
```diff
@@ -190,14 +190,28 @@ func (i *iterator) nextReader() bool {
 	return true
 }
 
-func filterFiles(opts Options, files []File, predicate FileFilterPredicate) []File {
-	filteredFiles := make([]File, 0, len(files))
-	for _, f := range files {
+func filterFiles(opts Options, files []FileOrError, predicate FileFilterPredicate) ([]File, []ErrorWithPath) {
```
prateek: as above, can move this to be used within the `Files()` method.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -69,6 +72,38 @@ type commitLogSource struct {
...
+}
+
+func newCommitLogSourceDataAndIndexMetrics(scope tally.Scope) commitLogSourceDataAndIndexMetrics {
+	return commitLogSourceDataAndIndexMetrics{
+		data:  newCommitLogSourceMetrics(scope),
+		index: newCommitLogSourceMetrics(scope),
+	}
+}
+
+type commitLogSourceMetrics struct {
```
prateek: huge +1 to this.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -275,6 +323,7 @@ func (s *commitLogSource) ReadData(
 	}
 
 	// Read / M3TSZ encode all the datapoints in the commit log that we need to read.
+	s.metrics.data.readingCommitlogs.Update(1)
```
prateek: could you do same thing we normally do - use an atomic and have a background routine emitting the guage regularly. that way you'd get more than a single datapoint indicating what's going on.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -682,7 +751,7 @@ func (s *commitLogSource) newReadCommitLogPredBasedOnAvailableSnapshotFiles(
 	shardsTimeRanges result.ShardTimeRanges,
 	snapshotFilesByShard map[uint32]fs.FileSetFilesSlice,
 ) (
-	func(f commitlog.File) bool,
+	commitlog.FileFilterPredicate,
```
prateek: +1 again.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -1355,6 +1433,17 @@ func (s *commitLogSource) ReadIndex(
 			indexResults, indexOptions, indexBlockSize, resultOptions)
 	}
 
+	if iterErr := iter.Err(); iterErr != nil {
+		// Log the error and mark that we encountered corrupt data, but don't
```
prateek: lol what i wouldn't give for macros.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -1380,9 +1469,47 @@ func (s *commitLogSource) ReadIndex(
...
+) (bool, error) {
+	if !s.opts.ReturnUnfulfilledForCorruptCommitlogFiles() {
+		return false, nil
+	}
+
+	if !encounteredCorruptData {
+		return false, nil
+	}
+
+	couldObtainDataFromPeers, err := s.couldObtainDataFromPeers(
```
prateek: hm - i thought you were going to check if it's a single node for this case. why use the more extended peer condition?

src/dbnode/storage/cleanup.go
```diff
@@ -332,7 +352,47 @@ func (m *cleanupManager) commitLogTimes(t time.Time) ([]commitlog.File, error) {
...
+					"commitlog file error did not contain path: %v", err)
+				// Continue because we want to try and clean up the remining files instead of erroring out.
+				continue
+			}
+
+			m.opts.InstrumentOptions().Logger().Errorf(
+				"encountered err: %v reading commit log file: %v info during cleanup, marking file for deletion",
+				errorWithPath.Error(), errorWithPath.Path())
+
+			// TODO(rartoul): Leave this out until we have a way of distinguishing between a corrupt commit
```
prateek: +1

src/dbnode/persist/fs/commitlog/files.go
richardartoul: Sorting is important because you want to try and read the commit log files in order in the bootstrapper

src/dbnode/persist/fs/commitlog/files_test.go
richardartoul: yep

src/dbnode/persist/fs/commitlog/files.go
richardartoul: Your interface change suggestion is good though, just made it

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Seemed like it was slightly better to not hand off to the peers bootstrapper if it can't help anyways. Think I should remove it?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -300,9 +379,17 @@ func (s *commitLogSource) ReadData(
 			blockStart: dp.Timestamp.Truncate(blockSize),
 		}
 	}
+	doneReadingCommitlogs()
```
robskillington: Is it a waste to just emit the gauges the whole time this function is running?  Then we could do it all at the start and defer all these cleanups.  It would be better mainly because it would defend against an early return not stopping the loop which would cause a memory leak (for loop continuing to hold ref to the commit log source forever.

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -1319,11 +1421,13 @@ func (s *commitLogSource) ReadIndex(
 	)
 
 	// Start by reading any available snapshot files.
+	doneReadingSnapshots := s.metrics.index.emitReadingSnapshots()
```
robskillington: If we could do this with a defer somehow that would also be nicer, is there some state it can just read for the whole execution of the function and emit one or zero instead of needing to explicitly emit based on control flow?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -1380,9 +1503,44 @@ func (s *commitLogSource) ReadIndex(
...
 	}
 
+	shouldReturnUnfulfilled, err := s.shouldReturnUnfulfilled(
+		encounteredCorruptData, ns, shardsTimeRanges, opts)
+	if err != nil {
+		return nil, err
+	}
+
+	if shouldReturnUnfulfilled {
+		indexResult.SetUnfulfilled(shardsTimeRanges)
```
robskillington: We still return our data right? (I know this is likely true, just want to verify).  That way we can join whatever we had at least.

src/dbnode/storage/cleanup_test.go
```diff
@@ -710,3 +720,27 @@ func TestCleanupManagerCommitLogTimesMultiNS(t *testing.T) {
 	require.True(t, contains(filesToCleanup, time10))
 	require.True(t, contains(filesToCleanup, time20))
 }
+
+func TestCleanupManagerDeletesCorruptCommitLogFiles(t *testing.T) {
+	// TODO(rartoul): Re-enable this once https://github.com/m3db/m3/issues/1078
+	// is resolved.
```
robskillington: Can you add a note to the issue to "be sure to re-enable test `TestCleanupManagerDeletesCorruptCommitLogFiles` with the same PR that fixes this issue".

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -69,6 +72,68 @@ type commitLogSource struct {
...
+
+type commitLogSourceDataAndIndexMetrics struct {
+	data  commitLogSourceMetrics
+	index commitLogSourceMetrics
+}
+
+func newCommitLogSourceDataAndIndexMetrics(scope tally.Scope) commitLogSourceDataAndIndexMetrics {
+	return commitLogSourceDataAndIndexMetrics{
+		data:  newCommitLogSourceMetrics(scope),
+		index: newCommitLogSourceMetrics(scope),
```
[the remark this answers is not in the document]
richardartoul: oh duh, thanks

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: see comment below

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Yeah there isn't any obvious state we could read to make a decision (any state we could read would have to be added based on control flow.)

I think we either leave it as is, or we can simplify it by not distinguishing between reading commitlogs and snapshotting and just having a gauge for "commitlog bootstrapping data" and "commitlog bootstrapping index"

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
richardartoul: Yeah this should have no effect on what data we return (pretty sure my prop test still verifies the data is there) so as long as the caller doesn't ignore the data it will get merged in

src/dbnode/storage/cleanup_test.go
richardartoul: sure

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source.go
```diff
@@ -184,9 +197,13 @@ func (s *commitLogSource) ReadData(
 	}
 
 	var (
-		fsOpts         = s.opts.CommitLogOptions().FilesystemOptions()
-		filePathPrefix = fsOpts.FilePathPrefix()
+		// Emit bootstrapping gauge for duration of ReadData
+		doneBootstrapping      = s.metrics.data.emitBootstrapping
```
robskillington: Hm, did you mean this?:
```go
		doneBootstrapping      = s.metrics.data.emitBootstrapping()
```

robskillington: Re: why didn't this break, as a meta point if we wanted to be more "type safe" about it we could make it return an `io.Closer` than calls the method.

```go
import "github.com/m3db/m3x/close"

var _ close.Closer = closerFn(nil)

// Implement close.Closer
func (f closerFn) Close() {
	f()
}

// Now to call it

func myThing() close.Closer {
	return closerFn(func() { /* code */})
}
```

richardartoul: Fixed the function call.  I get what you're saying, but seems like it really hurts readability for something that would have been caught by a test if it was really important (we don't currently test whether metrics are being emitted properly.) I think I'm just gonna make the fix and leave as is for now

PR #1067, opened 2018-10-11
Reduce snapshot bootstrapper logging level

prateek:


prateek:
@richardartoul yeah it's too noisy IMO. I'm going to land this and we should do https://github.com/m3db/m3/issues/1077 soon.

PR #1075, opened 2018-10-12
Fix import path

cw9:


src/cmd/services/m3collector/config/config.go
```diff
@@ -21,7 +21,7 @@
 package config
 
 import (
-	"github.com/m3db/m3aggregator/client"
+	"github.com/m3db/m3/src/aggregator/client"
```
prateek: was this breaking master? if not, it's weird that it wasn't. could you ensure we don't have `github.com/m3db/m3aggregator` in either glide files

cw9: It's not breaking master, we don't ping m3agg in glide.yaml but glide pulled it in, after this change it will not show up in glide.lock

prateek: hey that's weird. i didn't realise that glide will pull in files that aren't mentioned in glide.lock when we run `glide install`, does it?

cw9: it can pull in files that aren't mentioned in `glide.yaml` it will figure out a version and put it in `glide.lock`

cw9: `glide.yaml` is only needed when you need to pin a particular version for a dep

prateek: It's totally allowed to add stuff to `glide.lock|yaml` when doing `glide init` or `glide update`. But it definitely should not do that when running `glide install` if a `glide.lock` exists, never seen that behaviour before so not sure i buy it.

prateek: ah ok, i see it wasn't in glide.yaml but was in glide.lock. this all makes sense now

PR #1083, opened 2018-10-15
Map of buckets in buffer; setup for out of order writes

justinjc:


src/dbnode/storage/shard.go
```diff
@@ -671,7 +671,6 @@ func (s *dbShard) tickAndExpire(
 				}
 			}
 			r.activeBlocks += result.ActiveBlocks
-			r.openBlocks += result.OpenBlocks
```
prateek: why remove this?

src/dbnode/retention/types.go
```diff
@@ -69,4 +69,10 @@ type Options interface {
 	// BlockDataExpiryAfterNotAccessedPeriod returns the period that blocks data should
 	// be expired after not being accessed for a given duration
 	BlockDataExpiryAfterNotAccessedPeriod() time.Duration
+
+	// SetOutOfOrderWritesEnabled sets whether writes to any time is enabled
+	SetOutOfOrderWritesEnabled(value bool) Options
```
prateek: hm this would be better placed `namespace.Options()` IMO, what do you think?

src/dbnode/storage/series/buffer.go
```diff
@@ -182,250 +164,115 @@ func (b *dbBuffer) Write(
...
-			// stream being present.
-			err = errMoreThanOneStreamAfterMerge
-			return
-		}
+func (b *dbBuffer) Snapshot(
+	ctx context.Context,
+	mType metricType,
+	blockStart time.Time,
+) (xio.SegmentReader, error) {
+	if mType == allMetricTypes {
```
prateek: why do this?

src/dbnode/storage/series/buffer.go
```diff
@@ -500,18 +445,212 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
-	drained           bool
+	bucketPool        *dbBufferBucketPool
+}
+
+func (b *dbBufferBuckets) resetTo(
+	start time.Time,
+	opts Options,
+	bucketPool *dbBufferBucketPool,
+) {
+	b.buckets = b.buckets[:0]
```
justinjc: Set elements to nil as well or else they won't be GC'd.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
-	now := b.nowFn()
-	pastMostBucketStart := now.Truncate(b.blockSize).Add(-1 * b.blockSize)
-	bucketNum := (pastMostBucketStart.UnixNano() / int64(b.blockSize)) % bucketsLen
-	result := 0
-	for i := int64(0); i < bucketsLen; i++ {
-		idx := int((bucketNum + i) % bucketsLen)
-		curr := pastMostBucketStart.Add(time.Duration(i) * b.blockSize)
-		result += fn(now, b, idx, curr)
+	// A call to flush can only be for warm writes
+	block, err := buckets.toBlock(WarmWrite)
```
richardartoul: Do we need to convert it to a block just so we can flush it?

Also, what happens to bootstrapped blocks here? Seems like they'll just get lost

richardartoul: If I'm right about the bootstrapped blocks being lost we'll probably need an integration test for that which would basically be the same as the two ones you deleted which were no longer applicable (for rotation)

src/dbnode/storage/series/buffer.go
```diff
@@ -500,18 +475,205 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
+	annotation []byte,
+	wType WriteType,
+) error {
+	return b.writableBucketCreate(wType).write(timestamp, value, unit, annotation)
+}
+
+func (b *dbBufferBuckets) merge(wType WriteType) (mergeResult, error) {
+	var res mergeResult
+	for _, bucket := range b.buckets {
+		// Only makes sense to merge buckets that are writable
```
richardartoul: why? Because the other ones have been flushed and will be evicted soon?

src/dbnode/storage/series/buffer.go
```diff
@@ -500,18 +475,205 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
+	t time.Time,
+	wType WriteType,
+) *dbBufferBucket {
+	bucket := b.bucketPool.Get()
+	bucket.resetTo(t, wType, b.opts)
+	b.buckets = append([]*dbBufferBucket{bucket}, b.buckets...)
+	return bucket
+}
+
+func (b *dbBufferBuckets) toBlock(wType WriteType) (block.DatabaseBlock, error) {
```
richardartoul: I feel like the name of this method should clarify that it only creates a block for the currently writable bucket (it ignores unwritable buckets as well as bootstrapped blocks)

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
-	if b.buckets[idx].needsDrain(now, start) {
-		return 1
+func (b *dbBuffer) Stats() bufferStats {
+	return bufferStats{
+		wiredBlocks: len(b.bucketsMap),
 	}
-	return 0
 }
 
 func (b *dbBuffer) Tick() bufferTickResult {
```
richardartoul: Tick probably need to evict the bootstrapped blocks too right?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
+		return FlushOutcomeErr, err
+	}
 
-		// Direct indexing is safe because canRead guarantees us at least one stream
-		res = streams[0]
-	})
+	// Don't need to check error here because a non-nil block means that a
+	// writable bucket exists
+	bucket, _ := buckets.writableBucket(WarmWrite)
+	bucket.version = version
```
richardartoul: probably need to set the version on the bootstrapped blocks too so we can evict them in the next tick

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
 		// data is important for use with understanding access patterns, etc.
-		bucket.setLastRead(b.nowFn())
-	})
+		buckets.setLastRead(b.nowFn())
+	}
 
 	return res
 }
 
+func (b *dbBuffer) sortedBucketsKeys(ascending bool) []xtime.UnixNano {
```
richardartoul: So I think the reason this used to accept a function was to avoid allocating a slice that definetly escapes. Is there a reason we're switching to doing it this way now?

src/dbnode/storage/series/buffer.go
```diff
@@ -500,18 +475,205 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
 }
 
-type dbBufferBucket struct {
-	opts              Options
+func (b *dbBuffer) newBucketsAt(
+	t time.Time,
+) *dbBufferBuckets {
+	buckets := b.bucketsPool.Get()
+	buckets.resetTo(t, b.opts, b.bucketPool)
+	b.bucketsMap[xtime.ToUnixNano(t)] = buckets
```
richardartoul: All the places where this is called will check this doesn't exist already presumably?

src/dbnode/storage/series/buffer.go
```diff
@@ -500,18 +475,205 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
+	b.bucketsCache[0] = newBuckets
+}
+
+func (b *dbBuffer) removeBucketsAt(blockStart time.Time) {
+	buckets, exists := b.bucketsAt(blockStart)
+	if !exists {
+		return
+	}
+
+	b.bucketsPool.Put(buckets)
```
richardartoul: Should you reset them first to nil out any pointers?

src/dbnode/storage/series/buffer.go
```diff
@@ -500,18 +475,205 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
+func (b *dbBufferBuckets) lastRead() time.Time {
+	return time.Unix(0, atomic.LoadInt64(&b.lastReadUnixNanos))
+}
+
+func (b *dbBufferBuckets) bootstrap(bl block.DatabaseBlock) {
+	// TODO(juchan): what is a "cold" bootstrap?
+	b.writableBucketCreate(WarmWrite).addBlock(bl)
+}
+
+func (b *dbBufferBuckets) writableBucket(wType WriteType) (*dbBufferBucket, bool) {
```
richardartoul: A lot of your logic is dependent on this function, but I think it can mess things up in the case of a failed flush. For example:

```
1) Node bootstraps, stores a block on the writeableBucket
2) Node begins flush, marks the writeable bucket as being version 1 now (not writable)
3) Flush fails
4) Flush begins again, but now it will skip the data from the previous bucket (which was persisted, but to an incomplete flush that will be lost)
```

richardartoul: This might just be an issue with the flushing logic....could the flush logic just flush all buckets for a given time that it has? Or does that not work

src/dbnode/storage/series/types.go
```diff
@@ -280,3 +291,22 @@ func NewStats(scope tally.Scope) Stats {
...
+
+	// WarmWrite represents warm writes (within the buffer past/future window)
+	WarmWrite
+
+	// ColdWrite represents cold writes (outside the buffer past/future window)
+	ColdWrite
+)
+
+// WriteOptions define different options for a write
+type WriteOptions struct {
```
robskillington: Remind me: were we going to avoid this being determined by the caller?

src/dbnode/storage/block/wired_list.go
```diff
@@ -231,7 +231,7 @@ func (l *WiredList) processUpdateBlock(v DatabaseBlock) {
 	// the updateCh even though its an update for a closed block. For the same reason, the wired list
 	// can receive blocks that were not retrieved from disk because the closed block was returned to
 	// a pool and then re-used.
-	unwireable := !entry.closed && entry.wasRetrievedFromDisk
```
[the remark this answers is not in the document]
robskillington: Hm, yeah this seems like we should be extra careful about.

Also is there locking semantics when we call `wiredListEntry()`?  What happens if it block gets closed/reused during this method call (i.e. returns `entry.closed == false && entry.wasRetrievedFromDisk == true` but before you put it back in, that changes to `entry.closed == true` and then to `entry.closed == false && entry.wasRetrievedFromDisk == false`).

src/dbnode/storage/errors/types.go
```diff
@@ -32,4 +32,8 @@ var (
 
 	// ErrTooPast is returned for a write which is too far in the past.
 	ErrTooPast = xerrors.NewInvalidParamsError(errors.New("datapoint is too far in the past"))
+
+	// ErrColdWritesNotEnabled is returned for when non-realtime metrics are disabled
+	// and a write is too far in the past or future
+	ErrColdWritesNotEnabled = xerrors.NewInvalidParamsError(errors.New("datapoint too far in the past or future. Adjust the values of bufferPast and bufferFuture or enable non-realtime metrics for this namespace"))
```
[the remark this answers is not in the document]
robskillington: minor nit: Also can we make this error shorter? It's pretty long to have to read in logs from the ingester.

src/dbnode/storage/series/buffer.go
justinjc: Done

src/dbnode/storage/series/buffer.go
```diff
@@ -521,10 +683,13 @@ type inOrderEncoder struct {
 
 func (b *dbBufferBucket) resetTo(
 	start time.Time,
+	wType WriteType,
+	opts Options,
 ) {
 	// Close the old context if we're resetting for use
```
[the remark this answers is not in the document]
justinjc: I think this comment is not referring to a literal context object, but more figuratively so on internal encoders/blocks.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
+					// that the version has successfully persisted to disk
+					bucket.finalize()
+					b.bucketPool.Put(bucket)
+					continue
+				}
 
-func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
-	// Perform a drain and reset if necessary
-	mergedOutOfOrderBlocks := bucketDrainAndReset(now, b, idx, start)
+				newBuckets = append(newBuckets, bucket)
```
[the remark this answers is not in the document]
justinjc: [SliceTricks](https://github.com/golang/go/wiki/SliceTricks#filtering-without-allocating) :)

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
-// forEachBucketAsc iterates over the buckets in time ascending order
-// to read bucket data
-func (b *dbBuffer) forEachBucketAsc(fn func(*dbBufferBucket)) {
-	for i := 0; i < bucketsLen; i++ {
-		idx := (b.pastMostBucketIdx + i) % bucketsLen
-		fn(&b.buckets[idx])
+func (b *dbBuffer) Flush(
+	ctx context.Context,
+	blockStart time.Time,
+	id ident.ID,
```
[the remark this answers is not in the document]
justinjc: I agree and I didn't like this either... but this function will not exist past the first milestone (there won't be any direct flushing in the final design), so I settled with this.

src/dbnode/storage/series/buffer.go
justinjc: Good catch, will fix.

src/dbnode/storage/series/buffer.go
justinjc: Non-writable buckets are guaranteed to be merged and not written to after that.

src/dbnode/storage/series/buffer.go
justinjc: Actually, this should contain the bootstrapped blocks as well.

When a block gets bootstrapped, the writable bucket will store that block. The `toBlock` function will then return a block encapsulating all the data within the bucket, merging bootstrapped blocks and encoders. Sometimes there's no bootstrapped block to merge with in which case we're just converting it to a block.

src/dbnode/storage/series/buffer.go
justinjc: Good point, I'll rename the function to specify writable buckets only. This function actually does take into account bootstrapped blocks because it does a merge to create a block containing all data within the bucket.

src/dbnode/storage/series/buffer.go
justinjc: When bootstrapping a block, it actually goes into the writable bucket (the same bucket where writes go to). Maybe I should rename writable bucket to mutable bucket to be clearer.

src/dbnode/storage/series/buffer.go
justinjc: Yep, that's in the `nonEvictedBuckets` part of the code. Bootstrapped blocks and incoming writes both live in the same bucket (the writable bucket).

src/dbnode/storage/series/series.go
```diff
@@ -384,50 +381,11 @@ func (s *dbSeries) FetchBlocksMetadata(
...
-	}
-
-	// There is already an existing block, perform a (lazy) merge.
-	return existingBlock.Merge(newBlock)
-}
-
 func (s *dbSeries) addBlockWithLock(b block.DatabaseBlock) {
 	b.SetOnEvictedFromWiredList(s.blockOnEvictedFromWiredList)
-	s.blocks.AddBlock(b)
+	s.cachedBlocks.AddBlock(b)
```
richardartoul: Can we add here (or in AddBlock) a check that emits an invariant violation if we overwrite an existing block since that should never happen and we removed the merge logic?

src/dbnode/storage/series/series.go
```diff
@@ -690,19 +574,19 @@ func (s *dbSeries) Close() {
 		// by the WiredList and should not be closed here. They will eventually
 		// be evicted and closed by  the WiredList when it needs to make room
 		// for new blocks.
-		for _, block := range s.blocks.AllBlocks() {
+		for _, block := range s.cachedBlocks.AllBlocks() {
 			if !block.WasRetrievedFromDisk() {
 				block.Close()
```
richardartoul: Should we just remove this? There should never be anything in here that was not retrieved from disk so there should be nothing to close. Can maybe keep it in and instead of closing emit an invariant violated

src/dbnode/storage/series/series.go
```diff
@@ -690,19 +574,19 @@ func (s *dbSeries) Close() {
...
 			}
 		}
 	default:
-		s.blocks.RemoveAll()
+		s.cachedBlocks.RemoveAll()
```
richardartoul: this will close them yeah?

src/query/generated/assets/openapi/spec.yml
```diff
@@ -683,12 +683,17 @@ definitions:
         type: "boolean"
       indexOptions:
         $ref: "#/definitions/IndexOptions"
+      coldWritesEnabled:
```
richardartoul: wanna leave this out for now so we don't confuse people?

src/query/generated/assets/openapi/spec.yml
```diff
@@ -683,12 +683,17 @@ definitions:
...
+        type: "boolean"
   RetentionOptions:
     type: "object"
     properties:
       retentionPeriodNanos:
         type: "integer"
         format: "int64"
+      futureRetentionPeriodNanos:
```
richardartoul: same comment

src/dbnode/storage/series/buffer.go
```diff
@@ -22,40 +22,37 @@ package series
...
-
-const (
-	computeBucketIdx computeBucketIdxOp = iota
-	computeAndResetBucketIdx
+	bucketsCacheSize       = 2
+	defaultBucketsPoolSize = 2
+	// In the most common case, there would only be one bucket in a
+	// buckets slice, i.e. it gets written to, flushed, and then the buckets
+	// get evicted from the map
+	defaultBucketPoolSize = defaultBucketsPoolSize
```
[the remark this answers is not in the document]
justinjc: The plural one is the list of different bucket versions per block. I've renamed this `bucketVersions` instead.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
-		// Needs reset
-		b.DrainAndReset()
+	if wType == ColdWrite && !b.coldWritesEnabled {
+		return m3dberrors.ErrColdWritesNotEnabled
 	}
 
-	return b.buckets[idx].write(timestamp, value, unit, annotation)
+	blockStart := timestamp.Truncate(b.blockSize)
+	buckets := b.bucketsAtCreate(blockStart)
+	b.putBucketInCache(buckets)
```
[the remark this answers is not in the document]
justinjc: Renamed `buckets` to `bucketVersions`

src/dbnode/storage/block/wired_list.go
```diff
@@ -315,6 +313,7 @@ func (l *WiredList) insertAfter(v, at DatabaseBlock) {
 				).Errorf("wired list tried to close a block that was not from disk")
 			})
 		}
+		bl.Close()
```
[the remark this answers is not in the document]
justinjc: Nice catch.

robskillington: Hm does this double close need removing?

src/dbnode/storage/namespace/options.go
```diff
@@ -85,14 +90,15 @@ func (o *options) Validate() error {
...
+		retention       = o.retentionOpts.RetentionPeriod()
+		futureRetention = o.retentionOpts.FutureRetentionPeriod()
+		dataBlockSize   = o.retentionOpts.BlockSize()
+		indexBlockSize  = o.indexOpts.BlockSize()
 	)
 	if indexBlockSize <= 0 {
 		return errIndexBlockSizePositive
 	}
-	if retention < indexBlockSize {
+	if retention+futureRetention < indexBlockSize {
```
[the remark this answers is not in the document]
robskillington: Yeah each needs to be verified.

src/dbnode/storage/series/buffer.go
```diff
@@ -89,90 +99,68 @@ type databaseBuffer interface {
...
-	blockSize         time.Duration
-	bufferPast        time.Duration
-	bufferFuture      time.Duration
+	opts           Options
+	nowFn          clock.NowFn
+	blockRetriever QueryableBlockRetriever
+
+	bucketsMap map[xtime.UnixNano]BufferBucketVersions
+	// Cache of buckets to avoid map lookup of above
+	bucketsCache       [bucketsCacheSize]BufferBucketVersions
```
robskillington: This should be an array of structs yeah? Otherwise they won't be alloc'd when this struct is alloc'd.

src/dbnode/storage/series/buffer.go
```diff
@@ -89,90 +99,68 @@ type databaseBuffer interface {
...
-	pastMostBucketIdx int
-	buckets           [bucketsLen]dbBufferBucket
-	blockSize         time.Duration
-	bufferPast        time.Duration
-	bufferFuture      time.Duration
+	opts           Options
+	nowFn          clock.NowFn
+	blockRetriever QueryableBlockRetriever
+
+	bucketsMap map[xtime.UnixNano]BufferBucketVersions
```
robskillington: We can just use direct pointers into the cache here yeah?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +169,271 @@ func (b *dbBuffer) Write(
...
-	idx := b.writableBucketIdx(timestamp)
-	if b.buckets[idx].needsReset(bucketStart) {
-		// Needs reset
-		b.DrainAndReset()
+	if wType == ColdWrite {
+		if !b.coldWritesEnabled {
+			return m3dberrors.ErrColdWritesNotEnabled
+		}
+
+		if now.Add(b.futureRetentionPeriod).Before(timestamp) {
```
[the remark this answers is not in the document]
robskillington: Ah yeah, where does the past check get done? I can't seem to see it.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
-	now := b.nowFn()
-	futureLimit := now.Add(1 * b.bufferFuture)
-	pastLimit := now.Add(-1 * b.bufferPast)
-	if !futureLimit.After(timestamp) {
-		return m3dberrors.ErrTooFuture
-	}
-	if !pastLimit.Before(timestamp) {
-		return m3dberrors.ErrTooPast
+	wType := wopts.WriteType
+	if wType == UndefinedWriteType {
```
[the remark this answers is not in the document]
robskillington: Can you add a method to `WriteOptions` that returns the resolved write type?

i.e. `wopts.ResolveWriteType()` or named similarly that will default it to what it is if it's undefined?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +169,271 @@ func (b *dbBuffer) Write(
...
+	blockStart := timestamp.Truncate(b.blockSize)
+	buckets := b.bucketsAtCreate(blockStart)
+	b.putBucketInCache(buckets)
+	return buckets.Write(timestamp, value, unit, annotation, wType)
 }
 
-func (b *dbBuffer) writableBucketIdx(t time.Time) int {
-	return int(t.Truncate(b.blockSize).UnixNano() / int64(b.blockSize) % bucketsLen)
-}
+func (b *dbBuffer) writeType(timestamp time.Time, writeTime time.Time) WriteType {
```
robskillington: Agreed, `now` is better than `writeTime`

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
-		if !b.buckets[i].canRead() {
-			continue
-		}
-		if i == writableIdx {
-			stats.openBlocks++
-		}
-		stats.wiredBlocks++
-	}
-	return stats
+	return ColdWrite
```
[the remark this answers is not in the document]
robskillington: Agreed, where does this get handled?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +169,271 @@ func (b *dbBuffer) Write(
...
+		blockStart := tNano.ToTime()
+		if !retriever.IsBlockRetrievable(blockStart) {
+			continue
+		}
 
-func bucketTick(now time.Time, b *dbBuffer, idx int, start time.Time) int {
-	// Perform a drain and reset if necessary
-	mergedOutOfOrderBlocks := bucketDrainAndReset(now, b, idx, start)
+		// Avoid allocating a new backing array
+		version := retriever.RetrievableBlockVersion(blockStart)
```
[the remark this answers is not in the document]
robskillington: Hm yeah agreed, passing in the latest flush version is probably a good idea.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
 
-func (b *dbBuffer) DrainAndReset() drainAndResetResult {
-	// Avoid capturing any variables with callback
-	mergedOutOfOrder := b.computedForEachBucketAsc(computeAndResetBucketIdx, bucketDrainAndReset)
-	return drainAndResetResult{
+		// Try to merge any out of order encoders to amortize the cost of a drain
+		r, err := buckets.merge(WarmWrite)
+		if err != nil {
+			log := b.opts.InstrumentOptions().Logger()
+			log.Errorf("buffer merge encode error: %v", err)
```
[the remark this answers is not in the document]
robskillington: We can probably skip the metric, if this happens to any significant degree it'll flood logs so bad.

There needs to be a code error for this to ever happen, merging should always work unless there's a code bug.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
+	if err != nil {
+		return nil, err
 	}
 
-	if b.buckets[idx].needsReset(start) {
-		// Reset bucket
-		b.buckets[idx].resetTo(start)
+	streams := buckets.streams(ctx, WarmWrite)
+	// We may need to merge again here because the regular merge method does
+	// not merge buckets that have different versions.
```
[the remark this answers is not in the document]
robskillington: Doesn't the latest version always have all the data?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +169,271 @@ func (b *dbBuffer) Write(
...
 
-	b.forEachBucketAsc(func(bucket *dbBufferBucket) {
-		if err != nil {
-			// Something already went wrong and we want to return the error to the caller
-			// as soon as possible instead of continuing to do work.
-			return
+	// In the majority of cases, there will only be one block to persist here.
+	// Only when a flush fails
+	for _, block := range blocks {
+		if block == nil {
```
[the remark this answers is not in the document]
robskillington: Hm why would there be nil coming back?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +169,271 @@ func (b *dbBuffer) Write(
...
-		// we stream from will mark their internal context as dependent on that of the passed
-		// context, and the Encoder's that we stream from actually perform a data copy and
-		// don't share a reference.
-		streams := bucket.streams(ctx)
-		if len(streams) != 1 {
-			// Should never happen as the call to merge above should result in only a single
-			// stream being present.
-			err = errMoreThanOneStreamAfterMerge
-			return
+		err = persistFn(id, tags, segment, checksum)
```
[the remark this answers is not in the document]
robskillington: Agreed - but I want to understand how it's possible to have the same blockStart?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +169,271 @@ func (b *dbBuffer) Write(
...
-		if !bucket.start.Before(end) {
+
+	b.forEachBucketAsc(func(bv BufferBucketVersions) {
+		if !bv.StartTime().Before(end) || !start.Before(bv.StartTime().Add(b.blockSize)) {
 			return
 		}
 
-		res = append(res, bucket.streams(ctx))
+		res = append(res, bv.Streams(ctx, WarmWrite))
+		res = append(res, bv.Streams(ctx, ColdWrite))
```
robskillington: Hm if there are no cold writes, wouldn't this put nil here?  Probably want to avoid for that case I assume?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +169,271 @@ func (b *dbBuffer) Write(
...
+	buckets := b.bucketsMap
+	keys := make([]xtime.UnixNano, len(buckets))
+	i := 0
+	for k := range buckets {
+		keys[i] = k
+		i++
+	}
+	sort.Slice(keys, func(i, j int) bool {
+		return keys[i].Before(keys[j])
+	})
```
robskillington: This anon function is going to allocate.

richardartoul: Could also just implement the sort interface to avoid the alloc no? Also, just for my own curiosity, how do you know it will alloc? Isn't there a chance the compiler knows it doesn't escape, or does it never do that?

src/dbnode/storage/series/buffer.go
robskillington: Sounds good.

src/dbnode/storage/series/buffer.go
robskillington: So although the func seems to be inlined and does not escape, the call to sort.Slice I believe makes `keys` escape:
```
r@series[juchan/bucket-map]$ go build -gcflags "-m" .  2>&1 | sort | fgrep "buffer.go:4"
./buffer.go:409:14: make([]"github.com/m3db/m3/vendor/github.com/m3db/m3x/time".UnixNano, 0, len(buckets)) escapes to heap
./buffer.go:413:12: keys escapes to heap
./buffer.go:413:19: (*dbBuffer).forEachBucketAsc func literal does not escape
./buffer.go:413:19: can inline (*dbBuffer).forEachBucketAsc.func1
...
```

justinjc: @richardartoul just tried to implement the sort interface, but that solution still makes the thing you pass to `sort.Sort` escape.

I'll keep a separate slice of sorted keys and maintain that.

src/dbnode/storage/series/buffer.go
```diff
@@ -89,90 +99,60 @@ type databaseBuffer interface {
...
 	return b
 }
 
-func (b *dbBuffer) Reset(opts Options) {
+func (b *dbBuffer) Reset(blockRetriever QueryableBlockRetriever, opts Options) {
 	b.opts = opts
 	b.nowFn = opts.ClockOptions().NowFn()
 	ropts := opts.RetentionOptions()
+	b.blockRetriever = blockRetriever
+	bucketsPoolOpts := pool.NewObjectPoolOptions().SetSize(defaultBucketsPoolSize)
```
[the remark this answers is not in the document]
justinjc: That makes more sense. I'll make these into global pools instead.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
+	mergedOutOfOrder := 0
+	retriever := b.blockRetriever
+	for tNano, buckets := range b.bucketsMap {
+		if blockStart := tNano.ToTime(); retriever.IsBlockRetrievable(blockStart) {
+			// Avoid allocating a new backing array
+			newBuckets := buckets.buckets[:0]
+			version := retriever.RetrievableBlockVersion(blockStart)
+			for _, bucket := range buckets.buckets {
+				bVersion := bucket.version
+				// TODO(juchan): deal with ColdWrite too
```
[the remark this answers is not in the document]
justinjc: One idea would be to change the `fileOpState` struct so that it maintains two versions - one for cold writes and one for warm writes. Then we can change:
`FlushState(blockStart time.Time) fileOpState`
to something like:
`FlushState(blockStart time.Time, wType WriteType) fileOpState`

src/dbnode/storage/series/buffer.go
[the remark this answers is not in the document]
justinjc: Good idea, done.

src/dbnode/storage/series/series.go
justinjc: The only time we overwrite an existing block here is if `OnRetrieveBlock` is called multiple times. In this scenario, it seems correct to overwrite the existing block, since the data received will be at least as recent as the existing block. Multiple calls to `OnRetrieveBlock` also overwrites an existing block in today's code.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +169,271 @@ func (b *dbBuffer) Write(
...
-	return result
-}
 
-func (b *dbBuffer) Snapshot(ctx context.Context, blockStart time.Time) (xio.SegmentReader, error) {
-	var (
-		res xio.SegmentReader
-		err error
-	)
+	// A call to flush can only be for warm writes.
+	blocks, err := buckets.ToBlocks(WarmWrite)
```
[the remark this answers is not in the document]
justinjc: Yep, makes sense. I've redone this function to return all streams instead to avoid converting encoded data into blocks.

src/dbnode/storage/series/buffer.go
```diff
@@ -505,87 +476,273 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
+		return buckets
+	}
 
-	bopts := b.opts.DatabaseBlockOptions()
-	encoder := bopts.EncoderPool().Get()
-	encoder.Reset(start, bopts.DatabaseBlockAllocSize())
+	return b.newBucketsAt(t)
+}
+
+func (b *dbBuffer) putBucketInCache(newBuckets BufferBucketVersions) {
```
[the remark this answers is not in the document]
justinjc: Wouldn't this allow the possibility of having two of the same pointer in the cache? For example, if you have a cache consisting of `A,B,C,D`, then access X, then Y, then X again, the cache will be `X,Y,X,A`. Shouldn't it be `X,Y,A,B` instead?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +161,281 @@ func (b *dbBuffer) Write(
...
-	return bufferTickResult{
-		mergedOutOfOrderBlocks: mergedOutOfOrder,
-	}
-}
+	mergedOutOfOrder := 0
+	retriever := b.blockRetriever
+	for tNano, buckets := range b.bucketsMap {
+		if blockStart := tNano.ToTime(); retriever.IsBlockRetrievable(blockStart) {
+			// Avoid allocating a new backing array
+			newBuckets := buckets.buckets[:0]
```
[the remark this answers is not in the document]
justinjc: Sorry, missed this comment earlier. This no longer applies after using flush state "snapshots".

src/dbnode/storage/series/buffer.go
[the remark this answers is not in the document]
robskillington: Is there any way we can just have one copy of the merge code and just call `mergeOrNoOp` right before we do anything that requires the data to be merged?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +167,288 @@ func (b *dbBuffer) Write(
...
+	var stream xio.SegmentReader
+	numStreams := len(streams)
+	if numStreams == 1 {
+		stream = streams[0]
+	} else {
+		// In the majority of cases, there will only be one stream to persist
+		// here. Only when a previous flush fails midway through a shard will
+		// there be buckets for previous versions. In this case, we need to try
+		// to flush them again, so we merge them together to one stream and
+		// persist it.
```
robskillington: Hm, similarly here can we find a way to reuse the merge method and just pass options to return merged and not replace the current data there?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +167,288 @@ func (b *dbBuffer) Write(
...
+			log := b.opts.InstrumentOptions().Logger()
+			log.Errorf(errBucketMapCacheNotInSync)
+			return nil
 		}
 
-		res = append(res, bucket.streams(ctx))
+		if streams := bv.streams(ctx, WarmWrite); streams != nil {
+			res = append(res, streams)
+		}
+		if streams := bv.streams(ctx, ColdWrite); streams != nil {
```
robskillington: nit: Slightly more correct for both these would be `streams := ...; len(streams) > 0  {  /* do append */ }`?

src/dbnode/storage/series/buffer.go
```diff
@@ -473,119 +460,326 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
+		if t.Equal(removeTime) {
+			b.inOrderBlockStarts = append(starts[:i], starts[i+1:]...)
+			return
+		}
+	}
+}
 
+// BufferBucketVersions is a container for different versions (from
+// different flushes to disk) of buffer buckets in the database.
+type BufferBucketVersions struct {
```
robskillington: Does this need to be an exported type?

src/dbnode/storage/series/buffer.go
```diff
@@ -473,119 +460,326 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
 
-func (b *dbBufferBucket) needsDrain(
-	now time.Time,
-	targetStart time.Time,
-) bool {
-	retentionOpts := b.opts.RetentionOptions()
-	blockSize := retentionOpts.BlockSize()
-	bufferPast := retentionOpts.BufferPast()
+// BufferBucket is a bucket in the buffer.
+type BufferBucket struct {
```
robskillington: Does this need to be exported?

src/dbnode/storage/series/reader.go
```diff
@@ -117,6 +117,18 @@ func (r Reader) readersWithBlocksMapAndBuffer(
...
+		// resultsBlock holds the results from one block. The flow is:
+		// 1) Look in the cache for metrics for a block.
+		// 2) If there is nothing in the cache, try getting metrics from disk.
+		// 3) Regardless of (1) or (2), look for metrics in the series buffer.
+		//
+		// It is important to look for data in the series buffer one block at
+		// a time within this loop so that the returned results contain data
+		// from blocks in chronological order. Failure to do this will result
+		// in an out of order error in the MultiReaderIterator on query.
+		resultsBlock := []xio.BlockReader{}
```
robskillington: Just to avoid this allocating if not necessary, can we use `var resultsBlock []xio.BlockReader`?

src/dbnode/storage/series/reader.go
```diff
@@ -126,39 +138,45 @@ func (r Reader) readersWithBlocksMapAndBuffer(
...
-					results = append(results, []xio.BlockReader{streamedBlock})
+					resultsBlock = append(resultsBlock, streamedBlock)
 					// NB(r): Mark this block as read now
 					block.SetLastReadTime(now)
 					if r.onRead != nil {
 						r.onRead.OnReadBlock(block)
 					}
 				}
-				continue
+				retrievedFromCache = true
```
robskillington: Hm interesting, I see the issue.  Quite "oh dear".

src/dbnode/storage/series/reader.go
```diff
@@ -126,39 +138,45 @@ func (r Reader) readersWithBlocksMapAndBuffer(
...
+		if seriesBuffer != nil {
+			bufferResults := seriesBuffer.ReadEncoded(ctx, blockAt, blockAt.Add(size))
+			// Multiple block results may be returned here (for the same block
+			// start) - one for warm writes and another for cold writes.
+			for _, bufferRes := range bufferResults {
+				resultsBlock = append(resultsBlock, bufferRes...)
+			}
 		}
+
+		results = append(results, resultsBlock)
```
robskillington: Perhaps only do this if `len(resultsBlock) > 0`?  Otherwise we'll have a lot of nil's for blocks that don't have any results, and each one of them gets sent over RPC.

src/dbnode/storage/series/reader.go
richardartoul: Haha yeah this was a fun one to debug...

src/dbnode/storage/series/buffer.go
justinjc: Sure, makes sense.

src/dbnode/storage/series/reader.go
justinjc: Sure.

src/dbnode/storage/series/buffer.go
justinjc: Since we want to have a database-level pool of these, we use this when passing pools via our Options.

src/dbnode/storage/series/buffer.go
justinjc: Since we want to have a database-level pool of these, we use this when passing pools via our Options.

src/cmd/services/m3dbnode/config/pooling.go
```diff
@@ -125,6 +125,16 @@ var (
 			refillLowWaterMark:  defaultRefillLowWaterMark,
 			refillHighWaterMark: defaultRefillHighWaterMark,
 		},
+		"bufferBucket": poolPolicyDefault{
+			size:                262144,
```
justinjc: Any thoughts on what a good default here would be?

src/cmd/services/m3dbnode/config/config_test.go
```diff
@@ -547,6 +555,14 @@ func TestConfiguration(t *testing.T) {
...
       initialBatchSize: 128
       maxBatchSize: 100000
+    bufferBucketPool:
+      size: 2
+      lowWatermark: 0.01
+      highWatermark: 0.02
+    bufferBucketVersionsPool:
+      size: 2
+      lowWatermark: 0.01
+      highWatermark: 0.02
```
robskillington: Maybe we set the watermarks to 0 since these are so low?

src/cmd/services/m3dbnode/config/pooling.go
robskillington: Hm, maybe just `65536`.

robskillington:
Hey this is looking a lot better, taking a final pass - just wanted to ask a question, did you notice that the server side write latency increased rather significantly after deploying?

Screen Shot 2019-04-06 at 4 12 29 PM

Have you seen this in the test cluster too?

justinjc:
Ah no, I didn't notice the server side write latency increasing after deploying. It doesn't look like there was much change when I deployed to the test cluster, I wonder what's causing the spikes in prod:
graph

I can deploy the exact build from prod to the test cluster to see if we see that behavior. The only difference between the build I deployed in test and the prod one is that my change was rebased off a different commit though.

src/dbnode/storage/series/types.go
```diff
@@ -283,3 +329,42 @@ func NewStats(scope tally.Scope) Stats {
...
+
+	// WarmWrite represents warm writes (within the buffer past/future window).
+	WarmWrite
+
+	// ColdWrite represents cold writes (outside the buffer past/future window).
+	ColdWrite
+)
+
+// WriteOptions define different options for a write.
+type WriteOptions struct {
```
Haijuncao: can we follow other Options pattern, where there is an interface, and use a package private struct to implement the WriteOptions interface? 

Consider making WriteOptions immutable as well, where SetXXX will produce a new option.

src/dbnode/storage/series/types.go
```diff
@@ -283,3 +329,42 @@ func NewStats(scope tally.Scope) Stats {
...
+	WriteType WriteType
+}
+
+// ResolveWriteType returns whether a write is a cold write or warm write.
+func (w *WriteOptions) ResolveWriteType(
```
Haijuncao: Can we make this method a NewWriteType function? Instead of a method on WriteOptions? Ideally, WriteOptions do not have much logic, just some defaulter/getter/setter.

Haijuncao: If you are close to land, it is ok not to make these changes in short notice, I would be happy to include the suggested change in my pr (that make use of the WriteOptions to pass down schema). Let me know which way you prefer.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +167,268 @@ func (b *dbBuffer) Write(
...
-	idx := b.writableBucketIdx(timestamp)
-	if b.buckets[idx].needsReset(bucketStart) {
-		// Needs reset
-		b.DrainAndReset()
+	if wType == ColdWrite {
+		if !b.coldWritesEnabled {
+			return false, m3dberrors.ErrColdWritesNotEnabled
+		}
+
+		if now.Add(-b.retentionPeriod).Truncate(b.blockSize).After(timestamp) {
```
robskillington: Hm, I think even if cold writes are enabled you want to only allow `now-retention` not `(now-retention).truncate(blocksize)`.

You still should only be able to write into the past as far back as retention is enabled for your namespace.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +167,268 @@ func (b *dbBuffer) Write(
...
+			return false, m3dberrors.ErrTooPast
+		}
+
+		if !now.Add(b.futureRetentionPeriod).Truncate(b.blockSize).Add(b.blockSize).After(timestamp) {
```
robskillington: Same here, we shouldn't be truncating by block size when checking the bounds.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +167,268 @@ func (b *dbBuffer) Write(
...
-			return
-		}
+		stream = encoder.Stream()
+		encoder.Close()
+	}
 
-		if !blockStart.Equal(bucket.start) {
-			return
-		}
+	segment, err := stream.Segment()
```
robskillington: Does this segment ever get finalized?  Seems like it should get finalized before this method finishes but it doesn't seem to?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +167,268 @@ func (b *dbBuffer) Write(
...
-			}
+
+		if streams := buckets.streams(ctx, WarmWrite); len(streams) > 0 {
+			res = append(res, block.NewFetchBlockResult(start, streams, nil))
 		}
-		if !found {
-			return
+
+		if streams := buckets.streams(ctx, ColdWrite); len(streams) > 0 {
+			res = append(res, block.NewFetchBlockResult(start, streams, nil))
```
robskillington: Hm don't think this is correct, we should only have a single `FetchBlockResult` for a single block `start` time (e.g. you should append the streams together then create one of these).

Maybe we should have some type of test somewhere that enforces this too (since it's obvious this isn't enforced by the tests, which it should be - an integration test would be good for this case obviously).

src/dbnode/storage/series/buffer.go
```diff
@@ -473,119 +440,346 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
-		if !start.Before(bucket.start.Add(blockSize)) || !bucket.start.Before(end) {
-			return
+
+		bv, exists := b.bucketVersionsAt(blockStart)
+		if !exists {
+			// Invariant violated. This means the keys in the bucket map does
+			// not match the sorted keys cache, which should never happen.
+			log := b.opts.InstrumentOptions().Logger()
+			log.Errorf(errBucketMapCacheNotInSync)
+			return nil
```
robskillington: Should return an error rather than returning nil here.

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +167,268 @@ func (b *dbBuffer) Write(
...
-		if !bucket.start.Before(end) {
-			return
+
+		bv, exists := b.bucketVersionsAt(blockStart)
+		if !exists {
+			// Invariant violated. This means the keys in the bucket map does
+			// not match the sorted keys cache, which should never happen.
+			log := b.opts.InstrumentOptions().Logger()
+			log.Errorf(errBucketMapCacheNotInSync)
+			return nil
```
robskillington: Should return an error rather than returning nil here.

src/dbnode/storage/series/buffer.go
robskillington: I see, makes sense.

src/dbnode/storage/series/buffer.go
```diff
@@ -473,119 +440,346 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
+	b.opts = opts
 	atomic.StoreInt64(&b.lastReadUnixNanos, 0)
-	b.drained = false
+	b.bucketPool = bucketPool
 }
 
-func (b *dbBufferBucket) finalize() {
-	b.resetEncoders()
-	b.resetBootstrapped()
+func (b *BufferBucketVersions) streams(ctx context.Context, wType WriteType) []xio.BlockReader {
```
robskillington: This method is an if/else on the `WriteType` but I see a few places you call this twice (to get a result for each WriteType) and then concat them together. 

Would it perhaps be better to create an options struct so you can just call it once for those cases? i.e.

```
type streamsOptions struct {
  filterWriteType bool
  writeType WriteType
}

func (b *BufferBucketVersions) streams(ctx context.Context, opts streamsOptions) []xio.BlockReader {
	var res []xio.BlockReader
	for _, bucket := range b.buckets {
		if !opts.filterWriteType || bucket.wType == opts.writeType {
			res = append(res, bucket.streams(ctx)...)
		}
	}

 	return res
}
```

src/dbnode/storage/series/buffer.go
```diff
@@ -473,119 +440,346 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
+	return res, nil
+}
+
+func (b *BufferBucketVersions) removeBucketsUpToVersion(version int) {
+	// Avoid allocating a new backing array.
+	nonEvictedBuckets := b.buckets[:0]
+
+	for _, bucket := range b.buckets {
+		bVersion := bucket.version
+		// TODO(juchan): deal with ColdWrite too.
```
robskillington: Can you elaborate in this comment what at least might need to happen with ColdWrite when we do want to come back and "deal with ColdWrite"?

src/dbnode/storage/series/buffer.go
justinjc: Makes sense.a

src/dbnode/storage/series/buffer.go
justinjc: Makes sense.

src/dbnode/storage/series/buffer.go
justinjc: Hmm yeah that makes sense. I think I was modeling based off of the `Flush` func in series [here](https://github.com/m3db/m3/blob/master/src/dbnode/storage/series/series.go#L617). That doesn't seem to finalize the segment either - is that different somehow or should that finalize the segment as well?

src/dbnode/storage/series/buffer.go
justinjc: I guess I've always questioned why we had these "should never happen" invariant violations logs as opposed to returning an error and dealing with it in the application. There are [other places](https://github.com/m3db/m3/blob/master/src/dbnode/storage/shard.go#L385) in which we log and don't return an error. Thoughts? Do you think it's better to change the function return and deal with the invariant violation here?

src/dbnode/storage/series/buffer.go
justinjc: See above comment.

src/dbnode/storage/series/buffer.go
justinjc: Ah thanks for catching this, I feel like this would have been an annoying bug to figure out later on. I'll include an integration test in a separate PR.

src/dbnode/storage/series/buffer.go
justinjc: Good idea, I like it.

src/dbnode/storage/series/types.go
justinjc: Agreed. This seems like a simple enough change; I'll make these changes today and ping you when done.

justinjc:
Just one more stat:
graph

The first deploy is for this bucket maps change, and the second deploy is current M3DB master. I don't see a big difference here either.

src/dbnode/storage/series/buffer.go
```diff
@@ -473,119 +440,346 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
+			return
+		}
+	}
+}
+
+// BufferBucketVersions is a container for different versions (from
+// different flushes to disk) of buffer buckets in the database.
+type BufferBucketVersions struct {
+	buckets           []*BufferBucket
+	start             time.Time
```
Haijuncao: what is this start time for? 

each BufferBucket struct has a start time as well.

how do you decide which buckets go into which version?

src/dbnode/storage/series/buffer.go
```diff
@@ -89,53 +99,50 @@ type databaseBuffer interface {
...
+	// Cache of buckets to avoid map lookup of above.
+	bucketVersionsCache [bucketsCacheSize]*BufferBucketVersions
+	// This is an in order slice of the block starts in the bucketsMap.
+	// We maintain this to avoid sorting the map keys adhoc when we want to
+	// perform operations in chronological order.
+	inOrderBlockStarts []time.Time
+	bucketVersionsPool *BufferBucketVersionsPool
+	bucketPool         *BufferBucketPool
+
+	blockSize             time.Duration
```
Haijuncao: should we replace term "block" with "bucket" everywhere inside Buffer?

src/dbnode/storage/series/buffer.go
```diff
@@ -181,287 +188,262 @@ func (b *dbBuffer) Write(
...
+		if !now.Add(b.futureRetentionPeriod).Truncate(b.blockSize).Add(b.blockSize).After(timestamp) {
+			return false, m3dberrors.ErrTooFuture
+		}
 	}
 
-	return b.buckets[idx].write(timestamp, value, unit, annotation)
+	blockStart := timestamp.Truncate(b.blockSize)
+	buckets := b.bucketVersionsAtCreate(blockStart)
+	b.putBucketVersionsInCache(buckets)
+	return buckets.write(timestamp, value, unit, annotation, wType)
```
Haijuncao: can we pass down writeOpts all the way down to bucket?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -860,8 +861,14 @@ func (s *service) Write(tctx thrift.Context, req *rpc.WriteRequest) error {
...
-		ctx, s.pools.id.GetStringID(ctx, req.NameSpace), s.pools.id.GetStringID(ctx, req.ID),
-		xtime.FromNormalizedTime(dp.Timestamp, d), dp.Value, unit, dp.Annotation,
+		ctx,
+		s.pools.id.GetStringID(ctx, req.NameSpace),
+		s.pools.id.GetStringID(ctx, req.ID),
+		xtime.FromNormalizedTime(dp.Timestamp, d),
+		dp.Value,
+		unit,
+		dp.Annotation,
+		nil,
```
Haijuncao: pass down default writeOpts (instead of nil).

src/dbnode/storage/series/types.go
```diff
@@ -283,3 +329,26 @@ func NewStats(scope tally.Scope) Stats {
...
+
+	// WarmWrite represents warm writes (within the buffer past/future window).
+	WarmWrite
+
+	// ColdWrite represents cold writes (outside the buffer past/future window).
+	ColdWrite
+)
+
+// WriteOptions define different options for a write.
+type WriteOptions interface {
```
richardartoul: Does this need to be constructed for each write or can we share one globally for all writes for a namespace? I'm just asking because we don't want to have to allocate a new interface for each write

Haijuncao: WriteOptions will be implemented by some concrete struct (we probably want to use pointer receiver)... there is no problem sharing the same opt per namespace. 

The important thing is that Options is immutable, so that it can be safely shared.

src/dbnode/network/server/tchannelthrift/node/service.go
Haijuncao: @justinjc ,  you can leave it as is (passing nil). I am going to construct writeOpt and pass it down

@richardartoul I will make sure namespace do not construct new writeopts unless schema has changed.

src/dbnode/storage/series/buffer.go
Haijuncao: nvm, please ignore that. I will pass down schema directly (so that we avoid allocating writeOpts for every write).

src/dbnode/storage/series/buffer.go
```diff
@@ -820,59 +987,118 @@ func (b *dbBufferBucket) merge() (mergeResult, error) {
...
+	streams []xio.SegmentReader,
+	opts Options,
+) (encoding.Encoder, time.Time, error) {
+	bopts := opts.DatabaseBlockOptions()
+	encoder := opts.EncoderPool().Get()
+	encoder.Reset(blockStart, bopts.DatabaseBlockAllocSize())
+	iter := bopts.MultiReaderIteratorPool().Get()
+	defer func() {
+		iter.Close()
+	}()
```
robskillington: nit: `defer iter.Close()` will achieve the same thing.

src/dbnode/storage/series/buffer.go
```diff
@@ -820,59 +987,118 @@ func (b *dbBufferBucket) merge() (mergeResult, error) {
...
-		newBlock.Reset(b.start, blockSize, encoder.Discard())
-
-		// The single encoder is already discarded, no need to call resetEncoders
-		// just remove it from the list of encoders
-		b.encoders = b.encoders[:0]
 		b.resetBootstrapped()
-
-		return discardMergedResult{newBlock, 0}, nil
+		// Already merged as a single encoder.
+		return b.encoders[0].encoder.Stream(), nil
```
robskillington: Need to register finalizer here from the context so that the copied segment will get returned.

e.g.
```go
stream := b.encoders[0].encoder.Stream()
ctx.RegisterFinalizer(stream)
return stream, nil
```

src/dbnode/storage/series/buffer.go
```diff
@@ -820,59 +987,118 @@ func (b *dbBufferBucket) merge() (mergeResult, error) {
...
+	// reset, and the merged encoder appended as the only encoder in the
+	// bucket.
+	if !b.hasJustSingleEncoder() {
+		return nil, errIncompleteMerge
+	}
 
-	newBlock := b.opts.DatabaseBlockOptions().DatabaseBlockPool().Get()
-	blockSize := b.opts.RetentionOptions().BlockSize()
-	newBlock.Reset(b.start, blockSize, merged.Discard())
+	return b.encoders[0].encoder.Stream(), nil
```
robskillington: Same here - need to register finalizer here from the context so that the copied segment will get returned.

e.g.
```go
stream := b.encoders[0].encoder.Stream()
ctx.RegisterFinalizer(stream)
return stream, nil
```

Notice that for the bootstrap case because it takes a context on the call to stream it will do the register finalizer step for you, but here we have to do it ourselves since we're dealing with just an encoder.

src/dbnode/storage/series/buffer.go
robskillington: So I added some comments to the buffer `toStream(...)` method.  If you make those changes then we don't need to finalizer here (which is better than actually finalizing here).

src/dbnode/storage/series/buffer.go
robskillington: Sounds good, yeah definitely one of those that will likely break in the future unless we add a test soon.  Thanks 👍

src/dbnode/storage/series/buffer.go
robskillington: Those places there is no way to actually return the error because the structure of the code (i.e. OnRetrieveBlock is an internal call for caching a block, so there is no RPC to ultimately return an error to at the top level of the calling method).

The invariant pattern is just so that if screwed up stuff actually does happen we can alert off of it.  We should still always surface an error to the caller as best as possible, rather than silently letting the operation finish.

src/dbnode/storage/series/buffer.go
robskillington: Added a response^

src/dbnode/storage/series/types.go
robskillington: Can we revert this to just be a struct?  

The options interfaces we have everywhere is when we want to share a single huge object with lots of long lived data structures (so they can all share the same ref to lots of fields).

These options though are passed down the stack and are not held on by any longly lived object so it's fine to keep as just a struct.

src/dbnode/storage/series/buffer.go
justinjc: Oh I see, thanks for explaining that.

src/dbnode/storage/series/buffer.go
justinjc: Cool, changed.

src/dbnode/storage/series/buffer.go
robskillington: Still need to take these truncate statements out, just FYI.

src/dbnode/storage/series/buffer.go
robskillington: Still need to take these truncate and add statements out, just FYI.

src/dbnode/storage/series/buffer.go
```diff
@@ -183,287 +183,263 @@ func (b *dbBuffer) Write(
...
 		}
-		if !bucket.start.Before(end) {
-			return
+
+		bv, exists := b.bucketVersionsAt(blockStart)
+		if !exists {
+			// Invariant violated. This means the keys in the bucket map does
+			// not match the sorted keys cache, which should never happen.
+			log := b.opts.InstrumentOptions().Logger()
+			log.Errorf(errBucketMapCacheNotInSync)
```
robskillington: Can you use the invariant violated method `EmitAndLogInvariantViolation(...)` in m3/x?
https://github.com/m3db/m3/blob/2228f1327a66bc9b0fcbc360b509099e382d8ba9/src/x/instrument/invariant.go#L71

That will guarantee a metric is emitted.

src/dbnode/storage/series/buffer.go
```diff
@@ -22,40 +22,38 @@ package series
...
-)
-
 const (
-	// bucketsLen is three to contain the following buckets:
-	// 1. Bucket before current window, can be drained or not-yet-drained
-	// 2. Bucket currently taking writes
-	// 3. Bucket for the future that can be taking writes that is head of
-	// the current block if write is for the future within bounds
-	bucketsLen = 3
+	errBucketMapCacheNotInSync = "[invariant violated] bucket map keys do not match sorted keys cache"
```
robskillington: When using `EmitAndLogInvariantViolation` you shouldn't add the prefix yourself:
https://github.com/m3db/m3/blob/2228f1327a66bc9b0fcbc360b509099e382d8ba9/src/x/instrument/invariant.go#L71

src/dbnode/storage/series/buffer.go
```diff
@@ -472,122 +448,360 @@ func (b *dbBuffer) FetchBlocksMetadata(
...
 		}
-		if !start.Before(bucket.start.Add(blockSize)) || !bucket.start.Before(end) {
-			return
+
+		bv, exists := b.bucketVersionsAt(blockStart)
+		if !exists {
+			// Invariant violated. This means the keys in the bucket map does
+			// not match the sorted keys cache, which should never happen.
+			log := b.opts.InstrumentOptions().Logger()
+			log.Errorf(errBucketMapCacheNotInSync)
```
robskillington: Here too, can you use the invariant violated method `EmitAndLogInvariantViolation(...)` in m3/x?
https://github.com/m3db/m3/blob/2228f1327a66bc9b0fcbc360b509099e382d8ba9/src/x/instrument/invariant.go#L71

That will guarantee a metric is emitted.

src/dbnode/network/server/tchannelthrift/node/service.go
justinjc: I've removed write options from this diff since this logic doesn't rely on that. We can discuss/work on that separately in another diff.

src/dbnode/storage/series/buffer.go
justinjc: Hmm I think the term block fits more in certain areas in the buffer. While buckets map to blocks, they don't really logically have a size to them, only a reference to the block start that it corresponds to. Buckets refer to a bucket of encoders, but the buffer also deals with blocks (of data), which is a different concept.

src/dbnode/storage/series/buffer.go
```diff
@@ -89,53 +99,50 @@ type databaseBuffer interface {
...
+	opts  Options
+	nowFn clock.NowFn
+
+	bucketsMap map[xtime.UnixNano]*BufferBucketVersions
+	// Cache of buckets to avoid map lookup of above.
+	bucketVersionsCache [bucketsCacheSize]*BufferBucketVersions
+	// This is an in order slice of the block starts in the bucketsMap.
+	// We maintain this to avoid sorting the map keys adhoc when we want to
+	// perform operations in chronological order.
+	inOrderBlockStarts []time.Time
```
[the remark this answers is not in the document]
justinjc: While buckets do have start times, those start times are referencing the starts of blocks, so I think we should be calling these what they are fundamentally referring to.

It would be convenient to use `time.Time` everywhere, since the helper functions are useful, but using `time.Time` in maps is not good practice since it contains timezone and other data, making equality comparison difficult.

src/dbnode/storage/series/buffer.go
```diff
@@ -89,53 +99,50 @@ type databaseBuffer interface {
...
-	buckets           [bucketsLen]dbBufferBucket
-	blockSize         time.Duration
-	bufferPast        time.Duration
-	bufferFuture      time.Duration
+	opts  Options
+	nowFn clock.NowFn
+
+	bucketsMap map[xtime.UnixNano]*BufferBucketVersions
+	// Cache of buckets to avoid map lookup of above.
+	bucketVersionsCache [bucketsCacheSize]*BufferBucketVersions
```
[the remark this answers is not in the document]
justinjc: For the vast majority of the case, we'll be constantly accessing the map for the current bucket (for real time metrics). After benchmarking, we saw that accessing maps was significantly more expensive than going through an array, so we cache the pointers of buckets that we will most likely be using.

The array size is 2 because at a block boundary, warm writes can be for either block.

PR #1084, opened 2018-10-15
Add Time and Timestamp

benraskin92:


src/query/block/scalar.go
```diff
@@ -85,11 +91,18 @@ func (b *Scalar) SeriesIter() (SeriesIter, error) {
 func (b *Scalar) Close() error { return nil }
 
 // Value returns the value for the scalar block
-func (b *Scalar) Value() float64 { return b.val }
+func (b *Scalar) Value(idx int) float64 {
+	var t time.Time
+	if idx != -1 {
```
benraskin92: This actually should never be the case so not sure if it's worth handling the error

src/query/block/scalar.go
```diff
@@ -57,23 +57,30 @@ func (b *Scalar) StepIter() (StepIter, error) {
...
-			vals: []float64{b.val},
-		},
+		meta:    b.meta,
+		s:       b.s,
 		numVals: steps,
 		idx:     -1,
 	}, nil
 }
 
+// ScalarFunc determines whether to return a constant or time t in seconds
```
nikunjgit: ScalarFunc determines the function to apply to generate the value at each step

src/query/block/scalar.go
nikunjgit: no need, it's fine if we panic on a bug here. Also, don't have the if block, and just assume idx >=0

src/query/block/scalar.go
```diff
@@ -57,23 +57,30 @@ func (b *Scalar) StepIter() (StepIter, error) {
...
+type ScalarFunc func(t time.Time) float64
+
 // SeriesIter returns a SeriesIterator
 func (b *Scalar) SeriesIter() (SeriesIter, error) {
 	bounds := b.meta.Bounds
 	steps := bounds.Steps()
 	vals := make([]float64, steps)
 	for i := range vals {
-		vals[i] = b.val
+		t, err := bounds.TimeForIndex(i)
```
nikunjgit: can be optimized. t := bounds.Start, and on every iteration, t = t.Add(bounds.StepSize)

src/query/functions/unconsolidatedfunc/timestamp.go
```diff
@@ -0,0 +1,130 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package unconsolidatedfunc
```
nikunjgit: just call it unconsolidated ?

src/query/functions/unconsolidatedfunc/timestamp.go
```diff
@@ -0,0 +1,130 @@
...
+		return err
+	}
+
+	for index := 0; iter.Next(); index++ {
+		step, err := iter.Current()
+		if err != nil {
+			return err
+		}
+
+		var values []float64
```
nikunjgit: intialize as len(step.Values()) ?

nikunjgit: also use Memset to set it to NaNs() and if len(dps) == 0 , continue

src/query/functions/unconsolidatedfunc/timestamp.go
```diff
@@ -0,0 +1,130 @@
...
+		for _, dps := range step.Values() {
+			if len(dps) == 0 {
+				values = append(values, math.NaN())
+				continue
+			}
+
+			values = append(values, float64(dps[len(dps)-1].Timestamp.Unix()))
```
nikunjgit: this is the only part related to timestamp. Maybe just consider making this a generic Node which takes in a function which returns a value

benraskin92: Will do this another time should we add another function to this package.

src/query/block/scalar.go
```diff
@@ -32,14 +32,14 @@ import (
 // allowing them to treat this as a regular block, while at the same time
 // having an option to optimize by accessing the scalar value directly instead
 type Scalar struct {
-	val  float64
 	meta Metadata
+	s    ScalarFunc
```
arnikola: Nit 'func' or 'scalarFn'

src/query/parser/promql/parse.go
```diff
@@ -153,10 +154,12 @@ func (p *parseState) walk(node pql.Node) error {
...
-		p.edges = append(p.edges, parser.Edge{
-			ParentID: p.lastTransformID(),
-			ChildID:  opTransform.ID,
-		})
+		if op.OpType() != scalar.TimeType {
+			p.edges = append(p.edges, parser.Edge{
+				ParentID: p.lastTransformID(),
+				ChildID:  opTransform.ID,
+			})
+		}
```
arnikola: Nit newline

src/query/block/scalar.go
```diff
@@ -57,23 +57,27 @@ func (b *Scalar) StepIter() (StepIter, error) {
...
-		},
+		meta:    b.meta,
+		s:       b.s,
 		numVals: steps,
 		idx:     -1,
 	}, nil
 }
 
+// ScalarFunc determines the function to apply to generate the value at each step
+type ScalarFunc func(t time.Time) float64
```
arnikola: This seems really hacky that the base op needs a function that takes in time even when it doesn't need to. Can you localize the changes required rather than adding functions that necessarily take unnecessary params?

src/query/block/scalar_test.go
```diff
@@ -41,13 +41,9 @@ var (
...
 func TestScalarBlock(t *testing.T) {
-	block := NewScalar(val, bounds)
+	block := NewScalar(func(_ time.Time) float64 { return val }, bounds)
 
-	if scalar, ok := block.(*Scalar); !ok {
-		require.FailNow(t, "unexpected type for new scalar blocks")
-	} else {
-		assert.Equal(t, val, scalar.Value())
-	}
+	require.IsType(t, block, &Scalar{})
```
arnikola: :+1:

src/query/functions/binary/binary.go
```diff
@@ -49,7 +51,7 @@ func processBinary(
 			return nil, errLeftScalar
 		}
 
-		lVal := scalarL.Value()
+		lVal := scalarL.Value(time.Time{})
```
arnikola: Feels very hacky; can you make time function scalar blocks a different type

src/query/functions/unconsolidated/timestamp.go
```diff
@@ -0,0 +1,131 @@
...
+		return err
+	}
+
+	for index := 0; iter.Next(); index++ {
+		step, err := iter.Current()
+		if err != nil {
+			return err
+		}
+
+		values := make([]float64, len(step.Values()))
```
arnikola: Nit: pull out step.values into a var

src/query/functions/unconsolidated/timestamp.go
```diff
@@ -0,0 +1,131 @@
...
+		ts.Memset(values, math.NaN())
+		for i, dps := range step.Values() {
+			if len(dps) == 0 {
+				continue
+			}
+
+			values[i] = float64(dps[len(dps)-1].Timestamp.Unix())
```
arnikola: Should this take the timestamp of the last non nan point or is it ok since they're unconsolidated

PR #1086, opened 2018-10-15
Verify checkpoint files are completely written out when checking if they exist

richardartoul:
The existing implementation would assume that if a checkpoint file existed on disk that everything was fine, but we found that in some cases it was possible to end up with a checkpoint file of size 0 (no digest). In that case, we want to proceed as if the checkpoint file did not exist as it was not completely written out.


src/dbnode/persist/fs/files.go
```diff
@@ -1126,8 +1127,31 @@ func NextIndexSnapshotFileIndex(filePathPrefix string, namespace ident.ID, block
 	return currentSnapshotIndex + 1, nil
 }
 
+// CompleteCheckpointFileExists returns whether a checkpoint file exists, and if so,
+// is it complete.
+func CompleteCheckpointFileExists(filePath string) (bool, error) {
+	f, err := os.Stat(filePath)
```
prateek: should this have the same `strings.Contains` check?

richardartoul: what would the check be? This one is specifically written to be used for checkpoint files, and if its not a checkpoint file you'll probably get an error anyways because its unlikely the file you're checking is exactly 4 bytes

src/dbnode/persist/fs/files_test.go
```diff
@@ -282,29 +282,38 @@ func TestTimeAndVolumeIndexFromFileSetFilename(t *testing.T) {
...
 
 func TestFileExists(t *testing.T) {
-	dir := createTempDir(t)
-	defer os.RemoveAll(dir)
 
-	shard := uint32(10)
-	start := time.Now()
-	shardDir := ShardDataDirPath(dir, testNs1ID, shard)
-	err := os.MkdirAll(shardDir, defaultNewDirectoryMode)
+	var (
```
prateek: mind making a new test which only tests the `checkpointFileSizeBytes` condition we're worried about and adding a little blurb about why it exists.

src/dbnode/persist/fs/write.go
```diff
@@ -40,6 +40,10 @@ import (
 	xtime "github.com/m3db/m3x/time"
 )
 
+const (
+	checkpointFileSizeBytes = 4
```
prateek: maybe have this and `digest/digestLen` be exported types and ensure they're the same in a test.

richardartoul: done

src/dbnode/persist/fs/files.go
robskillington: I think prateek just means, should you check that the filepath they're checking for definitely is a checkpoint file (i.e. make sure that it passes the checkpoint suffix)?  I think that makes sense.

src/dbnode/persist/fs/read_write_test.go
```diff
@@ -253,6 +253,12 @@ func TestSimpleReadWrite(t *testing.T) {
 	readTestData(t, r, 0, testWriterStart, entries)
 }
 
+func TestCheckpointFileSizeBytesSize(t *testing.T) {
+	// These values need to match so that the logic for determining whether
+	// a checkpoint file is complete or not remains correct.
+	require.Equal(t, digest.DigestLenBytes, CheckpointFileSizeBytes)
```
robskillington: Instead of this, why don't you just make `CheckpointFileSizeBytes = digest.DigestLenBytes`?

richardartoul: Yeah I discussed this with Prateek and we like the idea that the tests would bring if you changed one and not the other just to make the person think about the change more

src/dbnode/persist/fs/files.go
richardartoul: Ah ok, I guess I just wasn't worried about it because this function is harder to misuse but I'll add it

richardartoul: done

PR #1089, opened 2018-10-15
Add logs in m3aggregator election manager

cw9:
Also changed the default configuration to speed up m3agg election in simple docker pipeline

### DIFF
--- a/src/query/api/v1/handler/placement/delete_test.go
+++ b/src/query/api/v1/handler/placement/delete_test.go
@@ -78,7 +78,13 @@ func TestPlacementDeleteHandler_Safe(t *testing.T) {
 		var (
 			mockClient, mockPlacementService = SetupPlacementTest(t)
 			handlerOpts                      = NewHandlerOptions(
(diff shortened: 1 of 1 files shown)


src/query/api/v1/handler/placement/common.go
```diff
@@ -75,8 +75,11 @@ const (
 	// HeaderDryRun is the header used to specify whether this should be a dry run.
 	HeaderDryRun = "Dry-Run"
 
-	defaultM3AggMaxAggregationWindowSize = 5 * time.Minute
-	defaultM3AggWarmupDuration           = time.Minute
+	defaultM3AggMaxAggregationWindowSize = time.Minute
+	// defaultM3AggWarmupDuration configures the buffer to account for the delay
+	// of propagating aggregator placement to clients, usually needed when there is
+	// a large amount of clients sending traffic to m3aggregator.
+	defaultM3AggWarmupDuration = 0
```
robskillington: Do we set this in config?  If so then I guess we'll pin to 1min from config?

PR #1090, opened 2018-10-15
Add ActiveLogs() API to commitlog and use it in the CleanupManager

richardartoul:
This P.R allows us to distinguish between corrupt commitlogs and active commitlogs (which sometimes look corrupt because the header info hasn't been flushed yet), allowing us to delete corrupt commitlog files safely.

### DIFF
--- a/docs/m3db/architecture/engine.md
+++ b/docs/m3db/architecture/engine.md
@@ -181,7 +181,7 @@ The ticking process runs continously in the background and is responsible for a 
 
 #### Merging all encoders
 
-M3TSZ is designed for compressing time series data in which each datapoint has a timestamp that is larger than the last encoded datapoint. For monitoring workloads this works very well because every subsequent datapoint is almost always larger than the previous one. However, real world systems are messy and occasionally out of order writes will be received. When this happens, M3DB will allocate a new encoder for the out of order datapoints. The multiple encoders need to be merged before flushing the data to disk, but to prevent huge memory spikes during the flushing process we continuously merge out of order encoders in the background.
+M3TSZ is designed for compressing time series data in which each datapoint has a timestamp that is larger than the last encoded datapoint. For monitoring workloads this works very well because every subsequent datapoint is almost always chronologically after the previous one. However, real world systems are messy and occasionally out of order writes will be received. When this happens, M3DB will allocate a new encoder for the out of order datapoints. The multiple encoders need to be merged before flushing the data to disk, but to prevent huge memory spikes during the flushing process we continuously merge out of order encoders in the background.
 
 #### Removing expired / flushed series and blocks from memory
 
--- a/docs/m3db/architecture/engine.md
+++ b/docs/m3db/architecture/engine.md
@@ -181,7 +181,7 @@ The ticking process runs continously in the background and is responsible for a 
 
 #### Merging all encoders
 
-M3TSZ is designed for compressing time series data in which each datapoint has a timestamp that is larger than the last encoded datapoint. For monitoring workloads this works very well because every subsequent datapoint is almost always larger than the previous one. However, real world systems are messy and occassionally out of order writes will be received. When this happens, M3DB will allocate a new encoder for the out of order datapoints. The multiple encoders need to be merged before flushing the data to disk, but to prevent huge memory spikes during the flushing process we continuously merge out of order encoders in the background.
+M3TSZ is designed for compressing time series data in which each datapoint has a timestamp that is larger than the last encoded datapoint. For monitoring workloads this works very well because every subsequent datapoint is almost always larger than the previous one. However, real world systems are messy and occasionally out of order writes will be received. When this happens, M3DB will allocate a new encoder for the out of order datapoints. The multiple encoders need to be merged before flushing the data to disk, but to prevent huge memory spikes during the flushing process we continuously merge out of order encoders in the background.
 
 #### Removing expired / flushed series and blocks from memory
 
--- a/src/dbnode/integration/disk_cleanup_test.go
+++ b/src/dbnode/integration/disk_cleanup_test.go
@@ -76,7 +76,6 @@ func TestDiskCleanup(t *testing.T) {
 	}
 	writeDataFileSetFiles(t, testSetup.storageOpts, md, shard, fileTimes)
 	for _, clTime := range fileTimes {
-		// Need to generate valid commit log files otherwise cleanup will fail.
 		data := map[xtime.UnixNano]generate.SeriesBlock{
 			xtime.ToUnixNano(clTime): nil,
 		}
@@ -89,6 +88,15 @@ func TestDiskCleanup(t *testing.T) {
 	// and commit logs at now will be deleted
 	newNow := now.Add(retentionPeriod).Add(2 * blockSize)
 	testSetup.setNowFn(newNow)
+	// This isn't great, but right now the commitlog will only ever rotate when writes
+	// are received, so we need to issue a write after changing the time to force the
+	// commitlog rotation. This won't be required once we tie commitlog rotation into
+	// the snapshotting process.
+	testSetup.writeBatch(testNamespaces[0], generate.Block(generate.BlockConfig{
+		IDs:       []string{"foo"},
+		NumPoints: 1,
+		Start:     newNow,
+	}))
 
 	// Check if files have been deleted
 	waitTimeout := 30 * time.Second
--- a/src/dbnode/persist/fs/commitlog/commit_log.go
+++ b/src/dbnode/persist/fs/commitlog/commit_log.go
@@ -62,6 +62,7 @@ type completionFn func(err error)
 
 type commitLog struct {
 	writerState writerState
+	flushState  flushState
 	closedState closedState
 	// Associated with the closedState, but stored separately since
 	// it does not require the closedState lock to be acquired before
@@ -93,7 +94,6 @@ type writerState struct {
 	sync.RWMutex
 	writer         commitLogWriter
 	writerExpireAt time.Time
-	lastFlushAt    time.Time
 	activeFile     *File
 }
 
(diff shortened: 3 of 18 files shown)


src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -184,6 +189,21 @@ func (l *commitLog) Open() error {
 	return nil
 }
 
+func (l *commitLog) ActiveLogs() ([]File, error) {
```
justinjc: Why have this function return `[]File`? It seems like you return at most one `File`.

richardartoul: @prateek Was talking about having support for multiple commit log files at some point

docs/m3db/architecture/engine.md
```diff
@@ -181,7 +181,7 @@ The ticking process runs continously in the background and is responsible for a
...
+M3TSZ is designed for compressing time series data in which each datapoint has a timestamp that is larger than the last encoded datapoint. For monitoring workloads this works very well because every subsequent datapoint is almost always larger than the previous one. However, real world systems are messy and occasionally out of order writes will be received. When this happens, M3DB will allocate a new encoder for the out of order datapoints. The multiple encoders need to be merged before flushing the data to disk, but to prevent huge memory spikes during the flushing process we continuously merge out of order encoders in the background.
```
prateek: nit: do you want to say `almost always chronologically after the previous one` instead of `almost always larger than the previous one`

prateek: ps can't tell what changed in this line

src/dbnode/integration/disk_cleanup_test.go
```diff
@@ -89,6 +88,15 @@ func TestDiskCleanup(t *testing.T) {
 	// and commit logs at now will be deleted
 	newNow := now.Add(retentionPeriod).Add(2 * blockSize)
 	testSetup.setNowFn(newNow)
+	// This isn't great, but right now the commitlog will only ever rotate when writes
```
prateek: nice to see our tests actually catch this kinda thing

src/dbnode/storage/cleanup.go
```diff
@@ -286,26 +297,42 @@ func (m *cleanupManager) cleanupNamespaceSnapshotFiles(earliestToRetain time.Tim
...
-	// If we cared about 100% correctness, we would optimize this by retaining a smarter data
-	// structure (e.g. interval tree), but for our use-case, it's safe to assume that commit logs
-	// are only retained for a period of 1-2 days (at most), after we which we'd live we with the
-	// data loss.
+	namespaces, err := m.database.GetOwnedNamespaces()
+	if err != nil {
+		return nil, err
+	}
 
+	// We list the commit log files on disk before we determine what the currently active commitlog
```
prateek: +1 for explanation and example

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -213,32 +250,42 @@ func (l *commitLog) flushEvery(interval time.Duration) {
...
+	// is guaranteed to still be present), but it does need to acquire an exclusive lock when the writer
+	// is opened, closed, or set to nil (any operation that could change the pointer value of
+	// writerState.writer or writerState.activeFile). In other words, this function can be thought of
+	// as having an implied read lock at all times that is occasionally upgraded to an exclusive lock
+	// for the purpose of mutating the writerState.
 	for write := range l.writes {
 		// For writes requiring acks add to pending acks
 		if write.completionFn != nil {
-			l.pendingFlushFns = append(l.pendingFlushFns, write.completionFn)
+			l.flushState.pendingFlushFns = append(l.flushState.pendingFlushFns, write.completionFn)
```
prateek: isn't this racy? i.e. you're modifying a field on `flushState` without a write Lock on it. but you use a Read lock when accessing the same field on line 326.

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -61,30 +61,45 @@ type commitLogFailFn func(err error)
...
 	writerExpireAt time.Time
-	closed         bool
-	closeErr       chan error
+	activeFile     *File
+}
 
-	metrics commitLogMetrics
+type closedState struct {
+	sync.RWMutex
+	closed bool
```
prateek: could you replace the `closedState` with an atomic. It'll make a lot of the code simpler

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -61,30 +61,45 @@ type commitLogFailFn func(err error)
 type completionFn func(err error)
 
 type commitLog struct {
-	sync.RWMutex
+	flushState  flushState
```
prateek: please add unit test for some of the races

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -61,30 +61,88 @@ type commitLogFailFn func(err error)
...
+	closedState closedState
+	writerState writerState
+	flushState  flushState
+	// Associated with the closedState, but stored separately since
+	// it does not require the closedState lock to be acquired before
+	// being accessed.
+	closeErr chan error
+
+	// TODO(r): replace buffered channel with concurrent striped
+	// circular buffer to avoid central write lock contention.
```
robskillington: Hm you can probably remove this comment, I think we just need to encode in parallel rather than need a better buffer.

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -184,6 +228,34 @@ func (l *commitLog) Open() error {
 	return nil
 }
 
+func (l *commitLog) ActiveLogs() ([]File, error) {
+	l.closedState.Lock()
+	defer l.closedState.Unlock()
```
robskillington: nit: This just needs to be RLock() and RUnlock() yeah?  Don't see it changing the closed state as far as I can tell?

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -184,6 +228,34 @@ func (l *commitLog) Open() error {
...
+		valueType: activeLogsValueType,
+		valueTypeFn: func(f File, e error) {
+			err = e
+			file = f
+			wg.Done()
+		},
+	}
+
+	wg.Wait()
+	return []File{file}, err
```
robskillington: Do you want to bother doing the `if err != nil { return nil, err}; return []File{file}, nil` so you only return something if and only if `err == nil`?  Just slightly more idiomatic Go to do so.

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -58,33 +58,72 @@ type writeCommitLogFn func(
 ) error
 type commitLogFailFn func(err error)
 
-type completionFn func(err error)
+type valueTypeFn func(f File, err error)
```
robskillington: For better reuse, which we'll need in the future, can we just rename this to `callbackFn`?

And maybe the first arg it would be better to be some composite event struct, so we can add cleanly to it in the future:

```go
func callbackFn func(result callbackResult)

type callbackResultType uint

const (
  activeLogsCallback callbackResultType = iota
  // ... more in the future
)

type callbackResult struct {
  resultType callbackResultType
  err error
  activeLogs activeLogsCallbackResult // to be used if and only if eventType == activeLogsResultCallbackEventType
}

type activeLogsCallbackResult struct {
  file *File
}

func (r callbackResult) activeLogsCallbackResult() (activeLogsCallbackResult, error) {
  if expectedType := r.resultType; expectedType != activeLogsCallback {
    return activeLogsCallbackResult{}, fmt.Errorf("wrong result type: expected=%d, actual=%d", expectedType, r.resultType)
  }
  if r.err != nil {
    return activeLogsCallbackResult{}, err
  }
  return r.activeLogs, nil
}

// Now from calling code:
func foo() {
  var (
    result activeLogsCallbackResult
    err error
  )
  writes <- commitLogWrite{
    valueType: activeLogsValueType,
    callbackFn: func(r callbackResult) {
      result, err = r.activeLogsCallbackResult()
      wg.Done()
    },
  }
  if err != nil {
    return nil, err
  }
  return result.file, nil
}
```

src/dbnode/persist/fs/commitlog/writer.go
```diff
@@ -75,7 +75,15 @@ type commitLogWriter interface {
 		annotation ts.Annotation,
 	) error
 
-	// Flush will flush the contents to the disk, useful when first testing if first commit log is writable
+	// Sync will ensure that all writes that have been issued to the writer have been
+	// FSync'd to disk.
+	Sync() error
```
robskillington: Perhaps `FlushAndSync()` considering that's what it's doing?

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: done

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: yeah thats fair, Its kind of messed up to return a non-empty slice in the error case

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: Yeah I can make this change, the reason I didn't is that the Rotate() API will need pretty much the same result as the ActiveLogs one but this is fine too

richardartoul: `resultType` seems a little overkill, I think I can just repurpose the `valueType` for now to infer the result type

richardartoul: `resultType` seems a little overkill, I think I can just repurpose the `valueType` for now to infer the result type.

richardartoul: done

src/dbnode/persist/fs/commitlog/writer.go
richardartoul: I want to stop using the word Flush entirely in the public interface (because its a little non-sensical, in this case "flush" means "flushed to chunk writer" whereas all an external caller will care about is "FSync'd to disk")

Getting rid of the flush method is outside the scope of this P.R but I'd rather not put it in this method name

PR #1096, opened 2018-10-16
add some problems to the troubleshooting section

BertHartm:
a few things I've hit or seen asked and answered

### DIFF
--- a/docs/troubleshooting/index.md
+++ b/docs/troubleshooting/index.md
@@ -10,10 +10,10 @@ These ports will not open until a namespace and placement have been created and 
 
 Double check your configuration against the [bootstrapping guide](../operational_guide/bootstrapping.md). The nodes will log what bootstrapper they are using and what time range they are using it for.
 
-If you don't seem to be using the filesystem bootstrapper, ensure that snapshotting is enabled for your namespace.
+If you're using the commitlog bootstrapper, and it seems to be slow, ensure that snapshotting is enabled for your namespace. Enabling snapshotting will require a node restart to take effect.
 
-If your node has been unable to snapshot for a long period of time, and is stuck in the commitlog bootstrapper, the peers bootstrapper may help as it uses less memory. This will only work if you have other nodes with replicas to peer bootstrap from.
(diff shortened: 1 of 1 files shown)


docs/troubleshooting/index.md
```diff
@@ -1 +1,19 @@
...
+
+## Ports 9001-9004 aren't open after starting m3db.
+
+These ports will not open until a namespace and placement have been created and the nodes have bootstrapped.
+
+## Bootstrapping is slow
+
+Double check your configuration against the [bootstrapping guide](../operational_guide/bootstrapping.md). The nodes will log what bootstrapper they are using and what time range they are using it for.
+
+If you don't seem to be using the filesystem bootstrapper, ensure that snapshotting is enabled for your namespace.
```
richardartoul: I'm a little confused by this sentence. What do you mean by "if you don't seem to be using the filesystem bootstrapper"? Snapshotting actually makes the commitlog boostrapper faster, but doesn't have any impact on the filesystem bootstrapper (which is reading immutable files, not the snapshots)

docs/troubleshooting/index.md
```diff
@@ -1 +1,19 @@
...
+
+If your node has been unable to snapshot for a long period of time, and is stuck in the commitlog bootstrapper, the peers bootstrapper may help as it uses less memory. This will only work if you have other nodes with replicas to peer bootstrap from.
```
richardartoul: Just a little cleanup: 

> If an m3db node hasn't been able to snapshot for awhile, or is stuck in the commitlog bootstrapping phase for a long time due to accumulating a large number of commitlogs, consider using the peers bootstrapper. In situations where a large number of commitlogs need to be read, the peers bootstrapper will outperform the commitlog bootstrapper (faster and less memory usage) due to the fact that it will receive already-compressed data from its peers. Keep in mind that this will only work with a replication factor of 3 or larger and if the nodes peers are healthy and bootstrapped. Review the [bootstrapping guide](../operational_guide/bootstrapping.md) for more information.

and

docs/troubleshooting/index.md
```diff
@@ -1 +1,19 @@
...
+
+## Things aren't scaling well and there are weird memory issues
```
richardartoul: Lol this title is a little broad. Maybe something like: "Nodes are crashing with memory allocation errors, but there is plenty of available memory"

BertHartm:
Thanks, wording updated as suggested (and I re-read the bootstrapping guide more closely as a side effect, which is always helpful)

PR #1103, opened 2018-10-17
[coordinator] Add tag search completion

arnikola:
Initial commit to propose shape of necessary proto files


src/query/generated/proto/rpcpb/query.proto
```diff
@@ -90,3 +92,45 @@ message M3Segment {
 	int64 startTime = 3;
 	int64 blockSize = 4;
 }
+
+message SearchRequest {
+	bytes target                = 1;
+	SearchType type             = 2;
+	repeated bytes targetTags   = 3;
+	map<string, bytes> metadata = 4;
```
nikunjgit: remove metadata and just use grpc metadata ?

src/query/generated/proto/rpcpb/query.proto
```diff
@@ -90,3 +92,45 @@ message M3Segment {
...
+
+enum SearchType {
+	// Requests tag names and values
+	DEFAULT = 0;
+	// Requests tag names only
+	TAGNAME = 1;
+}
+
+message SearchResponse {
+	uint32 hits                    = 1;
```
nikunjgit: Rob says we don't use hits

src/query/generated/proto/rpcpb/query.proto
```diff
@@ -90,3 +92,45 @@ message M3Segment {
 	int64 startTime = 3;
 	int64 blockSize = 4;
 }
+
+message SearchRequest {
+	bytes target                = 1;
```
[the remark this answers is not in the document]
arnikola: Echoing statsdex convention; this is the 'query string' analogue

src/query/generated/proto/rpcpb/query.proto
arnikola: Nice, happy to drop.

src/query/generated/proto/rpcpb/query.proto
arnikola: Was considering this, sounds fair going forwards since this is only here to account for some specific internal fields

src/query/generated/proto/rpcpb/query.proto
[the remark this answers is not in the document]
arnikola: I'll try changing to tagmatchers for now and see how well it gels from the other side

src/query/generated/proto/rpcpb/query.proto
```diff
@@ -90,3 +92,45 @@ message M3Segment {
 	int64 startTime = 3;
 	int64 blockSize = 4;
 }
+
+message SearchRequest {
+	oneof matchers {
+		TagMatchers tagMatchers = 1;
+	}
+	SearchType type           = 2;
+	repeated bytes targetTags = 3;
```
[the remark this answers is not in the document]
arnikola: Sounds good

src/query/tsdb/remote/codecs.go
```diff
@@ -269,3 +269,10 @@ func decodeTagMatchers(rpcMatchers *rpc.TagMatchers) (models.Matchers, error) {
 
 	return models.Matchers(matchers), nil
 }
+
+// DecodeSearchRequest decodes rpc search request to read query and read options
+func DecodeSearchRequest(
+	req *rpc.SearchRequest,
+) (*storage.FetchQuery, error) {
+	return &storage.FetchQuery{}, nil
```
robskillington: nit: Can we declare `var errNotImplemented = errors.New("not implemented")` and `return errNotImplemented` here until implemented?  Then at least it is transparent about behavior.

src/query/tsdb/remote/compressed_codecs.go
```diff
@@ -434,3 +443,11 @@ func DecodeCompressedFetchResponse(
 		nil,
 	), nil
 }
+
+// EncodeToCompressedSearchResult encodes SearchResults to a compressed search result
+func EncodeToCompressedSearchResult(
+	results *storage.SearchResults,
+	iterPools encoding.IteratorPools,
+) (*rpc.SearchResponse, error) {
+	return nil, nil
```
robskillington: same nit: Can we declare `var errNotImplemented = errors.New("not implemented")` and `return errNotImplemented` here until implemented?  Then at least it is transparent about behavior.

src/query/tsdb/remote/server.go
```diff
@@ -113,3 +114,49 @@ func (s *grpcServer) Fetch(
...
+		select {
+		case pools = <-poolCh:
+		case err = <-errCh:
+			return err
+		}
+	}
+
+	response, err := EncodeToCompressedSearchResult(results, pools)
+	if err != nil {
+		logger.Error("unable to encode search result", zap.Error(err))
```
robskillington: I think you need to `return err` here or else it will try and send a nil message?

src/query/tsdb/remote/codecs.go
arnikola: Sure, wasn't too worried about it since this PR should implement the function before the meat of it is up for review :)

src/query/tsdb/remote/server.go
arnikola: Sgtm

src/query/tsdb/remote/compressed_codecs.go
arnikola: Similar comment in that this PR should eventually implement this function

src/query/generated/proto/rpcpb/query.proto
```diff
@@ -90,3 +92,44 @@ message M3Segment {
 	int64 startTime = 3;
 	int64 blockSize = 4;
 }
+
+message SearchRequest {
+	oneof matchers {
+		TagMatchers tagMatchers     = 1;
+	}
+	SearchType type               = 2;
+	repeated bytes filterNameTags = 3;
```
robskillington: nit: Maybe we could create another message called `SearchRequestOptions` that contains this `filterNameTags` field?  That way in the future we can combine all the search options fields together in one logical group.

Not a big deal, but from an API design angle it's a little more clean.

src/query/generated/proto/rpcpb/query.proto
```diff
@@ -90,3 +92,44 @@ message M3Segment {
...
+	bytes compressedTags = 2;
+}
+
+message TagProperties {
+	repeated TagProperty properties = 1;
+}
+
+message TagProperty {
+	bytes key             = 1;
+	uint32 weight         = 2;
```
robskillington: I don't think we need weight for now (we were only getting that because it was provided), can we remove this and add it back when we have a use for it?

src/query/storage/m3/storage.go
```diff
@@ -174,80 +173,73 @@ func (s *m3storage) FetchTags(
...
+	for i, result := range tagResult {
+		m, err := storage.FromM3IdentToMetric(result.ID, result.Iter, s.tagOptions)
+		if err != nil {
+			return nil, err
+		}
+
+		metrics[i] = m
+	}
+
+	cleanup()
```
robskillington: Can you move this to a `defer cleanup()` before `metrics := make(model.Metrics...` so that it will get called in the case of an early return in the for loop?

src/query/storage/m3/types.go
```diff
@@ -48,4 +49,18 @@ type Querier interface {
 		query *genericstorage.FetchQuery,
 		options *genericstorage.FetchOptions,
 	) (encoding.SeriesIterators, Cleanup, error)
+
+	SearchRaw(
```
robskillington: Hm, instead of FetchRaw and SearchRaw (which match the M3DB method names), maybe we should call these FetchCompressed and SearchCompressed?  In the future we may want to change the M3DB method names too, compressed is definitely a little more accurate than raw.

src/query/tsdb/remote/compressed_codecs.go
```diff
@@ -434,3 +444,90 @@ func DecodeCompressedFetchResponse(
...
+		tag := iter.Current()
+		tagName := tag.Name.Bytes()
+		for _, filter := range filterTagNames {
+			if bytes.Equal(filter, tagName) {
+				tags.Append(tag)
+			}
+		}
+	}
+
+	if err := iter.Err(); err != nil {
```
robskillington: Hm do we also close the iterator from somewhere?  Perhaps we need to `iter.Close()` after checking for `iter.Err()`?

src/query/generated/proto/rpcpb/query.proto
```diff
@@ -90,3 +92,44 @@ message M3Segment {
...
+	// Requests tag names and values
+	DEFAULT = 0;
+	// Requests tag names only
+	TAGNAME = 1;
+}
+
+message SearchResponse {
+	oneof value {
+		M3TagProperties compressed = 1;
+		TagProperties decompressed = 2;
```
robskillington: Hey, I just realized that it's a little more complicated that we first thought.  There's actually two APIs that we need.

There's basically two types of queries:
1. Find series by tags, similar to Prometheus "Find series by label matchers":
https://prometheus.io/docs/prometheus/latest/querying/api/#finding-series-by-label-matchers
2. Find tag values, similar to Prometheus "Querying label values"
https://prometheus.io/docs/prometheus/latest/querying/api/#querying-label-values

You can see for the second API, on our current query service we actually flatten the series together (you don't return any "per series" results):
```go
func renderSearchQueryResultsJSON(
	w http.ResponseWriter,
	result *index.MetadataTags,
) error {
	jw := streamjson.NewWriter(w)
	jw.BeginObject()

	// ...

	jw.BeginObjectField("tags")
	jw.BeginArray()
	// Sort for tests and deterministic output
	index.SortTagProperties(result.Tags)
	for _, tag := range result.Tags {
		jw.BeginObject()

		jw.BeginObjectField("key")
		jw.WriteString(tag.Key)

		// ...

		jw.BeginObjectField("values")

		jw.BeginArray()
		// Sort for tests and deterministic output
		sort.Strings(tag.Values)
		for _, value := range tag.Values {
			jw.WriteString(value)
		}
		jw.EndArray()

		jw.EndObject()
	}
	jw.EndArray()

	jw.EndObject()

	return jw.Close()
}
```

src/query/tsdb/remote/compressed_codecs.go
arnikola: Won't the underlying data be released in the cleanup() called in server.Search? If we close here we lose the data in `tags` don't we?

src/query/generated/proto/rpcpb/query.proto
arnikola: Nice :)

Actually had it deleted in latest WIP so works out

src/query/storage/m3/types.go
arnikola: Sounds good, touching enough of things around this area, I might as well

src/query/storage/m3/storage.go
arnikola: Good call

src/query/generated/proto/rpcpb/query.proto
arnikola: Ah gotcha, yeah I was trying to bridge between the current search endpoints we had in coordinator/query and the existing remote endpoint we have.

If I'm understanding correctly, you're saying that this PR satisfies what you described as query type 1, right? If it does neither I'll be a very sad boy haha

src/query/tsdb/remote/client.go
```diff
@@ -197,7 +188,114 @@ func (c *grpcClient) FetchTags(
...
+		return nil, err
+	}
+
+	request, err := encodeSearchRequest(query)
+	if err != nil {
+		return nil, err
+	}
+
+	// Send the id from the client to the remote server so that provides logging
+	id := logging.ReadContextID(ctx)
```
nikunjgit: tracing handles this automatically. Maybe add a todo to switch id propogation to opentracing?

src/query/pools/pool_wrapper.go
```diff
@@ -77,3 +79,34 @@ func (w *PoolWrapper) IteratorPools() (
 	w.watchersErr = append(w.watchersErr, watcherErr)
 	return false, nil, nil, watcher, watcherErr
 }
+
+// WaitForIteratorPools will block until iterator pools are available.
+// If given a timeout of 0, will block indefinitely
+func (w *PoolWrapper) WaitForIteratorPools(
```
nikunjgit: @richardartoul do you guys not need something similar ?

src/query/storage/completed_tags.go
```diff
@@ -0,0 +1,115 @@
...
+
+// MergeWith merges two CompleteTagsResults together, handling deduplication.
+// Once all merges are completed, Finalize() should be called to give sorted results.
+func (r *CompleteTagsResult) MergeWith(other *CompleteTagsResult) error {
+	nameOnly := r.CompleteNameOnly
+	if nameOnly != other.CompleteNameOnly {
+		return errors.New("cannot merge tag results with name only tag results")
+	}
+
+	if !r.merged {
```
nikunjgit: you don't need merged, just use if r.seenMap == nil ?

src/query/storage/completed_tags.go
```diff
@@ -0,0 +1,115 @@
...
+		r.merged = true
+		r.seenMap = make(map[string]int, len(r.CompletedTags))
+		for i, tag := range r.CompletedTags {
+			r.seenMap[string(tag.Name)] = i
```
nikunjgit: interestingly this is optimized by Go: https://github.com/golang/go/issues/3512

src/query/storage/completed_tags.go
```diff
@@ -0,0 +1,115 @@
...
+// Once all merges are completed, Finalize() should be called to give sorted results.
+func (r *CompleteTagsResult) MergeWith(other *CompleteTagsResult) error {
+	nameOnly := r.CompleteNameOnly
+	if nameOnly != other.CompleteNameOnly {
+		return errors.New("cannot merge tag results with name only tag results")
+	}
+
+	if !r.merged {
+		r.merged = true
+		r.seenMap = make(map[string]int, len(r.CompletedTags))
```
nikunjgit: seenMap can be better named, maybe to namePositionMap ?

src/query/storage/completed_tags.go
```diff
@@ -0,0 +1,115 @@
...
+	r.merged = false
+	for _, tag := range r.CompletedTags {
+		tag.finalize()
+	}
+
+	sort.Sort(alphabeticalTagSort(r.CompletedTags))
+}
+
+func (t *CompletedTag) mergeWith(other CompletedTag) {
+	if !t.merged {
```
nikunjgit: same comment about avoiding merged

src/query/storage/completed_tags.go
```diff
@@ -0,0 +1,115 @@
...
+
+import (
+	"bytes"
+	"errors"
+	"sort"
+)
+
+// MergeWith merges two CompleteTagsResults together, handling deduplication.
+// Once all merges are completed, Finalize() should be called to give sorted results.
+func (r *CompleteTagsResult) MergeWith(other *CompleteTagsResult) error {
```
nikunjgit: this is a bit weird design to me. Consider having a CompleteTagsBuilder which has this method. Instead of Finalize, the building can have Build() method which returns the CompleteTagsResult. This way you don't end up having things like seenMap defined on the CompleteTagsResult struct

src/query/storage/completed_tags.go
```diff
@@ -0,0 +1,115 @@
...
+	r.seenMap = nil
+	r.merged = false
+	for _, tag := range r.CompletedTags {
+		tag.finalize()
+	}
+
+	sort.Sort(alphabeticalTagSort(r.CompletedTags))
+}
+
+func (t *CompletedTag) mergeWith(other CompletedTag) {
```
nikunjgit: same comment about using a builder as above

src/query/storage/m3/multi_fetch_tags_result.go
```diff
@@ -23,19 +23,60 @@ package m3
...
 	xerrors "github.com/m3db/m3x/errors"
 )
 
-type multiFetchTagsResult struct {
+const initSize = 10
+
+type multiSearchResult struct {
 	sync.Mutex
-	result    *storage.SearchResults
+	seenIters []client.TaggedIDsIterator // track known iterators to avoid leaking
```
nikunjgit: again, seems like you need a builder here to avoid the need to have FinalResult()

arnikola: I'm going to leave this one for now so it more closely matches multi_fetch_result.go, but this is essentially a builder under a different name

src/query/server/server_test.go
```diff
@@ -192,8 +192,8 @@ func waitForServerHealthy(t *testing.T, port int) {
 }
 
 type queryServer struct {
-	reads int
-	mu    sync.Mutex
+	reads, searches, tagCompletes int
```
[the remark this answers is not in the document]
arnikola: Currently nothing; plans to use them once we have an endpoint hooked up, when we can call a query against that endpoint and verify these counts increase

src/query/tsdb/remote/client.go
arnikola: For rpc calls too? This is to send the request id over the wire so the server can track what happens to it

src/query/storage/completed_tags.go
arnikola: Nice, as far as I can tell this implies the general pattern is to use `map[string(name)]` rather than using
```
n := string(name)
map[n]
```
right?

src/query/storage/completed_tags.go
arnikola: Refactored a bit due to builder approach

src/query/storage/completed_tags.go
arnikola: Refactored it to use a builder. I think the builder approach is likely a little less efficient, but much cleaner to read through and use

src/query/storage/completed_tags.go
arnikola: Refactored here too

src/query/storage/completed_tags.go
nikunjgit: yup

src/query/storage/completed_tags.go
nikunjgit: less efficient how ?

arnikola: This approach is creating additional data structures to do the value merge rather than doing it on the tags themselves

Pretty minimal impact overall and this isn't a hot path so not worried

src/query/generated/proto/rpcpb/query.proto
```diff
@@ -90,3 +93,64 @@ message M3Segment {
...
+	repeated bytes values = 2;
+}
+
+message TagProperties {
+	repeated TagProperty properties = 1;
+}
+
+message SearchResponse {
+	oneof value {
+		TagProperties decompressed = 1;
```
robskillington: Hm for this case, there should really only be one tag value per tag key, should we create two types `TagProperty` for the `SearchResponse` case and `TagValues` for the `CompleteTagsResponse` case that has multiple tag values (to match the naming of the `TagNames` type)?

robskillington: Otherwise this looks good 👍

src/query/storage/completed_tags.go
```diff
@@ -0,0 +1,141 @@
...
+			builder := completedTagBuilder{}
+			builder.add(tag.Values)
+			b.tagBuilders[string(tag.Name)] = builder
+		}
+	}
+
+	return nil
+}
+
+type alphabeticalTagSort []CompletedTag
```
robskillington: nit: We usually call these things `nameOfTypeBySort` so this might be named `CompletedTagsByName`.  Since it's not exported it's not too big of an issue, but if you need to export it eventually that's the naming scheme we usually use.

arnikola: Sounds good

src/query/storage/m3/multi_fetch_tags_result.go
nikunjgit: ok

src/query/tsdb/remote/client.go
nikunjgit: yeah it does that for you

src/query/tsdb/remote/server.go
```diff
@@ -72,35 +77,32 @@ func (s *grpcServer) Fetch(
...
-		logger.Error("unable to decode fetch query", zap.Any("error", err))
+		logger.Error("unable to decode fetch query", zap.Error(err))
 		return err
 	}
 
-	result, cleanup, err := s.storage.FetchRaw(ctx, storeQuery, nil)
+	result, cleanup, err := s.storage.FetchCompressed(
+		ctx,
+		storeQuery,
+		storage.NewFetchOptions(),
```
nikunjgit: should we send signal on the kill channel by handling the ctx timeout ?

src/query/tsdb/remote/server.go
```diff
@@ -113,3 +115,55 @@ func (s *grpcServer) Fetch(
...
+	searchQuery, err := decodeSearchRequest(message)
+	if err != nil {
+		logger.Error("unable to decode search query", zap.Error(err))
+		return err
+	}
+
+	results, cleanup, err := s.storage.SearchCompressed(
+		ctx,
+		searchQuery,
+		storage.NewFetchOptions(),
```
nikunjgit: handle context timeout ?

src/query/tsdb/remote/server.go
```diff
@@ -72,35 +77,32 @@ func (s *grpcServer) Fetch(
...
+	)
 	defer cleanup()
 	if err != nil {
 		logger.Error("unable to fetch local query", zap.Error(err))
 		return err
 	}
 
-	available, pools, err, poolCh, errCh := s.poolWrapper.IteratorPools()
+	pools, err := s.poolWrapper.WaitForIteratorPools(poolTimeout)
```
nikunjgit: instead of using a separate pooltimeout, maybe use the ctx timeout ?

src/query/tsdb/remote/server.go
```diff
@@ -113,3 +115,55 @@ func (s *grpcServer) Fetch(
...
+		searchQuery,
+		storage.NewFetchOptions(),
+	)
+	defer cleanup()
+	if err != nil {
+		logger.Error("unable to search tags", zap.Error(err))
+		return err
+	}
+
+	pools, err := s.poolWrapper.WaitForIteratorPools(poolTimeout)
```
nikunjgit: use ctx timeout here ?

src/query/tsdb/remote/server.go
arnikola: I'd rather keep this separate; it should never really have to wait since the iterator pools should spin up very soon after startup

src/query/tsdb/remote/server.go
arnikola: I'd rather keep this separate; it should never really have to wait since the iterator pools should spin up very soon after startup

nikunjgit: okay, are you trying to have only on startup ? Then probably using sync.once might be a better way instead of always waiting on a timeout. Also, doesn't look like m3db guys are doing this anywhere. The reason I suggested ctx timeout is so that we have a consistent number per request instead of different functions defining different timeouts.

arnikola: Will change to only on startup. M3 blocks until the pools are intialized, but we want to start m3 asynchronously and take their pools when available.

PR #1104, opened 2018-10-17
Add sort and sort_desc

benraskin92:


src/query/parser/promql/parse.go
```diff
@@ -160,11 +160,15 @@ func (p *parseState) walk(node pql.Node) error {
 			}
 		}
 
-		op, err := NewFunctionExpr(n.Func.Name, argValues, stringValues)
+		op, ok, err := NewFunctionExpr(n.Func.Name, argValues, stringValues)
```
arnikola: Would prefer this check here for n.func.name and return nil if it's a sort, rather than adding awkward third return param

src/query/functions/linear/sort.go
```diff
@@ -0,0 +1,32 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package linear
+
+const (
+	// NB: Because Prometheus's sort and sort_desc only look at the last value,
+	// these functions are essentially noops in M3 as we don't support instant queries.
+
+	// SortType returns timeseries elements sorted by their values, in ascending order.
```
arnikola: This file seems unnecessary? Would prefer to keep these in parser, as this doesn't add too much context

PR #1105, opened 2018-10-17
Remove self-exclusion when peer bootstrapping

robskillington:
This removes the special casing of the origin node counting towards bootstrap consistency.  

For the case where you need to bootstrap a range, you really cannot rely on local data being a substitute for fulfilling that range at some later time in the bootstrap process.

Especially if we start using this API for repairs.


richardartoul:
@robskillington @prateek Do we need to get this merged now that I'm working on repairs?

robskillington:
As per conclusion of this issue, I'd like to close this for the time being: https://github.com/m3db/m3/issues/2458.

PR #1109, opened 2018-10-17
[query] api/placement: 404 on deleting absent inst

schallert:
Provides better feedback to clients making decisions based on delete
responses.

### DIFF
--- a/src/query/api/v1/handler/placement/delete.go
+++ b/src/query/api/v1/handler/placement/delete.go
@@ -126,6 +126,14 @@ func (h *DeleteHandler) ServeHTTP(serviceName string, w http.ResponseWriter, r *
 			return
 		}
 
+		_, ok := curPlacement.Instance(id)
+		if !ok {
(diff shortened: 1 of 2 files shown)


src/query/api/v1/handler/placement/delete.go
```diff
@@ -126,6 +126,14 @@ func (h *DeleteHandler) ServeHTTP(serviceName string, w http.ResponseWriter, r *
 			return
 		}
 
+		_, ok := curPlacement.Instance(id)
+		if !ok {
+			logger.Info("instance not found in placement", zap.String("instance", id))
+			err := fmt.Errorf("instance %s not found in placement", id)
+			xhttp.Error(w, err, http.StatusNotFound)
```
cw9: Should this be StatusBadRequest?

schallert: I wanted NotFound so that it's more clear to clients what went wrong, and that they don't have to retry since the instance they were trying to delete is already deleted

PR #1110, opened 2018-10-17
 Remove CacheAllMetadata caching policy

justinjc:


src/dbnode/storage/block/types.go
```diff
@@ -168,10 +168,6 @@ type DatabaseBlock interface {
 	// merged during Stream().
 	HasMergeTarget() bool
 
-	// IsRetrieved returns whether the block is already retrieved. Only
```
prateek: Hm, shouldn't this be required for some cache policies (e.g. LRU/RecentlyRead)?

src/dbnode/storage/shard.go
```diff
@@ -1451,18 +1445,16 @@ func (s *dbShard) FetchBlocksMetadata(
...
 		// API will only return active block metadata (mutable and cached)
 		// hence this call is invalid
 		return nil, nil, fmt.Errorf(
 			"fetch blocks metadata v1 endpoint invalid with cache policy: %s",
 			s.opts.SeriesCachePolicy().String())
 	}
 
-	// For v1 endpoint we always include cached blocks because when using
-	// CacheAllMetadata the blocks will appear cached
+	// For v1 endpoint we always include cached blocks
```
prateek: Should be able to remove the V1 endpoint code at this point too /cc @robskillington

prateek:
@justinjc could you update the buildkite pipelines to remove the integration test target for this policy

src/dbnode/storage/block/types.go
justinjc: I trusted the comment! I believe the other caching policies use `IsCachedBlock()` and "retrieved" means we have metadata but we may not have the data for it.

src/dbnode/storage/block/block.go
```diff
@@ -263,13 +263,6 @@ func (b *dbBlock) HasMergeTarget() bool {
 	return hasMergeTarget
 }
 
-func (b *dbBlock) IsRetrieved() bool {
```
richardartoul: Thank god, I never understood this. Does this mean you can remove the retriever field from the block as well as any logic dependent on that being present?

The following functions still seem to be using the retriever field:

```
streamWithRLock
resetNewBlockStartWithLock
resetRetrievableWithLock
```

We might even get a small perf boost cause this is will be million of less pointers for the G.C to chase

src/dbnode/storage/block/types.go
richardartoul: Yeah I think `IsRetrieved` was specific to that policy (you can see the same logic implemented in the `IsCachedBlock` function. I think we definitely want to get rid of this function and move towards using `IsCachedBlock` or honestly at this point we can delete `IsCachedBlock` as well and only use `WasRetrievedFromDisk since they will have the exact same logic at that point. The only place we use `IsCachedBlock` is in `series.FetchBlocksMetadata` and its just an optimization there (and not a particularly accurate one at that since it doesn't work in the `recently_read` policy)

@robskillington

src/dbnode/storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -343,11 +343,8 @@ func (s *peersSource) logFetchBootstrapBlocksFromPeersOutcome(
 // the series will either be held in memory, or removed from memory once
 // flushing has completed.
 // Once everything has been flushed to disk then depending on the series
-// caching policy the function is either done, or in the case of the
-// CacheAllMetadata policy we loop through every series and make every block
-// retrievable (so that we can retrieve data for the blocks that we're caching
-// the metadata for).
-// In addition, if the caching policy is not CacheAll or CacheAllMetadata, then
+// caching policy the function is either done.
```
richardartoul: This sentence terminates too early now

src/dbnode/storage/series/series.go
```diff
@@ -235,29 +226,9 @@ func (s *dbSeries) updateBlocksWithLock() (updateBlocksResult, error) {
 		}
 
 		if shouldUnwire {
-			switch cachePolicy {
```
richardartoul: So much cleanup :D

src/dbnode/storage/shard.go
richardartoul: delete all the things

src/dbnode/storage/block/block.go
```diff
@@ -263,13 +263,6 @@ func (b *dbBlock) HasMergeTarget() bool {
...
-	b.RLock()
-	retrieved := b.retriever == nil
-	b.RUnlock()
-	return retrieved
-}
-
 func (b *dbBlock) WasRetrievedFromDisk() bool {
```
richardartoul: Another random thing, I think we might be able to delete the `OnRetrieveBlock` method in the DatabaseBlock. I think its primary purpose was to allow us to separate the following two actions:

1) Resetting a block
and
2) Fetching the blocks data

Without the CacheAllMetadata policy, these two steps will never be separate so you can probably combine it all into one function call. For example in the series we do this:

```golang
b.ResetRetrievable(startTime, blockSize, s.blockRetriever, metadata)
	// Use s.id instead of id here, because id is finalized by the context whereas
	// we rely on the G.C to reclaim s.id. This is important because the block will
	// hold onto the id ref, and (if the LRU caching policy is enabled) the shard
	// will need it later when the WiredList calls its OnEvictedFromWiredList method.
	// Also note that ResetRetrievable will mark the block as not retrieved from disk,
	// but OnRetrieveBlock will then properly mark it as retrieved from disk so subsequent
	// calls to WasRetrievedFromDisk will return true.
	b.OnRetrieveBlock(s.id, tags, startTime, segment)
```

It seems like that could just become

```golang
b.Reset(startTime, blockSize, segment)
````

And then we could delete the `OnRetrieveBlock` and `ResetRetrievable` methods (and their helper methods like `resetRetrievableWithLock`) since they're only required in a world where data can be lazily fetched

Comment (richardartoul) [reply src/dbnode/storage/block/block.go]: @robskillington

File: src/dbnode/storage/shard.go (hunk 2 above)
Comment (prateek) [reply src/dbnode/storage/shard.go]: lets definitely def delete all V1 code/tests. Don't do the rename tho - the 1-1 parity between go methods and the thrift code is nice. Plus it means if we ever add v3 or w/e it'd be a lot more explicit

File: src/dbnode/client/session.go [hunk 7]
```diff
@@ -1988,18 +1957,8 @@ func (s *session) streamBlocksMetadataFromPeers(
...
-						peer, start, end, currPageToken, metadataCh, progress)
-				case FetchBlocksMetadataEndpointV2:
-					currPageToken, err = s.streamBlocksMetadataFromPeerV2(namespace, shardID,
-						peer, start, end, currPageToken, metadataCh, resultOpts, progress)
-				default:
-					// Should never happen - we validate the version before this function is
-					// ever called
-					err = xerrors.NewNonRetryableError(errInvalidFetchBlocksMetadataVersion)
-				}
+				currPageToken, err = s.streamBlocksMetadataFromPeerV2(namespace, shardID,
```
robskillington: nit: Maybe rename `streamBlocksMetadataFromPeerV2` to just `streamBlocksMetadataFromPeer`?

src/dbnode/client/session.go
```diff
@@ -2034,168 +1993,6 @@ func (s *session) streamBlocksMetadataFromPeers(
 // TODO(r): Delete this once we delete the V1 code path
 type pageToken interface{}
```
robskillington: We can delete this now too, since the page token will always be of the same type now (that is `[]byte`).

src/dbnode/client/session_fetch_bulk_blocks_test.go
```diff
@@ -1914,14 +1779,9 @@ type MockTChanNodes []*rpc.MockTChanNode
...
 	opts AdminOptions,
-	isV2 bool,
 ) {
 	for _, client := range c {
-		if isV2 {
-			expectFetchMetadataAndReturnV2(client, result, opts)
-		} else {
-			expectFetchMetadataAndReturn(client, result, opts)
-		}
+		expectFetchMetadataAndReturnV2(client, result, opts)
```
robskillington: nit: Maybe we can rename this from `expectFetchMetadataAndReturnV2` to just `expectFetchMetadataAndReturn`?

src/dbnode/client/session_fetch_bulk_blocks_test.go
```diff
@@ -2198,64 +2003,12 @@ func expectFetchMetadataAndReturnV2(
 type fetchMetadataReqMatcher struct {
 	shard        int32
 	limit        int64
-	pageToken    *int64
 	pageTokenV2  []byte
```
robskillington: nit: Maybe we can rename `pageTokenV2` to `pageToken` now?

src/dbnode/storage/block/block.go
```diff
@@ -55,7 +55,6 @@ type dbBlock struct {
 
 	mergeTarget DatabaseBlock
 
-	retriever  DatabaseShardBlockRetriever
 	retrieveID ident.ID
```
robskillington: You can remove `retrieveID` and `wasRetrievedFromDisk` too.

robskillington: Can you rename `retrieveID` since it's not used for retrieving any more, it's just used for accounting purposes, mainly in the wired list.

So just rename it to `seriesID` here and in the `wiredListEntry` it should also be renamed to `seriesID`.  

That sounds good?

src/dbnode/storage/block/block.go
```diff
@@ -186,26 +167,6 @@ func (b *dbBlock) Checksum() (uint32, error) {
 	return b.checksum, nil
 }
 
-func (b *dbBlock) OnRetrieveBlock(
```
richardartoul: @justinjc Now that you've deleted this method, how does the seriesID get set? I think you might need to make seriesID a required argument for the `Reset` method otherwise it will mess up the LRU caching policy

src/dbnode/storage/block/types.go
robskillington: Yeah let's remove it, don't think this is plausible now.

src/dbnode/storage/block/block.go
justinjc: As discussed offline, going to make a separate `ResetFromDisk` function that's basically `Reset` with setting the `seriesID`

src/dbnode/storage/types.go
```diff
@@ -371,7 +371,7 @@ type databaseShard interface {
 		starts []time.Time,
 	) ([]block.FetchBlockResult, error)
 
-	// FetchBlocksMetadataV2 retrieves blocks metadata.
+	// FetchBlocksMetadata retrieves blocks metadata.
 	FetchBlocksMetadataV2(
```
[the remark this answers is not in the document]
justinjc: Oops, I think we want to keep the "V2" in the name in this one because it's part of the chain from RPC code. I'll revert this comment.

PR #1115, opened 2018-10-19
[coordinator] Add all pprof endpoints

### LINKED ISSUE #1106: pprof handler in m3coordinator doesn't expose all endpoints
eg I can't seem to get a goroutine dump from the coordinator pprof endpoint.

prateek:
fixes #1106

### DIFF
--- a/src/query/api/v1/httpd/handler.go
+++ b/src/query/api/v1/httpd/handler.go
@@ -23,7 +23,7 @@ package httpd
 import (
 	"encoding/json"
 	"net/http"
-	"net/http/pprof"
+	_ "net/http/pprof"
 	"time"
 
 	clusterclient "github.com/m3db/m3/src/cluster/client"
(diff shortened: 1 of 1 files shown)


src/query/api/v1/httpd/handler.go
```diff
@@ -199,7 +198,7 @@ func (h *Handler) registerHealthEndpoints() {
 
 // Endpoints useful for profiling the service
 func (h *Handler) registerProfileEndpoints() {
-	h.Router.HandleFunc(pprofURL, pprof.Profile)
+	h.Router.PathPrefix("/debug/pprof/").Handler(http.DefaultServeMux)
```
benraskin92: Make this a `const` to match the pattern?

benraskin92: Actually i kind of prefer this instead

prateek: Yeah me too :) 

The const makes this harder to follow, not easier.

PR #1117, opened 2018-10-19
Fix flaky test: serialize/TestDecoderLifecycle

prateek:


src/x/serialize/decoder.go
```diff
@@ -37,6 +37,7 @@ var (
 type decoder struct {
 	checkedData checked.Bytes
 	data        []byte
+	nextCalls   int
```
prateek: i don't like doing it this way but I can't seem to come up with a combination of `length` and `remaining` that conveys the same information.

PR #1119, opened 2018-10-20
Use a default pool size of 10 in development for query

andrewmains12:
Another small tweak--I found that when I got panics running locally, the stack trace was huge due to the large number of goroutines we had spawned. Most of them were from the `readWorkerPool` and `writeWorkerPool` in query, since default config gives you 4096 threads. I added config to the local query config to shrink this down to a starting size of 10, which is much more manageable.

Let me know if I should propagate that through some of the other config files as well.

### DIFF
--- a/src/query/README.md
+++ b/src/query/README.md
@@ -46,7 +46,7 @@ Build m3coordinator binary:
 
 Run m3coordinator binary:
 
-    $ ./bin/m3coordinator --config.file docker/coordinator.yml
+    $ ./bin/m3coordinator -f src/query/config/m3query-dev-etcd.yml
 
 Run Prometheus Docker image:
 
--- a/src/query/config/m3query-dev-etcd.yml
+++ b/src/query/config/m3query-dev-etcd.yml
@@ -1,0 +1,62 @@
+# m3query configuration for local development setup. Mostly the same as m3query-local-etcd.yml, but using fewer
+# resources (threads primarily).
+
+listenAddress:
+  type: "config"
+  value: "0.0.0.0:7201"
+
+metrics:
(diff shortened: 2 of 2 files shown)


prateek:
This file is used in the default [dockerfile](https://github.com/m3db/m3/blob/master/docker/m3query/Dockerfile#L24 ) for `query` which makes me a bit apprehensive. Maybe start a new convention to have default development config files be `service-dev-<type>.yaml`, e.g. `m3query-dev-etcd.yaml`

andrewmains12:
> Maybe start a new convention to have default development config files be service-dev-<type>.yaml, e.g. m3query-dev-etcd.yaml

Good call! I'll move it.

src/query/config/m3query-local-etcd.yml
```diff
@@ -50,3 +50,10 @@ clusters:
       backgroundHealthCheckFailLimit: 4
       backgroundHealthCheckFailThrottleFactor: 0.5
 
+readWorkerPoolPolicy:
+  grow: true
```
arnikola: Consider setting grow to false if this is only used for Dev/local testing; grow is unbounded and can lead to the issue you're trying to avoid here by spawning lots of workers.

Local load under prom is pretty light so 10 wokers should plenty

richardartoul:
Agree with what @prateek said, I think people use these as their starting points so the values should be sane and then we should have a separate file for development if thats what we want.

andrewmains12:
Thanks for the reviews all; I updated it to use `m3query-dev-etcd.yml` (though I didn't introduce the convention for other services).

PR #1124, opened 2018-10-23
Add a very basic limit on number of resolved datapoints

andrewmains12:
The approach here is taken directly from prom code; we basically want to prevent queries which go from the beginning of time at a nanosecond step size.

The limit is configured by `limits` in query config:

```
limits:
  maxComputedDatapoints: 20000
```
[edit] It defaults to being disabled, so as to not disrupt existing queries. 

I've also added some basic tests for config, just to make sure my validation was working.

Let me know if you want any code anywhere different than where I have it; I did what seemed convenient and in the existing style, but I'm totally open to moving things.

### DIFF
--- a/src/query/api/v1/handler/prometheus/native/read.go
+++ b/src/query/api/v1/handler/prometheus/native/read.go
@@ -27,6 +27,7 @@ import (
 	"net/http"
 	"sort"
 
+	"github.com/m3db/m3/src/cmd/services/m3query/config"
 	"github.com/m3db/m3/src/query/api/v1/handler"
 	"github.com/m3db/m3/src/query/block"
 	"github.com/m3db/m3/src/query/executor"
@@ -57,8 +58,9 @@ var (
 
 // PromReadHandler represents a handler for prometheus read endpoint.
 type PromReadHandler struct {
-	engine  *executor.Engine
-	tagOpts models.TagOptions
+	engine    *executor.Engine
+	tagOpts   models.TagOptions
+	limitsCfg *config.LimitsConfiguration
 }
 
 // ReadResponse is the response that gets returned to the user
@@ -75,10 +77,12 @@ type blockWithMeta struct {
 func NewPromReadHandler(
 	engine *executor.Engine,
 	tagOpts models.TagOptions,
-) http.Handler {
+	limitsCfg *config.LimitsConfiguration,
+) *PromReadHandler {
 	return &PromReadHandler{
-		engine:  engine,
-		tagOpts: tagOpts,
+		engine:    engine,
+		tagOpts:   tagOpts,
+		limitsCfg: limitsCfg,
 	}
 }
 
@@ -94,6 +98,11 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	if params.Debug {
 		logger.Info("Request params", zap.Any("params", params))
+	}
+
+	if err := h.validateRequest(&params); err != nil {
+		xhttp.Error(w, err, http.StatusBadRequest)
+		return
 	}
 
 	result, err := h.read(ctx, w, params)
@@ -194,6 +203,23 @@ func (h *PromReadHandler) read(
 	}
 
 	return sortedBlocksToSeriesList(sortedBlockList)
+}
+
+func (h *PromReadHandler) validateRequest(params *models.RequestParams) error {
+	// Impose a rough limit on the number of returned time series. This is intended to prevent things like
+	// querying from the beginning of time with a 1s step size.
+	// Approach taken directly from prom.
+	numSteps := int64(params.End.Sub(params.Start) / params.Step)
+	if h.limitsCfg.MaxComputedDatapoints > 0 && numSteps > h.limitsCfg.MaxComputedDatapoints {
+		return fmt.Errorf(
+			"querying from %v to %v with step size %v would result in too many datapoints "+
+				"(end - start / step > %d). Either decrease the query resolution (?step=XX), decrease the time window, " +
+				"or increase the limit (`limits.maxComputedDatapoints`)",
+			params.Start, params.End, params.Step, h.limitsCfg.MaxComputedDatapoints,
+		)
+	}
+
+	return nil
 }
 
 func drainResultChan(resultsChan chan executor.Query) {
--- a/src/cmd/services/m3query/config/config.go
+++ b/src/cmd/services/m3query/config/config.go
@@ -43,6 +43,12 @@ const (
 	M3DBStorageType BackendStorageType = "m3db"
 )
 
+// defaultLimitsConfiguration is applied if `limits` isn't specified.
+var defaultLimitsConfiguration = &LimitsConfiguration{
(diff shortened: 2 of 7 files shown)


src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -75,10 +82,15 @@ type blockWithMeta struct {
 func NewPromReadHandler(
 	engine *executor.Engine,
 	tagOpts models.TagOptions,
-) http.Handler {
+	limitsCfg *config.LimitsConfiguration,
+) *PromReadHandler {
+	if limitsCfg == nil {
```
nikunjgit: do this in the server.go method instead of the handler ? I'd assume other handlers might need the limitsconfiguration too

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -196,6 +213,21 @@ func (h *PromReadHandler) read(
 	return sortedBlocksToSeriesList(sortedBlockList)
 }
 
+func (h *PromReadHandler) validateRequest(params *models.RequestParams) error {
+	// For safety, limit the number of returned points per timeseries.
+	// This is sufficient for 60s resolution for a week or 1h resolution for a year.
```
nikunjgit: remove ?

src/query/api/v1/handler/prometheus/native/read.go
andrewmains12: Good call, will move.

andrewmains12: Hmm looking at `server.go`, there's not a super obvious place in `Run` to put it. Should I start an `applyDefaults` function or something to that effect? Would be good to centralize that.

andrewmains12: I put just above handler construction for now--probably not ideal, so lmk if you have any suggestions on that front.

src/query/api/v1/handler/prometheus/native/read.go
andrewmains12: done

src/query/server/server.go
```diff
@@ -205,6 +210,11 @@ func Run(runOpts RunOptions) {
 
 	engine := executor.NewEngine(backendStorage, scope.SubScope("engine"))
 
+	// TODO: centralize application of defaults?
+	if cfg.Limits == nil {
```
arnikola: General pattern seems to be using a generator for the configs which is passed a cfg.Limits object which it then can decode with defaults as appropriate

andrewmains12: Also, definitely not going to add it here, but any thoughts on https://github.com/imdario/mergo? Seems like a natural fit for this kind of thing (and avoids all of the `if field == empty` boilerplate).

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -75,10 +77,12 @@ type blockWithMeta struct {
 func NewPromReadHandler(
 	engine *executor.Engine,
 	tagOpts models.TagOptions,
-) http.Handler {
+	limitsCfg *config.LimitsConfiguration,
```
robskillington: Can you either do a nil check and return error if this is not set, or just pass it by value?  

Otherwise if someone doesn't set the limits config then we'll get a nil exception when we call validate request while handling a request at runtime I believe?

src/query/server/server.go
```diff
@@ -72,6 +72,11 @@ var (
 		Namespace: "default",
 		Retention: 2 * 24 * time.Hour,
 	}
+
+	defaultLimitsConfiguration = &config.LimitsConfiguration{
+		// from prom: This is sufficient for 60s resolution for a week or 1h resolution for a year.
+		MaxComputedDatapoints: 11000,
```
robskillington: Hey since we have a lot of users using our default retention of 4 weeks (720h/30d) which is on a lot of our guides, and also a similar retention that we run with internally, can you perhaps make that fit that case?  

i.e. 1m at 30 days, so 43,200 max computed datapoints?

andrewmains12: Absolutely, good call!

src/query/server/server.go
andrewmains12: Ah cool thanks, will go with that pattern.

src/query/server/server.go
andrewmains12: Perhaps we should make it larger than that (86400--double)? Since it's such a crude limit, I'd rather folks not hit it unless they're doing something really awful (would be an annoying UX and probably would not be what saved us from anything)

src/query/server/server.go
andrewmains12: Doh nvm; misread you.

andrewmains12: Slightly broader discussion: any thoughts on introducing `base.yml` (or `defaults.yml`) files for services, and using `xconfig.LoadFiles` to do the merging? Things I like about that approach:

- Defaults are very explicitly defined, and in a central location (and thus easy to find).
- Avoids unnecessary boiler plate from getter methods or `if cfg == nil` checks throughout the code.

cc. @arnikola

src/query/server/server.go
andrewmains12: Set it at 86400 for the moment (which also has the nice property of being == 1s for 1 day); lmk if we should do a different number!

src/query/server/server.go
andrewmains12: FYI: ended up making limits off by default (so as to not disrupt anybody's existing expensive queries), which had the side effect of simplifying the defaults. `<= 0` now implies no limit, which is hopefully intuitive.

src/query/server/server.go
andrewmains12: After a bit more discussion, I actually disabled this by default, so as to not disrupt existing expensive queries. We can flag it on in our internal deployment soon (will keep an eye on 400's).

src/query/api/v1/handler/prometheus/native/read.go
arnikola: nit: just wondering why this is a pointer, I think we usually try to keep configs immutable

andrewmains12: Ahh yeah, this is one of my pet peeves with golang--the guidance on when to use pointers vs not is unclear imo. My intent with using a pointer here isn't to allow mutation (totally agreed with leaving configs immutable after construction), but more to prevent unnecessary copying. Now, you could very reasonably argue that this struct is tiny, making copying quite cheap, and you'd be correct. However, we'll almost certainly be adding fields here, and I in general don't know when the size cutoff makes it better to pass structs by value instead of by reference. As a result, I tend to pass most things by reference unless they're explicitly going to remain small (`Datapoint` is a good example of this). 

Very welcome to other arguments on this, and sorry for the longish response! This is definitely something I'd love more clear guidance on :)

arnikola: Sure, no worries! I think in situations like this where we only use it once on server startup it's ok to copy the struct to get immutability

I think configs in general we copy where possible

PR #1131, opened 2018-10-25
Add FOSSA license scan to CI

robskillington:
Thanks to @liftM for the help getting this to work cleanly. 

Once we have a regular passing scan we'll start using `fossa-test` to fail builds.


.fossa.yml
```diff
@@ -0,0 +1,82 @@
+# Generated by FOSSA CLI (https://github.com/fossas/fossa-cli)
```
prateek: Created a followup about keeping this in sync with the makefile (#1138)

prateek:
Looking at the buildfailures, it appears to be caused by retool failing to pull down the tools. Might need to turn on the retool logging to figure it out (we redirect retool build/sync output atm)

PR #1132, opened 2018-10-25
[Aggregator Client] Drain queue asynchronously when closing writer

cw9:
Currently closing a writer is blocked on draining the queue within it. When the size of the queue is large, it could take a long time for it to drain. Which would cause the writer close to take a long time.

On the other hand the writer manager needs to close the writer with a lock when an instance is removed from the placement. A delayed writer close could block writes from acquiring the lock.

This changes changes the `RemoveInstance` behavior to synchronously remove the writer from its map and calls `writer.Close()` asynchronously.

### DIFF
--- a/src/aggregator/client/queue.go
+++ b/src/aggregator/client/queue.go
@@ -101,7 +101,7 @@ type instanceQueue interface {
 	// Enqueue enqueues a data buffer.
 	Enqueue(buf protobuf.Buffer) error
 
-	// Close closes the queue.
+	// Close closes the queue, it blocks until the queue is drained.
 	Close() error
 }
 
--- a/src/aggregator/client/writer.go
+++ b/src/aggregator/client/writer.go
@@ -147,10 +147,19 @@ func (w *writer) Close() error {
 		return errInstanceWriterClosed
 	}
 	w.closed = true
-	if err := w.flushWithLock(); err != nil {
+
+	err := w.flushWithLock()
+	if err != nil {
 		w.metrics.flushErrors.Inc(1)
+		w.log.Errorf("error flushing writer, %v", err)
 	}
-	return w.queue.Close()
+	go func() {
+		if err := w.queue.Close(); err != nil {
+			w.log.Errorf("error closing queue, %v", err)
+		}
+	}()
+
+	return err
 }
 
 func (w *writer) encodeWithLock(
--- a/src/aggregator/client/writer.go
+++ b/src/aggregator/client/writer.go
@@ -148,18 +148,18 @@ func (w *writer) Close() error {
 	}
 	w.closed = true
 
-	err := w.flushWithLock()
-	if err != nil {
-		w.metrics.flushErrors.Inc(1)
-		w.log.Errorf("error flushing writer, %v", err)
-	}
 	go func() {
+		if err := w.flushWithLock(); err != nil {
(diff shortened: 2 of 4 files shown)


src/aggregator/client/queue.go
```diff
@@ -101,7 +101,7 @@ type instanceQueue interface {
 	// Enqueue enqueues a data buffer.
 	Enqueue(buf protobuf.Buffer) error
 
-	// Close closes the queue.
+	// Close closes the queue, it blocks until the queue is drained.
```
richardartoul: Isnt this comment wrong? You're making it so it doesn't block.

richardartoul: Oh nevermind i see what you did

src/aggregator/client/writer.go
```diff
@@ -147,10 +147,19 @@ func (w *writer) Close() error {
 		return errInstanceWriterClosed
 	}
 	w.closed = true
-	if err := w.flushWithLock(); err != nil {
-		w.metrics.flushErrors.Inc(1)
-	}
-	return w.queue.Close()
+
+	go func() {
+		if err := w.flushWithLock(); err != nil {
```
robskillington: Hm you no longer hold the lock in this goroutine.  I think you might want to `w.Lock()` and `defer w.Unlock()` from inside the async goroutine too?

robskillington: Either that or conditionally unlock, and if you get to this goroutine, `defer w.Unlock()` from the top of the goroutine execution start?

robskillington: I assume that `w.queue.Close()` you can safely do from an async goroutine, but `flushWithLock()` it doesn't look like it.

Maybe you can rename `flushWithLock()` to `flushWithEncodersAndQueue(...)` take the params it works on as arguments then nil them out during close and pass them yourself async from the goroutine?

richardartoul:
Would this P.R be simpler to reason about if we kept the Writer.Close() method synchronous and made the writerManager kick off the close in a background goroutine after deleting it from its internal map? @chaowang

cw9:
> Would this P.R be simpler to reason about if we kept the Writer.Close() method synchronous and made the writerManager kick off the close in a background goroutine after deleting it from its internal map? @chaowang

It does seem simpler to do it this way, I'll give it a try

src/aggregator/client/writer.go
cw9: flushWithLock should also be safe because `w.closed` was set to true, so no other writes will be accepted after that, but Richie's suggestion seems cleaner to do it asynchronously in `mgr.RemoveInstance()`, I'll go with that route instead

src/aggregator/client/writer.go
```diff
@@ -464,5 +464,5 @@ func newRefCountedWriter(instance placement.Instance, opts Options) *refCountedW
 }
 
 func (rcWriter *refCountedWriter) Close() {
-	rcWriter.instanceWriter.Close() // nolint: errcheck
+	go rcWriter.instanceWriter.Close() // nolint: errcheck
```
robskillington: Can you add a `NOTE`/`NB` here to why this is asynchronous?

cw9: Sure, sounds good

Issue #996, opened 2018-10-01
Health endpoints should start before topology client initialization


schallert:
We're currently waiting for DB topology to be initialized before starting up the thrift + HTTP services, which in turn serve `/health` and `/bootstrapped` endpoints. This creates a circular dependency in cases where an orchestration system may not try to create an initial placement until a host responds to `/health` or `/bootstrapped`.

It would be best if these endpoints were available as soon as the process came up.

cc @richardartoul @prateek

schallert:
One other thing to add on here that I just noticed: because these endpoints don't start until after DB initialization, we also can't scrape metrics until they're up, which can be tricky for debugging.

PR #1134, opened 2018-10-25
[scripts] bootstrapped: healthy if no namespaces

schallert:
Ref #996

### DIFF
--- a/scripts/m3dbnode_bootstrapped.sh
+++ b/scripts/m3dbnode_bootstrapped.sh
@@ -48,11 +48,23 @@ if [ "$RES" -ne 0 ]; then
   exit 0
 fi
 
-curl -sSf "$COORD_NAMESPACE_ENDPOINT" >/dev/null
+curl -sSf -o "$COORD_TMPFILE" "$COORD_NAMESPACE_ENDPOINT"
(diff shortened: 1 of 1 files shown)


scripts/m3dbnode_bootstrapped.sh
```diff
@@ -48,14 +48,26 @@ if [ "$RES" -ne 0 ]; then
...
-curl -sSf "$COORD_NAMESPACE_ENDPOINT" >/dev/null
+curl -sSf -o "$COORD_TMPFILE" "$COORD_NAMESPACE_ENDPOINT"
 RES=$?
 
 if [ "$RES" -eq 22 ]; then
   echo "Received 4xx from namespace endpoint $COORD_NAMESPACE_ENDPOINT"
   exit 0
 fi
 
+# If there IS a placement but NO namespace, then the dbnode will respond
```
prateek: Hm how does this distinguish from the time when there is a namespace in etcd but it hasn't yet been picked up by the node at startup?

schallert: Good point, I hadn't thought of that. Any advice to get around the issue without waiting for #996?

PR #1136, opened 2018-10-26
[query] Add tag completion HTTP endpoint onto query

arnikola:
Adds the endpoint to query which will allow tag completion


src/query/api/v1/handler/prometheus/tags/common.go
```diff
@@ -0,0 +1,154 @@
...
+	jw.EndArray()
+
+	jw.EndObject()
+
+	return jw.Close()
+}
+
+func renderResultsJSON(
+	w io.Writer,
+	result storage.CompleteTagsResult,
```
andrewmains12: Nit: looks like the result type is passed around as a pointer everywhere else; any reason to not follow the same convention here?

src/query/api/v1/handler/prometheus/tags/complete_tags.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/m3db/m3/src/query/api/v1/handler"
+	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/query/util/logging"
+	"github.com/m3db/m3/src/x/net/http"
+
+	"go.uber.org/zap"
+)
+
+const (
+	// SearchTagsURL is the url for searching tags
```
andrewmains12: Repeating @nikunjgit 's question from chat, how does this endpoint compare to prom's `label` API? (https://prometheus.io/docs/prometheus/latest/querying/api/#querying-label-values)

arnikola: Added the extra endpoint for prom-style label-values

src/query/models/matcher.go
```diff
@@ -108,3 +110,47 @@ func (m Matchers) String() string {
 
 	return buffer.String()
 }
+
+// TODO: make this more robust, handle types other than MatchEqual
+func matcherFromString(s string) (Matcher, error) {
```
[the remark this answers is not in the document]
andrewmains12: Stopgap for how long? If this is serving prod at any point, it's worth writing some tests for it imo. Also, this should be a pretty trivial one to test, no?

src/query/api/v1/handler/prometheus/tags/common.go
```diff
@@ -0,0 +1,154 @@
...
+	for _, tag := range results {
+		jw.WriteString(string(tag.Name))
+	}
+
+	jw.EndArray()
+
+	return jw.Close()
+}
+
+func renderDefaultResultsJSON(
```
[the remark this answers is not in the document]
andrewmains12: Good to know! I can ask around as well.

src/query/api/v1/handler/prometheus/tags/common.go
```diff
@@ -0,0 +1,177 @@
...
+	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/query/util/json"
+	"github.com/m3db/m3/src/x/net/http"
+
+	"github.com/prometheus/common/route"
+)
+
+const (
+	nameReplace         = "name"
```
nikunjgit: maybe reuse these from native/common.go ?

src/query/api/v1/handler/prometheus/tags/common.go
```diff
@@ -0,0 +1,177 @@
...
+		case "default":
+			tagQuery.CompleteNameOnly = false
+		case "tagNamesOnly":
+			tagQuery.CompleteNameOnly = true
+		default:
+			return nil, xhttp.NewParseError(errors.ErrInvalidResultParamError, http.StatusBadRequest)
+		}
+	}
+
+	filterNameTags := r.Form[filterNameTagsParam]
```
nikunjgit: check for existence ?

src/query/api/v1/handler/prometheus/tags/common.go
```diff
@@ -0,0 +1,177 @@
...
+	}
+
+	return &tagQuery, nil
+}
+
+func parseTagValuesToQuery(
+	r *http.Request,
+) *storage.CompleteTagsQuery {
+	ctx := r.Context()
+	name := route.Param(ctx, nameReplace)
```
nikunjgit: vars := mux.Vars(r)
	name, ok := vars[nameReplace]

src/query/api/v1/handler/prometheus/tags/common.go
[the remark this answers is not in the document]
nikunjgit: So i copied over statsdex bc i didn't know about easyjson. We can port over if you have seen it work well in the past.

src/query/api/v1/handler/prometheus/tags/common.go
```diff
@@ -0,0 +1,177 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package tags
```
nikunjgit: i think we made native and remote packages to differentiate between native endpoints vs remote endpoint. Maybe this belongs to native ?

nikunjgit: you can put it directly in native or add tags package inside it.

src/query/api/v1/handler/prometheus/tags/common.go
```diff
@@ -0,0 +1,177 @@
...
+			models.Matcher{
+				Type:  models.MatchEqual,
+				Name:  []byte(name),
+				Value: []byte{},
+			},
+		},
+	}
+}
+
+func parseQuery(r *http.Request) (string, error) {
```
nikunjgit: reuse code instead of copy ?

src/query/api/v1/handler/prometheus/tags/tag_values.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/query/util/logging"
+	"github.com/m3db/m3/src/x/net/http"
+
+	"go.uber.org/zap"
+)
+
+const (
+	// TagValuesURL is the url for tag values.
+	TagValuesURL = handler.RoutePrefixV1 + "/:" + nameReplace + "/values"
```
nikunjgit: use fmt.Sprintf to make this let confusing ?

src/query/api/v1/handler/prometheus/tags/tag_values.go
```diff
@@ -0,0 +1,79 @@
...
+) http.Handler {
+	return &CompleteTagsHandler{
+		storage: storage,
+	}
+}
+
+func (h *TagValuesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
+	logger := logging.WithContext(ctx)
+	w.Header().Set("Content-Type", "application/json")
```
nikunjgit: move this into a common place ? All the handlers seem to be setting these headers

src/query/api/v1/handler/prometheus/tags/tag_values.go
nikunjgit: why nameReplace ? This shouldn't be configurable by user

src/query/storage/fanout/storage.go
```diff
@@ -128,6 +128,34 @@ func (s *fanoutStorage) FetchTags(ctx context.Context, query *storage.FetchQuery
...
+	ctx context.Context,
+	query *storage.CompleteTagsQuery,
+	options *storage.FetchOptions,
+) (*storage.CompleteTagsResult, error) {
+	var accumulatedTags storage.CompleteTagsResultBuilder
+
+	for _, store := range s.stores {
+		// TODO: once complete tags enabled locally,
+		// this should aggregate and merge results from all storages.
+		if store.Type() == storage.TypeRemoteDC {
```
nikunjgit: ugly :( Just use filters as done in read/write ?

src/query/api/v1/handler/prometheus/tags/common.go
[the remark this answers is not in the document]
andrewmains12: Yeah it definitely did what it needed to for flipr-daemon-go, and I don't recall encountering any weird behavior with it. I can try integrating it as a cleanup item at some point.

src/query/api/v1/handler/prometheus/tags/tag_values.go
andrewmains12: +1, we could make this middleware.

src/query/api/v1/handler/prometheus/tags/common.go
arnikola: These aren't in native/common.go, that's more timeout params etc

src/query/api/v1/handler/prometheus/tags/common.go
arnikola: It's okay to take the default here if it doesn't exist in the configs, then the filter name tags is empty anyway

src/query/api/v1/handler/prometheus/tags/common.go
arnikola: This is a very different parseQuery function to the others; not much to reuse here

src/query/api/v1/handler/prometheus/tags/tag_values.go
arnikola: To make it easier to pull out the variable here. It's fairly similar to the way Prom is doing the var extraction here

src/query/api/v1/handler/prometheus/tags/common.go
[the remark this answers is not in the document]
arnikola: Sounds like a plan, but out of scope for this PR, would rather keep it focussed on adding the endpoint and avoid touching a bunch of unrelated things. Could open an issue for it?

src/query/storage/fanout/storage.go
arnikola: It uses a different query type to read/write

src/query/api/v1/handler/prometheus/tags/complete_tags.go
```diff
@@ -0,0 +1,84 @@
...
+	opts := storage.NewFetchOptions()
+	result, err := h.storage.CompleteTags(ctx, &query, opts)
+	if err != nil {
+		logger.Error("unable to complete tags", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
+		return
+	}
+
+	// TODO: Support multiple result types
+	w.Header().Set("Content-Type", "application/json")
```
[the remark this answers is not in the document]
andrewmains12: Makes sense to me!

src/query/storage/fanout/storage.go
nikunjgit: I meant, maybe add a new filter for tags ?

src/query/api/v1/handler/prometheus/tags/common.go
nikunjgit: it wouldn't panic ?

src/query/api/v1/handler/prometheus/tags/tag_values.go
nikunjgit: I meant, shouldn't this always be   "/:name/values"  ?? Don't see how it can be dynamic

arnikola: Using fmt.Sprintf doesn't allow this to be a const unfortunately

arnikola: It's not dynamic; using a constant from the parser so that it can reuse it in lookups vs having two potential lookup values, which could lead to bugs if only one is changed.

In any case, since it will get extracted by the parser, it can be any value as long as it matches the URL param as well as the parser

arnikola:
Github is doing something weird, closing in favour of https://github.com/m3db/m3/pull/1175

PR #1141, opened 2018-10-29
Fix lookback delta check

nikunjgit:


src/query/ts/values.go
```diff
@@ -94,13 +94,13 @@ func (d Datapoints) AlignToBounds(bounds models.Bounds) []Datapoints {
 
 		for dpIdx < numDatapoints && !d[dpIdx].Timestamp.After(t) {
 			point := d[dpIdx]
+			dpIdx++
```
benraskin92: Add a test to catch this :)

PR #1146, opened 2018-10-31
Stop consuming query resources after timeout

nikunjgit:


src/query/api/v1/handler/close.go
```diff
@@ -28,33 +28,23 @@ import (
...
 		notify := notifier.CloseNotify()
 		go func() {
 			// Wait for either the request to finish
 			// or for the client to disconnect
 			select {
-			case <-done:
 			case <-notify:
 				logger.Warn("connection closed by client")
-				close(closing)
+				cancel()
```
andrewmains12: +1 on correct context usage :).

src/query/executor/engine.go
```diff
@@ -168,5 +152,5 @@ func (e *Engine) ExecuteExpr(ctx context.Context, parser parser.Parser, opts *En
 
 // Close kills all running queries and prevents new queries from being attached.
 func (e *Engine) Close() error {
-	return e.tracker.Close()
+	return nil
```
andrewmains12: If we're not actually closing anything at the moment, should we get rid of `Close`, or add a TODO?

src/query/storage/m3/storage.go
```diff
@@ -160,6 +160,15 @@ func (s *m3storage) FetchRaw(
 
 	wg.Wait()
 
+	// Check if the query was interrupted.
+	select {
+	case <-ctx.Done():
+		return nil, noop, ctx.Err()
+	case <-options.KillChan:
```
andrewmains12: Do we actually need a separate `killChan` in addition to the context? Could a cancellable context be used to send that signal all the way through?

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -128,8 +128,7 @@ func (h *PromReadHandler) read(
 
 	opts := &executor.EngineOptions{}
 	// Detect clients closing connections
-	abortCh, _ := handler.CloseWatcher(ctx, w)
-	opts.AbortCh = abortCh
```
andrewmains12: Is `abortCh` now unused? Could we eliminate it entirely (in favor of using contexts like this PR does)?

src/query/executor/engine.go
nikunjgit: might still need close to flush some metrics/traces in the future. Will leave it here for now. I'll update the comment

src/query/api/v1/handler/prometheus/native/read.go
nikunjgit: will remove

PR #1151, opened 2018-10-31
Add configuration for downsampler

cw9:
This pr adds a configuration struct for downsampler.

### DIFF
--- a/src/cmd/services/m3coordinator/downsample/options.go
+++ b/src/cmd/services/m3coordinator/downsample/options.go
@@ -159,7 +159,7 @@ type agg struct {
 // Configuration configurates a downsampler.
 type Configuration struct {
 	// AggregationTypes configs the aggregation types.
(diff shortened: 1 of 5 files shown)


src/cmd/services/m3coordinator/downsample/options.go
```diff
@@ -155,7 +156,37 @@ type agg struct {
 	pools                  aggPools
 }
 
-func (o DownsamplerOptions) newAggregator() (agg, error) {
+// Configuration configurates a downsampler.
+type Configuration struct {
+	// AggregationTypes configs the aggregation types.
+	AggregationTypes *aggregation.TypesConfiguration
```
prateek: wanna give this a `yaml:...` name too

cw9: Good catch, that one slipped

PR #1156, opened 2018-11-05
Set wasRetrievedFromDisk properly

justinjc:
Prior to being [removing the CacheAllMetadata policy](https://github.com/m3db/m3/pull/1110/files#diff-8605e9ad35bdfdb5c194ed171978939dL206), this used to be set in the `OnRetrieveBlock` function. Properly set it here.

### DIFF
--- a/src/dbnode/client/fetch_state_leakcheckpool_gen_test.go
+++ b/src/dbnode/client/fetch_state_leakcheckpool_gen_test.go
@@ -26,8 +26,11 @@ package client
 
 import (
 	"fmt"
+
 	"runtime/debug"
+
 	"sync"
+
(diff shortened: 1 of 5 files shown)


richardartoul:
@justinjc Ok before we fixed this I wanted to dig in and figure out why the integration tests didn't catch anything. It turns out this doesn't really break anything because the WiredList luckily has some protection against not storing things that were not retrieved from disk. As a result, these blocks were getting evicted constantly but there was no data loss or double closes which makes me feel better. Basically the practical implication of this bug would be a performance regression. That said, can you at leas add a unit test to the block package to make sure this value gets set properly

PR #1157, opened 2018-11-06
Implement the ability to write in batches in M3DB database

richardartoul:


src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -45,17 +45,25 @@ var (
 	timeZero = time.Time{}
 )
 
+// WritesBatch is a batch of Write.
+type WritesBatch []Write
```
robskillington: Can we wrap this up in a struct (which we could then pool)?

src/dbnode/persist/fs/commitlog/types.go
```diff
@@ -107,7 +113,9 @@ type Series struct {
 	ID ident.ID
 
 	// Tags are the series tags
-	Tags ident.Tags // FOLLOWUP(prateek): wire Tags to commit log writer
+	Tags ident.Tags
+
+	TagIter ident.TagIterator
```
robskillington: Just curious, why do we have both now?  Is it an either/or argument?

src/dbnode/persist/fs/commitlog/commit_log.go
robskillington: Can we also rename it to `WriteBatch` and drop the `s`?  Just since in other places where we have write batches that's what we call them.

src/dbnode/persist/fs/commitlog/types.go
robskillington: Can we create a wrapper type `TagsOrTagsIter` struct and perhaps a `type TagsType uint` enum like to specify which one to use?

src/cmd/services/m3dbnode/config/config_test.go
```diff
@@ -206,6 +206,13 @@ db:
             size: 8192
             lowWatermark: 0.01
             highWatermark: 0.02
+        writeBatchPool:
+            initialBatchSize: 1024
```
robskillington: Should we use 128 to match the default write ingestion batch size?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -824,50 +827,59 @@ func (s *service) WriteBatchRaw(tctx thrift.Context, req *rpc.WriteBatchRawReque
 	pooledReq.writeReq = req
 	ctx.RegisterFinalizer(pooledReq)
 
-	nsID := s.newPooledID(ctx, req.NameSpace, pooledReq)
-
 	var (
-		errs               []*rpc.WriteBatchRawError
-		success            int
+		nsID = s.newPooledID(ctx, req.NameSpace, pooledReq)
+
```
robskillington: nit: Remove empty line here?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -889,57 +901,66 @@ func (s *service) WriteTaggedBatchRaw(tctx thrift.Context, req *rpc.WriteTaggedB
 	pooledReq.writeTaggedReq = req
 	ctx.RegisterFinalizer(pooledReq)
 
-	nsID := s.newPooledID(ctx, req.NameSpace, pooledReq)
-
 	var (
-		errs               []*rpc.WriteBatchRawError
-		success            int
+		nsID = s.newPooledID(ctx, req.NameSpace, pooledReq)
+
```
robskillington: nit: Remove empty line here too?

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -443,20 +455,54 @@ func (l *commitLog) write() {
...
-			l.log.Errorf("failed to write to commit log: %v", err)
+		if write.write.writeBatch == nil {
+			// Handle individual write case
+			numDequeued = 1
+			write := write.write.write
+			err := l.writerState.writer.Write(write.Series,
+				write.Datapoint, write.Unit, write.Annotation)
+			if err != nil {
+				l.handleWriteErr(err)
+				continue
```
robskillington: Hm do we really want to continue here?  We won't dequeue the num writes in queue if so.

Should we add a test to catch this too perhaps?  Or find some way to have a single code path?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -1263,10 +1295,55 @@ func (r *writeBatchPooledReq) Finalize() {
 		r.writeTaggedReq = nil
 	}
 
+	r.nonRetryableErrors = 0
+	r.retryableErrors = 0
+	if cap(r.errs) <= writeBatchPooledReqPoolMaxErrorsSliceSize {
+		r.errs = r.errs[:0]
+	} else {
+		// Slice grew too large, throw it away and let a knew one be
```
robskillington: nit: Typo here.

src/dbnode/storage/database.go
```diff
@@ -156,6 +169,8 @@ func NewDatabase(
 		errors:       xcounter.NewFrequencyCounter(opts.ErrorCounterOptions()),
 		errWindow:    opts.ErrorWindowForLoad(),
 		errThreshold: opts.ErrorThresholdForLoad(),
+
```
robskillington: nit: Remove empty line.

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: Sync'd offline, found a single code path solution

src/cmd/services/m3dbnode/config/config_test.go
richardartoul: This is just a random config forwards/backwards test, but I can update it here

PR #1160, opened 2018-11-07
Make commitlog msgpack encoder really, really fast.

richardartoul:
The basic gist of this P.R is that the old code had too many function calls. Function call overhead in Golang is small, but when every function does nothing more than write a few bytes to an array in memory, the function call overhead begins to dominate. As a result, what we do in this P.R is create a fast path that manually inlines (read: copy-pastes) a lot of the helper functions. This makes the code more verbose and more difficult to maintain, but we get an over 10x speedup in exchange which is worth it as this is the hottest code path in M3DB and bottlenecks everything else.

Old path log entry
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/persist/fs/msgpack
BenchmarkLogEntryDecoder-8   	 3000000	       611 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/m3db/m3/src/dbnode/persist/fs/msgpack	2.423s
Success: Benchmarks passed.
```

New path log entry
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/persist/fs/msgpack
BenchmarkLogEntryEncoderFast-8   	30000000	        43.9 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/m3db/m3/src/dbnode/persist/fs/msgpack	1.381s
Success: Benchmarks passed.
```

Old path log metadata
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/persist/fs/msgpack
BenchmarkLogMetadataEncoder-8   	 5000000	       235 ns/op	     130 B/op	       0 allocs/op
PASS
ok  	github.com/m3db/m3/src/dbnode/persist/fs/msgpack	1.526s
Success: Benchmarks passed.
```

New path log metadata
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/persist/fs/msgpack
BenchmarkLogMetadataEncoderFast-8   	50000000	        31.9 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/m3db/m3/src/dbnode/persist/fs/msgpack	1.648s
Success: Benchmarks passed.
````

### CODE REVIEW

File: src/dbnode/persist/fs/msgpack/encoder_fast.go [hunk 1]
```diff
@@ -0,0 +1,348 @@
...
+// than write a few bytes to an in-memory array the function call overhead begins to dominate,
+// especially when each call to this function results in dozens of such helper function calls.
+//
+// Manually inlining the msgpack encoding results in a lot of code duplication for this one path, but
+// we pay the price because this is the most frequently called function in M3DB and it indirectly
+// applies back-pressure on every other part of the system via the commitlog queue. As a result, almost
+// any performance gains that can be had in this function are worth it.
+//
+// Before modifying this function, please run the BenchmarkLogEntryEncoderFast benchmark as a small
+// degration in this functions performance can have a substantial impact on M3DB.
```
isaachier: `degradation` typo

src/dbnode/persist/fs/commitlog/writer.go
```diff
@@ -129,6 +130,7 @@ func newCommitLogWriter(
 		sizeBuffer:         make([]byte, binary.MaxVarintLen64),
 		seen:               bitset.NewBitSet(defaultBitSetLength),
 		logEncoder:         msgpack.NewEncoder(),
+		logEncoderBuff:     make([]byte, 0, 16384),
```
robskillington: Nit: make size a const (I know this was me hah).

src/dbnode/persist/fs/msgpack/encoder_decoder_bench_test.go
```diff
@@ -0,0 +1,81 @@
...
+		if err != nil {
+			panic(err)
+		}
+	}
+}
+
+var benchmarkBuf []byte
+
+func BenchmarkLogEntryEncoderFast(b *testing.B) {
+	var (
```
robskillington: N1

src/dbnode/persist/fs/msgpack/encoder_fast.go
```diff
@@ -0,0 +1,348 @@
...
+	b = encodeFloat64(b, entry.Value)
+	b = encodeVarUint64(b, uint64(entry.Unit))
+	b = encodeBytes(b, entry.Annotation)
+
+	return b, nil
+}
+
+// encodeVarUint64 is used in EncodeLogEntryFast which is a very hot path.
+// As a result, many of the function calls in this function have been
+// manually inlined to reduce function call overhead.
```
robskillington: So I’ve read this code a few times, the manual inlining makes a lot of sense.  Having said that it might be worth using Genny to code gen the code so that if we really need to change parts of the code it affects every party (DRY it up).  

Genny should be able to code gen all the inlined parts, it’s more work for sure, so I’m fine if with TODO it but we should definitely do this at some point (otherwise the copy paste manual inlining thing may become a pattern and code gen is definitely better than what manual copy paste to inline things).

Thoughts?

src/dbnode/persist/fs/msgpack/encoder_fast.go
```diff
@@ -0,0 +1,348 @@
...
+		buf[4] = byte(v)
+	}
+
+	b = append(b, data...)
+	return b
+}
+
+// encodeBytesLen is not used, it is left here to demonstrate what a manually inlined
+// version of this function should look like.
+func encodeBytesLen(b []byte, l int) []byte {
```
robskillington: Is this called? Looks Inlined elsewhere.  This is where codgen would help (some being reused while others being used as real calls, everything can just always be inlined potentially).

src/dbnode/persist/fs/msgpack/encoder_fast.go
richardartoul: Yeah I thought about doing that too, but I’d prefer to leave as a TODO for now. Mainly because:

1. Hopefully this doesn’t become a pattern for us
2. This is really unplanned work and there’s like 5 other things I’m supposed to be doing that are not performance improvements and I have a feeling getting the genny thing right will take a little bit of back and forth

src/dbnode/persist/fs/msgpack/encoder_fast.go
richardartoul: Nope not called (if it is, it’s because I forgot to in-line it). I just left these in as “documention” per the comment

src/dbnode/persist/fs/commitlog/writer.go
richardartoul: Haha will fix

src/dbnode/persist/fs/commitlog/writer.go
```diff
@@ -224,22 +237,30 @@ func (w *writer) Write(
 		metadata.Namespace = series.Namespace.Bytes()
 		metadata.Shard = series.Shard
 		metadata.EncodedTags = encodedTags
-		w.metadataEncoder.Reset()
-		if err := w.metadataEncoder.EncodeLogMetadata(metadata); err != nil {
+
+		w.metadataEncoderBuff = w.metadataEncoderBuff[:0]
+		var err error
+
+		w.metadataEncoderBuff, err = msgpack.EncodeLogMetadataFast(w.metadataEncoderBuff, metadata)
```
robskillington: p.s. I realized this can be collapsed somewhat to do the `[:0]` just as you pass it to the EncodeLogMetadataFast method:
```go
var err error
w.metadataEncoderBuff, err = msgpack.EncodeLogMetadataFast(w.metadataEncoderBuff[:0], metadata)
```

src/dbnode/persist/fs/commitlog/writer.go
```diff
@@ -224,22 +237,30 @@ func (w *writer) Write(
...
 	logEntry.Value = datapoint.Value
 	logEntry.Unit = uint32(unit)
 	logEntry.Annotation = annotation
-	w.logEncoder.Reset()
-	if err := w.logEncoder.EncodeLogEntry(logEntry); err != nil {
+
+	w.logEncoderBuff = w.logEncoderBuff[:0]
+
+	var err error
+	w.logEncoderBuff, err = msgpack.EncodeLogEntryFast(w.logEncoderBuff, logEntry)
```
robskillington: Same here, can collapse to `w.logEncoderBuff, err = msgpack.EncodeLogEntryFast(w.logEncoderBuff[:0], logEntry)`

src/dbnode/persist/fs/msgpack/encoder_fast.go
isaachier: This PR makes sense only if you believe Go won't inline your function in the near future. Try debugging this the explanation here in the Go compiler: https://github.com/golang/go/blob/master/src/cmd/compile/internal/gc/inl.go#L5-L25.

richardartoul: I think the benchmarks demonstrate that this P.R is worth it. Even if Golang inlining gets better in the near-future, this function is so hot and the number of function calls we removed so high (we're talking about literally dozens of function calls per iteration that have been removed) I highly doubt that the Go compiler will be able to achieve a similar effect anytime soon

PR #1162, opened 2018-11-12
[api] Implement replace endpoint

### LINKED ISSUE #1095: api: Implement replace endpoint
We currently have Add and Remove endpoints, but not Replace which is a common operation for us.

schallert:
Adds an endpoint for replacing an instance across all services.

Closes #1095


docs/operational_guide/placement_configuration.md
```diff
@@ -126,16 +126,31 @@ After sending the add command you will need to wait for the M3DB cluster to reac
 
 #### Removing a Node
 
-Send a DELETE request to the `/api/v1/placement/<NODE_ID>` endpoint.
+Send a DELETE request to the `/api/v1/services/m3db/placement/<NODE_ID>` endpoint.
```
richardartoul: nice thanks, I think the legacy stuff still works but good to update the docs to the new stuff

docs/operational_guide/placement_configuration.md
```diff
@@ -126,16 +126,31 @@ After sending the add command you will need to wait for the M3DB cluster to reac
...
 After sending the delete command you will need to wait for the M3DB cluster to reach the new desired state. You'll know that this has been achieved when the placement shows that all shards for all hosts are in the `Available` state.
 
 #### Replacing a Node
 
-Currently, the best way to replace a node (due to hardware failure or any other reason) is to perform a node remove followed by a node add. We will eventually support node replacement as a single operation, but that is currently not implemented.
-
+Send a POST request to the `/api/v1/services/m3db/placement/replace` endpoint containing hosts to replace and candidates to replace it with.
 
+```bash
+curl -X POST <M3_COORDINATOR_HOST_NAME>:<M3_COORDINATOR_PORT(default 7201)>/api/v1/services/m3db/placement/replace -d '{
```
richardartoul: Does this all work out correctly with various numbers of inputs and outputs? I.E if I want to replace one node with two, or remove two and add one etc

src/query/api/v1/handler/placement/add_test.go
```diff
@@ -154,7 +154,7 @@ func TestPlacementAddHandler_SafeOK(t *testing.T) {
 		)
 		switch serviceName {
 		case M3AggregatorServiceName:
-			req = httptest.NewRequest(AddHTTPMethod, M3DBAddURL, strings.NewReader(`{"instances":[{"id": "host1","isolation_group": "rack1","zone": "test","weight": 1,"endpoint": "http://host1:1234","hostname": "host1","port": 1234}]}`))
+			req = httptest.NewRequest(AddHTTPMethod, M3AggAddURL, strings.NewReader(`{"instances":[{"id": "host1","isolation_group": "rack1","zone": "test","weight": 1,"endpoint": "http://host1:1234","hostname": "host1","port": 1234}]}`))
```
richardartoul: +1 thanks

src/query/api/v1/handler/placement/replace.go
```diff
@@ -0,0 +1,154 @@
...
+		return
+	}
+
+	placement, err := h.Replace(serviceName, r, req)
+	if err != nil {
+		status := http.StatusInternalServerError
+		if _, ok := err.(unsafeAddError); ok {
+			status = http.StatusBadRequest
+		}
+		logger.Error("unable to replace instance", zap.Any("error", err))
```
arnikola: nit: `zap.Error(err)`

src/query/api/v1/handler/placement/replace.go
```diff
@@ -0,0 +1,154 @@
...
+		xhttp.Error(w, err, status)
+		return
+	}
+
+	placementProto, err := placement.Proto()
+	if err != nil {
+		logger.Error("unable to get placement protobuf", zap.Any("error", err))
```
arnikola: nit: `zap.Error(err)`

src/query/api/v1/handler/placement/replace.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	// M3Coordinator isn't sharded, can't check if its shards are available.
+	if serviceName != M3CoordinatorServiceName {
+		if err := validateAllAvailable(curPlacement); err != nil {
+			return nil, err
+		}
+	}
+
+	newPlacement, err := algo.ReplaceInstances(curPlacement, req.LeavingInstanceIDs, candidates)
```
richardartoul: can you add a comment about why you use the algo / check and set stuff

src/query/api/v1/handler/placement/replace.go
```diff
@@ -0,0 +1,154 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	if err := service.CheckAndSet(newPlacement, version); err != nil {
+		return nil, err
+	}
+
+	return newPlacement.SetVersion(version + 1), nil
```
richardartoul: A little annoying that CheckAndSet doesn't set the version, but this is guaranteed to be correct right (version +1)?

src/query/generated/assets/openapi/spec.yml
```diff
@@ -339,6 +339,35 @@ paths:
           description: ""
           schema:
             $ref: "#/definitions/GenericError"
+  /services/m3db/placement/replace:
+    post:
+      tags:
+      - "M3DB Placement"
+      - "M3DB"
+      summary: "Replace an M3DBNode"
```
richardartoul: do we concatenate M3DB and Node together in the other summaries?

src/query/api/v1/handler/placement/replace.go
```diff
@@ -0,0 +1,154 @@
...
+
+	resp := &admin.PlacementGetResponse{
+		Placement: placementProto,
+		Version:   int32(placement.GetVersion()),
+	}
+
+	xhttp.WriteProtoMsgJSONResponse(w, resp, logger)
+}
+
+func (h *ReplaceHandler) parseRequest(r *http.Request) (*admin.PlacementReplaceRequest, *xhttp.ParseError) {
```
arnikola: nit: maybe have this return a generic `error` since you're unwrapping it in the caller anyway? Fine either way

src/query/api/v1/handler/placement/replace.go
```diff
@@ -0,0 +1,154 @@
...
+	return req, nil
+}
+
+// Replace replaces instances.
+func (h *ReplaceHandler) Replace(
+	serviceName string,
+	httpReq *http.Request,
+	req *admin.PlacementReplaceRequest,
+) (placement.Placement, error) {
+
```
arnikola: nit: extra newline

src/query/api/v1/handler/placement/replace.go
```diff
@@ -0,0 +1,154 @@
...
+	serviceOpts := NewServiceOptions(serviceName, httpReq.Header, h.M3AggServiceOptions)
+	service, algo, err := ServiceWithAlgo(h.ClusterClient, serviceOpts, h.nowFn())
+	if err != nil {
+		return nil, err
+	}
+
+	if req.Force {
+		newPlacement, _, err := service.ReplaceInstances(req.LeavingInstanceIDs, candidates)
+		if err != nil {
```
arnikola: nit: can probably just `return newPlacement, err` here

src/query/api/v1/handler/placement/replace.go
```diff
@@ -0,0 +1,154 @@
...
+const (
+	// ReplaceHTTPMethod is the HTTP method for the the replace endpoint.
+	ReplaceHTTPMethod = http.MethodPost
+
+	replacePathName = "replace"
+)
+
+var (
+	// M3DBReplaceURL is the url for the m3db replace handler (method POST).
+	M3DBReplaceURL = path.Join(handler.RoutePrefixV1, M3DBServicePlacementPathName, replacePathName)
```
richardartoul: You wrote documentation for m3db/m3coordiantor/m3agg but only created a route for M3DB. Wanna create the other routes? Also we should get @cw9 to confirm there are no other special changes we need to make to support M3Agg replaces (when I did the m3agg stuff I looked at the r2admin code to see if there was anything special but Chao can probably just answer this question quickly)

docs/operational_guide/placement_configuration.md
schallert: It should so long as that algorithm considers that change valid, since we don't do much rather than be a glorified wrapper around that. @cw9 can chime in on the algo stuff

src/query/api/v1/handler/placement/replace.go
schallert: Will do

src/query/api/v1/handler/placement/replace.go
schallert: It currently is but admittedly that's tied to the etcd implementation I think. The API CheckAndSet uses [under the hood](https://github.com/m3db/m3/blob/a2fab37bae1b9144354eba854bcb050a7d6398b3/src/cluster/kv/types.go#L141) propagates the version but we [drop that](https://github.com/m3db/m3/blob/a2fab37bae1b9144354eba854bcb050a7d6398b3/src/cluster/placement/storage/storage.go#L102) when the placement service does CheckAndSet.

We could (and probably should) refactor that but it would likely require updating a bunch of interface implementations. I can open an issue.

src/query/api/v1/handler/placement/replace.go
schallert: Will do

src/query/generated/assets/openapi/spec.yml
schallert: Nope, will fix

docs/operational_guide/placement_configuration.md
```diff
@@ -64,10 +64,10 @@ The instructions below all contain sample curl commands, but you can always revi
 
 #### Placement Initialization
 
-Send a POST request to the `/api/v1/placement/init` endpoint
+Send a POST request to the `/api/v1/services/m3db/placement/init` endpoint
```
justinjc: I think right now it's a little ambiguous as to when to use `/api/v1/...` and when to use others like `/api/v1/services/m3db/...`. There are other mentions of `/api/v1/placement` like in `docs/how_to` (Github docs) and in our docker integration tests. It would be great to get some clarification on when to use which, but perhaps that's for a larger discussion in another PR.

README.md
```diff
@@ -22,7 +22,7 @@ More information:
 make m3dbnode (note that we currently require at least Go 1.10 or higher)
 
 # run it with the sample configuration
-./bin/m3dbnode -f ./src/dbnode/config/m3dbnode-local-etcd.yml
+./bin/m3dbnode -f ./src/dbnode/example/m3db-node-config.yaml
```
robskillington: Looks like you deleted this file?  Did you want to keep the `./src/dbnode/config/m3dbnode-local-etcd`.yaml` path here?

docs/operational_guide/placement_configuration.md
richardartoul: @justinjc It should always be the /services/ ones, we just kept the old names in to not make a breaking change. Anywhere we have a URL without the service name should just be updated to include it

src/query/api/v1/handler/placement/replace.go
```diff
@@ -0,0 +1,154 @@
...
+		return newPlacement, nil
+	}
+
+	curPlacement, version, err := service.Placement()
+	if err != nil {
+		return nil, err
+	}
+
+	// M3Coordinator isn't sharded, can't check if its shards are available.
+	if serviceName != M3CoordinatorServiceName {
```
robskillington: Hm, can we refactor this into a `IsStateless(serviceName string) bool` method?  I see that the placement delete handler also does something like this, which could change to just:

Before
```go
	switch serviceName {
	case M3CoordinatorServiceName:
		// There are no unsafe placement changes because M3Coordinator is stateless
		force = true
	}
```

After:
```go
force := IsStateless(serviceName)
```

src/query/api/v1/handler/placement/replace.go
robskillington: Hm, you could just do a get placement operation to avoid this possibly?

src/query/api/v1/handler/placement/replace.go
schallert: Hmm, I see what you mean but since we use ParseError in all the other handler handlers I'm going to stick to local convention

src/query/generated/assets/openapi/spec.yml
robskillington: Other M3DB ones:
```
 summary: "Add an instance to the placement"
```

Maybe "Replace an instance in the placement"? *shrug*

src/query/api/v1/handler/placement/replace.go
schallert: Updating. Re: the delete handler we can't just set force outright because we may end up setting it to false when it was set to true for a stateful service, but added a check to simplify

src/query/api/v1/handler/placement/replace.go
robskillington: Hey can you return perhaps just:
```go
p, _, err := service.Placement()
return p, err
```

Just so before we have some wrapper that does this for you in a way that can be reused, you at least just read back what you wrote?

cw9: That read after CAS breaks the atomicity, the current implementation is the only way to get atomicity in placement change with validation, a more proper solution is [this](https://github.com/m3db/m3cluster/issues/243)

docs/operational_guide/placement_configuration.md
cw9: this is fine, the algo basically calls the `add()` and then call the `remove()` on non sharded placements

src/query/api/v1/handler/placement/replace.go
cw9: You might need to change `ServiceWithAlgo` function to make sure it's using the same logic with r2admin, in there we had functions like `GetOptionsForReplace`, Richies work probably only moved  `GetOptions` over.

schallert: @cw9 do we need more logic past what's here? https://github.com/m3db/m3/blob/a2fab37bae1b9144354eba854bcb050a7d6398b3/src/query/api/v1/handler/placement/common.go#L262

PR #1163, opened 2018-11-13
Reduce noise from multi errors by returning last

arnikola:
Previously, was returning the concatenated error messages for all
errors seen, which got very noisy if a single failure was causing each
request to error. Now returns only last error if any seen.


glide.lock
```diff
@@ -1,5 +1,5 @@
-hash: e18d8ec10bc3e1cefa5c87365838c7c97ba003d552733931b5e4b9a2515d4235
-updated: 2018-10-23T16:48:16.261449-04:00
+hash: 760468137d3b67dc5978441265e564f5afe9f526fb4c556464eb3a4505df8dbc
```
nikunjgit: remove this file from the diff

PR #1166, opened 2018-11-13
[cluster] placement: propagate version in storage

### LINKED ISSUE #1165: Placement storage CheckAndSet should expose version
The KV [CheckAndSet](https://github.com/m3db/m3/blob/a2fab37bae1b9144354eba854bcb050a7d6398b3/src/cluster/kv/types.go#L141) API returns a version number of the updated K/V pair. However when the Placement CheckAndSet [calls](https://github.com/m3db/m3/blob/a2fab37bae1b9144354eba854bcb050a7d6398b3/src/cluster/placement/storage/storage.go#L102-L121) KV CheckAndSet it drops the version it receives. We should propagate this version so users can depend on the updated version of a placement.

schallert:
Closes #1165

### DIFF
--- a/src/cluster/placement/algo/sharded_helper.go
+++ b/src/cluster/placement/algo/sharded_helper.go
@@ -99,6 +99,7 @@ type helper struct {
 	groupToWeightMap    map[string]uint32
 	totalWeight         uint32
 	rf                  int
+	version             int
 	uniqueShards        []uint32
 	instances           map[string]placement.Instance
 	maxShardSetID       uint32
@@ -198,6 +199,7 @@ func newHelper(p placement.Placement, targetRF int, opts placement.Options) plac
 		instances:     make(map[string]placement.Instance, p.NumInstances()),
 		uniqueShards:  p.Shards(),
 		maxShardSetID: p.MaxShardSetID(),
+		version:       p.GetVersion(),
 		log:           opts.InstrumentOptions().Logger(),
 		opts:          opts,
 	}
@@ -404,7 +406,8 @@ func (ph *helper) generatePlacement() placement.Placement {
 		SetIsSharded(true).
 		SetIsMirrored(ph.opts.IsMirrored()).
 		SetCutoverNanos(ph.opts.PlacementCutoverNanosFn()()).
-		SetMaxShardSetID(maxShardSetID)
+		SetMaxShardSetID(maxShardSetID).
+		SetVersion(ph.version)
 }
 
 func (ph *helper) placeShards(
--- a/src/cluster/placement/service/service.go
+++ b/src/cluster/placement/service/service.go
@@ -163,11 +163,7 @@ func (ps *placementService) RemoveInstances(instanceIDs []string) (placement.Pla
 		return nil, err
 	}
 
-	_, err = ps.CheckAndSet(p, p.GetVersion())
-	if err != nil {
-		return nil, err
-	}
-	return p, nil
+	return ps.CheckAndSet(p, p.GetVersion())
 }
 
 func (ps *placementService) ReplaceInstances(
@@ -201,7 +197,7 @@ func (ps *placementService) ReplaceInstances(
 		addedInstances = append(addedInstances, addedInstance)
 	}
 
-	_, err = ps.CheckAndSet(p, p.GetVersion())
+	p, err = ps.CheckAndSet(p, p.GetVersion())
 	if err != nil {
 		return nil, nil, err
 	}
--- a/src/cluster/placement/service/service.go
+++ b/src/cluster/placement/service/service.go
@@ -86,7 +86,10 @@ func (ps *placementService) BuildInitialPlacement(
 	}
 
 	_, err = ps.SetIfNotExist(p)
-	return p, err
+	if err != nil {
+		return nil, err
+	}
+	return p, nil
 }
 
 func (ps *placementService) AddReplica() (placement.Placement, error) {
@@ -104,7 +107,10 @@ func (ps *placementService) AddReplica() (placement.Placement, error) {
 	}
 
 	_, err = ps.CheckAndSet(p, v)
-	return p, err
+	if err != nil {
+		return nil, err
+	}
+	return p, nil
 }
 
 func (ps *placementService) AddInstances(
@@ -137,7 +143,10 @@ func (ps *placementService) AddInstances(
 	}
 
 	_, err = ps.CheckAndSet(p, v)
-	return p, addingInstances, err
+	if err != nil {
+		return nil, nil, err
+	}
+	return p, addingInstances, nil
 }
 
 func (ps *placementService) RemoveInstances(instanceIDs []string) (placement.Placement, error) {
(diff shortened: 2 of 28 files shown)


src/cluster/placement/service/service.go
```diff
@@ -85,7 +85,8 @@ func (ps *placementService) BuildInitialPlacement(
 		return nil, err
 	}
 
-	return p, ps.SetIfNotExist(p)
+	_, err = ps.SetIfNotExist(p)
```
cw9: change to the style of
```
if _, err = ps.SetIfNotExist(p); err != nil {
  return nil, err
}
return p, nil
```
in all similar places in this file?

src/cluster/placement/storage/storage.go
```diff
@@ -50,98 +52,93 @@ func NewPlacementStorage(store kv.Store, key string, opts placement.Options) pla
...
+func (s *storage) CheckAndSetProto(p proto.Message, version int) (int, error) {
 	if err := s.helper.ValidateProto(p); err != nil {
-		return err
+		return errorVersionValue, err
 	}
 
 	if s.opts.Dryrun() {
 		s.logger.Info("this is a dryrun, the operation is not persisted")
-		return nil
+		return 0, nil
```
cw9: errorVersionValue instead of 0?

src/cluster/placement/storage/storage_test.go
```diff
@@ -28,6 +28,7 @@ import (
 	"github.com/m3db/m3/src/cluster/kv/mem"
 	"github.com/m3db/m3/src/cluster/placement"
 
+	"github.com/stretchr/testify/assert"
```
cw9: nit: require instead of assert?

schallert: What's the value of using require here? Some of these seem like tests where even if the version isn't correct we'd want to debug other failures in the test rather than have to fix a version issue and push a new commit to make further test progress

src/cluster/placement/types.go
```diff
@@ -440,14 +440,14 @@ const (
 // Storage provides read and write access to placement.
 type Storage interface {
 	// Set writes a placement.
-	Set(p Placement) error
+	Set(p Placement) (int, error)
```
robskillington: Hey, see how hard this is but I think either of these two approaches is a good idea:

1)
So, we can either update the description for these methods to be:
```go
// Set writes a placement and returns the version or an error.
Set(p Placement) (int, error)
```

2)
Or we can:
```go
package kv

// Version is the version of a key value pair.
type Version int
```

```go
package placement

type Storage interface {
  // Set writes a placement.
  Set(p Placement) (kv.Version, error)

  // ...
}
```

src/cluster/placement/storage/storage_test.go
cw9: Usually I prefer to use only one of `assert` or `require` in a test unless both are really needed

src/cluster/placement/types.go
robskillington: (If it's not clear, my preference is number two, but yeah I could see how it might be intimidating to add this change - having said that if you scope the change super tight then it shouldn't be too full on to make the change)

src/cluster/placement/storage/storage.go
```diff
@@ -50,98 +52,93 @@ func NewPlacementStorage(store kv.Store, key string, opts placement.Options) pla
...
+func (s *storage) CheckAndSetProto(p proto.Message, version int) (int, error) {
 	if err := s.helper.ValidateProto(p); err != nil {
-		return err
+		return errorVersionValue, err
 	}
 
 	if s.opts.Dryrun() {
 		s.logger.Info("this is a dryrun, the operation is not persisted")
-		return nil
+		return errorVersionValue, nil
```
cw9: nit: probably not ideal to apply the `+1` logic here but this should be `version+1` for dryrun?

schallert: Ehh, I'd rather it be a constant for dry runs. `version+1` is only the case because that's an implementation detail of etcd. Plus between when you do a dry run and the actual command someone else could've changed the placement.

src/cluster/placement/types.go
cw9: Or you could do
```
Set(p Placement) (Placement, error)
```
so the users can do 
```
p, err := storage.Set(p)
if err != nil {...}
v := p.Version()
```

src/cluster/placement/storage/storage.go
cw9: It just feels weird to return an `errVersionValue` for this case when it's not an error, but I guess that's probably fine since this is just dryrun, the version would show up in logs after you run m3admin though, it does reduce readability of the log somewhat.

src/cluster/placement/types.go
```diff
@@ -440,14 +440,14 @@ const (
...
 
 	// CheckAndSet writes a placement if the current version
 	// matches the expected version.
-	CheckAndSet(p Placement, version int) error
+	CheckAndSet(p Placement, version int) (int, error)
 
 	// SetIfNotExist writes a placement.
-	SetIfNotExist(p Placement) error
+	SetIfNotExist(p Placement) (int, error)
```
cw9: Since you are here, could you change
`Placement() (Placement, int, error)` to `Placement() (Placement,error)`?

src/cluster/placement/service/service.go
```diff
@@ -151,14 +163,18 @@ func (ps *placementService) RemoveInstances(instanceIDs []string) (placement.Pla
 		return nil, err
 	}
 
-	return p, ps.CheckAndSet(p, v)
+	_, err = ps.CheckAndSet(p, p.GetVersion())
```
cw9: Can just return `ps.CheckAndSet` now?

src/cluster/placement/service/service.go
```diff
@@ -185,11 +201,15 @@ func (ps *placementService) ReplaceInstances(
 		addedInstances = append(addedInstances, addedInstance)
 	}
 
-	return p, addedInstances, ps.CheckAndSet(p, v)
+	_, err = ps.CheckAndSet(p, p.GetVersion())
```
cw9: `p, err = ps.CheckAndSet(...)`

src/cluster/placement/algo/sharded_helper.go
```diff
@@ -404,7 +406,8 @@ func (ph *helper) generatePlacement() placement.Placement {
 		SetIsSharded(true).
 		SetIsMirrored(ph.opts.IsMirrored()).
 		SetCutoverNanos(ph.opts.PlacementCutoverNanosFn()()).
-		SetMaxShardSetID(maxShardSetID)
+		SetMaxShardSetID(maxShardSetID).
+		SetVersion(ph.version)
```
cw9: Is this needed? I actually prefer to not have version in the placement algorithm, because it's generating a temporary placement between last version and next version, I'd prefer this to be set to a default/meaningless value like `0` to avoid user depend on the version on this temporary placement, I see you are not using this value in the apis as well? If that's the case, I'd prefer to not set this value here.

schallert: Discussed offline, we'll clean this up separately

PR #1170, opened 2018-11-14
Implement commitlog rotation as well as writing out, cleaning up, and bootstrapping using snapshot metadata files

richardartoul:
# Summary

This P.R implements the following functionality:

- [X] Rotate commitlogs, generate snapshot ID, and write out snapshot metadata/checkpoint files for each snapshot.
- [X] Rewrite snapshot/commitlog cleanup logic to take new snapshotting process into account.
- [X] Most minimal possible changes to commitlog bootstrapping logic to make it work with the new snapshot metadata files.
- [X] Support a fully transparent upgrade for users (even those using the commitlog bootstrapper) by keeping the bootstrapping logic backwards-compatible with previous version of M3DB.
- [X] Removal of the concept of commitlog blocksize.

# Review guide 

Unfortunately its very difficult to make this P.R any smaller because as soon as writing out the snapshot metadata files is implemented, the cleanup logic needs to be implemented, and if the cleanup logic is implemented then the bootstrapping logic needs to be implemented as well.

Luckily, while this is a large P.R, there are much more deletions than additions, and the vast majority of the new code is a simplification of previously much more complicated code. I,E a lot of the code that was rewritten went from very complicated logic to very simple logic.

Places to pay particular attention to in the review:

1. Cleanup logic in `cleanup.go`
2. Commitlog rotation logic in `commitlog.go` (removed the concept of rotating files by time entirely, and commitlog files will only be rotated when an explicit call to `RotateLogs()` is made.
3. Removal of the concept of a commitlog block size, and also while we retained the existing commitlog filename structure, all commitlog files will have a name that defaults to a blockStart of time.Unix(0, 0). Because the commitlog files have info headers and we don't need to rely on their names at all, I believe we can remove the blockStart from their name entirely, but I'd like to leave it out of this P.R as its already too big.
4. Changes to the PersistManager to support writing out the snapshot ID and commitlog ID when the entire snapshot process completes in `persist_manager.go`
5. Changes to the commitlog bootstrapping logic (mostly code deletion) that make it work for both the previous implementation and the current implementation. This is accomplished by simplifying the logic to: Read the most recent snapshot file for every namespace/shard/block combination and ALL commitlog files that exist on disk. This will be slow, but correct, and a subsequent P.R will make this significantly more efficient will still remaining backwards compatible with files generated from previous versions of M3DB (trying to make minimal possible changes to bootstrapping in this P.R to keep it from growing obscenely large). Also confidence that this is backwards compatible is maintained because I left the existing prop test unmodified which generates combinations of snapshots/commitlogs that look like they came from an older version of M3DB.
6. The logic for rotating the commitlog, generating the snapshot ID, and performing the snapshot in `flush.go`


src/dbnode/storage/cleanup.go
```diff
@@ -283,195 +274,181 @@ func (m *cleanupManager) cleanupExpiredNamespaceDataFiles(earliestToRetain time.
...
-		return nil, err
+		return err
 	}
 
-	activeCommitlogs, err := m.activeCommitlogs.ActiveLogs()
-	if err != nil {
-		return nil, err
+	// Assert that the snapshot metadata files are indeed sorted.
+	lastMetadataIndex := int64(-1)
+	for _, snapshotMetadata := range sortedSnapshotMetadatas {
```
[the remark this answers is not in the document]
richardartoul: not necessary, but I've been trying to add more assertions like this throughout the codebase in situations where they are cheap to make our lives easier

src/dbnode/storage/cleanup.go
```diff
@@ -26,16 +26,22 @@ import (
...
 	"github.com/m3db/m3x/ident"
+	xlog "github.com/m3db/m3x/log"
 
+	"github.com/pborman/uuid"
 	"github.com/uber-go/tally"
 )
 
-type commitLogFilesFn func(commitlog.Options) ([]commitlog.File, []commitlog.ErrorWithPath, error)
+type commitLogFilesFn func(commitlog.Options) (persist.CommitlogFiles, []commitlog.ErrorWithPath, error)
+type sortedSnapshotMetadataFilesFn func(fs.Options) ([]fs.SnapshotMetadata, []fs.SnapshotMetadataErrorWithPaths, error)
```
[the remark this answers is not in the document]
richardartoul: hmm....thats fair, let me see if I can rework it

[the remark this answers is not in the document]
richardartoul: actually I like the logic where its sorted, but maybe instead of checking if its sorted and assuming that in the interface I can just re-sort cause it should be cheap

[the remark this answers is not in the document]
richardartoul: ok updated, let me know what you think

src/dbnode/storage/flush.go
```diff
@@ -130,56 +134,35 @@ func (m *flushManager) Flush(
...
+	err = flushPersist.DoneFlush()
+	if err != nil {
+		multiErr = multiErr.Add(err)
+	}
 
-	if shouldSnapshot {
-		if multiErr.NumErrors() == 0 {
-			m.lastSuccessfulSnapshotStartTime = tickStart
-		}
+	// TODO: is this missing minimum time between snapshots?
```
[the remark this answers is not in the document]
richardartoul: this was actually more about preventing nodes with small amounts of data from snapshotting too much

src/dbnode/storage/flush.go
```diff
@@ -206,6 +189,50 @@ func (m *flushManager) Flush(
...
+	rotatedCommitlogID persist.CommitlogFile,
+) error {
+	snapshotID := uuid.NewUUID()
+
+	snapshotPersist, err := m.pm.StartSnapshotPersist(snapshotID)
+	if err != nil {
+		return err
+	}
+
+	m.setState(flushManagerSnapshotInProgress)
```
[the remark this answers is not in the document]
richardartoul: no, it will just remain in that state until the index flush begins which will cause a state transition

src/dbnode/storage/flush.go
```diff
@@ -130,56 +134,35 @@ func (m *flushManager) Flush(
...
+		err = m.flushNamespaceWithTimes(
+			ns, shardBootstrapTimes, flushTimes, flushPersist)
+		if err != nil {
+			multiErr = multiErr.Add(err)
 		}
-		m.maxBlocksSnapshottedByNamespace.Update(float64(maxBlocksSnapshottedByNamespace))
 	}
 
-	// mark data flush finished
-	multiErr = multiErr.Add(flush.DoneData())
```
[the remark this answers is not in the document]
richardartoul: it just felt easier to follow to me if the error path is clearly delimited

src/dbnode/generated/proto/snapshot/snapshot_metadata.proto
```diff
@@ -4,5 +4,10 @@ package snapshot;
 message Metadata {
   int64 snapshotIndex = 1;
   bytes snapshotUUID = 2;
-  bytes commitlogID = 3;
+  CommitlogID commitlogID = 3;
```
robskillington: nit: `commitLogID`?

src/dbnode/generated/proto/snapshot/snapshot_metadata.proto
```diff
@@ -4,5 +4,10 @@ package snapshot;
...
+}
+
+message CommitlogID {
```
robskillington: nit: `CommitLogID`?

src/cmd/services/m3dbnode/config/config.go
```diff
@@ -193,8 +193,8 @@ type CommitLogPolicy struct {
 	// enough for almost all workloads assuming a reasonable batch size is used.
 	QueueChannel *CommitLogQueuePolicy `yaml:"queueChannel"`
 
-	// The commit log block size.
-	BlockSize time.Duration `yaml:"blockSize" validate:"nonzero"`
+	// Deprecated. Left in struct to keep old YAMLs parseable.
+	DeprecatedBlockSize *time.Duration `yaml:"blockSize"`
```
robskillington: Interesting, good call - just wondering how we should track some of this stuff so we remove at v1.

Should we get some labels and/or test cases that can be discovered by scanning the repo easily?

src/dbnode/integration/commitlog_bootstrap_only_reads_required_files_test.go
```diff
@@ -33,9 +33,11 @@ import (
...
 
 func TestCommitLogBootstrapOnlyReadsRequiredFiles(t *testing.T) {
-	if testing.Short() {
-		t.SkipNow() // Just skip if we're doing a short run
-	}
+	// TODO(rartoul): Temporarily disabled until a subsequent P.R that will
+	// improve and simplify the commitlog bootstrapping logic. This is fine
+	// because this integration test protects against performance regressions
+	// not correctness.
+	t.SkipNow()
```
robskillington: Ok cool, can we open an issue for this and link to it from here?  Just so we're tracking it as an issue too?

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -125,9 +126,8 @@ func (f *flushState) getLastFlushAt() time.Time {
 }
 
 type writerState struct {
-	writer         commitLogWriter
-	writerExpireAt time.Time
-	activeFile     *File
+	writer     commitLogWriter
+	activeFile *persist.CommitlogFile
```
robskillington: Yeah, can we make this `persist.CommitLogFile` for consistency?  i.e. there is a `CommitLogWriter` etc so there is already precedence for camelcasing the two words CommitLog

src/dbnode/persist/fs/commitlog/files.go
```diff
@@ -27,20 +27,43 @@ import (
...
+	for _, f := range files {
+		if int(f.Index) >= newIndex {
+			newIndex = int(f.Index + 1)
+		}
+	}
+
+	for ; ; newIndex++ {
+		var (
+			prefix   = opts.FilesystemOptions().FilePathPrefix()
+			filePath = fs.CommitlogFilePath(prefix, time.Unix(0, 0), newIndex)
```
robskillington: Can you make `time.Unix(0,0)` a var at the top of this file and put a comment perhaps next to where you use it about why it's no longer required to be any value and can be zero?

```go
var timeNone = time.Unix(0,0)
```

src/dbnode/persist/fs/files.go
```diff
@@ -1315,6 +1329,16 @@ func OpenWritable(filePath string, perm os.FileMode) (*os.File, error) {
 	return os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, perm)
 }
 
+// CommitlogFilePath returns the path for a commitlog file.
+func CommitlogFilePath(prefix string, start time.Time, index int) string {
```
robskillington: nit: `CommitLogFilePath` (camel case the "commit" "log" words)

src/dbnode/persist/fs/msgpack/decoder.go
```diff
@@ -368,8 +368,11 @@ func (dec *Decoder) decodeLogInfo() schema.LogInfo {
 		return emptyLogInfo
 	}
 	var logInfo schema.LogInfo
-	logInfo.Start = dec.decodeVarint()
-	logInfo.Duration = dec.decodeVarint()
+
+	// Deprecated, have to decode anyways for backwards compatibility, but we ignore the values.
```
robskillington: Hm, we can remove the fields and just call `dec.decodeVarint()` twice here possibly.  Thoughts on that approach instead?  Not too bullish on either, but seems a little cleaner to remove the fields from the LogInfo struct.

src/dbnode/persist/fs/persist_manager.go
```diff
@@ -497,15 +534,60 @@ func (pm *persistManager) closeData() error {
...
+		CommitlogIdentifier: commitLogIdentifier,
+	})
+	if err != nil {
+		return fmt.Errorf("error writing out snapshot metadata file: %v", err)
+	}
+
+	return pm.doneShared()
+}
+
+func (pm *persistManager) doneShared() error {
```
robskillington: nit: Perhaps slightly better method name is `emitMetricsAndReset()`?

src/dbnode/persist/types.go
```diff
@@ -43,6 +45,26 @@ type PreparedDataPersist struct {
 	Close   DataCloser
 }
 
+// CommitlogFiles represents a slice of commitlog files.
+type CommitlogFiles []CommitlogFile
```
robskillington: nit: `CommitlogFiles` -> `CommitLogFiles`

src/dbnode/persist/types.go
```diff
@@ -43,6 +45,26 @@ type PreparedDataPersist struct {
...
+	for _, f := range c {
+		if f.FilePath == path {
+			return true
+		}
+	}
+	return false
+}
+
+// CommitlogFile represents a commit log file and its associated metadata.
+type CommitlogFile struct {
```
robskillington: nit: `CommitlogFile` -> `CommitLogFile`

src/dbnode/storage/cleanup.go
```diff
@@ -283,195 +275,193 @@ func (m *cleanupManager) cleanupExpiredNamespaceDataFiles(earliestToRetain time.
...
-func commitlogsContainPath(commitlogs []commitlog.File, path string) bool {
-	for _, f := range commitlogs {
-		if path == f.FilePath {
-			return true
-		}
+		filesToDelete = append(filesToDelete, errorWithPath.Path())
 	}
 
-	return false
+	return nil
```
robskillington: Hm should this not be `return finalErr` or just `return`?  

Can we add a test that makes sure errors truly come back if the `multiErr` is not nil?

src/dbnode/storage/flush.go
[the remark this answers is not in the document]
robskillington: Do we verify that state is `flushManagerSnapshotInProgress` when the index flush starts?  Or does that happen in the method that calls this one?

src/dbnode/storage/flush.go
```diff
@@ -200,12 +169,63 @@ func (m *flushManager) Flush(
...
+				multiErr = multiErr.Add(detailedErr)
+			}
+		}
+	}
+	m.maxBlocksSnapshottedByNamespace.Update(float64(maxBlocksSnapshottedByNamespace))
+
+	err = snapshotPersist.DoneSnapshot(snapshotID, rotatedCommitlogID)
+	if err != nil {
+		multiErr = multiErr.Add(err)
+	}
```
robskillington: nit: instead of `err != nil { multiErr = multiErr.Add(err) }` instead just:
```go
multiErr = multiErr.Add(snapshotPersist.DoneSnapshot(snapshotID, rotatedCommitlogID))
```

Since it will do the err == nil check for you?

src/dbnode/persist/fs/msgpack/decoder.go
richardartoul: I'd like to keep them in because the structs actually make for good ad-hoc documentation of the file format on disk (we keep all the fields in order)

src/dbnode/storage/cleanup.go
richardartoul: This is actually handled by the `defer` statement above which will reset the `finalErr` var to any errors in the `multiErr`. I can change it to `return finalErr` to make it less confusing and I'll make sure we have a test

src/cmd/services/m3dbnode/config/config.go
richardartoul: how about just: `TODO(v1): remove`

src/dbnode/storage/flush.go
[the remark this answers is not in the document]
richardartoul: We set it to `flushManagerIndexFLushInProgress` right before we do the index flush:

```
	m.setState(flushManagerIndexFlushInProgress)
		var (
			indexOpts    = ns.Options().IndexOptions()
			indexEnabled = indexOpts.Enabled()
		)
		if !indexEnabled {
			continue
		}
		multiErr = multiErr.Add(ns.FlushIndex(indexFlush))
	}
	multiErr = multiErr.Add(indexFlush.DoneIndex())
```

src/dbnode/integration/commitlog_bootstrap_only_reads_required_files_test.go
richardartoul: created a master task with all commitlog rotation / snapshotting remaining work: #1383

src/dbnode/persist/fs/commitlog/files.go
richardartoul: sure, done. In the future we can just remove it entirely which I'm pretty sure is backwards compatible (because we just list all the files in the directory then read their heads and ignore their filenames now) but I didn't want to tackle it in this P.R

src/dbnode/storage/cleanup.go
richardartoul: Ok I just verified existing logic was correct but I added a test case for it too

src/dbnode/storage/flush.go
richardartoul: yeah @justinjc had the same comment. I prefer the explicit error handling but I'll change it

PR #1175, opened 2018-11-14
[query] Add tag completion HTTP endpoint onto query

arnikola:
Adds the endpoint to query which will allow tag completion


src/query/models/matcher.go
```diff
@@ -108,3 +110,49 @@ func (m Matchers) String() string {
...
 }
+
+// TODO: make this more robust, handle types other than MatchEqual
+func matcherFromString(s string) (Matcher, error) {
+	ss := strings.Split(s, ":")
+	l := len(ss)
+	if len(ss) > 2 {
+		return Matcher{}, errors.New("invalid arg length for matcher")
+	}
+	fmt.Println(l)
```
nikunjgit: remove ?

src/query/models/matcher.go
```diff
@@ -108,3 +110,49 @@ func (m Matchers) String() string {
...
+
+	if l == 0 || len(ss[0]) == 0 {
+		return Matcher{}, errors.New("empty matcher")
+	}
+
+	if l == 1 {
+		return Matcher{
+			Type:  MatchEqual,
```
nikunjgit: this is a bit weird since you can in theory also have match not equal, match regex, etc.

src/query/storage/completed_tags.go
```diff
@@ -81,10 +81,13 @@ func (s completedTagsByName) Less(i, j int) bool {
 }
 
 func (b *completeTagsResultBuilder) Build() CompleteTagsResult {
-	result := make([]CompletedTag, len(b.tagBuilders))
+	result := make([]CompletedTag, 0, len(b.tagBuilders))
 	if b.nameOnly {
 		for name := range b.tagBuilders {
-			result = append(result, CompletedTag{Name: []byte(name)})
```
nikunjgit: oops, did it break any test now ?

src/query/models/matcher.go
arnikola: Addressed in the TODO; this is a bit annoying to do since it requires query parsing on our side

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -81,3 +97,145 @@ func ParseRequestTimeout(r *http.Request) (time.Duration, error) {
...
+		return nil, errors.ErrNoName
+	}
+
+	nameBytes := []byte(name)
+	return &storage.CompleteTagsQuery{
+		CompleteNameOnly: false,
+		FilterNameTags:   [][]byte{nameBytes},
+		TagMatchers: models.Matchers{
+			models.Matcher{
+				Type:  models.MatchEqual,
```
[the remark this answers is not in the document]
arnikola: Prom supports exact matchers for this endpoint

src/query/storage/completed_tags.go
arnikola: This was only added in this PR; once tests were added it failed a few

src/query/models/matcher.go
```diff
@@ -108,3 +110,47 @@ func (m Matchers) String() string {
 
 	return buffer.String()
 }
+
+// TODO: make this more robust, handle types other than MatchEqual
+func matcherFromString(s string) (Matcher, error) {
+	ss := strings.Split(s, ":")
+	l := len(ss)
```
nikunjgit: call it length ?

src/query/models/matcher.go
```diff
@@ -108,3 +110,47 @@ func (m Matchers) String() string {
...
+	if len(ss) > 2 {
```
nikunjgit: u already have a variable for that

arnikola: I just wanted to bump by characters per commit metrics

Issue #990, opened 2018-09-30
Coordinator MultiFetchResult Followups


prateek:
Followups from #989:
- [ ] Pooling multiResult (and underlying map)
- [ ] Using a genny map instead of using string key
- [ ] More unit coverage around correctness 
- [ ] Allowing configurable option when merging (instead of only picking highest res)

Issue #988, opened 2018-09-30
storage/local should use a genny-generated map so that a string does not need to be allocated for each key


richardartoul:
The ts/m3db/storage package was using a map[ident.ID] which is broken because even if two IDs are the same, their pointers will not be if they came from separate allocations. The tests were passing anyways, but once we updated the implementing of ident.StringID() the tests started breaking. We replaced the map[ident.ID] with a map[string] to make the build pass, but this will be really slow because a string needs to be allocated for each key. We should auto-generate a map that can handle ident.ID as the key.

PR #1177, opened 2018-11-14
[coordinator] Add genny generated maps to coordinator

arnikola:
Generates maps for multiresult merges, and uses them where appropriate.

Resolving https://github.com/m3db/m3/issues/988 and several points from https://github.com/m3db/m3/issues/990


src/query/storage/m3/multiresults/multi_fetch_result.go
```diff
@@ -166,24 +169,24 @@ func (r *multiResult) addOrUpdateDedupeMap(
...
 	for _, iter := range newIterators.Iters() {
-		id := iter.ID().String()
+		id := iter.ID().Bytes()
 
-		existing, exists := r.dedupeMap[id]
+		existing, exists := r.dedupeMap.Get(id)
 		if !exists {
 			// Does not exist, new addition
-			r.dedupeMap[id] = multiResultSeries{
+			r.dedupeMap.Set(id, multiResultSeries{
```
prateek: more a question: should this be `SetUnsafe`? i.e. can you rely on the bytes underlying `id` for the lifecycle you need in this request-response cycle?

src/query/functions/binary/common.go
```diff
@@ -109,8 +109,7 @@ func combineMetaAndSeriesMeta(
 	tags := models.NewTags(leftTags.Len(), leftTags.Opts)
 
 	for _, t := range leftTags.Tags {
-		name := string(t.Name)
-		if otherTag, ok := otherTags[name]; ok {
+		if otherTag, ok := otherTags[string(t.Name)]; ok {
```
prateek: should otherTags use a genny map too?

src/query/functions/aggregation/take.go
```diff
@@ -68,9 +68,10 @@ func NewTakeOp(
 
 // takeOp stores required properties for take ops
 type takeOp struct {
-	params   NodeParams
-	opType   string
-	takeFunc takeFunc
+	params      NodeParams
+	opType      string
+	opTypeBytes []byte
```
prateek: question: why have both opType and opTypeBytes?

src/query/storage/m3/types.go
```diff
@@ -83,15 +67,7 @@ type MultiFetchTagsResult interface {
 	)
 	// FinalResult returns a deduped list of tag iterators with
 	// corresponding series ids
-	FinalResult() ([]MultiTagResult, error)
+	FinalResult() ([]multiresults.MultiTagResult, error)
```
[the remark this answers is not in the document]
arnikola: Oops, this was actually already moved into `multiresults` and renamed to `MultiSearchResultBuilder`, but missed removing it from here

src/query/storage/m3/multiresults/multi_fetch_result.go
arnikola: Good call, these are all for builders that are much shorter lived than the objects they're building

src/query/functions/aggregation/take.go
arnikola: Just looking to avoid having to cast it to and from `string` where necessary

src/query/functions/binary/common.go
arnikola: Updated it to do so; would be a smaller map that would be used much less often than the others so was less concerned about putting a genny map here initially

src/query/functions/aggregation/base.go
```diff
@@ -72,9 +72,10 @@ func NewAggregationOp(
 
 // baseOp stores required properties for the baseOp
 type baseOp struct {
-	params NodeParams
-	opType string
-	aggFn  aggregationFn
+	params      NodeParams
+	opType      string
```
nikunjgit: confusing to have both opType and opTypeBytes. Maybe just have a OpType struct and hide it behind that ? Also, these functions are created once per query so i'm not sure what we are trying to optimize here ? This definitely makes it harder to understand.

src/query/functions/binary/common.go
```diff
@@ -80,10 +80,23 @@ var (
 	errNoMatching              = errors.New("vector matching parameters must be provided for binary operations between series")
 )
 
-func tagMap(t models.Tags) map[string]models.Tag {
-	m := make(map[string]models.Tag, t.Len())
+type tags models.Tag
+
+func toTagMap(t models.Tags) *tagMap {
```
nikunjgit: does this show up on the flamegraph ? This is done only once per series I assume so I'm not sure if we should be optimizing at the cost of simplicity.

src/query/util/execution/parallel.go
```diff
@@ -52,13 +52,10 @@ func ExecuteParallel(ctx context.Context, requests []Request) error {
...
 	g, ctx := errgroup.WithContext(ctx)
 	for _, req := range requests {
-		// Need to use a separate func since g.Go doesn't take input
 		req := req
-		func() {
-			g.Go(func() error {
-				return req.Process(ctx)
-			})
-		}()
+		g.Go(func() error {
```
nikunjgit: remove this change from the diff ?

src/query/functions/aggregation/base.go
arnikola: Fair, will standardize

src/query/functions/binary/tag_new_map_gen.go
```diff
@@ -0,0 +1,91 @@
...
+
+import (
+	"bytes"
+
+	"github.com/m3db/m3x/pool"
+
+	"github.com/cespare/xxhash"
+)
+
+// Copyright (c) 2018 Uber Technologies, Inc.
```
benraskin92: ??

arnikola:
When we pick this up again, may have to redo a lot of this work due to staleness

PR #1188, opened 2018-11-20
Respect numShards in /database/create

justinjc:


src/query/api/v1/handler/database/create_test.go
```diff
@@ -175,6 +175,106 @@ func TestLocalType(t *testing.T) {
...
+						"weight": 1,
+						"endpoint": "http://localhost:9000",
+						"shards": [],
+						"shardSetId": 0,
+						"hostname": "localhost",
+						"port": 9000
+					}
+				},
+				"replicaFactor": 0,
+				"numShards": 0,
```
richardartoul: Why would you expect 0 shards in the response?

justinjc:
Not in the real world, but in this case, we'd just be testing whether whether our [placement mock return](https://github.com/m3db/m3/pull/1188/files#diff-3a103a2ddb031aa6013f148ef4ddc14dR197) is what we expect.

The real thing we're testing in this new unit test is whether `51` gets passed in correctly [here](https://github.com/m3db/m3/pull/1188/files#diff-3a103a2ddb031aa6013f148ef4ddc14dR212).

src/query/api/v1/handler/database/create_test.go
```diff
@@ -175,6 +175,106 @@ func TestLocalType(t *testing.T) {
...
+						"repairEnabled": false,
+						"retentionOptions": {
+							"retentionPeriodNanos": "86400000000000",
+							"blockSizeNanos": "3600000000000",
+							"bufferFutureNanos": "120000000000",
+							"bufferPastNanos": "600000000000",
+							"blockDataExpiry": true,
+							"blockDataExpiryAfterNotAccessPeriodNanos": "300000000000"
+						},
+						"snapshotEnabled": false,
```
robskillington: Hm why is snapshotEnabled false when using the database/create method?

PR #1191, opened 2018-11-21
Update changelog for v0.4.8

richardartoul:


CHANGELOG.md
```diff
@@ -2,7 +2,23 @@ Changelog
 =========
 # (unreleased)
 
-# 0.4.7
+# 0.4.8 (2018-10-20)
+
+- **Coordinator**: Fix bugs in database creation API (error not displayed in logs and provided number of shards not honored)
```
robskillington: Since this is a fix, can you format as (to be consistent with other changelogs):
```md
- FIX **Coordinator**: 
```

robskillington:
To make consistent with the other changelogs we need to also add PR numbers (see other versions).

I can add a linter at some point so this is enforced perhaps.

PR #1195, opened 2018-11-22
Add support for concurrent index block queries

robskillington:


robskillington:
Ready to review now

src/dbnode/runtime/runtime_options.go
```diff
@@ -38,6 +38,10 @@ const (
 	// DefaultBootstrapConsistencyLevel is the default bootstrap consistency level
 	DefaultBootstrapConsistencyLevel = topology.ReadConsistencyLevelMajority
 
+	// DefaultIndexDefaultQueryTimeout is the hard timeout value to use if none is
+	// specified for a specific query, zero specifies no timeout.
+	DefaultIndexDefaultQueryTimeout = time.Minute
```
richardartoul: What value are you seeing in prod? I would have thought 15->30 would be a more reasonable default

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
+
+		// terminate early if we know we don't need any more results
+		if opts.Limit > 0 && results.Size() >= opts.Limit {
+			results.Lock()
+			exhaustive = false
+			results.Unlock()
+			break
+		}
+
+		execBlockQuery := func() {
```
richardartoul: The variable capture above and allocating this closure each time isn't required. I think you can reduce it all down to a single closure allocation by moving the function declaration above the loop and then changing the definition of the function to be:

```golang
func(block index.Block) {
    ...
}
```

and then down below it becomes

```golang
workers.Go(func() {
    execBlockQuery(block) // Capture happens for free here
    wg.Done()
})
```

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
+				// we have hit the limit and don't need to query any more.
+				exhaustive = false
+				return
+			}
+		}
+
+		if timeoutApplied := timeout > 0; !timeoutApplied {
+			// No timeout, just wait blockingly for a worker
+			wg.Add(1)
+			workers.Go(func() {
```
richardartoul: Can we add a comment here (or somewhere else) with a TODO to replace this with the pooled worker pool? Previously we were creating goroutines proportional to number of queries, now its proportional to number of queries * number of blocks. Still a probably small number, but I can see it having some impact if these stacks need to grow to 4/8KiB

richardartoul: In the comment you can basically say: "Swap this out with PooledWorkerPool once we have an implementation with timeout support" or something like that

src/dbnode/storage/index.go
richardartoul: Should reduce number of allocated functions per block from 2->1

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
+
+			if !currExhaustive {
+				// i.e. block had more data but we stopped early, we know
+				// we have hit the limit and don't need to query any more.
+				exhaustive = false
+				return
+			}
+		}
+
+		if timeoutApplied := timeout > 0; !timeoutApplied {
```
richardartoul: This would read better if it was `applyTimeout` or `shouldApplyTimeout`

src/dbnode/runtime/runtime_options.go
robskillington: Yeah, it could be, I just don't want to necessarily break some users existing queries.  I could be convinced using 30s instead though for sure.

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
+			})
+			continue
+		}
+
+		// Need to apply timeout
+		timedOut := false
+		now := i.nowFn()
+		if timeLeft := deadline.Sub(now); timeLeft > 0 {
+			wg.Add(1)
+			running := workers.GoWithTimeout(func() {
```
richardartoul: the `running` variable is kind of confusing. Maybe:

```golang
timedOut = !workers.GoWithTimeout(func() {
    ...
}
```

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
+			// Exceeded our deadline waiting for this block's query to start
+			return index.QueryResults{},
+				fmt.Errorf("index query exceeded timeout: %s", timeout.String())
+		}
+	}
+
+	// Wait for queries to finish
+	wg.Wait()
+
+	// NB(r): Dont' need to hold lock for concurrent results since
```
richardartoul: Should this comment go over the call to `results.Results()` below instead of the error check here?

richardartoul: Dont' -> Don't

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
+	// at this point all block queries must have finished.
+	// If we ever to early aborts we'll have to make sure that the results
```
richardartoul: This comment is a little confusingly written. I think the confusion lies around the fact that we already have early aborts, but I think what you're discussing here is being able to abort a goroutine that has already started querying. In that case, you'll need to make sure that you actually wait for all the goroutines to properly abort before reaching this point so they don't have access to the results, so maybe:

"We don't need to hold a lock for the results here since at this point all the block queries (which are the only other accessors) must have finished. If we ever add support for aborting block queries that have already started, we will need to make sure that we add some additional synchronization here OR make sure that that we wait until all the worker goroutines have finished "aborting" by this point so we know that nothing else will try and access the results." or something like that

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
+	return i.state.runtimeOpts.defaultQueryTimeout
+}
+
+func (i *nsIndex) overriddenOptsForQueryWithLock(
+	opts index.QueryOptions,
+) index.QueryOptions {
+	// Override query response limit if needed
 	if i.state.runtimeOpts.maxQueryLimit > 0 && (opts.Limit == 0 ||
 		int64(opts.Limit) > i.state.runtimeOpts.maxQueryLimit) {
 		i.logger.Debugf("overriding query response limit, requested: %d, max-allowed: %d",
```
richardartoul: Nit: I think queries are serializable now (at least the idx package has a `String` method on the query struct.

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
 	// some determinism about the results returned.
 	for _, start := range i.state.blockStartsDescOrder {
 		block, ok := i.state.blocksByTime[start]
 		if !ok { // should never happen
-			return index.QueryResults{}, i.missingBlockInvariantError(start)
+			return nil, i.missingBlockInvariantError(start)
 		}
 
-		// ensure the block has data requested by the query
+		if queryRange.IsEmpty() {
```
richardartoul: nit: This might be a little less confusing if it was the first part of the loop. As is, I was confused why you would need to check this after doing the block lookup, but its not related to that at all

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
+}
 
-		// terminate if queryRange doesn't need any more data
-		queryRange = queryRange.RemoveRange(blockRange)
-		if queryRange.IsEmpty() {
-			break
-		}
-	}
+func (i *nsIndex) newConcurrentResults(ctx context.Context) *index.ConcurrentResults {
+	results := i.opts.IndexOptions().ResultsPool().Get()
```
richardartoul: want to just throw this pool onto the struct itself like we do with the worker pool so you don't have to do the triple function call?

src/dbnode/storage/index.go
```diff
@@ -909,11 +1035,15 @@ func (i *nsIndex) Close() error {
...
 		return errDbIndexAlreadyClosed
 	}
+
 	i.state.closed = true
 
 	var multiErr xerrors.MultiError
 	multiErr = multiErr.Add(i.state.insertQueue.Stop())
 
+	// Wait for inflight queries to finish before closing blocks
+	i.queriesWg.Wait()
```
richardartoul: Might be safer (or less likely to deadlock when changes are made) if you do this wait outside the lock. So:

```golang
i.state.Lock()
defer i.state.Unlock()

if !i.isOpenWithRLock() {
    return errDbIndexAlreadyClosed
}

// Signal our intent to close so that we stop accepting reads / writes.
i.state.closed = true

i.state.Unlock()

// Wait for in-flight queries to finish before continuing.
i.queriesWg.Wait()

// Reacquire the lock so we can complete the shutdown.
// Not even sure you need to re-acquire the lock at this point if you're willing
// to remove the defer at the top since technically no one else can call Close()
// after you mark is closed.
i.state.Lock()

var multiErr xerror.MultiError
multiErr = multiErr.Add(i.state.insertQueue.Stop())

...
```

richardartoul: You understand the lifecycle better than I do, so take it or leave it

src/dbnode/storage/index/block.go
```diff
@@ -253,30 +257,74 @@ func (b *block) Query(
 		return false, err
 	}
 
-	var (
-		size       = results.Size()
-		brokeEarly = false
-	)
+	size := results.Size()
```
richardartoul: wanna add the var back? I think it was cleaner. I assume you removed it unintentionally while refactoring

src/dbnode/storage/index/block.go
```diff
@@ -253,30 +257,74 @@ func (b *block) Query(
...
+	limitedResults := false
 	execCloser := safeCloser{closable: exec}
 	iterCloser := safeCloser{closable: iter}
 
+	// NB(r): iter.Next() when moving between segments will actually
+	// execute the search phase which can take a long time, hence
+	// we only retrieve the results lock when we add a batch of documents
+	// to the results set.
+	batch := b.docsPool.Get()
+	// Use documentArrayPoolCapacity to as max batch to avoid growing outside
```
richardartoul: "Use the maximum capacity of the array pool as the max batch size to avoid growing outside the allowed pool capacity."

src/dbnode/storage/index/block.go
```diff
@@ -253,30 +258,74 @@ func (b *block) Query(
...
+
 	defer func() {
 		iterCloser.Close()
 		execCloser.Close()
+		b.docsPool.Put(batch)
 	}()
 
+	// NB(r): This query method only called once per block so is relatively
+	// cheap to declared as a lambda.
+	flushBatch := func() error {
```
richardartoul: could you eliminate the allocation entirely if you had a named function / method with the signature:

```golang
func flushBatch(batch []Document, results *ConcurrentResults) ([]Document, error)
```

and then you could use it like:

```golang
batch = append(batch, d)
if len(batch) < maxBatch {
    continue
}
batch, err = flushBatch(batch, results)
if err != nil {
    ...
```

Maybe overkill, but the anonymous func doesn't seem like its closing over many vars

src/dbnode/storage/index/block.go
```diff
@@ -253,30 +258,74 @@ func (b *block) Query(
 		return false, err
 	}
 
-	var (
-		size       = results.Size()
-		brokeEarly = false
-	)
+	size := results.Size()
+	limitedResults := false
```
richardartoul: can you just call this exhaustive and return it at the end instead of doing the limitedResults -> exhaustive translation

src/dbnode/storage/index/options.go
```diff
@@ -33,6 +34,14 @@ import (
 const (
 	// defaultIndexInsertMode sets the default indexing mode to synchronous.
 	defaultIndexInsertMode = InsertSync
+
+	// documentArrayPool size in general: 256*256*sizeof(doc.Document)
+	// = 256 * 256 * 16
+	// = 1mb (but with Go's heap probably 2mb)
+	// TODO(r): Make this configurable in a followup change.
```
richardartoul: We're probably gonna want this tunable really soon so we don't have to ask people to recompile when they're testing our changes, especially if their read workloads are very different from ours

src/dbnode/storage/index/results.go
```diff
@@ -84,25 +84,14 @@ func (r *results) Add(d doc.Document) (added bool, size int, err error) {
 func (r *results) tags(fields doc.Fields) ident.Tags {
 	tags := r.idPool.Tags()
 	for _, f := range fields {
-		tags.Append(ident.Tag{
-			Name:  r.copyBytes(f.Name),
-			Value: r.copyBytes(f.Value),
-		})
+		tags.Append(r.idPool.CloneTag(ident.Tag{
```
richardartoul: is this just a cleanup?

robskillington: Yup, this wasn't being pooled before, now it is (it always could have been).

src/dbnode/storage/index/block.go
robskillington: Reads worse, it was actually like this before.

src/dbnode/storage/index/options.go
robskillington: This is now only going to be used by the compactor, so it should be tiny (we never will have 256 concurrent compactions, and compactor just uses a single pool).

src/dbnode/storage/index.go
```diff
@@ -716,77 +723,196 @@ func (i *nsIndex) Query(
...
 ) (index.QueryResults, error) {
 	i.state.RLock()
-	defer i.state.RUnlock()
 	if !i.isOpenWithRLock() {
+		i.state.RUnlock()
 		return index.QueryResults{}, errDbIndexUnableToQueryClosed
 	}
 
-	// override query response limit if needed.
+	// Track this as an inflight query that needs to finish
```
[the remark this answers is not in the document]
robskillington: Done.

src/dbnode/storage/index.go
```diff
@@ -715,80 +730,298 @@ func (i *nsIndex) Query(
...
+		wg.Wait()
+	} else {
+		// Need to abort early if timeout hit.
+		timeLeft := deadline.Sub(i.nowFn())
+		if timeLeft <= 0 {
+			return index.QueryResults{}, fmt.Errorf("index query timed out: %s", timeout.String())
+		}
+
+		var (
+			ticker  = time.NewTicker(timeLeft)
```
richardartoul: I know Prateek has talked about time tickers not getting scheduled properly in certain situations and preferring to use loops with time.Sleep which apparently has some special hooks into the runtime, although then you'd have to implement polling so ehhhh. Probably fine as is

src/dbnode/storage/index.go
```diff
@@ -905,18 +1138,19 @@ func (i *nsIndex) CleanupExpiredFileSets(t time.Time) error {
 
 func (i *nsIndex) Close() error {
```
richardartoul: is there not a "Tick()" loop? How does the ticking get stopped

src/dbnode/runtime/runtime_options.go
robskillington: Yup, I opened here:
https://github.com/m3db/m3/issues/1220

src/dbnode/storage/index.go
robskillington: There is a tick loop yeah, see `Tick()` on `nsIndex`.

PR #1197, opened 2018-11-24
Add compaction of mutable segments for index blocks

robskillington:
This change uses the compaction planner to constantly move mutable segments from the index block into a compacted state.


src/dbnode/storage/index.go
```diff
@@ -332,7 +435,7 @@ func (i *nsIndex) WriteBatch(
 
 		// Re-sort the batch by initial enqueue order
 		if numErrs := batch.NumErrs(); numErrs > 0 {
-			// Restore the sort order from whene enqueued for the caller
+			// Restore the sort order from when enqueued for the caller
```
richardartoul: nit: `.`

src/dbnode/storage/index.go
```diff
@@ -343,28 +446,33 @@ func (i *nsIndex) WriteBatch(
...
 	pastLimit := now.Add(-1 * i.bufferPast)
-	writeBatchFn := i.writeBatchForBlockStartWithRLock
-	for _, batch := range batches {
-		// Ensure timestamp is not too old/new based on retention policies and that
-		// doc is valid.
-		batch.ForEach(func(idx int, entry index.WriteBatchEntry,
+	// NB(r): Release lock early to avoid writing batches impacting ticking
+	// speed, etc.
+	// Sometimes foreground compaction can take a long time during heavy inserts.
+	// Each lookup to ensureBlockPresent checks that index is still open, etc.
```
richardartoul: This comment could use a little clarification too. Looks like notes for a comment to write later lol

src/dbnode/storage/index.go
```diff
@@ -466,6 +580,13 @@ func (i *nsIndex) Bootstrap(
 	return multiErr.FinalError()
 }
 
+func (i *nsIndex) BootstrapsDone() uint {
+	i.state.RLock()
+	result := i.state.bootstrapsDone
```
richardartoul: Renaming this field and the method `NumBootstrapsDone()` would be more clear. I thought `bootstrapsDone` was an enum at first

src/m3ninx/index/segment/builder/builder.go
```diff
@@ -0,0 +1,240 @@
...
+	opts      Options
+	newUUIDFn util.NewUUIDFn
+
+	offset postings.ID
+
+	batchSizeOne index.Batch
+	docs         []doc.Document
+	idSet        *IDsMap
+	fields       *fieldsMap
+	uniqueFields [][]byte
```
richardartoul: Would it be better to have some kind of theoretical cap on this? Seems like it can grow unbounded

src/dbnode/storage/index/block.go
```diff
@@ -94,26 +162,45 @@ func NewBlock(
...
-	)
+	docsPool := opts.DocumentArrayPool()
 
-	// FOLLOWUP(prateek): use this to track segments when we have multiple segments in a Block.
-	postingsOffset := postings.ID(0)
-	seg, err := mem.NewSegment(postingsOffset, opts.MemSegmentOptions())
+	foregroundCompactor, err := compaction.NewCompactor(docsPool,
+		documentArrayPoolCapacity,
+		opts.SegmentBuilderOptions(),
+		opts.FSTSegmentOptions())
```
richardartoul: You need to set the registry cell size to zero here once we fix the underlying lib right?

src/dbnode/storage/index/block.go
```diff
@@ -94,26 +162,45 @@ func NewBlock(
...
+	if err != nil {
+		return nil, err
+	}
 
-		startTime: startTime,
-		endTime:   startTime.Add(blockSize),
-		blockSize: blockSize,
-		opts:      opts,
-		nsMD:      md,
+	segmentBuilder, err := builder.NewBuilderFromDocuments(opts.SegmentBuilderOptions())
```
richardartoul: Are there any other types of builders? The name is a little confusing because it makes it seem like the constructor should accept a slice of documents, but its not a huge deal

src/dbnode/storage/index/block.go
```diff
@@ -94,26 +162,45 @@ func NewBlock(
...
+	if err != nil {
+		return nil, err
+	}
+
+	blockSize := md.Options().IndexOptions().BlockSize()
+	iopts := opts.InstrumentOptions()
+	b := &block{
+		state:               blockStateOpen,
+		segmentBuilder:      segmentBuilder,
+		startTime:           startTime,
```
richardartoul: presumably this was truncated already and such? might be nice to rename it to `blockStart` if so

src/dbnode/storage/index/block.go
```diff
@@ -128,32 +215,254 @@ func (b *block) EndTime() time.Time {
...
+	for _, seg := range b.backgroundSegments {
+		segs = append(segs, compaction.Segment{
+			Age:     seg.Age(),
+			Size:    seg.Segment().Size(),
+			Type:    segments.FSTType,
+			Segment: seg.Segment(),
+		})
+	}
+
+	plan, err := compaction.NewPlan(segs, b.opts.BackgroundCompactionPlannerOptions())
```
richardartoul: Do you need to check if the segments are `Compactable` first? I I saw that method in `plan.go` but it doesn't seem to be used anywhere

src/dbnode/storage/index/block.go
```diff
@@ -128,32 +215,254 @@ func (b *block) EndTime() time.Time {
...
+		return
+	}
+
+	if len(plan.Tasks) == 0 {
+		return
+	}
+
+	// Kick off compaction
+	b.compactingBackground = true
+	go func() {
```
richardartoul: Not a huge fan of this pattern where you kick off a goroutine that at the end maybe kicks itself off again after checking some state...

Could the block just have a single-threaded background routine that checks for background segments and compacts them? seems simpler and you already have a `Close()` method for cleanup.

src/dbnode/storage/index/block.go
```diff
@@ -162,32 +471,242 @@ func (b *block) WriteBatch(inserts *WriteBatch) (WriteBatchResult, error) {
...
-		expectedReaders++
+func (b *block) foregroundCompactWithBuilder(builder segment.DocumentsBuilder) error {
+	// We inserted some documents, need to compact immediately into a
+	// foreground segment.
+	b.Lock()
+	foregroundSegments := b.foregroundSegments
+	b.Unlock()
+
+	segs := make([]compaction.Segment, 0, len(foregroundSegments)+1)
+	segs = append(segs, compaction.Segment{
```
richardartoul: intentional that the newest one comes first, or doesn't master?

src/dbnode/storage/index/block.go
```diff
@@ -162,32 +471,242 @@ func (b *block) WriteBatch(inserts *WriteBatch) (WriteBatchResult, error) {
...
+
+	plan, err := compaction.NewPlan(segs, b.opts.ForegroundCompactionPlannerOptions())
+	if err != nil {
+		return err
+	}
+
+	// Check plan.
+	var planErr error
+	switch {
+	case len(builder.Docs()) == 0:
```
richardartoul: Can this check be moved further up? Seems like you could have determined this at the beginning

src/dbnode/storage/index/block.go
```diff
@@ -162,32 +471,242 @@ func (b *block) WriteBatch(inserts *WriteBatch) (WriteBatchResult, error) {
...
 	}
+
+	plan, err := compaction.NewPlan(segs, b.opts.ForegroundCompactionPlannerOptions())
+	if err != nil {
+		return err
+	}
+
+	// Check plan.
+	var planErr error
+	switch {
```
richardartoul: switch statement seems a little weird to me since you're just trying to return an error. How is it different from just doing:

```golang
if len(plan.Tasks) == 0 {
    return errForegroundComapctorNoPlan
}

if !taskHasNilSegment(plan.Tasks[0]) {
    return errForegroundCompactorBadPlanFirstTask
}
...
```

src/dbnode/storage/index/block.go
```diff
@@ -162,32 +471,242 @@ func (b *block) WriteBatch(inserts *WriteBatch) (WriteBatchResult, error) {
...
+	case len(builder.Docs()) == 0:
+		// Must be compacting with a builder that has some inserted documents.
+		planErr = errForegroundCompactorNoBuilderDocs
+	case len(plan.Tasks) == 0:
+		// Must generate a plan since we gave it a mutable segment type to compact.
+		planErr = errForegroundCompactorNoPlan
+	case !taskHasNilSegment(plan.Tasks[0]):
```
richardartoul: I read through `plan.go` and still don't understand this.

src/dbnode/storage/index/block.go
```diff
@@ -162,32 +471,242 @@ func (b *block) WriteBatch(inserts *WriteBatch) (WriteBatchResult, error) {
...
+				xlog.NewField("cumulativeSize", summary.CumulativeSize),
+			).Debug("planned foreground compaction task")
+		}
+	}
+
+	// Run the plan.
+	sw := b.metrics.foregroundCompactionPlanRunLatency.Start()
+	defer sw.Stop()
+
+	// Run the first task, without resetting the builder.
```
richardartoul: is resetting the builder expensive?

src/dbnode/storage/index/block.go
```diff
@@ -162,32 +471,242 @@ func (b *block) WriteBatch(inserts *WriteBatch) (WriteBatchResult, error) {
...
+
+	result := b.addCompactedSegmentFromSegments(b.foregroundSegments,
+		segments, compacted)
+	b.foregroundSegments = result
+
+	return nil
+}
+
+func (b *block) cleanupForegroundCompactWithLock() {
+	// Check if need to close all the compacted segments due to
```
richardartoul: if we*

src/dbnode/storage/index/block.go
richardartoul: You have more context than I do, but after reading through the whole file the flow of operations is kind of hard to track. As far as I can tell, at a given time you can have both forground compactions going on, as well as background compactions.

Not saying we should necessarily change it now, but I wonder if it would be easier to understand this code if it was modeled as a pipeline where the flow is:

Write batch --> Foreground compactor (single threaded) --> Background compactor (single threaded)

and then that would be implemented with two goroutines and some channels.

src/dbnode/storage/index/compaction/compactor.go
```diff
@@ -0,0 +1,349 @@
...
+		buff:         bytes.NewBuffer(nil),
+	}, nil
+}
+
+// Compact will take a set of segments and compact them into an immutable
+// FST segment, if there is a single mutable segment it can directly be
+// converted into an FST segment, otherwise an intermediary mutable segment
+// (reused by the compactor between runs) is used to combine all the segments
+// together first before compacting into an FST segment.
+// Note: This is thread safe and only a single compaction may happen at a time.
```
richardartoul: not thread safe

src/dbnode/storage/index/compaction/compactor.go
```diff
@@ -0,0 +1,349 @@
...
+
+	return compacted, nil
+}
+
+func (c *Compactor) mmapAndAppendCloser(
+	fromBytes []byte,
+	closers *closers,
+) ([]byte, error) {
+	// Copy bytes to new mmap region to hide from the GC
+	mmapedResult, err := mmap.Bytes(int64(len(fromBytes)), mmap.Options{
```
richardartoul: will all these mmaps usually at least a few KiB in size? Every mmap costs us at least one full page so we should be careful to avoid allocating lots of them for small byte slices

src/dbnode/storage/index/segments.go
```diff
@@ -18,37 +18,30 @@
...
-var undefinedFSTTermsOffsetsMapFinalizeFn fstTermsOffsetsMapFinalizeFn = func([]byte) {
-	// NB: intentionally not defined to force users of the map to not
-	// allocate extra copies.
-	panic("not implemented")
+func (s *readableSeg) Segment() segment.Segment {
+	return s.segment
+}
+
+func (s *readableSeg) Age() time.Duration {
+	return time.Since(s.createdAt)
```
richardartoul: Should this be using a nowFn? I don't think we explicitly do `time.Now()` anywhere in the codebase

src/dbnode/storage/index/segments.go
```diff
@@ -18,37 +18,30 @@
...
+	segment   segment.Segment
 }
 
-var undefinedFSTTermsOffsetsMapCopyFn fstTermsOffsetsMapCopyFn = func([]byte) []byte {
-	// NB: intentionally not defined to force users of the map to not
-	// allocate extra copies.
-	panic("not implemented")
+func newReadableSeg(seg segment.Segment) *readableSeg {
+	return &readableSeg{
+		createdAt: time.Now(),
```
richardartoul: same comment as below

src/dbnode/storage/namespace.go
```diff
@@ -602,6 +602,13 @@ func (n *dbNamespace) QueryIDs(
 		n.metrics.queryIDs.ReportError(n.nowFn().Sub(callStart))
 		return index.QueryResults{}, errNamespaceIndexingDisabled
 	}
+
+	if n.reverseIndex.BootstrapsDone() < 1 {
```
richardartoul: interesting...wasn't this handled at the RPC level before? Whats the change

src/dbnode/storage/namespace/dynamic.go
```diff
@@ -107,8 +107,8 @@ func newDynamicRegistry(opts DynamicOptions) (Registry, error) {
 	}
 
 	logger := opts.InstrumentOptions().Logger()
-	logger.Info(`waiting for dynamic namespace registry initialization.
-		If this takes a long time, make sure that a namespace is configured`)
+	logger.Info("waiting for dynamic namespace registry initialization, " +
```
richardartoul: lol +1

src/m3ninx/index/segment/fst/fst_terms_postings_iterator.go
```diff
@@ -0,0 +1,117 @@
...
+	sgmt "github.com/m3db/m3/src/m3ninx/index/segment"
+	"github.com/m3db/m3/src/m3ninx/postings"
+	postingsroaring "github.com/m3db/m3/src/m3ninx/postings/roaring"
+	"github.com/pilosa/pilosa/roaring"
+)
+
+// postingsIterRoaringPoolingConfig uses a configuration that avoids allocating
+// any containers in the roaring bitmap, since these roaring bitmaps are backed
+// by mmaps and don't have any native containers themselves.
+var postingsIterRoaringPoolingConfig = roaring.ContainerPoolingConfiguration{
```
richardartoul: I think if you don't want any containers you can just use the non pooling one?

src/m3ninx/persist/reader.go
```diff
@@ -116,3 +119,58 @@ func newSafeIndexSegmentFileSetCloser(fileset IndexSegmentFileSet) io.Closer {
 	}
 	return x.NewSafeMultiCloser(closers...)
 }
+
+// NewMmapedIndexSegmentFile returns an IndexSegmentFile backed by the provided bytes and FD.
+// The returned object assumes ownership of the input fd, and mmap-ed bytes.
+func NewMmapedIndexSegmentFile(
```
richardartoul: Why does this keep track of the `fd` at all? You don't need the fd to maintain an mmap

src/dbnode/storage/index/block.go
robskillington: The compaction planner is the one that actually looks at that and then will return it as "UnusedSegments" if they aren't actually compactable.

src/dbnode/storage/index/block.go
robskillington: Yeah so the thing is the background compaction loop needs to keep going indefinitely until it really can't anymore.  That's why you can't trigger it necessarily from the foreground goroutine and have it kick of sequentially each time.  

Some times it will be in the middle of a background compaction, and you need to make sure it kicks itself back off again at the end of it's current run.  That's why you have to have when it finishes, it needs to check if it needs to run again or not.

src/dbnode/storage/index/block.go
robskillington: Doesn't really matter, the planner sorts everything first and does its own selection/grouping, etc regardless of input order.

src/dbnode/storage/index/block.go
robskillington: Sure thing, can do.

src/dbnode/storage/index/block.go
robskillington: It's not, I thought I'd just do a final sanity check and group them all together.  

I'll break it up to just the default Go-ism of if err/ret, if err/ret, etc to match what you suggest.

src/dbnode/storage/index/block.go
robskillington: Yeah so... it's kinda fucked up, so basically the planner thinks everything is always a `segment.Segment` and there's no way for me to put the `Builder` on the `compaction.Segment` struct on the `Segment segment.Segment` field.  

So what I do is set all the other metadata on the `compaction.Segment` struct but leave `Segment segment.Segment` field as `nil`.  It still does the right thing selecting that to be the first one to combine with other segments, but I just basically need to make sure that it always wants to compact it, so I can safely know when I see a `nil` segment I basically need to do a compaction with the builder I have on hand.

I can't really think of a way to do this nicely, possibly I could create a smaller scoped interface to pass to the `compaction` code so the `Builder` could be passed in place there too?  Thoughts?

src/dbnode/storage/index/block.go
robskillington: So this is confusing, but basically the first task is required to always include the builder (because its a mutable segment so must be compacted immediately by the settings we specify for the planner for foreground compaction), that's why I do the sanity check making sure the first task is the `nil` Segment that signifies it represents the builder.  

The builder already has a ton of documents from the `WriteBatch(...)` call, hence must not be reset, otherwise the documents it has doesn't get compacted into a new segment or with another segment.

src/dbnode/storage/index/compaction/compactor.go
```diff
@@ -0,0 +1,349 @@
...
+
+	// Cleanup incase we run into issues
+	defer func() {
+		if !success {
+			closers.Close()
+		}
+	}()
+
+	c.buff.Reset()
+	if err := c.writer.WritePostingsOffsets(c.buff); err != nil {
```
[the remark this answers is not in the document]
robskillington: Yeah, would be pretty hard I think... wish there was like a mmap backed bytes.Buffer that grew by 2x and realloc'd each time.  Although that'd get expensive too since it's a syscall each time.

src/dbnode/storage/index/compaction/compactor.go
robskillington: Yeah most of these will be like 10mb at least, it's 3 really huge files basically.

src/dbnode/storage/index/segments.go
robskillington: Yeah should do, I'll update it.

src/dbnode/storage/index/segments.go
robskillington: Yup, sounds good.

src/dbnode/storage/namespace.go
robskillington: I think I found that actually the query service when testing shadow queries was actually returning for purely index queries a lot actually when not bootstrapped, which is definitely incorrect since we don't have any of the data and only in memory (incomplete data).

src/dbnode/storage/namespace/dynamic.go
robskillington: Heh, I was reading the logs and like... why is there no timestamp for this log line?  Ohh... it's actually a multiline log line with an unencoded endline hah.

src/m3ninx/index/segment/builder/multi_segments_builder.go
```diff
@@ -0,0 +1,200 @@
...
+
+func (b *builderFromSegments) Reset(offset postings.ID) {
+	b.offset = offset
+
+	// Reset the documents slice
+	var emptyDoc doc.Document
+	for i := range b.docs {
+		b.docs[i] = emptyDoc
+	}
+	b.docs = b.docs[:0]
```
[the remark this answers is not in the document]
robskillington: Same answer as before, let me know what you think

src/m3ninx/index/segment/builder/multi_segments_builder.go
```diff
@@ -0,0 +1,200 @@
...
+	offset  postings.ID
+	// duplicates is a lookup of document IDs are duplicates
+	// in this segment, that is documents that are already
+	// contained by other segments and hence should not be
+	// returned when looking up documents.
+	duplicatesAsc []postings.ID
+}
+
+// NewBuilderFromSegments returns a new builder from segments.
+func NewBuilderFromSegments(opts Options) (segment.SegmentsBuilder, error) {
```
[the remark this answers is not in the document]
robskillington: Hm, no reason I suppose, can remove it.

src/m3ninx/index/segment/fst/fst_terms_postings_iterator.go
robskillington: We want the containers to be reused when Reset() is called on them and then they are used again though.  

Otherwise it starts allocating no?   I thought if we don't specify `NewBitmapWithPooling(...)` that containers never get returned when a call to reset is used?  We reuse a single bitmap very frequently since we're iterating over all terms from an existing FST segment (call UnmarshalBinary on each terms over and over again) as we're iterating over the entire FST and reading the postings list for every term in the segment.

src/m3ninx/persist/reader.go
robskillington: Can you safely close the FD though?  Wasn't 100% sure of that.  Probably can I suppose?

robskillington: https://stackoverflow.com/questions/17490033/do-i-need-to-keep-a-file-open-after-calling-mmap-on-it

I suppose we can then?

src/m3ninx/index/segment/builder/builder.go
richardartoul: If it dies after a block rotation I'm cool with it

src/dbnode/storage/index/block.go
richardartoul: yeah ended up seeing that later, looks good

src/dbnode/storage/index/block.go
richardartoul: Are you certain? I did a quick grep of the codebase and can't find any code that calls `segment.Compactable()` except for `plan_test.go`

src/dbnode/storage/index/block.go
richardartoul: Yeah I guess it just seems like you could have some backgroung goroutines and some message passing or something like that, and then the background routine can receive a close signal and do any remaining work before it shuts down if it needs to keep going after a call to Close(), but I'll leave it to you. I'm sure it could be one of those things where you try an alternative that seems cleaner and it actually turns out way worse

src/dbnode/storage/index/block.go
richardartoul: kk cool I think that would be good, the fact that it diverged made me stop and spend some time on it to see if it was doing something unexpected

src/dbnode/storage/index/compaction/compactor.go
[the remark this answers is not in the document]
richardartoul: Yeah...I'm like this close to writing a crappy `malloc` implementation to manage our mmaps so we can use them more aggressively lol

src/dbnode/storage/namespace.go
richardartoul: huh interesting...I guess is there a reason we can't use the bootstrap state on the namespace?

src/m3ninx/index/segment/builder/multi_segments_builder.go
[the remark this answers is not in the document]
richardartoul: as long as it gets cleared eventually I'm cool with it.

src/m3ninx/index/segment/builder/multi_segments_builder.go
```diff
@@ -0,0 +1,200 @@
...
+	// Reset all entries in ID set
+	b.idSet.Reset()
+
+	// Reset the segments metadata
+	b.segmentsOffset = 0
+	var emptySegment segmentMetadata
+	for i := range b.segments {
+		b.segments[i] = emptySegment
+	}
+	b.segments = b.segments[:0]
```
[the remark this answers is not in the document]
richardartoul: some response

src/m3ninx/index/segment/fst/fst_terms_postings_iterator.go
richardartoul: Ok cool you want the pooling. What I'm saying is with the configuration you have:

```golang
var postingsIterRoaringPoolingConfig = roaring.ContainerPoolingConfiguration{
 	AllocateArray:                   false,
	AllocateRuns:                    false,
	AllocateBitmap:                  false,
	MaxCapacity:                     128,
	MaxKeysAndContainersSliceLength: 128 * 10,
}
```

Every call to Reset will discard all the containers. If you want to keep them around (and there is only one bitmap so you don't mind it getting kind of big) I would do something like:

```golang
var postingsIterRoaringPoolingConfig = roaring.ContainerPoolingConfiguration{
 	AllocateArray:                   true,
	AllocateRuns:                    true,
	AllocateBitmap:                  true,
	MaxCapacity:                     128,
	MaxKeysAndContainersSliceLength: 128 * 10,
}
```

Although I'd like to get this P.R merged in: https://github.com/m3db/pilosa/pull/4/

and then you can just do:

```golang
roaring.NewContainerWithDefaultPoolingConfiguration(128)
```

src/m3ninx/persist/reader.go
richardartoul: Yeah I think its fine, see our `seeker.Open()` implementation: https://github.com/m3db/m3/blob/master/src/dbnode/persist/fs/seek.go#L159

We open all the files, defer a close on all the FDs, and then mmap what we need.

src/dbnode/storage/index/block.go
robskillington: I'll just remove `Compactable`, it doesn't really make sense for us.

robskillington: Done.

src/dbnode/storage/index/block.go
robskillington: Yeah I played with a few different ways, and this way was actually cleanest.  Especially since both foreground and compaction could be modeled similarly.

src/dbnode/storage/index/block.go
robskillington: I changed it so it counts the number of "builders" used, should make it a little more simple.

src/cmd/services/m3dbnode/config/config_test.go
```diff
@@ -538,6 +542,10 @@ db:
         size: 8192
         lowWatermark: 0.01
         highWatermark: 0.02
+    postingsListPool:
```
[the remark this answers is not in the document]
robskillington: Not yet, although it'll be useful to have in the future.

src/dbnode/storage/namespace.go
robskillington: When shards are added/removed and namespace needs another bootstrap, the bootstrap get's reset to "Bootstrapping" and we still need to return data at that point:
```go
n.bootstrapState = Bootstrapping
```

src/dbnode/storage/index/compaction/compactor.go
```diff
@@ -0,0 +1,358 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
richardartoul: 2019

robskillington: As talked, going to leave these as a fair few were from 2018 hah.

PR #1199, opened 2018-11-26
Add TLA+ specification for commitlog / snapshotting / rotation / cleanup logic.

richardartoul:
This is the first TLA+ specification for M3DB and is meant to be a relatively simple piece that we can use as a basis for introducing additional specifications. This spec models how M3DB acks writes by storing them in the commitlog, as well as the background commit log rotation / snapshotting (including snapshot failures) / cleanup process at a very high level. 

Running TLC with the following parameters:

```
numWrites <- 40
minNumWritesForPersistence <- 20
numClients <- 3
minNumWritesForCleanup <- 20
```

and requiring the following invariants be checked at each step:

```
AllAckedWritesAreBootstrappable
```

results in TLC finding no invariant violations along with the following model checking results

screen shot 2018-11-25 at 9 30 14 pm

### DIFF
--- a/specs/README.md
+++ b/specs/README.md
@@ -1,0 +1,20 @@
+# Specifications
+
+This directory contains specifications for various M3 components that are used for documentation and/or verification.
+
+## TLA+ Specifications
+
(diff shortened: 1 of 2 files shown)


richardartoul:
@schallert Sure, done

PR #1200, opened 2018-11-26
Add snapshot ID field to info files

richardartoul:


src/dbnode/persist/fs/files.go
```diff
@@ -74,25 +74,27 @@ type FileSetFile struct {
...
+	CachedSnapshotID   []byte
 	filePathPrefix     string
 }
 
-// SnapshotTime returns the SnapshotTime for the given FileSetFile. Value is meaningless
-// if the the FileSetFile is a flush instead of a snapshot.
-func (f *FileSetFile) SnapshotTime() (time.Time, error) {
-	if !f.CachedSnapshotTime.IsZero() {
+// SnapshotTimeAndID returns the snapshot time and id for the given FileSetFile.
+// Value is meaningless if the the FileSetFile is a flush instead of a snapshot.
```
robskillington: Can we validate it's a snapshot fileset file with a name check or something to return an error for this case?

src/dbnode/persist/fs/msgpack/encoder.go
```diff
@@ -53,21 +53,32 @@ type Encoder struct {
 	legacy legacyEncodingOptions
 }
 
+type legacyEncodingIndexInfoVersion int
+
+const (
+	// List in reverse order to ensure default value is current version.
+	legacyEncodingIndexVersionCurrent legacyEncodingIndexInfoVersion = iota
+	legacyEncodingIndexVersionV2
+	legacyEncodingIndexVersionV1
```
robskillington: To avoid Current being the default value, just create a const that is the Default that points to the current?

The opposite direction is just so much more natural and clean (avoiding capturing things like the default value being the zero value, which seems brittle). 

```go
const (
  legacyEncodingIndexInfoVersionCurrent = legacyEncodingIndexVersionCurrent

  legacyEncodingIndexInfoVersionV1 legacyEncodingIndexInfoVersion = iota
  legacyEncodingIndexInfoVersionV2
  legacyEncodingIndexInfoVersionV3 
)
```

PR #1202, opened 2018-11-26
[query] Add endpoint for matching series

arnikola:
Adds an endpoint for matching results to match prometheus endpoint

### DIFF
--- a/src/query/api/v1/handler/prometheus/common.go
+++ b/src/query/api/v1/handler/prometheus/common.go
@@ -29,12 +29,15 @@ import (
 
 	"github.com/m3db/m3/src/query/errors"
 	"github.com/m3db/m3/src/query/models"
+	xpromql "github.com/m3db/m3/src/query/parser/promql"
 	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/query/util"
 	"github.com/m3db/m3/src/query/util/json"
 	"github.com/m3db/m3/src/x/net/http"
 
 	"github.com/golang/snappy"
 	"github.com/gorilla/mux"
+	"github.com/prometheus/prometheus/promql"
 )
 
 const (
@@ -51,6 +54,7 @@ const (
 
 var (
 	matchValues = []byte("*")
+	maxTime     = time.Unix(1<<63-62135596801, 999999999)
 )
 
 // ParsePromCompressedRequest parses a snappy compressed request from Prometheus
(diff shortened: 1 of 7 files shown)


src/query/api/v1/handler/prometheus/common.go
```diff
@@ -239,3 +298,64 @@ func RenderTagCompletionResultsJSON(
...
+
+		jw.EndObject()
+		return
+	}
+
+	firstResult := completedTags[0]
+	name := string(firstResult.Name)
+
+	for _, value := range firstResult.Values {
+		copiedTags := make([]tag, len(tags)+1)
```
nikunjgit: can you move this up and reuse ?

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -239,3 +298,64 @@ func RenderTagCompletionResultsJSON(
...
+	}
+}
+
+func writeTags(
+	jw *json.Writer,
+	results []*storage.CompleteTagsResult,
+) {
+	for _, result := range results {
+		jw.BeginArray()
+		writeTagsHelper(jw, result.CompletedTags, []tag{})
```
nikunjgit: pass nil as last argument instead of empty slice

src/query/api/v1/handler/prometheus/common_test.go
```diff
@@ -75,3 +77,72 @@ func TestTimeoutParse(t *testing.T) {
...
+type writer struct {
+	value string
+}
+
+func (w *writer) Write(p []byte) (n int, err error) {
+	w.value = string(p)
+	return len(p), nil
+}
+
+func makeResult() []*storage.CompleteTagsResult {
```
nikunjgit: test a case with no tags ?

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -152,6 +156,61 @@ func parseTagCompletionQuery(r *http.Request) (string, error) {
...
+	r *http.Request,
+	tagOptions models.TagOptions,
+) (*storage.SeriesMatchQuery, *xhttp.ParseError) {
+	r.ParseForm()
+	matcherValues := r.Form["match[]"]
+	if len(matcherValues) == 0 {
+		return nil, xhttp.NewParseError(errors.ErrInvalidMatchers, http.StatusBadRequest)
+	}
+
+	start, err := parseTimeWithDefault(r, "start", time.Time{})
```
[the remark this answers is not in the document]
arnikola: Prometheus does from timeMin to timeMax so this should mirror that. We shouldn't get issues with a long time range since we don't fetch any points

Potentially could be an issue if we have old IDs in a glacier cluster, but even then it shouldn't be too bad if the match query is well constructed

src/query/api/v1/handler/prometheus/common.go
arnikola: Sounds good

src/query/api/v1/handler/prometheus/common.go
arnikola: Sounds good

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -152,6 +156,61 @@ func parseTagCompletionQuery(r *http.Request) (string, error) {
...
+	if err != nil {
+		return nil, xhttp.NewParseError(err, http.StatusBadRequest)
+	}
+
+	end, err := parseTimeWithDefault(r, "end", maxTime)
```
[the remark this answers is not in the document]
arnikola: Changed it to time.now

src/query/api/v1/handler/prometheus/common.go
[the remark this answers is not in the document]
arnikola: Changed it to a 40 day lookback after conferring offline

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -152,6 +155,61 @@ func parseTagCompletionQuery(r *http.Request) (string, error) {
...
+	r *http.Request,
+	tagOptions models.TagOptions,
+) (*storage.SeriesMatchQuery, *xhttp.ParseError) {
+	r.ParseForm()
+	matcherValues := r.Form["match[]"]
+	if len(matcherValues) == 0 {
+		return nil, xhttp.NewParseError(errors.ErrInvalidMatchers, http.StatusBadRequest)
+	}
+
+	start, err := parseTimeWithDefault(r, "start", time.Now().Add(time.Hour*24*-40))
```
nikunjgit: move the default to a constant ?

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -152,6 +155,61 @@ func parseTagCompletionQuery(r *http.Request) (string, error) {
...
+	if err != nil {
+		return nil, xhttp.NewParseError(err, http.StatusBadRequest)
+	}
+
+	end, err := parseTimeWithDefault(r, "end", time.Now())
```
nikunjgit: move the default to a constant ? eg: endDefault

src/query/api/v1/handler/prometheus/common.go
arnikola: It uses time.Now, should not be a constant

src/query/api/v1/handler/prometheus/common.go
arnikola: It uses time.Now, should not be a constant

PR #1209, opened 2018-11-27
[Aggregator][Ingest Aggregated Metrics - Part 1] Set Aggregated Value in Metric Elements

jeromefroe:
Allow callers to set aggregated values in the metric elements.


jeromefroe:
Talked offline, going with a different approach to support proxying metrics.

PR #1210, opened 2018-11-27
[integration] docker: add test for coord+db colo

schallert:
We unintentionally broke colo'd dbnode + coordinator downsampling in a
recent PR (#1205), but didn't catch it in tests. This adds an integration test
for such. We expect this PR to fail until pending async downsampling PR (#1206)
lands.

Verified that this 1) fails on master and 2) worked at `HEAD~2` prior to
breaking PR #1205.

### DIFF
--- a/scripts/docker-integration-tests/prometheus-colo/m3dbnode.yml
+++ b/scripts/docker-integration-tests/prometheus-colo/m3dbnode.yml
@@ -1,0 +1,250 @@
+coordinator:
+  listenAddress:
+    type: "config"
+    value: "0.0.0.0:7201"
+
+  metrics:
+    scope:
+      prefix: "coordinator"
+    prometheus:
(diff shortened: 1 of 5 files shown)


scripts/docker-integration-tests/prometheus-colo/m3dbnode.yml
```diff
@@ -0,0 +1,250 @@
...
+
+  clusters:
+  - namespaces:
+      - namespace: agg
+        type: aggregated
+        retention: 10h
+        resolution: 15s
+      - namespace: unagg
+        type: unaggregated
+        retention: 10m
```
robskillington: Yay, testing the downsampler code path too =]

schallert:
Hit some issues with this PR where it wasn't failing as I expected prior to landing the fix. Will investigate further.

PR #1211, opened 2018-11-27
Use a random test port in server_test.go

andrewmains12:
I noticed that my tests were failing locally while I had a query instance running, and tracked it down to this. Pretty small; just switches the port to a different fixed port. Random ephemeral would be ideal, but that would require passing a `net.Listener` through to `Run` as described [here](https://stackoverflow.com/questions/43424787/how-to-use-next-available-port-in-http-listenandserve), and it didn't seem worth it.

### DIFF
--- a/src/query/server/server_test.go
+++ b/src/query/server/server_test.go
@@ -47,10 +47,12 @@ import (
 	"google.golang.org/grpc"
 )
 
+var queryPort = 25123
+
 var configYAML = `
 listenAddress:
   type: "config"
-  value: "127.0.0.1:7201"
+  value: "127.0.0.1:25123"
 
 metrics:
   scope:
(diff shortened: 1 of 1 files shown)


src/query/server/server_test.go
```diff
@@ -47,10 +47,12 @@ import (
 	"google.golang.org/grpc"
 )
 
+var queryPort = 25123
+
 var configYAML = `
 listenAddress:
   type: "config"
-  value: "127.0.0.1:7201"
+  value: "127.0.0.1:25123"
```
arnikola: Might be worth templating with sprintf? Not really a concern since it's a test

andrewmains12: Thought about it, but I figured it'd be more difficult to look for the %d's in a long config string than to just have the hardcoding.

PR #1213, opened 2018-11-28
[query] Add ts datapoint to unconsolidated block converter

arnikola:
Converts ts.datapoints into unconsolidated blocks, given some consolidation information.

### DIFF
--- a/src/query/storage/inmemory/converter.go
+++ b/src/query/storage/inmemory/converter.go
@@ -1,0 +1,62 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


src/query/storage/inmemory/converter.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	"time"
+
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/query/ts"
+)
+
+func datapointsToBlocks(
```
andrewmains12: What's the eventual use case for this?

arnikola: Was an approach for Ben's new stuff spun out of my existing PR, but he decided to go a different way. Will close

PR #1215, opened 2018-11-28
Port the x/cost package from statsdex

andrewmains12:
This is a prerequisite diff for cost accounting in m3query. It lifts the `cost` package from statsdex, modifying it slightly to work in the `m3` monorepo context, and tweaking some interfaces that I found inconvenient. I use this in https://github.com/m3db/m3/pull/1207 to implement datapoint based cost accounting. 

Note: I haven't maintained the statsdex commit history here, but I definitely can if desired.


src/x/cost/limit_manager_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"github.com/m3db/m3/src/cluster/generated/proto/commonpb"
+	"github.com/m3db/m3/src/cluster/kv/mem"
+
+	"github.com/fortytw2/leaktest"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDynamicLimitManager(t *testing.T) {
+	var (
```
arnikola: More of a question, but if all vars have init values, should we do it like this, or initialize them one by one with `var := x`?

src/x/cost/types.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"math"
+)
+
+// Cost represents the cost of an operation.
+type Cost float64
+
+const (
+	// MaxCost is the maximum cost of an operation.
+	MaxCost = Cost(math.MaxFloat64)
```
arnikola: This seems a bit high? Also if it's just a default value maybe it shouldn't be exported

robskillington:
Just to start the discussion, I was hoping to get a question on our strategy when it comes to limits.

I wanted to ask the question why we are porting our soft limits implementation (based on datapoints, etc) when I thought we wanted to go after providing harder limits (i.e. bytes, etc) in the new query service?

cc @nikunjgit

andrewmains12:
> Just to start the discussion, I was hoping to ask the question on our strategy when it comes to limits.
> 
> I wanted to ask why we are porting our soft limits implementation (based on datapoints, etc) when I thought we wanted to go after providing harder limits (i.e. bytes, etc) in the new query service?
> 
> cc @nikunjgit

Excellent question, which I hope will be mostly answered in full in the [followup PR](https://github.com/m3db/m3/pull/1207) to this one. Right now my strategy there is to count and limit datapoints at two points: as they're decompressed and as they're readded to query blocks during query processing. I *think* that this ought to make the limits "harder", since it should capture most of the memory usage of the query engine as it happens. Limits are currently still expressed in terms of datapoints, but I can definitely switch that to bytes if that seems like a more intuitive way to configure things (which thinking about it, I think it probably does :) ).

I think that this approach parallels that of InfluxDB reasonably closely; they use their `Allocator` type (which handles resource limit enforcement). 

Is that closer to the approach you were considering @robskillington ?

src/x/cost/limit_manager_test.go
andrewmains12: Typically I'd do it `var := x` style, but given that this code is directly lifted from statsdex I think I'll leave it unless you feel strongly about it (since I don't want to go through and change small style things everywhere on it).

arnikola: Probably better to work towards a more consistent style for open source, but this is a super minor nit

src/x/cost/limit_manager.go
```diff
@@ -0,0 +1,178 @@
...
+		store,
+		kvEnabledKey,
+		m.enabled,
+		defaultEnabled,
+		watchOpts,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("unable to watch key '%s': %v", kvEnabledKey, err)
+	}
+	m.enabledWatcher = enabledWatcher
```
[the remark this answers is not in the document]
andrewmains12: What's your desired goal state with these? 

```
m.enabledWatcher = enabledWatcher
return m, nil
```
?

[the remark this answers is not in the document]
andrewmains12: Btw, bit of a personal rant (not to be taken too seriously :) )

My personal strong opinion about strong style opinions of this variety is that they should be fixed if one of the following is met:

1. Readability is *significantly* impacted (and I would consider that rare--think a crazy long line).
2. They violate a rule imposed by an automated tool (e.g. gofmt).
3. They violate a rule imposed by a standard style guide for the project.

Otherwise, I wouldn't really consider it worth the time.

That said, all of that is just my opinion, and I will adjust these if you like!

src/x/cost/limit_manager_test.go
andrewmains12: Haha yeah fair enough--fixed!

src/x/cost/limit_manager.go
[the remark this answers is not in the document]
arnikola: Probably something like this:

```
   return nil, fmt.Errorf("unable to watch key '%s': %v", kvEnabledKey, err)
  }
 
 m.enabledWatcher = enabledWatcher
 return m, nil
}
```

I think general style approach we were doing is always newline after a `}`, but that may be a query-specific quirk.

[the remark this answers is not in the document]
andrewmains12: Haha sounds good. I'll do that for future PR's, but I think I'm going to leave them here, since there are probably more instances than we can reasonably find without an automated tool (due to the fact that this is copied in code).

src/x/cost/enforcer.go
```diff
@@ -0,0 +1,168 @@
...
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
```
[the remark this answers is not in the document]
arnikola: Fun (?) question; should these be 2019 now to reflect what it will be when merged, or is it when the PR was initially authored? I think 2018 is correct, but who knows :p

src/x/cost/enforcer.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+// Report is a report on the cost limits of an Enforcer.
+type Report struct {
+	Cost
+	Error error
+}
+
+// EnforcerIF instances enforce cost limits for operations.
+type EnforcerIF interface {
```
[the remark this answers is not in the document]
arnikola: I tend to lean towards m3 style when possible, but that's probably an artifact of m3 being the first golang codebase I've worked on, so ymmv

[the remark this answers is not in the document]
andrewmains12: Yeah this is definitely a case where it's worth going with the style of the code base instead of introducing an external "standard".

src/x/cost/enforcer.go
```diff
@@ -0,0 +1,168 @@
...
+func (e *enforcer) Clone() Enforcer {
+	return &enforcer{
+		LimitManager: e.LimitManager,
+		tracker:      NewTracker(),
+		costMsg:      e.costMsg,
+		metrics:      e.metrics,
+	}
+}
+
+func (e *enforcer) checkLimit(cost Cost, limit Limit) error {
```
nikunjgit: checkLimit is called both from State() and Add(). Would e.metrics.overLimit show incorrect values ? Maybe we put the metrics only in Add() ?

src/x/cost/types.go
```diff
@@ -0,0 +1,62 @@
...
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package cost provides facilities for estimating the cost of operations
+// and enforcing limits on them.
+package cost
```
nikunjgit: maybe comment about the flow here ? Eg: is limit manager global ? Is limit per query ? Where does the actual enforcement happen ?

src/x/cost/enforcer.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+// Report is a report on the cost limits of an enforcer.
+type Report struct {
+	Cost
+	Error error
+}
+
+// Enforcer instances enforce cost limits for operations.
+type Enforcer interface {
```
nikunjgit: this should also go in types.go and be tied into the documentation

src/x/cost/types.go
andrewmains12: Sounds good! I think the comment is actually relatively accurate; this package is mostly a genericish library of utilities for estimating cost; actual usage of it is in https://github.com/m3db/m3/pull/1207 (which has the per query/global logic).

src/x/cost/enforcer.go
andrewmains12: Sure I can move it. What do you mean tied into the documentation? `godocs` should already pick this up, no?

src/x/cost/enforcer.go
andrewmains12: Good catch--`State` definitely shouldn't bump any metrics here. I'll change it.

PR #1216, opened 2018-11-28
Support m3msg server with messageFn

cw9:
Currently, we encode the aggregated metric in msgpack format and msgpack decoding iterator can be reused per consumer/connection, so we have m3msg server implementation that takes in lambda to process consumer/connection.

In the future we plan to encode aggregated metrics in protoub format, which no longer reuses decoder per consumer/connection, so I implemented the m3msg server to accept message lambda which is easier for users who don't need the consumer/connection level control to use.

### DIFF
--- a/src/msg/consumer/options.go
+++ b/src/msg/consumer/options.go
@@ -21,6 +21,7 @@
 package consumer
 
 import (
+	"errors"
 	"time"
 
 	"github.com/m3db/m3/src/msg/protocol/proto"
@@ -142,6 +143,7 @@ func (opts *options) SetInstrumentOptions(value instrument.Options) Options {
 
 type serverOptions struct {
 	consumeFn ConsumeFn
+	messageFn MessageFn
 	sOpts     server.Options
 	cOpts     Options
 }
@@ -164,6 +166,16 @@ func (opts *serverOptions) SetConsumeFn(value ConsumeFn) ServerOptions {
 	return &o
 }
 
+func (opts *serverOptions) MessageFn() MessageFn {
+	return opts.messageFn
+}
+
+func (opts *serverOptions) SetMessageFn(value MessageFn) ServerOptions {
+	o := *opts
+	o.messageFn = value
+	return &o
+}
+
 func (opts *serverOptions) ServerOptions() server.Options {
 	return opts.sOpts
 }
@@ -182,4 +194,14 @@ func (opts *serverOptions) SetConsumerOptions(value Options) ServerOptions {
 	o := *opts
 	o.cOpts = value
 	return &o
+}
+
+func (opts *serverOptions) Validate() error {
+	if opts.consumeFn == nil && opts.messageFn == nil {
+		return errors.New("no consumeFn nor messageFn defined")
+	}
+	if opts.consumeFn != nil && opts.messageFn != nil {
+		return errors.New("both consumeFn and messageFn defined")
+	}
+	return nil
 }
--- a/src/msg/consumer/options.go
+++ b/src/msg/consumer/options.go
@@ -21,13 +21,11 @@
 package consumer
 
 import (
-	"errors"
 	"time"
 
 	"github.com/m3db/m3/src/msg/protocol/proto"
 	"github.com/m3db/m3x/instrument"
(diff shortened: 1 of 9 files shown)


andrewmains12:
What's the context for this?

src/msg/consumer/options.go
```diff
@@ -164,6 +166,16 @@ func (opts *serverOptions) SetConsumeFn(value ConsumeFn) ServerOptions {
 	return &o
 }
 
+func (opts *serverOptions) MessageFn() MessageFn {
+	return opts.messageFn
+}
+
+func (opts *serverOptions) SetMessageFn(value MessageFn) ServerOptions {
+	o := *opts
+	o.messageFn = value
```
andrewmains12: What's the distinction between `consumeFn` and `messageFn`? What's `messageFn` intended to do?

src/msg/consumer/types.go
```diff
@@ -118,6 +118,9 @@ type Options interface {
 // ConsumeFn processes the consumer.
 type ConsumeFn func(c Consumer)
 
+// MessageFn processes the message.
+type MessageFn func(m Message)
```
andrewmains12: Nit: doc string could maybe use a bit more detail--what sort of processing is it intended to do?

cw9:
> What's the context for this?

Added more description to the pr

src/msg/consumer/options.go
cw9: Both `consumeFn` and `messageFn` needs to be implemented by user, if the user wants to process the consumer then they can pass in a `consumeFn`, this is useful when users can reuse resource on consumer level. When users don't care about controlling the consumer, they could simply pass in a `messageFn` that processes the message.

src/msg/consumer/types.go
cw9: It's up to the users how they want to process the message, e.g: it could be indexing the message, or ingest the message.

src/msg/consumer/server.go
```diff
@@ -62,3 +69,43 @@ func (h *handler) Handle(conn net.Conn) {
 }
 
 func (h *handler) Close() {}
+
+type messageHandler struct {
```
jeromefroe: Would it be simpler to make this public so you don't have to change `serverOptions` and `NewServer`?

src/msg/consumer/options.go
andrewmains12: Ah ok cool, thanks for the explanation! Maybe add to the docs to explain when to use `consumerFn` vs `messageFn`? I think that would get at what I was asking.

src/msg/consumer/server.go
cw9: I could make this public, but I'd still keep the changes in `NewServer`, seems easier to use that way. Users just need to implement `ConsumeFn` or `MessageFn` without needing to know what kind of handler to create.

cw9: Plus the handler itself is meaningless without the context of `server.Server`

src/msg/consumer/options.go
cw9: Sounds good, added more comments

src/msg/consumer/server.go
jeromefroe: I see, I was just thinking the following might be simpler:

```go
func (h *MessageHandler) Consume(c consumer.Consumer) {
	var (
		msgErr error
		msg    Message
	)
	for {
		msg, msgErr = c.Message()
		if msgErr != nil {
			break
		}
		h.messageFn(msg)
	}
	if msgErr != nil && msgErr != io.EOF {
		h.opts.InstrumentOptions().Logger().WithFields(log.NewErrField(msgErr)).Errorf("could not read message from consumer")
	}
	c.Close()
}
```

Then to use it one could do:

```go
msgHandler := NewMessageHandler(msgFn, handlerOpts)
opts := NewServerOptions().SetConsumeFn(msgHandler.Consume)
server := NewServer(addr, opts)
```

and you wouldn't need to update any of the code any place else.

cw9: Yeah that's a good idea, I did some refactoring based on the feedback

PR #1218, opened 2018-11-29
Add validate debug Prom endpoint

benraskin92:


src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,141 @@
...
+
+	var promDebug prometheus.PromDebug
+	if err := json.Unmarshal(data, &promDebug); err != nil {
+		logger.Error("unable to unmarshal data", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
+		return
+	}
+
+	// fmt.Println(promDebug.Input)
+	s := debug.NewStorage(promDebug.Input)
```
nikunjgit: debug is not an informative package name. Maybe something like proxy ?

src/query/storage/debugger/storage.go
```diff
@@ -0,0 +1,137 @@
...
+	fetchResult, err := s.Fetch(ctx, query, options)
+	if err != nil {
+		return block.Result{}, err
+	}
+
+	fmt.Println("fetch blocks dps: ", fetchResult.SeriesList[0].Values().Datapoints())
+	return storage.FetchResultToBlockResult(fetchResult, query)
+}
+
+func DpToTS(promReadResp prometheus.PromResp, tagOptions models.TagOptions) ts.SeriesList {
```
nikunjgit: PromResultToSeriesList ?

src/query/storage/debugger/storage.go
```diff
@@ -0,0 +1,137 @@
...
+				},
+			},
+		)
+	}
+
+	return seriesList
+}
+
+func (s *debugStorage) Type() storage.Type {
+	s.RLock()
```
nikunjgit: don't need any locks here

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -239,3 +239,30 @@ func RenderTagCompletionResultsJSON(
...
+	Value     string
+}
+
+// todo(braskin): use `Datapoint` instead of interface{} in values
+type PromResp struct {
+	Status string `json:"status"`
+	Data   struct {
+		ResultType string `json:"resultType"`
+		Result     []struct {
+			Metric struct {
```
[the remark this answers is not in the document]
benraskin92: Yeah I will - gonna clean up the pr and add some tests

src/query/storage/debugger/storage_test.go
```diff
@@ -0,0 +1,58 @@
...
+		{1543434985.200, "12"},
+		{1543434995.200, "14"},
+	}
+
+	promResult.Data.Result = append(promResult.Data.Result,
+		struct {
+			Metric map[string]string `json:"metric"`
+			Values [][]interface{}   `json:"values"`
+		}{
+			Values: vals,
```
benraskin92: Add tags here and test below

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -361,3 +361,31 @@ func RenderSeriesMatchResultsJSON(
...
 }
+
+// nolint: structcheck, megacheck, unused
+type Datapoint struct {
+	Timestamp float64
+	Value     string
+}
+
+// nolint: structcheck, megacheck, unused
+type Datapoints []Datapoint
```
arnikola: Why nolint? If this is unused should delete it

src/query/api/v1/handler/prometheus/debugger/sample_data.json
```diff
@@ -0,0 +1,278 @@
+{
```
arnikola: Why not add in a sample with 100,000 datapoints to pad lines of code 😛  ?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,266 @@
...
+		scope:       scope,
+		readHandler: h,
+	}
+}
+
+func (h *PromDebugHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
+	logger := logging.WithContext(ctx)
+
+	data, err := ioutil.ReadAll(r.Body)
```
arnikola: Will need to close `r.Body`, `defer r.Body.Close()` before this line should do it

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	s, err := debug.NewStorage(promDebug.Input)
+	if err != nil {
+		logger.Error("unable to create storage", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
+		return
+	}
+
+	h.engine = executor.NewEngine(s, h.scope.SubScope("debug_engine"))
```
arnikola: Should this be creating a new engine each time you run?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,266 @@
...
+
+	mismatches, err := validate(tsListToMap(promResults), tsListToMap(results))
+	if err != nil {
+		logger.Error("error validating results", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
arnikola: Is this necessary in light of https://github.com/m3db/m3/pull/1217 ?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,266 @@
...
+
+type mismatch struct {
+	seriesName       string
+	promVal, m3Val   float64
+	promTime, m3Time time.Time
+}
+
+func validate(prom, m3 map[string]*ts.Series) ([][]mismatch, error) {
+	if len(prom) != len(m3) {
+		return [][]mismatch{}, errors.New("number of Prom series not equal to number of M3 series")
```
arnikola: You can return nil as first arg

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,266 @@
...
+	}
+
+	var mismatches [][]mismatch
```
arnikola: Initialize this with a length

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,266 @@
...
+
+	for id, promSeries := range prom {
+		var (
+			m3Series *ts.Series
+			exists   bool
+		)
+
+		if m3Series, exists = m3[id]; !exists {
+			return [][]mismatch{}, fmt.Errorf("series with id %s does not exist in M3 results", id)
```
arnikola: Should this error out or should it add this as a mismatched series?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,266 @@
...
+		}
+
+		promdps := promSeries.Values().Datapoints()
+		m3dps := m3Series.Values().Datapoints()
+
+		var mismatchList []mismatch
```
arnikola: initialize this with a length

src/query/storage/debugger/storage.go
```diff
@@ -0,0 +1,149 @@
...
+		return block.Result{}, err
+	}
+
+	return storage.FetchResultToBlockResult(fetchResult, query)
+}
+
+// PromResultToSeriesList converts a prom result to a series list
+func PromResultToSeriesList(promReadResp prometheus.PromResp, tagOptions models.TagOptions) ([]*ts.Series, error) {
+	results := promReadResp.Data.Result
+	// seriesList := make(map[string]*ts.Series, len(results))
```
arnikola: Delete this

src/query/storage/debugger/storage.go
```diff
@@ -0,0 +1,149 @@
...
+	seriesList := make([]*ts.Series, len(results))
+
+	for i, result := range results {
+		dps := make(ts.Datapoints, len(result.Values))
+		for i, dp := range result.Values {
+			dps[i].Timestamp = time.Unix(0, int64(dp[0].(float64))*int64(time.Second))
```
arnikola: Should probably check that len(dp) == 2

src/query/storage/debugger/storage.go
```diff
@@ -0,0 +1,149 @@
...
+
+			tagsIdx++
+		}
+
+		tags := models.Tags{
+			Opts: tagOptions,
+			Tags: tagSlice,
+		}
+
+		name, _ := tags.Name()
```
arnikola: Will need to check second param and error if it's not found

src/query/storage/debugger/storage.go
```diff
@@ -0,0 +1,149 @@
...
+			if err != nil {
+				return nil, err
+			}
+			dps[i].Value = s
+		}
+
+		tagSlice := make([]models.Tag, len(result.Metric))
+		tagsIdx := 0
+		for name, val := range result.Metric {
+			tagSlice[tagsIdx] = models.Tag{
```
arnikola: Can just have `tagSlice = append(tagSlice...` rather than keeping the `tagsIdx` variable

src/query/storage/debugger/storage.go
```diff
@@ -0,0 +1,149 @@
...
+				Name:  []byte(name),
+				Value: []byte(val),
+			}
+
+			tagsIdx++
+		}
+
+		tags := models.Tags{
```
arnikola: Use `models.NewTags` and `tags.AddTags` instead

src/query/api/v1/handler/prometheus/debugger/validate_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	req, _ := http.NewRequest("POST", PromDebugURL+"?start=1543431465&end=1543435005&step=14&query=go_gc_duration_seconds", newBody())
+	recorder := httptest.NewRecorder()
+	debugHandler.ServeHTTP(recorder, req)
+
+	var mismatches MismatchesJSON
+	require.NoError(t, json.Unmarshal(recorder.Body.Bytes(), &mismatches))
+
+	assert.Len(t, mismatches.MismatchesList, 1)
+	assert.Len(t, mismatches.MismatchesList[0].Mismatches, 1)
```
arnikola: Pull `mismatches.MismatchesList[0].Mismatches` into a variable

src/query/api/v1/handler/prometheus/debugger/sample_data.json
benraskin92: Don't worry, I'm deleting this file :)

src/query/api/v1/handler/prometheus/debugger/validate.go
benraskin92: Yeah I discussed this with @nikunjgit and it's fine since it's a debug endpoint

src/query/api/v1/handler/prometheus/debugger/validate.go
benraskin92: I think error out since we return in the error the id of the series that's missing

src/query/api/v1/handler/prometheus/debugger/validate.go
benraskin92: We don't know the length upfront though so that wouldn't really save us anything, right? For example, every series could have a mismatch or none can.

src/query/api/v1/handler/prometheus/debugger/validate.go
benraskin92: Same comment as above, we don't know the length upfront.

src/query/storage/debugger/storage.go
benraskin92: Isn't indexing in more optimal?

benraskin92: Nevermind, didn't see `models.NewTags()`

src/query/api/v1/handler/prometheus/debugger/sample_data.json
arnikola: Make sure to make it a separate commit to really get those numbers up ;)

src/query/api/v1/handler/prometheus/debugger/validate.go
arnikola: When we don't know lengths beforehand we usually unit to a default value like 10 or something so it doesn't do a ton of copies at low numbers

src/query/api/v1/handler/prometheus/debugger/sample_data.json
benraskin92: Of course

src/query/api/v1/handler/prometheus/debugger/validate_test.go
```diff
@@ -149,7 +149,10 @@ func TestValidateEndpoint(t *testing.T) {
 	require.NoError(t, json.Unmarshal(recorder.Body.Bytes(), &mismatches))
 
 	assert.Len(t, mismatches.MismatchesList, 1)
-	assert.Len(t, mismatches.MismatchesList[0].Mismatches, 1)
-	assert.Equal(t, "__name__=go_gc_duration_seconds,instance=localhost:9090,job=prometheus,quantile=1,", mismatches.MismatchesList[0].Mismatches[0].Name)
-	assert.Equal(t, 0.012203, mismatches.MismatchesList[0].Mismatches[0].M3Val)
+
```
arnikola: nit: a little less whitespace? Typically we only add it after closing brackets

src/query/storage/debugger/storage.go
```diff
@@ -85,32 +85,32 @@ func PromResultToSeriesList(promReadResp prometheus.PromResp, tagOptions models.
 	for i, result := range results {
 		dps := make(ts.Datapoints, len(result.Values))
 		for i, dp := range result.Values {
+			if len(dp) != 2 {
+				continue
```
arnikola: This should probably error rather than just skipping it

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -84,13 +85,26 @@ func NewPromReadHandler(
...
 
 func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	result, params, err := h.ServeHTTPWithEngine(w, r, h.engine)
+	if err != nil {
+		return
+	}
+
+	// TODO: Support multiple result types
+	w.Header().Set("Content-Type", "application/json")
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
arnikola: Don't think this is necessary

src/query/storage/debugger/storage.go
```diff
@@ -0,0 +1,153 @@
...
+
+	for i, result := range results {
+		dps := make(ts.Datapoints, len(result.Values))
+		for i, dp := range result.Values {
+			if len(dp) != 2 {
+				return nil, errors.New("misformatted datapoint found")
+			}
+
+			dps[i].Timestamp = time.Unix(0, int64(dp[0].(float64))*int64(time.Second))
+			s, err := strconv.ParseFloat(dp[1].(string), 64)
```
arnikola: Might want to check type assertations work out here and above

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+				}
+
+				if !math.IsNaN(m3dps[m3idx].Value) {
+					break
+				}
+
+				m3idx++
+			}
+
+			if m3idx > len(m3dps)-1 {
```
arnikola: Can you refactor the loop a bit to avoid having to do this here?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+
+		if len(mismatchList) > 0 {
+			mismatches = append(mismatches, mismatchList)
+		}
+	}
+
+	return mismatches, nil
+}
+
+func createMismatch(name string, promVal, m3Val float64, promTime, m3Time time.Time) mismatch {
```
arnikola: This function doesn't really do anything, can get rid of it?

src/query/storage/debugger/storage.go
andrewmains12: Maybe better just to resolve the TODO and use the datapoint type there? Doesn't seem necessary to have `Values` as `[][]interface{}`

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -361,3 +361,22 @@ func RenderSeriesMatchResultsJSON(
 
 	return jw.Close()
 }
+
```
andrewmains12: Add some docs (either to the README, or linked from there) including sample usage?

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -239,3 +239,30 @@ func RenderTagCompletionResultsJSON(
...
+	Data   struct {
+		ResultType string `json:"resultType"`
+		Result     []struct {
+			Metric struct {
+				Name     string `json:"__name__"`
+				Instance string `json:"instance"`
+				Job      string `json:"job"`
+				Quantile string `json:"quantile"`
+			} `json:"metric"`
+			Values [][]interface{} `json:"values"`
```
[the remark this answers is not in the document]
andrewmains12: +1

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+	// PromDebugURL is the url for the Prom and m3query debugging tool
+	PromDebugURL = handler.RoutePrefixV1 + "/debug/validate_query"
+
+	// PromDebugHTTPMethod is the HTTP method used with this resource.
+	PromDebugHTTPMethod = http.MethodPost
+
+	mismatchCapacity = 10
+)
+
+// PromDebugHandler represents a handler for prometheus debug endpoint.
```
andrewmains12: Comment could be more informative--what does the prom debug endpoint do?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+}
+
+func (h *PromDebugHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
+	logger := logging.WithContext(ctx)
+
+	defer r.Body.Close()
+
+	data, err := ioutil.ReadAll(r.Body)
```
andrewmains12: Nit: you can read the stream directly like so: https://play.golang.org/p/IRWuwl2uY0E . Saves some err checking :)

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+	if err != nil {
+		logger.Error("unable to create storage", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
+		return
+	}
+
+	h.engine = executor.NewEngine(s, h.scope.SubScope("debug_engine"))
+	results, _, err := h.readHandler.ServeHTTPWithEngine(w, r, h.engine)
+	if err != nil {
+		logger.Error("unable to read data", zap.Error(err))
```
andrewmains12: Why is this a bad request? Seems like it might be a 500 instead?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+		logger.Error("error validating results", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+
+	if err := renderDebugMismatchResultsJSON(w, mismatches); err != nil {
+		logger.Error("unable to write back mismatch data", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
```
andrewmains12: Same--this is definitely a server error, and not bad request.

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+
+func validate(prom, m3 map[string]*ts.Series) ([][]mismatch, error) {
+	if len(prom) != len(m3) {
+		return nil, errors.New("number of Prom series not equal to number of M3 series")
+	}
+
+	mismatches := make([][]mismatch, 0, mismatchCapacity)
+
+	for id, promSeries := range prom {
+		var (
```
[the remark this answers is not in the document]
andrewmains12: +1 -- would let you get rid of the separate declaration.

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+			return nil, fmt.Errorf("series with id %s does not exist in M3 results", id)
+		}
+
+		promdps := promSeries.Values().Datapoints()
+		m3dps := m3Series.Values().Datapoints()
+
+		mismatchList := make([]mismatch, 0, mismatchCapacity)
+
+		m3idx := 0
+		m3dp := m3dps[m3idx]
```
andrewmains12: Check that the lengths match?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+
+		for _, promdp := range promdps {
+			if math.IsNaN(promdp.Value) && !math.IsNaN(m3dp.Value) {
+				mismatchList = append(mismatchList, createMismatch(id, promdp.Value, m3dp.Value, promdp.Timestamp, m3dp.Timestamp))
+			}
+
+			for {
```
andrewmains12: Nit: I think this would be equivalently readable as a `for` condition, maybe with a comment

```
// advance past NaN's
for ; m3idx < len(m3dps) && math.IsNaN(m3dps[m3idx].Value); m3idx++ {}
```

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+func renderDebugMismatchResultsJSON(
+	w io.Writer,
+	mismatches [][]mismatch,
+) error {
+	jw := qjson.NewWriter(w)
+	jw.BeginObject()
+
+	jw.BeginObjectField("mismatches_list")
```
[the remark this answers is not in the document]
andrewmains12: +1 -- I think it'd be worth wrapping the response in a top-level object at the least, so we can extend it later:

```
{
  "correct": true,
  "mismatches": ...
}
```

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+			if math.IsNaN(promdp.Value) && !math.IsNaN(m3dp.Value) {
+				mismatchList = append(mismatchList, createMismatch(id, promdp.Value, m3dp.Value, promdp.Timestamp, m3dp.Timestamp))
+			}
+
+			for {
+				if m3idx > len(m3dps)-1 {
+					break
+				}
+
+				if !math.IsNaN(m3dps[m3idx].Value) {
```
andrewmains12: What's the reasoning behind advancing past NaN's in the M3 response? Shouldn't we be checking that both sides of the comparison have NaN's in that case?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+
+			m3dp = m3dps[m3idx]
+			if (promdp.Value != m3dp.Value && !math.IsNaN(promdp.Value)) || !promdp.Timestamp.Equal(m3dp.Timestamp) {
+				mismatchList = append(mismatchList, createMismatch(id, promdp.Value, m3dp.Value, promdp.Timestamp, m3dp.Timestamp))
+			}
+
+			m3idx++
+		}
+
+		if len(mismatchList) > 0 {
```
andrewmains12: Seems like mismatchList should also be keyed by series name (same as the input)? That is, should it be a map instead?

src/query/api/v1/handler/prometheus/native/read.go
andrewmains12: Correct; `CORS` now gets set by middleware. Content-Type is still necessary though (we should probably factor that out too).

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -84,13 +85,26 @@ func NewPromReadHandler(
...
+// ServeHTTPWithEngine returns query results from the storage
+func (h *PromReadHandler) ServeHTTPWithEngine(w http.ResponseWriter, r *http.Request, engine *executor.Engine) ([]*ts.Series, models.RequestParams, error) {
 	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
 	logger := logging.WithContext(ctx)
 
 	params, rErr := parseParams(r)
 	if rErr != nil {
 		xhttp.Error(w, rErr.Inner(), rErr.Code())
-		return
+		return nil, emptyReqParams, rErr
```
andrewmains12: Will we end up calling `xhttp.Error` twice here? (looks like we call it again in the caller). Maybe we can just render results + errors in the caller. That would also eliminate the weird dichotomy between writing errors here but returning results.

src/query/storage/debugger/storage.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/query/ts"
+)
+
+type debugStorage struct {
+	seriesList []*ts.Series
+}
+
+// NewStorage creates a new debug storage instance.
```
andrewmains12: Comment (or the type comment) could use more detail; this is kind of a "X is X" comment.

src/query/api/v1/handler/prometheus/common.go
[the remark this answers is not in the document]
benraskin92: I'll do this in a follow up pr - it's a bit more involved unfortunately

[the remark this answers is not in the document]
andrewmains12: Sure fair enough. Involved how so? Are there actual different types that get stuffed in there?

src/query/api/v1/handler/prometheus/debugger/validate.go
arnikola: M3query populates NaNs at steps between datapoints, which prom does not, it just has the next point

andrewmains12: Cool gtk. @benraskin92 maybe worth adding that as a comment here?

src/query/api/v1/handler/prometheus/debugger/validate.go
```diff
@@ -0,0 +1,269 @@
...
+
+	mismatches, err := validate(tsListToMap(promResults), tsListToMap(results))
+	if err != nil {
+		logger.Error("error validating results", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
+		return
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+
```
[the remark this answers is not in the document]
benraskin92: Not sure. @nikunjgit you know?

src/query/storage/debugger/storage.go
benraskin92: I'll do this in a follow up pr

src/query/api/v1/handler/prometheus/debugger/validate.go
benraskin92: I like having this, makes it easier.

andrewmains12: If you're keeping it, rename to `newMismatch`?

arnikola: Doesn't really change much, and end up copying a bunch of variables without much upside

src/query/api/v1/handler/prometheus/debugger/validate.go
benraskin92: Most of the time Prom doesn't have NaN's (they just dont have any datapoint) and when they do I have a check above

src/query/api/v1/handler/prometheus/common.go
benraskin92: I'll add docs in follow up pr (have that on my to-do)

src/query/api/v1/handler/prometheus/debugger/validate.go
andrewmains12: 👍 makes sense. Add a brief comment?

src/query/api/v1/handler/prometheus/debugger/validate.go
benraskin92: promdps and m3dps?

src/query/api/v1/handler/prometheus/debugger/validate.go
benraskin92: I like it

src/query/api/v1/handler/prometheus/debugger/validate.go
andrewmains12: Yeah, though I suppose the lengths wouldn't match in the presence of NaN's. If prom never has NaN's, maybe the better comparison would be to filter out NaN's from the M3 response, and then compare the two lists. Would that work?

src/query/api/v1/handler/prometheus/native/read.go
benraskin92: Hmm not quite sure I understand, you mean move the `xhttp.Error` code to `ServeHTTP`?

src/query/api/v1/handler/prometheus/common.go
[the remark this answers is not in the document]
benraskin92: Well I think we'd need to write a custom unmarshaller, which isn't the end of the world, but id rather get this out so at least internally we can start debugging stuff then work on performance (not that this tool really needs to be that performant...)

src/query/api/v1/handler/prometheus/debugger/validate.go
benraskin92: Yeah but then at that point, we might as well just do the validation?

PR #1219, opened 2018-11-29
Add protobuf based conding for aggregated metrics

cw9:
Added protobuf based encoder and decoder for aggregated metrics


src/metrics/encoding/protobuf/aggregated_decoder.go
```diff
@@ -0,0 +1,93 @@
...
+
+package protobuf
+
+import (
+	"github.com/m3db/m3/src/metrics/generated/proto/metricpb"
+	"github.com/m3db/m3/src/metrics/policy"
+)
+
+// AggregatedDecoder is a decoder for decoding aggregated metrics.
+type AggregatedDecoder interface {
```
andrewmains12: Do we need the interface? That is, will we ever have other implementations of this? Also, what's the advantage of this wrapper instead of just using the protobuf object directly?

cw9: yeah, in this case since the implementation is so straight forward and I don't think I'll ever need to mock the interface, I'm going to just use struct instead. 

Using this wrapper around protobuf allows us to provide a `Close()` function to handle pooling operations, it also converts the protobuf types to object types.

PR #1222, opened 2018-11-30
[query] Use encoded series iterator blocks from fetch requests

arnikola:
Adds a new block structure that wraps a list of seriesIterators, satisfying both consolidated and unconsolidated block interfaces.


src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (s *encodedUnconsolidatedStep) Time() time.Time          { return s.time }
+func (s *encodedUnconsolidatedStep) Values() []xts.Datapoints { return s.values }
+
+func (it *encodedStepIterUnconsolidated) CurrentUnconsolidated() (
+	block.UnconsolidatedStep,
+	error,
+) {
+	it.mu.RLock()
```
andrewmains12: Does this actually need to be thread safe? Iterators often times are not (with no ill effects given proper usage). Are the other iterators thread safe?

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -0,0 +1,143 @@
...
+func (b *encodedBlock) stepIterUnconsolidated() block.UnconsolidatedStepIter {
+	return &encodedStepIterUnconsolidated{
+		meta:        b.meta,
+		seriesMeta:  b.seriesMetas,
+		seriesIters: b.seriesBlockIterators,
+		lastBlock:   b.lastBlock,
+	}
+}
+
+type encodedUnconsolidatedStep struct {
```
andrewmains12: Do we need a new implementation of this? Could we not just share the other implementation (which is identical afaict)?

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -0,0 +1,143 @@
...
+	if !anyNext {
+		it.exhausted = true
+	}
+
+	return anyNext
+
+}
+
+func (it *encodedStepIterUnconsolidated) StepCount() int {
+	panic("cannot request step count from unconsolidated step iterator")
```
andrewmains12: Seems dangerous, no? What guarantees do we have that this won't get called on this type?

src/query/ts/m3db/encoded_unconsolidated_series_iterator.go
```diff
@@ -0,0 +1,110 @@
...
+	// }
+
+	// return block.NewUnconsolidatedSeries(values, block.SeriesMeta{
+	// 	Tags: s.Tags,
+	// 	Name: s.Name(),
+	// }), nil
+
+	series := block.NewUnconsolidatedSeries(values, it.seriesMeta[it.idx])
+	it.mu.RUnlock()
+	return series, nil
```
andrewmains12: Should we be checking `iter.Err()` somewhere in here?

src/query/storage/m3/storage.go
```diff
@@ -88,11 +88,47 @@ func (s *m3storage) Fetch(
...
+		Blocks: blocks,
+	}, nil
+}
+
 func (s *m3storage) FetchBlocks(
 	ctx context.Context,
 	query *storage.FetchQuery,
 	options *storage.FetchOptions,
 ) (block.Result, error) {
+	if !options.UseDecodedBlocks {
```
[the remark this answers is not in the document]
arnikola: Sure, was just a bit reticent about bigbang swapping to this method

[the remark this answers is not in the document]
nikunjgit: it's not really in use right now so good time to make those big changes

src/query/test/test_series_iterator.go
```diff
@@ -180,3 +182,84 @@ func BuildTestSeriesIterator() (encoding.SeriesIterator, error) {
...
+}
+
+// BuildCustomIterator builds a custom iterator with bounds
+func BuildCustomIterator(
+	dps [][]Datapoint,
+	testTags map[string]string,
+	seriesID, seriesNamespace string,
+	start time.Time,
+	blockSize, stepSize time.Duration,
+) (encoding.SeriesIterator, *models.Bounds, error) {
```
nikunjgit: why return pointer to Bounds ?

nikunjgit: oops its a test, nvm

src/query/ts/m3db/consolidator.go
```diff
@@ -0,0 +1,110 @@
...
+
+func removeStale(
+	earliestLookback time.Time,
+	dps []ts.Datapoint,
+) []ts.Datapoint {
+	var (
+		i  int
+		dp ts.Datapoint
+	)
+	for _, dp = range dps {
```
nikunjgit: Can be made simpler

for i, dp := range dps {

if !...{ return dps[i:] }

)

src/query/ts/m3db/consolidator.go
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		i++
+	}
+
+	return dps[i:]
+}
+
+type lookbackConsolidator struct {
+	lookback         time.Duration
```
nikunjgit: what's lookback here ?

src/query/ts/m3db/consolidator.go
```diff
@@ -0,0 +1,110 @@
...
+	stepSize         time.Duration
+	earliestLookback time.Time
+	consolidated     []float64
```
nikunjgit: any reason why it should store consolidated ? Make pooling harder in the future

src/query/ts/m3db/consolidator_single.go
```diff
@@ -0,0 +1,90 @@
...
+package m3db
+
+import (
+	"math"
+	"time"
+
+	"github.com/m3db/m3/src/dbnode/ts"
+)
+
+type singleLookbackConsolidator struct {
```
nikunjgit: can't consolidator handle single lookback as well ?

nikunjgit: also, need some tests for it if we do decide to keep it

src/query/ts/m3db/convert.go
```diff
@@ -54,31 +57,56 @@ func (b blockReplicas) Less(i, j int) bool {
 	return b[i].start.Before(b[j].start)
 }
 
-// ConvertM3DBSeriesIterators converts m3db SeriesIterators to SeriesBlocks
-// which are used to construct Blocks for query processing.
-func ConvertM3DBSeriesIterators(iterators encoding.SeriesIterators, iterAlloc encoding.ReaderIteratorAllocate) ([]SeriesBlocks, error) {
-	defer iterators.Close()
-	multiSeriesBlocks := make([]SeriesBlocks, iterators.Len())
+// ConvertM3DBSeriesIterators converts series iterators to iterator blocks
+func ConvertM3DBSeriesIterators(
```
nikunjgit: when would you have multiple blocks from series iterators ?

src/query/ts/m3db/convert.go
```diff
@@ -54,31 +57,56 @@ func (b blockReplicas) Less(i, j int) bool {
...
-		blockReplicas, err := blockReplicasFromSeriesIterator(seriesIterator, iterAlloc)
+// ConvertM3DBSegmentedBlockIterators converts series iterators to a list of blocks
+func ConvertM3DBSegmentedBlockIterators(
+	iterators encoding.SeriesIterators,
+	iterAlloc encoding.ReaderIteratorAllocate,
+	tagOptions models.TagOptions,
+	bounds *models.Bounds,
+	pools encoding.IteratorPools,
+) ([]block.MultipurposeBlock, error) {
+	return nil, fmt.Errorf("WIP: splitting blocks will require lookback propagation")
```
nikunjgit: this is a bit confusing. Is the code below not needed at all ?

src/query/ts/m3db/convert.go
```diff
@@ -92,22 +120,23 @@ func blockReplicasFromSeriesIterator(seriesIterator encoding.SeriesIterator, ite
...
 			}
-			// todo(braskin): pooling
+
 			iter := encoding.NewMultiReaderIterator(iterAlloc, nil)
 			iter.Reset(readers, start, bs)
-
 			inserted := false
-			for i := range blockReplicas {
-				if blockReplicas[i].start.Equal(start) {
+			for i := range replicas {
```
nikunjgit: replicas is confusing since that has replicas too ? Also, why not just _, replica := range replicas

src/query/ts/m3db/convert.go
```diff
@@ -21,11 +21,14 @@
 package m3db
```
nikunjgit: can we have a high level doc of how our conversion is working ? We can put that in the user docs since that would help other storages to integrate with us if needed

src/query/ts/m3db/convert.go
```diff
@@ -166,14 +192,20 @@ func seriesBlocksFromBlockReplicas(blockReplicas []blockReplica, seriesIterator
...
 
-		series.Blocks[i] = SeriesBlock{
-			start:          filterValuesStart,
-			end:            filterValuesEnd,
-			seriesIterator: valuesIter,
-		}
+		// NB(braskin): we should be careful when directly accessing the series iterators.
+		// Instead, we should access them through the SeriesBlock.
+		blockBuilder.add(
+			&models.Bounds{
```
nikunjgit: why a pointer to bounds ?

src/query/ts/m3db/convert.go
```diff
@@ -117,12 +146,17 @@ func blockReplicasFromSeriesIterator(seriesIterator encoding.SeriesIterator, ite
...
-	sort.Sort(blockReplicas)
-
-	return blockReplicas, nil
+	sort.Sort(replicas)
+	return replicas, nil
 }
 
-func seriesBlocksFromBlockReplicas(blockReplicas []blockReplica, seriesIterator encoding.SeriesIterator) (SeriesBlocks, error) {
+func seriesBlocksFromBlockReplicas(
+	blockBuilder *encodedBlockBuilder,
```
nikunjgit: why pointer to builder ?

src/query/ts/m3db/convert.go
```diff
@@ -166,14 +192,20 @@ func seriesBlocksFromBlockReplicas(blockReplicas []blockReplica, seriesIterator
...
+		// NB(braskin): we should be careful when directly accessing the series iterators.
+		// Instead, we should access them through the SeriesBlock.
+		blockBuilder.add(
+			&models.Bounds{
+				Start:    filterValuesStart,
+				Duration: block.blockSize,
+				StepSize: stepSize,
+			},
+			iter,
+			i == replicaLength,
```
nikunjgit: move this out into a variable for better readability

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,206 @@
...
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/functions/utils"
+	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/storage"
+)
+
+type consolidationSettings struct {
+	consolidationFn ConsolidationFunc
+	currentTime     time.Time
+	bounds          *models.Bounds
```
nikunjgit: why pointer for bounds ?

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+type encodedBlock struct {
+	lastBlock            bool
+	meta                 block.Metadata
+	tagOptions           models.TagOptions
+	consolidation        *consolidationSettings
```
nikunjgit: why pointer here ?

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,206 @@
...
+)
+
+type consolidationSettings struct {
+	consolidationFn ConsolidationFunc
+	currentTime     time.Time
+	bounds          *models.Bounds
+}
+
+type encodedBlock struct {
+	lastBlock            bool
```
nikunjgit: this is weird

src/query/ts/m3db/consolidator.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	c.datapoints[i] = append(c.datapoints[i], dp)
+}
+
+func (c *lookbackConsolidator) consolidate() []float64 {
+	c.earliestLookback = c.earliestLookback.Add(c.stepSize)
+	for i, dps := range c.datapoints {
+		c.consolidated[i] = c.fn(dps)
+		c.datapoints[i] = removeStale(c.earliestLookback, dps)
```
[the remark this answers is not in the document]
arnikola: It does; the next step is added prior to removing stale, so that next iteration will have all the stale points removed.

src/query/storage/types.go
```diff
@@ -76,9 +76,12 @@ func (q *FetchQuery) String() string {
 	return q.Raw
 }
 
-// FetchOptions represents the options for fetch query
+// FetchOptions represents the options for fetch query.
 type FetchOptions struct {
+	// Limit is the maximum number of series to return.
 	Limit int
+	// UseDecodedBlocks determines if decoded blocks should be used.
+	UseDecodedBlocks bool
```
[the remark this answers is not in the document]
arnikola: Sure

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,206 @@
...
+	meta                 block.Metadata
+	tagOptions           models.TagOptions
+	consolidation        *consolidationSettings
+	seriesMetas          []block.SeriesMeta
+	seriesBlockIterators []encoding.SeriesIterator
+}
+
+// verify encodedSeriesIter satisfies both block and unconsolidated block
+var (
+	_ block.Block               = (*encodedBlock)(nil)
```
nikunjgit: rather have MakeEncodedBlock return block.Block. Also, not sure about it being both a block and unconsolidatedblock

src/query/block/types.go
```diff
@@ -96,8 +105,8 @@ type SeriesIter interface {
 type UnconsolidatedSeriesIter interface {
 	Iterator
 	SeriesMetaIter
-	// Current returns the current series for the block
-	Current() (UnconsolidatedSeries, error)
+	// CurrentUnconsolidated returns the current series for the block
+	CurrentUnconsolidated() (UnconsolidatedSeries, error)
```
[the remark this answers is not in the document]
arnikola: To have a single block that would satisfy both Consolidated and Unconsolidated block interfaces, otherwise it would have conflicts. Originally I had a single SeriesIterator and StepIterator that satisfied both interfaces, so had to make the change here, too, but since split it up, so can revert this

src/query/block/types.go
```diff
@@ -119,8 +128,8 @@ type StepIter interface {
 type UnconsolidatedStepIter interface {
 	Iterator
 	StepMetaIter
-	// Current returns the current step for the block
-	Current() (UnconsolidatedStep, error)
+	// CurrentUnconsolidated returns the current step for the block
+	CurrentUnconsolidated() (UnconsolidatedStep, error)
```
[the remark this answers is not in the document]
arnikola: To have a single block that would satisfy both Consolidated and Unconsolidated block interfaces, otherwise it would have conflicts. Originally I had a single SeriesIterator and StepIterator that satisfied both interfaces, so had to make the change here, too, but since split it up, so can revert this

src/query/block/types.go
```diff
@@ -39,22 +47,23 @@ type Block interface {
...
 type UnconsolidatedBlock interface {
-	// StepIter returns a StepIterator
-	StepIter() (UnconsolidatedStepIter, error)
-	// SeriesIter returns a SeriesIterator
-	SeriesIter() (UnconsolidatedSeriesIter, error)
+	io.Closer
+	unconsolidatedBlock
+}
+
+type unconsolidatedBlock interface {
```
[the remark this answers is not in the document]
arnikola: To be able to satisfy both interfaces

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,206 @@
...
+	generateMetas bool,
+) encodedBlock {
+	b := encodedBlock{
+		seriesBlockIterators: seriesBlockIterators,
+		tagOptions:           tagOptions,
+		consolidation:        consolidation,
+		lastBlock:            lastBlock,
+	}
+
+	if generateMetas {
```
nikunjgit: why not generate this every time ?

src/query/block/types.go
```diff
@@ -22,15 +22,23 @@ package block
...
+	"io"
 	"math"
 	"time"
 
 	"github.com/m3db/m3/src/query/models"
 	"github.com/m3db/m3/src/query/ts"
 )
 
+// MultipurposeBlock is a block that satisfies consolidated and unconsolidated blocks
+type MultipurposeBlock interface {
```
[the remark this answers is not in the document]
arnikola: This one is both consolidated and unconsolidated

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,206 @@
...
+}
+
+func (b *encodedBlock) WithMetadata(
+	meta block.Metadata,
+	seriesMetas []block.SeriesMeta,
+) (block.Block, error) {
+	return b.withMetadata(meta, seriesMetas)
+}
+
+func (b *encodedBlock) WithMetadataUnconsolidated(
```
nikunjgit: these functions are confusing. What's metadata for unconsolidated vs consolidated block. How's the caller to know which one to use ? Shouldn't they always be the same

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: I think it's probably not necessary at the moment, but the next few commits for the block iterator are likely to require this to be threadsafe since there may be async communications between them

src/query/ts/m3db/consolidator.go
arnikola: Nice

src/query/ts/m3db/encoded_block_iterator.go
arnikola: This will be necessary for when series are split into blocks; the last block has a distinct that it excludes any values at EndTime, whereas other blocks include it.

src/query/ts/m3db/encoded_block_iterator.go
arnikola: It's for the builder case; when building up these blocks, there's no point generating metadatas until all blocks are present and the `build()` method is called.

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// verify encodedSeriesIter satisfies both block and unconsolidated block
+var (
+	_ block.Block               = (*encodedBlock)(nil)
+	_ block.UnconsolidatedBlock = (*encodedBlock)(nil)
+)
+
+// MakeEncodedBlock builds an encoded block
+func MakeEncodedBlock(
```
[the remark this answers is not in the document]
arnikola: Sounds good

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: My thinking was that this is an invalid method to call on an unconsolidated step iterator, as we have no way of knowing step count; however this actually gives the projected step count post consolidation, so it's doable

src/query/ts/m3db/consolidator.go
arnikola: This is the same lookback as `models.LookbackDelta` for prom consolidation

src/query/ts/m3db/consolidator.go
arnikola: It's more effective until pooling is added; when it is added, it will init on consolidator start, then release to the pool after the consolidator is cleaned up.

src/query/ts/m3db/consolidator_single.go
arnikola: It can, but gets annoying dealing with wrapping/unwrapping arrays unnecessarily. Will add tests

src/query/ts/m3db/convert.go
arnikola: When I enable splitting series iterators into block iterators, this will return multiple blocks.

src/query/ts/m3db/convert.go
arnikola: The below code is fine for lookback = 0s; however there will need to be changes to update it to be able to handle lookback times since blocks in a series will need to communicate with other blocks in that series to provide lookback values.

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
andrewmains12: So it definitely makes sense to me that we can't effectively answer this question for this class; we have to iterate through the block in order to know how many steps it contains. Given that, my real question is whether we actually need this method. I worry about a case where we pass through a `encodedStepIterUnconsolidated` as the interface, call the method and then get a panic. 

I think there are two potential safe approaches here, depending on what's needed:

1. Get rid of this method from the interface (if that's possible).
2. If it's not, use upcasting to check when it's safe to use. That is,

```
type StepCounted interface {
    StepCount() int
}

// when using it
func Process(b block.Block) {

var stepCount int = someDefault
if sc, ok := b.(StepCounted); ok {
   stepCount = sc.StepCount()
}
}
```
Does that make sense?

src/query/ts/m3db/convert.go
arnikola: Done

src/query/block/types.go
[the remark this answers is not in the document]
arnikola: Spoke offline, split.

src/query/block/types.go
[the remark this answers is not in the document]
arnikola: Spoke offline, split into two block types.

src/query/storage/m3/storage.go
```diff
@@ -88,11 +88,47 @@ func (s *m3storage) Fetch(
...
+		StepSize: query.Interval,
+	}
+
+	multiPurposeBlocks := m3db.ConvertM3DBSeriesIterators(
+		raw,
+		s.tagOptions,
+		bounds,
+	)
+
+	blocks := make([]block.Block, len(multiPurposeBlocks))
```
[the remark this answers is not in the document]
arnikola: Refactored

src/query/ts/m3db/convert.go
arnikola: Will put one together. It'll be a bit more involved since it will cover how split blocks will work so may take a bit longer to put together unfortunately

src/query/ts/m3db/encoded_block_iterator.go
arnikola: Refactored to do it this way

src/query/ts/m3db/encoded_block_iterator.go
arnikola: Changed all of these

src/query/ts/m3db/encoded_block_iterator.go
arnikola: Refactored a bit.

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: Refactored to use existing

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: Sorry, my response may have been a bit inaccurate.

While we can't get the actual step count for unconsolidated series, `UnconsolidatedStepIterator.StepCount()` should return the expected step count for that step iterator, post-consolidation. As such it returns `bounds.StepSize()` to mirror how the other unconsolidated step iterator performs.

arnikola: If anything, might be a bit of a naming issue on that func

src/query/ts/m3db/encoded_unconsolidated_series_iterator.go
arnikola: Addedchecks where appropriate

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
andrewmains12: Ok cool, so we can get rid of the panic then? That works for me.

src/query/ts/m3db/encoded_block_iterator.go
andrewmains12: Longer discussion, but are you guys familiar with the "accept interfaces return structs" dictum? Not the pattern in this repo, but it is elsewhere within Uber. https://medium.com/@cep21/what-accept-interfaces-return-structs-means-in-go-2fe879e25ee8

andrewmains12: Also mentioned in [the go libraries talk](https://docs.google.com/presentation/d/1phfzoGQqEpSQVXRiy0Zpq5v2USy9bc6Bc2XRrV0R9sA/edit)

src/query/storage/block.go
```diff
@@ -115,7 +115,10 @@ func (m *multiBlockWrapper) WithMetadata(
 	}
 
 	if m.unconsolidated != nil {
-		unconsolidated, err = m.unconsolidated.WithMetadata(meta, seriesMetas)
+		unconsolidated, err = m.unconsolidated.WithMetadata(
```
nikunjgit: what's your line limit set ? I think two parameters on one line are fine, splitting everything out hurts readability a bit

src/query/ts/m3db/consolidator.go
nikunjgit: Isn't that the earliestLookback ? Maybe change the names a bit to make it clear ?

src/query/ts/m3db/consolidator.go
nikunjgit: maybe i'm missing something. Isn't consolidated used only in consolidate() ? Then how is it effective to store it on the struct

src/query/ts/m3db/consolidator.go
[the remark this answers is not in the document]
nikunjgit: i meant c.fn(dps) consolidated over stale datapoints ?

src/query/ts/m3db/consolidator_single.go
nikunjgit: lot of code duplication :( you're just using singleConsolidator at one place so i'll suggest merging.

src/query/ts/m3db/convert.go
nikunjgit: should we be coding before figuring out how to do it ? :P It was just a suggestion to reduce back n forth and avoiding having you to redo lot of things during reviews but i don't mind either way.

src/query/ts/m3db/convert.go
nikunjgit: there's no way to test the code below :( Maybe move this out into a different function and have a test ?

src/query/ts/m3db/encoded_block_iterator.go
nikunjgit: then remove from here and call generateMetas from build() ?

src/query/ts/m3db/convert.go
nikunjgit: maybe add that TODO?

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (s *encodedStep) Time() time.Time   { return s.time }
+func (s *encodedStep) Values() []float64 { return s.values }
+
+func (it *encodedStepIter) Current() (block.Step, error) {
+	it.mu.RLock()
+	currentTime := it.currentTime
+	if !it.started || currentTime.After(it.bounds.End()) {
+		it.mu.RUnlock()
+		panic("out of bounds")
```
nikunjgit: return error ?

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,256 @@
...
+type peekValue struct {
+	started  bool
+	finished bool
+	point    ts.Datapoint
+}
+
+type encodedStepIter struct {
+	mu           sync.RWMutex
+	lastBlock    bool
+	exhausted    bool
```
nikunjgit: don't think you need this. Next can just use the bounds to figure out if its exhausted. Also, if there are no values left in the m3db iterator, it should still return NaNs.

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/m3db/m3/src/dbnode/encoding"
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/storage"
+	xts "github.com/m3db/m3/src/query/ts"
+)
+
+type encodedStepIterUnconsolidated struct {
+	mu          sync.RWMutex
+	lastBlock   bool
+	exhausted   bool
```
nikunjgit: same comment regarding exhausted as above

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,256 @@
...
+	started      bool
+	currentTime  time.Time
+	err          error
```
nikunjgit: why store this ?

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,256 @@
...
+					point: ts.Datapoint{
+						Timestamp: dp.Timestamp,
+						Value:     dp.Value,
+					},
+					started: true,
+				}
+			}
+
+			if err := iter.Err(); err != nil {
+				it.err = err
```
nikunjgit: just have this return the error instead of storing

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	if !anyNext {
+		it.exhausted = true
+	}
+}
+
+// Need to run an initial step; if there are any values
+// that appear at exactly the start, they must be added.
+func (it *encodedStepIter) initialStep() {
```
nikunjgit: not super clear if this is needed as a special case, will go through it again

src/query/ts/m3db/encoded_block_iterator.go
nikunjgit: i'm not sure I follow. Shouldn't all blocks exclude the EndTime since endTime for block1 is same as start time for block2 so if you include endtime for block1, its double counted ?

src/query/ts/m3db/encoded_step_iterator.go
arnikola: I think iterators typically panic if you call `Current` while out of bounds, but can do either way

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: Initial approach used `exhausted` to shortcircuit checking iterators; can get rid of it.

src/query/ts/m3db/encoded_step_iterator.go
arnikola: It's a little weird for `Next()` to return an error; all other iterators we use store an error which is accessed as a separate function

src/query/ts/m3db/encoded_block_iterator.go
arnikola: Because of the way the internal blockIterators that make up a seriesIterator are structured, values at the block boundary can go in either block; there will only be 1 value but it may be in either one. Because of this, we always want to take values at the end in all blocks except the last, which would be the exclusive end of the entire query.

src/query/ts/m3db/convert.go
arnikola: It's not currently in use so no tests are necessary; this is the code required to transform a series iterator into series iterator blocks

src/query/ts/m3db/convert.go
arnikola: Haven't started the other section yet, sho shouldn't be a problem there :)

src/query/ts/m3db/consolidator.go
arnikola: lookback is the actual duration, `earliestLookback` is the cutoff time for incoming values

src/query/ts/m3db/encoded_block_iterator.go
arnikola: Refactored a bit

src/query/storage/block.go
arnikola: Really? Find this a bit easier to read personally

src/query/ts/m3db/consolidator.go
arnikola: Renamed to `lookbackDuration`

src/query/ts/m3db/consolidator_single.go
arnikola: I'm also only using `consolidator` in one place.

There's a couple of subtle differences; i.e. the single needs to reset whereas the other does not; would rather keep separate types here

src/query/ts/m3db/consolidator.go
arnikola: This way it lives for the lifetime of the consolidator; we're not making output arrays constantly by re-using the result array.

src/query/ts/m3db/consolidator.go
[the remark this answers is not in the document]
arnikola: No. They are not stale when c.fn runs, as it updates the earliest lookback, and thus the stale values, for next consolidation.

src/query/ts/m3db/encoded_step_iterator.go
arnikola: Currently it inits the seriesPeek here. This will become more of a special case on the next phase, so would prefer to keep it separate.

src/query/ts/m3db/encoded_step_iterator.go
arnikola: Next cannot error, so errors will need to be checked when requesting `Current()`

src/query/ts/m3db/encoded_step_iterator.go
arnikola: Removed it here.

src/query/ts/m3db/encoded_block_iterator.go
nikunjgit: hmm interesting, will take a look

src/query/ts/m3db/encoded_step_iterator.go
nikunjgit: Current() here already returns an error in the second arg, why do we have to force panic when we can avoid that

src/query/ts/m3db/consolidator.go
nikunjgit: ah makes sense. I didn't realize consolidate() is being called in Current(). Might be worth resetting c.consolidated within consolidate first to avoid extraneous values ?

src/query/ts/m3db/consolidator.go
```diff
@@ -0,0 +1,105 @@
...
+
+	c.datapoints[i] = append(c.datapoints[i], dp)
+}
+
+func (c *lookbackConsolidator) consolidate() []float64 {
+	// Update earliest lookback then remove stale values for the next
+	// iteration of the datapoint set.
+	c.earliestLookback = c.earliestLookback.Add(c.stepSize)
+	for i, dps := range c.datapoints {
+		c.consolidated[i] = c.fn(dps)
```
nikunjgit: reset consolidated  first ?

src/query/ts/m3db/consolidator.go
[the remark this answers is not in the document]
nikunjgit: If dps are not stale in line 100, how can they get stale in line 101 ?

src/query/ts/m3db/consolidator_single.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+func (c *singleLookbackConsolidator) consolidate() float64 {
+	c.earliestLookback = c.earliestLookback.Add(c.stepSize)
+	c.consolidated = c.fn(c.datapoints)
+	c.datapoints = removeStale(c.earliestLookback, c.datapoints)
+	return c.consolidated
+}
+
+func (c *singleLookbackConsolidator) empty() bool {
```
nikunjgit: don't really need this, consolidate() can just do empty check and return math.NaN()

src/query/ts/m3db/consolidator_single.go
nikunjgit: can you add some comments on the singleLookbackConsolidator struct and also on lockbackConsolidator struct ? That should make it easier for us to follow

src/query/ts/m3db/consolidator_single.go
```diff
@@ -0,0 +1,90 @@
...
+	return len(c.datapoints) == 0
+}
+
+func (c *singleLookbackConsolidator) reset(
```
nikunjgit: can you move such pooling outside the consolidator and keep it simple ? One way would be that the build single consolidator takes in the datapoints and is just called every time instead of reset() being called.

src/query/ts/m3db/convert.go
nikunjgit: let's just comment it out then ?

src/query/ts/m3db/encoded_block_iterator.go
nikunjgit: if you count that point within block1, what would be the timestamp associated with it ? If its timestamp is the end, then it might get excluded by downstream since they exclude bounds.End

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,228 @@
...
+		it.nextConsolidatedForStep(i)
+	}
+}
+
+// Need to run an initial step; if there are any values
+// that appear at exactly the start, they must be added.
+func (it *encodedStepIter) initialStep() {
+	it.seriesPeek = make([]peekValue, len(it.seriesIters))
+	for i, iter := range it.seriesIters {
+		if iter.Next() {
```
nikunjgit: flip to reduce nesting ?

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,228 @@
...
+				it.seriesPeek[i] = peekValue{
+					point: ts.Datapoint{
+						Timestamp: dp.Timestamp,
+						Value:     dp.Value,
+					},
+					started: true,
+				}
+			}
+
+			if err := iter.Err(); err != nil {
```
nikunjgit: should this be the first check, and skip doing other things if err ?

src/query/ts/m3db/encoded_step_iterator.go
nikunjgit: that iterator is interesting since it returns true and may have an error.

arnikola: SeriesIterator itself does the same. `Next()` should never error

src/query/ts/m3db/encoded_step_iterator.go
arnikola: `if !iter.Next()` is really weird semantics, not worth it for nesting.

src/query/ts/m3db/consolidator.go
arnikola: `c.fn(dps)` should set 'used up' values with whatever the default case for the consolidation function is

src/query/ts/m3db/consolidator.go
arnikola: I think it's better to let the consolidation function decide what its default values should be

src/query/ts/m3db/consolidator.go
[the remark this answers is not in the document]
arnikola: They're not stale for the current step, but are stale for the next step. This cleans them up for next time Current is called

src/query/ts/m3db/consolidator_single.go
arnikola: This allows us to shortcircuit lookups when there are no points

src/query/ts/m3db/encoded_block_iterator.go
arnikola: Downstream consumers are going to have to be changed when the multiblock read is enabled in either case, which is when this will be necessary

src/query/ts/m3db/encoded_step_iterator.go
arnikola: I can change this but it doesn't really matter so much; will be intending to refactor how these works to mirror SeriesIterator, where neither `Next` or `Current` can error, but there's an `Err` method to check if any errors have occured

src/query/ts/m3db/encoded_step_iterator.go
arnikola: It's a weird consequence of Current returning a value and an error that this happens like this. Once I refactor Current to return only a value, and add an Error method, we should be able to switch it out to be a bit more semantically correct. Since I want to refactor consumers already, I may roll that into the future work

src/query/ts/m3db/encoded_step_iterator.go
arnikola: Refactored this a bit to make it more correct

src/query/ts/m3db/convert.go
arnikola: It's the next part of this, doesn't really need a todo since it should be fixed soon

src/query/ts/m3db/consolidator_single.go
arnikola: Why though?

This consolidator is built to make consolidating a single series easier; reset just clears the values for the next series to be able to consolidate.

src/query/ts/m3db/convert.go
arnikola: It gets errored out already

src/query/ts/m3db/consolidator_single.go
arnikola: Done

src/query/storage/block.go
nikunjgit: okay, I can understand that you want to write code this way but can you avoid reformatting code you're not changing ?

src/query/ts/m3db/consolidator.go
nikunjgit: so maybe change c.consolidated -> c.pooledConsolidated ? Also, i meant reset it before the for loop so that we avoid the problem of sending extraneous values. Currently, if consolidated and datapoints length doesn't match up, it might take old values and get hard to debug

src/query/ts/m3db/consolidator.go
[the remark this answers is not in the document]
nikunjgit: let's move it out into another function then. Conslidate() shouldn't be responsible to doing this

src/query/ts/m3db/consolidator_single.go
nikunjgit: To separate responsibility ? I think it's fine for now.

src/query/ts/m3db/convert.go
nikunjgit: Some linters complain about unreachable code. Usually, it's not a great practice to keep unreachable code around, just in case you don't clean it up later. I'm not sure what's your hesitation here. I'm fine if you leave it as is for now.

src/query/ts/m3db/encoded_block_iterator.go
nikunjgit: Can you add some comments here ? Anyone else looking at the code may have similar questions.

src/query/ts/m3db/encoded_block_iterator_builder.go
```diff
@@ -0,0 +1,87 @@
...
+		)
+	}
+
+	block := b.blocksAtTime[str]
+	block.seriesBlockIterators = append(block.seriesBlockIterators, iter)
+	b.blocksAtTime[str] = block
+}
+
+func (b *encodedBlockBuilder) build() []block.Block {
+	blocks := make([]encodedBlock, 0, len(b.blocksAtTime))
```
nikunjgit: just make this block.Block to avoid the casting later ?

src/query/ts/m3db/encoded_series_iterator.go
```diff
@@ -0,0 +1,130 @@
...
+		consolidator: consolidator,
+	}
+}
+
+func (it *encodedSeriesIter) Current() (block.Series, error) {
+	it.mu.RLock()
+	iter := it.seriesIters[it.idx]
+	values := make([]float64, it.bounds.Steps())
+	xts.Memset(values, math.NaN())
+	i := 0
```
nikunjgit: you can avoid keeping this by values := make([]float64, 0, it.bounds.Steps()) and later just appending to values

src/query/ts/m3db/encoded_series_iterator.go
```diff
@@ -0,0 +1,130 @@
...
+
+			if !ts.After(currentTime) {
+				it.consolidator.addPoint(dp)
+				break
+			}
+		}
+	}
+
+	if err := iter.Err(); err != nil {
+		return block.Series{}, err
```
nikunjgit: Unlock ? Although you get a little perf bump avoiding defer but it's very easy to miss unlocking. I'm not sure if we should be going to these small optimizations at this point, but this is fine for now if you don't want to change it

src/query/storage/block.go
arnikola: This was changed earlier before the revert

src/query/ts/m3db/convert.go
arnikola: Not a great practice to commit big commented out blocks of code, either :p

src/query/ts/m3db/consolidator.go
arnikola: It's not pooled consolidated, it's just a slice of current consolidated values. Length of `consolidated` and `datapoints` will always match since neither can change

src/query/ts/m3db/encoded_series_iterator.go
arnikola: Fixed for now; Will be reconsidering how we propagate errs out of here so can revisit this then

src/query/ts/m3db/encoded_series_iterator.go
arnikola: Unfortunately that would make it awkward to fill up the values with nans if the consolidator is empty later on

src/query/ts/m3db/consolidator.go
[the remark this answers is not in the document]
arnikola: Renamed to consolidateAndMoveToNext; will need a slightly different consolidator soon

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+
+func (it *encodedStepIterUnconsolidated) Next() bool {
+	it.mu.Lock()
+	defer it.mu.Unlock()
+	if it.exhausted {
+		return false
+	}
+
+	if !it.started {
+		it.validIters = make([]bool, len(it.seriesIters))
```
andrewmains12: I think it'd be more clear if we did this on construction (could add a constructor to help with that).

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+		if err := iter.Err(); err != nil {
+			it.err = err
+		}
+	}
+
+	if !anyNext {
+		it.exhausted = true
+	}
+
+	return anyNext
```
andrewmains12: Should we be returning `next = true` in case of error? Or maybe we don't have to handle that case explicitly, if we assume that the underlying iterators do it.

src/query/ts/m3db/consolidator.go
```diff
@@ -0,0 +1,109 @@
...
+// consolidator is a helper for consolidating series in a step-wise
+// fashion. It takes a 'step' of values, which represents a vertical
+// slice of time across a list of series, and consolidates when a
+// valid step has been reached.
+type consolidator interface {
+	addPointForIterator(ts.Datapoint, int)
+	consolidate() []float64
+}
+
+func buildConsolidator(
```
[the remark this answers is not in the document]
arnikola: Sure

src/query/ts/m3db/consolidator.go
```diff
@@ -0,0 +1,109 @@
...
+	datapoints       [][]ts.Datapoint
+	fn               ConsolidationFunc
+}
+
+// consolidator is a helper for consolidating series in a step-wise
+// fashion. It takes a 'step' of values, which represents a vertical
+// slice of time across a list of series, and consolidates when a
+// valid step has been reached.
+type consolidator interface {
+	addPointForIterator(ts.Datapoint, int)
```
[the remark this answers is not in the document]
arnikola: I was trying to hide the internals from other files in the package; may refactor and put the consolidators in a separate package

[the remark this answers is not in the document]
arnikola: Sure, should be fixed by the refactor

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: Mentioned elsewhere, but we have a bit of a weird model with how we do Next() and Current(). If this returns false, we would never see the error since the downstream would assume that the iterator is exhausted, and wouldn't call `Current()`. I want to move to a model closer to seriesIterators, where we have a specialized `.Err()` method on the iterator, in which case this can return false.

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: Hopefully this should be addressed a bit in the design doc for the next phase

src/query/ts/m3db/encoded_block_iterator.go
arnikola: Added a brief comment here about why this bool exists; will expand on why when the logic for choosing to add/drop last point is added

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func (it *encodedStepIter) nextConsolidatedForStep(i int) {
+	peek := it.seriesPeek[i]
+	if peek.finished {
+		// No next value in this iterator
+		return
+	}
+
+	if peek.started {
```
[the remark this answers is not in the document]
arnikola: I think it's just a rather complex set of logic; not sure if it would make it much clearer, and we may just be kicking complexity a bit further down the stack

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -0,0 +1,134 @@
...
+	seriesIters []encoding.SeriesIterator
+}
+
+func (it *encodedStepIterUnconsolidated) Current() (
+	block.UnconsolidatedStep,
+	error,
+) {
+	it.mu.RLock()
+	if !it.started || it.exhausted {
+		it.mu.RUnlock()
```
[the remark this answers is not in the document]
arnikola: Deferring comes with a bit of a hit and this will get called pretty often. Probably a bit premature optimization, but there shouldn't be a lot of early returns, especially after the `Err` method is created on the iterator

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: Pulled it into the generator for now

src/query/ts/m3db/consolidators/consolidator_single.go
```diff
@@ -0,0 +1,97 @@
...
+	"time"
+
+	"github.com/m3db/m3/src/dbnode/ts"
+)
+
+// SingleLookbackConsolidator is a helper for consolidating a full single
+// series. It has some differences with the step consolidator in that it
+// collects points for a single series, and is reset when the next series
+// needs to be consolidated.
+type SingleLookbackConsolidator struct {
```
benraskin92: So this is like a series consolidator? Maybe rename to `SeriesLookbackConsolidator` and the other one can be `StepLookbackConsolidator`?

src/query/ts/m3db/consolidators/consolidator_single.go
```diff
@@ -0,0 +1,97 @@
...
+	fn ConsolidationFunc,
+) *SingleLookbackConsolidator {
+	datapoints := make([]ts.Datapoint, 0, initLength)
+
+	return &SingleLookbackConsolidator{
+		lookbackDuration: lookbackDuration,
+		stepSize:         stepSize,
+		earliestLookback: startTime.Add(-1 * lookbackDuration),
+		consolidated:     math.NaN(),
+		datapoints:       datapoints,
```
benraskin92: Maybe just initialize this here `datapoints: make([]ts.Datapoint, 0, initLength),`

src/query/ts/m3db/consolidators/consolidator.go
```diff
@@ -0,0 +1,108 @@
...
+func (c *LookbackConsolidator) AddPointForIterator(
+	dp ts.Datapoint,
+	i int,
+) {
+	if dp.Timestamp.Before(c.earliestLookback) {
+		// this datapoint is too far in the past, it can be dropped.
+		return
+	}
+
+	c.datapoints[i] = append(c.datapoints[i], dp)
```
benraskin92: Check to make sure `len(c.datapoints)-1 > i`?

src/query/ts/m3db/encoded_unconsolidated_series_iterator.go
```diff
@@ -0,0 +1,88 @@
...
+	for iter.Next() {
+		dp, _, _ := iter.Current()
+		values = append(values,
+			[]xts.Datapoint{{
+				Timestamp: dp.Timestamp,
+				Value:     dp.Value,
+			}})
+	}
+
+	if err := iter.Err(); err != nil {
```
benraskin92: nit: `if iter.Err() != nil { ... }`

benraskin92: Also do you need to unlock here?

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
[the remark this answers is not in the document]
andrewmains12: *Locking* comes with a bit of a hit :-p. Let's have that discussion about the locking--I'm still not sure what the use case is for the concurrent iterator.

src/query/ts/m3db/encoded_step_iterator.go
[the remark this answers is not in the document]
andrewmains12: That's fair, though I would argue peek in isolation is actually pretty simple, whereas combined with everything else it's harder to reason about. That said, the function's still under 100 lines or so, so I can live with the complexity here.

src/query/ts/m3db/consolidator.go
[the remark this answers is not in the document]
andrewmains12: > I was trying to hide the internals from other files in the package; 

Fair enough, though I don't think an interface really achieves this goal for you. It's all still in the same package, and it's not like these methods are implementation details; they're the public interface of the class. If anything, it just makes getting to the code harder for no reason. I'd say that we should eliminate it unless we actually intend to have other implementations.

This rant will probably come up again--I think we overuse interfaces in the code base in general :)

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,230 @@
...
+
+	if err := iter.Err(); err != nil {
+		it.err = err
+	}
+}
+
+func (it *encodedStepIter) nextConsolidated() {
+	end := it.bounds.End()
+	// Check that current time is not before end since end is exclusive
+	if it.currentTime.After(end) {
```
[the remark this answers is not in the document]
arnikola: I fear that this won't think it would necessarily save us a lot and might make some of the logic a bit more convoluted

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,230 @@
...
+	step := &encodedStep{
+		time:   currentTime,
+		values: it.consolidator.consolidate(),
+	}
+
+	it.mu.RUnlock()
+	return step, nil
+}
+
+func (it *encodedStepIter) nextConsolidatedForStep(i int) {
```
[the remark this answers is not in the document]
arnikola: Added

src/query/ts/m3db/consolidators/types.go
```diff
@@ -0,0 +1,47 @@
...
+		}
+	}
+
+	return math.NaN()
+}
+
+const initLength = 10
+
+// Set NaN to a variable makes tests easier.
+var nan = math.NaN()
```
benraskin92: Kind of weird to have this here since you don't use it. I would move it to tests.

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,168 @@
...
+		bounds:          bounds,
+	}
+
+	bl := newEncodedBlock(
+		seriesBlockIterators,
+		tagOptions,
+		consolidation,
+		lastBlock,
+	)
+	err := bl.generateMetas()
```
benraskin92:
```
if err := bl.generateMetas(); err != nil {
   return nil, err
}
```

src/query/ts/m3db/encoded_block_iterator.go
```diff
@@ -0,0 +1,168 @@
...
+	tags, metas := utils.DedupeMetadata(b.seriesMetas)
+	b.seriesMetas = metas
+	b.meta = block.Metadata{
+		Tags:   tags,
+		Bounds: b.consolidation.bounds,
+	}
+}
+
+func (b *encodedBlock) generateMetas() error {
+	err := b.buildSeriesMeta()
```
benraskin92:
```
if err := b.buildSeriesMeta(); err != nil {
  return err
}
```

src/query/ts/m3db/encoded_block_iterator_builder.go
```diff
@@ -0,0 +1,83 @@
...
+	tagOptions      models.TagOptions
+	consolidationFn consolidators.ConsolidationFunc
+}
+
+func newEncodedBlockBuilder(
+	tagOptions models.TagOptions,
+	consolidationFn consolidators.ConsolidationFunc,
+) *encodedBlockBuilder {
+	return &encodedBlockBuilder{
+		blocksAtTime:    make(map[string]encodedBlock, 10),
```
benraskin92: make `10` a const?

src/query/ts/m3db/consolidators/types.go
arnikola: It's in multiple tests, weird to have it in a single test and used in another

src/query/ts/m3db/encoded_block_iterator.go
arnikola: Don't know how valid it is, but a bit iffy about doing this pattern when the `err := XXX` has side effects

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,236 @@
...
+
+func (s *encodedStep) Time() time.Time   { return s.time }
+func (s *encodedStep) Values() []float64 { return s.values }
+
+func (it *encodedStepIter) Current() (block.Step, error) {
+	it.mu.RLock()
+	currentTime := it.currentTime
+	if !it.started || currentTime.After(it.bounds.End()) {
+		it.mu.RUnlock()
+		return nil, fmt.Errorf("out of bounds")
```
benraskin92: nit: `errors.New(...)`

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -0,0 +1,236 @@
...
+			return
+		}
+	}
+
+	iter := it.seriesIters[i]
+	// Read through iterator until finding a data point outside of the
+	// range of this consolidated step; then consolidate those points into
+	// a value, set the next peek value, and return true.
+	for iter.Next() {
+		dp, _, _ := iter.Current()
```
benraskin92: Check error?

arnikola: If seriesIter fails on a `Next` it will return false, so it will not enter this codeblock. Instead, error is checked after the `Next` block is finished

PR #1223, opened 2018-11-30
Add TLA+ specification for future M3DB version-based flushing

richardartoul:
- [X] Model versioned series buffer bucket flushes, including concurrent flushing, ticking, and writing
- [ ] Model  series blocks cache (and its invalidation)

### DIFF
--- a/specs/dbnode/flush/FlushVersion.tla
+++ b/specs/dbnode/flush/FlushVersion.tla
@@ -1,6 +1,8 @@
 ---------------------------- MODULE FlushVersion ----------------------------
 EXTENDS Integers, TLC
 
+\* Constants used to limit the potential search space when
+\* running the TLC model checker.
 CONSTANT maxNumWrites
 CONSTANT maxNumFlushes
 CONSTANT maxNumTicks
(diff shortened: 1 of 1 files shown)


specs/dbnode/flush/FlushVersion.tla
```diff
@@ -0,0 +1,253 @@
...
+\*       from memory process).
+\*
+\*
+\*    3. One process responsible for the "Ticking" process that will periodically evict
+\*       every bucket from the BucketsInMemory set whose LastSuccessfulFlushVersion is >=
+\*       CurrentLastSuccessfulFlushVersion and also > 0.
+\*
+\*
+\* The specification verifies the system by allowing all of these processes to run
+\* concurrently (as they do in M3DB) and asserting that the invariant that we can
```
justinjc: How does this get run concurrently? I just wonder if this will always run the three processes in the same order or pattern such that certain cases won't be caught.

specs/dbnode/flush/FlushVersion.tla
```diff
@@ -0,0 +1,253 @@
+---------------------------- MODULE FlushVersion ----------------------------
+EXTENDS Integers, TLC
+
+\* Constants used to limit the potential search space when
+\* running the TLC model checker.
+CONSTANT maxNumWrites
+CONSTANT maxNumFlushes
+CONSTANT maxNumTicks
```
justinjc: Do you want to include some recommended values for these as a rough guideline?

specs/dbnode/flush/FlushVersion.tla
richardartoul: It runs all processes and generates all possible interleaving of actions :D

PR #1224, opened 2018-12-02
m3msg: Add message pool max reuse size to avoid large messages overwhelming the pool

robskillington:


src/msg/consumer/config.go
```diff
@@ -30,13 +30,42 @@ import (
...
+	AckBufferSize             *int                      `yaml:"ackBufferSize"`
+	ConnectionWriteBufferSize *int                      `yaml:"connectionWriteBufferSize"`
+	ConnectionReadBufferSize  *int                      `yaml:"connectionReadBufferSize"`
+}
+
+// MessagePoolConfiguration is the message pool configuration
+// options, which extends the default object pool configuration.
+type MessagePoolConfiguration struct {
+	// Size is the size of the pool.
+	Size int `yaml:"size"`
```
richardartoul: Could you just embed pool.ObjectPoolConfiguration here instead of duplicating the fields?

robskillington: Yeah I tried that, it failed to populate the fields unfortunately.

Issue #641, opened 2018-05-15
Use a lockfile in filePathPrefix


prateek:
Already experienced configurations when multiple processes point to the same directory. Should catch this using a lockfile.

richardartoul:
Lets make sure we choose an implementation that automatically releases the lock if the process dies or the host restarts so we don't add an additional manual task that needs to be performed to recover from an abrupt failure.

edkolev:
I'll try to implement this, please assign to me.

Regarding the implementation, seems like we can't rely on automatic cleanup (due to hard restart, OOM killer, etc). Maybe the lock file could contain the pid of the process. Then, when we try to create a lock file, we'll ignore an existing lock file if it contains a dead PID (checking if a process is alive might be tricky though - maybe sending signal `0` could do it). Does something like this sound like a good approach?

richardartoul:
The PID solution seems brittle to me, especially since they get recycled and such. I did a little investigation and it seems like the ideal solution is to use the [fcntl system call](http://man7.org/linux/man-pages/man2/fcntl.2.html) - specifically see the "Advisory record locking" and "Open file description locks" sections. Specifically it says:

```
 As well as being removed by an explicit F_UNLCK, record locks are automatically released when the process terminates.
```

[This document](https://gavv.github.io/blog/file-locks/) also has a good summary.

It looks like these locks are automatically released when a process dies, and they're also allocated in memory so you should always end up in a clean state when a machine reboots.

richardartoul:
@edkolev 

Sure. I think this is a fairly re-usable thing that we may want for multiple services, so maybe we can put it in a shared location. Right now I'm thinking here would be good: https://github.com/m3db/m3/tree/master/src/x

Make it a standalone package / function so that we can use it in different places, and then in a separate P.R we can add it to M3DB.

edkolev:
Hey, Richard, I now see that I've provided wrong information in my comment above - the lock *file* is never removed automatically (when the process exits or is killed). Instead, just the *lock* on that file is removed. This makes sense because only the lock resides in the memory, not the file itself.

richardartoul:
@edkolev That seems like a reasonable solution to me right? It doesn't really matter whether the file is cleaned up or not as long as there is only ever one.

edkolev:
@richardartoul Correct, it should do the job - only one process will be able to hold a lock, and cleaning up locks will be optional.

richardartoul:
[...]

PR #1225, opened 2018-12-03
Initial lockfile implementation

edkolev:
This is an initial lockfile implementation, as discussed in #641  

I decided to add a file `scripts/lockfile/lockfile.go` which is used for testing.

### DIFF
--- a/scripts/lockfile/lockfile.go
+++ b/scripts/lockfile/lockfile.go
@@ -28,7 +28,7 @@ func main() {
 		exitWithUsage()
 	}
 
-	lock, err := lockfile.Create(path)
+	lock, err := lockfile.Acquire(path)
 	if err != nil {
 		os.Exit(1)
 	}
@@ -38,6 +38,6 @@ func main() {
 	}
 
 	if rmLock != "0" {
-		lock.Remove()
+		lock.Release()
 	}
 }
--- a/scripts/lockfile/lockfile.go
+++ b/scripts/lockfile/lockfile.go
@@ -1,0 +1,41 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"strconv"
+	"time"
+
+	"github.com/m3db/m3/src/x/lockfile"
+)
+
+func exitWithUsage() {
+	fmt.Printf("Usage: %s PATH SLEEP_SECONDS SHOULD_REMOVE_LOCK\n", path.Base(os.Args[0]))
+	os.Exit(1)
+}
+
+func main() {
+	if len(os.Args) != 4 {
+		exitWithUsage()
+	}
+
+	path, sleepStr, rmLock := os.Args[1], os.Args[2], os.Args[3]
+	sleep, err := strconv.Atoi(sleepStr)
+	if err != nil {
+		exitWithUsage()
+	}
+
+	lock, err := lockfile.Create(path)
+	if err != nil {
+		os.Exit(1)
+	}
+
+	if sleep > 0 {
+		time.Sleep(time.Duration(sleep) * time.Second)
+	}
+
+	if rmLock != "0" {
+		lock.Remove()
+	}
+}
--- a/scripts/lockfile/lockfile.go
+++ b/scripts/lockfile/lockfile.go
@@ -13,7 +13,9 @@ import (
 )
 
 func exitWithUsage() {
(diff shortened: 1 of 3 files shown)


edkolev:
@richardartoul I think this PR is ready to be reviewed. Please let me know if the PR can be improved.

scripts/lockfile/lockfile.go
```diff
@@ -0,0 +1,43 @@
...
+	"os"
+	"path"
+	"strconv"
+	"time"
+
+	"github.com/m3db/m3/src/x/lockfile"
+)
+
+func exitWithUsage() {
+	fmt.Printf("Usage: %s PATH SLEEP_SECONDS SHOULD_REMOVE_LOCK\n", path.Base(os.Args[0]))
```
richardartoul: nit: "Usage: %s <LOCK_FILE_PATH> <NUM_SECONDS_TO_SLEEP> <SHOULD_CLEANUP_LOCK>\n"

richardartoul: and then maybe one more line with a concrete example

src/x/lockfile/lockfile.go
```diff
@@ -0,0 +1,77 @@
...
+)
+
+type Lockfile struct {
+	file os.File
+}
+
+// Create creates the given file path if it doesn't exist and
+// obtains an exclusive lock on it. An error is returned if the lock
+// has been obtained by another process.
+func Create(path string) (*Lockfile, error) {
```
richardartoul: Could we call these `Lock` and `Unlock` or `Acquire` and `Release`

src/x/lockfile/lockfile.go
```diff
@@ -0,0 +1,77 @@
...
+	file, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, 0666)
```
richardartoul: what happens if some of the intermediary required directories don't exist? Will this still work or do you need a call to `os.MkDirAll()`

src/x/lockfile/lockfile.go
```diff
@@ -0,0 +1,77 @@
...
+	ft := &syscall.Flock_t{
+		Pid:  int32(os.Getpid()),
+		Type: syscall.F_UNLCK,
+	}
+
+	if err := syscall.FcntlFlock(lf.file.Fd(), syscall.F_SETLK, ft); err != nil {
+		return errors.Wrap(err, "failed releasing lock")
+	}
+
+	if err := os.Remove(lf.file.Name()); err != nil {
```
richardartoul: is it idiomatic to remove and then close?

src/x/lockfile/lockfile.go
edkolev: the caller will have to call `MkDirAll()` or he won't be able to obtain the lock

src/x/lockfile/lockfile.go
edkolev: I did some quick research but I didn't see any example of first remove then close -- I modified the code to close, then remove. I was thinking that there might be a race condition:
- process A closes the file handle
- process B creates the file and obtains the lock
- process A remove the files
- at this point B has a lock on a file which has been removed from the file system

But as I said, the code now closes, then removes, the above scenario doesn't seem likely. Let me know if you think we should dig deeper here and I'll try to reproduce the above scenario to check if it's a real concern.

src/x/lockfile/lockfile.go
edkolev: renamed

edkolev:
@richardartoul 

The PR is now updated with the feedback. I checked that `fcntl` is specified in POSIX, while `flock` isn't (which is the opposite to what I remembered). [posix link](http://pubs.opengroup.org/onlinepubs/9699919799/functions/fcntl.html)

Should I squash the commits?

Thanks, and please let me know if I can further improve the PR.

src/x/lockfile/lockfile.go
richardartoul: @edkolev If you delete a file that another process has a handle on, the file won't actually be deleted until that process releases the file

src/x/lockfile/lockfile.go
richardartoul: Can we add it here so callers don't have to think about it? Maybe can even be an additional function:

```golang
func CreateAndAcquire(path string, newDirMode os.FileMode) (*LockFile, error) {
    if err := os.MkdirAll(paths.Dir(path), newDirMode); err != nil {
        return nil, err
    }

    return Acquire(path)
}
```

PR #1229, opened 2018-12-03
[query] fixes NPE on tag completion with no backing stores

arnikola:


andrewmains12:
Noice thanks for the quick fix!

PR #1235, opened 2018-12-07
WIP Use apache thrift bytes pool to share dbnode server's bytes pool

robskillington:
Need to measure performance impact.


robskillington:
The performance of this was really really bad, abandoning.

PR #1240, opened 2018-12-11
Log ingest error

cw9:
Log the m3db write error in ingester for debugging

### DIFF
--- a/src/cmd/services/m3coordinator/ingest/ingest.go
+++ b/src/cmd/services/m3coordinator/ingest/ingest.go
@@ -22,6 +22,7 @@ package ingest
 
 import (
 	"context"
+	"math/rand"
 	"time"
 
 	"github.com/m3db/m3/src/cmd/services/m3coordinator/server/m3msg"
@@ -34,6 +35,7 @@ import (
 	"github.com/m3db/m3/src/x/serialize"
 	xerrors "github.com/m3db/m3x/errors"
 	"github.com/m3db/m3x/instrument"
+	"github.com/m3db/m3x/log"
 	"github.com/m3db/m3x/pool"
 	"github.com/m3db/m3x/retry"
 	xsync "github.com/m3db/m3x/sync"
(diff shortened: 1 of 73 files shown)


src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -136,11 +143,18 @@ type ingestOp struct {
...
+	return rand.Float32() < op.logSampleRate
+}
+
 func (op *ingestOp) ingest() {
 	if err := op.resetWriteQuery(); err != nil {
 		op.m.ingestError.Inc(1)
 		op.callback.Callback(m3msg.OnRetriableError)
 		op.p.Put(op)
+		if op.shouldLog() {
+			op.logger.Infof("could not reset ingest op, %v", err)
```
robskillington: Maybe `.Errorf`?

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -151,6 +165,9 @@ func (op *ingestOp) ingest() {
 		}
 		op.m.ingestError.Inc(1)
 		op.p.Put(op)
+		if op.shouldLog() {
+			op.logger.Infof("could not write ingest op, %v", err)
```
robskillington: Maybe `.Errorf`?

src/cmd/services/m3coordinator/ingest/ingest.go
```diff
@@ -119,13 +124,15 @@ func (i *Ingester) Ingest(
...
-	m         ingestMetrics
-	attemptFn retry.Fn
-	ingestFn  func()
+	s             storage.Appender
+	r             retry.Retrier
+	it            id.SortedTagIterator
+	p             pool.ObjectPool
+	m             ingestMetrics
+	logger        log.Logger
+	logSampleRate float32
```
robskillington: You might want to use an atomic int so that the very first error always comes out?

i.e. 
```go
import "github.com/uber-go/atomic"

type thing struct {
  logEvery int32
  numLogs *atomic.Int32
}

func newThing(sampleRate float64) *thing {
	return &thing{numLogs: atomic.NewInt32(0), logEvery: int32(float64(1.0/sampleRate)}
}

func (t *thing) shouldLog() bool {
	return t.numLogs.Inc(1)-1%t.logEvery
}
```

PR #1244, opened 2018-12-15
Add M3QL support to native prom read handler

benraskin92:


src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -312,3 +311,47 @@ func renderResultsInstantaneousJSON(
...
+		for i := 0; i < s.Len(); i++ {
+			dp := s.Values().DatapointAt(i)
+			jw.BeginArray()
+			jw.WriteFloat64(dp.Value)
+			jw.WriteInt(int(dp.Timestamp.Unix()))
+			jw.EndArray()
+		}
+		jw.EndArray()
+
+		jw.BeginObjectField("step_size_ms")
```
benraskin92: Figure out if apps team needs cost accounting fields. If not, I'll just leave them out (and maybe put a `todo`)

src/query/api/v1/handler/prometheus/native/common_test.go
```diff
@@ -186,7 +186,7 @@ func TestRenderResultsJSON(t *testing.T) {
 func TestRenderInstantaneousResultsJSON(t *testing.T) {
 	start := time.Unix(1535948880, 0)
 	buffer := bytes.NewBuffer(nil)
-	params := models.RequestParams{}
+	// params := models.RequestParams{}
```
benraskin92: Get rid of this

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -97,9 +98,14 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
+	if strings.ToLower(r.Header.Get("X-M3-Render-Format")) == "m3ql" {
+		w.Header().Set("Content-Type", "application/json")
+		renderM3QLResultsJSON(w, result, params)
+		return
+	}
+
 	// TODO: Support multiple result types
 	w.Header().Set("Content-Type", "application/json")
-	w.Header().Set("Access-Control-Allow-Origin", "*")
```
benraskin92: We can remove, right? @andrewmains12

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -97,9 +98,14 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if strings.ToLower(r.Header.Get("X-M3-Render-Format")) == "m3ql" {
```
nikunjgit: instead of this, can you add Format to RequestParams and then parse this header at the same place other things are parsed ? This would prevent us from adding adhoc parsing in our handlers.

src/query/api/v1/handler/prometheus/native/common.go
nikunjgit: lets keep it since we already have the info

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -97,9 +98,14 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		w.Header().Set("Content-Type", "application/json")
```
arnikola: Set this one line above so you don't have to copy it for each branch

src/query/api/v1/handler/prometheus/native/common.go
benraskin92: we do?

src/query/models/params.go
```diff
@@ -24,6 +24,14 @@ import (
 	"time"
 )
 
+// Format describes what format to return the data in
+type Format int
+
+const (
+	// M3QL returns results in M3QL format
+	M3QL Format = iota
```
arnikola: nit: `FormatType` and `FormatM3QL`

arnikola: Also, do we need `FormatPromQL`?

src/query/api/v1/handler/prometheus/native/read.go
andrewmains12: Yup should be good.

src/query/models/params.go
```diff
@@ -40,6 +48,7 @@ type RequestParams struct {
 	Query      string
 	Debug      bool
 	IncludeEnd bool
+	FormatType FormatType
```
arnikola: Won't this default to 0, which is equivalent to FormatM3QL; this would probably cause issues when trying to parse Prom. Might be easiest to add the prom option as the 0 default option

PR #1249, opened 2018-12-17
[query] Proof of concept of asynchronous block iterators

arnikola:
Adds asynchronous query execution for blocks. Has two different approaches:
1) Single channel per `Current()` call, with caller determining execution
2) Value channel that will handle iteration through `Current()` values in the iterator as available, and adds them to the channel with an appropriate index

Currently a race condition exists within the step channel approach
Next step is feedback/benchmarking to determine which approach is best, and refactoring more functions to allow async execution, then introducing actual async.

### DIFF
--- a/src/query/functions/aggregation/base_async.go
+++ b/src/query/functions/aggregation/base_async.go
@@ -1,0 +1,164 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package aggregation
+
+import (
+	"sync"
+
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/executor/transform"
+	"github.com/m3db/m3/src/query/functions/utils"
+	"github.com/m3db/m3/src/query/parser"
+	xsync "github.com/m3db/m3x/sync"
+)
+
+type baseNodeAsync struct {
+	op         baseOp
+	controller *transform.Controller
+
+	workerPool xsync.PooledWorkerPool
+}
+
+// Node creates an execution node
+func (o baseOp) AsyncNode(
+	controller *transform.Controller,
+	workerPool xsync.PooledWorkerPool,
+) transform.OpNodeAsync {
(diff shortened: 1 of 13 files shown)


arnikola:
Benchmarks added, average results across 10 runs:


1 Series,   100 Steps
Steps: 20000 runs, 83403.8 ns/op,  29250.8 allocs/op
Chan:  50000 runs, 25404.4 ns/op,  13686 allocs/op

100 Series, 100 Steps
Steps: 8000  runs, 203090 ns/op,   147566 allocs/op
Chan:  10000 runs, 142081 ns/op,   131903 allocs/op

10k Series, 100 Steps
Steps: 100 runs, 11473111.4 ns/op, 11334547.2 allocs/op
Chan:  100 runs, 12721183.8 ns/op, 11317352.9 allocs/op

100 Series, 10k Steps
Steps: 100 runs, 15840414.9 ns/op, 11696304.5 allocs/op
Chan:  100 runs, 10403478.3 ns/op, 10052040.3 allocs/op

10k Series, 10k Steps
Steps: 2.5 runs, 507341991 ns/op,  827119034.7 allocs/op
Chan:  2.8 runs, 499723487 ns/op,  823376332.7 allocs/op

src/query/functions/aggregation/base_async.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	nextBlock := builder.Build()
+	defer nextBlock.Close()
+	return n.controller.Process(nextBlock)
+}
+
+// ProcessValueChannel processes the block asynchronously using the value channel.
+func (n *baseNodeAsync) ProcessValueChannel(ID parser.NodeID, b block.AsyncBlock) error {
+	stepIter, err := b.AsyncStepIter()
```
andrewmains12: We don't actually use `stepIter` as an iterator here, which is confusing. Maybe we should just have a `ValuesChannel` method directly on the block?

src/query/functions/aggregation/base_async_benchmark_test.go
```diff
@@ -0,0 +1,242 @@
...
+func BenchmarkSteps10kSeries100Step(b *testing.B) {
+	benchmarkSteps(b, 10000, 100)
+}
+
+func BenchmarkValueChannel10kSeries100Step(b *testing.B) {
+	benchmarkValueChannel(b, 10000, 100)
+}
+
+func BenchmarkSteps100Series10kStep(b *testing.B) {
+	benchmarkSteps(b, 100, 10000)
```
andrewmains12: Not that it matters too much, but this could read nicely as a table driven benchmark:

```
func BenchmarkValueChannel(b *testing.B) {
   cases := []struct{
          Series int
          Step int
   }{
      {100, 100},
      {10000, 100},
      ...
    }

     for _, tc := range cases {
           b.Run("...", func(b *testing.B) {
                	benchmarkSteps(b, tc.Series, tc.Steps)
           })
     }
}
```

arnikola: Didn't know we could apply table driven tests to benchmarks but I guess it makes sense

src/query/ts/m3db/async_encoded_step_iterator.go
```diff
@@ -0,0 +1,274 @@
...
+	stepCount := it.StepCount()
+	ch := make(chan (block.IndexedStep), stepCount)
+	workerPools := it.workerPools
+	var wg sync.WaitGroup
+	wg.Add(it.StepCount())
+	for i := 0; it.Next(); i++ {
+		idx := i
+		currentTime := it.currentTime
+		currentValues := it.currentValues
+		workerPools.Go(func() {
```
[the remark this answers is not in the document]
arnikola: Currently it does, as soon as some of the blocks will need to start blocking, this will change. Idea is to build up from the simple case of being able to use async blocks in a synchronous-ish fashion to one where some steps are blocked until additional data comes in.

Making this async at the moment mirrors the potential for out-of-order writes, to ensure that the output can be put back together again correctly on the function side

src/query/functions/aggregation/base_async.go
arnikola: Good point; I'll play around with adding a different reader on the block and see how that works

If it ends up being a bit messy, might change the method name to `AsyncStepIter.ReadAll()` or something similar

src/query/ts/m3db/async_encoded_step_iterator.go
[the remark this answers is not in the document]
andrewmains12: Ah ok cool, fair enough. If the only purpose of this implementation is testing, let's move it to a test maybe?

arnikola:
Putting this on ice until a use case other than Prometheus lookback becomes apparent; can carry on splitting blocks for other data models and drop down to a full series approach when servicing Prometheus queries

PR #1256, opened 2018-12-19
[query] Hide m3 iterator expansion behind a query flag 

arnikola:
There seem to be issues with the new iterator expansion logic for very small datasets. This hides that logic behind a `iters=true` flag on the query until that issue is resolved

### DIFF
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -48,6 +48,7 @@ const (
 	queryParam        = "query"
 	stepParam         = "step"
 	debugParam        = "debug"
+	useIterParam      = "iters"
 	endExclusiveParam = "end-exclusive"
 
 	formatErrStr = "error parsing param: %s, error: %v"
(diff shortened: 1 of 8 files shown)


src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -188,6 +199,16 @@ func parseInstantaneousParams(r *http.Request) (models.RequestParams, *xhttp.Par
 		params.Debug = debug
 	}
 
+	// Skip useIterators if unable to parse useIterators param
```
benraskin92: Duplicate?

benraskin92: Pull into func?

src/query/storage/m3/storage.go
```diff
@@ -93,6 +93,15 @@ func (s *m3storage) FetchBlocks(
 	query *storage.FetchQuery,
 	options *storage.FetchOptions,
 ) (block.Result, error) {
+	if !options.UseIterators {
+		fetchResult, err := s.Fetch(ctx, query, options)
+		if err != nil {
+			return block.Result{}, err
+		}
+
+		return storage.FetchResultToBlockResult(fetchResult, query)
```
benraskin92: we still have tests for this, yeah?

PR #1258, opened 2018-12-20
Add CompleteTags method for M3 storage and by default enable it

robskillington:
Remote storage was hard coded to be the only storage allowed to return complete tags results, this sets it to local only by default and allows overriding any of the read/write/complete tags filter in config.

It also adds CompleteTags for M3 storage so it can be used for local storage CompleteTags queries.  This is for now just aggregating results of a search query, but in the future the DB could natively aggregate and return specifically just the unique tag values over the wire to the query server.


nikunjgit:
This has a lot of extraneous commits, maybe rebase ?

PR #1260, opened 2018-12-20
[query] Refactor to PR to use existing accumulators

arnikola:
Just some updates to Rob's proposed changes that use existing accumulator logic rather than handling it in the fetch tags function itself. Considering the time zone difference, figured might be easier to put up updated PR rather than leaving comments.

Currently the query looks good, but there is some issue with how we're using M3 `FetchTaggedIDs` endpoint

### DIFF
--- a/src/query/storage/m3/storage.go
+++ b/src/query/storage/m3/storage.go
@@ -240,65 +240,41 @@ func (s *m3storage) CompleteTags(
 	}
 
 	// TODO: instead of aggregating locally, have the DB aggregate it before
-	// sending results back
+	// sending results back.
 	fetchQuery := &storage.FetchQuery{
 		TagMatchers: query.TagMatchers,
 	}
+
 	results, cleanup, err := s.SearchCompressed(ctx, fetchQuery, options)
+	defer cleanup()
+
 	if err != nil {
 		return nil, err
 	}
 
-	defer func() { _ = cleanup() }()
-
-	tagsToTagValues := make(map[string]map[string]struct{})
+	accumulatedTags := storage.NewCompleteTagsResultBuilder(query.CompleteNameOnly)
 	for _, elem := range results {
-		for elem.Iter.Next() {
+		tags := make([]storage.CompletedTag, 0, elem.Iter.Len())
+		for i := 0; elem.Iter.Next(); i++ {
 			tag := elem.Iter.Current()
-			tagName := tag.Name.String()
-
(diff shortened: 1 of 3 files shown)


src/query/api/v1/handler/prometheus/remote/tag_values.go
```diff
@@ -82,5 +82,9 @@ func (h *TagValuesHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// TODO: Support multiple result types
-	prometheus.RenderTagCompletionResultsJSON(w, result)
+	err = prometheus.RenderTagValuesResultsJSON(w, result)
```
nikunjgit: maybe do error check separately ? On calling this function with the io.Writer, it might seem as if you can still have error while something is already written back.

src/query/storage/m3/storage.go
```diff
@@ -240,65 +241,57 @@ func (s *m3storage) CompleteTags(
...
 	for _, elem := range results {
-		for elem.Iter.Next() {
-			tag := elem.Iter.Current()
-			tagName := tag.Name.String()
-
-			values, ok := tagsToTagValues[tagName]
-			if !ok {
-				values = make(map[string]struct{})
-				tagsToTagValues[tagName] = values
+		it := elem.Iter
```
nikunjgit: maybe move this out into a separate function ?

src/query/storage/m3/storage.go
```diff
@@ -240,65 +241,57 @@ func (s *m3storage) CompleteTags(
...
+		tags := make([]storage.CompletedTag, 0, it.Len())
+		for i := 0; it.Next(); i++ {
+			tag := it.Current()
+			name := tag.Name.Bytes()
+			found := false
+			for _, filterName := range query.FilterNameTags {
```
nikunjgit: consider separating this into Any() function ? https://gobyexample.com/collection-functions

src/query/api/v1/handler/prometheus/remote/tag_values.go
arnikola: It only completes the write in the flush method in the `jw.Close()` call, so if it errors, nothing is written.

src/query/storage/m3/storage.go
arnikola: It doesn't really seem worth it, there's not a lot of logic here

Issue #1259, opened 2018-12-20
Remove m3 iterator expansion flag in query


nikunjgit:
We should get rid of the flag introduced in https://github.com/m3db/m3/pull/1256 once the logic is stable.

PR #1266, opened 2018-12-24
Arnikola/fix compressed iterator bugs

arnikola:
Fixes bugs in compressed block iterator processing

Bugs fixed:
1. Unconsolidated series iterator was returning series of an incorrect
length, since it would not align correctly to bounds. This could cause
panics in query execution depending on downstream functions.

2. Consolidated series iterator would attempt to fill in values if its
internal consolidator still had valid values, without ensuring that those
values were still needed

3. Fetch function was closing blocks too early which caused values to be
lost in compressed block iterators. This was not an issue before since
the previous implementation was not backed by pooled arrays so closing
was less impactful

4. Tags common to all series in a block now correctly added to output (this was an issue in the decompressed case, too)

Also resolves https://github.com/m3db/m3/issues/1259 by flipping the flag and defaulting to compressed iterators

### DIFF
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -48,8 +48,12 @@ const (
 	queryParam        = "query"
 	stepParam         = "step"
 	debugParam        = "debug"
-	useIterParam      = "iters"
 	endExclusiveParam = "end-exclusive"
+
+	// NB: this will force series decompression on fetch. Will be left for a few
+	// releases to help debug any issues that may arise with compressed series
+	// block fetch
+	useLegacyParam = "legacy"
 
 	formatErrStr = "error parsing param: %s, error: %v"
 )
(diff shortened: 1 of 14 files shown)


src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -138,11 +142,12 @@ func parseParams(r *http.Request) (models.RequestParams, *xhttp.ParseError) {
 	return params, nil
 }
 
-func parseDebugAndIterFlags(r *http.Request) (bool, bool) {
+func parseDebugAndLegacyFlags(r *http.Request) (bool, bool) {
```
nikunjgit: should these be two methods so that its easier to get rid of legacy later ?

arnikola: I was thinking of adding a third flag which would enable the split block approach to test that when that is added, might split it out then to have a debug parser and a "fallback" parser

PR #1269, opened 2018-12-27
Add a debug endpoint with information for host and heapdump

lyuboraykov:
It returns a zip file with the dumped data.
The zip archive contains multiple files, each one containing data from
the providers.

This is just an example of two things to be dumped.
I'll add more once we get the approach nailed down.


src/x/debugdump/debug.go
```diff
@@ -0,0 +1,78 @@
...
+
+package debugdump
+
+import (
+	"archive/zip"
+	"fmt"
+	"io"
+)
+
+// DataProvider returns an instance of a provider for debug information
```
richardartoul: super nit: period at the end of all comments

src/x/debugdump/debug.go
```diff
@@ -0,0 +1,78 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
richardartoul: nit: 2019

src/x/debugdump/debug.go
```diff
@@ -0,0 +1,78 @@
...
+package debugdump
+
+import (
+	"archive/zip"
+	"fmt"
+	"io"
+)
+
+// DataProvider returns an instance of a provider for debug information
+type DataProvider interface {
```
richardartoul: The name of this package is already so well scoped that I wonder if this could just be called provider.

I.E

`debugdump.Provider` vs. `debugdump.DataProvider`

That feels a little more intuitive to me because then it seems more obvious that a provider is providing debug information as opposed to jus "data" but I don't have super strong feelings

src/x/debugdump/debug.go
```diff
@@ -0,0 +1,78 @@
...
+	// ProvideData writes it's debug information into the provided writer
+	ProvideData(w io.Writer) error
```
richardartoul: Maybe just call this `Write`?

src/x/debugdump/debug.go
richardartoul: Alternatively:

```
type Source interface {
    Write(w io.Writer) error
}
```

Seems reasonable to me too

src/x/debugdump/debug.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// DataDumper represents a collection of InfoProviders
+type DataDumper interface {
```
richardartoul: Not a big fan of this name. How do you feel about something like: `Writer` or `ArchiveWriter` or `ZipWriter` etc.

src/x/debugdump/debug.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package debugdump
```
richardartoul: It might be nice to just have a "debug" package that contains this functionality so that we can add more debugging utilities into this package later.

src/x/debugdump/debug.go
```diff
@@ -0,0 +1,78 @@
...
+type DataProvider interface {
+	// ProvideData writes it's debug information into the provided writer
+	ProvideData(w io.Writer) error
+}
+
+// DataDumper represents a collection of InfoProviders
+type DataDumper interface {
+	// Dump writes a ZIP file in the provided writer
+	// the archive contains the dumps of all the providers in separate files
+	DumpData(io.Writer) error
```
richardartoul: Same comment about the name

src/x/debugdump/debug_test.go
```diff
@@ -0,0 +1,76 @@
...
+	return nil
+}
+
+func TestDumpData(t *testing.T) {
+	dataDumper := NewDataDumper()
+	fp := &fakeProvider{}
+	dataDumper.RegisterProvider("test", fp)
+	buff := bytes.NewBuffer([]byte{})
+	err := dataDumper.DumpData(buff)
+	require.NotZero(t, buff.Len())
```
richardartoul: Can you make this actually assert that the correct data was written into the buffer?

src/x/debugdump/debug_test.go
```diff
@@ -0,0 +1,76 @@
...
+func (f *fakeProvider) ProvideData(w io.Writer) error {
+	f.called = true
+	if f.shouldErr {
+		return errors.New("bad provide")
+	}
+	w.Write([]byte("test"))
+	return nil
+}
+
+func TestDumpData(t *testing.T) {
```
richardartoul: It looks like the `archive/zip` package also has support for reading zip files. Do you mind adding a test that:

1. Registers two different providers
2. Generates the archive
3. Reads the archive and asserts on the bytes in each file

That way we have an end-to-end test of writing and reading

src/x/debugdump/host.go
```diff
@@ -0,0 +1,59 @@
...
+func (h *hostDataProvider) ProvideData(w io.Writer) error {
+	wd, err := os.Getwd()
+	if err != nil {
+		return err
+	}
+	hostInfo := &hostInformation{
+		PID: os.Getpid(),
+		CWD: wd,
+	}
+	jsonData, err := json.Marshal(hostInfo)
```
richardartoul: Wanna use [MarshalIndent](https://stackoverflow.com/questions/19038598/how-can-i-pretty-print-json-using-go) to get slightly easier to read output?

src/x/debugdump/host_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestHostDataProvider(t *testing.T) {
+	prov := NewHostDataProvider()
+	buff := bytes.NewBuffer([]byte{})
+	prov.ProvideData(buff)
+	require.NotZero(t, buff.Len())
```
richardartoul: Maybbe just assert you're able to unmarshal without errors

src/x/debugdump/heap.go
```diff
@@ -0,0 +1,39 @@
...
+
+package debugdump
+
+import (
+	"io"
+	"runtime/pprof"
+)
+
+// heapDumpProvider provides a heap dump of the process
+type heapDumpProvider struct{}
```
richardartoul: How do you feel about adding a CPU provider as well? That would make this P.R significantly more useful once its landed.

I think it should be as simple as something like:

```golang
type cpuProfileProvider struct {
    duration time.Duration
}

func (c *cpuProfileProvider) ProvideData(w io.Writer) error {
    err := pprof.StartCPUProfile(w); err != nil {
        return err
    }

    time.Sleep(h.duration)
    pprof.StopCPUProfile()
}

Ideally we'd be able to control the duration at runtime too, so maybe add a sync.Mutex into the struct and then a public method for modifying the duration, but you get the idea.

File: src/x/debugdump/debug.go [hunk 13]
```diff
@@ -0,0 +1,78 @@
...
+	// RegisterProvider adds a new provider the aggregator
+	RegisterProvider(string, DataProvider) error
+}
+
+type dataDumper struct {
+	providers map[string]DataProvider
+}
+
+// NewDataDumper returns an instance of an DataDumper
+func NewDataDumper() DataDumper {
```
richardartoul: Wanna add a constructor function with defaults too? I.E `NewDefaultDataDumper()` or whatever and when you call it returns one that is already initialized with all the default providers.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -21,11 +21,14 @@
...
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"sync"
 	"time"
 
+	"github.com/m3db/m3/src/x/debugdump"
+
```
hodgesds: Extra space not needed.

src/x/debugdump/debug.go
```diff
@@ -0,0 +1,78 @@
...
+	return &dataDumper{
+		providers: make(map[string]DataProvider),
+	}
+}
+
```
hodgesds: This should have a comment something similar to:
```
// RegisterProvider implements the DataDumper interface.
```

src/x/debugdump/debug.go
```diff
@@ -0,0 +1,78 @@
...
+func (i *dataDumper) RegisterProvider(dumpFileName string, p DataProvider) error {
+	if _, ok := i.providers[dumpFileName]; ok {
+		return fmt.Errorf("dumpfile already registered %s", dumpFileName)
+	}
+	i.providers[dumpFileName] = p
+	return nil
+}
+
```
hodgesds: Similar as above, but also include how errors are handled.

src/x/debugdump/heap.go
lyuboraykov: Yeah, makes perfect sense. Though I'm not sure about controlling the duration, you'd have to keep the object somewhere, because the ZipWriter doesn't expose the Sources.
Or it could have a `Source(filename string) Source`.

.gitignore
```diff
@@ -29,6 +29,9 @@ _tools/
 debug
 debug.test
 
+# debug package
+!src/x/debug
```
richardartoul: what is this for?

src/x/debug/debug.go
```diff
@@ -0,0 +1,82 @@
...
+
+package debug
+
+import (
+	"archive/zip"
+	"fmt"
+	"io"
+)
+
+// Source provides functions for fething debug information from a single debug source.
```
richardartoul: fething ->fetching

src/x/debug/debug.go
```diff
@@ -0,0 +1,82 @@
...
+type Source interface {
+	// Write writes it's debug information into the provided writer
```
richardartoul: nit: period (I really need to write a linter for this...)

PR #1270, opened 2018-12-27
[M3agg] Add flush handler to flush aggregated metrics in protobuf

cw9:
New flush handler to flush aggregated metrics in protobuf

Placeholder for config changes, could turn on sending/receiving metrics in protobuf by toggling the config in the future, need to coordinate external users.

### DIFF
--- a/src/aggregator/aggregator/handler/config.go
+++ b/src/aggregator/aggregator/handler/config.go
@@ -54,6 +54,7 @@ var (
 	errNoWriterConfiguration                    = errors.New("no writer configuration")
 	errNoDynamicOrStaticBackendConfiguration    = errors.New("neither dynamic nor static backend was configured")
 	errBothDynamicAndStaticBackendConfiguration = errors.New("both dynamic and static backend were configured")
+	errInvalidShardingConfiguration             = errors.New("invalid sharding configuration, missing hash type or total shards")
 )
 
 // FlushHandlerConfiguration configures flush handlers.
@@ -92,7 +93,7 @@ func (c FlushHandlerConfiguration) NewHandler(
 			return nil, err
 		}
 		if hc.DynamicBackend != nil {
-			handlers, sharderRouters, err = hc.DynamicBackend.newHandler(
+			handlers, sharderRouters, err = hc.DynamicBackend.constructDynamicBackend(
 				handlers,
 				sharderRouters,
 				cs,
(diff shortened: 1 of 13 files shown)


src/aggregator/aggregator/handler/config.go
```diff
@@ -237,7 +314,7 @@ func (c *dynamicBackendConfiguration) NewSharderRouter(
 		r = trafficcontrol.NewRouter(tc, r, scope)
 	}
 	return SharderRouter{
-		SharderID: sharding.NewSharderID(c.HashType, c.TotalShards),
+		SharderID: sharding.NewSharderID(*c.HashType, *c.TotalShards),
```
jeromefroe: Is it ok to dereference these fields without checking if they `nil` first? In `newHandler` you first check that `c.ProtobufEnabled` is not `nil` before dereferencing it.

src/aggregator/aggregator/handler/writer/protobuf.go
```diff
@@ -0,0 +1,185 @@
...
+	return w, nil
+}
+
+func (w *protobufWriter) Write(mp aggregated.ChunkedMetricWithStoragePolicy) error {
+	if w.closed {
+		w.metrics.writerClosed.Inc(1)
+		return errWriterClosed
+	}
+	var encodeNanos int64
+	if w.encodingTimeSamplingRate > 0 && w.randFn() < w.encodingTimeSamplingRate {
```
jeromefroe: nit: Is the `w.encodingTimeSamplingRate > 0` check redundant since if `w.encodingTimeSamplingRate` is 0 then `w.randFn() < w.encodingTimeSamplingRate` will always be false?

cw9: I moved this code from another place, I think it's still useful as when w.encodingTimeSamplingRate is less than 0 we skip the randFn() call

src/aggregator/aggregator/handler/config.go
```diff
@@ -91,15 +92,17 @@ func (c FlushHandlerConfiguration) NewHandler(
 			return nil, err
 		}
 		if hc.DynamicBackend != nil {
-			sharderRouter, err := hc.DynamicBackend.NewSharderRouter(
+			handlers, sharderRouters, err = hc.DynamicBackend.newHandler(
+				handlers,
```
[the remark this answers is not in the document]
cw9: I can't check sharderRouter against nil because it's a struct, unless I return pointer which is uglier, I'll keep this for now and remove the sharderRouter code once migration is done, then I can just append the handler here.

src/aggregator/aggregator/handler/config.go
cw9: done

PR #1271, opened 2018-12-29
[query] Refactor block iterators, fix unconsolidated series iterator

arnikola:
Changes query's iterators to match M3 Iterators; i.e. not erroring on `Current` methods but having a dedicated

Also moves all processing onto the `Next` function, allowing `Current` to be a much simpler method in all iterators, and consolidating complexity in `Next`

Also fixes an error with unconsolidated series encoded block iterators


src/query/block/column.go
```diff
@@ -237,19 +261,25 @@ func (m *columnBlockSeriesIter) SeriesCount() int {
...
+			m.values[i] = cols[i].Values[m.idx]
+		}
 	}
 
-	return NewSeries(values, m.seriesMeta[m.idx]), nil
+	return next
+}
+
+func (m *columnBlockSeriesIter) Current() Series {
+	return NewSeries(m.values, m.seriesMeta[m.idx])
```
andrewmains12: Just checking--does our usage of the iterator ensure that we never use the `values` slice after another call to `Next`?

i.e. nothing like:

```
firstValues := iter.Current()
iter.Next()
nextValues := iter.Current()
``` 

Maybe worth adding a comment to the `Next` method saying that it invalidates previously returned arrays, and that users should copy if they need to keep that data around?

src/query/block/types.go
```diff
@@ -46,122 +48,125 @@ type Block interface {
 // UnconsolidatedBlock represents a group of unconsolidated series across a time bound
 type UnconsolidatedBlock interface {
 	io.Closer
-	// StepIter returns an UnconsolidatedStepIter
+	// StepIter returns a step-wise block iterator, giving unconsolidated values
```
andrewmains12: Thanks a lot for updating these to make them more informative :)

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -21,99 +21,95 @@
...
-		} else {
-			it.validIters[i] = false
-		}
-
-		if err := iter.Err(); err != nil {
-			it.err = err
+	it.idx++
+	if it.idx == 0 {
+		// decode the series on initial Next() call
+		if success := it.decodeSeries(); !success {
```
andrewmains12: Am I reading correctly that this change switches us back to eager decompression? What's the reasoning behind the switch back?

src/query/block/column.go
```diff
@@ -115,23 +117,34 @@ func (c *colBlockIter) StepCount() int {
 
 // Next returns true if iterator has more values remaining
 func (c *colBlockIter) Next() bool {
```
nikunjgit: maybe write comments regarding the flow ? Eg: check for Next() if that's false, then also check for Err() ?

src/query/block/scalar.go
```diff
@@ -166,12 +174,15 @@ func (it *scalarSeriesIter) Next() bool {
 	return it.idx == 0
 }
 
-func (it *scalarSeriesIter) Current() (Series, error) {
+func (it *scalarSeriesIter) Current() Series {
 	if it.idx != 0 {
-		return Series{}, fmt.Errorf("invalid scalar index: %d, numVals: 1", it.idx)
+		// Indicates an error with the caller, having either not called Next() or attempted
```
nikunjgit: In most other places, we don't check if Next() has been called before Current() ? Also, having a panic here while you also have an Err() field is confusing. Maybe just get rid of this check ?

src/query/block/types.go
nikunjgit: +1

src/query/block/types.go
```diff
@@ -46,122 +48,125 @@ type Block interface {
...
 
-// SeriesMeta is metadata data for the series
+// SeriesMeta is metadata data for the series.
 type SeriesMeta struct {
 	Tags models.Tags
 	Name string
 }
 
-// Iterator is the base iterator
+// Iterator is the base iterator.
```
nikunjgit: Write a bit about how the flow should work (as mentioned above) ?

src/query/ts/m3db/encoded_unconsolidated_series_iterator.go
```diff
@@ -21,53 +21,55 @@
...
 func (it *encodedSeriesIterUnconsolidated) Next() bool {
-	it.mu.Lock()
+	if it.err != nil {
+		return false
+	}
+
 	it.idx++
 	next := it.idx < len(it.seriesIters)
-	it.mu.Unlock()
+	if next {
```
nikunjgit: flip to reduce nesting ?

src/query/block/column.go
arnikola: SGTM. Considering that these are interface methods though, might get rid of the comments here and make the iterator ones more meaningful

src/query/block/column.go
arnikola: Added a comment to reflect this. I think failing anything else coming up with more priority, my next task would include adding pooling to these iterators, which should make it safer.

Also this one will only be used by LegacyBlockProcessing, which shouldn't be around much longer and is not the default, so less worried about it.

arnikola: Actually, just going to copy the array instead as this would be a nightmare to debug if it happens :p

src/query/block/scalar.go
arnikola: Sure, just a lot of iterators work on top of slices so if you do something like call `Current` before `Next` or after `Next` is false, you'll get an out of bounds panic, so was to replicate that. Happy to delete it, it's not really a useful case, had it here so that it would be easier to detect if we were doing something wrong in the caller.

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: Yeah, so the original implementation was incorrect, so had to fix it. I had a few questions with unconsolidated step iteration as a concept, since it's a bit odd, and was thinking of maybe just getting rid of it in general.

Talked offline and will fix this up to do it lazily the right way instead.

src/query/block/column.go
```diff
@@ -237,19 +252,30 @@ func (m *columnBlockSeriesIter) SeriesCount() int {
 	return len(cols[0].Values)
 }
 
+func (m *columnBlockSeriesIter) Err() error {
+	return nil
```
benraskin92: Add comment that this is `no-op`?

src/query/api/v1/handler/prometheus/native/read_common.go
```diff
@@ -224,27 +228,18 @@ func insertSortedBlock(
...
-	blockStepCount := blockStepIter.StepCount()
-	if stepCount != blockStepCount {
-		return nil, fmt.Errorf("mismatch in number of steps for the"+
-			"block, wanted: %d, found: %d", stepCount, blockStepCount)
-	}
-
 	// Binary search to keep the start times sorted
 	index := sort.Search(len(blockList), func(i int) bool {
-		return blockList[i].meta.Bounds.Start.Before(blockMeta.Bounds.Start)
+		return blockList[i].meta.Bounds.Start.After(blockMeta.Bounds.Start)
```
benraskin92: Why change this?

arnikola: Because it was wrong; it sorted in the wrong order

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -21,99 +21,150 @@
...
+		it.seriesPeek[i].started = false
+		// Currently at a potentially viable data point.
+		// Record previously peeked value, and all potentially valid values.
+		it.accumulator.AddPointForIterator(point, i)
 
-		values[i] = []xts.Datapoint{{
-			Timestamp: dp.Timestamp,
-			Value:     dp.Value,
-		}}
+		// If at boundary, add the point as the current value.
```
benraskin92: Where do you add the point?

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -21,99 +21,150 @@
...
 
-	if it.err != nil {
-		it.mu.RUnlock()
-		return nil, it.err
+func (it *encodedStepIterUnconsolidated) Current() block.UnconsolidatedStep {
+	points := it.accumulator.AccumulateAndMoveToNext()
+	return storage.NewUnconsolidatedStep(it.stepTime, points)
+}
+
+func (it *encodedStepIterUnconsolidated) nextForStep(
```
benraskin92: Can you add high level comment explaining why we need this function?

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
```diff
@@ -21,99 +21,150 @@
...
+			// This point exists further than the current step; set peeked value
+			// to this point.
+			it.seriesPeek[i].point = dp
+			it.seriesPeek[i].started = true
+			return
+		}
+	}
 }
 
 func (it *encodedStepIterUnconsolidated) Next() bool {
```
benraskin92: Given that the `Next()` functions are pretty similar for series and step, think it's worth moving the common parts into a common function?

src/query/storage/unconsolidated_test.go
```diff
@@ -0,0 +1,155 @@
...
+
+	i := 0
+	for iter.Next() {
+		series := iter.Current()
+		dps := datapointsToFloatSlices(t, series.Datapoints())
+		assert.Equal(t, expected[i], dps)
+		i++
+	}
+
+	assert.Equal(t, len(expected), i)
```
benraskin92: `assert.Len()`

src/query/ts/m3db/consolidators/step_accumulator.go
```diff
@@ -0,0 +1,119 @@
...
+
+// AddPointForIterator adds a datapoint to a given step if it's within the valid
+// time period; otherwise drops it silently, which is fine for accumulation.
+func (c *StepLookbackAccumulator) AddPointForIterator(
+	dp ts.Datapoint,
+	i int,
+) {
+	if dp.Timestamp.Before(c.earliestLookback) {
+		// this datapoint is too far in the past, it can be dropped.
+		return
```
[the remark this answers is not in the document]
arnikola: No; would be a lot of annoying spam

src/query/ts/m3db/convert.go
```diff
@@ -35,15 +37,24 @@ const (
...
+	readStart time.Time
 	blockSize time.Duration
 	replicas  []encoding.MultiReaderIterator
 }
 
 type seriesBlocks []seriesBlock
 
+func (b seriesBlock) String() string {
+	return fmt.Sprint("BlockSize:", b.blockSize.Hours(), " blockStart:",
+		b.blockStart.Format("3:04:05PM"), " readStart:", b.readStart.Format("3:04:05PM"),
```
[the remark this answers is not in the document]
arnikola: That's just how [go rolls](https://golang.org/pkg/time/#pkg-constants)

src/query/storage/unconsolidated_test.go
arnikola: Unfortunately assert.Len gives pretty garbage output if it fails:
`"[%!s(int=1) %!s(int=2) %!s(int=3) %!s(int=4)]" should have 100 item(s), but has 4`
vs
`Not equal: 100 (expected) != 4 (actual)`

If `expected` doesn't have good String methods, you'll get a bunch of unintelligible spam

src/query/README.md
```diff
@@ -145,7 +146,8 @@ Input JSON:
   }
 }
 ```
-Full request: `curl -X POST 'localhost:7201/api/v1/debug/validate_query?start=1543431465&end=1543435045&step=14s&query=go_gc_duration_seconds%7Bquantile%3D%221%22%7D*2' -d @<input_json_file> --header "Content-Type: application/json"`
```
arnikola: Oops, reverting this

src/query/README.md
```diff
@@ -145,7 +146,8 @@ Input JSON:
...
+
+Full request: `curl -X POST 'http://localhost:7201/api/v1/debug/validate_query?query=sum_over_time(process_cpu_seconds_total%7Binstance%3D~%22m3db_seed%3A7203%22%7D%5B60s%5D)&start=1549035092&end=1549035177&step=1' -d @lol.json --header "Content-Type: application/json"`
```
benraskin92: lol? lol

src/query/block/column.go
```diff
@@ -113,30 +109,38 @@ func (c *colBlockIter) StepCount() int {
...
-// Current returns the current step
-func (c *colBlockIter) Current() (Step, error) {
-	col := c.columns[c.idx]
-	t, err := c.meta.Bounds.TimeForIndex(c.idx)
-	// TODO: Test panic case
-	if err != nil {
-		panic(err)
+	c.timeForStep, c.err = c.meta.Bounds.TimeForIndex(c.idx)
+	if c.err != nil {
+		return false
```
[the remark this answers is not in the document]
arnikola: The changed iterator flow here sets the error to `err` on the `colBlockIter`; it's up to the caller to check `iter.Error()` is nil after iteration's finished

src/query/ts/m3db/encoded_series_iterator.go
```diff
@@ -22,7 +22,6 @@ package m3db
```
benraskin92: Maybe not in this PR, but we should organize all of these iterator files a bit better.

src/query/README.md
robskillington: Maybe revert to `@<input_json_file>`?

src/query/executor/transform/lazy.go
```diff
@@ -86,25 +88,42 @@ func (s *stepIter) StepCount() int {
...
 
 func (s *stepIter) Next() bool {
-	return s.iter.Next()
+	if s.err != nil {
+		return false
+	}
+
+	next := s.iter.Next()
+	if !next {
+		s.err = s.iter.Err()
```
robskillington: Hm, instead of adding this here, perhaps in `Err()` we can instead just return it as well when it's asked for?  Just otherwise this logic is just a little bit more difficult to follow (also if the inner iterator gets an error for some other reason later, it will properly propagate, unlike here where it can only get set during a call to next?):

```go
func (s *stepIter) Err() error {
	if s.err != nil {
		return s.err
	}
	return s.iter.Err()
}
```

src/query/executor/transform/lazy.go
```diff
@@ -123,17 +142,32 @@ func (s *seriesIter) Close() {
...
 
 func (s *seriesIter) Next() bool {
-	return s.iter.Next()
+	if s.err != nil {
+		return false
+	}
+
+	next := s.iter.Next()
+	if !next {
+		s.err = s.iter.Err()
```
robskillington: Same comment here.

src/query/storage/unconsolidated_test.go
```diff
@@ -0,0 +1,155 @@
...
+	vals := make([][]float64, len(dps))
+	for i, dp := range dps {
+		vals[i] = dp.Values()
+	}
+
+	return vals
+}
+
+func TestUnconsolidatedStep(t *testing.T) {
+	expected := [][][]float64{
```
robskillington: Nice 👍

src/query/storage/unconsolidated_test.go
robskillington: Agreed with artem, assert.Len is pretty disappointing.

src/query/storage/validator/storage.go
```diff
@@ -115,17 +116,23 @@ func PromResultToSeriesList(promReadResp prometheus.PromResp, tagOptions models.
...
+					Name:  []byte(name),
+					Value: []byte(val),
+				})
+			}
 		}
 
 		name, exists := tags.Name()
 		if !exists {
-			return nil, errors.New("metric name does not exist")
+			// return nil, errors.New("metric name does not exist at all ever")
```
robskillington: Hm, don't we want to restore this return err?  What possibility would we want to allow for empty series names?

src/query/ts/m3db/consolidators/step_accumulator.go
```diff
@@ -0,0 +1,92 @@
...
+
+	c.datapoints[i] = append(c.datapoints[i], xts.Datapoint{
+		Timestamp: dp.Timestamp,
+		Value:     dp.Value,
+	})
+}
+
+// AccumulateAndMoveToNext consolidates the current values and moves the
+// consolidator to the next given value, purging stale values.
+func (c *StepLookbackAccumulator) AccumulateAndMoveToNext() []xts.Datapoints {
```
robskillington: Interesting 👍

src/query/ts/m3db/convert.go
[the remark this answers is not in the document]
robskillington: Can you add `"3:04:05PM"` as a constant perhaps to top of this file, like `blockTimeFormat = ...`?

src/query/ts/m3db/convert.go
```diff
@@ -123,6 +151,11 @@ func blockReplicasFromSeriesIterator(
 	pools encoding.IteratorPools,
 ) (seriesBlocks, error) {
 	blocks := make(seriesBlocks, 0, bounds.Steps())
+	var pool encoding.MultiReaderIteratorPool
+	if pools != nil {
+		pool = pools.MultiReaderIterator()
```
robskillington: 🎉 nice

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -32,14 +30,7 @@ import (
...
 )
 
-type peekValue struct {
-	started  bool
-	finished bool
-	point    ts.Datapoint
-}
-
 type encodedStepIter struct {
-	mu           sync.RWMutex
```
robskillington: Re: Removing the mutex, is there no concurrent access now?

src/query/api/v1/handler/prometheus/native/read_common.go
```diff
@@ -171,22 +165,31 @@ func sortedBlocksToSeriesList(blockList []blockWithMeta) ([]*ts.Series, error) {
 		seriesIters[i] = seriesIter
 	}
 
-	numValues := firstStepIter.StepCount() * len(blockList)
+	numValues := 0
+	for _, block := range blockList {
+		b, _ := block.block.StepIter()
```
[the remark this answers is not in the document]
arnikola: Oh yeah, nice catch. Had added this path quick and dirty to test it was working correctly, then missed cleaning up and adding the error case here.

src/query/executor/transform/lazy.go
arnikola: Nice one, will do 👍

src/query/executor/transform/lazy.go
arnikola: 👍

src/query/ts/m3db/encoded_step_iterator.go
arnikola: Yeah there wasn't really any before either, concurrent access on an iterator is a bit weird but was going to be necessary for a proposed lookback fix for split blocks.

src/query/ts/m3db/encoded_series_iterator.go
arnikola: Yeah that's fair, it's kind of in a weird path right now

src/query/storage/validator/storage.go
arnikola: Added a comment explaining why it's fine to set to `default` 👍

src/query/ts/m3db/encoded_unconsolidated_step_iterator.go
arnikola: It's a bit weird, but the point has been added to the consolidator at this point and it's already at the boundary so there's no reason to keep looking in the iterator until `Next` is called again. I agree the comment here is a bit misleading/crap though, will fix it up

src/query/storage/unconsolidated_test.go
benraskin92: cool, good to know

src/query/storage/validator/storage.go
robskillington: Ah interesting.  Thx for the detailed response.

src/query/storage/unconsolidated_test.go
```diff
@@ -0,0 +1,155 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
robskillington: nit: 2019

src/query/storage/validator/storage.go
```diff
@@ -115,17 +116,29 @@ func PromResultToSeriesList(promReadResp prometheus.PromResp, tagOptions models.
...
 		}
 
+		metricName := string(tagOptions.MetricName())
 		tags := models.NewTags(len(result.Metric), tagOptions)
 		for name, val := range result.Metric {
-			tags = tags.AddTag(models.Tag{
-				Name:  []byte(name),
-				Value: []byte(val),
-			})
+			if name == metricName {
```
benraskin92: Thanks for fixing!

src/query/ts/m3db/convert.go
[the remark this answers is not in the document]
benraskin92: +1

src/query/ts/m3db/convert.go
```diff
@@ -174,6 +204,47 @@ func blockReplicasFromSeriesIterator(
...
+			numberToFill++
+		}
+
+		iterStart = iterStart.Add(stepSize * time.Duration(numberToFill))
+	}
+
+	// Update read starts for existing blocks.
+	for i, bl := range blocks {
+		blocks[i].readStart = iterStart
+		fillSize := bl.blockStart.Add(bl.blockSize).Sub(iterStart)
```
benraskin92: Pull some of this common stuff out into a function

PR #1272, opened 2019-01-01
Regen copyrights for 2019

justinjc:
Happy New Year!


robskillington:
Hey sorry - we should investigate why this fix isn't being applied to the files in question:
https://github.com/m3db/tools/commit/0eb8069d49114a3721ebd94ede7c653efadf6a46

The license gen should skip year mismatches when regen'ing code.

justinjc:
The license gen skip doesn't work because when files get regenerated (e.g. mockgen recreating a mock file), the new files won't have a license on them, to which license gen will happily place the copyright blurb with the new year on it.

Here's an attempt to not have to deal with copyright year changes: https://github.com/m3db/m3/pull/1275

PR #1275, opened 2019-01-02
Ignore copyright year changes in gen'd files

justinjc:


scripts/auto-gen-helpers.sh
```diff
@@ -21,9 +21,26 @@ remove_matching_files() {
     find $DIR -type f -name "$FILE_PATTERN" -exec rm -f {} \;
 }
 
+revert_copyright_only_change() {
+    # We don't want to make a change to a file if the only change
+    # is the copyright year. We can't check this in add_license because a newly-
+    # generated file will not contain the copyright notice and thus it will
+    # add in the copyright (with the new year).
+    local FILE=$0
+    numDiffLines=$(git diff --text -U0 $FILE | # Get file text diffs with no context.
```
justinjc: Hoping some scripting wizard can come up with something cleaner here...

src/dbnode/client/fetch_state_leakcheckpool_gen_test.go
```diff
@@ -26,11 +26,8 @@ package client
 
 import (
 	"fmt"
-
```
justinjc: These three `*_gen_test.go` files should have always had these extra blank lines removed during generation, but were not because the command `gen_cleanup_dir '*_gen.go'` have always omitted them.

PR #1276, opened 2019-01-02
[query] Refactor encoded block conversion to use options

arnikola:
Refactors the list of args taken in by encoded block conversion to avoid long argument lists and consolidate all option permutations into one object.

### DIFF
--- a/src/query/ts/m3db/encoded_block_iterator_builder.go
+++ b/src/query/ts/m3db/encoded_block_iterator_builder.go
@@ -21,6 +21,8 @@
 package m3db
 
 import (
+	"time"
+
 	"github.com/m3db/m3/src/dbnode/encoding"
 	"github.com/m3db/m3/src/query/block"
 	"github.com/m3db/m3/src/query/models"
@@ -59,10 +61,13 @@ func (b *encodedBlockBuilder) add(
 	str := start.String()
 	if _, found := b.blocksAtTime[str]; !found {
 		// Add a new encoded block
+		// NB: the lookback will always be 0, as splitting series into blocks is not
+		// supported with positive lookback durations.
 		b.blocksAtTime[str] = newEncodedBlock(
 			[]encoding.SeriesIterator{},
 			b.tagOptions,
 			consolidation,
+			time.Duration(0),
 			lastBlock,
 		)
 	}
--- a/src/query/ts/m3db/options.go
+++ b/src/query/ts/m3db/options.go
@@ -1,0 +1,140 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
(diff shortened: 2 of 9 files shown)


src/query/ts/m3db/types.go
```diff
@@ -0,0 +1,60 @@
...
+	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/ts/m3db/consolidators"
+)
+
+// Options describes the options for encoded block converters.
+type Options interface {
+	// SetLookbackDuration sets the lookback duration.
+	SetLookbackDuration(time.Duration) Options
+	// GetLookbackDuration returns the lookback duration.
+	GetLookbackDuration() time.Duration
```
benraskin92: I would make these just `LookbackDuration()` (i.e. without the `Get` prefix) - I think that's Go standard convention

src/query/ts/m3db/options.go
```diff
@@ -0,0 +1,139 @@
...
+	opts.lookbackDuration = lookback
+	return &opts
+}
+
+func (o *encodedBlockOptions) GetLookbackDuration() time.Duration {
+	return o.lookbackDuration
+}
+
+func (o *encodedBlockOptions) SetConsolidationFunc(
+	fn consolidators.ConsolidationFunc,
```
benraskin92: Can you put these on one line? We should come up with a convention..

benraskin92: And the other places where you only have one argument.

src/query/ts/m3db/options.go
```diff
@@ -0,0 +1,139 @@
...
+	opts := *o
+	opts.pools = p
+	return &opts
+}
+
+func (o *encodedBlockOptions) GetIteratorPools() encoding.IteratorPools {
+	return o.pools
+}
+
+func (o *encodedBlockOptions) Validate() error {
```
benraskin92: Check `LookbackDuration > 0`

arnikola: Good call; I'll put the check on `SetLookbackDuration` and limit it to >= 0 instead

src/query/ts/m3db/types.go
arnikola: Good call, I swear I saw `Get..` in an existing `Options` so was going by that convention, but will update.

src/query/ts/m3db/options.go
arnikola: Was going by line length mostly, put it  on a single line now

Should get a standard convention at some point though

src/query/ts/m3db/encoded_block_iterator_builder.go
```diff
@@ -59,10 +61,13 @@ func (b *encodedBlockBuilder) add(
 	str := start.String()
 	if _, found := b.blocksAtTime[str]; !found {
 		// Add a new encoded block
+		// NB: the lookback will always be 0, as splitting series into blocks is not
```
benraskin92: In the future, we plan to support this, right?

src/query/ts/m3db/options.go
```diff
@@ -0,0 +1,156 @@
...
+		return false
+	}
+
+	return o.splitSeries
+}
+
+func (o *encodedBlockOptions) SetLookbackDuration(lookback time.Duration) Options {
+	// NB: lookback cannot be negative.
+	if lookback < 0 {
+		lookback = 0
```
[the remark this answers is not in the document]
arnikola: Yeah fair enough, looks like that's how other Options do it

src/query/ts/m3db/encoded_block_iterator_builder.go
arnikola: Potentially; it may not be worth dealing with the added complexity in this case.

PR #1283, opened 2019-01-04
[query] splits series into blocks, filling in missing blocks

arnikola:
Adds tests and fills in blocks in sparse series with empty iterators, allowing downstream functions to work correctly.

### DIFF
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -157,16 +157,7 @@ func parseDebugFlag(r *http.Request) bool {
 	return debug
 }
 
-var (
-	validBlockTypes = []models.FetchedBlockType{
-		models.TypeSingleBlock,
-		models.TypeMultiBlock,
-		models.TypeDecodedBlock,
-	}
-)
-
 func parseBlockType(r *http.Request) models.FetchedBlockType {
-	var parseBlockType models.FetchedBlockType
 	// Use default block type if unable to parse blockTypeParam.
 	useLegacyVal := r.FormValue(blockTypeParam)
 	if useLegacyVal != "" {
@@ -175,15 +166,16 @@ func parseBlockType(r *http.Request) models.FetchedBlockType {
 			logging.WithContext(r.Context()).Warn("unable to parse useLegacy flag", zap.Error(err))
 		}
 
-		// If invalid value, return default.
-		if intVal < 0 || int(intVal) >= len(validBlockTypes) {
+		blockType := models.FetchedBlockType(intVal)
+		// Ignore error from receiving an invalid block type, and return default.
+		if blockType.Validate() != nil {
 			return models.TypeSingleBlock
 		}
 
-		return validBlockTypes[intVal]
+		return blockType
 	}
 
-	return parseBlockType
+	return models.TypeSingleBlock
 }
 
 // parseInstantaneousParams parses all params from the GET request
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -49,11 +49,7 @@ const (
 	stepParam         = "step"
 	debugParam        = "debug"
 	endExclusiveParam = "end-exclusive"
-
-	// NB: this will force series decompression on fetch. Will be left for a few
-	// releases to help debug any issues that may arise with compressed series
-	// block fetch
-	useLegacyParam = "legacy"
+	blockTypeParam    = "block-type"
 
 	formatErrStr = "error parsing param: %s, error: %v"
 )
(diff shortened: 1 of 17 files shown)


src/query/models/params.go
```diff
@@ -39,6 +39,25 @@ const (
...
+// transformed from storage type.
+type FetchedBlockType uint8
+
+const (
+	// TypeSingleBlock represents a single block which contains each encoded fetched
+	// series. Default block type for Prometheus queries.
+	TypeSingleBlock FetchedBlockType = iota
+	// TypeMultiBlock represents multiple blocks, each containing a time-based slice
+	// of encoded fetched series. Default block type for Prometheus queries.
+	TypeMultiBlock
```
nikunjgit: both are defaults ?

src/query/storage/m3/storage.go
```diff
@@ -105,13 +105,20 @@ func (s *m3storage) FetchBlocks(
 	query *storage.FetchQuery,
 	options *storage.FetchOptions,
 ) (block.Result, error) {
-	if options.UseLegacy {
+	opts := s.opts
+	switch options.BlockType {
```
nikunjgit: this flow is a little confusing since in only one case you return and others you don't

nikunjgit: one way to avoid that could be to move everything after the switch into another function and then call that both from models.TypeMultiBlock and also a default case

src/query/ts/m3db/encoded_block_iterator_builder.go
```diff
@@ -58,31 +70,143 @@ func (b *encodedBlockBuilder) add(
...
-			consolidation,
-			time.Duration(0),
-			lastBlock,
-		)
+	for idx, bl := range b.blocksAtTime {
+		if bl.time.Equal(start) {
+			block := bl.block
+			block.seriesBlockIterators = append(block.seriesBlockIterators, iter)
+			b.blocksAtTime[idx].block = block
+
```
nikunjgit: remove line ?

src/query/ts/m3db/encoded_block_iterator_builder.go
```diff
@@ -58,31 +70,143 @@ func (b *encodedBlockBuilder) add(
...
+			}, nil)
+
+			block := bl.block
+			block.seriesBlockIterators = append(block.seriesBlockIterators, iter)
+			b.blocksAtTime[blockIdx].block = block
+		}
+	}
+
+	for _, bl := range b.blocksAtTime {
+		sort.Sort(seriesIteratorByID(bl.block.seriesBlockIterators))
```
nikunjgit: move this out ? I assume you have to do this independent of backfill ?

src/query/models/params.go
arnikola: Fixed comment 👍

src/query/storage/m3/storage.go
arnikola: Switched it to `if` to make it a bit clearer, you're right that there's not a lot of point on a `switch` there

src/query/ts/m3db/encoded_block_iterator_builder.go
arnikola: 👍

src/query/ts/m3db/encoded_block_iterator_builder.go
arnikola: Sounds good. Was originally intending to remove the rest of the backfill logic and just do the sort here, but makes more sense separately.

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -157,16 +154,36 @@ func parseDebugAndLegacyFlags(r *http.Request) (bool, bool) {
...
+	useLegacyVal := r.FormValue(blockTypeParam)
 	if useLegacyVal != "" {
-		useLegacy, err = strconv.ParseBool(r.FormValue(useLegacyParam))
+		intVal, err := strconv.ParseInt(r.FormValue(blockTypeParam), 10, 8)
 		if err != nil {
 			logging.WithContext(r.Context()).Warn("unable to parse useLegacy flag", zap.Error(err))
 		}
+
+		// If invalid value, return default.
+		if intVal < 0 || int(intVal) >= len(validBlockTypes) {
```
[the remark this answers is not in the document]
arnikola: :+1: Refactored it

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -157,16 +154,36 @@ func parseDebugAndLegacyFlags(r *http.Request) (bool, bool) {
 		}
 	}
 
-	// Skip useLegacy if unable to parse useLegacy param
-	useLegacyVal := r.FormValue(useLegacyParam)
+	return debug
+}
+
+var (
+	validBlockTypes = []models.FetchedBlockType{
```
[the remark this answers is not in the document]
arnikola: :+1: Refactored it

PR #1285, opened 2019-01-04
Refix flaky test.

andrewmains12:
The previous version of this test contained a race condition where the select on <-ctx.Done() was hit before context cancellation, causing the request to go through.

```
if f.wait != nil {
    <-f.wait
}

select {
case <-ctx.Done():
	return ctx.Err()
default:
	f.processed = true
	return nil
}
```

Sequence of events in the happy path:

```
W = worker, C = coordinator thread

W Receive on f.wait
C ctx.Cancel()
W select ctx.Done(), default
W Receive ctx.Done()
```

Failing path:

```
W Receive on f.wait
W select ctx.Done, default
W Hit default (set processed = true, causing failure)
C ctx.Cancel()
```

Fix is to just unconditionally wait for the context to be cancelled, and assert that it was in fact cancelled (as opposed to some timeout or something).

Running tests with -test.count 1000 repros the race consistently on master; it doesn't show up on the branch.

### DIFF
--- a/src/query/util/execution/parallel_test.go
+++ b/src/query/util/execution/parallel_test.go
@@ -25,6 +25,7 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/fortytw2/leaktest"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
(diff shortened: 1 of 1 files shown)


src/query/util/execution/parallel_test.go
```diff
@@ -78,13 +79,34 @@ func TestOrderedParallel(t *testing.T) {
...
 	requests := make([]Request, 3)
 	signalChan := make(chan bool)
-	requests[0] = &request{order: 0, wait: signalChan}
+
+	var cancelErr error
+	requests[0] = funcRequest(func(ctx context.Context) error {
+		// wait for the second goroutine to finish
+		<-signalChan
+
+		// wait for cancellation. This will hang if there's a bug here (i.e. the context doesn't get cancelled)
```
arnikola: Any way we can have it detect this case without waiting for the timeout to kill test execution?

src/query/util/execution/parallel_test.go
```diff
@@ -78,13 +79,34 @@ func TestOrderedParallel(t *testing.T) {
 }
 
 func TestSingleError(t *testing.T) {
+	defer leaktest.Check(t)()
```
arnikola: nit: Is this necessary? I think CI runs with `--race` but I may be mistaken as to what this does

src/query/util/execution/parallel_test.go
andrewmains12: Yeah we can add an explicit timeout here, but I don't know that we need to. `leaktest` will fail the test if we do hang. I'll add a comment explaining as much.

src/query/util/execution/parallel_test.go
andrewmains12: Ah yeah, this is serving a different purpose--it checks for goroutines that leak after the end of the test. For instance (from their README):

// Default "Check" will poll for 5 seconds to check that all
// goroutines are cleaned up
func TestPool(t *testing.T) {
    defer leaktest.Check(t)()

    go func() {
        for {
            time.Sleep(time.Second)
        }
    }()
}

PR #1286, opened 2019-01-05
[query] Change Tag ID generation to avoid possible collisions

arnikola:
Changes Tags.ID() method to return a byte slice representing the ID rather than allocating strings.


src/query/models/tags.go
```diff
@@ -50,40 +49,23 @@ func EmptyTags() Tags {
...
-	}
-
-	return strBuilder.String()
-}
-
-// IDMarshalTo writes out the ID representation
-// of the tags into the provided buffer.
-func (t Tags) IDMarshalTo(b []byte) []byte {
+// ID returns a byte slice representation of the tags.
+func (t Tags) ID() []byte {
```
robskillington: Can we also fix this ID so that it doesn't generate potentially colliding metric IDs?

i.e.
```
series 1, tags={"set_values":"something=foo,someother=bar"}
# generates this ID:
set_values=something=foo,someother=bar,

series 2, tags={"set_values":"something=foo","someother":"bar"}
# generates same ID:
set_values=something=foo,someother=bar,
```

In Prometheus they use `strconv.Quote(...)` to avoid this:
https://github.com/prometheus/tsdb/blob/master/labels/labels.go#L43-L58

You could refactor this method from the Go std library to use just []byte as well:
https://golang.org/src/strconv/quote.go

There's some sub-functions but it wouldn't be too difficult to do this properly by using a slightly modified version of it:
```go
func appendQuotedWith(buf []byte, s string, quote byte, ASCIIonly, graphicOnly bool) []byte {
	buf = append(buf, quote)
	for width := 0; len(s) > 0; s = s[width:] {
		r := rune(s[0])
		width = 1
		if r >= utf8.RuneSelf {
			r, width = utf8.DecodeRuneInString(s)
		}
		if width == 1 && r == utf8.RuneError {
			buf = append(buf, `\x`...)
			buf = append(buf, lowerhex[s[0]>>4])
			buf = append(buf, lowerhex[s[0]&0xF])
			continue
		}
		buf = appendEscapedRune(buf, r, quote, ASCIIonly, graphicOnly)
	}
	buf = append(buf, quote)
	return buf
}
```

arnikola: Good call. Would you recommend going this way or maybe doing something like pre-processing 
tag lengths and then appending it to the start of the ID, so it would look like: 

`series 1`
"set_values":"something=foo,someother=bar" -> `0.set_values:something=foo,someother=bar`

`series 2`
"set_values":"something=foo","someother=bar" -> `0,24.set_values:something=foosomeother:bar`

src/query/models/strconv/int_printer.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2018 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
```
arnikola: Not sure how this got detected as a moved file, gut I deleted the resolver and added this file instead

andrewmains12:
Nice! Mind posting some benchmark results to get a sense of the improvement here?

src/query/models/strconv/checker.go
```diff
@@ -0,0 +1,63 @@
...
+	tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, // 0xB0-0xBF
+	tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, // 0xC0-0xCF
+	tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, // 0xD0-0xDF
+	tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, // 0xE0-0xEF
+	tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, tt, // 0xF0-0xFF
+}
+
+// NeedToEscape returns true if the byte slice contains characters that will
+// need to be escaped when quoting the slice.
+func NeedToEscape(bb []byte) bool {
```
andrewmains12: Do these `*Escape` functions come from `strconv`? If not, should we move them to a separate package? It'd be easier to reason about this package given a clear separation of our modifications and copied code.

src/query/models/tags.go
```diff
@@ -93,37 +88,118 @@ func (t Tags) IDLen() int {
...
+func (t Tags) quotedID() []byte {
+	needEscaping, length := t.escapingAndLength()
+	if needEscaping == nil {
+		return t.quoteIDSimple(length)
+	}
+
+	// TODO: pool these bytes.
+	id := make([]byte, length)
+	idx := 0
+	for i, tag := range t.Tags {
```
andrewmains12: A potential readability refactor for your consideration: could we build up this function in terms of a method to write out a single tag to a `[]byte`? Something like:

```
length := 0
for _, t :=` range t.Tags {
    sliceLen += t.SerializedLength()  // this would take into account NeedsEscaping, etc
}

id := make([]byte, length)
for _, t := range t.Tags {
    t.WriteTo(id)
}
```

src/query/models/tags.go
```diff
@@ -93,37 +88,118 @@ func (t Tags) IDLen() int {
...
 
-		found := false
-		for _, n := range excludeKeys {
-			if bytes.Equal(n, tag.Name) {
-				found = true
-				break
+		if needEscaping[i*2+1] {
+			idx = strconv.Quote(id, tag.Value, idx)
+		} else {
+			idx = strconv.QuoteSimple(id, tag.Value, idx)
```
andrewmains12: How much benefit does having `Quote` and `QuoteSimple` give you? Code would definitely be simpler without the distinction (though I could see how this would be faster).

src/query/models/strconv/quote.go
```diff
@@ -0,0 +1,293 @@
...
+	dst[idx] = quote
+	idx++
+	idx += copy(dst[idx:], src)
+	dst[idx] = quote
+	return idx + 1
+}
+
+// EscapedLength computes the length required for a byte slice to hold
+// a quoted byte slice.
+func EscapedLength(src []byte) int {
```
andrewmains12: Is this copied code or new?

src/query/models/tags.go
```diff
@@ -93,37 +88,118 @@ func (t Tags) IDLen() int {
...
+	idx := 0
+	for _, tag := range t.Tags {
+		idx += copy(id[idx:], tag.Name)
+		idx = strconv.QuoteSimple(id, tag.Value, idx)
+	}
+
+	return id
+}
+
+func (t Tags) escapingAndLength() ([]bool, int) {
```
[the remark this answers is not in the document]
andrewmains12: One potential way to make this signature more clear without compromising perf (I think): 

```
type tagEscaping struct {
    EscapeName bool
    EscapeValue bool
}

func (t Tags) escapingAndLength() ([]tagEscaping, int) {
...
```
https://play.golang.org/p/ctnDxYvBxyK shows that the size of the two is the same (i.e. an array of 2 of the struct takes up the same space as 4 bools)

src/query/models/tags.go
arnikola: It's a really large difference;

```
10_tags,_2_len_easy-8         	 5000000	       328 ns/op	      64 B/op	       1 allocs/op
10_tags,_2_len_hard-8         	 3000000	       583 ns/op	     112 B/op	       2 allocs/op

100_tags,_2_len_easy-8        	  500000	      3207 ns/op	     704 B/op	       1 allocs/op
100_tags,_2_len_hard-8        	  300000	      5511 ns/op	    1104 B/op	       2 allocs/op

1000_tags,_2_len_easy-8       	   50000	     31613 ns/op	    8194 B/op	       1 allocs/op
1000_tags,_2_len_hard-8       	   30000	     59363 ns/op	   12292 B/op	       2 allocs/op

10_tags,_10_len_easy-8        	 3000000	       397 ns/op	     144 B/op	       1 allocs/op
10_tags,_10_len_hard-8        	 1000000	      1257 ns/op	     272 B/op	       2 allocs/op

100_tags,_10_len_easy-8     		  500000	      4007 ns/op	    1536 B/op	       1 allocs/op
100_tags,_10_len_hard-8       	  100000	     13479 ns/op	    2896 B/op	       2 allocs/op

1000_tags,_10_len_easy-8      	   30000	     41068 ns/op	   16388 B/op	       1 allocs/op
1000_tags,_10_len_hard-8      	   10000	    155339 ns/op	   29325 B/op	       2 allocs/op

10_tags,_100_len_easy-8       	 1000000	      1176 ns/op	    1152 B/op	       1 allocs/op
10_tags,_100_len_hard-8       	  200000	      9091 ns/op	    2080 B/op	       2 allocs/op

100_tags,_100_len_easy-8      	  200000	     10434 ns/op	   10880 B/op	       1 allocs/op
100_tags,_100_len_hard-8      	   20000	     90329 ns/op	   21969 B/op	       2 allocs/op

1000_tags,_100_len_easy-8     	   10000	    108446 ns/op	  106519 B/op	       1 allocs/op
1000_tags,_100_len_hard-8     	    2000	    908084 ns/op	  215157 B/op	       4 allocs/op

10_tags,_1000_len_easy-8      	  200000	      6704 ns/op	   10240 B/op	       1 allocs/op
10_tags,_1000_len_hard-8      	   20000	     89744 ns/op	   20512 B/op	       2 allocs/op

100_tags,_1000_len_easy-8     	   20000	     63663 ns/op	  106501 B/op	       1 allocs/op
100_tags,_1000_len_hard-8     	    2000	    850395 ns/op	  205065 B/op	       2 allocs/op

1000_tags,_1000_len_easy-8    	    2000	    650311 ns/op	 1008189 B/op	       3 allocs/op
1000_tags,_1000_len_hard-8    	     200	   8405176 ns/op	 2014826 B/op	      22 allocs/op
```

For most likely cases, there's a 3-10x increase, and this is just with quote characters; if using any long unicode this difference would get even worse.

andrewmains12: Ah ok, good to know. Which benchmark are these from? Do they compare preprocessing to determine which `Quote` method is needed followed by `Quote`/`QuoteSimple` to just running `Quote`?

arnikola:
Benchmark results for relevant tag number/lengths:

```
10__T,_10__L_legacy_str_-8           	 5000000	       383 ns/op	     224 B/op	       2 allocs/op
10__T,_10__L_legacy_byte-8         	10000000	       225 ns/op	     144 B/op	       1 allocs/op
10__T,_10__L_prepended__-8         	 5000000	       382 ns/op	     224 B/op	       2 allocs/op
10__T,_10__L_quoted_easy-8         	 5000000	       353 ns/op	     144 B/op	       1 allocs/op
10__T,_10__L_quoted_hard-8         	 1000000	      1230 ns/op	     272 B/op	       2 allocs/op

100_T,_10__L_legacy_str_-8         	  500000	             3322 ns/op	    2432 B/op	       2 allocs/op
100_T,_10__L_legacy_byte-8         	 1000000	      1993 ns/op	    1536 B/op	       1 allocs/op
100_T,_10__L_prepended__-8         	  500000	      3303 ns/op	    2432 B/op	       2 allocs/op
100_T,_10__L_quoted_easy-8         	  500000	      3618 ns/op	    1536 B/op	       1 allocs/op
100_T,_10__L_quoted_hard-8         	  200000	     11919 ns/op	    2896 B/op	       2 allocs/op

10__T,_100_L_legacy_str_-8         	 2000000	       635 ns/op	    1232 B/op	       2 allocs/op
10__T,_100_L_legacy_byte-8         	 3000000	       442 ns/op	    1152 B/op	       1 allocs/op
10__T,_100_L_prepended__-8         	 2000000	       628 ns/op	    1232 B/op	       2 allocs/op
10__T,_100_L_quoted_easy-8         	 1000000	      1036 ns/op	    1152 B/op	       1 allocs/op
10__T,_100_L_quoted_hard-8         	  200000	      9033 ns/op	    2080 B/op	       2 allocs/op

100_T,_100_L_legacy_str_-8         	  300000	      5474 ns/op	   11776 B/op	       2 allocs/op
100_T,_100_L_legacy_byte-8         	  500000	      3690 ns/op	   10880 B/op	       1 allocs/op
100_T,_100_L_prepended__-8         	  300000	      5454 ns/op	   11776 B/op	       2 allocs/op
100_T,_100_L_quoted_easy-8         	  200000	     10157 ns/op	   10880 B/op	       1 allocs/op
100_T,_100_L_quoted_hard-8         	   20000	             92368 ns/op	   21969 B/op	       2 allocs/op
```

Basically, the prepended byte is similar in performance to the original `string` version, and the byte version of the original is the quickest (but also prone to collisions). The prepended version is the fastest, but it's currently not reversible, and is not extremely human readable. Up to you guys which should be default, it's currently `quoted`.

src/query/models/strconv/quote.go
```diff
@@ -0,0 +1,293 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package strconv
```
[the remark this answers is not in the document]
arnikola: Yeah, this refactors the `strconv` `quote.go` approach in two main ways;
- changes string functions to bytes
- works on a pre-set byte slice and inserts at indices rather than working on `append`

The ID function is called on each incoming write, so it's fairly hot.

src/query/models/tags.go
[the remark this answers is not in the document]
andrewmains12: Cool makes sense. Actually now that I think about it, what if we just moved the `Quote` vs simple quote decision and actual quoting into a single method? That is, make a method which: 

1. Checks the string for characters needing escaping + calculate the length
2. Allocates the appropriate byte array.
3. Calls `Quote` or `QuoteSimple` as appropriate

Perf wise I think it would be equivalent, and it would abstract away some of the details from `Tags`, allowing this class to just call `Quote`, which does the right thing and is performant. 

Thoughts?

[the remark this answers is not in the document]
arnikola: Unfortunately we need to calculate the length for the whole tag set, otherwise we're allocating a bunch of temporary buffers per-tag, then putting them together. If `Tags` handles this itself, it can determine the total buffer size for the whole ID

src/query/functions/aggregation/take.go
```diff
@@ -68,9 +68,10 @@ func NewTakeOp(
 
 // takeOp stores required properties for take ops
 type takeOp struct {
-	params   NodeParams
-	opType   string
-	takeFunc takeFunc
+	params      NodeParams
+	opType      string
+	opTypeBytes []byte
```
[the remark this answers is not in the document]
arnikola: Ah damn, thought I'd reverted these, good catch.

src/query/models/tags.go
```diff
@@ -93,37 +88,118 @@ func (t Tags) IDLen() int {
...
+		idx += copy(id[idx:], tag.Value)
 	}
 
-	h := fnv.New64a()
-	h.Write(b)
-	return h.Sum64()
+	return id
+}
+
+func prependTagLengthMeta(dst []byte, lengths []int) int {
```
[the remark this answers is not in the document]
arnikola: Sure, sounds good. Was thinking `prefix` or something as well, but will probably go with `write`

src/query/models/strconv/checker.go
arnikola: It's a super cut down version of `utf8.DecodeRune`

src/query/models/tags.go
arnikola: This is benchmarking the entire `ID()` method; `easy` has just valid characters, whereas `hard` has just `"`. The problem is that `utf8.DecodeRune` is quite a slow operation, which is called once per byte, and incurs a fair cost.

src/query/models/tags.go
```diff
@@ -93,37 +88,118 @@ func (t Tags) IDLen() int {
...
+	idx := 0
+	for _, l := range lengths {
+		idx = strconv.WriteInteger(dst, l, idx)
+	}
+
+	return idx
+}
+
+// idLen returns the length of the ID that would be generated from the tags.
```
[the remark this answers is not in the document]
arnikola: Good catch

andrewmains12:
Broader question about the schemes--which of the schemes are we recommending people to use? Seems like the choice has significant implications, since it changes the data that gets written (I think). Is this something that we actually need to make a config choice? We definitely need to fix the current scheme, but it seems like we should maybe just pick a correct new format and roll with that unless there are certain scenarios under which one scheme would be significantly better than the other. Thoughts?

vdarulis:
I may be just playing the devil's advocate - but why not just use a simple binary encoding (uint16 len+raw bytes per field) instead, and just keep the quoting for debugging/human-readable representation?

vdarulis:
> Unless I'm misunderstanding you, that seems pretty close to the TypePrependMeta scheme, other than it using written out ints rather than bytes (Rob was a bit iffy about writing bytes representing weird control characters to the DB)
Yep, just that just without the int->str conversion.

robskillington:
@vdarulis so this is so you can use the APIs (especially the human readable ones like HTTP+JSON) without getting garbaged binary in your terminal.

It's also important to have readable IDs for a ton of other reasons too.

src/query/models/config.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+)
+
+var validIDSchemes = []IDSchemeType{
+	TypeLegacy,
+	TypeQuoted,
+	TypePrependMeta,
+}
+
+func (t IDSchemeType) validateIDSchemeType() error {
```
robskillington: Maybe worth just make this an exported method?  i.e. `Validate() error` so that if we use it outside of YAML parsing somewhere we can call validate on it?

vdarulis:
> @vdarulis so this is so you can use the APIs (especially the human readable ones like HTTP+JSON) without getting garbage binary in your terminal.
> 
> It's also important to have readable IDs for a ton of other reasons too.

Makes sense, JSON is nasty to deal with if you have binary data.

src/query/models/config.go
```diff
@@ -0,0 +1,79 @@
...
+func (t IDSchemeType) String() string {
+	switch t {
+	case TypeDefault:
+		return ""
+	case TypeLegacy:
+		return "legacy"
+	case TypeQuoted:
+		return "quoted"
+	case TypePrependMeta:
+		return "prependMeta"
```
robskillington: Maybe make this underscore separated, to match our other `String()` methods on enum like values?

For example: 
https://github.com/m3db/m3/blob/master/src/dbnode/storage/series/policy.go#L60-L72

src/query/models/config.go
arnikola: Sure, sounds good.

More of a general style question, but is it better to leave functions unexported initially, then export when needed, or to have "expected" behaviour like `Validate()` public even if it's not used anywhere? I guess leaving it private can lead to accidentally rewriting functionality considering how go has functions floating around everywhere in a package :p

src/query/models/config.go
arnikola: sgtm :+1: should've checked for existing conventions

andrewmains12:
> @vdarulis so this is so you can use the APIs (especially the human readable ones like HTTP+JSON) without getting garbage binary in your terminal.
> 
> It's also important to have readable IDs for a ton of other reasons too.

Makes sense to me. @arnikola  If readability is one of the goals, which scheme should we go for? Seems like `quoted` is probably going to be more readable overall than `prependMeta`, but I could be wrong. 

cc. @robskillington @vdarulis

src/query/models/tags_test.go
```diff
@@ -242,3 +288,133 @@ func TestTagAppend(t *testing.T) {
...
+	name                string
+	tagCount, tagLength int
+}{
+	{"10  Tags 10  Length", 10, 10},
+	{"100 Tags 10  Length", 100, 10},
+	{"10  Tags 100 Length", 10, 100},
+	{"100 Tags 100 Length", 100, 100},
+}
+
+const typeQuotedEscaped = IDSchemeType(100)
```
andrewmains12: What's with the magic number here?

arnikola: Just a big number to use as an "invalid" id scheme type

src/query/models/tags.go
```diff
@@ -193,29 +193,24 @@ func (t Tags) prependMetaID() []byte {
...
+	idx := writer.WriteIntegers(dst, lengths, sep, 0)
+	dst[idx] = finish
+	return idx + 1
 }
 
 func (t Tags) prependMetaLen() (int, []int) {
-	idLen := 0
-	tagLengths := make([]int, len(t.Tags))
+	idLen := 1 // account for separator
+	tagLengths := make([]int, len(t.Tags)*2)
```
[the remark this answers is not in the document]
arnikola: Would rather just have a float array here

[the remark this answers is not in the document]
andrewmains12: Floats? But sure, this is ok as is.

src/cmd/services/m3query/config/config.go
```diff
@@ -199,9 +199,16 @@ type RPCConfiguration struct {
 // Currently only name, but can expand to cover deduplication settings, or other
 // relevant options.
 type TagOptionsConfiguration struct {
+	// Version specifies the version number of tag options. Defaults to 0.
```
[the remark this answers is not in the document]
robskillington: Agreed, we will have a point in time where we do based on a "YAML" config version change which type (i.e. we'll use quoted by default), but we don't have that mechanism right now.

src/query/models/tags.go
```diff
@@ -93,37 +91,177 @@ func (t Tags) IDLen() int {
...
-	h.Write(b)
-	return h.Sum64()
+	// TODO: pool these bytes
+	lastIndex := len(t.Tags) - 1
+	id := make([]byte, idLen)
+	idx := 0
+	for i, tt := range t.Tags[:lastIndex] {
+		idx = tt.writeAtIndex(id, needEscaping[i], idx)
+		id[idx] = sep
+		fmt.Println(string(id))
```
robskillington: Remove printlns?

src/query/models/tags.go
```diff
@@ -93,37 +91,177 @@ func (t Tags) IDLen() int {
...
+		idx++
+	}
+	fmt.Println(string(id))
```
robskillington: Remove printlns?

src/query/models/tags.go
```diff
@@ -93,37 +91,177 @@ func (t Tags) IDLen() int {
...
+
+	t.Tags[lastIndex].writeAtIndex(id, needEscaping[lastIndex], idx)
+	fmt.Println(string(id))
```
robskillington: Remove printlns?

src/query/models/tags.go
arnikola: Big oops

src/query/models/tags_test.go
```diff
@@ -22,78 +22,136 @@ package models
...
-	var (
-		tags = createTags(false)
-		b    = tags.IDMarshalTo([]byte{})
-	)
-	assert.Equal(t, []byte("t1=v1,t2=v2,"), b)
-	assert.Equal(t, tags.IDLen(), len(b))
+func TestLongTagNewIDOutOfOrderQuoted(t *testing.T) {
+	tags := testLongTagIDOutOfOrder(t, TypeQuoted)
+	actual := tags.ID()
+	assert.Equal(t, []byte(`t1="v1",t2="v2",t3="v3",t4="v4"`), actual)
```
robskillington: nit: Just as discussed, to exact match the Prometheus ID construction, adding the `{` and `}` should get us 100% matching.

robskillington: I think the only reason we wouldn't want to do that, is if M3QL ends up using the same ID format. Which it probably should TBH.

arnikola: Should be fixed 👍

PR #1287, opened 2019-01-09
Use IsBootstrappedAndDurable instead of IsBootstrapped in health checks

richardartoul:
The existing health checks use IsBootstrapped() to determine if a node is healthy and bootstrapped which is fine in the normal case, but in the situation where a topology change has occurred this is problematic.

The reason its problematic in that scenario is that after a topology change, as a result of #1183 the node will not mark its shards as available until a snapshot has occurred. This is an issue for our automated tooling which will assume that once a node is "healthy" it can proceed with adding more nodes because we don't want the tooling to proceed until the previous node has marked all its shards as available otherwise we can end up in weird shards state (I.E R.F = 3 but we have 2 shard states AVAILABLE, 1 LEAVING, and 2 Initializing and 2/5 will not achieve quorum so writes / reads will begin to fail)

### DIFF
--- a/src/dbnode/storage/cluster/database.go
+++ b/src/dbnode/storage/cluster/database.go
@@ -151,6 +151,49 @@ func (d *clusterDB) Close() error {
 	return d.stopActiveTopologyWatch()
 }
 
+// IsBootstrappedAndDurable determines whether the database is bootstrapped
+// and durable, meaning that it could recover all data in memory using only
+// the local disk.
+//
+// The logic here is a little tricky because there are two levels of
+// IsBootstrappedAndDurable():
+//
+// The first level is this method which exists on the clustered database. It is
+// used by our health check endpoints and tooling in general to determine when
+// it is safe to continue a deploy or performing topology changes. In that case,
+// we only need to determine two things:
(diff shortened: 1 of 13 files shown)


src/dbnode/storage/cluster/database.go
```diff
@@ -151,6 +151,46 @@ func (d *clusterDB) Close() error {
...
+	if !d.Database.IsBootstrapped() {
+		return false
+	}
+
+	_, ok := d.topo.(topology.DynamicTopology)
+	if !ok {
+		// If the topology is not dynamic, then the only thing we care
+		// about is whether the node is bootstrapped or not because the
+		// concept of durability as it relates to shard state doesn't
+		// make sense if we're using a static topology.
```
robskillington: Hm is this true?  Should we perhaps be calling `d.Database.IsBootstrappedAndDurable()` so that at least in the case where the topology is not dynamic it's had at least one snapshot, etc?

richardartoul: spoke offline, existing implementation should be good but I added more context in the comments.

Issue #1274, opened 2019-01-01
Incorrect OpenAPI Endpoint Paths


swtch1:
## General Issue

The OpenAPI documentation references an incorrect path for requests.  Removing `services/m3db/` from the path corrects the issue.

1. What service is experiencing the issue?
OpenAPI Documentation

2. What is the configuration of the service?
K8s installation using [kubernetes guide](https://m3db.github.io/m3/how_to/kubernetes/)

3. How are you using the service?
Early testing

4. Is there a reliable way to reproduce the behavior? If so, please provide detailed instructions.
- Install service on Kubernetes
- Expose service on port 7201
- execute GET on OpenAPI documented endpoint http://127.0.0.1:7201/api/v1/services/m3db/namespace  # FAILURE
- execute GET on k8s guide documented endpoint http://127.0.0.1:7201/api/v1/namespace  # SUCCESS

It appears this path prefix is incorrect across all of the API docs and is likely a simple fix, but definitely confusing when the direct request from the docs doesn't work.

richardartoul:
@swtch1 Thanks for the bug report! I'll take a look at this today.

richardartoul:
Should be fixed by #1292 once it lands

richardartoul:
Resolved

PR #1292, opened 2019-01-11
Change namespace URLs to be scoped by service

richardartoul:
Our documentation (and all of our other APIs) are scoped by service, but namespace APIs weren't scoped by service properly. This P.R scopes them by service, but continues to register the old APIs for backwards compatibility. 

Addresses #1274

### DIFF
--- a/src/query/api/v1/handler/namespace/common.go
+++ b/src/query/api/v1/handler/namespace/common.go
@@ -22,6 +22,7 @@ package namespace
 
 import (
 	"fmt"
+	"path"
 
 	clusterclient "github.com/m3db/m3/src/cluster/client"
 	"github.com/m3db/m3/src/cluster/kv"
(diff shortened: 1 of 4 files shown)


src/query/api/v1/handler/namespace/common.go
```diff
@@ -33,8 +34,22 @@ import (
 )
 
 const (
-	// M3DBNodeNamespacesKey is the KV key that holds namespaces
+	// M3DBServiceName is the service name for M3DB.
+	M3DBServiceName = "m3db"
+
+	// ServicesPathName is the services part of the API path.
+	ServicesPathName = "services"
```
andrewmains12: Nit: I'm not 100% sure these constants add much value; I think it may just make the URL's a bit harder to determine by quick inspection. Maybe worth just leaving them inlined?

richardartoul: Yeah I was just following the pattern (that admitabbly I set up) in this file: https://github.com/m3db/m3/blob/master/src/query/api/v1/handler/placement/common.go

andrewmains12: Fair enough! I don't have too strong an opinion about it, so up to you.

PR #1299, opened 2019-01-16
Use Seek() and Read() APIs for index/data files instead of mmap + share expensive seeker resources among seekers 

richardartoul:


src/dbnode/persist/fs/index_lookup.go
```diff
@@ -41,22 +39,17 @@ type nearestIndexOffsetLookup struct {
...
-	msgpackDecoder *msgpack.Decoder
-	isClone        bool
+	isClone       bool
 }
 
 func newNearestIndexOffsetLookup(
 	summaryIDsOffsets []xmsgpack.IndexSummaryToken,
 	summariesMmap []byte,
-	decoderStream xmsgpack.DecoderStream,
+	decoderStream xmsgpack.ByteDecoderStream,
```
robskillington: Does this get used anymore?  Maybe don't need to pass it in anymore.

src/dbnode/persist/fs/msgpack/decoder.go
```diff
@@ -328,7 +349,11 @@ func (dec *Decoder) decodeIndexEntry() schema.IndexEntry {
 		return indexEntry
 	}
 
-	indexEntry.EncodedTags, _, _ = dec.decodeBytes()
+	if dec.byteReader != nil {
```
robskillington: Hm why does line 337 use `bytesPool == nil` and this if/else use `dec.byteReader != nil`?  Shouldn't it be the same condition?

robskillington: Maybe make this a helper function? Like `func (dec *Decoder) decodeBytesDirect() bool`?  Then you can do:
```go
if dec.decodeBytesDirect() {
  //...
} else {
  //...
}
```

src/dbnode/persist/fs/retriever.go
```diff
@@ -321,28 +325,34 @@ func (r *blockRetriever) fetchBatch(
...
+				decoder.Reset(tags)
 				req.tags = decoder
 			}
+		} else {
+			// If we didn't transfer ownership of the tags to the decoder above, then we
+			// no longer need them and we can can finalize them.
+			if tags := req.indexEntry.EncodedTags; tags != nil {
+				tags.DecRef()
+				tags.Finalize()
+			}
```
robskillington: Regardless of whether we call on retrieve or not, since we're now pooling `req.indexEntry.ID` should we always also `req.indexEntry.ID.Finalize()`?

src/dbnode/persist/fs/seek.go
```diff
@@ -143,32 +143,33 @@ type fileSetSeeker interface {
...
+	blockStart time.Time,
+	resources ReusableSeekerResources,
+) error {
 	if s.isClone {
 		return errClonesShouldNotBeOpened
 	}
 
-	shardDir := ShardDataDirPath(s.filePathPrefix, namespace, shard)
-	var infoFd, indexFd, dataFd, digestFd, bloomFilterFd, summariesFd *os.File
+	shardDir := ShardDataDirPath(s.opts.filePathPrefix, namespace, shard)
```
robskillington: Hm do we still need the local var if right below we're setting `s.shardDir = shardDir`?

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -378,20 +390,21 @@ func (m *seekerManager) newOpenSeeker(
...
 	// Set the unread buffer to reuse it amongst all seekers.
 	seeker.setUnreadBuffer(m.unreadBuf.value)
 
-	if err := seeker.Open(m.namespace, shard, blockStart); err != nil {
+	resources := m.getSeekerResources()
+	if err := seeker.Open(m.namespace, shard, blockStart, m.getSeekerResources()); err != nil {
+		m.putSeekerResources(resources)
 		return nil, err
 	}
+	m.putSeekerResources(resources)
```
robskillington: Are you calling `m.getSeekerResources()` twice here?  Seems like the first one doesn't get used.

Did you want something more like this?:
```go
	resources := m.getSeekerResources()
	err := seeker.Open(m.namespace, shard, blockStart, resources)
	m.putSeekerResources(resources)

	if err != nil {
		return nil, err
	}

```

src/x/panicmon/executor_test.go
```diff
@@ -287,6 +287,7 @@ func TestExecutorPassSignals(t *testing.T) {
 	procInfo.Signal(syscall.SIGUSR1)
 
 	child.Wait()
+	ex.closeAndWait()
```
robskillington: Oh this was flaking?

src/dbnode/persist/fs/index_lookup.go
richardartoul: yeah it's still used to parse the mmap (to generate the summaries slice) and all that so we still have to pass it in, it's just not retained by the struct anymore.

src/dbnode/persist/fs/seek.go
```diff
@@ -383,55 +368,96 @@ func (s *seeker) SeekByIndexEntry(entry IndexEntry) (checked.Bytes, error) {
...
+	resources.xmsgpackDecoder.Reset(resources.fileDecoderStream)
 
 	idBytes := id.Bytes()
-	// Prevent panic's when we're scanning to the end of the buffer
-	for stream.Remaining() != 0 {
-		entry, err := s.decoder.DecodeIndexEntry()
-		// Should never happen, either something is really wrong with the code or
-		// the file on disk was corrupted
+	for {
+		currOffset, err := s.indexFd.Seek(0, os.SEEK_CUR)
```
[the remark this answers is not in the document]
richardartoul: So like 395 is not actually changing the offset, thats actually just how you ask the fd to tell you what the current offset is.

Basically what happens is we:

```
1. Go to the indexLookup thing and it will give us an offset that is a good starting point for scanning
2. Line 382 we seek to the position that the indexLookup told us would be a good start
3. We reset a decoder with our fileDecoderStream (basically a fd wrapped in a bufio.reader)
4. We call entry.DecodeIndexEntry() in a tight loop until we've either: found the entry, gone too far, or reached the end of a file. The further seeks are handled automatically by the fileDecoderStream/bufioReader/fd
```

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -120,13 +125,20 @@ func NewSeekerManager(
 	opts Options,
 	fetchConcurrency int,
 ) DataFileSetSeekerManager {
+	reusableSeekerResourcesPool := make(
+		chan ReusableSeekerResources, reusableSeekerResourcesPoolSize)
+	for i := 0; i < reusableSeekerResourcesPoolSize; i++ {
+		reusableSeekerResourcesPool <- NewReusableSeekerResources(opts)
```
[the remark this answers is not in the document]
richardartoul: Thats...a good point. I'll fix this

src/dbnode/persist/fs/seek_manager.go
richardartoul: yeah that was a mistake thanks

src/x/panicmon/executor_test.go
richardartoul: Yeah its been flaking for a while, but I fixed it in another P.R. Let me rebase and this change will go away

src/dbnode/persist/fs/seek.go
richardartoul: good point

src/dbnode/persist/fs/seek.go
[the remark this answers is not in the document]
richardartoul: Do you want me to add a comment or something with basically what I just wrote above?

src/dbnode/persist/fs/retriever.go
```diff
@@ -561,7 +573,10 @@ func (req *retrieveRequest) onCallerOrRetrieverDone() {
 
 func (req *retrieveRequest) Reset(segment ts.Segment) {
 	req.reader.Reset(segment)
-	req.resultWg.Done()
+	if req.err == nil {
+		// If there was an error, we've already called done.
```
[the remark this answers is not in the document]
richardartoul: Yeah so I didn't change how these methods are called, its just that a bug in my code revealed that in certain error scenarios we would end up with a negative waitgroup. I looked through the code and it wasn't immediately obvious to me what the issue was. I can probably try and look through the code and see exactly how it could happen and fix it? I figured these guards would be good either way

src/dbnode/persist/fs/msgpack/decoder.go
richardartoul: its a remnant of a previous implementation I had where the decoder stored a checked bytes pool and you had to make decisions based on whether or not various interfaces were implemented...

I think for now the right thing to do is to check if the bytes pool is passed, it feels more explicit. If you try and use the decoder the wrong way (pass nil for bytes pool, but also try and use a file instead of a stream backed by a byte slice) the decoder will give you an appropriate error `errorCalledDecodeBytesWithoutByteStreamDecoder`

richardartoul: Basically if the bytes pool is present the caller is signaling that they want to use the implementation that doesn't involve referencing the backing byte slice implementation.

I tried 3-4 different approaches and this is the one that came out the cleanest, but I'm open to suggestions

src/dbnode/persist/fs/retriever.go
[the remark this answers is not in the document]
richardartoul: Actually I'm looking through the code now, are you sure this used to be correct?

We have this bit of code:

```golang
for _, req := range reqs {
		entry, err := seeker.SeekIndexEntry(req.id, seekerResources)
		if err != nil && err != errSeekIDNotFound {
			req.onError(err)
			continue
		}

		if err == errSeekIDNotFound {
			req.notFound = true
		}
		req.indexEntry = entry
	}
```

and later on in the function we do this to each `req` in a loop:

```golang
// Complete request.
req.onRetrieved(seg)
```

The only way `req.onRetrieved(seg)` won't be called is if:

```golang
// Only try to seek the ID if it exists, otherwise we'll get a checksum
// mismatch error because default offset value for indexEntry is zero.
if !req.notFound {
    data, err = seeker.SeekByIndexEntry(req.indexEntry, seekerResources)
        if err != nil && err != errSeekIDNotFound {
            req.onError(err)
            continue
        }
}
```

however `req.notFound = true` doesn't get set in the error case. So I guess we can fix it by changing the code to:

```golang
// Only try to seek the ID if it exists and there haven't been any errors already, otherwise we'll get a checksum mismatch error because default offset value for indexEntry is zero.
if !req.notFound {
    data, err = seeker.SeekByIndexEntry(req.indexEntry, seekerResources)
        if err != nil && err != errSeekIDNotFound {
            req.onError(err)
            continue
        }
}
```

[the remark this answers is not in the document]
richardartoul: I ended up adding a variable: `foundAndHasNoError = !req.notFound && req.err == nil` and tweaking the code slightly, let me know what you think. Might be easiest to discuss in person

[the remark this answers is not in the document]
richardartoul: Ok forget all that, after realizing how buggy the existing implementation was I added some unit tests and just generally refactored it to be correct. Let me know what you think

src/dbnode/persist/fs/index_lookup.go
robskillington: Can we remove the param from this constructor then if it doesn't get set on the struct? (it literally looks like an unused var for this method)

src/dbnode/persist/fs/retriever.go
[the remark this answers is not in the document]
robskillington: Nice, yeah good call - this looks good now. Thanks for adding the tests too, nice work.

src/dbnode/persist/fs/seek.go
```diff
@@ -383,55 +367,96 @@ func (s *seeker) SeekByIndexEntry(entry IndexEntry) (checked.Bytes, error) {
...
+			resources.decodeIndexEntryBytesPool)
+		if err != nil {
+			// Should never happen, either something is really wrong with the code or
+			// the file on disk was corrupted.
+			return IndexEntry{}, err
+		}
+		if entry.ID == nil {
+			// Should never happen, either something is really wrong with the code or
+			// the file on disk was corrupted.
+			return IndexEntry{}, errDecodedIndexEntryHadNoID
```
robskillington: Want to use the invariant violated stuff for these two code paths?

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -378,20 +393,21 @@ func (m *seekerManager) newOpenSeeker(
...
 	// Set the unread buffer to reuse it amongst all seekers.
 	seeker.setUnreadBuffer(m.unreadBuf.value)
 
-	if err := seeker.Open(m.namespace, shard, blockStart); err != nil {
+	resources := m.getSeekerResources()
+	if err := seeker.Open(m.namespace, shard, blockStart, resources); err != nil {
+		m.putSeekerResources(resources)
 		return nil, err
 	}
+	m.putSeekerResources(resources)
```
robskillington: Can you simplify this code so you don't have to call `putSeekerResources(...)` from two different code paths?

e.g.
```go
	resources := m.getSeekerResources()
	err := seeker.Open(m.namespace, shard, blockStart, resources)
	m.putSeekerResources(resources)

	if err != nil {
		return nil, err
	}
```

src/dbnode/persist/fs/seek.go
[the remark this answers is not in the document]
richardartoul: I just added a comment to it

src/dbnode/persist/fs/index_lookup.go
richardartoul: done

PR #1301, opened 2019-01-17
Expose vellum regexp

benraskin92:


src/m3ninx/index/regexp.go
```diff
@@ -62,7 +62,10 @@ func CompileRegex(r []byte) (CompiledRegex, error) {
 	if err != nil {
 		return CompiledRegex{}, err
 	}
-	compiledRegex := CompiledRegex{Simple: simpleRE}
+	compiledRegex := CompiledRegex{
+		Simple:       simpleRE,
+		VellumSyntax: vellumRe,
```
richardartoul: nit: Would it be worth renaming this field to something like: "UnanchoredSyntax" if we're gonna start using it in other places?

src/m3ninx/index/types.go
```diff
@@ -83,10 +84,11 @@ type Readable interface {
 // CompiledRegex is a collection of regexp compiled structs to allow
 // amortisation of regexp construction costs.
 type CompiledRegex struct {
-	Simple      *regexp.Regexp
-	FST         *vregex.Regexp
-	PrefixBegin []byte
-	PrefixEnd   []byte
+	Simple       *regexp.Regexp
+	FST          *vregex.Regexp
+	VellumSyntax *syntax.Regexp
```
robskillington: Could you call this `FSTSyntax` to be consistent with the other field which is called `FST`?

PR #1303, opened 2019-01-17
Fix cloning for segment readers

arnikola:
Segment readers were reusing the same underlying bytes, and not cloning
correctly, which lead to issues when splitting seriesIterators and
closing the parent iterator.


src/dbnode/ts/segment.go
```diff
@@ -123,3 +124,34 @@ func (s *Segment) Finalize() {
...
+func (s *Segment) Clone(pool pool.CheckedBytesPool) Segment {
+	var (
+		checkedHead, checkedTail checked.Bytes
+	)
+
+	head := s.Head.Bytes()
+	ch := make([]byte, len(head))
+	copy(ch, head)
+	tail := s.Tail.Bytes()
+	ct := make([]byte, len(tail))
```
robskillington: Hey instead of `make` and copying in here, if the pool is not nil you should just copy into that?

i.e.
```go
head := s.Head.Bytes()
tail := s.Tail.Bytes()

if pool != nil {
  checkedHead = pool.Get(len(head))
  checkedHead.IncRef()
  checkedHead.AppendAll(head)
  checkedHead.DecRef()
  checkedTail = pool.Get(len(tail))
  checkedTail.IncRef()
  checkedTail.AppendAll(tail)
  checkedTail.DecRef()
}
```

src/dbnode/ts/segment.go
```diff
@@ -123,3 +124,34 @@ func (s *Segment) Finalize() {
...
+		checkedTail = pool.Get(len(ct))
+		checkedTail.IncRef()
+		checkedTail.Reset(ct)
+		checkedTail.DecRef()
+	} else {
+		checkedHead = checked.NewBytes(ch, nil)
+		checkedTail = checked.NewBytes(ct, nil)
+	}
+
+	return NewSegment(checkedHead, checkedTail, s.Flags)
```
robskillington: You'll want to return FinalizeHead&FinalizeTail since we copy both of them here.

src/dbnode/ts/segment.go
arnikola: Oh... duh

Was actually going to remove the `make([]byte...` before submitting but somehow managed to forget :|

src/dbnode/ts/segment.go
arnikola: Nice, good point.

src/dbnode/x/xio/io_mock.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by MockGen. DO NOT EDIT.
 // Source: github.com/m3db/m3/src/dbnode/x/xio (interfaces: SegmentReader,SegmentReaderPool)
 
-// Copyright (c) 2018 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
```
arnikola: Would we want to manually revert this?

src/dbnode/persist/fs/retriever.go
```diff
@@ -582,6 +582,16 @@ func (req *retrieveRequest) Clone() (xio.SegmentReader, error) {
 	return req.reader.Clone()
 }
 
+func (req *retrieveRequest) DeepClone(
```
robskillington: Hm, what's deep clone used for?

richardartoul: I think its so that the Retriever code path doesn't have to clone the bytes again (which have already just been cloned from disk). I need to look through the retriever code path before we merge this, but I do think we want to avoid cloning twice

PR #1305, opened 2019-01-18
Correctly return error for failed json unmarshal

arnikola:
Unmarshaling error was not being returned correctly, which could cause null pointer references later on.

### DIFF
--- a/src/query/api/v1/handler/json/write.go
+++ b/src/query/api/v1/handler/json/write.go
@@ -124,7 +124,9 @@ func (h *WriteJSONHandler) parseRequest(r *http.Request) (*WriteQuery, *xhttp.Pa
 	}
 
 	var writeQuery *WriteQuery
-	json.Unmarshal(js, &writeQuery)
+	if err = json.Unmarshal(js, &writeQuery); err != nil {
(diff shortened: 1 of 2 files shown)


src/query/api/v1/handler/json/write.go
```diff
@@ -124,7 +124,9 @@ func (h *WriteJSONHandler) parseRequest(r *http.Request) (*WriteQuery, *xhttp.Pa
 	}
 
 	var writeQuery *WriteQuery
-	json.Unmarshal(js, &writeQuery)
+	if err = json.Unmarshal(js, &writeQuery); err != nil {
```
andrewmains12: Nit: I would do `err :=` here; that way we're definitely only assigning within that scope.

arnikola: The err doesn't persist past this scope, don't know if that changes your mind at all 📦

andrewmains12: Yeah something of a style thing; we don't need beyond the `if` scope, so might as well leave it in the smaller scope. Not hugely important.

arnikola: Sure, linter sometimes complains about shadowing; doesn't make a difference logically at least :)

PR #1306, opened 2019-01-18
Reduce unit test memory usage

richardartoul:


.buildkite/pipeline.yml
```diff
@@ -21,13 +21,15 @@ steps:
     <<: *common
   - name: "Unit"
```
schallert: Can you make these `Unit %n` etc.

src/dbnode/client/session_test.go
```diff
@@ -56,7 +56,16 @@ type testEnqueueFn func(idx int, op op)
...
+	_testSessionOpts = NewOptions().
+		SetCheckedBytesWrapperPoolSize(1).
+		SetFetchBatchOpPoolSize(1).
+		SetHostQueueOpsArrayPoolSize(1).
+		SetSeriesIteratorPoolSize(1).
+		SetTagDecoderPoolSize(1).
+		SetTagEncoderPoolSize(1).
+		SetWriteOpPoolSize(1).
+		SetWriteTaggedOpPoolSize(1).
+		SetSeriesIteratorPoolSize(1)
```
robskillington: Yeah, this is much better, TY.

PR #1307, opened 2019-01-18
Fix tag propagation for temporal functions

andrewmains12:
I ran into this bug locally today--temporal functions don't properly propagate series tags (thanks to @arnikola for suggesting this as the bug!).

For instance, given series:

```
coordinator_engine_datapoints{type="fetched"} 1386
coordinator_engine_datapoints{type="generated"} 104
```
a query like `increase(coordinator_engine_datapoints[5s])` will return
```
{
  "__name__": "coordinator_engine_datapoints",
  "instance": "host.docker.internal:7203",
  "job": "coordinator",
  "role": "remote"
}
{
  "__name__": "coordinator_engine_datapoints",
  "instance": "host.docker.internal:7203",
  "job": "coordinator",
  "role": "remote"
}
```
dropping tags. Query the same range without `increase` gives all tags, as expected.

Fix is simple; we weren't copying tags into the new block's `SeriesMetas`; now we do.

### DIFF
--- a/src/query/functions/temporal/base_test.go
+++ b/src/query/functions/temporal/base_test.go
@@ -53,10 +53,10 @@ func (p *processor) Process(dps ts.Datapoints, _ time.Time) float64 {
 	return sum
 }
 
-func compareCacheState(t *testing.T, bNode *baseNode, bounds models.Bounds, state []bool, debugMsg string) {
+func compareCacheState(t *testing.T, node *baseNode, bounds models.Bounds, state []bool, debugMsg string) {
 	actualState := make([]bool, len(state))
 	for i := range state {
-		_, exists := bNode.cache.get(bounds.Next(i).Start)
+		_, exists := node.cache.get(bounds.Next(i).Start)
 		actualState[i] = exists
 	}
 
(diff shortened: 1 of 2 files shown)


andrewmains12:
Note: I also refactored the tests to deduplicate a bit; I thought I would need it, and figured I might as well leave it even though I didn't.

src/query/functions/temporal/base_test.go
```diff
@@ -502,4 +406,12 @@ func TestSingleProcessRequest(t *testing.T) {
...
 	assert.Equal(t, sink.Values[0], []float64{50, 40, 30, 20, 10}, "first series is 10 - 14 which sums to 60, the current block first series is 0-4 which sums to 10, we need 5 values per aggregation")
 	assert.Equal(t, sink.Values[1], []float64{75, 65, 55, 45, 35}, "second series is 15 - 19 which sums to 85 and second series is 5-9 which sums to 35")
+
+	// processSingleRequest renames the series to use their ids; reflect this in our expectation.
+	expectedSeriesMetas := make([]block.SeriesMeta, len(seriesMetas))
+	require.Equal(t, len(expectedSeriesMetas), copy(expectedSeriesMetas, seriesMetas))
+	expectedSeriesMetas[0].Name = "t1=v1,"
+	expectedSeriesMetas[1].Name = "t1=v2,"
+
+	assert.Equal(t, expectedSeriesMetas, sink.Metas, "Process should pass along series meta, renaming by ID")
```
arnikola: nit: `renaming to ID`

src/query/functions/temporal/base_test.go
```diff
@@ -121,351 +121,255 @@ func TestBaseWithB0(t *testing.T) {
...
+	assert.Len(t, tc.Sink.Values, 0, "nothing processed yet")
+	compareCacheState(t, tc.Node, tc.Bounds, []bool{false, true}, "B1 cached")
 
-	err = node.Process(parser.NodeID(0), blocks[0])
+	err = tc.Node.Process(parser.NodeID(0), tc.Blocks[0])
 	require.NoError(t, err)
-	assert.Len(t, sink.Values, 4, "output from both blocks")
-	compareCacheState(t, bNode, bounds, []bool{false, false}, "everything removed from cache")
-	blks, err := bNode.cache.multiGet(bounds, 2, false)
+	assert.Len(t, tc.Sink.Values, 4, "output from both tc.Blocks")
```
arnikola: nit: copy/paste error?

andrewmains12: Whoops yeah thanks.

src/query/functions/temporal/base_test.go
andrewmains12: 👍

PR #1308, opened 2019-01-19
[query] Add graphite query path

arnikola:
Adds graphite query endpoint to m3query. This is largely a port from our internal legacy service with some additional transform steps that contain the blast radius to a few codepaths.

TODO:
- [ ] Readme describing that this code is largely migrated from existing project and has not been modernized
- [x] Update graphite string transform to tags to [mirror write path](https://github.com/m3db/m3/pull/1310/files#diff-a928a4a419a386b39396d79d5febca63R149)
- [x] Add globbing detection for queries 
- [x] Tests for new plumbing, and holes in existing coverage
- [x] LTTB downsampling when maxDataPoints is specified


src/query/graphite/common/_blacklist_testgo
```diff
@@ -0,0 +1,89 @@
+package common
```
robskillington: Think we can delete this file (blacklist)

src/query/graphite/common/_blacklistgo
```diff
@@ -0,0 +1,229 @@
+package common
```
robskillington: Same with this one.

arnikola: Oops forgot to delete the ones in /common/

src/query/graphite/graphite/graphite_types_test.go
```diff
@@ -0,0 +1,149 @@
...
+
+func TestDatapointAccess(t *testing.T) {
+	data, err := ioutil.ReadFile("testdata/no-results.json")
+	require.Nil(t, err)
+
+	var results RenderResults
+	err = json.Unmarshal(data, &results)
+	require.Nil(t, err)
+	require.Equal(t, 1, len(results))
+	require.Equal(t, "exp.sjc1.timers.cn.production.cn37-sjc1.city_load.monterrey.total.p95", results[0].Target)
```
robskillington: We need to replace all IDs in testdata with more generic, non-real IDs.

src/query/graphite/graphite/identify_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+const (
+	statsdStat    = "dispatch.production.san_francisco.uberx.drivers"
+	malformedStat = "dispatchproductionsan_franciscouberxdrivers"
```
robskillington: Similar, need to change both these two stats to more test data-ish looking names.

src/query/graphite/graphite/identify_test.go
```diff
@@ -0,0 +1,44 @@
...
+	dashedString  = "some-other-delimiter"
+)
+
+func TestExtractNthEmptyMetric(t *testing.T) {
+	assert.Equal(t, "", ExtractNthMetricPart("", 0))
+	assert.Equal(t, 0, CountMetricParts(""))
+}
+
+func TestExtractNthMetricPartNoDots(t *testing.T) {
+	nodots := "dispatchproductionsan_franciscouberxdrivers"
```
robskillington: Similar, need to change both these two stats to more test data-ish looking names.

src/query/graphite/graphite/identify_test.go
```diff
@@ -0,0 +1,44 @@
...
+	assert.Equal(t, nodots, ExtractNthMetricPart(malformedStat, 0))
+	assert.Equal(t, 1, CountMetricParts(malformedStat))
+}
+
+func TestExtractNthMetricPartStandardCase(t *testing.T) {
+	assert.Equal(t, "dispatch", ExtractNthMetricPart(statsdStat, 0))
+	assert.Equal(t, "production", ExtractNthMetricPart(statsdStat, 1))
+	assert.Equal(t, "drivers", ExtractNthMetricPart(statsdStat, 4))
```
robskillington: Similar, need to change both these two stats to more test data-ish looking names.

src/query/graphite/graphite/metric_type_test.go
```diff
@@ -0,0 +1,130 @@
...
+
+	ca = ConsolidationFuncForMetricType(MetricType(999))
+	assert.Equal(t, ts.ConsolidationAvg, ca)
+}
+
+func TestMetricTypeFromID(t *testing.T) {
+	assert.Equal(t, metric.UnknownType, MetricTypeFromID("servers.m3-kv05-dca1.disk.sdd.rkb"))
+	assert.Equal(t, metric.CounterType, MetricTypeFromID("stats.sjc1.counts.m3+calls+dc=sjc1"))
+	assert.Equal(t, metric.GaugeType, MetricTypeFromID("stats.sjc1.gauges.m3+task+dc=sjc1"))
+	assert.Equal(t, metric.TimerType, MetricTypeFromID("stats.sjc1.timers.haproxy.beagle.response_time.p99"))
```
robskillington: Similar, need to change both these two stats to more test data-ish looking names.

src/query/graphite/graphite/metric_type.go
```diff
@@ -0,0 +1,117 @@
+package graphite
+
+import (
```
robskillington: Actually wonder if we need the metric type detection at all, think we can possibly do without it.

Let me see where this tricks up from.

src/query/graphite/graphite/retention_policies.go
```diff
@@ -0,0 +1,106 @@
...
+type RetentionPolicy struct {
+	TTL           time.Duration
+	UnitPerStep   time.Duration
+	Consolidation ts.ConsolidationApproach
+	RoundToUnit   bool
+}
+
+var (
+	// NB(mmihic): These need to be ordered
+	graphiteRetentionPeriods = []*RetentionPeriod{
```
robskillington: Don't think we need any of these retention period definitions.

Let me see where this tricks up from, I'm pretty certain we can remove them.

src/query/graphite/graphite/testdata/no-results.json
```diff
@@ -0,0 +1 @@
...30838350],[null,1430838360],[null,1430838370],[null,1430838380],[null,1430838390],[null,1430838400],[null,1430838410],[null,1430838420],[null,1430838430],[null,1430838440],[null,1430838450],[null,1430838460],[null,1430838470],[null,1430838480],[null,1430838490],[null,1430838500],[null,1430838510],[null,1430838520],[null,1430838530],[null,1430838540],[null,1430838550],[null,1430838560],[null,1430838570],[null,1430838580],[null,1430838590],[null,1430838600],[null,1430838610],[null,1430838620],[null,1430838630],[null,1430838640],[null,1430838650],[null,1430838660],[null,1430838670],[null,1430838680],[null,1430838690],[null,1430838700],[null,1430838710],[null,1430838720],[null,1430838730],[null,1430838740],[null,1430838750],[null,1430838760],[null,1430838770],[null,1430838780],[null,1430838790],[null,1430838800],[null,1430838810],[null,1430838820],[null,1430838830],[null,1430838840],[null,1430838850],[null,1430838860],[null,1430838870],[null,1430838880],[null,1430838890],[null,1430838900],[null,1430838910],[null,1430838920],[null,1430838930],[null,1430838940],[null,1430838950],[null,1430838960],[null,1430838970],[null,1430838980],[null,1430838990],[null,1430839000],[null,1430839010]]}]
```
robskillington: Yeah need to update this metric ID in here to be more test data-ish.

src/query/graphite/lexer/lexer_test.go
```diff
@@ -0,0 +1,142 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+type lexerTestData struct {
+	input          string
+	expectedTokens []Token
+	reservedIdents map[string]TokenType
+}
+
+func TestLexer(t *testing.T) {
```
robskillington: Similar for this method, need to remove all of the non-test data IDs and make them more synthetic/test data like (e.g. foo/bar/baz/etc)

src/query/graphite/ts/series.go
```diff
@@ -0,0 +1,645 @@
...
+	vals      Values
+	ctx       context.Context
+
+	// The Specification is the path that was used to generate this timeseries,
+	// typically either the query, or the function stack used to transform
+	// specific results.
+	Specification string
+
+	// Metric tags.
+	Tags map[string]string
```
robskillington: We can get rid of tags since this is only relevant to graphite.

src/query/graphite/ts/series.go
```diff
@@ -0,0 +1,645 @@
...
+
+	// options for graphing
+	GraphOptions map[string]string
+
+	// consolidationFunc specifies how the series will be consolidated when the
+	// number of data points in the series is more than the maximum number allowed.
```
robskillington: We can get rid of GraphOptions, this isn't relevant for a graphite only implementation.

src/query/ts/series.go
```diff
@@ -21,15 +21,18 @@
...
 	"github.com/m3db/m3/src/query/models"
 )
 
 // Series is the public interface to a block of timeseries values.  Each block has a start time,
 // a logical number of steps, and a step size indicating the number of milliseconds represented by each point.
 type Series struct {
-	name string
-	vals Values
-	Tags models.Tags
+	resolution int
```
robskillington: Can we make this `time.Duration` instead of just an int?

It's a loss of precision to turn this into millis and seems more type safe to just use `time.Duration` everywhere rather than an `int` that is named `blahMillis`.

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -83,6 +83,23 @@ func (r *multiResult) Close() error {
...
+	encoding.SeriesIterators, []storage.Attributes, error) {
+	iters, err := r.FinalResult()
+	if err != nil {
+		return nil, nil, err
+	}
+
+	i := 0
+	attrs := make([]storage.Attributes, iters.Len())
+	for _, res := range r.dedupeMap {
+		attrs[i] = res.attrs
```
robskillington: Hm, we don't always have the dedupe map in the case of we only got one set of iterators.

You'd have to set the `attrs[i] = r.seenFirstAttrs` for that case.

Perhaps check if `len(r.dedupeMap) != iters.Len()` and if so just return `r.seenFirstAttrs` for all of the series?

src/query/graphite/ts/series.go
arnikola: It's a bit more involved since it's used in a bunch of tests, might do a more specialized diff to get rid of it

src/query/ts/series.go
arnikola: Sure sounds good, was just doing it quick and dirty for the conversion to millis later on

src/query/storage/m3/multi_fetch_result.go
arnikola: Good call

src/query/api/v1/handler/graphite/native/render_parser.go
```diff
@@ -0,0 +1,217 @@
...
+	// the server's local time in order to take advantage of possibly higher-resolution data.
+	// In the future we could potentially distinguish absolute time and relative time and only
+	// use the time range for policy resolution, although we need to be careful when passing
+	// the range for cross-DC queries.
+	isRealTimeQuery := now.Sub(p.Until) < realTimeQueryThreshold
+	isLargeRangeQuery := p.Until.Sub(p.From) > queryRangeShiftThreshold
+	if isRealTimeQuery && isLargeRangeQuery {
+		p.From = p.From.Add(queryRangeShift)
+		p.Until = p.Until.Add(queryRangeShift)
+	}
```
robskillington: We can probably just get rid of all this for now.

src/query/api/v1/handler/graphite/native/search.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/query/util/logging"
+	"github.com/m3db/m3/src/x/net/http"
+
+	"go.uber.org/zap"
+)
+
+const (
+	// GraphiteSearchURL is the url for searching graphite paths.
+	GraphiteSearchURL = handler.RoutePrefixGraphiteV1 + "/search"
```
robskillington: Think this has to be `/find`.

src/query/api/v1/handler/graphite/native/search.go
```diff
@@ -0,0 +1,116 @@
...
+			if bytes.Equal(partName, tag.Name) {
+				index = i
+				break
+			}
+		}
+
+		value := tags[index].Value
+		// If this value has already been encountered, check if
+		if hadExtra, seen := seenMap[string(value)]; seen {
+			if hadExtra {
```
robskillington: Combine these two if's perhaps?  
```go
if hadExtra, seen := seenMap[string(value)]; seen && hadExtra {
  continue
}
```

src/query/graphite/common/context.go
```diff
@@ -0,0 +1,315 @@
...
+}
+
+// ContextOptions provides the options to create the context with
+type ContextOptions struct {
+	Start     time.Time
+	End       time.Time
+	Engine    QueryEngine
+	LocalOnly bool
+	UseCache  bool
+	UseM3DB   bool
```
robskillington: Can remove LocalOnly, UseCache and UseM3DB here.

src/query/graphite/common/context.go
```diff
@@ -0,0 +1,315 @@
...
+	Timeout   time.Duration
+	Zone      string
```
robskillington: Can also remove Zone.

src/query/graphite/common/context.go
arnikola: Ah damn, removed them from regular context object but missed cleaning up contextBase

PR #1310, opened 2019-01-22
Add Graphite ingestion

richardartoul:


src/cmd/services/m3coordinator/ingest/write.go
```diff
@@ -0,0 +1,188 @@
...
+			MetricsType: storage.UnaggregatedMetricsType,
+		},
+	})
+}
+
+func (d *downsamplerAndWriter) WriteBatch(
+	ctx context.Context,
+	iter DownsampleAndWriteIter,
+) error {
+	if d.downsampler != nil {
```
richardartoul: start goroutines first then do this then wait

src/query/api/v1/handler/prometheus/remote/write.go
```diff
@@ -126,104 +125,43 @@ func (h *PromWriteHandler) parseRequest(r *http.Request) (*prompb.WriteRequest,
...
-	}
+	iter := newPromTSIter(r.Timeseries)
+	return h.downsamplerAndWriter.WriteBatch(ctx, iter)
+}
 
-	var multiErr xerrors.MultiError
-	multiErr = multiErr.Add(writeUnaggErr)
-	multiErr = multiErr.Add(writeAggErr)
-	return multiErr.FinalError()
+type promTSIter struct {
```
richardartoul: generate models.Tags upfront so we don't need to allocate them twice

richardartoul: and ts.Datapoints

scripts/docker-integration-tests/common.sh
```diff
@@ -36,3 +36,99 @@ function retry_with_backoff {
...
+      "flushEnabled": true,
+      "writesToCommitLog": true,
+      "cleanupEnabled": true,
+      "snapshotEnabled": true,
+      "repairEnabled": false,
+      "retentionOptions": {
+        "retentionPeriodNanos": 172800000000000,
+        "blockSizeNanos": 7200000000000,
+        "bufferFutureNanos": 600000000000,
+        "bufferPastNanos": 600000000000,
```
robskillington: Nit: Probably better to use string and durations? i.e. `"bufferPast": "10m"` etc

scripts/docker-integration-tests/carbon/test.sh
```diff
@@ -0,0 +1,33 @@
...
+
+echo "Run m3dbnode and m3coordinator containers"
+docker-compose -f ${COMPOSE_FILE} up -d dbnode01
+docker-compose -f ${COMPOSE_FILE} up -d coordinator01
+
+# think of this as a defer func() in golang
+function defer {
+  docker-compose -f ${COMPOSE_FILE} down || echo "unable to shutdown containers" # CI fails to stop all containers sometimes
+}
+trap defer EXIT
```
robskillington: Haha 👍 good old trap

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
```diff
@@ -0,0 +1,216 @@
...
+	InstrumentOptions instrument.Options
+	WorkerPool        xsync.PooledWorkerPool
+}
+
+// NewIngester returns an ingester for carbon metrics.
+func NewIngester(
+	downsamplerAndWriter ingest.DownsamplerAndWriter,
+	opts Options,
+) (m3xserver.Handler, error) {
+	// TODO(rartoul): Validate options.
```
robskillington: Maybe add `Validate() error` method to the `Options` struct itself?

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
```diff
@@ -0,0 +1,216 @@
...
+
+			// TODO(rartoul): Context with timeout?
+			err = i.downsamplerAndWriter.Write(
+				context.Background(), tags, datapoints, xtime.Second)
+			if err != nil {
+				logger.Errorf("err writing carbon metric: %s, err: %s",
+					string(name), err)
+			}
+
+			wg.Done()
```
robskillington: This `wg.Done()` is not called if early return up top.

Perhaps just pay the cost of using `defer wg.Done()` at the top of the goroutine?

src/query/server/server.go
```diff
@@ -261,6 +265,59 @@ func Run(runOpts RunOptions) {
...
+		ingester, err := ingestcarbon.NewIngester(carbonIngestDownsamplerAndWriter, ingestcarbon.Options{
+			InstrumentOptions: carbonIOpts,
+			WorkerPool:        workerPool,
+		})
+		if err != nil {
+			logger.Fatal("unable to create carbon ingester: %v", zap.Error(err))
+		}
+
+		// TODO: Default constants
+		listenAddress := fmt.Sprintf("0.0.0.0:7204")
```
vdarulis: just `"0.0.0.0:7204"` or const

src/query/server/server.go
```diff
@@ -261,6 +265,59 @@ func Run(runOpts RunOptions) {
...
+		} else {
+			carbonWorkerPoolOpts = xsync.NewPooledWorkerPoolOptions().
+				SetGrowOnDemand(true).
+				SetKillWorkerProbability(0.001)
+			// TODO: Constants?
+			carbonWorkerPoolSize = 1024
+		}
+		workerPool, err := xsync.NewPooledWorkerPool(carbonWorkerPoolSize, carbonWorkerPoolOpts)
+		if err != nil {
+			logger.Fatal("unable to create worker pool for carbon ingester: %v", zap.Error(err))
```
vdarulis: drop `: %v"` as you're adding a zap.Error field

src/cmd/services/m3query/config/config.go
```diff
@@ -127,12 +130,20 @@ type LimitsConfiguration struct {
...
 
 	// M3Msg is the configuration for m3msg server.
 	M3Msg m3msg.Configuration `yaml:"m3msg"`
 }
 
+// TODO(rartoul): Make this enabled by default.
+// CarbonConfiguration is the configuration for the carbon server.
+type CarbonConfiguration struct {
+	Enabled        bool   `yaml:"enabled"`
+	MaxConcurrency int    `yaml:"max_concurrency"`
```
robskillington: For consistency, make this camelcase YAML key?  i.e. `maxConcurrency`

src/cmd/services/m3query/config/config.go
```diff
@@ -127,12 +130,20 @@ type LimitsConfiguration struct {
...
+	ListenAddress  string `yaml:"listen_address"`
```
robskillington: For consistency, make these both camelcase YAML key?  i.e. `maxConcurrency`

src/cmd/services/m3coordinator/ingest/write.go
```diff
@@ -0,0 +1,213 @@
...
+
+	return nil
+}
+
+func (d *downsamplerAndWriter) WriteBatch(
+	ctx context.Context,
+	iter DownsampleAndWriteIter,
+) error {
+	var (
+		wg       = &sync.WaitGroup{}
```
vdarulis: just `sync.WaitGroup`?

src/query/server/server.go
```diff
@@ -261,6 +265,59 @@ func Run(runOpts RunOptions) {
...
+			carbonWorkerPoolOpts = xsync.NewPooledWorkerPoolOptions().
+				SetGrowOnDemand(false).
+				SetInstrumentOptions(carbonIOpts)
+			carbonWorkerPoolSize = cfg.Carbon.MaxConcurrency
+		} else {
+			carbonWorkerPoolOpts = xsync.NewPooledWorkerPoolOptions().
+				SetGrowOnDemand(true).
+				SetKillWorkerProbability(0.001)
+			// TODO: Constants?
+			carbonWorkerPoolSize = 1024
```
robskillington: Yeah maybe make this a constant?

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
richardartoul: good catch.

scripts/docker-integration-tests/carbon/m3coordinator.yml
```diff
@@ -0,0 +1,54 @@
...
+  samplingRate: 1.0
+  extended: none
+
+clusters:
+  - namespaces:
+      - namespace: agg
+        type: aggregated
+        retention: 10h
+        resolution: 15s
+      - namespace: unagg
```
arnikola: nit: add a TODO to remove this once unaggregated is no longer mandatory?

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
```diff
@@ -0,0 +1,265 @@
...
+	err = i.downsamplerAndWriter.Write(ctx, tags, datapoints, xtime.Second)
+	if err != nil {
+		i.logger.Errorf("err writing carbon metric: %s, err: %s",
+			string(name), err)
+		i.metrics.err.Inc(1)
+		return false
+	}
+
+	if cleanup != nil {
+		cleanup()
```
robskillington: This doesn't get called if `err != nil`.

The common way to do this apparently is:
```go
ctx, cancel = context.WithTimeout(ctx, i.opts.Timeout)

// ...

err = i.downsamplerAndWriter.Write(ctx, tags, datapoints, xtime.Second)
if cancel != nil {
  cancel()
}

// now deal with err
if err != nil {
  // ...
  return
}
```

robskillington: Example here:
https://github.com/uber/tchannel-go/blob/a815a8e7cbcb4ebcd05a73a84f7fd2237dd8a44f/health.go#L127-L131

src/cmd/services/m3query/config/config.go
```diff
@@ -94,10 +101,47 @@ type Configuration struct {
...
+				ListenAddress: defaultCarbonIngesterListenAddress,
+				Timeout:       &defaultCarbonIngesterTimeout,
+			},
+		}
+	}
+
+	if c.Carbon.Ingestion.Enabled == nil {
+		// If they have set any carbon configuration and haven't explicitly marked
+		// the ingestion as disabled, then assume they want it enabled.
+		c.Carbon.Ingestion.Enabled = &defaultCarbonIngesterEnabled
```
robskillington: Perhaps you could just make all these values setters use getters instead?

i.e.
```go
func (c *CarbonIngestionConfiguration) EnabledValue() bool {
  if c.Enabled != nil {
    return *c.Enabled
  }
  return defaultCarbonIngestConfigurationEnabled
}
```

That way there's no way any field could be nil when accessing it it.

src/cmd/services/m3query/config/config.go
```diff
@@ -94,10 +101,47 @@ type Configuration struct {
...
 
+	// Carbon is the carbon configuration.
+	Carbon *CarbonConfiguration `yaml:"carbon"`
+
 	// Limits specifies limits on per-query resource usage.
 	Limits LimitsConfiguration `yaml:"limits"`
 }
 
+// CarbonConfiguration returns the carbon configuration.
+func (c *Configuration) CarbonConfiguration() *CarbonConfiguration {
```
robskillington: Can we make this return `CarbonConfiguration` instead of `*CarbonConfiguration`?  That way it's always not nil when used as a local var returned from this method.

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
richardartoul: ah thanks, this wasn't broken when I first wrote it but added a hasty refactor. Luckily leaking these the G.C will eventually get them but good catch I'll clean it up proper.

src/cmd/services/m3query/config/config.go
richardartoul: ended up not needing this accessor at all. Made everything non-point and then used the getter pattern you described above.

PR #1311, opened 2019-01-23
Upgrade GRPC version

richardartoul:


glide.yaml
```diff
@@ -196,7 +196,7 @@ import:
 
   # To avoid conflicting packages not resolving the latest GRPC
   - package: google.golang.org/grpc
-    version: ~1.7.3
+    version: ^1.14.0
```
arnikola: I think this may break round-robin routing for coordinator rpc calls

PR #1314, opened 2019-01-25
Add samples appender options to allow aggregating with retention/resolution

robskillington:


src/cmd/services/m3coordinator/downsample/samples_appender.go
```diff
@@ -52,6 +56,40 @@ func (a samplesAppender) AppendGaugeSample(value float64) error {
 	return a.agg.AddUntimed(sample, a.stagedMetadatas)
 }
 
+func (a *samplesAppender) AppendCounterTimedSample(value int64, t time.Time) error {
```
richardartoul: Super nit but Datapoint structs are usually:

```
type Datapoint struct {
    Timestamp time.Time
     value int64
}
```

Which makes me think time should proceed value in the API

robskillington: Sure thing, sounds good.

PR #1315, opened 2019-01-25
Add query how to

benraskin92:


docs/how_to/query.md
```diff
@@ -0,0 +1,47 @@
+# Setting up m3query
+
+## Introduction
+
+m3query is used to query data that is stored in M3DB. For instance, if you are using the Prometheus remote write endpoint with [m3coordinator](../integrations/prometheus.md), you can use m3query instead of the Prometheus remote read endpoint. By doing so, you get all of the benefits of m3query's engine. Furthermore, since m3query provides a Prometheus compatible API, you use 3rd party graphing and alerting solutions like Grafana.
```
andrewmains12: What are our benefits? :).

docs/how_to/query.md
```diff
@@ -0,0 +1,47 @@
...
+
+<!-- TODO: link to m3query docs -->
+
+## Configuration
+
+Before setting up m3query, make sure that you have at least [one M3DB node running](single_node.md). In order to start m3query, you need to configure a `yaml` file, that will be used to connect to M3DB. Here is a link to a [sample config](https://github.com/m3db/m3/blob/master/src/query/config/m3query-local-etcd.yml) file that is used for an embedded etcd cluster.
```
andrewmains12: We have a Dockerfile for `m3query` right? Should we suggest using that?

andrewmains12: Nit: it might make sense to have two sections, one on `Running` (which could mention the Dockerfile), and one on `Configuration`. Right now this section is doing both.

benraskin92: We do, but I think it just uses a default config file so not sure we should recommend using it.

andrewmains12: Looks like it's using this exact config https://github.com/m3db/m3/blob/master/docker/m3query/Dockerfile#L24 . That said, we should probably make sure it works before recommending it haha. In the meantime, can we provide specific instructions on building on running? e.g.

```
make m3query
./bin/m3query -f ./src/query/config/m3query-local-etcd.yml
```

docs/how_to/query.md
richardartoul: Do we capitalize? I think we capitalize M3DB usually but not sure about query.

richardartoul: `you use` -> `you can use`

docs/how_to/query.md
```diff
@@ -0,0 +1,47 @@
+# Setting up m3query
+
+## Introduction
+
+m3query is used to query data that is stored in M3DB. For instance, if you are using the Prometheus remote write endpoint with [m3coordinator](../integrations/prometheus.md), you can use m3query instead of the Prometheus remote read endpoint. By doing so, you get all of the benefits of m3query's engine. Furthermore, since m3query provides a Prometheus compatible API, you use 3rd party graphing and alerting solutions like Grafana.
+
+<!-- TODO: link to m3query docs -->
```
richardartoul: What docs are you talking about? Is this referring to the configuration docs we don't have yet?

docs/how_to/query.md
richardartoul: +1 andrew and then also you say that the config is used for an embedded etcd cluster but thats kind of confusing because its the m3db node that runs the embedded etcd cluster

docs/how_to/query.md
```diff
@@ -0,0 +1,47 @@
...
+
+### etcd
+
+The configuration file linked above uses an embedded etcd cluster, which is fine for development purposes. However, if you wish to use this in production, you will want an external etcd cluster.
+
+<!-- TODO: link to etcd operational guide -->
+
+## Aggregation
+
+You will notice that in the setup linked above, M3DB has just one unaggregated namespace configured. If you want aggregated metrics, you will need to set up an aggregated namespace in M3DB **and** in the m3query configuration. It it important to note that all writes go to all namespaces so as long as you include all namespaces in your query config, you will be querying all namespaces. Aggregation is done strictly by the query service. For example if you have an aggregated namespace setup in M3DB named `metrics_10s_48h`, you can add the following to the query config:
```
richardartoul: Do we always call it m3query and never m3coordinator? I know people in gitter seem to call it the coordinator a lot. Sounds kind of weird to say: "aggregation is done strictly by the query service"

richardartoul: This is kind of why I think this should be in its own "How to" document.

Basically in the "How to setup aggregation" document we can explain the difference between M3Query and M3Coordinator a little bit, then you can talk about the fact that you need to add namespaces to M3DB, then you can show how to add the namespaces to the config, and then you can link to both the Prometheus and Carbon integration documents.

docs/how_to/query.md
benraskin92: I haven't been capitalizing `m3query`

docs/how_to/query.md
benraskin92: Will remove - meant to link to a high level summary page of m3query

docs/how_to/query.md
benraskin92: I want this file to just be for the query engine. Otherwise people are gonna get very confused. I agree we should have a separate section somewhere explaining the differences between query and coordinator. I'll work on writing something up. Maybe in a `Terms` or `Glossary` section.

docs/how_to/query.md
```diff
@@ -0,0 +1,56 @@
...
+### Running
+
+You can run m3query by either building and running the binary yourself:
+
+```
+make m3query
+./bin/m3query -f ./src/query/config/m3query-local-etcd.yml
+```
+
+Or you can run it with Docker using the Docker file located at `$GOPATH/src/github.com/m3db/m3/docker/m3query/Dockerfile`.
```
andrewmains12: Nit: specific docker command here would be more convenient imo.

PR #1316, opened 2019-01-25
Add native read metrics

benraskin92:


src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -56,9 +57,28 @@ var (
...
+	fetchErrorsServer tally.Counter
+	fetchErrorsClient tally.Counter
+	fetchTimerSuccess tally.Timer
+	maxDatapoints     tally.Gauge
+}
+
+func newPromReadMetrics(scope tally.Scope) promReadMetrics {
+	return promReadMetrics{
+		fetchSuccess:      scope.Counter("fetch.success"),
+		fetchErrorsServer: scope.Tagged(map[string]string{"code": "5XX"}).Counter("fetch.errors"),
```
andrewmains12: These are generic HTTP metrics--maybe worth factoring them out and applying them to every endpoint (as middleware)?

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -100,9 +127,13 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	if params.FormatType == models.FormatM3QL {
 		renderM3QLResultsJSON(w, result, params)
+		h.promReadMetrics.fetchSuccess.Inc(1)
+		timer.Stop()
```
andrewmains12: Nit: should probably just defer `timer.Stop()`

benraskin92: so I don't want it to record unless it's a success.

PR #1321, opened 2019-01-28
Add jaeger/opentracing tracing to m3query

andrewmains12:
This diff adds opentracing support to m3query. Spans are started for every HTTP request (courtesy middleware from `github.com/opentracing-contrib/go-stdlib`). I then have child spans started for compilation + execution, with drilldowns into each node of the query. Here's a sample trace:

Query is `sum(increase(coordinator_engine_datapoints{type="fetched"}[10s]))`

jaeger_trace_no_children

One implementation note: node instrumentation is a bit less clean than I would have liked; since all nodes call their children before exiting, i.e.:

```
nextBlock := doStuff()
n.controller.Process(nextBlock)
```
I instrumented them like:

```
startSpan()
nextBlock := doStuff()
stopSpan()

n.controller.Process(nextBlock)
```

The alternative would be to do:

```
startSpan()
nextBlock := doStuff()

n.controller.Process(nextBlock)
stopSpan()
```
but that gives unnecessary nesting imo:

jaeger_tracing_children


src/query/DATAMODEL.md
```diff
@@ -0,0 +1,5 @@
+
```
benraskin92: ??

src/query/api/v1/handler/prometheus/native/read_common.go
```diff
@@ -45,6 +48,15 @@ func read(
 	ctx, cancel := context.WithTimeout(reqCtx, params.Timeout)
```
benraskin92: Maybe delete this ctx if we're not gonna use it? 

`_, cancel := context.WithTimeout(reqCtx, params.Timeout)`

src/query/config/m3query-dev-etcd.yml
```diff
@@ -66,3 +66,6 @@ limits:
     maxFetchedDatapoints: 5000000
   global:
     maxFetchedDatapoints: 500000000
+
```
benraskin92: Should we update any of the other `yaml`/config files?

src/query/executor/request.go
```diff
@@ -108,63 +103,29 @@ func (r *Request) plan(ctx context.Context, nodes parser.Nodes, edges parser.Edg
...
 	}
 
-	sp.finish(nil)
 	return pp, nil
 }
 
-func (r *Request) execute(ctx context.Context, pp plan.PhysicalPlan) (*ExecutionState, error) {
-	sp := startSpan(r.engine.metrics.executingHist, r.engine.metrics.executing)
+func (r *Request) generateExecutionState(ctx context.Context, pp plan.PhysicalPlan) (*ExecutionState, error) {
+	sp, ctx := opentracing.StartSpanFromContext(ctx, "generateExecutionState")
```
benraskin92: Not sure about the naming semantics for opentracing, but do they recommend camel case or using `_` or something else?

src/query/executor/transform/controller.go
```diff
@@ -40,7 +40,16 @@ func (t *Controller) AddTransform(node OpNode) {
 // Process performs processing on the underlying transforms
 func (t *Controller) Process(queryCtx *models.QueryContext, block block.Block) error {
 	for _, ts := range t.transforms {
+
```
benraskin92: I agree - don't put in controller

src/query/executor/transform/exec.go
```diff
@@ -0,0 +1,54 @@
...
+
+	"github.com/opentracing/opentracing-go"
+)
+
+type simpleOpNode interface {
+	OpNode
+	ProcessBlock(queryCtx *models.QueryContext, ID parser.NodeID, b block.Block) (block.Block, error)
+}
+
+// ProcessSimpleBlock is a utility for OpNode instances which simply propagate their data after doing their own
```
benraskin92: Nice comment!

src/query/executor/request.go
```diff
@@ -108,63 +103,29 @@ func (r *Request) plan(ctx context.Context, nodes parser.Nodes, edges parser.Edg
...
+	defer sp.Finish()
```
benraskin92: What should we do about errors and stopping the spans?

src/query/executor/transform/exec_test.go
```diff
@@ -0,0 +1,33 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package transform
+
+import "testing"
+
+func TestProcessSimpleBlock(t *testing.T) {
+	t.Run("closes next block", func(t *testing.T) {
+
```
benraskin92: remove extra line?

src/query/executor/transform/exec_test.go
```diff
@@ -0,0 +1,33 @@
...
+	})
+
+	t.Run("errors on process error", func(t *testing.T) {
+
```
benraskin92: remove extra line?

src/query/functions/fetch.go
```diff
@@ -104,6 +107,12 @@ func (n *FetchNode) Execute(queryCtx *models.QueryContext) error {
 		Enforcer:  queryCtx.Enforcer,
 		Scope:     queryCtx.Scope,
 	})
+}
+
+// Execute runs the fetch node operation
+func (n *FetchNode) Execute(queryCtx *models.QueryContext) error {
+	ctx := queryCtx.Ctx
+	blockResult, err := n.fetch(queryCtx.Ctx, queryCtx)
```
benraskin92: use `ctx` instead of `queryCtx.Ctx`?

src/query/functions/utils/params.go
```diff
@@ -0,0 +1,34 @@
...
+// THE SOFTWARE.
+
+package utils
+
+// StaticParams is a simple string Params implementation, useful for when no data other than OpType is needed.
+type StaticParams string
+
+// String simply returns s.
+func (s StaticParams) String() string {
+	return string(s)
```
benraskin92: Do you need to make this a `string()` or can you just return `s`? Might be a noob go question..

src/query/functions/utils/params.go
```diff
@@ -0,0 +1,34 @@
...
+}
+
+// OpType simply returns s for StaticParams.
+func (s StaticParams) OpType() string {
+	return string(s)
```
benraskin92: Same here

src/query/functions/utils/params.go
andrewmains12: Nah totally reasonable question. https://play.golang.org/p/HOEP-EYYyNC doesn't compile unless you explicitly mark the conversion (though my understanding is that it's a completely compile time check with no runtime implications, since the underlying types are identical) https://golang.org/ref/spec#Conversions

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -94,10 +97,24 @@ func NewPromReadHandler(
...
 	if respErr != nil {
-		xhttp.Error(w, respErr.Err, respErr.Code)
+		logging.ReadContextID(ctx)
+		w.WriteHeader(respErr.Code)
+		xhttp.WriteJSONResponse(w, errorWithID{
+			ErrorResponse: xhttp.ErrorResponse{
+				Error: respErr.Err.Error(),
+			},
+			RqID: logging.ReadContextID(ctx),
+		}, logging.WithContext(ctx))
```
arnikola: Do we need to log id in the error returned to the client? The failing request ID is still logged since the logger has the context id on it

src/cmd/services/m3query/config/config.go
```diff
@@ -50,6 +50,8 @@ type Configuration struct {
 	// Metrics configuration.
 	Metrics instrument.MetricsConfiguration `yaml:"metrics"`
 
+	Tracing instrument.TracingConfiguration `yaml:"tracing"`
```
arnikola: Just curious of what happens if this section is not added to the config; does it just turn tracing off?

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -114,6 +131,7 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 // ServeHTTPWithEngine returns query results from the storage
 func (h *PromReadHandler) ServeHTTPWithEngine(w http.ResponseWriter, r *http.Request, engine *executor.Engine) ([]*ts.Series, models.RequestParams, *RespError) {
 	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
+	sp := opentracingutil.SpanFromContextOrRoot(ctx)
```
arnikola: Might be better to just put this in the error handler, since it's not used aside from error reporting so there's little reason to create it on every request

src/query/api/v1/httpd/handler.go
```diff
@@ -96,14 +99,20 @@ func NewHandler(
 ) (*Handler, error) {
 	r := mux.NewRouter()
 
-	// apply middleware. Just CORS for now, but we could add more here as needed.
-	withMiddleware := &cors.CORSHandler{
+	// apply middleware.
+	var withMiddleware http.Handler = &cors.Handler{
```
arnikola: As a nit, I think this may read slightly cleaner as 
```
corsMiddleware := &cors.Handler{...}
...
withMiddleware := nethttp.Middleware(opentracing.GlobalTracer(), corsMiddleware, ...)

File: src/query/config/m3query-dev-etcd.yml (hunk 3 above)
Comment (arnikola) [reply src/query/config/m3query-dev-etcd.yml]: Actually I like the idea of having it on by default; any chance we could get it into start_m3.sh as well?

File: src/query/executor/engine.go [hunk 17]
```diff
@@ -132,7 +133,7 @@ func (e *Engine) ExecuteExpr(ctx context.Context, parser parser.Parser, opts *En
 	}()
 
 	req := newRequest(e, params)
-	defer req.finish()
+
```
arnikola: Should this still be deferring the finish?

src/query/executor/transform/exec_test.go
```diff
@@ -0,0 +1,143 @@
...
+func (testSimpleNode) Params() parser.Params {
+	panic("implement me")
+}
+
+func (testSimpleNode) ProcessBlock(queryCtx *models.QueryContext, ID parser.NodeID, b block.Block) (block.Block, error) {
+	panic("implement me")
+}
+
+func TestProcessSimpleBlock(t *testing.T) {
+
```
arnikola: nit: extra newline

src/query/executor/transform/exec_test.go
```diff
@@ -0,0 +1,143 @@
...
+type testSimpleNode struct {
+}
+
+func (testSimpleNode) Params() parser.Params {
+	panic("implement me")
+}
+
+func (testSimpleNode) ProcessBlock(queryCtx *models.QueryContext, ID parser.NodeID, b block.Block) (block.Block, error) {
+	panic("implement me")
```
arnikola: I don't think you're using `testSimpleNode` anywhere; remove it?

src/query/executor/transform/exec_test.go
```diff
@@ -0,0 +1,143 @@
...
+		ChildNode   *MockOpNode
+		Node        *MocksimpleOpNode
+		ResultBlock *block.MockBlock
+		SourceBlock block.Block
+		QueryCtx    *models.QueryContext
+	}
+
+	setup := func(t *testing.T) (*testContext, func()) {
+		id := parser.NodeID("foo")
+
```
arnikola: nit: can probably condense some of these newlines

src/query/executor/transform/lazy.go
```diff
@@ -32,6 +33,12 @@ type sinkNode struct {
 	block block.Block
 }
 
+// Params for sinkNode returns an params object for lazy sinks. It doesn't appear in queries (since sink nodes are
```
arnikola: nit: `a params object`; also do we typically add comments for interface functions?

src/query/util/opentracing/context.go
```diff
@@ -0,0 +1,74 @@
...
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
+)
+
+// alias GlobalTracer() so we can mock out the tracer without impacting test's outside the package
```
arnikola: nit: `tests`

src/query/util/opentracing/context.go
```diff
@@ -0,0 +1,74 @@
...
+var getGlobalTracer = opentracing.GlobalTracer
+
+// SpanFromContextOrRoot is the same as opentracing.SpanFromContext, but instead of returning nil, it starts a
```
arnikola: nit: `it starts a a`

src/query/util/opentracing/context.go
```diff
@@ -0,0 +1,74 @@
...
+// a new root span if ctx doesn't already have an associated span, instead of returning nil. Use this over
+// opentracing.StartSpanFromContext if you need access to the current span, but don't want to start a child span.
+// Note: the span returned by this function won't really be configured, and won't be attached to the context; if you
```
arnikola: nit: usually use `NB`

src/query/util/opentracing/context.go
```diff
@@ -0,0 +1,74 @@
...
+	var span opentracing.Span
+	if parentSpan := opentracing.SpanFromContext(ctx); parentSpan != nil {
+		opts = append(opts, opentracing.ChildOf(parentSpan.Context()))
+		tracer := parentSpan.Tracer()
+		span = tracer.StartSpan(operationName, opts...)
+	} else {
+		tracer := getGlobalTracer()
+		span = tracer.StartSpan(operationName, opts...)
+	}
+	return span, opentracing.ContextWithSpan(ctx, span)
```
arnikola: nit: newline

src/query/util/opentracing/context.go
```diff
@@ -0,0 +1,74 @@
...
+// a new root span if ctx doesn't already have an associated span, instead of returning nil. Use this over
+// opentracing.StartSpanFromContext if you need access to the current span, but don't want to start a child span.
+// Note: the span returned by this function won't really be configured, and won't be attached to the context; if you
+// want a proper span, start one at the root and pass it in.
+func SpanFromContextOrRoot(ctx context.Context) opentracing.Span {
+	sp := opentracing.SpanFromContext(ctx)
+	if sp != nil {
+		return sp
+	}
+	return opentracing.StartSpan("SpanFromContextOrRoot - dummy")
```
arnikola: nit: newline

src/query/test/executor/transform.go
```diff
@@ -48,6 +49,11 @@ type SinkNode struct {
 	Metas  []block.SeriesMeta
 }
 
+// Params returns the Params associated with this node.
```
arnikola: It's a bit inconsistent with commenting on interface methods; comments missing on a few of these `Params()` methods

I'd err on the side of generally removing them unless the particular `Params` method is doing something tricky and needs further explanation

src/query/server/server.go
```diff
@@ -125,9 +130,24 @@ func Run(runOpts RunOptions) {
...
+
+	tracer, traceCloser, err := cfg.Tracing.NewTracer(serviceName, scope, logger)
+	if err != nil {
+		logger.Fatal("could not initialize tracing", zap.Error(err))
+	}
+
+	defer traceCloser.Close()
+
+	if _, ok := tracer.(opentracing.NoopTracer); ok {
+		logger.Info("tracing disabled; set `tracing.backend` to enable")
```
arnikola: 👍

src/query/functions/fetch.go
```diff
@@ -86,15 +88,16 @@ func (o FetchOp) Node(controller *transform.Controller, storage storage.Storage,
 	}
 }
 
-// Execute runs the fetch node operation
-func (n *FetchNode) Execute(queryCtx *models.QueryContext) error {
+func (n *FetchNode) fetch(queryCtx *models.QueryContext) (block.Result, error) {
+	ctx := queryCtx.Ctx
+	sp, _ := opentracingutil.StartSpanFromContext(ctx, "fetch")
```
arnikola: Just curious why we're dropping the context from here? Shouldn't we update ctx to `sp, ctx := opentracingutil.StartSpanFromContext(ctx, "fetch")`?

src/query/functions/aggregation/count_values.go
```diff
@@ -88,6 +88,10 @@ type countValuesNode struct {
 	controller *transform.Controller
 }
 
+func (n *countValuesNode) Params() parser.Params {
```
arnikola: Shouldn't this function (and other snowflakes existing outside of various `base.go` files) be refactored to the `ProcessBlock(queryCtx *models.QueryContext, ID parser.NodeID, b block.Block)` model?

src/query/executor/transform/exec.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/m3db/m3/src/query/parser"
+	"github.com/m3db/m3/src/query/util/opentracing"
+)
+
+type simpleOpNode interface {
+	Params() parser.Params
+	ProcessBlock(queryCtx *models.QueryContext, ID parser.NodeID, b block.Block) (block.Block, error)
+}
+
+// ProcessSimpleBlock is a utility for OpNode instances which simply propagate their data after doing their own
```
arnikola: Doesn't this describe every block type?

src/query/executor/transform/exec.go
```diff
@@ -0,0 +1,53 @@
...
+package transform
+
+import (
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/parser"
+	"github.com/m3db/m3/src/query/util/opentracing"
+)
+
+type simpleOpNode interface {
```
arnikola: It's a bit weird having an interface that a bunch of nodes satisfy be private

src/query/executor/transform/exec_test.go
```diff
@@ -0,0 +1,143 @@
...
+	t.Run("closes next block", func(t *testing.T) {
+		tctx, closer := setup(t)
+		defer closer()
+
+		configureSuccessfulNode(tctx)
+
+		require.NoError(t, doCall(tctx))
+	})
+
+	t.Run("errors on process error", func(t *testing.T) {
```
arnikola: nit: same comment about the `t.Run`, seems to swallow up useful information for failing tests

src/query/functions/temporal/base.go
```diff
@@ -197,7 +202,21 @@ func (c *baseNode) Process(queryCtx *models.QueryContext, ID parser.NodeID, b bl
...
-	return c.processCompletedBlocks(processRequests, maxBlocks)
+	blocks, err := c.processCompletedBlocks(queryCtx, processRequests, maxBlocks)
+	defer closeBlocks(blocks)
+
+	if err != nil {
+		return err
+	}
+
+	return fmt.Errorf("Asdf")
+	// return c.propagateNextBlocks(processRequests, blocks, maxBlocks)
```
arnikola: Might want to revert this 😛

src/query/functions/aggregation/count_values.go
arnikola: Yeah it's a bit annoying unfortunately; I guess it's a downside that we have to deal with in return for the nice properties of functions calling their own downstream processing

andrewmains12: Yeah for a brief moment I was like "we should just explicitly walk the graph instead" and then I realized that would be a large rewrite 😄

docs/operational_guide/monitoring.md
```diff
@@ -0,0 +1,74 @@
...
+[docs](https://www.jaegertracing.io/docs/1.9/getting-started/).
+
+
+#### Finding slow queries
+
+To find prom queries longer than <threshold>, filter for `minDuration >= <threshold>` on
+`operation="GET /api/v1/query_range"`.
+
+Sample query:
+http://localhost:16686/search?end=1548876672544000&limit=20&lookback=1h&maxDuration&minDuration=1ms&operation=GET%20%2Fapi%2Fv1%2Fquery_range&service=m3query&start=1548873072544000
```
arnikola: nit: may be nicer to wrap it in `\`\`` characters otherwise it'll be a real link :)

src/query/api/v1/handler/prometheus/native/read.go
arnikola: Sure, doesn't seem like it would cause issues. We've only got tracing on read right? Might want to add the same thing to `read_instantaneous`, too... Would it make sense to take this block and put it into `xhttp` as a method like `WriteJSONResponseWithRequestID`?

src/query/config/m3query-dev-etcd.yml
arnikola: That would be pretty cool :)

Not sure how much effort that would entail; if it's a larger task maybe just throw up an Issue for it and leave it for future work?

src/query/executor/engine.go
arnikola: Maybe add a quick comment saying that downstream methods will be closing the request?

Also what happens if we hit an error on `req.compile` or `req.plan`, will we miss the finish call?

src/query/executor/transform/exec_test.go
arnikola: Now that I think about it, mostly had problems with it in tabled tests, when iterating over them... when one of twenty fails it's sometimes a pain to find exactly which test failed exactly where.

src/query/executor/transform/lazy.go
arnikola: Think this one may have fallen through the cracks (thanks github for collapsing comments for some reason 🙄 )

src/query/functions/aggregation/count_values.go
arnikola: Yeah, was looking at making a few nicety fixes to DAG generation/execution/mapping, and it's a pretty hefty beast unfortunately to make it more usable for things like this (although it would definitely not go astray!)

src/query/functions/aggregation/take.go
```diff
@@ -111,11 +111,19 @@ type takeNode struct {
...
+func (n *takeNode) Params() parser.Params {
+	return n.op
+}
+
 // Process the block
 func (n *takeNode) Process(queryCtx *models.QueryContext, ID parser.NodeID, b block.Block) error {
+	return transform.ProcessSimpleBlock(n, n.controller, queryCtx, ID, b)
+}
+
+func (n *takeNode) ProcessBlock(queryCtx *models.QueryContext, ID parser.NodeID, b block.Block) (block.Block, error) {
```
arnikola: 👍

src/query/functions/binary/base.go
```diff
@@ -142,6 +147,14 @@ func (n *baseNode) Process(queryCtx *models.QueryContext, ID parser.NodeID, b bl
 	return n.controller.Process(queryCtx, nextBlock)
 }
 
+func (n *baseNode) processWithTracing(queryCtx *models.QueryContext, lhs block.Block, rhs block.Block) (block.Block, error) {
```
arnikola: Correct me if I'm wrong, but this will show up on the span as two function calls, one of which will be pretty much instantaneous, right? Should be fine, and probably not worth overengineering here to avoid that case (like starting the span only when both LHS and RHS satisfied), but just trying to get an idea of how it will look :)

arnikola: That being said, it might be necessary to add a parent span to `NewScalarOp`, since there is a potential (highly unlikely, but still) scenario where you have a binary function between a series and a scalar, e.g. `up + 1`, and somehow the `up` block resolves first, which may cut off your trace early if I'm understanding the flow correctly?

src/query/util/opentracing/context_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"testing"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/mocktracer"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestStartSpanFromContext(t *testing.T) {
+	t.Run("uses local tracer if available", func(t *testing.T) {
```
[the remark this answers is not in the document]
arnikola: Explained issue above, doesn't seem like a relevant concern here :)

src/query/functions/temporal/base.go
```diff
@@ -197,7 +202,20 @@ func (c *baseNode) Process(queryCtx *models.QueryContext, ID parser.NodeID, b bl
 		}
 	}
 
-	return c.processCompletedBlocks(processRequests, maxBlocks)
+	blocks, err := c.processCompletedBlocks(queryCtx, processRequests, maxBlocks)
+	defer closeBlocks(blocks)
+
+	if err != nil {
```
arnikola: So probably(definitely!) out of scope for this PR, but would be good to revisit temporal functions at some point and add in some logic that if any incoming block errored, it should drop any incoming blocks with the same error instead of trying to add them and starting processing before hitting it... May add an issue regarding this?

src/query/functions/temporal/base.go
```diff
@@ -240,17 +260,36 @@ func (c *baseNode) processCompletedBlocks(processRequests []processRequest, maxB
...
-func (c *baseNode) processSingleRequest(request processRequest) error {
+// processCompletedBlocks processes all blocks for which all dependent blocks are present
+func (c *baseNode) processCompletedBlocks(queryCtx *models.QueryContext, processRequests []processRequest, maxBlocks int) ([]block.Block, error) {
+	sp, _ := opentracingutil.StartSpanFromContext(queryCtx.Ctx, c.op.OpType())
+	defer sp.Finish()
+
+	blocks := make([]block.Block, len(processRequests))
+	for i, req := range processRequests {
+		bl, err := c.processSingleRequest(req)
+
```
arnikola: nit: delete this newline and add it between closing brace and `blocks[i]`

src/query/api/v1/handler/prometheus/native/read.go
andrewmains12: Yeah good call, I shouldn't have been lazy about it here :). I'll factor it out.

src/query/executor/transform/exec_test.go
andrewmains12: Ah yes, totally agreed on that! I always try to use searchable names in those cases, which helps, but it can still be annoying.

docs/operational_guide/monitoring.md
andrewmains12: Feature not a bug I think; if you're running jaeger locally, you should just be able to click and do the search.

src/query/functions/binary/base.go
andrewmains12: > Correct me if I'm wrong, but this will show up on the span as two function calls, one of which will be pretty much instantaneous, right?

Ah yeah good question. It actually shows up as a single trace, since I am in fact starting the span only when both LHS and RHS are satisfied.

Here's a sample:

screen shot 2019-02-15 at 3 21 15 pm

arnikola: Nice!

src/query/functions/temporal/base.go
arnikola: The general function of temporal functions is that blocks come in, and are added to a cache. When enough blocks come in to fully calculate a downstream block, it is processed (which can error).  At the moment, if the processing fails, that particular request will fail but there will still be in-flight blocks coming into the temporal function, which do not necessarily need to be computed since the entire query will error anyway

andrewmains12: Ah ok cool, thanks for the clarification. That seems like a kind of generic problem--other parts of the query should be cancelled once one part fails. Do we have any general cancellation mechanism already? I suppose we could check the context periodically during each node's `Process`; that would probably be the idiomatic way to do it. No idea how often "periodically" should be though haha.

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -167,6 +170,9 @@ func (h *PromReadHandler) ServeHTTPWithEngine(
 
 	result, err := read(ctx, engine, h.tagOpts, w, params)
 	if err != nil {
+		sp := opentracingutil.SpanFromContextOrRoot(ctx)
+		sp.LogFields(opentracinglog.Error(err))
+		opentracingext.Error.Set(sp, true)
```
arnikola: Should you have `sp.Finish()` here?

src/query/api/v1/handler/prometheus/native/read_common.go
```diff
@@ -45,6 +48,15 @@ func read(
...
 	defer cancel()
 
+	sp := opentracingutil.SpanFromContextOrRoot(ctx)
+	sp.LogFields(
+		opentracinglog.String("params.query", params.Query),
+		opentracingutil.Time("params.start", params.Start),
+		opentracingutil.Time("params.end", params.End),
+		opentracingutil.Time("params.now", params.Now),
+		opentracingutil.Duration("params.step", params.Step),
+	)
```
arnikola: Should you have `sp.Finish()` here?

src/query/api/v1/httpd/handler.go
```diff
@@ -102,13 +105,19 @@ func NewHandler(
 ) (*Handler, error) {
 	r := mux.NewRouter()
 
-	// apply middleware. Just CORS for now, but we could add more here as needed.
-	withMiddleware := &cors.Handler{
+	// apply middleware.
```
arnikola: nit: maybe change comment to `// Add CORS to endpoint middleware`?

src/query/api/v1/httpd/handler.go
```diff
@@ -102,13 +105,19 @@ func NewHandler(
...
+	withMiddleware := http.Handler(&cors.Handler{
 		Handler: r,
 		Info: &cors.Info{
 			"*": true,
 		},
-	}
+	})
+
+	// add jaeger tracing to our endpoints
```
arnikola: nit: maybe `// Add Jaeger tracing to endpoint middleware`?

docs/operational_guide/monitoring.md
```diff
@@ -0,0 +1,74 @@
...
+
+TODO
+
+## Tracing
+
+M3DB is integrated with [opentracing](https://opentracing.io/) to provide
+insight into query performance and errors.
+
+### Configuration
+Currently, only https://www.jaegertracing.io/ is supported as a backend.
```
arnikola: nit: maybe `only [Jaeger](https://www.jaegertracing.io/) is supported as a tracing backend.`

docs/operational_guide/monitoring.md
```diff
@@ -0,0 +1,74 @@
...
+
+To enable it, set tracing.backend to "jaeger":
+
+```
+tracing:
```
arnikola: Do we have a config with this in it? Might be good to link it here if so

docs/operational_guide/monitoring.md
```diff
@@ -0,0 +1,74 @@
...
+#### Finding queries with errors
+
+Search for `error=true` on `operation="GET /api/v1/query_range"`
+http://localhost:16686/search?operation=GET%20%2Fapi%2Fv1%2Fquery_range&service=m3query&tags=%7B%22error%22%3A%22true%22%7D
+
+#### Finding 500 (Internal Server Error) responses
+
+Search for `http.status_code=500`.
+
+http://localhost:16686/search?limit=20&lookback=24h&maxDuration&minDuration&operation=GET%20%2Fapi%2Fv1%2Fquery_range&service=m3query&start=1548802430108000&tags=%7B"http.status_code"%3A"500"%7D
```
arnikola: nit: Could be worth adding some screenshots of your local traces for some of these to illustrate what's happening, and maybe demonstrate value to people unfamiliar with tracing?

docs/operational_guide/monitoring.md
```diff
@@ -0,0 +1,74 @@
...
+
+To enable it, set tracing.backend to "jaeger":
+
+```
+tracing:
+    backend: jaeger  # enables jaeger with default configs
+    jaeger:
+        # optional configuration for jaeger -- see
+        # https://github.com/jaegertracing/jaeger-client-go/blob/master/config/config.go#L37
+        # for options
```
arnikola: Worth calling out that sampling is heavily encouraged? Otherwise it's likely to create a ton of data

src/query/util/opentracing/context.go
```diff
@@ -0,0 +1,76 @@
...
+// opentracing.StartSpanFromContext if you need access to the current span, but don't want to start a child span.
+// NB: the span returned by this function won't really be configured, and won't be attached to the context; if you
+// want a proper span, start one at the root and pass it in.
+func SpanFromContextOrRoot(ctx context.Context) opentracing.Span {
+	sp := opentracing.SpanFromContext(ctx)
+	if sp != nil {
+		return sp
+	}
+
+	return opentracing.StartSpan("SpanFromContextOrRoot - dummy")
```
arnikola: Where does this string appear re: trace? Could be good to make it a constant if nothing else

src/query/util/logging/log.go
```diff
@@ -127,8 +128,14 @@ func withResponseTimeLoggingFunc(
 	return func(w http.ResponseWriter, r *http.Request) {
 		startTime := time.Now()
 		rqCtx := NewContextWithGeneratedID(r.Context())
+		rqID := ReadContextID(rqCtx)
```
arnikola: nit: Move this into the `if sp != nil {..` block since it's not used if sp is nil?

src/query/util/httperrors/errors.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/m3db/m3/src/query/util/logging"
+	"github.com/m3db/m3/src/x/net/http"
+)
+
+type errorWithID struct {
+	xhttp.ErrorResponse
+	RqID string `json:"rqID"`
+}
+
+// ErrorWithReqID writes an xhttp.ErrorResponse with an added request id (RqId) field read from the request
```
arnikola: nit1: could be better to call it `ErrorWithReq` rather than specifying it must be a reqID?

nit2: Split the NB out
```
// context.
//
// NB: ...
```

docs/operational_guide/monitoring.md
```diff
@@ -0,0 +1,74 @@
+## Metrics
+
+TODO
+
+## Logs
+
+TODO
```
arnikola: Probably better to remove these?

src/query/api/v1/httpd/handler.go
andrewmains12: Ah yeah I meant the comment to be "apply all middleware". I'll just factor out a function.

docs/operational_guide/monitoring.md
andrewmains12: Sure, though I think I'll do that as a separate PR; this one's got enough files already :)

src/query/api/v1/httpd/handler.go
```diff
@@ -139,6 +136,23 @@ func NewHandler(
 	return h, nil
 }
 
+func applyMiddleware(base http.Handler) http.Handler {
+	rtn := http.Handler(&cors.Handler{
```
benraskin92: Better var name?

src/query/api/v1/httpd/handler.go
```diff
@@ -139,6 +136,23 @@ func NewHandler(
...
+		Handler: base,
+		Info: &cors.Info{
+			"*": true,
+		},
+	})
+
+	// apply jaeger middleware,which will start a span
```
benraskin92: `middleware,which` > `middleware, which`

src/query/util/opentracing/context.go
```diff
@@ -0,0 +1,80 @@
...
+// See https://github.com/opentracing/opentracing-go/issues/149 for more discussion.
+func StartSpanFromContext(ctx context.Context, operationName string, opts ...opentracing.StartSpanOption) (opentracing.Span, context.Context) {
+	var span opentracing.Span
+	if parentSpan := opentracing.SpanFromContext(ctx); parentSpan != nil {
+		opts = append(opts, opentracing.ChildOf(parentSpan.Context()))
+		tracer := parentSpan.Tracer()
+		span = tracer.StartSpan(operationName, opts...)
+	} else {
+		tracer := getGlobalTracer()
+		span = tracer.StartSpan(operationName, opts...)
```
benraskin92: Pull out `span = tracer.StartSpan(operationName, opts...)` since its in both conditional statements?

src/query/util/opentracing/context_test.go
```diff
@@ -0,0 +1,105 @@
...
+		sp, ctx := StartSpanFromContext(context.Background(), "foo")
+		sp.Finish()
+
+		assertHasSpans(t, mtr, []string{"foo"})
+		assert.NotNil(t, opentracing.SpanFromContext(ctx))
+	})
+}
+
+func TestStartSpanFromContextOrRoot(t *testing.T) {
+
```
benraskin92: nit: newline

src/query/util/httperrors/errors.go
```diff
@@ -0,0 +1,50 @@
...
+	RqID string `json:"rqID"`
+}
+
+// ErrorWithReqInfo writes an xhttp.ErrorResponse with an added request id (
+// RqId) field read from the request context.
+//
+// NB: RqID is currently a query specific concept,
+// which is why this doesn't exist in xhttp proper.
+// We can add it later if we propagate the request id concept to that package as well.
+func ErrorWithReqInfo(w http.ResponseWriter, r *http.Request, err error, code int) {
```
benraskin92: nit: i think Rob told me once that if error is an argument in a function, it should come last 🤷‍♂️

PR #1322, opened 2019-01-28
Make lookback duration configurable

benraskin92:


src/query/models/params.go
```diff
@@ -34,10 +34,10 @@ const (
...
 )
 
-// LookbackDelta determines the time since the last sample after which a time
-// series is considered stale (inclusive).
-// TODO: Make this configurable
-var LookbackDelta = time.Minute
+// // LookbackDelta determines the time since the last sample after which a time
+// // series is considered stale (inclusive).
+// // TODO: Make this configurable
+// var LookbackDelta = time.Minute
```
benraskin92: Remove this commented out code

src/query/server/server.go
```diff
@@ -77,6 +77,7 @@ var (
 	}
 
 	defaultCarbonIngesterWorkerPoolSize = 1024
+	defaultLookbackDuration             = 5 * time.Minute
```
benraskin92: This is what Prom uses

src/query/server/server.go
```diff
@@ -162,6 +163,10 @@ func Run(runOpts RunOptions) {
 		logger.Fatal("could not create tag options", zap.Error(err))
 	}
 
+	if cfg.LookbackDuration == nil {
+		cfg.LookbackDuration = &defaultLookbackDuration
```
arnikola: This should be validated to be >= 0

benraskin92: Good call

arnikola: nit: Might be better to put this check on a validate function, but not really a big deal

robskillington: Can you add this as a method on the config object?

i.e.
```go
func (c Configuration) LookbackDurationOrDefault() (time.Duration, error) {
  if c.LookbackDuration == nil {
    return DefaultLookbackDuration, nil
  }
  v := *c.LookbackDuration
  if v < 0 {
    return nil, anError
  }
  return v, nil
}
```

src/query/server/server.go
```diff
@@ -173,11 +172,11 @@ func Run(runOpts RunOptions) {
...
 
-	if cfg.LookbackDuration == nil {
-		cfg.LookbackDuration = &defaultLookbackDuration
-	} else if *cfg.LookbackDuration < 0 {
-		logger.Fatal("lookbackDuration must be > 0")
+	lookbackDuration, err := cfg.LookbackDurationOrDefault()
+	if err != nil {
+		logger.Fatal("error validating LookbackDuration", zap.Error(err))
 	}
+	cfg.LookbackDuration = &lookbackDuration
```
arnikola: nit newline

PR #1325, opened 2019-01-29
[query] Add build exporter metrics to m3query

arnikola:


src/query/server/server.go
```diff
@@ -140,6 +140,14 @@ func Run(runOpts RunOptions) {
 		}
 	}()
 
+	iopts := instrumentOptions.SetMetricsScope(
```
richardartoul: won't this subscope all the metrics for the whole service with `build_info`?

src/query/server/server.go
```diff
@@ -140,6 +140,14 @@ func Run(runOpts RunOptions) {
...
+		instrumentOptions.MetricsScope().SubScope("build_info"))
+	buildReporter := instrument.NewBuildReporter(iopts)
+	if err := buildReporter.Start(); err != nil {
+		logger.Fatal("could not connect to metrics", zap.Error(err))
```
richardartoul: maybe: `could not start build reporter`

src/query/server/server.go
richardartoul: Oh I see its a new var. Yeah either dont subscope or just give the var a name that implies it should not be used for anything other than the build reporter

PR #1327, opened 2019-01-29
Improve carbon ingestion performance

richardartoul:


src/cmd/services/m3coordinator/ingest/carbon/ingest.go
```diff
@@ -41,13 +41,20 @@ import (
...
+	"github.com/m3db/m3x/pool"
 	m3xserver "github.com/m3db/m3x/server"
 	xsync "github.com/m3db/m3x/sync"
 	xtime "github.com/m3db/m3x/time"
 
 	"github.com/uber-go/tally"
 )
 
+const (
+	maxResourcePoolNameSize = 200
```
robskillington: Maybe bump this to 256 for "power of 2-ness".  Also IDs can get long in graphite land, heh.

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
```diff
@@ -138,16 +165,20 @@ func (i *ingester) Handle(conn net.Conn) {
 	logger.Debug("handling new carbon ingestion connection")
 	for s.Scan() {
 		name, timestamp, value := s.Metric()
-		// TODO(rartoul): Pool.
+
+		resources := i.getLineResources()
 		// Copy name since scanner bytes are recycled.
-		name = append([]byte(nil), name...)
+		resources.name = append(resources.name, name...)
```
robskillington: Maybe `append(resources.name[:0], name...)` to be safe

scripts/development/m3_stack/docker-compose.yml
```diff
@@ -11,6 +11,10 @@ services:
       - "./m3dbnode.yml:/etc/m3dbnode/m3dbnode.yml"
     environment:
       - M3DB_HOST_ID=m3db_seed
+    expose:
+      - "9004"
+    ports:
+      - "0.0.0.0:9004:9004"
```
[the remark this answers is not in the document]
richardartoul: sure

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
```diff
@@ -106,7 +112,23 @@ func NewIngester(
...
 
+	poolOpts := pool.NewObjectPoolOptions().
+		SetInstrumentOptions(opts.InstrumentOptions).
+		SetRefillLowWatermark(0).
+		SetRefillHighWatermark(0).
+		SetSize(defaultResourcePoolSize)
+
+	resourcePool := pool.NewObjectPool(poolOpts)
+	resourcePool.Init(func() interface{} {
+		return lineResources{
```
robskillington: Think it's better to make this `&lineResources{...}` just in case when it gets boxed back to `interface{}` during the `putLineResources(...)` method it gets reallocated on the heap (even though it would just be the wrapping struct itself needing to be reallocated)?

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
richardartoul: should I go higher than 256?

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
```diff
@@ -115,17 +137,22 @@ func NewIngester(
 			opts.InstrumentOptions.MetricsScope()),
 
 		rules: compiledRules,
+
+		lineResourcesPool: resourcePool,
 	}, nil
 }
 
 type ingester struct {
+	ctx                  context.Context
```
robskillington: Can we make this per connection, e.g. at top of `Handle(...)` we have a `ctx := context.Background()`?  That way at least each connection gets its own unique context.  (And there's not likely to be more than hundreds of connections, so that should avoid it being alloc heavy).

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
richardartoul: benchmarked it and you're right, will fix

src/cmd/services/m3coordinator/ingest/carbon/ingest.go
```diff
@@ -356,6 +404,34 @@ func compileRules(rules CarbonIngesterRules) ([]ruleAndRegex, error) {
...
+		cap(l.tags) > maxPooledTagsSize
+
+	if tooLargeForPool {
+		return
+	}
+
+	// Reset.
+	l.name = l.name[:0]
+	l.datapoints[0] = ts.Datapoint{}
+	l.tags = l.tags[:0]
```
robskillington: nit: Might want to free the tag pointers before setting to `[:0]`, i.e.
```go
for i := range l.tags {
  l.tags[i] = models.Tag{}
}
l.tags = l.tags[:0]
```

src/cmd/tools/carbon_load/main/main.go
```diff
@@ -0,0 +1,146 @@
...
+
+			wg.Done()
+		}()
+	}
+
+	start := time.Now()
+	go func() {
+		time.Sleep(*duration)
+
+		fmt.Println("Beginning shutdown...")
```
robskillington: nit: For consistency, use lowercase line messages?

src/cmd/tools/carbon_load/main/main.go
```diff
@@ -0,0 +1,146 @@
...
+		close(closeCh)
+	}()
+
+	wg.Wait()
+
+	durationOfBench := time.Now().Sub(start).Seconds()
+	fmt.Println("Average QPS: ", float64(numSuccess)/durationOfBench)
```
robskillington: nit: For consistency, use lowercase line messages?

PR #1328, opened 2019-01-30
[query] Fanout options to select namespaces

arnikola:


src/query/graphite/storage/m3_wrapper.go
```diff
@@ -130,8 +130,14 @@ func (s *m3WrappedStore) FetchByQuery(
 	m3ctx, cancel := context.WithTimeout(ctx.RequestContext(), opts.Timeout)
 	defer cancel()
 
-	m3result, err := s.m3.Fetch(m3ctx, m3query,
-		storage.NewFetchOptions())
+	fetchOptions := storage.NewFetchOptions()
+	fetchOptions.FanoutOptions = &storage.FanoutOptions{
+		FanoutUnaggregated:        storage.FanoutForceDisable,
+		FanoutAggregated:          storage.FanoutForceEnable,
+		FanoutAggregatedOptimized: storage.FanoutForceEnable,
```
robskillington: I think we want this to be `FanoutForceDisable` for the optimized flag.

src/query/storage/m3/storage.go
```diff
@@ -507,27 +512,46 @@ func (s *m3storage) writeSingle(
...
-	unaggregatedRetention := unaggregated.Options().Attributes().Retention
-	unaggregatedStart := now.Add(-1 * unaggregatedRetention)
-	if unaggregatedStart.Before(start) || unaggregatedStart.Equal(start) {
-		// Highest resolution is unaggregated, return if it can fulfill it
-		return namespaceCoversAllQueryRange, ClusterNamespaces{unaggregated}, nil
+	type unaggregatedNamespaceDetails struct {
+		unaggregated ClusterNamespace
+		retention    time.Duration
+		enabled      bool
+	}
```
robskillington: Maybe move this type to top of this file?

scripts/docker-integration-tests/carbon/test.sh
```diff
@@ -57,5 +57,3 @@ t=$(date +%s)
 echo "foo.min.catch-all.baz 10 $t" | nc 0.0.0.0 7204
 echo "foo.min.catch-all.baz 20 $t" | nc 0.0.0.0 7204
 echo "Attempting to read mean aggregated carbon metric"
-ATTEMPTS=10 TIMEOUT=1 retry_with_backoff read_carbon foo.min.catch-all.baz 15
```
arnikola: Accidentally deleted

src/query/storage/m3/cluster_resolver.go
```diff
@@ -0,0 +1,299 @@
...
+	if opts.FanoutUnaggregated == storage.FanoutForceDisable {
+		return unaggregatedNamespaceDetails{satisfies: disabled}
+	}
+
+	retention := unaggregated.Options().Attributes().Retention
+	unaggregatedStart := now.Add(-1 * retention)
+	if unaggregatedStart.Before(start) || unaggregatedStart.Equal(start) {
+		return unaggregatedNamespaceDetails{
+			clusterNamespace: unaggregated,
+			satisfies:        fullySatisfiesRange,
```
robskillington: nit: Is it worth including `retention` in return for this code path?  In case it's needed in future and we don't find out that it's set to zero with tests or something?

src/query/storage/m3/cluster_resolver.go
```diff
@@ -0,0 +1,299 @@
...
+			ClusterNamespaces{unaggregated.clusterNamespace},
+			nil
+	}
+
+	// The filter function will drop namespaces which do not cover the entire
+	// query range from contention.
+	//
+	// NB: if fanout aggregation is forced on, the filter instead forces clusters
+	// that do not cover the range to be set as partially aggregated.
+	filterFunc := func(namespace ClusterNamespace) bool {
```
robskillington: nit: Maybe we call this `coversRangeFilter` to declare intent?

src/query/graphite/storage/m3_wrapper.go
```diff
@@ -129,9 +129,14 @@ func (s *m3WrappedStore) FetchByQuery(
 	m3query := translateQuery(query, opts)
 	m3ctx, cancel := context.WithTimeout(ctx.RequestContext(), opts.Timeout)
 	defer cancel()
+	fetchOptions := storage.NewFetchOptions()
+	fetchOptions.FanoutOptions = &storage.FanoutOptions{
+		FanoutUnaggregated:        storage.FanoutForceDisable,
+		FanoutAggregated:          storage.FanoutForceEnable,
```
robskillington: Hm, should we specifying Default perhaps here?

src/query/storage/m3/cluster_resolver.go
```diff
@@ -0,0 +1,299 @@
...
+				result = append(result, n)
+			}
+		}
+
+		return namespaceCoversAllQueryRange, result, nil
+	}
+
+	// No need to regenerate aggregated namespace without a filter if aggregated
+	// fanout is force enabled.
+	if opts.FanoutAggregated != storage.FanoutForceEnable {
```
robskillington: Hm isn't the default to enable fanning out to aggregated namespaces in general?  

i.e. if `ForceEnable` and `Default` should both always return aggregated namespaces, and for `FanoutAggregatedOptimized` that should just determine whether to include or not include namespaces with `downsampleOpts.All` into the complete or partial bucket.

So I would've thought for `opts.FanoutAggregated == FanoutForceDisable` we would just always return the unaggregated namespace, and no namespaces if there is no unaggregated namespace at all.  And then we shouldn't check this flag any later I would've thought?

src/query/storage/m3/cluster_resolver.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	if len(r.completeAggregated) == 0 {
+		// Absolutely no complete aggregated namespaces, need to fanout to all
+		// partial aggregated namespaces as well as the unaggregated cluster
+		// as we have no idea which has the longest retention.
+		result := r.partialAggregated
+		// If unaggregated namespace can partially satisfy this range, add it as a
+		// fanout contender.
+		if unaggregated.satisfies == partiallySatisfiesRange {
```
robskillington: Is this check necessary?  If it fully satisifies the query range it would've already be returned on line 87 as part of the first check no?

src/query/storage/m3/cluster_resolver.go
```diff
@@ -0,0 +1,299 @@
...
+			result = append(result, unaggregated.clusterNamespace)
+		}
+
+		// If any namespace currently in contention does not cover the entire query
+		// range, set query fanout type to namespaceCoversPartialQueryRange.
+		for _, n := range result {
+			clusterStart := now.Add(-1 * n.Options().Attributes().Retention)
+			if clusterStart.After(start) {
```
robskillington: Unfortunately you need to do `clusterStart.Equal(..) || clusterStart.After(...)` to take into account the case where the start and cluster start is an exact match.

src/query/storage/m3/cluster_resolver.go
```diff
@@ -0,0 +1,299 @@
...
+	all ClusterNamespaces,
+	slices reusedAggregatedNamespaceSlices,
+	filter func(ClusterNamespace) bool,
+	opts *storage.FanoutOptions,
+) reusedAggregatedNamespaceSlices {
+	// Reset reused slices.
+	slices = slices.reset(len(all))
+
+	// Force disable fanout to any aggregated namespaces.
+	if opts.FanoutAggregated == storage.FanoutForceDisable {
```
robskillington: As discussed before, if `FanoutAggregated` is actually disabled, then we can just probably return immediately after checking the unaggregated namespace. (i.e. return unaggregated as long as that is default or enabled) and then don't need to check it again at all.

src/query/storage/m3/cluster_resolver.go
arnikola: Yeah, it's to check that it's not `disabled` here... I didn't want to use two bools like `enabled, coversRange bool` since they're kinda linked

src/query/storage/m3/cluster_resolver.go
arnikola: Sure, sounds good.

src/query/storage/m3/cluster_resolver.go
arnikola: Good call.

src/query/storage/m3/cluster_resolver.go
arnikola: Ah damn I thought I flipped it to `!clusterStart.Before(..)`

src/query/storage/m3/cluster_resolver.go
arnikola: Yeah, it gets to the same conclusion but with more steps; will try and short circuit it instead

src/query/storage/m3/cluster_resolver.go
```diff
@@ -0,0 +1,299 @@
...
+	// have all the data).
+	for _, namespace := range all {
+		nsOpts := namespace.Options()
+		if nsOpts.Attributes().MetricsType != storage.AggregatedMetricsType {
+			// Not an aggregated cluster.
+			continue
+		}
+
+		// If fanout is forced, ignore the filter.
+		if opts.FanoutAggregated != storage.FanoutForceEnable {
```
robskillington: I was thinking we would not actually need to skip the check here, since if `FanoutAggregatedOptimized` is forced disabled, then you'll always run this function with `filter == nil` because you won't take the path of `if len(r.completeAggregated) > 0` after running the first call to `aggregatedNamespaces(...)`.

src/query/storage/m3/storage.go
```diff
@@ -52,6 +53,7 @@ type queryFanoutType uint
 const (
 	namespaceCoversAllQueryRange queryFanoutType = iota
 	namespaceCoversPartialQueryRange
+	namespaceInvalid
```
robskillington: nit: Maybe make this the first value in the enum?

src/query/storage/m3/cluster_resolver.go
```diff
@@ -0,0 +1,293 @@
...
+		clusters.UnaggregatedClusterNamespace(), opts)
+	if unaggregated.satisfies == fullySatisfiesRange {
+		return namespaceCoversAllQueryRange,
+			ClusterNamespaces{unaggregated.clusterNamespace},
+			nil
+	}
+
+	if opts.FanoutAggregated == storage.FanoutForceDisable {
+		if unaggregated.satisfies == partiallySatisfiesRange {
+			return namespaceCoversAllQueryRange,
```
robskillington: nit: Probably better to return `namespaceCoversPartialQueryRange` here since it only partially satisfies the range?

PR #1333, opened 2019-01-30
Update m3db single node docs

benraskin92:


docs/how_to/single_node.md
```diff
@@ -16,9 +16,11 @@ directory on your host for durability:
 
 ```
 docker pull quay.io/m3/m3dbnode:latest
-docker run -p 7201:7201 -p 9003:9003 --name m3db -v $(pwd)/m3db_data:/var/lib/m3db quay.io/m3/m3dbnode:latest
+docker run -p 7201:7201 -p 7203:7203 -p 9003:9003 --name m3db -v $(pwd)/m3db_data:/var/lib/m3db -v $GOPATH/src/github.com/m3db/m3/src/dbnode/config/m3dbnode-local-etcd.yml:/etc/m3dbnode/m3dbnode.yml quay.io/m3/m3dbnode:latest
```
robskillington: Hm, you shouldn't have to add the config as a file mount - it will be auto added I believe, as per line 26 of the Dockerfile:
https://github.com/m3db/m3/blob/master/docker/m3dbnode/Dockerfile#L26

docs/how_to/single_node.md
```diff
@@ -16,9 +16,11 @@ directory on your host for durability:
...
 ```
 
+**Note:** This setup runs `M3DB` and `m3coordinator` as one application and should only be used for testing/development purposes. If you want to run a clustered `M3DB` with a separate `m3coordinator`, please [see here](cluster_hard_way.md).
```
robskillington: nit: replace "M3DB" with "m3dbnode" so it matches the binary names, since you're using the binary name of coordinator and not "M3Coordinator".

i.e. they should both be "m3dbnode" and "m3coordinator" or they should both be "M3DB" and "M3Coordinator".

docs/integrations/prometheus.md
```diff
@@ -92,3 +92,11 @@ remote_read:
 remote_write:
   - url: "http://localhost:7201/api/v1/prom/remote/write"
 ```
+
+Also, if you want to see metrics generated by the coordinator/query process such as number of queries, etc., you should add the following job under `scrape_configs`:
```
robskillington: nit: replace "coordinator/query" with "m3coordinator or m3query" for consistency?

src/dbnode/config/m3dbnode-local-etcd.yml
```diff
@@ -3,6 +3,44 @@ coordinator:
     type: "config"
     value: "0.0.0.0:7201"
 
+  clusters:
```
robskillington: This is not necessary for the embedded m3dbnode case, it will auto create a client for the local node.  This is adding a pretty large overhead of config that's not necessary.

benraskin92:
Closed in favor of https://github.com/m3db/m3/pull/1335

PR #1339, opened 2019-01-31
[dbnode] allow replacing of seed node + docs

schallert:
Replacing a seed node requires some extra steps and was previously not
feasible using config alone. This adds a config flag to the seed node
list to allow a new seed node to join the etcd cluster, and adds docs
for replacing a seed node.

### DIFF
--- a/docs/operational_guide/placement_configuration.md
+++ b/docs/operational_guide/placement_configuration.md
@@ -56,6 +56,9 @@ This value should be in the form of <M3DB_NODE_LISTEN_PORT> and identifies the p
 
 ### Placement Operations
 
+**NOTE**: If you find yourself performing operations on seed nodes, please refer to the seed node-specific sections
+below before making changes.
+
 The instructions below all contain sample curl commands, but you can always review the API documentation by navigating to
 
 `http://<M3_COORDINATOR_HOST_NAME>:<CONFIGURED_PORT(default 7201)>/api/v1/openapi` or our [online API documentation](https://m3db.io/openapi/).
--- a/docs/operational_guide/placement_configuration.md
+++ b/docs/operational_guide/placement_configuration.md
@@ -136,6 +136,8 @@ After sending the delete command you will need to wait for the M3DB cluster to r
 
 #### Replacing a Node
 
+**NOTE**: If using embedded etcd and replacing a seed node, please read the section below.
+
 Send a POST request to the `/api/v1/services/m3db/placement/replace` endpoint containing hosts to replace and candidates to replace it with.
 
 ```bash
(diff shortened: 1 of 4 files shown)


docs/operational_guide/placement_configuration.md
```diff
@@ -154,3 +156,70 @@ curl -X POST <M3_COORDINATOR_HOST_NAME>:<M3_COORDINATOR_PORT(default 7201)>/api/
...
+  - hostID: host2
+    endpoint: http://1.1.1.2:2380
+  - hostID: host4
+    clusterState: existing
+    endpoint: http://1.1.1.4:2380
+```
+
+4. Start M3DB on `host4`.
+
+5. On all other seed nodes, update their `initialCluster` list to be exactly equal to the list on `host4` from step 3.
```
richardartoul: maybe obvious but should this also say: "Note, this means that you'll have to restart them one at a time" or something like that?

docs/operational_guide/placement_configuration.md
```diff
@@ -154,3 +156,70 @@ curl -X POST <M3_COORDINATOR_HOST_NAME>:<M3_COORDINATOR_PORT(default 7201)>/api/
...
+
+6. Follow the steps from `Replacing a Seed Node` to replace `host3` with `host4` in the M3DB placement.
+
+Once the M3DB replace finishes, you may want to rolling restart the other seed nodes to ensure they have an up-to-date
```
richardartoul: ah ok you mention it here, might be better up above

docs/operational_guide/placement_configuration.md
```diff
@@ -154,3 +156,70 @@ curl -X POST <M3_COORDINATOR_HOST_NAME>:<M3_COORDINATOR_PORT(default 7201)>/api/
     ]
 }'
 ```
+
+#### Replacing a Seed Node
```
richardartoul: Maybe add a disclaimer at the top of the file in bold saying something like: "If you're replacing an M3DB seed node please make sure to read the seed node specific instructions below."

docs/operational_guide/placement_configuration.md
```diff
@@ -136,6 +136,8 @@ After sending the delete command you will need to wait for the M3DB cluster to r
 
 #### Replacing a Node
```
richardartoul: Can we add similar sections for add and remove? Basically in the add section say its not recommended to make new nodes seed nodes and if you find yourself in a situation where you need more than 3 M3DB seed nodes you should just run an external cluster and then something similar for remove?

docs/operational_guide/placement_configuration.md
schallert: I mention this in `Replacing a Node`, we want it at the top as well?

richardartoul: Yeah sorry I mean more like: "If you're performing any placement operations on an M3DB seed node..." Just a general disclaimer incase they miss this specific heading

PR #1342, opened 2019-01-31
Get fetch timeout from db config

benraskin92:


src/query/api/v1/httpd/handler.go
```diff
@@ -131,8 +132,12 @@ func (h *Handler) RegisterRoutes() error {
 	).Methods(openapi.HTTPMethod)
 	h.router.PathPrefix(openapi.StaticURLPrefix).Handler(logged(openapi.StaticHandler()))
 
+	timeoutOpts := &prometheus.TimeoutOpts{
+		FetchTimeout: h.embeddedDbCfg.Client.FetchTimeout,
```
benraskin92: @r and @richardartoul are we guaranteed that `embeddedDbCfg.Client != nil` here?

src/query/api/v1/httpd/handler_test.go
```diff
@@ -55,10 +57,14 @@ func makeTagOptions() models.TagOptions {
...
-func setupHandler(store storage.Storage) (*Handler, error) {
+func setupHandler(t *testing.T, store storage.Storage) (*Handler, error) {
 	downsamplerAndWriter := ingest.NewDownsamplerAndWriter(store, nil, testWorkerPool)
-	return NewHandler(downsamplerAndWriter, makeTagOptions(), executor.NewEngine(store, tally.NewTestScope("test", nil)), nil, nil,
+	h, err := NewHandler(downsamplerAndWriter, makeTagOptions(), executor.NewEngine(store, tally.NewTestScope("test", nil)), nil, nil,
 		config.Configuration{}, nil, tally.NewTestScope("", nil))
+	require.NoError(t, err)
+	h.embeddedDbCfg = &dbconfig.DBConfiguration{Client: client.Configuration{FetchTimeout: 15 * time.Second}}
+
+	return h, nil
```
arnikola: No need to return nil here, instead change the return to just be `*Handler`

src/query/api/v1/httpd/handler.go
arnikola: Rather than having this, have the handler take in a `TimeoutOpts` in `NewHandler` and set it on the Handler stuct

src/query/api/v1/handler/prometheus/remote/read_test.go
```diff
@@ -46,6 +47,10 @@ import (
 
 var (
 	promReadTestMetrics = newPromReadMetrics(tally.NewTestScope("", nil))
+
+	timeoutOpts = &prometheus.TimeoutOpts{
```
arnikola: Add a test to check that the timeout is being honoured?

src/query/api/v1/httpd/handler.go
richardartoul: I don't think its guaranteed to be set

benraskin92: Okay maybe I'll just set it here then if it's nil

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -47,15 +47,20 @@ const (
 	filterNameTagsParam = "tag"
 	errFormatStr        = "error parsing param: %s, error: %v"
 
-	// TODO: get timeouts from configs
-	maxTimeout     = time.Minute
-	defaultTimeout = time.Second * 15
+	maxTimeout = 5 * time.Minute
+	// DefaultTimeout is the default timeout for fetch requests
+	DefaultTimeout = 15 * time.Second
```
arnikola: It feels a bit weird to export this from here; I think it would read better if you had this hardcoded backup value in `src/query/api/v1/httpd/handler.go` and keep it unexported.

Then in `common_test` hardcode the 15 seconds as the test value (might be better to up to 30 in light of user comments on gitter?)

src/query/api/v1/handler/prometheus/remote/read_test.go
```diff
@@ -79,11 +87,29 @@ func TestPromReadParsing(t *testing.T) {
...
+		promReadMetrics: promReadTestMetrics,
+		timeoutOpts: &prometheus.TimeoutOpts{
+			FetchTimeout: 2 * time.Minute,
+		},
+	}
+
+	req, _ := http.NewRequest("POST", PromReadURL, test.GeneratePromReadBody(t))
+	dur, err := prometheus.ParseRequestTimeout(req, promRead.timeoutOpts.FetchTimeout)
+	require.NoError(t, err)
+	assert.Equal(t, 2*time.Minute, dur)
```
arnikola: Nice!

Can you add a test with a default, and an invalid timeout value?

src/query/api/v1/handler/prometheus/common.go
benraskin92: Yeah I went back and forth on this. I'll put in `handler.go`

src/query/api/v1/handler/prometheus/remote/read_test.go
benraskin92: We have these tests in the native package. Not sure we need to duplicate.

src/query/api/v1/handler/prometheus/common_test.go
```diff
@@ -64,17 +64,17 @@ func TestTimeoutParse(t *testing.T) {
 	req, _ := http.NewRequest("POST", "dummy", nil)
 	req.Header.Add("timeout", "1ms")
 
-	timeout, err := ParseRequestTimeout(req)
+	timeout, err := ParseRequestTimeout(req, time.Second)
 	assert.NoError(t, err)
 	assert.Equal(t, timeout, time.Millisecond)
```
arnikola: Shouldn't this fail?

src/query/api/v1/httpd/handler.go
```diff
@@ -105,6 +110,17 @@ func NewHandler(
 		},
 	}
 
+	var timeoutOpts = &prometheus.TimeoutOpts{}
+	if embeddedDbCfg == nil {
+		timeoutOpts.FetchTimeout = defaultTimeout
+	} else {
+		if embeddedDbCfg.Client.FetchTimeout < 0 {
```
arnikola: Quick test to check this?

src/query/api/v1/handler/prometheus/common_test.go
arnikola: Oops never mind github hid the top bit

PR #1345, opened 2019-01-31
[ci] docker builds via buildkite

schallert:
Up until now we've been using Quay automated builds, which doesn't give
us all the granularity we want in tags. That is, every build Quay
creates it tags `latest` and `master` depending on the ref it builds
from.

What we really want is `master` to always point to `origin/master` and
`latest` to point to the latest tagged release, along with having a tag
for that release.

This PR modifies our buildkite pipeline to check if `HEAD` is
`origin/master` OR tagged, and if so dynamically add another step to the
pipeline to actually do the docker build. This minimizes blocking in the
queue for other jobs.

### DIFF
--- a/.buildkite/docker-pipeline.yml
+++ b/.buildkite/docker-pipeline.yml
@@ -1,0 +1,16 @@
+steps:
+  - wait
+  - name: ":docker: build"
+    command: "scripts/docker_build.sh"
+    agents:
+      queue: builder
+    timeout_in_minutes: 40
+    retry:
+      automatic:
+        limit: 1
+      manual: true
+    plugins:
+      - docker-login#v2.0.1:
+          server: quay.io
+          username: m3db+buildkite
+          password-env: QUAY_M3DB_TOKEN
--- a/.buildkite/docker-pipeline.yml
+++ b/.buildkite/docker-pipeline.yml
@@ -2,8 +2,6 @@ steps:
   - wait
   - name: ":docker: build"
     command: "docker/build.sh"
-    env:
-      DRYRUN: "true"
     agents:
       queue: builder
     timeout_in_minutes: 60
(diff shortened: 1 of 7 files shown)


.buildkite/docker-pipeline.yml
```diff
@@ -0,0 +1,16 @@
+steps:
+  - wait
+  - name: ":docker: build"
+    command: "scripts/docker_build.sh"
+    agents:
+      queue: builder
+    timeout_in_minutes: 40
```
robskillington: 60mins perhaps just to round up? Heh.

scripts/docker_build.sh
```diff
@@ -0,0 +1,60 @@
...
+# This script creates builds according to our custom build policy, which is:
+# - "master" will always point to the most recent build on master
+# - "latest" will refer to the latest tagged release
+# - Each release "foo" will have a tag "foo"
+#
+# This script is a noop if HEAD is not origin/master OR tagged.
+
+set -exuo pipefail
+
+IMAGES="m3dbnode m3coordinator m3query m3nsch"
```
robskillington: Interesting, should we think about centralizing this in a JSON document or something for the images we build for quay?

scripts/docker_build.sh
```diff
@@ -0,0 +1,60 @@
...
+done
+
+# Clean up
+CLEANUP_IMAGES=$(docker images | grep "$TAG_BASE" | awk '{print $3}' | sort | uniq)
+for IMG in $CLEANUP_IMAGES; do
+  docker rmi -f "$IMG"
+done
+
+docker system prune -f
+find /tmp -name '*m3-docker' -print0 | xargs -0 sudo rm -fv
```
robskillington: Should we do this in a `trap` perhaps?

scripts/docker_build.sh
schallert: Yeah good idea. I like the minimal dependencies we have so far, but maybe I can this in a JSON config and generate scripts from that or something.

schallert:
Planning to get back by EOW.

PR #1351, opened 2019-02-03
[query] M3DB dashboard now uses templated data source

arnikola:
This allows users to flip between Prometheus query processing and
M3Query processing. Also adds a new dashboard that overlays M3Query on
the same graph as Prometheus Query, allowing for easy comparisons.

### DIFF
--- a/docker/grafana/Dockerfile
+++ b/docker/grafana/Dockerfile
@@ -3,13 +3,13 @@ FROM grafana/grafana:latest
 COPY ./docker/grafana/datasource.yaml /etc/grafana/provisioning/datasources/datasource.yaml
 COPY ./docker/grafana/dashboards.yaml /etc/grafana/provisioning/dashboards/all.yaml
 
+COPY ./integrations/grafana/m3db_dashboard.json /tmp/m3db_dashboard.json
+RUN mkdir -p /tmp/grafana_dashboards
+RUN cp /tmp/m3db_dashboard.json  /tmp/grafana_dashboards/m3db_dashboard.json
+
 # Need to replace datasource template variable with name of actual data source so auto-import
 # JustWorksTM. Use a temporary directory to host the dashboards since the default
 # directory is owned by root.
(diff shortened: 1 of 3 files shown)


arnikola:
> Do we need to have two dashboards? Perhaps we can just modify the existing one without needing to add the overlaid_dashboard? Although I might be completely misunderstanding it, if it's required in some way to have two then that makes sense, but if not maybe we can keep just the one?

So the idea with the overlaid_dashboard is ~~to pad my lines of code~~ that it's a bit easier to see any discrepancies between Prometheus and m3query results if they're both present on a single graph, rather than having to flip the `$datasource` variable between them. 

As an anecdotal example, I was trying to fix an issue where the results were off by a factor of 2x, which was almost impossible to see when flipping the datasource as the shape was roughly the same, and only the scales on the axis had changed

docker/grafana/Dockerfile
```diff
@@ -3,9 +3,13 @@ FROM grafana/grafana:latest
 COPY ./docker/grafana/datasource.yaml /etc/grafana/provisioning/datasources/datasource.yaml
 COPY ./docker/grafana/dashboards.yaml /etc/grafana/provisioning/dashboards/all.yaml
 
+COPY ./integrations/grafana/m3db_dashboard.json /tmp/m3db_dashboard.json
+RUN mkdir -p /tmp/grafana_dashboards
```
richardartoul: can you just run this command first and avoid the first copy above? Not seeing the point of the two copies. I think what it did before was copy it to tmp, make the new directory, then use awk to create a new file in the new folder which made a little more sense, but right now this first copy seems like basically a no-op

richardartoul: also it looks like you deleted the awk command for `m3db_dashboard`, could you make sure that dashboard still works before merging this?

richardartoul: oh nevermind I see how it works with the template var now.

arnikola: Yeah, ran the m3db_dashboard and it; I think it defaults to m3query

Good call on the copy, will fix it up

PR #1352, opened 2019-02-04
Fix development env m3_stack for Prometheus scraping wrong port

robskillington:


robskillington:
PS we should add an integration tests that checks m3_stack is working as expected

PR #1354, opened 2019-02-06
Add etcd operational guide

richardartoul:


docs/operational_guide/etcd.md
```diff
@@ -0,0 +1,81 @@
+# etcd
```
benraskin92: Do we want to have recs on how many external etcd hosts one should have? is there some sort of formula based on the size of the m3db cluster?

docs/operational_guide/etcd.md
```diff
@@ -0,0 +1,81 @@
...
+
+The M3 stack leverages M3 as a distributed key-value storage to:
```
benraskin92: leverage `etcd` you mean?

docs/operational_guide/etcd.md
schallert: There's some etcd docs on the subject we can link to: https://github.com/etcd-io/etcd/blob/v3.3.11/Documentation/faq.md#what-is-failure-tolerance

docs/operational_guide/etcd.md
```diff
@@ -0,0 +1,81 @@
...
+        cacheDir: /var/lib/m3kv
+        etcdClusters:
+            - zone: embedded
+              endpoints:
+                  - http://external_etcd1:2379
+                  - http://external_etcd2:2379
+                  - http://external_etcd3:2379
+```
+
+**Note**: `M3DB` placements and namespaces are stored in `etcd` so if you want to switch to an external `etcd` cluster you'll need to recreate all your placements and namespaces.
```
schallert: `etcdctl` has a feature to copy data from 1 cluster to another: https://github.com/etcd-io/etcd/blob/v3.3.11/etcdctl/doc/mirror_maker.md

Maybe we can link that here?

richardartoul:
@schallert good call, done

PR #1355, opened 2019-02-06
Add query (postings list) cache to M3DB index

richardartoul:


src/dbnode/storage/index/query_cache.go
```diff
@@ -0,0 +1,274 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package index
+
+import (
+	"sync"
+	"time"
+
+	"github.com/hashicorp/golang-lru"
```
prateek: JFYI - that package has an (ARC implementation)[https://github.com/hashicorp/golang-lru/blob/master/arc.go#L9] which has a (complicated)[https://news.ycombinator.com/item?id=17317031] background (is still under patent from Oracle). For paranoia's sake, can you pick another package/roll your own

richardartoul: spoke offline, but ended up rolling my own.

src/cmd/services/m3dbnode/config/cache.go
```diff
@@ -22,10 +22,19 @@ package config
 
 import "github.com/m3db/m3/src/dbnode/storage/series"
 
+var (
+	defaultPostingsListCacheSize   = 256000
```
robskillington: Maybe keep this to the power of 2, i.e. 2^18 or `2<<17`

src/cmd/services/m3dbnode/config/cache.go
```diff
@@ -38,6 +47,20 @@ func (c CacheConfigurations) SeriesConfiguration() SeriesCacheConfiguration {
...
 }
 
+// PostingsListConfiguration returns the postings list cache configuration
+// or default if none is specified.
+func (c CacheConfigurations) PostingsListConfiguration() PostingsListCacheConfiguration {
+	if c.PostingsList == nil {
+		return PostingsListCacheConfiguration{
+			Size:        &defaultPostingsListCacheSize,
+			CacheRegexp: &defaultPostingsListCacheRegexp,
+			CacheTerms:  &defaultPostingsListCacheTerms,
```
robskillington: Why not just leave this all `nil` and let the `ThingOrDefault()` handle the case to provide the default value?

src/dbnode/server/server.go
```diff
@@ -241,14 +241,35 @@ func Run(runOpts RunOptions) {
 		runtimeOpts = runtimeOpts.SetMaxWiredBlocks(lruCfg.MaxBlocks)
 	}
 
+	// Setup postings list cache.
+	var (
+		plCacheConfig  = cfg.Cache.PostingsListConfiguration()
+		plCacheSize    = plCacheConfig.SizeOrDefault()
+		plCacheOptions = index.PostingsListCacheOptions{
+			InstrumentOptions: opts.InstrumentOptions().
+				SetMetricsScope(scope.SubScope("query-cache")),
```
robskillington: Maybe `query-postingslist-cache`?

src/dbnode/storage/index/block.go
```diff
@@ -833,9 +833,24 @@ func (b *block) AddResults(
...
+	var (
+		plCache         = b.opts.PostingsListCache()
+		readThroughOpts = b.opts.ReadThroughSegmentOptions()
+		segments        = results.Segments()
+	)
+	readThroughSegments := make([]segment.Segment, 0, len(segments))
+	for _, seg := range segments {
+		readThroughSeg := seg
+		if _, ok := seg.(segment.MutableSegment); !ok {
+			// Only wrap the immutable segments with a read through cache.
```
robskillington: Honestly, if it's AddResults path, it's fine to wrap the mutable ones too here (we never actually write to them, they're essentially frozen).

I'll also fix this up so this never happens and we always deliver FSTs here soon.

src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/uber-go/tally"
+)
+
+// PatternType is an enum for the various pattern types. It allows us
+// separate them logically within the cache.
+type PatternType int
+
+const (
+	// PatternTypeRegexp indicates that the pattern is of type regexp.
+	PatternTypeRegexp PatternType = 0
```
robskillington: nit: Use iota?
```go
PatternTypeRegexp PatternType = iota
PatternTypeTerm
```

src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -0,0 +1,251 @@
...
+	// PatternTypeTerm indicates that the pattern is of type term.
+	PatternTypeTerm PatternType = 1
+
+	reportLoopInterval = time.Second
```
robskillington: This is pretty tight, want to make it 5s or 10s?

src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+	}()
+
+	return func() { close(doneCh) }
+}
+
+// Report will emit metrics about the status of the cache.
+func (q *PostingsListCache) Report() {
+	var size float64
+	var capacity float64
```
robskillington: nit:
```go
var size, capacity float64
```

or a var block

src/dbnode/storage/index/postings_list_cache_lru.go
```diff
@@ -0,0 +1,233 @@
...
+	patternType PatternType,
+	pl postings.List,
+) (evicted bool) {
+	// Check for existing item.
+	uuidArray := segmentUUID.Array()
+	if uuidEntries, ok := c.items[uuidArray]; ok {
+		key := newPatternAndPatternType(pattern, patternType)
+		if ent, ok := uuidEntries[key]; ok {
+			c.evictList.MoveToFront(ent)
+			ent.Value.(*entry).postingsList = pl
```
robskillington: Interesting, how often would this happen?  Just basically due to two queries coming in and both not hitting the cache I assume, then both results get added?

src/dbnode/storage/index/read_through_segment.go
```diff
@@ -0,0 +1,183 @@
...
+// MatchTerm returns a cached posting list or queries the underlying
+// segment if their is a cache miss.
+func (s *readThroughSegmentReader) MatchTerm(
+	field []byte, term []byte,
+) (postings.List, error) {
+	if s.postingsListCache == nil || !s.opts.CacheTerms {
+		return s.Reader.MatchTerm(field, term)
+	}
+
+	termString := string(term)
```
robskillington: At some point we could make these all bytes and use our generated map, but for now these will be so small comparitive to everything else it's probably fine to just stringify for now.

src/dbnode/storage/index/results.go
```diff
@@ -156,10 +157,25 @@ func (r *results) Reset(nsID ident.ID) {
...
+
 	r.Reset(nil)
 
 	if r.pool == nil {
 		return
 	}
 	r.pool.Put(r)
 }
+
+func (r *results) NoFinalize() {
```
robskillington: Hm, I can't actually see where this is used - is this still used?

richardartoul: No, I just had already written and tested it and figure'd we'd want it for query caching later so might as well leave it in. Want me to take it out?

src/cmd/services/m3dbnode/config/cache.go
richardartoul: sure

src/dbnode/storage/index/block.go
richardartoul: Yeah thats fair, I'm a little hesitant to allow wrapping of mutable segments though just for correctness purposes (I know you're gonna do the F.S.T thing but would suck for this to break months/years down the line). I'll do the mutable segment check, but i'll still allow wrapping if its "sealed"

richardartoul: Actually spoke offline and settled on not wrapping the mutable segments due to complications it introduces. We're going to delete the concept of a mutable segment soon anyways so this will become a moot point.

src/dbnode/storage/index/read_through_segment.go
richardartoul: yeah I think considering how expensive the operations we're caching here are this wouldn't really matter. Also I think our map requires wrapping things in an ident.ID?

src/dbnode/storage/index/postings_list_cache_lru.go
richardartoul: Yeah basically. I think technically it would still work if you didn't do this, but you'd be wasting space and have items in the linked list that weren't reachable from the map which would be a problem because thats how we do purges.

This is also how the LRU implementation I modeled this one off of did it.

richardartoul: Added a comment and updated the tests to make sure they break if this logic is changed.

PR #1356, opened 2019-02-07
Fix race/panic in index query code

richardartoul:


src/dbnode/storage/index_query_concurrent_test.go
```diff
@@ -248,6 +274,8 @@ func testNamespaceIndexHighConcurrentQueries(
 				rangeEnd := blockStarts[k].Add(test.indexBlockSize)
 
 				ctx := context.NewContext()
+				defer ctx.Close()
```
robskillington: Really shouldn't defer from within a for loop, but.. ok haha.

PR #1364, opened 2019-02-09
[ci] docker builds via buildkite

schallert:
Up until now we've been using Quay automated builds, which doesn't give
us all the granularity we want in tags. That is, every build Quay
creates it tags `latest` and `master` depending on the ref it builds
from.

What we really want is `master` to always point to `origin/master` and
`latest` to point to the latest tagged release, along with having a tag
for that release.

This PR modifies our buildkite pipeline to check if `HEAD` is
`origin/master` OR tagged, and if so dynamically add another step to the
pipeline to actually do the docker build. This minimizes blocking in the
queue for other jobs.


schallert:
Had to close and reopen this PR to get CI working (FYI @robskillington )

.buildkite/scripts/check_do_docker.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -exuo pipefail
+
+# This script checks if we should do a docker build, and if so adds a step in
```
prateek: nit: mind adding some kind of check to ensure this is only attempted in on a CI host - something like `[[ -z BUILDKITE ]] || (echo "..."; exit 1)`

docker/images.json
```diff
@@ -0,0 +1,23 @@
+{
```
prateek: mind adding a readme in this directory explaining what/where this file is used

prateek: nvm just saw the other files in the diff

.buildkite/docker-pipeline.yml
```diff
@@ -0,0 +1,20 @@
+steps:
```
prateek: nit: mind renaming file to `image-release-pipeline.yml` or something else to indicate what it's for

docker/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+3. For all images, and for all releases, there will be an image tagged `${IMAGE}:${RELEASE}`.
+
+## Builds
+
+This directory contains the Dockerfiles, configs, and build scripts for building images according to the above policy. `images.json` contains the base repo that will be used to tag images, as well as the config for each image. For example, with the following config:
+
+```json
+{
+  "image_base": "quay.io/m3db",
```
prateek: nit: mind using `quay.io/m3` as the example in this file

docker/build.sh
```diff
@@ -0,0 +1,78 @@
...
+  if [[ -z "$DRYRUN" ]]; then
+    docker push "$1"
+  else
+    echo "would push $1"
+  fi
+}
+
+CONFIG="docker/images.json"
+
+IMAGES="$(<$CONFIG jq -er '.images | to_entries | map(.key)[]')"
```
prateek: could you ensure `[[ -f $CONFIG ]]` before this

docker/build.sh
```diff
@@ -0,0 +1,78 @@
...
+# If the current commit is exactly origin/master, push a tag for "master".
+if [[ "$CURRENT_SHA" == "$MASTER_SHA" ]]; then
+  TAGS_TO_PUSH="${TAGS_TO_PUSH} master"
+fi
+
+if [[ -z "$TAGS_TO_PUSH" ]]; then
+  exit 0
+fi
+
+for IMAGE in $IMAGES; do
```
prateek: nit: could you add a little more logging to the loops below

docker/build.sh
```diff
@@ -0,0 +1,78 @@
...
+set -exo pipefail
+
+function cleanup() {
+  docker system prune -f
+  find /tmp -name '*m3-docker' -print0 | xargs -0 rm -fv
+}
+
+trap cleanup EXIT
+
+function push() {
```
prateek: nit: mind calling this pushimage or something longer than just push. The usage of this in the loop made me think it was related to `pushd` until I remembered this decl

PR #1367, opened 2019-02-11
[query] Add quantile_over_time

benraskin92:


src/query/functions/temporal/aggregation.go
```diff
@@ -85,28 +92,42 @@ func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
 			aggFunc: aggregationFunc,
 		}
 
+		if optype == QuantileType {
+			if len(args) != 2 {
```
arnikola: This is pretty messy; do a similar approach to what's there for holt winters instead

src/query/parser/promql/types.go
```diff
@@ -204,7 +204,7 @@ func NewFunctionExpr(
 
 	case temporal.AvgType, temporal.CountType, temporal.MinType,
 		temporal.MaxType, temporal.SumType, temporal.StdDevType,
-		temporal.StdVarType:
+		temporal.StdVarType, temporal.QuantileType:
```
arnikola: Add to parse test

src/query/functions/temporal/aggregation.go
```diff
@@ -85,28 +92,42 @@ func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
...
 }
 
 type aggNode struct {
-	op         baseOp
-	controller *transform.Controller
-	aggFunc    func([]float64) float64
+	op             baseOp
+	controller     *transform.Controller
+	aggFunc        func([]float64, float64) float64
+	quantileScalar float64
```
martin-mao: Do we have to add this quantileScalar at the generic agg func layer? Doesn't seem scalable given how many different types of agg funcs we can have. Can we just do the args checking and parsing once we get to the functions that need them?

src/query/functions/temporal/aggregation.go
```diff
@@ -85,28 +92,42 @@ func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
...
 }
 
 func (a *aggNode) Process(datapoints ts.Datapoints, _ time.Time) float64 {
-	return a.aggFunc(datapoints.Values())
+	return a.aggFunc(datapoints.Values(), a.quantileScalar)
 }
 
-func avgOverTime(values []float64) float64 {
+func avgOverTime(values []float64, _ float64) float64 {
```
martin-mao: Not great that we modify all the func signatures for a var that isn't used in most of these functions. If it needs to be this way, can we wrap it in an options or something at least so the function signature doesn't have to ever change again? (But only if you have to pass this all the way down, see comment above).

martin-mao: Or consider putting this in some sort of context object that you propagate instead of a specific options for agg functions.

andrewmains12: +1. Would it make sense to pass in `scalar` as a constructor argument to the quantile `aggFunc`? That way we separate out the "configuration" of the `aggFunc` from it's actual execution. A note on how that could work: 

1. We can switch `aggFunc` to be an interface (functions can implement interfaces, so this is actually less work than it seems), to allow for `quantile` to hold configuration (`scalar`)
2. Pass `scalar` to `quantile` when we initialize the node instead (threading through `NewAggOp` and `temporal.baseOp.Node`)

src/query/functions/temporal/aggregation.go
```diff
@@ -192,3 +213,60 @@ func sumAndCount(values []float64) (float64, float64) {
 
 	return sum, count
 }
+
+func quantileOverTime(values []float64, scalar float64) float64 {
+	valuesSlice := make(valsSlice, 0, len(values))
```
andrewmains12: Add a comment about why copying the array is necessary?

src/query/functions/temporal/aggregation.go
```diff
@@ -192,3 +220,62 @@ func sumAndCount(values []float64) (float64, float64) {
 
 	return sum, count
 }
+
+func makeQuantileOverTimeFn(scalar float64) aggFunc {
+	return func(values []float64) float64 {
+		valuesSlice := make(valsSlice, 0, len(values))
+		for _, v := range values {
+			valuesSlice = append(values, v)
```
martin-mao: What is this loop trying to achieve? This seems to be overwriting valuesSlice on each iteration with the whole of values + the current value from values. You end up with just the values slice + the very last value in values in valuesSlice. Is that what you want?

src/query/functions/temporal/base.go
```diff
@@ -49,15 +49,27 @@ type baseOp struct {
...
+	var (
+		durArg   interface{}
+		duration time.Duration
+		ok       bool
+	)
+
+	if operatorType != QuantileType {
+		durArg = args[0]
+	} else {
+		durArg = args[1]
```
martin-mao: I know you check outside of this function, but is it worth checking len args in this func just in case? There's some check above for len(args) != 1, but does that cover everything?

src/query/functions/temporal/aggregation.go
```diff
@@ -78,6 +82,30 @@ func (a aggProcessor) Init(op baseOp, controller *transform.Controller, opts tra
 	}
 }
 
+// NewQuantileOp create a new base temporal transform for quantile_over_time func
```
arnikola: nit: `.` at end of comment

src/query/functions/temporal/aggregation.go
```diff
@@ -78,6 +82,30 @@ func (a aggProcessor) Init(op baseOp, controller *transform.Controller, opts tra
...
+func NewQuantileOp(args []interface{}, optype string) (transform.Params, error) {
+	if optype == QuantileType {
```
arnikola: Flip the conditional to reduce nesting

src/query/functions/temporal/aggregation.go
```diff
@@ -78,6 +82,30 @@ func (a aggProcessor) Init(op baseOp, controller *transform.Controller, opts tra
...
+		if len(args) != 2 {
+			return emptyOp, fmt.Errorf("invalid number of args for %s: %d", QuantileType, len(args))
+		}
+
+		scalar, ok := args[0].(float64)
+		if !ok {
+			return emptyOp, fmt.Errorf("unable to cast to scalar argument: %v for %s", args[1], QuantileType)
```
arnikola: This prints args[1] but you're checking args[0]; do we ever use args[1]? Does it need two args?

src/query/functions/temporal/aggregation.go
```diff
@@ -192,3 +220,62 @@ func sumAndCount(values []float64) (float64, float64) {
...
+//
+// This slice will be sorted.
+// If 'values' has zero elements, NaN is returned.
+// If q<0, -Inf is returned.
+// If q>1, +Inf is returned.
+func quantile(q float64, values valsSlice) float64 {
+	if len(values) == 0 {
+		return math.NaN()
+	}
+	if q < 0 {
```
arnikola: nit: newlines

src/query/functions/temporal/aggregation.go
```diff
@@ -192,3 +220,62 @@ func sumAndCount(values []float64) (float64, float64) {
...
+	s[i], s[j] = s[j], s[i]
+}
+
+// qauntile calculates the given quantile of a slice of values.
+//
+// This slice will be sorted.
+// If 'values' has zero elements, NaN is returned.
+// If q<0, -Inf is returned.
+// If q>1, +Inf is returned.
+func quantile(q float64, values valsSlice) float64 {
```
arnikola: Nit: keep same name for the `q` variable throughout rather than having it be `scalar` when you parse it

src/query/functions/temporal/aggregation.go
```diff
@@ -78,6 +82,30 @@ func (a aggProcessor) Init(op baseOp, controller *transform.Controller, opts tra
...
+		if len(args) != 2 {
+			return emptyOp, fmt.Errorf("invalid number of args for %s: %d", QuantileType, len(args))
+		}
+
+		scalar, ok := args[0].(float64)
+		if !ok {
+			return emptyOp, fmt.Errorf("unable to cast to scalar argument: %v for %s", args[1], QuantileType)
+		}
+
+		aggregationFunc := makeQuantileOverTimeFn(scalar)
```
arnikola: You can add an optimization here is `scalar < 0 || scalar > 1` by instead having a function that looks like:

```
func makeInvalidQuantileFn(inf float64) aggFunc {
 return func(_ []float64) float64 {
  return inf
 }
}
```
And passing the correctly signed infinity in

src/query/functions/temporal/aggregation.go
andrewmains12: +1--looks like it should be `valuesSlice`? You could probably also use `copy` for this (assuming that's the goal)

src/query/functions/temporal/aggregation.go
```diff
@@ -192,3 +220,62 @@ func sumAndCount(values []float64) (float64, float64) {
...
+	if len(values) == 0 {
+		return math.NaN()
+	}
+	if q < 0 {
+		return math.Inf(-1)
+	}
+	if q > 1 {
+		return math.Inf(+1)
+	}
+	sort.Sort(values)
```
andrewmains12: Is this the approach prom (and/or M3QL) take to quantile computation? Not that we can necessarily do better (without approximating a la https://dl.acm.org/citation.cfm?id=1054027), just curious.

benraskin92: This is from Prom which we need to match

src/query/functions/temporal/base.go
```diff
@@ -49,15 +49,27 @@ type baseOp struct {
 // skipping lint check for a single operator type since we will be adding more
 // nolint : unparam
 func newBaseOp(args []interface{}, operatorType string, processorFn MakeProcessor) (baseOp, error) {
-	if operatorType != HoltWintersType && operatorType != PredictLinearType {
+	if operatorType != HoltWintersType && operatorType != PredictLinearType && operatorType != QuantileType {
```
andrewmains12: Seconding +arnikola's earlier comment here; why not handle this in the same way that we do `HoltWinters` and `PredictLinear`, and handle this casing in `NewFunctionExpr`? Handling the argument extraction in 2 places is unnecessarily confusing imo.

src/query/functions/temporal/aggregation_test.go
```diff
@@ -223,6 +235,18 @@ var testCasesNaNs = []testCase{
 			{math.NaN(), math.NaN(), math.NaN(), math.NaN(), math.NaN()},
 		},
 	},
+	{
+		name:   "quantile_over_time",
+		opType: QuantileType,
+		afterBlockOne: [][]float64{
+			{math.NaN(), math.NaN(), math.NaN(), math.NaN(), math.NaN()},
```
andrewmains12: Do we have some tests that don't include `NaN`'s? Might be a useful simple case to cover.

src/query/functions/temporal/aggregation.go
andrewmains12: If this is creating a `QuantileOp`, why does it need `optype` as input? Anything other than `QuantileType` is an error, so it doesn't seem worth having the argument (fwiw, `HoltWinters` doesn't take one).

src/query/functions/temporal/aggregation.go
```diff
@@ -192,3 +220,62 @@ func sumAndCount(values []float64) (float64, float64) {
 
 	return sum, count
 }
+
+func makeQuantileOverTimeFn(scalar float64) aggFunc {
```
andrewmains12: Thanks for this refactor, much cleaner this way.

src/query/functions/temporal/base.go
benraskin92: Eh I think it's fine (we do the same thing in the Holt-Winters function) + newBaseOp() is a private function

src/query/functions/temporal/aggregation.go
arnikola: There can be a problem with NaNs actually; we don't want them to pollute the output since they don't exist in Prom's implementation. Instead of just doing a `sort` here, it would be better to put the through a sanitize function that will remove NaNs from the values, then sort.

On the bright side, this means we can use the inbuilt int slice sorter, and not have to worry about the `valsSlice` type

andrewmains12: 👍 perfect thanks!

src/query/functions/temporal/aggregation_test.go
benraskin92: Yeah we should add tests for all of the temporal functions with some NaNs - ill do this in a separate PR.

src/query/functions/temporal/aggregation.go
```diff
@@ -78,14 +82,52 @@ func (a aggProcessor) Init(op baseOp, controller *transform.Controller, opts tra
...
+	a := aggProcessor{
+		aggFunc: aggregationFunc,
+	}
+
+	return newBaseOp(duration, QuantileType, a)
+}
+
 // NewAggOp creates a new base temporal transform with a specified node.
 func NewAggOp(args []interface{}, optype string) (transform.Params, error) {
 	if aggregationFunc, ok := aggFuncs[optype]; ok {
```
andrewmains12: Nit: flip the conditional, since the `if` is a lot larger now.

src/query/functions/temporal/base.go
```diff
@@ -48,18 +48,7 @@ type baseOp struct {
...
-			return emptyOp, fmt.Errorf("invalid number of args for %s: %d", operatorType, len(args))
-		}
-	}
-
-	duration, ok := args[0].(time.Duration)
-	if !ok {
-		return emptyOp, fmt.Errorf("unable to cast to scalar argument: %v for %s", args[0], operatorType)
-	}
-
+func newBaseOp(duration time.Duration, operatorType string, processorFn MakeProcessor) (baseOp, error) {
```
arnikola: :+1:

src/query/functions/temporal/aggregation.go
```diff
@@ -78,14 +82,52 @@ func (a aggProcessor) Init(op baseOp, controller *transform.Controller, opts tra
...
+	}
+
+	q, ok := args[0].(float64)
+	if !ok {
+		return emptyOp, fmt.Errorf("unable to cast to quantile argument: %v for %s", args[0], QuantileType)
+	}
+
+	duration, ok := args[1].(time.Duration)
+	if !ok {
+		return emptyOp, fmt.Errorf("unable to cast to scalar argument: %v for %s", args[1], QuantileType)
```
arnikola: Can we push all of these parsing functions up into `promql/types.go`, similar to how [this does it](https://github.com/m3db/m3/blob/master/src/query/parser/promql/types.go#L97)?

A small gotcha is that any numeric expression is valid in Prom, i.e. `quantile_over_time( (1*2)/3, up[5m] )` is a valid query but would fail here and in many of the other temporal functions.

benraskin92: Really don't wanna change this anymore. Also this query works in ours:

```
quantile_over_time((1*2)/3, service_writeTaggedBatchRaw_latency{instance=~"$instance",quantile="0.99"}[1m])               
```

Issue #1365, opened 2019-02-09
Remaining index optimizations


robskillington:
Tasks:
- [x] Review and merge the postings list cache P.R. https://github.com/m3db/m3/pull/1355.
- [ ] Modify the index flushing code to rely on mmap'd memory so that we can flush one segment instead of 4 without using a ton of memory during flush.
- [ ] Modify the database/create and database/namespace/create APIs to default to creating index blocksizes that are 2x larger than whatever data block size we recommend.
- [ ] Undo the hacky local to bitmap pilosa pooling implemented in our fork of pilosa and spend a few days trying to implement proper container pooling, and then get that merged back upstream.
- [ ] Update the FS bootstrapper (and even the peers bootstrapper) to always return FST segments rather than memory segments, along the way fixing the invariant violated issue in the FS bootstrapper that leads to a index block not being written to disk when index block size doesn't align with retention exactly.
- [ ] Take a blocking profile to see if postings list cache is highly contended

richardartoul:
#1382 addresses bullet 3

PR #1370, opened 2019-02-11
Add postings list cache to M3DB index

richardartoul:


src/dbnode/storage/index/postings_list_cache_lru.go
```diff
@@ -0,0 +1,237 @@
...
+// is keyed by pattern and pattern type.
+//
+// As a result, when a segment is ready to be closed, they can call into the cache with their
+// UUID and we can efficiently remove all the entries corresponding to that segment from the
+// LRU. The specialization has the additional nice property that we don't need to allocate everytime
+// we add an item to the LRU due to the interface{} conversion.
+type postingsListLRU struct {
+	size      int
+	evictList *list.List
+	items     map[[16]byte]map[patternAndPatternType]*list.Element
```
prateek: nit: s/`[16]byte]`/`uuid.Array`/g

src/dbnode/storage/index/postings_list_cache_lru.go
```diff
@@ -0,0 +1,237 @@
...
+// that were resolved by running a given query against a particular segment. Normally
+// an key in the LRU would look like:
+//
+// type key struct {
+//    segmentUUID uuid.UUID
+//    pattern     string
+//    patternType PatternType
+// }
+//
+// However, some of the postings lists that we will store in the LRU have a fixed lifecycle
```
prateek: hm not sure why you need all this protection here. If a segment is closed, how would we attempt a `Get()` on it?

src/dbnode/storage/index/read_through_segment.go
```diff
@@ -0,0 +1,183 @@
...
+// MatchTerm returns a cached posting list or queries the underlying
+// segment if their is a cache miss.
+func (s *readThroughSegmentReader) MatchTerm(
+	field []byte, term []byte,
+) (postings.List, error) {
+	if s.postingsListCache == nil || !s.opts.CacheTerms {
+		return s.Reader.MatchTerm(field, term)
+	}
+
+	termString := string(term)
```
prateek: can you avoid these allocs?

src/dbnode/storage/index/options.go
```diff
@@ -98,11 +98,13 @@ type opts struct {
 	docArrayPool                    doc.DocumentArrayPool
 	foregroundCompactionPlannerOpts compaction.PlannerOptions
 	backgroundCompactionPlannerOpts compaction.PlannerOptions
+	postingsListCache               *PostingsListCache
```
prateek: could you add a `!= nil` in the `Validate()`

src/dbnode/storage/index/postings_list_cache_lru.go
richardartoul: Yeah I reasoned through the entire lifecycle with @robskillington and I think the conclusion we came to was that this wasn't necessary at all, but we didn't like the idea of leaving objects in the cache that if access would cause a segmentation fault. Since there was a cheap way to prevent that from happening we decided to be overly cautious

src/dbnode/storage/index/read_through_segment.go
richardartoul: we could, but not without doing something complicated or "unsafe" right now. I'm not worried about it, compared to the 50GiB in postings lists we're still allocating every few minutes this is nothing and I'd rather avoid more complexity here and focus on solving the other performance issues

src/dbnode/storage/index/options.go
richardartoul: sure

src/dbnode/storage/index/results.go
```diff
@@ -156,10 +157,25 @@ func (r *results) Reset(nsID ident.ID) {
...
+
 	r.Reset(nil)
 
 	if r.pool == nil {
 		return
 	}
 	r.pool.Put(r)
 }
+
+func (r *results) NoFinalize() {
```
robskillington: I don't see this actually being used, can we remove this?

scripts/development/m3_stack/m3dbnode.yml
```diff
@@ -56,6 +56,8 @@ db:
   cache:
     series:
       policy: lru
+    postingsList:
```
prateek: how would you feel about moving this to the `index:` section?

src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -0,0 +1,253 @@
...
+		pattern,
+		PatternTypeTerm)
+}
+
+func (q *PostingsListCache) get(
+	segmentUUID uuid.UUID,
+	pattern string,
+	patternType PatternType,
+) (postings.List, bool) {
+	// No RLock because a Get() operation mutates the LRU.
```
prateek: what's the contention on this thing like? could do a bunch of things to make it less expensive (e.g. approximate the lru, sample updates, etc)

Fine to do this as a follow up too.

src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -0,0 +1,253 @@
...
+func (q *PostingsListCache) PurgeSegment(segmentUUID uuid.UUID) {
+	q.Lock()
+	q.lru.PurgeSegment(segmentUUID)
+	q.Unlock()
+}
+
+// StartReportLoop starts a background process that will call Report()
+// on a regular basis and returns a function that will end the background
+// process.
+func (q *PostingsListCache) StartReportLoop() func() {
```
prateek: Instead of exposing the external report loop controls, how about exposing `Start()` and `Close()` on the cache? Seems more in line with other things we do

src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -0,0 +1,253 @@
...
+	regexp *postingsListCacheMethodMetrics
+	term   *postingsListCacheMethodMetrics
+
+	size     tally.Gauge
+	capacity tally.Gauge
+}
+
+func newPostingsListCacheMetrics(scope tally.Scope) *postingsListCacheMetrics {
+	return &postingsListCacheMetrics{
+		regexp: newPostingsListCacheMethodMetrics(scope.SubScope("regexp")),
```
prateek: probably better to have these as tags instead of subscopes. would allow for querying with the m3ql functions better

src/dbnode/storage/index/postings_list_cache_lru.go
prateek: lol OK. +1 for the paranoia =)

Maybe also describe that motivation in the file comment. Currently reads like it's necessary as opposed to additional safety

src/dbnode/storage/index/postings_list_cache_lru.go
```diff
@@ -0,0 +1,237 @@
...
+		}
+	}
+}
+
+// Len returns the number of items in the cache.
+func (c *postingsListLRU) Len() int {
+	return c.evictList.Len()
+}
+
+// Keys returns a slice of the keys in the cache, from oldest to newest. Mostly
```
prateek: if it's only used for testing, can you move it to the `_test.go` file

src/dbnode/storage/index/postings_list_cache_test.go
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/pborman/uuid"
+	"github.com/stretchr/testify/require"
+)
+
+const (
+	numTestPlEntries = 1000
+)
+
+var (
+	// Filled in by init().
```
prateek: nit: move the init method below the var decl.

src/dbnode/storage/index/read_through_segment.go
prateek: 👍 sg

prateek: maybe add TODO as a reminder for the future?

scripts/development/m3_stack/m3dbnode.yml
richardartoul: Don't have super strong feelings about it, but from an operator perspective seems nice to be able to tune all the caches in one place.

src/dbnode/storage/index/postings_list_cache.go
richardartoul: doesn't show up in flame graphs at all, and at the load we're currently handling I don't think its highly contended (100s of queries/s at peak, maybe looking back 100 blocks each is 10,000 locks per second which isn't much). Would rather keep implementation simple (already kind of sucks I had to specialize it) until it becomes an issue which I don't think will happen until we have 1-2 orders of magnitude improvement in general query performance

src/dbnode/storage/index/postings_list_cache.go
richardartoul: Yeah it just seems a little weird because there are no resources that need to be released and no background processes going on, so Open()/Close() didn't seem to make a lot of sense.

scripts/development/m3_stack/m3dbnode.yml
richardartoul: actually now I'm torn :/ maybe I'll just do index since thats where your gut said it should go. Should be ok once this P.R lands #1371 configs should be a lot simpler

richardartoul: although we don't currently have anything under the index field...

prateek: Nah we have the maxconcurrentquery knob there already.

It makes more sense to me under `index:` cause the rest of the code, e.g. IndexOptions are grouped as such too

src/dbnode/storage/index/postings_list_cache.go
prateek: wouldn't show up in the cpu flamegraphs necessarily - did you try the blocking?.

mainly curious if this is the cause of latency spikes you mentioned but we can investigate this separately. don't want to block the PR.

src/dbnode/storage/index/postings_list_cache.go
prateek: I mean the bg report loop is a resource that needs to be controlled. The current method feels leaky - passing lifecycle hooks outside the function.

Could alternatively do it with a better ctor, something like the Tally Scope idiom.

```
cache, closer, err := index.NewPostingsCache(...)
if err != nil {
...
}
defer closer()
```

src/dbnode/storage/index/postings_list_cache.go
prateek: still applicable?

src/dbnode/storage/index/postings_list_cache_lru.go
richardartoul: cool updated the comment

src/dbnode/storage/index/read_through_segment.go
richardartoul: Sure left a TODO

scripts/development/m3_stack/m3dbnode.yml
richardartoul: Ok sure will move it there

src/dbnode/storage/index/postings_list_cache.go
richardartoul: ah...no I did not check blocking. I will get a blocking profile next time I do benchmarking and see if this is an issue. My guess is that its not the root cause of the latency spikes but I'd love to be wrong as this would be an easy fix lol

src/dbnode/storage/index/postings_list_cache.go
richardartoul: yeah I like that one. I'll do that

src/dbnode/storage/index/postings_list_cache.go
richardartoul: sure thing

src/dbnode/storage/index/postings_list_cache_lru.go
richardartoul: done

src/dbnode/storage/index/postings_list_cache_test.go
richardartoul: sure

scripts/development/m3_stack/m3dbnode.yml
richardartoul: @prateek Although to be fair the YAML config is for humans, so I don't think how the rest of the code is organized should be the decision maker. Its more about where it makes sense for an operator to look at and modify it

richardartoul: I think I might leave as is. We're gonna keep adding caches I think its nice all in one place.

src/dbnode/storage/index/postings_list_cache.go
richardartoul: Added an additional bullet point to #1365 to track

scripts/development/m3_stack/m3dbnode.yml
prateek: Sure, I don't feel strongly either way

src/dbnode/server/server.go
```diff
@@ -1162,15 +1182,16 @@ func withEncodingAndPoolingOptions(
...
-			policy.IndexResultsPool,
-			scope.SubScope("index-results-pool")))
-
-	postingsListOpts := poolOptions(
-		policy.PostingsListPool,
-		scope.SubScope("postingslist-pool"))
-	postingsList := postings.NewPool(postingsListOpts, roaring.NewPostingsList)
+	var (
+		resultsPool = index.NewResultsPool(
+			poolOptions(
```
prateek: nit: putting these in the same lines reads better

src/dbnode/server/server.go
```diff
@@ -1162,15 +1182,16 @@ func withEncodingAndPoolingOptions(
...
+				policy.IndexResultsPool,
+				scope.SubScope("index-results-pool")))
+		postingsListOpts = poolOptions(
```
prateek: here too

src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -0,0 +1,267 @@
...
+func (q *PostingsListCache) PurgeSegment(segmentUUID uuid.UUID) {
+	q.Lock()
+	q.lru.PurgeSegment(segmentUUID)
+	q.Unlock()
+}
+
+// startReportLoop starts a background process that will call Report()
+// on a regular basis and returns a function that will end the background
+// process.
+func (q *PostingsListCache) startReportLoop() func() {
```
prateek: +1, i like it

src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -0,0 +1,267 @@
...
+
+	"github.com/pborman/uuid"
+	"github.com/uber-go/tally"
+)
+
+// PatternType is an enum for the various pattern types. It allows us
+// separate them logically within the cache.
+type PatternType int
+
+type Closer func()
```
prateek: nit: need a comment on the exported type decl

src/dbnode/storage/index/postings_list_cache.go
prateek: nit: change return type to `Closer`

src/dbnode/storage/index/postings_list_cache_lru.go
```diff
@@ -0,0 +1,225 @@
...
+// }
+//
+// However, some of the postings lists that we will store in the LRU have a fixed lifecycle
+// because they reference mmap'd byte slices which will eventually be unmap'd. To prevent
+// these postings lists that point to unmap'd regions from remaining in the LRU, we want to
+// support the ability to efficiently purge the LRU of any postings list that belong to a
+// given segment. This isn't technically required for correctness as once a segment has been
+// closed, its old postings list in the LRU will never be accessed again (since they are only
+// addressable by that segments UUID), but we purge them from the LRU before closing the segment
+// anyways as an additional safety precaution.
```
prateek: +1, nice

src/dbnode/storage/index/postings_list_cache_test.go
```diff
@@ -0,0 +1,345 @@
...
+}
+
+type testEntry struct {
+	segmentUUID  uuid.UUID
+	pattern      string
+	patternType  PatternType
+	postingsList postings.List
+}
+
+func TestSimpleLRUBehavior(t *testing.T) {
```
prateek: +1 for the tests, nice coverage

src/dbnode/storage/index/types.go
```diff
@@ -94,6 +94,11 @@ type Results interface {
 	// including returning it to a backing pool.
 	Finalize()
 
+	// NoFinalize marks the Results such that a subsequent call to Finalize() will
+	// be a no-op and will not return the object to the pool or release any of its
+	// resources.
+	NoFinalize()
```
prateek: this is only used in tests afaict, can you remove it?

richardartoul: synced with Rob offline, since its already written and tested gonna leave it in so we can use for the query cache

PR #1372, opened 2019-02-12
[query] Adds histogram_quantile function

arnikola:


arnikola:
Going to close and re-open to try to get buildkite healthy

PR #1373, opened 2019-02-12
[query] Adds histogram_quantile function

arnikola:


src/query/functions/linear/histogram_quantile.go
```diff
@@ -0,0 +1,376 @@
...
+func NewHistogramQuantileOp(
+	args []interface{},
+	opType string,
+) (parser.Params, error) {
+	if len(args) != 1 {
+		return emptyOp, fmt.Errorf(
+			"invalid number of args for histogram_quantile: %d", len(args))
+	}
+
+	if opType != HistogramQuantileType {
```
benraskin92: Check this first

arnikola: Was copying the pattern [from clamp.go](https://github.com/m3db/m3/blob/master/src/query/functions/linear/clamp.go#L49)

Wouldn't mind dealing with these correctly within the parser in a later pr, like the binary and aggregation functions are done at the moment.

src/query/functions/linear/histogram_quantile_test.go
```diff
@@ -0,0 +1,283 @@
...
+		{upperBound: 1, value: 1},
+		{upperBound: 2, value: 2},
+	})
+	assert.True(t, math.IsNaN(actual))
+
+	actual = bucketQuantile(0.5, []bucketValue{
+		{upperBound: 1, value: 1},
+		{upperBound: math.Inf(1), value: 22},
+	})
+	assert.Equal(t, float64(1), actual)
```
benraskin92: Add test with neg infinity and > 2 buckets?

src/query/functions/linear/histogram_quantile_test.go
```diff
@@ -0,0 +1,283 @@
...
+
+	return sink.Values
+}
+
+var (
+	inf  = math.Inf(+1)
+	ninf = math.Inf(-1)
+)
+
+func TestQuantileFunctionFilteringWithoutA(t *testing.T) {
```
benraskin92: `WithoutA`?

benraskin92: Same with `WithQ`?

src/query/functions/linear/histogram_quantile_test.go
```diff
@@ -0,0 +1,283 @@
...
+	op, err := NewHistogramQuantileOp(args, HistogramQuantileType)
+	assert.NoError(t, err)
+
+	assert.Equal(t, HistogramQuantileType, op.OpType())
+	assert.Equal(t, "type: histogram_quantile", op.String())
+}
+
+func testQuantileFunctionWithQ(t *testing.T, q float64) [][]float64 {
+	args := make([]interface{}, 0, 1)
+	args = append(args, q)
```
benraskin92: Just pass in the args instead of having to construct them here?

src/query/functions/linear/histogram_quantile.go
```diff
@@ -0,0 +1,376 @@
...
+		return err
+	}
+
+	meta := stepIter.Meta()
+	seriesMetas := utils.FlattenMetadata(meta, stepIter.SeriesMeta())
+	bucketedSeries := gatherSeriesToBuckets(seriesMetas)
+
+	q := n.op.q
+	if q < 0 || q > 1 {
+		return processInvalidQuantile(q, bucketedSeries, meta, stepIter, n.controller)
```
benraskin92: Seems kind of weird to split this out like this. Maybe just have one function called `processQuantile` and put the inf logic in there.

src/query/functions/linear/histogram_quantile.go
```diff
@@ -0,0 +1,376 @@
...
+)
+
+const (
+	// HistogramQuantileType calculates the quantile for histogram buckets.
+	//
+	// NB: each sample must contain a tag with a bucket name (given by tag
+	// options) that denotes the upper bound of that bucket; series without this
+	// tag are ignored.
+	HistogramQuantileType = "histogram_quantile"
+	initIndexBucketLength = 10
```
benraskin92: Any specific reason to use 10?

src/query/functions/linear/histogram_quantile.go
```diff
@@ -0,0 +1,376 @@
...
+	if err != nil {
+		return err
+	}
+
+	for index := 0; stepIter.Next(); index++ {
+		step := stepIter.Current()
+		values := step.Values()
+		bucketValues := make([]bucketValue, 0, initIndexBucketLength)
+
+		aggregatedValues := make([]float64, 0, len(bucketedSeries))
```
benraskin92: You should be able to just pre-allocate this with a length and index in below when adding to it. Looks like you always add to it.

src/query/models/tags_test.go
```diff
@@ -226,18 +226,56 @@ func TestUpdateName(t *testing.T) {
...
+	value := []byte("n")
+	tags = tags.SetBucket(value)
+	actual, found = tags.Get(name)
+	assert.True(t, found)
+	assert.Equal(t, value, actual)
+	actual, found = tags.Bucket()
+	assert.True(t, found)
+	assert.Equal(t, value, actual)
+
+	value2 := []byte("abc")
```
benraskin92: Add test for empty val? e.g. `[]byte("")`

src/cmd/services/m3query/config/config.go
```diff
@@ -299,6 +299,10 @@ type TagOptionsConfiguration struct {
 	// If not provided, defaults to `__name__`.
 	MetricName string `yaml:"metricName"`
 
+	// BucketName specifies the tag name that corresponds to the metric's bucket.
+	// If not provided, defaults to `le`.
```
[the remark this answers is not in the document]
benraskin92: Nvm

src/query/functions/linear/histogram_quantile.go
```diff
@@ -0,0 +1,376 @@
...
+		sign = -1
+	}
+
+	setValue := math.Inf(sign)
+	outValues := make([]float64, len(bucketedSeries))
+	ts.Memset(outValues, setValue)
+
+	for index := 0; stepIter.Next(); index++ {
+		cloned := make([]float64, len(outValues))
+		copy(cloned, outValues)
```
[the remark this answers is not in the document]
arnikola: Nice catch, copied from the regular path, forgot to fix this one

src/query/functions/linear/histogram_quantile.go
```diff
@@ -0,0 +1,376 @@
...
+func processValidQuantile(
+	q float64,
+	bucketedSeries bucketedSeries,
+	meta block.Metadata,
+	stepIter block.StepIter,
+	controller *transform.Controller,
+) error {
+	sanitizeBuckets(bucketedSeries)
+
+	builder, err := setupBuilder(bucketedSeries, meta, stepIter, controller)
```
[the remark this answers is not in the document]
arnikola: Need to set a bunch of meta, seriesMetas, step counts unfortunately

src/cmd/services/m3query/config/config.go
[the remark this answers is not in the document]
arnikola: Yep :+1:

src/query/functions/linear/histogram_quantile_test.go
arnikola: Good idea

src/query/functions/linear/histogram_quantile_test.go
arnikola: Oh I copied some setup stuff from an agg function that filtered by an `A` tag, will update this name to make more sense

src/query/functions/linear/histogram_quantile_test.go
arnikola: Unfortunately it's pretty annoying to build a `[]interface{..}`, and that's what we need to pass into the function; since we only ever need the `q` value here, it should be fine to build the interface slice here

src/query/models/tags_test.go
arnikola: Sure; not sure if it's necessary a special case though

src/query/functions/linear/histogram_quantile.go
arnikola: Will do; was fine using `append` here since didn't want to add a tracking `idx` variable

src/query/functions/linear/histogram_quantile.go
arnikola: If we do that, will need to do the comparison for each step

src/query/functions/linear/histogram_quantile.go
arnikola: Seems to be the general convention for initing arrays we don't know the length of

Here it has (by chance) a bit more context in that the general rule of thumb for histograms is no more than 10 buckets

src/query/functions/linear/histogram_quantile.go
```diff
@@ -0,0 +1,375 @@
...
+}
+
+func setupBuilder(
+	bucketedSeries bucketedSeries,
+	meta block.Metadata,
+	stepIter block.StepIter,
+	controller *transform.Controller,
+) (block.Builder, error) {
+	metas := make([]block.SeriesMeta, len(bucketedSeries))
+	idx := 0
```
benraskin92: Why have this var? Just include `idx` in the for loop below?

src/query/functions/linear/histogram_quantile_test.go
```diff
@@ -0,0 +1,291 @@
...
+)
+
+func TestQuantileFunctionForInvalidQValues(t *testing.T) {
+	actual := testQuantileFunctionWithQ(t, -1)
+	assert.Equal(t, [][]float64{{ninf, ninf, ninf, ninf, ninf}}, actual)
+	actual = testQuantileFunctionWithQ(t, 1.1)
+	assert.Equal(t, [][]float64{{inf, inf, inf, inf, inf}}, actual)
+
+	actual = testQuantileFunctionWithQ(t, 0.8)
+	test.EqualsWithNansWithDelta(t, [][]float64{{15.6, 20, 11.6, 13.4, 0.8}}, actual, 0.00001)
```
benraskin92: Add a case with some `NaNs`

src/query/functions/linear/histogram_quantile.go
arnikola: It's a map not an array

src/query/functions/linear/histogram_quantile.go
```diff
@@ -0,0 +1,375 @@
...
+			continue
+		}
+
+		excludeTags := [][]byte{tags.Opts.MetricName(), tags.Opts.BucketName()}
+		tagsWithoutKeys := tags.TagsWithoutKeys(excludeTags)
+		id := tagsWithoutKeys.ID()
+		if buckets, found := bucketsForID[string(id)]; !found {
+			// Add a single indexed bucket for this ID with the current index only.
+			newBuckets := make([]indexedBucket, 0, initIndexBucketLength)
+			newBucket := indexedBucket{
```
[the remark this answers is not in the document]
arnikola: Good call; this was originally going to have a list of indices that are valid for that bucket but the logic got really awkward. Changed it to add a bucket/index combo every time instead, but didn't update here to reflect that

src/query/functions/linear/histogram_quantile.go
benraskin92: 🤦‍♂️

PR #1374, opened 2019-02-13
 [M3agg] Remove msgpack encoding

cw9:


src/metrics/encoding/msgpack/wire_format.md
```diff
@@ -1,115 +0,0 @@
-## Wire format for unaggregated metrics
```
richardartoul: Do we have something like this for protobuf?

richardartoul:
@cw9 Are you gonna land this? :P

src/metrics/encoding/msgpack/wire_format.md
cw9: The .proto file should be pretty self explanatory, we had this for msgpack because the code itself was really hard to read lol

cw9:
> @cw9 Are you gonna land this? :P

There are some(<1%) collectors in prod that are still sending msgpack traffic to m3aggregator, probably due to puppet disabled, so I can't really land this thing :(

cw9:
Let me know when the Shunt migration is done, then I'll land this.

PR #1375, opened 2019-02-13
[index] Prevent duplicated writes to commit log

arnikola:
Propagates a `shouldWrite` flag from the buffer if a point with the same value at the same time exists, which determines if the point should be added to the commit log or not.


src/dbnode/storage/database.go
```diff
@@ -548,12 +548,12 @@ func (d *db) Write(
 		return err
 	}
 
-	series, err := n.Write(ctx, id, timestamp, value, unit, annotation)
+	series, shouldWrite, err := n.Write(ctx, id, timestamp, value, unit, annotation)
```
richardartoul: I think this makes more sense as `wasWritten` instead of `shouldWrite`. It makes more sense to me as the buffer returning a bool indicating whether it wrote or not as opposed to the buffer telling the commitlog what to do

src/dbnode/storage/database.go
```diff
@@ -585,12 +585,12 @@ func (d *db) WriteTagged(
 		return err
 	}
 
-	series, err := n.WriteTagged(ctx, id, tags, timestamp, value, unit, annotation)
+	series, shouldWrite, err := n.WriteTagged(ctx, id, tags, timestamp, value, unit, annotation)
```
richardartoul: same comment

src/dbnode/storage/database.go
```diff
@@ -694,17 +696,31 @@ func (d *db) writeBatch(
...
+			// error case so that the commitlog knows to skip this entry.
+			writes.SetOutcome(i, series, err)
+		}
+	}
+
+	if numSkipped > 0 {
+		// Sanitize the writes; if there are any skipping writes, they get dropped
+		// here. Ensure that there are remaining non-errored writes that need to go
+		// through remaining.
+		writes.Sanitize()
```
richardartoul: I don't like the idea of reshuffling this entire slice on the hot path (also, it might be fine, but I worry it might interfere with some of the error handling logic in the tchannel layer).

I think the commitlog itself already checks if the write had an error and if so skips it. Instead of sanitizing, could we just add a new method for each write so that the commitlog can do something like:

```golang
if write.ShouldWrite() {

}
```

and then

```golang
func (w *write) ShouldWrite() bool {
    return w.SkipWrite == False && w.Err == nil
}
```

src/dbnode/storage/series/buffer.go
```diff
@@ -614,7 +614,7 @@ func (b *dbBufferBucket) write(
 				// in a time window will still cause a flood of disk/CPU resource
 				// usage writing values to the commit log, even if the memory
 				// profile is lean as a side effect of this write being a no-op.
-				return nil
+				return false, nil
```
richardartoul: Wanna delete the TODO right above this line since you're fixing it? :D

src/dbnode/storage/series/series.go
```diff
@@ -282,11 +282,11 @@ func (s *dbSeries) Write(
 	value float64,
 	unit xtime.Unit,
 	annotation []byte,
-) error {
+) (bool, error) {
 	s.Lock()
-	err := s.buffer.Write(ctx, timestamp, value, unit, annotation)
+	shouldWrite, err := s.buffer.Write(ctx, timestamp, value, unit, annotation)
```
richardartoul: `shouldWrite` -> `wasWritten`

src/dbnode/storage/shard.go
```diff
@@ -816,10 +816,16 @@ func (s *dbShard) writeAndIndex(
 		commitLogSeriesID          ident.ID
 		commitLogSeriesTags        ident.Tags
 		commitLogSeriesUniqueIndex uint64
+		shouldWrite                = true
```
richardartoul: `shouldWrite` -> `wasWritten`

Also, seems more intuitive for default value to be `false`. Does that break something?

richardartoul: Oh I see why you default it to `true`. Want to add a comment about the default value I.E we err on the side of caution and want to write it to the commitlog always in the async case because we don't know if its gonna be written or not until later?

src/dbnode/storage/shard.go
```diff
@@ -816,10 +816,16 @@ func (s *dbShard) writeAndIndex(
...
 	)
 	if writable {
 		// Perform write
-		err = entry.Series.Write(ctx, timestamp, value, unit, annotation)
+		shouldWrite, err = entry.Series.Write(ctx, timestamp, value, unit, annotation)
+		if err != nil {
```
richardartoul: Any reason you moved the error handling code around? it looks like the old code only had to call `DecrementReaderWriterCount` once but this one has to do it multiple times

src/dbnode/storage/shard.go
```diff
@@ -804,7 +804,7 @@ func (s *dbShard) writeAndIndex(
 		// Retrieve the inserted entry
 		entry, err = s.writableSeries(id, tags)
```
richardartoul: I think technically just because the series appeared here doesn't actually mean the write succeeded. In practice, I don't think it matter, but maybe worth a comment. (You can imagine a situation where two writes for the same series come in and get batched together. The first one fails (due to some issue in the buffer) and the second one succeeds. Both write will "look" like they succeeded here, but you don't actually know if the buffer rejected it or not for being a dual-write

src/dbnode/storage/shard.go
```diff
@@ -1283,7 +1290,10 @@ func (s *dbShard) insertSeriesBatch(inserts []dbShardInsert) error {
 
 		if inserts[i].opts.hasPendingWrite {
 			write := inserts[i].opts.pendingWrite
-			err := entry.Series.Write(ctx, write.timestamp, write.value,
+			// NB: Since this is adding a new series, it will always write a new value
```
richardartoul: I would change this comment to say: "We ignore the `wasWritten` return argument because this is an async operation and there is nothing to do with the value right now. In the future, we could try and propagate the `wasWritten` argument back to the caller somehow using the waitgroup, but its not required right now."

src/dbnode/ts/write_batch.go
```diff
@@ -97,6 +97,21 @@ func (b *writeBatch) SetOutcome(idx int, series Series, err error) {
 	b.writes[idx].Err = err
 }
 
+func (b *writeBatch) SetSkipWrite(idx int) {
+	b.writes[idx].skipWrite = true
+}
+
+func (b *writeBatch) Sanitize() {
```
richardartoul: Probably want to remove this as discussed

src/dbnode/storage/database.go
arnikola: Sure, sounds good to me

src/dbnode/storage/series/buffer.go
arnikola: Sounds good!

src/dbnode/storage/database.go
arnikola: Sure, sounds good; I was thinking that the reslice was cheaper than having these series go through the entire commitlog write path only to skip at the write step itself.

src/dbnode/storage/shard.go
arnikola: Sure sounds good, added comment

src/dbnode/storage/shard.go
arnikola: So this should still call `DecrementReaderCount` once, depending on which operations error; the previous code would still do these calls:

```
commitLogSeriesID = entry.Series.ID()			commitLogSeriesID = entry.Series.ID()
commitLogSeriesTags = entry.Series.Tags()			commitLogSeriesTags = entry.Series.Tags()
commitLogSeriesUniqueIndex = entry.Index
```

for no reason if `.Write(..)` errored. Happy to revert it though if this is fine

src/dbnode/persist/fs/commitlog/commit_log_test.go
```diff
@@ -130,6 +131,7 @@ func (w testWrite) assert(
 	unit xtime.Unit,
 	annotation []byte,
 ) {
+	fmt.Println("Doing this, series", int(w.series.UniqueIndex), "no w", int(series.UniqueIndex))
```
richardartoul: delete

src/dbnode/storage/database.go
```diff
@@ -694,17 +695,23 @@ func (d *db) writeBatch(
 			errHandler.HandleError(write.OriginalIndex, err)
 		}
 
-		// Need to set the outcome in the success case so the commitlog gets the updated
-		// series object which contains identifiers (like the series ID) whose lifecycle
-		// live longer than the span of this request, making them safe for use by the async
-		// commitlog. Need to set the outcome in the error case so that the commitlog knows
-		// to skip this entry.
-		writes.SetOutcome(i, series, err)
+		if !wasWritten {
```
richardartoul: I think you still need to set the outcome I.E dont use an else statement otherwise we'll never report errors back to the caller.

Any way we can improve this with a unit test too?

richardartoul: like if err != nil then wasWritten will be false, but we won't record the error in the outcome this way

src/dbnode/storage/shard.go
richardartoul: is this no longer relevant? @arnikola Not sure what I wrote this in reference to

src/dbnode/storage/shard.go
richardartoul: yeah I think i'd prefer not to change this for now as I think its a little more brittle now (cleanup code more spread out)

src/dbnode/ts/types.go
```diff
@@ -38,6 +38,9 @@ type Write struct {
 // BatchWrite represents a write that was added to the
 // BatchWriter.
 type BatchWrite struct {
+	// Determines if this series should be written or not. If this is false, the
+	// series at this index will be cleaned on
```
richardartoul: cleaned on?

src/dbnode/storage/database.go
arnikola: Ah good catch. I'll make a change that if it errors, it will always `SetOutcome(i, _, err)` first

src/dbnode/ts/types.go
arnikola: Oops

src/dbnode/persist/fs/commitlog/commit_log_test.go
arnikola: Luckliy this is the println I left in rather than one of the more objectionable ones :p

src/dbnode/storage/shard.go
arnikola: I'm not super sure what it's referring to, sorry... Either way I think it should be fine since worst case scenario in this case is that this series will be written twice to the commit log, right?

src/dbnode/storage/database.go
```diff
@@ -694,17 +695,22 @@ func (d *db) writeBatch(
...
+			// This series has no additional information that needs to be written to
+			// the commit log; set this series to skip writing to the commit log.
+			writes.SetSkipWrite(i)
+		} else {
+			// Need to set the outcome in the success case so the commitlog gets the
+			// updated series object which contains identifiers (like the series ID)
+			// whose lifecycle lives longer than the span of this request, making them
+			// safe for use by the async commitlog. Need to set the outcome in the
+			// error case so that the commitlog knows to skip this entry.
+			writes.SetOutcome(i, series, err)
```
richardartoul: Not sure if I brought this up last time, but is there any reason we can't just `SetOutcome` in both cases? Feels a little safer to make sure everything gets updated anyways and I'm pretty sure its cheap since it literally just does this

```
func (b *writeBatch) SetOutcome(idx int, series Series, err error) {
	b.writes[idx].Write.Series = series
	b.writes[idx].Err = err
}
```

richardartoul: So basically always SetOutcome

and then 

```golang
if !wasWritten || err != nil {
    writes.SetSkipWrite(i)
}
```

richardartoul: Notice that I changed it to an OR because you always want to skip the write if it wasn't written or there was an error. I know the commitlog will skip it anyways if there was an error but this seems more semantically correct

arnikola: Sure, sounds good to me; makes sense that the readability is worth more here

PR #1376, opened 2019-02-14
Obtain a lock on `filePathPrefix` on dbnode startup. Close #641

### LINKED ISSUE #641: Use a lockfile in filePathPrefix

edkolev:
As discussed in #641, with this PR a lock on `filePathPrefix` will be acquired when starting a db node. Two dbnode processes will not be able to run with the same `filePathPrefix`.

In case another process already has a lock, this error message would be printed:
```
16:07:39.082121[F] could not acquire lock on /tmp/m3db/.lock: failed obtaining lock: resource temporarily unavailable
```

When the dbnode exits gracefully, the lock file will be removed (using a `defer`).

When the dbnode doesn't exit gracefully, another dbnode process will be able to obtain the lock. A bit more details: there's a distinction b/w the *lock file* which resides on the FS, and the *lock* which resides in memory; when the dbnode is killed ungracefully the *lock file* will remain on the FS, but the *lock* will be removed from memory; the next time a dbnode starts, it will be able to acquire a *lock* despite the fact that the *lock file* exists.

@richardartoul please let me know if this PR can be improved.

### DIFF
--- a/src/dbnode/server/server.go
+++ b/src/dbnode/server/server.go
@@ -28,6 +28,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"path"
 	"runtime"
 	"runtime/debug"
 	"syscall"
@@ -67,6 +68,7 @@ import (
 	"github.com/m3db/m3/src/m3ninx/postings"
 	"github.com/m3db/m3/src/m3ninx/postings/roaring"
 	xdocs "github.com/m3db/m3/src/x/docs"
+	"github.com/m3db/m3/src/x/lockfile"
 	"github.com/m3db/m3/src/x/mmap"
 	"github.com/m3db/m3/src/x/serialize"
 	xconfig "github.com/m3db/m3x/config"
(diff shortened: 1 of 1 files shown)


src/dbnode/server/server.go
```diff
@@ -318,6 +321,14 @@ func Run(runOpts RunOptions) {
 		SetForceIndexSummariesMmapMemory(cfg.Filesystem.ForceIndexSummariesMmapMemory).
 		SetForceBloomFilterMmapMemory(cfg.Filesystem.ForceBloomFilterMmapMemory)
 
+	// obtain a lock on `filePathPrefix`
```
richardartoul: Is this as high as we can easily move this code? Also, can you add a similar comment to the one you put in this P.R and leave it above this code?

edkolev: Sure, I've now extended the comment. The lock acquisition is also move up, it's almost the first thing done, with only the logger initialisation being before it. Let me know in case that's too early.

edkolev:
Also let me know if I need to squash.

edkolev:
> We've almost figured out how to run C.I for P.Rs from external contributors

Nice :)

The conflict is now resolved, commits squashed.

PR #1377, opened 2019-02-14
Updated pilosa import to explicitly use m3db fork

simonrobb:
Projects using `go mod` for dependency management have issues when M3 is a dependency, since `go mod` doesn't allow overriding the repo for a package in the way Glide does.


simonrobb:
Thanks. Yep `glide up` was fine 👍

PR #1378, opened 2019-02-14
Switch imports for forked Pilosa and fix various lint issues

richardartoul:


src/cmd/services/m3query/config/config.go
```diff
@@ -180,6 +179,26 @@ func (c Configuration) LookbackDurationOrDefault() (time.Duration, error) {
 	return v, nil
 }
 
+// LimitsOrDefault returns the specified limit configuration if provided, or the
+// default value otherwise.
+func (c Configuration) LimitsOrDefault() LimitsConfiguration {
```
simonrobb: Should this return `*LimitsConfiguration`?

simonrobb: nvm I see you've updated

PR #1382, opened 2019-02-14
Configure index block size in namespace/database create APIs

richardartoul:


src/query/api/v1/handler/database/create.go
```diff
@@ -84,33 +84,68 @@ const (
 
 	minRecommendCalculateBlockSize = 30 * time.Minute
 	maxRecommendCalculateBlockSize = 24 * time.Hour
+
+	// Amount to multiply a calculated data block size by to generate a reasonable
+	// index block size.
+	dataToIndexBlockSizeScalingFactor = 2
```
prateek: instead of this heuristic, could you add an `indexBlockSize` field to the `recommendedBlockSize` struct

richardartoul: Thats what I did originally, but it just ended up being 2x block size for each one anyways. Want me to do it anyways?

richardartoul: @prateek Pinging since I think this is the last blocker

prateek: yeah probably best to be explicit about it. for e.g. i don't think we'd want a 2w index block size in the largest e.g. you have.

PR #1384, opened 2019-02-15
Implement commitlog rotation as well as writing out, cleaning up, and bootstrapping using snapshot metadata files

richardartoul:
# Summary

This P.R implements the following functionality:

- [X] Rotate commitlogs, generate snapshot ID, and write out snapshot metadata/checkpoint files for each snapshot.
- [X] Rewrite snapshot/commitlog cleanup logic to take new snapshotting process into account.
- [X] Most minimal possible changes to commitlog bootstrapping logic to make it work with the new snapshot metadata files.
- [X] Support a fully transparent upgrade for users (even those using the commitlog bootstrapper) by keeping the bootstrapping logic backwards-compatible with previous version of M3DB.
- [X] Removal of the concept of commitlog blocksize.

# Review guide 

Unfortunately its very difficult to make this P.R any smaller because as soon as writing out the snapshot metadata files is implemented, the cleanup logic needs to be implemented, and if the cleanup logic is implemented then the bootstrapping logic needs to be implemented as well.

Luckily, while this is a large P.R, there are much more deletions than additions, and the vast majority of the new code is a simplification of previously much more complicated code. I,E a lot of the code that was rewritten went from very complicated logic to very simple logic.

Places to pay particular attention to in the review:

1. Cleanup logic in `cleanup.go`
2. Commitlog rotation logic in `commitlog.go` (removed the concept of rotating files by time entirely, and commitlog files will only be rotated when an explicit call to `RotateLogs()` is made.
3. Removal of the concept of a commitlog block size, and also while we retained the existing commitlog filename structure, all commitlog files will have a name that defaults to a blockStart of time.Unix(0, 0). Because the commitlog files have info headers and we don't need to rely on their names at all, I believe we can remove the blockStart from their name entirely, but I'd like to leave it out of this P.R as its already too big.
4. Changes to the PersistManager to support writing out the snapshot ID and commitlog ID when the entire snapshot process completes in `persist_manager.go`
5. Changes to the commitlog bootstrapping logic (mostly code deletion) that make it work for both the previous implementation and the current implementation. This is accomplished by simplifying the logic to: Read the most recent snapshot file for every namespace/shard/block combination and ALL commitlog files that exist on disk. This will be slow, but correct, and a subsequent P.R will make this significantly more efficient will still remaining backwards compatible with files generated from previous versions of M3DB (trying to make minimal possible changes to bootstrapping in this P.R to keep it from growing obscenely large). Also confidence that this is backwards compatible is maintained because I left the existing prop test unmodified which generates combinations of snapshots/commitlogs that look like they came from an older version of M3DB.
6. The logic for rotating the commitlog, generating the snapshot ID, and performing the snapshot in `flush.go`


src/dbnode/persist/fs/commitlog/files.go
```diff
@@ -27,20 +27,54 @@ import (
...
+	for ; ; newIndex++ {
+		var (
+			prefix = opts.FilesystemOptions().FilePathPrefix()
+			// We pass timeNone for the commitlog file block start because that field
+			// is no longer required and can just be zero.
+			// TODO(rartoul): It should actually be completely backwards compatible to
+			// change the commitlog filename structure (because we don't rely on the name
+			// for any information, we just list all the files in a directory and then
+			// read their encoded heads to obtain information about them), so in the future
+			// we can just get rid of this.
```
robskillington: Can you add a `TODO(v1): remove this when ok to be backwards incompatible`

src/dbnode/persist/fs/msgpack/encoder.go
```diff
@@ -266,8 +266,11 @@ func (enc *Encoder) encodeIndexSummary(summary schema.IndexSummary) {
 
 func (enc *Encoder) encodeLogInfo(info schema.LogInfo) {
 	enc.encodeNumObjectFieldsForFn(logInfoType)
-	enc.encodeVarintFn(info.Start)
-	enc.encodeVarintFn(info.Duration)
+
+	// Deprecated, have to encode anyways for backwards compatibility, but we ignore the values.
```
robskillington: Can you add a `TODO(v1): remove when we can make backwards incompatible changes with an upgrade to v1`

src/dbnode/persist/fs/commitlog/files.go
richardartoul: I think this is just a regular TODO because it is backwards compatible (existing code doesn't rely on filename). I just didn't want to make the change in this P.R

robskillington: Sure thing.

PR #1397, opened 2019-02-21
[m3nsch] support generating new unique metrics

schallert:
`m3nsch` supports generating load for a defined set of metrics, but does
not currently support generating new uniques over time. This allows
specifying a percentage of generated metrics that will be unique.


src/m3nsch/examples/sample_workload.go
```diff
@@ -1,3 +1,5 @@
+//+build never
```
prateek: lol why?

schallert: It was causing lint errors or some similar build error when I was testing and I wanted to ignore it, I'll loop back and try to figure out what that was

schallert:
Looks like something's off with my approach here. The below screenshot is from runs with `u` equal to `0.5`, `1.0`, and `0.2` respectively. The general goal is accomplished, in that uniques scale with `u`, but it's not the 1:1 behavior I was hoping for. I would have expected to see 3k unique/s when `u=1.0`. Not sure if I'm misunderstanding the write behavior or if my numbers are bad.

Admittedly I don't want to sink a lot of time into this if the general spirit is accomplished, so may merge as is after a bit more testing.

screen shot 2019-02-25 at 1 53 32 pm

prateek:
@schallert yep the behaviour you're seeing is expected. The agent has a bunch of "workers" (goroutines) which are assigned a set of metrics to generate traffic for every time the workload changes - [1], [2]. Subsequently, in each worker's event loop they just use the metrics they're assigned to create writes - [3]. 

Your change is only going thru the code-path at [1]/[2]. i.e. You're only creating new uniques when the workload is updated. You'll need to update the logic to be triggered during [3] if you want the uniques to be consistently generated. 

[1]: https://github.com/m3db/m3/blob/master/src/m3nsch/agent/agent.go#L166
[2]: https://github.com/m3db/m3/blob/master/src/m3nsch/agent/agent.go#L117-L149
[3]: https://github.com/m3db/m3/blob/master/src/m3nsch/agent/agent.go#L301

PR #1398, opened 2019-02-21
Add lru cache for query conversion

benraskin92:
W/O cache
screen shot 2019-02-21 at 1 06 07 pm

W/ cache
screen shot 2019-02-21 at 12 58 44 pm

### DIFF
--- a/scripts/development/m3_stack/m3coordinator.yml
+++ b/scripts/development/m3_stack/m3coordinator.yml
@@ -55,3 +55,7 @@ carbon:
 
 tagOptions:
   idScheme: quoted
+
+cache:
+  queryConversion:
+    size: 200
--- a/scripts/development/m3_stack/m3coordinator.yml
+++ b/scripts/development/m3_stack/m3coordinator.yml
@@ -55,7 +55,3 @@ carbon:
 
 tagOptions:
   idScheme: quoted
-
-cache:
-  queryConversion:
-    size: 200
--- a/src/cmd/services/m3dbnode/config/cache.go
+++ b/src/cmd/services/m3dbnode/config/cache.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/src/cmd/services/m3dbnode/config/cache.go
+++ b/src/cmd/services/m3dbnode/config/cache.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2019 Uber Technologies, Inc.
+// Copyright (c) 2017 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/src/cmd/services/m3query/config/config.go
+++ b/src/cmd/services/m3query/config/config.go
@@ -54,7 +54,7 @@ const (
 		"generation scheme is required in coordinator configuration settings. " +
 		"More information is available here: %s"
 
-	defaultQueryConversionCacheSize = 100
+	defaultQueryConversionCacheSize = 4096
 )
 
 var (
--- a/src/cmd/services/m3query/config/config.go
+++ b/src/cmd/services/m3query/config/config.go
@@ -53,6 +53,8 @@ const (
 	errNoIDGenerationScheme            = "error: a recent breaking change means that an ID " +
 		"generation scheme is required in coordinator configuration settings. " +
 		"More information is available here: %s"
+
+	defaultQueryConversionCacheSize = 100
 )
 
 var (
@@ -123,6 +125,9 @@ type Configuration struct {
 
 	// LookbackDuration determines the lookback duration for queries
 	LookbackDuration *time.Duration `yaml:"lookbackDuration"`
+
+	// Cache configurations.
+	Cache CacheConfigurations `yaml:"cache"`
 }
 
 // Filter is a query filter type.
@@ -144,6 +149,37 @@ type FilterConfiguration struct {
 	Read         Filter `yaml:"read"`
 	Write        Filter `yaml:"write"`
 	CompleteTags Filter `yaml:"completeTags"`
+}
+
+// CacheConfigurations is the cache configurations.
+type CacheConfigurations struct {
+	// QueryConversion cache policy.
+	QueryConversion *QueryConversionCacheConfiguration `yaml:"queryConversion"`
+}
+
+// QueryConversionCacheConfiguration is the query conversion cache configuration.
+type QueryConversionCacheConfiguration struct {
+	Size *int `yaml:"size"`
+}
+
+// QueryConversionCacheConfiguration returns the query conversion cache configuration
+// or default if none is specified.
+func (c CacheConfigurations) QueryConversionCacheConfiguration() QueryConversionCacheConfiguration {
+	if c.QueryConversion == nil {
+		return QueryConversionCacheConfiguration{}
+	}
+
+	return *c.QueryConversion
+}
+
+// SizeOrDefault returns the provided size or the default value is none is
+// provided.
+func (p *QueryConversionCacheConfiguration) SizeOrDefault() int {
+	if p.Size == nil {
+		return defaultQueryConversionCacheSize
+	}
+
+	return *p.Size
 }
 
 // LimitsConfiguration represents limitations on per-query resource usage. Zero or negative values imply no limit.
--- a/src/cmd/services/m3query/config/config.go
+++ b/src/cmd/services/m3query/config/config.go
@@ -127,7 +127,7 @@ type Configuration struct {
 	LookbackDuration *time.Duration `yaml:"lookbackDuration"`
 
 	// Cache configurations.
-	Cache CacheConfigurations `yaml:"cache"`
+	Cache CacheConfiguration `yaml:"cache"`
 }
 
 // Filter is a query filter type.
--- a/src/cmd/services/m3query/config/config.go
+++ b/src/cmd/services/m3query/config/config.go
@@ -184,11 +184,14 @@ func (q *QueryConversionCacheConfiguration) SizeOrDefault() int {
 
 // Validate validates the QueryConversionCacheConfiguration settings.
 func (q *QueryConversionCacheConfiguration) Validate() error {
-	if *q.Size <= 0 {
+	switch {
+	case q.Size == nil:
+		return nil
+	case *q.Size <= 0:
 		return fmt.Errorf("must provide a positive size for query conversion config, instead got: %d", *q.Size)
+	default:
+		return nil
 	}
-
-	return nil
 }
 
 // LimitsConfiguration represents limitations on per-query resource usage. Zero or negative values imply no limit.
--- a/src/cmd/services/m3query/config/config.go
+++ b/src/cmd/services/m3query/config/config.go
@@ -174,15 +174,21 @@ func (c CacheConfiguration) QueryConversionCacheConfiguration() QueryConversionC
 
 // SizeOrDefault returns the provided size or the default value is none is
 // provided.
-func (q *QueryConversionCacheConfiguration) SizeOrDefault() (int, error) {
-	switch {
-	case q.Size == nil:
-		return defaultQueryConversionCacheSize, nil
-	case *q.Size <= 0:
-		return -1, fmt.Errorf("must provide a positive size for query conversion config, instead got: %d", *q.Size)
-	default:
-		return *q.Size, nil
+func (q *QueryConversionCacheConfiguration) SizeOrDefault() int {
+	if q.Size == nil {
+		return defaultQueryConversionCacheSize
 	}
+
+	return *q.Size
+}
+
+// Validate validates the QueryConversionCacheConfiguration settings.
+func (q *QueryConversionCacheConfiguration) Validate() error {
+	if *q.Size <= 0 {
+		return fmt.Errorf("must provide a positive size for query conversion config, instead got: %d", *q.Size)
+	}
+
+	return nil
 }
 
 // LimitsConfiguration represents limitations on per-query resource usage. Zero or negative values imply no limit.
--- a/src/cmd/services/m3query/config/config.go
+++ b/src/cmd/services/m3query/config/config.go
@@ -151,8 +151,8 @@ type FilterConfiguration struct {
 	CompleteTags Filter `yaml:"completeTags"`
 }
 
-// CacheConfigurations is the cache configurations.
-type CacheConfigurations struct {
+// CacheConfiguration is the cache configurations.
+type CacheConfiguration struct {
 	// QueryConversion cache policy.
 	QueryConversion *QueryConversionCacheConfiguration `yaml:"queryConversion"`
 }
@@ -164,7 +164,7 @@ type QueryConversionCacheConfiguration struct {
 
 // QueryConversionCacheConfiguration returns the query conversion cache configuration
 // or default if none is specified.
-func (c CacheConfigurations) QueryConversionCacheConfiguration() QueryConversionCacheConfiguration {
+func (c CacheConfiguration) QueryConversionCacheConfiguration() QueryConversionCacheConfiguration {
 	if c.QueryConversion == nil {
 		return QueryConversionCacheConfiguration{}
 	}
@@ -174,12 +174,15 @@ func (c CacheConfigurations) QueryConversionCacheConfiguration() QueryConversion
 
 // SizeOrDefault returns the provided size or the default value is none is
 // provided.
-func (q *QueryConversionCacheConfiguration) SizeOrDefault() int {
-	if q.Size == nil {
-		return defaultQueryConversionCacheSize
+func (q *QueryConversionCacheConfiguration) SizeOrDefault() (int, error) {
+	switch {
+	case q.Size == nil:
+		return defaultQueryConversionCacheSize, nil
+	case *q.Size <= 0:
+		return -1, fmt.Errorf("must provide a positive size for query conversion config, instead got: %d", *q.Size)
+	default:
+		return *q.Size, nil
 	}
-
-	return *q.Size
 }
 
 // LimitsConfiguration represents limitations on per-query resource usage. Zero or negative values imply no limit.
--- a/src/cmd/services/m3query/config/config.go
+++ b/src/cmd/services/m3query/config/config.go
@@ -184,14 +184,11 @@ func (q *QueryConversionCacheConfiguration) SizeOrDefault() int {
 
 // Validate validates the QueryConversionCacheConfiguration settings.
 func (q *QueryConversionCacheConfiguration) Validate() error {
-	switch {
-	case q.Size == nil:
-		return nil
-	case *q.Size <= 0:
+	if q.Size != nil && *q.Size <= 0 {
 		return fmt.Errorf("must provide a positive size for query conversion config, instead got: %d", *q.Size)
-	default:
-		return nil
 	}
+
+	return nil
 }
 
 // LimitsConfiguration represents limitations on per-query resource usage. Zero or negative values imply no limit.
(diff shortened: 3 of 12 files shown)


src/cmd/services/m3query/config/config.go
```diff
@@ -53,6 +53,8 @@ const (
 	errNoIDGenerationScheme            = "error: a recent breaking change means that an ID " +
 		"generation scheme is required in coordinator configuration settings. " +
 		"More information is available here: %s"
+
+	defaultQueryConversionCacheSize = 100
```
benraskin92: Probably can be a lot higher?

src/cmd/services/m3query/config/config.go
```diff
@@ -120,6 +122,9 @@ type Configuration struct {
 
 	// LookbackDuration determines the lookback duration for queries
 	LookbackDuration *time.Duration `yaml:"lookbackDuration"`
+
+	// Cache configurations.
+	Cache CacheConfigurations `yaml:"cache"`
```
benraskin92: Added this instead of directly adding query conversion cache in case we have more cache options to add in the future

scripts/development/m3_stack/m3coordinator.yml
```diff
@@ -55,3 +55,7 @@ carbon:
```
benraskin92: Either add this to all configs or remove since it defaults

src/query/storage/query_conversion_lru.go
```diff
@@ -0,0 +1,113 @@
...
+		items:     make(map[interface{}]*list.Element),
+	}
+
+	return c, nil
+}
+
+// AddWithLock adds a value to the cache. Returns true if an eviction occurred.
+func (c *QueryConversionLRU) AddWithLock(key string, value idx.Query) (evicted bool) {
+	// Check for existing item
+	c.Lock()
```
andrewmains12: I'm a fan of the pattern the hashicorp implementation uses for this--implement all the datastructure operations without a lock, and then wrap that class in a thin class with a lock. Saves you on defer overhead while still maintaining lock safety fairly easily. Thoughts?

src/query/storage/query_conversion_lru.go
```diff
@@ -0,0 +1,113 @@
...
+	if evict {
+		c.removeOldest()
+	}
+
+	c.Unlock()
+	return evict
+}
+
+// GetWithLock looks up a key's value from the cache.
+func (c *QueryConversionLRU) GetWithLock(key string) (value idx.Query, ok bool) {
```
andrewmains12: If `key` is always a string, why is `items` `map[interface]`?

src/query/storage/query_conversion_lru.go
```diff
@@ -0,0 +1,113 @@
...
+	"container/list"
+	"fmt"
+	"sync"
+
+	"github.com/m3db/m3/src/m3ninx/idx"
+)
+
+// QueryConversionLRU implements a fixed size LRU cache
+type QueryConversionLRU struct {
+	sync.Mutex
```
andrewmains12: Nit: embedding is weird here imo (sort of implies that the LRU *is* a lock); I would just use an explicit field.

src/query/storage/query_conversion_lru.go
```diff
@@ -0,0 +1,113 @@
...
+// NewQueryConversionLRU constructs an LRU of the given size
+func NewQueryConversionLRU(size int) (*QueryConversionLRU, error) {
+	if size <= 0 {
+		return nil, fmt.Errorf("must provide a positive size, instead got: %d", size)
+	}
+
+	c := &QueryConversionLRU{
+		size:      size,
+		evictList: list.New(),
+		items:     make(map[interface{}]*list.Element),
```
andrewmains12: Size is known here, so let's use it in `make`

src/query/storage/query_conversion_lru.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	return c, nil
+}
+
+// AddWithLock adds a value to the cache. Returns true if an eviction occurred.
+func (c *QueryConversionLRU) AddWithLock(key string, value idx.Query) (evicted bool) {
```
andrewmains12: Nit: I don't think the `WithLock` suffixes are really necessary; we can comment the class as thread safe or not instead.

scripts/development/m3_stack/m3coordinator.yml
arnikola: I'd err towards default

src/query/storage/query_conversion_lru.go
```diff
@@ -0,0 +1,104 @@
...
+		size:      size,
+		evictList: list.New(),
+		items:     make(map[string]*list.Element, size),
+	}
+
+	return c, nil
+}
+
+// Add adds a value to the cache. Returns true if an eviction occurred.
+func (c *QueryConversionLRU) Add(key string, value idx.Query) (evicted bool) {
```
arnikola: Can you keep key as a []byte here, then cast it only when indexing into the map?

src/query/storage/query_conversion_lru.go
```diff
@@ -0,0 +1,104 @@
...
+// QueryConversionLRU implements a fixed size LRU cache
+type QueryConversionLRU struct {
+	size      int
+	evictList *list.List
+	items     map[string]*list.Element
+}
+
+// entry is used to hold a value in the evictList
+type entry struct {
+	key   string
```
arnikola: Can you keep key as a `[]byte`

src/query/storage/query_conversion_lru.go
arnikola: Also nit: `Set` rather than `Add`

src/query/storage/query_conversion_lru_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLRU(t *testing.T) {
+	lru, err := NewQueryConversionLRU(5)
+	require.NoError(t, err)
+
+	// test add and get
+	lru.Add("a", idx.NewTermQuery([]byte("foo"), []byte("bar")))
```
arnikola: Can you check the return type with `assert.False(t, lru.Add("a", idx.NewTermQuery([]byte("foo"), []byte("bar")))`

Also a test for overriding the value returning `true` and changing the value in the map?

src/query/storage/query_conversion_lru_test.go
```diff
@@ -0,0 +1,61 @@
...
+	assert.Equal(t, "term(foo, bar)", q.String())
+	q, ok = lru.Get("b")
+	require.True(t, ok)
+	assert.Equal(t, "term(biz, baz)", q.String())
+
+	// fill up the cache
+	lru.Add("c", idx.NewTermQuery([]byte("bar"), []byte("foo")))
+	lru.Add("d", idx.NewTermQuery([]byte("baz"), []byte("biz")))
+	lru.Add("e", idx.NewTermQuery([]byte("qux"), []byte("quz")))
+	lru.Add("f", idx.NewTermQuery([]byte("quz"), []byte("qux")))
```
arnikola: Can you check that the return types for the others are `false` and the return type for this one is `true`?

src/query/storage/m3/storage.go
```diff
@@ -72,19 +73,26 @@ func NewStorage(
...
 	lookbackDuration time.Duration,
-) Storage {
+	conversionCacheSize int,
+) (Storage, error) {
 	opts := m3db.NewOptions().
 		SetTagOptions(tagOptions).
 		SetLookbackDuration(lookbackDuration).
 		SetConsolidationFunc(consolidators.TakeLast)
 
+	conversionLRU, err := storage.NewQueryConversionLRU(conversionCacheSize)
```
arnikola: I think it would make a bit more sense to create the cache above and then pass the cache into the storage constructor?

src/query/storage/m3/storage.go
```diff
@@ -72,19 +73,26 @@ func NewStorage(
...
+	}
+
 	return &m3storage{
 		clusters:        clusters,
 		readWorkerPool:  readWorkerPool,
 		writeWorkerPool: writeWorkerPool,
 		opts:            opts,
 		nowFn:           time.Now,
-	}
+		conversionCache: &storage.QueryConvserionCache{LRU: conversionLRU},
```
arnikola: Can you make a constructor for this, taking in the LRU as an argument?

src/query/storage/index.go
```diff
@@ -90,16 +98,48 @@ func FetchOptionsToM3Options(fetchOptions *FetchOptions, fetchQuery *FetchQuery)
...
+	}
+
 	// Optimization for single matcher case.
 	if len(matchers) == 1 {
 		q, err := matcherToQuery(matchers[0])
 		if err != nil {
 			return index.Query{}, err
 		}
 
+		cache.LRU.Add(string(k), q)
```
arnikola: nit: rather than calling `.LRU.Add()`, can you put a thin wrapper `Add` function onto the query conversion cache?

src/query/storage/index.go
```diff
@@ -22,13 +22,21 @@ package storage
...
 
 	"github.com/m3db/m3/src/dbnode/storage/index"
 	"github.com/m3db/m3/src/m3ninx/idx"
 	"github.com/m3db/m3/src/query/models"
 	"github.com/m3db/m3x/ident"
 )
 
+// QueryConvserionCache represents the query conversion LRU cache
+type QueryConvserionCache struct {
+	mu sync.RWMutex
```
arnikola: Since you're never using the Read lock, might be better to just make this a simple `sync.Mutex`

src/query/storage/index.go
benraskin92: Any specific reason to do this?

src/query/storage/query_conversion_lru_test.go
benraskin92: Not sure I understand your second point here?

src/cmd/services/m3query/config/config.go
```diff
@@ -146,6 +151,37 @@ type FilterConfiguration struct {
 	CompleteTags Filter `yaml:"completeTags"`
 }
 
+// CacheConfigurations is the cache configurations.
+type CacheConfigurations struct {
```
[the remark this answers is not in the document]
benraskin92: I think it's better with the `s` as it implies it can be used for multiple cache configs.

src/cmd/services/m3query/config/config.go
```diff
@@ -146,6 +151,37 @@ type FilterConfiguration struct {
...
+// SizeOrDefault returns the provided size or the default value is none is
+// provided.
+func (q *QueryConversionCacheConfiguration) SizeOrDefault() int {
+	if q.Size == nil {
+		return defaultQueryConversionCacheSize
+	}
+
+	return *q.Size
+}
+
```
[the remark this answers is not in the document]
benraskin92: I check this when creating the new lru cache 
```
	if size <= 0 {
		return nil, fmt.Errorf("must provide a positive size, instead got: %d", size)
	}
```

src/query/storage/index.go
arnikola: Reads better, more clarity; also if at some point we want to add more logic to the add function we can keep it in one place

src/cmd/services/m3query/config/config.go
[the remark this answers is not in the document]
arnikola: Better to check it straight away so it errors out during setup rather than during execution of queries

src/cmd/services/m3query/config/config.go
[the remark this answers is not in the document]
arnikola: All other configs are `XyzConfiguration` even with multiple configs, should probably stick to the precedent here

src/query/storage/query_conversion_lru_test.go
arnikola: Something like

```
q, ok = lru.Get("a")
	require.True(t, ok)
	assert.Equal(t, "term(foo, bar)", q.String())
assert.False(t, lru.Add("a", idx.NewTermQuery([]byte("foo"), []byte("bar")))
assert.True(t, lru.Add("a", idx.NewTermQuery([]byte("foo"), []byte("baz")))
q, ok = lru.Get("a")
	require.True(t, ok)
	assert.Equal(t, "term(foo, baz)", q.String())
```

To ensure it overwrites

src/query/storage/m3/storage.go
```diff
@@ -72,19 +73,27 @@ func NewStorage(
...
+	// }
+
 	return &m3storage{
 		clusters:        clusters,
 		readWorkerPool:  readWorkerPool,
 		writeWorkerPool: writeWorkerPool,
 		opts:            opts,
 		nowFn:           time.Now,
-	}
+		// conversionCache: &storage.QueryConversionCache{LRU: conversionLRU},
```
benraskin92: Remove

src/query/storage/m3/storage.go
```diff
@@ -72,19 +73,27 @@ func NewStorage(
...
 	lookbackDuration time.Duration,
-) Storage {
+	queryConversionCache *storage.QueryConversionCache,
+) (Storage, error) {
 	opts := m3db.NewOptions().
 		SetTagOptions(tagOptions).
 		SetLookbackDuration(lookbackDuration).
 		SetConsolidationFunc(consolidators.TakeLast)
 
+	// conversionLRU, err := storage.NewQueryConversionLRU(conversionCacheSize)
```
benraskin92: Remove

scripts/development/m3_stack/m3coordinator.yml
```diff
@@ -55,3 +55,7 @@ carbon:
 
 tagOptions:
   idScheme: quoted
+
+cache:
+  queryConversion:
+    size: 200
```
arnikola: nit: this is very low

src/cmd/services/m3dbnode/config/cache.go
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2017 Uber Technologies, Inc.
+// Copyright (c) 2019 Uber Technologies, Inc.
```
arnikola: nit: revert

src/query/storage/index.go
```diff
@@ -22,13 +22,36 @@ package storage
...
 	"github.com/m3db/m3/src/m3ninx/idx"
 	"github.com/m3db/m3/src/query/models"
 	"github.com/m3db/m3x/ident"
 )
 
+// QueryConversionCache represents the query conversion LRU cache
+type QueryConversionCache struct {
+	mu sync.Mutex
+
+	LRU *QueryConversionLRU
```
arnikola: nit: can make this private now

src/query/storage/index_test.go
```diff
@@ -142,10 +149,22 @@ func TestFetchQueryToM3Query(t *testing.T) {
...
-			m3Query, err := FetchQueryToM3Query(fetchQuery)
+			m3Query, err := FetchQueryToM3Query(fetchQuery, cache)
 			require.NoError(t, err)
 			assert.Equal(t, test.expected, m3Query.String())
 		})
 	}
+}
+
+func TestQueryKey(t *testing.T) {
+	matchers := models.Matchers{
```
arnikola: Can you add a few matchers here, maybe with different `Type`s

src/query/storage/query_conversion_lru_test.go
```diff
@@ -0,0 +1,72 @@
...
+	require.False(t, evicted)
+	evicted = lru.Set([]byte("f"), idx.NewTermQuery([]byte("quz"), []byte("qux")))
+	require.True(t, evicted)
+
+	// make sure "a" is no longer in the cache
+	_, ok = lru.Get([]byte("a"))
+	require.False(t, ok)
+
+	// make sure "b" is still in the cache
+	_, ok = lru.Get([]byte("b"))
```
arnikola: nit: might as well check value here

src/cmd/services/m3query/config/config.go
```diff
@@ -146,6 +151,46 @@ type FilterConfiguration struct {
...
+// Validate validates the QueryConversionCacheConfiguration settings.
+func (q *QueryConversionCacheConfiguration) Validate() error {
+	if *q.Size <= 0 {
```
arnikola: Is 0 valid?

src/query/server/server.go
```diff
@@ -608,13 +608,33 @@ func newStorages(
...
 	cleanup := func() error { return nil }
 
-	localStorage := m3.NewStorage(
+	// Setup query conversion cache.
+	var (
+		conversionCacheConfig = cfg.Cache.QueryConversionCacheConfiguration()
+		conversionCacheSize   = conversionCacheConfig.SizeOrDefault()
+	)
+
+	if err := conversionCacheConfig.Validate(); err != nil {
```
arnikola: Should validate prior to calling SizeOrDefault here

src/query/storage/index_test.go
```diff
@@ -142,10 +149,69 @@ func TestFetchQueryToM3Query(t *testing.T) {
...
+					Value: []byte("v1"),
+				},
+			},
+		},
+	}
 
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			keyByte := queryKey(test.matchers)
+			assert.Equal(t, []byte(test.expected), keyByte)
```
arnikola: Can you add a test with multiple matchers

src/query/storage/index_test.go
```diff
@@ -142,10 +149,69 @@ func TestFetchQueryToM3Query(t *testing.T) {
 				Interval:    15 * time.Second,
 			}
 
-			m3Query, err := FetchQueryToM3Query(fetchQuery)
+			m3Query, err := FetchQueryToM3Query(fetchQuery, cache)
```
arnikola: Can this verify that values are written to cache?

src/query/storage/query_conversion_lru_test.go
```diff
@@ -0,0 +1,73 @@
...
+import (
+	"testing"
+
+	"github.com/m3db/m3/src/m3ninx/idx"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestLRU(t *testing.T) {
```
arnikola: Add a test for rewriting values

src/query/server/server.go
benraskin92: then we should just check in `SizeOrDefault()`. having a validate doesn't make much sense unless we need to validate multiple things

src/cmd/services/m3query/config/config.go
benraskin92: No

src/cmd/services/m3query/config/config.go
```diff
@@ -146,6 +151,49 @@ type FilterConfiguration struct {
...
+	if q.Size == nil {
+		return defaultQueryConversionCacheSize
+	}
+
+	return *q.Size
+}
+
+// Validate validates the QueryConversionCacheConfiguration settings.
+func (q *QueryConversionCacheConfiguration) Validate() error {
+	switch {
```
arnikola: Would be cleaner as

`if q.Size != nil && q.Size <= 0 { // return  error...`

src/query/server/server.go
arnikola: The benefit of validating like this is when we add config options to the validation method, won't have to worry about it being called in time, since it's already validated on startup

src/query/storage/query_conversion_lru_test.go
```diff
@@ -0,0 +1,78 @@
...
+
+	// make sure "b" is still in the cache
+	q, ok = lru.Get([]byte("b"))
+	require.True(t, ok)
+	assert.Equal(t, "term(biz, baz)", q.String())
+
+	// rewrite "e" and make sure nothing gets evicted
+	// since "e" is already in the cache.
+	evicted = lru.Set([]byte("e"), idx.NewTermQuery([]byte("qux"), []byte("quz")))
+	require.False(t, evicted)
```
arnikola: nit: check that the value of `e` changes?

src/cmd/services/m3query/config/config_test.go
```diff
@@ -162,3 +162,13 @@ func TestTagOptionsConfig(t *testing.T) {
 	assert.Equal(t, []byte("foo"), opts.BucketName())
 	assert.Equal(t, models.TypePrependMeta, opts.IDSchemeType())
 }
+
+func TestNegativeQueryConversionSize(t *testing.T) {
```
arnikola: Add test to validate nil case maybe? Up to you

PR #1399, opened 2019-02-21
Optimize OStream implementation

richardartoul:
When writing annotation heavy code, we spend 80% of our time in this routine:

screen shot 2019-02-21 at 1 53 13 pm

This P.R makes that code-path 10x faster

Before
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/encoding
BenchmarkWriteBytes-8   	  100000	     22507 ns/op	      20 B/op	       0 allocs/op
PASS
ok  	github.com/m3db/m3/src/dbnode/encoding	2.495s
Success: Benchmarks passed.
```

After
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/encoding
BenchmarkWriteBytes-8   	 1000000	      2226 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/m3db/m3/src/dbnode/encoding	2.271s
```


src/dbnode/encoding/ostream.go
```diff
@@ -73,22 +87,50 @@ func (os *ostream) hasUnusedBits() bool {
...
+
+	os.pos = np
+}
+
+// ensureCapacity ensures that there is at least capacity for n more bytes.
+func (os *ostream) ensureCapacityFor(n int) {
+	var (
+		currCap      = cap(os.rawBuffer)
+		currLen      = len(os.rawBuffer)
+		availableCap = currCap - currLen
```
martin-mao: Nit: Given that availableCap is only used in the next line, may as well just do the calc of missingCap in 1 line?

src/dbnode/encoding/ostream.go
```diff
@@ -180,5 +234,21 @@ func (os *ostream) Reset(buffer checked.Bytes) {
 
 // Rawbytes returns the Osteam's raw bytes
 func (os *ostream) Rawbytes() (checked.Bytes, int) {
-	return os.rawBuffer, os.pos
+	os.repairCheckedBytes()
+	return os.checked, os.pos
+}
+
+// repairCheckedBytes makes suer that the checked.Bytes wraps the rawBuffer as
```
martin-mao: Nit: Typo on sure.

src/dbnode/encoding/ostream.go
```diff
@@ -180,5 +234,21 @@ func (os *ostream) Reset(buffer checked.Bytes) {
 
 // Rawbytes returns the Osteam's raw bytes
 func (os *ostream) Rawbytes() (checked.Bytes, int) {
-	return os.rawBuffer, os.pos
+	os.repairCheckedBytes()
+	return os.checked, os.pos
```
martin-mao: This changes the cost accounting accuracy right? eg, checked now will only give the cost when this was called and not the underlying bytes which may be written to afterwards. Not sure if it matters, how long do we hold the ref from this call for and what is the cost accounting used for?

richardartoul: As far as I'm aware we don't use checked bytes for tracking cost, we use them for tracking ownership.

This changes the semantics slightly, but the way .the OStream is used this should be fine. Basically what I'm doing is while the OStream is holding onto the checked bytes, it uses the raw buffer to write to them, and only when it exposes the bytes to someone else does it use the checked bytes.

The only thing that could cause issues is if someone got a hold of the checked bytes from this thing, and then they kept writing to the ostream and the rawbuffer and checked bytes they grabbed earlier got out of sync, but thats not how we use this structure in the encoder luckily

src/dbnode/encoding/ostream.go
```diff
@@ -31,8 +31,24 @@ const (
...
+	// the checked.Bytes for every write is massive. As a result, we store both
+	// the rawBuffer that backs the checked.Bytes AND the checked.Bytes themselves
+	// in this struct.
+	//
+	// That way, whenever we're writing to the buffer we can avoid the cost accounting
+	// overhead entirely, but when the data needs to be exposed outside of this datastructure
+	// or ownership of the data needs to be transferred, then we use the checked.Bytes, which
+	// is when the accounting really matters anyways.
+	//
+	// The rawBuffered and checked.Bytes may get out of sync as the rawBuffer is written to,
```
robskillington: nit: `s/rawBuffered/rawBuffer`

PR #1404, opened 2019-02-25
[dbnode] emit limit warning less frequently

schallert:
Even though we only emit the warnings for 5 minutes, it's frustrating to
debug startup issues when the first 30 log messages can be warnings
about process limits. This modifies those warnings to be once a minute.

### DIFF
--- a/src/dbnode/server/server.go
+++ b/src/dbnode/server/server.go
@@ -63,7 +63,7 @@ import (
 	"github.com/m3db/m3/src/dbnode/storage/series"
 	"github.com/m3db/m3/src/dbnode/topology"
 	"github.com/m3db/m3/src/dbnode/ts"
-	"github.com/m3db/m3/src/dbnode/x/tchannel"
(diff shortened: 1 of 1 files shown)


prateek:
Lol, the duration was picked to be intentionally annoying. Wondering if you're hitting a limit that isn't worth fixing?

richardartoul:
@prateek @schallert Coloring our logs would make this less annoying

prateek:
Hm would be trivial to add in colouring based on log-level so not entirely opposed. How would you two feel about control characters in logs - reads poorly when people aren't enabling colour in their readers/terminals

schallert:
Even if we added the ability to disable the warnings with a config flag I'd be happy. Like I said this is totally selfish, but I'm spinning up test clusters frequently on new cloud provider instances and it's annoying to have to tweak these values on every VM just so I can read the logs more easily.

I'm not a huge fan of coloring unless it's **only** when a TTY is attached since like @prateek mentioned control characters can be a pain, and in these cases I'm tty-less.

Time for `schallert/m3` 😂 (kidding)

PR #1407, opened 2019-02-26
Add perf section to docs

benraskin92:


docs/how_to/query.md
```diff
@@ -85,6 +85,18 @@ An example of a configuration file with the ID generation scheme can be found [h
 
 If none of these options work for you, or you would like further clarification, please stop by our [gitter channel](https://gitter.im/m3db/Lobby) and we'll be happy to help you.
 
+## Performance configurations
```
arnikola: Might want to check with the db guys, this seems like it would be better suited in a general `Performance Tuning` section

docs/performance/m3query/index.md
```diff
@@ -0,0 +1,13 @@
+## Performance configurations
+
+Below are some common configurations related to performance for the query engine.
+
+### Query conversion LRU
+
+Before fetching data from M3DB, all queries are converted to a format that M3DB can understand. We have an LRU cache that caches the results of this conversion so that queries that are the same do not need to go through the conversion process each time. By default, the size of the cache is set to `4096`, however, this number is configurable:
```
arnikola: Maybe mention what would determine a good size? i.e. this number should roughly be the number of queries on all dashboards + number of alerts?

Also that if it's set to 0 it will no longer cache (if that's desired for whatever reason)

benraskin92: 👍 good call

docs/performance/m3query/index.md
```diff
@@ -0,0 +1,18 @@
+## Performance configurations
+
+Below are some common configurations related to performance for the query engine.
+
+### Query conversion LRU
+
+Before fetching data from M3DB, all queries are converted to a format that M3DB can understand. We have an LRU cache that caches the results of this
```
arnikola: Before fetching data from M3DB, all queries are converted to a format that M3DB can understand. Considering that the majority of these queries are repeated (dashboards and alerts being fairly fixed), the converted results are cached, up to a default max of 4096. To fine tune the size of the cache, the following configuration settings:

docs/performance/m3query/index.md
```diff
@@ -0,0 +1,18 @@
...
+however, this number is configurable:
+
+```yaml
+cache:
+  queryConversion:
+    size: <int>
+```
+
+**Note:** We recommend changing the size of the cache to be roughly the number of queries being executed (including both dashboards and alerts).
+Additionally, if you set this to `0`, caching will be turned off.
```
arnikola: Setting the size to `0` will turn this cache off.

PR #1412, opened 2019-02-28
Better error responses for unknown namespaces

prateek:
Return clearer RPC errors when requests use unknown namespaces

### DIFF
--- a/src/cmd/services/m3dbnode/main/main_index_test.go
+++ b/src/cmd/services/m3dbnode/main/main_index_test.go
@@ -25,6 +25,7 @@ package main_test
 import (
 	"fmt"
 	"strconv"
+	"strings"
 	"sync"
 	"testing"
 	"text/template"
@@ -37,6 +38,7 @@ import (
 	"github.com/m3db/m3/src/dbnode/client"
 	"github.com/m3db/m3/src/dbnode/kvconfig"
 	"github.com/m3db/m3/src/dbnode/server"
+	dberrors "github.com/m3db/m3/src/dbnode/storage/errors"
 	"github.com/m3db/m3/src/dbnode/storage/index"
 	m3ninxidx "github.com/m3db/m3/src/m3ninx/idx"
 	xconfig "github.com/m3db/m3x/config"
@@ -207,6 +209,20 @@ func TestIndexEnabledServer(t *testing.T) {
 		{value: 1.0, at: start, unit: xtime.Second},
 		{value: 2.0, at: start.Add(1 * time.Second), unit: xtime.Second},
 		{value: 3.0, at: start.Add(2 * time.Second), unit: xtime.Second},
+	}
+
+	// unknown ns test
+	randomNs := "random-ns"
+	for _, v := range values {
+		err := session.WriteTagged(ident.StringID(randomNs),
+			ident.StringID("foo"),
+			ident.NewTagsIterator(ident.NewTags(
+				ident.StringTag("foo", "bar"),
+				ident.StringTag("baz", "foo"),
+			)),
+			v.at, v.value, v.unit, nil)
+		require.Error(t, err)
+		require.True(t, strings.Contains(err.Error(), dberrors.NewUnknownNamespaceError(randomNs).Error()))
 	}
 
 	for _, v := range values {
--- a/src/dbnode/network/server/tchannelthrift/node/service.go
+++ b/src/dbnode/network/server/tchannelthrift/node/service.go
@@ -34,7 +34,6 @@ import (
 	tterrors "github.com/m3db/m3/src/dbnode/network/server/tchannelthrift/errors"
 	"github.com/m3db/m3/src/dbnode/storage"
 	"github.com/m3db/m3/src/dbnode/storage/block"
-	dberrors "github.com/m3db/m3/src/dbnode/storage/errors"
 	"github.com/m3db/m3/src/dbnode/storage/index"
 	"github.com/m3db/m3/src/dbnode/ts"
 	"github.com/m3db/m3/src/dbnode/x/xio"
@@ -409,12 +408,9 @@ func (s *service) FetchTagged(tctx thrift.Context, req *rpc.FetchTaggedRequest) 
 	}
 
 	queryResult, err := s.db.QueryIDs(ctx, ns, query, opts)
-	if dberrors.IsUnknownNamespaceError(err) {
+	if err != nil {
 		s.metrics.fetchTagged.ReportError(s.nowFn().Sub(callStart))
 		return nil, convert.ToRPCError(err)
-	} else if err != nil {
-		s.metrics.fetchTagged.ReportError(s.nowFn().Sub(callStart))
-		return nil, tterrors.NewInternalError(err)
 	}
 
 	response := &rpc.FetchTaggedResult_{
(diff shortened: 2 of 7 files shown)


src/cmd/services/m3dbnode/main/main_index_test.go
```diff
@@ -209,6 +211,20 @@ func TestIndexEnabledServer(t *testing.T) {
...
+	for _, v := range values {
+		err := session.WriteTagged(ident.StringID(randomNs),
+			ident.StringID("foo"),
+			ident.NewTagsIterator(ident.NewTags(
+				ident.StringTag("foo", "bar"),
+				ident.StringTag("baz", "foo"),
+			)),
+			v.at, v.value, v.unit, nil)
+		require.Error(t, err)
+		require.True(t, strings.Contains(err.Error(), dberrors.NewUnknownNamespaceError(randomNs).Error()))
```
benraskin92: Is there a particular reason you can't just have an `assert.Equal()` here?

src/cmd/services/m3dbnode/main/main_index_test.go
```diff
@@ -209,6 +211,20 @@ func TestIndexEnabledServer(t *testing.T) {
 		{value: 3.0, at: start.Add(2 * time.Second), unit: xtime.Second},
 	}
 
+	// unknown ns test
```
benraskin92: 👍

src/dbnode/network/server/tchannelthrift/node/service_test.go
```diff
@@ -295,7 +295,63 @@ func TestServiceQueryOverloaded(t *testing.T) {
...
+		NameSpace:      nsID,
+		Limit:          &limit,
+		ResultTimeType: rpc.TimeType_UNIX_SECONDS,
+	})
+	require.Error(t, err)
+	require.Equal(t, convert.ToRPCError(unknownErr), err)
+}
+
 func TestServiceFetch(t *testing.T) {
+
```
benraskin92: Remove newline?

src/dbnode/network/server/tchannelthrift/node/service_test.go
```diff
@@ -491,6 +582,55 @@ func TestServiceFetchBatchRaw(t *testing.T) {
...
+	start := time.Now().Add(-2 * time.Hour)
+	end := start.Add(2 * time.Hour)
+
+	start, end = start.Truncate(time.Second), end.Truncate(time.Second)
+
+	nsID := "metrics"
+	unknownErr := fmt.Errorf("unknown-err")
+
+	series := map[string][]struct {
```
benraskin92: Can you just do `series := map[string]ts.Datapoint`?

src/dbnode/network/server/tchannelthrift/node/service_test.go
```diff
@@ -394,6 +450,41 @@ func TestServiceFetchIsOverloaded(t *testing.T) {
...
+
+	start, end = start.Truncate(time.Second), end.Truncate(time.Second)
+	nsID := "metrics"
+	unknownErr := fmt.Errorf("unknown-err")
+
+	mockDB.EXPECT().
+		ReadEncoded(ctx, ident.NewIDMatcher(nsID), ident.NewIDMatcher("foo"), start, end).
+		Return(nil, unknownErr)
+
+	_, err := service.Fetch(tctx, &rpc.FetchRequest{
```
benraskin92: I would expect this test to return `unknown namespace: metrics` since it's called `TestServiceFetchUnknownNs`. Same for above (`TestServiceQueryUnknownNs`).

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -408,7 +409,10 @@ func (s *service) FetchTagged(tctx thrift.Context, req *rpc.FetchTaggedRequest)
 	}
 
 	queryResult, err := s.db.QueryIDs(ctx, ns, query, opts)
-	if err != nil {
+	if dberrors.IsUnknownNamespaceError(err) {
+		s.metrics.fetchTagged.ReportError(s.nowFn().Sub(callStart))
+		return nil, convert.ToRPCError(err)
```
robskillington: Hm, wonder if we should just wrap all of our methods that have an `error` type and always just call `convert.ToRPCError(err)` on `err` if it's not nil?

Then it could be like:
```go
if err != nil {
  if _, ok := err.(*rpc.Error); !ok {
    err = convert.ToRPCError(err)
  }
}

return any, other, arg, err
```

However this would require codegen changes unless we started doing this by reflection.

robskillington: Failing that, why don't we just always `convert.ToRPCError(...)` instead of these if/else?  

It will return internal error if the error is not wrapped by an invalid params error, I don't see it necessary to do an if/else here.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -843,7 +847,9 @@ func (s *service) WriteBatchRaw(tctx thrift.Context, req *rpc.WriteBatchRawReque
 	)
 
 	batchWriter, err := s.db.BatchWriter(nsID, len(req.Elements))
-	if err != nil {
+	if dberrors.IsUnknownNamespaceError(err) {
+		return convert.ToRPCError(err)
+	} else if err != nil {
```
robskillington: Same re: if/else here.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -874,7 +880,9 @@ func (s *service) WriteBatchRaw(tctx thrift.Context, req *rpc.WriteBatchRawReque
 	}
 
 	err = s.db.WriteBatch(ctx, nsID, batchWriter.(ts.WriteBatch), pooledReq)
-	if err != nil {
+	if dberrors.IsUnknownNamespaceError(err) {
+		return convert.ToRPCError(err)
+	} else if err != nil {
```
robskillington: Same re: if/else here.

src/cmd/services/m3dbnode/main/main_index_test.go
prateek: assert v require - I use require because it fails as soon as you hit the condition (unless their are defers in the codepath, which can get complicated). Assert fails at the end of the test so its output is weirder to grok usually.

src/dbnode/network/server/tchannelthrift/node/service_test.go
prateek: sure, done

src/dbnode/network/server/tchannelthrift/node/service_test.go
prateek: yeah def could, just following the pattern in this file to keep the tests uniform.

src/dbnode/network/server/tchannelthrift/node/service_test.go
prateek: renamed the tests to be called `...UnknownErr`. Intentionally using a different type of error in these tests, this way we have fewer dependencies across packages.

src/dbnode/network/server/tchannelthrift/node/service.go
prateek: fair enough, changed to always `convert.ToRPCError`

src/dbnode/network/server/tchannelthrift/node/service.go
prateek: changed to always `convert.ToRPCError`

src/dbnode/network/server/tchannelthrift/node/service.go
prateek: changed to always `convert.ToRPCError`

src/cmd/services/m3dbnode/main/main_index_test.go
benraskin92: Yeah I meant more `Equal` vs `True`, but I think the error you get back is like some long message so ignore my comment. `strings.Contains()` makes sense here.

PR #1414, opened 2019-02-28
ignore emacs temp files

Haijuncao:


.gitignore
```diff
@@ -1,3 +1,4 @@
+*~
```
justinjc: Is this for emacs?

Haijuncao: yes

PR #1417, opened 2019-03-01
[kv] Add capability to add a generic updating watcher on a KV store

arnikola:
Allows clients to watch generic values specialized to their needs, useful for clients which are watching a single complex KV field.


src/cluster/kv/util/lock_test.go
```diff
@@ -896,6 +898,94 @@ func TestWatchAndUpdateWithValidationTime(t *testing.T) {
...
 }
 
+// NB: uses a map[string]int64 as a standin for a generic type
+func TestWatchAndUpdateWithValidationGeneric(t *testing.T) {
+	testConfig := struct {
+		sync.RWMutex
+		v map[string]int64
+	}{}
+
+	valueFn := func() map[string]int64 {
```
prateek: hm isn't this racy? -- this function returns a reference to a map which users can use without holding lock. does the race detector not pick it up?

arnikola: Yeah, weirdly race detector didn't mind even over a bunch of writes; cloned the map instead just to be safe

PR #1419, opened 2019-03-02
WIP: [m3collector] implements prometheus remote write api

eahydra:
Hi, we had ran the m3collector with m3aggregator succeed, and we modified the Prometheus code that embedded the m3collector into Prometheus which implements Prometheus Storage Interface.

Now we have separate the code from our internal Prometheus version and merge into the m3collector that implements the Prometheus Remote Write API. And we want to feedback to M3DB

### DIFF
--- a/src/collector/api/v1/handler/prometheus/remote/write.go
+++ b/src/collector/api/v1/handler/prometheus/remote/write.go
@@ -1,0 +1,113 @@
+package remote
+
+import (
+	"fmt"
+	"io/ioutil"
+	"math"
+	"net/http"
+
+	"github.com/m3db/m3/src/collector/reporter"
+	"github.com/m3db/m3/src/metrics/metric/id"
+	"github.com/m3db/m3/src/query/models"
+	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/x/serialize"
+	"github.com/m3db/m3x/instrument"
+
+	"github.com/gogo/protobuf/proto"
+	"github.com/golang/snappy"
+	"github.com/prometheus/prometheus/prompb"
+)
+
+const (
+	WriteURL         = "/api/v1/write"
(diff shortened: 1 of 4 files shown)


robskillington:
@eahydra thank you for this!

src/collector/api/v1/handler/prometheus/remote/write.go
```diff
@@ -0,0 +1,113 @@
...
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+		}
+	}
+
+	w.WriteHeader(http.StatusOK)
+}
+
+func (h *PromWriteHandler) newMetricID(l []*prompb.Label) (id.ID, error) {
```
robskillington: We should probably move this logic to a central place so we don't duplicate this with the existing remote write logic. 

Perhaps we can put it in the `src/cmd/coordinator/downsample` package and export it?

src/collector/server/id_pool_types.go
```diff
@@ -0,0 +1,342 @@
...
+	"github.com/m3db/m3/src/x/serialize"
+	"github.com/m3db/m3x/checked"
+	"github.com/m3db/m3x/ident"
+	"github.com/m3db/m3x/pool"
+
+	"github.com/prometheus/common/model"
+)
+
+var (
+	defaultMetricNameTagName = []byte(model.MetricNameLabel)
```
robskillington: Similar to above in this file, we should probably export logic to a central place so we don't duplicate this with the existing remote write logic. 

Perhaps we can export these types from `src/cmd/coordinator/downsample`?

src/collector/api/v1/handler/prometheus/remote/write.go
eahydra: Yeah, I will refer to your comments to make changes.

PR #1420, opened 2019-03-02
construct service once and have it shared by tchannel and http servers

Haijuncao:


Haijuncao:
> Mostly LGTM, just two comments:
> 
> Hard to tell from C.I but you have these minor lint issues which is why its failing:
> 
> ```shell
> src/dbnode/network/server/httpjson/node/server.go:38:2:warning: unused struct field github.com/m3db/m3/src/dbnode/network/server/httpjson/node.server.ttopts (structcheck)
> --
>   | src/dbnode/network/server/httpjson/node/server.go:38:2:warning: field ttopts is unused (U1000) (unused)
> ```
> 
> Also, now that we've fixed this, could you bump the size of the the constant `writeBatchPooledReqPoolSize` from `1024` to `4096`? The reason I discovered this issue is I wanted to make that constant bigger and then I noticed we were allocating the pools twice.
> 
> Ideally the pool size would be controlled by config, but changing the constant is good enough for now.
I will make another diff for change default pool size. Maybe also make the pool size configurable.

PR #1421, opened 2019-03-02
Use Seek() and Read() APIs for index/data files instead of mmap + share expensive seeker resources among seekers

richardartoul:
Had to create a new P.R for C.I reasons, see previous comments / discussion on #1299


src/dbnode/persist/fs/msgpack/decoder.go
```diff
@@ -41,13 +43,17 @@ var (
 	emptyLogEntry               schema.LogEntry
 	emptyLogMetadata            schema.LogMetadata
 	emptyLogEntryRemainingToken DecodeLogEntryRemainingToken
-)
 
-var errorUnableToDetermineNumFieldsToSkip = errors.New("unable to determine num fields to skip")
+	errorUnableToDetermineNumFieldsToSkip          = errors.New("unable to determine num fields to skip")
+	errorCalledDecodeBytesWithoutByteStreamDecoder = errors.New("called decodeBytes with out byte stream decoder")
```
robskillington: nit: rename this to `errThing... =` rather than `errorThing... = ` for consistency with other packages/etc.

src/dbnode/persist/fs/seek.go
```diff
@@ -383,55 +369,106 @@ func (s *seeker) SeekByIndexEntry(entry IndexEntry) (checked.Bytes, error) {
...
+	resources.xmsgpackDecoder.Reset(resources.fileDecoderStream)
 
 	idBytes := id.Bytes()
-	// Prevent panic's when we're scanning to the end of the buffer
-	for stream.Remaining() != 0 {
-		entry, err := s.decoder.DecodeIndexEntry()
-		// Should never happen, either something is really wrong with the code or
-		// the file on disk was corrupted
+	for {
+		currOffset, err := s.indexFd.Seek(0, os.SEEK_CUR)
```
robskillington: Is this call necessary?  Seems this is not needed because `Seek(0, SEEK_CUR)` should cause no change as `SEEK_CUR` means "seek relative to the current offset",

robskillington: Can't we just keep track of `currOffset` ourselves and avoid this syscall?

robskillington: I have looked at the different interfaces, looks difficult to actually call like `.xmsgpackDecoder.Position()` or whatever which can call the reader underneath.

Maybe just add a comment here that in the future we can remove this Seek (and consequently syscall) call by tracking the position ourselves with some refactoring of interfaces/etc?

PR #1422, opened 2019-03-03
[m3nsch] script to more easily run on k8s

schallert:
Previously users had to `exec` into a pod and manually provide m3nsch
endpoints. This is unnecessary as we can gather the IPs of the agent
pods and pass that via the command line.

Also included a make target to generate the simple manifest bundle in CI
to avoid it becoming stale.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -77,6 +77,7 @@ SUBDIRS :=    \
 	m3ninx      \
 	aggregator  \
 	ctl         \
+	kube        \
 
 TOOLS :=               \
 	read_ids             \
@@ -247,6 +248,17 @@ test-ci-integration:
 
 define SUBDIR_RULES
 
+# We override the rules for `*-gen-kube` to just generate the kube manifest
+# bundle.
+ifeq ($(SUBDIR), kube)
+
+# Builds the single kube bundle from individual manifest files.
+all-gen-kube:
+	@echo "--- Generating kube bundle"
+	@./kube/scripts/build_bundle.sh
+
+else
+
 .PHONY: mock-gen-$(SUBDIR)
 mock-gen-$(SUBDIR): install-tools
 	@echo "--- Generating mocks $(SUBDIR)"
@@ -341,6 +353,8 @@ metalint-$(SUBDIR): install-gometalinter install-linter-badtime install-linter-i
 	@echo "--- metalinting $(SUBDIR)"
 	@(PATH=$(retool_bin_path):$(PATH) $(metalint_check) \
 		$(metalint_config) $(metalint_exclude) src/$(SUBDIR))
+
+endif
 
 endef
 
--- a/docs/how_to/single_node.md
+++ b/docs/how_to/single_node.md
@@ -14,8 +14,8 @@ To begin, first start up a Docker container with port `7201` (used to manage the
 directory on your host for durability:
 
 ```
-docker pull quay.io/m3/m3dbnode:latest
-docker run -p 7201:7201 -p 7203:7203 -p 9003:9003 --name m3db -v $(pwd)/m3db_data:/var/lib/m3db -v <PATH_TO_M3DB_CONFIG.yml>:/etc/m3dbnode/m3dbnode.yml quay.io/m3/m3dbnode:latest
+docker pull quay.io/m3db/m3dbnode:latest
+docker run -p 7201:7201 -p 7203:7203 -p 9003:9003 --name m3db -v $(pwd)/m3db_data:/var/lib/m3db -v <PATH_TO_M3DB_CONFIG.yml>:/etc/m3dbnode/m3dbnode.yml quay.io/m3db/m3dbnode:latest
 ```
 
 **Note:** For the single node case, we recommend that you start with this [sample config file](https://github.com/m3db/m3/blob/master/src/dbnode/config/m3dbnode-local-etcd.yml). If you inspect the file, you'll see that all the configuration is namespaced by `coordinator` or `db`. That's because this setup runs `M3DB` and `M3Coordinator` as one application. While this is convenient for testing and development, you'll want to run clustered `M3DB` with a separate `M3Coordinator` in production. You can read more about that [here.](cluster_hard_way.md).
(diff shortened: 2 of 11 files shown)


docs/how_to/single_node.md
```diff
@@ -14,8 +14,8 @@ To begin, first start up a Docker container with port `7201` (used to manage the
 directory on your host for durability:
 
 ```
-docker pull quay.io/m3/m3dbnode:latest
-docker run -p 7201:7201 -p 7203:7203 -p 9003:9003 --name m3db -v $(pwd)/m3db_data:/var/lib/m3db -v <PATH_TO_M3DB_CONFIG.yml>:/etc/m3dbnode/m3dbnode.yml quay.io/m3/m3dbnode:latest
+docker pull quay.io/m3db/m3dbnode:latest
```
robskillington: Eh?  Why are we changing the Quay repo people are using?

Have all the changes you been making for the :latest tag, etc been to the other Quay repo?

kube/README.md
```diff
@@ -46,10 +44,12 @@ $ kubectlc exec -it m3nsch-client-78987775f5-l5pvs
...
-You can view the stats of the ongoing benchmark via `m3nsch_server`'s prometheus endpoints:
+You can view the stats of the ongoing benchmark via `m3nsch_server`'s prometheus endpoints, or using our [grafana dashboard][dash]:
 ```
 $ kubectl port-forward svc/m3nch-agent 12580
 $ curl -sSf localhost:12580/metrics | rg write.+success
 # TYPE write_req_success counter
 write_req_success 146000
 ```
+
+[dash]: https://github.com/m3db/m3/blob/75661bc2e7fc806ee3424c068db6812455ba0878/integrations/grafana/m3nsch_dashboard.json
```
robskillington: Shouldn't this reference the master blob?  Otherwise they'll get an old dashboard if we ever update it no?

docs/how_to/single_node.md
schallert: It looks like we disabled builds for the other one, or that one only publishes tags and not latest `master`. I thought we wanted to deprecate it or something and only have our builds published to the same repo as our GH name, do we have a plan for that?

kube/README.md
schallert: I'm fine with making it master but I mostly wanted to future-proof this so the link still works if we ever move the dashboards around

kube/scripts/m3nsch_client.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# The script makes it easier to run m3nsch by gathering the endpoints of all
+# agents and constructing the base m3nsch_client command. You may provide extra
+# labels to target agent pods if you are running multiple m3nsch agent
+# deployments in the same namespace (i.e. for concurrent load tests against
+# multiple m3db clusters).
+#
+# Example usage:
```
prateek: very nifty

src/cmd/services/m3nsch_server/main/main.go
```diff
@@ -97,9 +97,7 @@ func newSessionFn(conf config.Configuration, iopts instrument.Options) m3nsch.Ne
 		if zone != svc.Zone {
 			return nil, fmt.Errorf("request zone %s did not match config zone %s", zone, svc.Zone)
 		}
-		if env != svc.Env {
-			return nil, fmt.Errorf("request env %s did not match config zone %s", env, svc.Env)
-		}
+		svc.Env = env
```
prateek: Don't have context for this change - why override?

kube/m3dbnode-statefulset.yaml
```diff
@@ -82,7 +82,7 @@ spec:
             weight: 10
       containers:
       - name: m3db
-        image: quay.io/m3/m3dbnode:latest
+        image: quay.io/m3db/m3dbnode:latest
```
prateek: could we use the `m3` repo for quay?

Makefile
```diff
@@ -247,6 +248,17 @@ test-ci-integration:
...
 define SUBDIR_RULES
 
+# We override the rules for `*-gen-kube` to just generate the kube manifest
+# bundle.
+ifeq ($(SUBDIR), kube)
+
+# Builds the single kube bundle from individual manifest files.
+all-gen-kube:
+	@echo "--- Generating kube bundle"
+	@./kube/scripts/build_bundle.sh
```
prateek: hm is there a way to verify if the generated bundle file is valid?

src/cmd/services/m3nsch_server/main/main.go
schallert: Ah I should've left a comment. AFAICT this was causing me to have to have a separate agent config for each environment I wanted to load test, since if my `env` flag didn't match the YAML it bailed.

What I wanted was to be able to have a bunch of agent deployments share one config with the etcd cluster and override env per-tier. Or be able to start a benchmark with a set of agents, stop them, and then resume them against a different env.

schallert: Edit: for some reason I didn't relax this check for zone, I forget but can change that if we want the behavior more aligned

kube/m3dbnode-statefulset.yaml
schallert: Ah yeah I'll get #1364 landed first and then land this. For a while `/m3` was stale

Makefile
schallert: Hmm, do you mean like valid Kube objects or just not garbage YAML? I'm not sure of a way to do k8s validation without a remote API server and `kubectl` install.

src/cmd/services/m3nsch_server/main/main.go
prateek: Makes sense. Maybe cleanup zone as you suggested too

Makefile
prateek: ideally kube level, for e.g. dunno how good it is but something like https://github.com/garethr/kubeval maybe might be nice?

Feel free to do as a followup, don't want to derail this PR

kube/m3dbnode-statefulset.yaml
prateek: 👍

Makefile
schallert: Ah nice, I like that idea. Will follow up.

PR #1426, opened 2019-03-04
Add support for compressing protobuf messages

richardartoul:
# Review Guide

## Known Issues

There are several known issues with this P.R that I will address in a follow up, but the P.R is already too large to address them all here: https://github.com/m3db/m3/issues/1471

## Features

- [x] Add new Proto encoder which implements encoding.Encoder and can compress a stream of Protobuf messages given a schema.
- [x] Add new Proto iterator which implements encoding.Iterator and can decompress a stream of Protobuf messages given a schema.
- [x] Add extremely rudimentary support for applying a single schema to an entire M3DB node based on a configuration file (this is temporary code that will be replaced by dynamic schema management later).
- [x] Add property test that verifies the behavior of the encoder and iterator by generating dynamic schemas and encoding/decoding many different messages.


src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -363,31 +363,34 @@ func (enc *encoder) writeFullFloatVal(val uint64) {
...
 func (enc *encoder) writeFloatXOR(val uint64) {
 	xor := enc.vb ^ val
-	enc.writeXOR(enc.xor, xor)
+	WriteXOR(enc.os, enc.xor, xor)
 	enc.xor = xor
 	enc.vb = val
 }
 
-func (enc *encoder) writeXOR(prevXOR, curXOR uint64) {
+func WriteXOR(
```
robskillington: This is fine for a proof of concept, I do think we'll want to use use the same float or int value compression logic though before we land on a final implementation.

richardartoul: I don't think so for two reasons:

1. We don't want this to be lossy because its expressly not for metrics data
2. The data is already typed (since we have schemas) so we already know if the data is a float or integer. If we want to do some special integer compression we can just do it on any fields that are explicitly of type integer

src/dbnode/encoding/proto/iterator.go
```diff
@@ -0,0 +1,878 @@
...
+)
+
+type iterator struct {
+	opts                   encoding.Options
+	err                    error
+	schema                 *desc.MessageDescriptor
+	stream                 encoding.IStream
+	consumedFirstMessage   bool
+	lastIterated           *dynamic.Message
+	lastIteratedAnnotation []byte
```
martin-mao: Why name this Annotation, isn't it just the proto bytes now and we're just using annotation due to interface refactoring issues?

src/dbnode/encoding/proto/iterator.go
```diff
@@ -0,0 +1,878 @@
...
+		}
+
+		prev := int64(it.customFields[i].prevFloatBits)
+		it.customFields[i].prevFloatBits = uint64(prev + (sign * diff))
+	}
+
+	return nil
+}
+
+// TODO: Share logic with encoder if possible
```
martin-mao: +1, logic looks identical.

src/dbnode/encoding/proto/iterator.go
```diff
@@ -0,0 +1,878 @@
...
+func (it *iterator) readFloatXOR(i int) (floatBits, xor uint64, err error) {
+	xor, err = it.readXOR(i)
+	if err != nil {
+		return 0, 0, err
+	}
+	prevFloatBits := it.customFields[i].prevFloatBits
+	return prevFloatBits ^ xor, xor, nil
+}
+
+func (it *iterator) readXOR(i int) (uint64, error) {
```
martin-mao: We should probably refactor the horizontal field encoding logic into some lower level encoding packages so you can re-use certain type encoding logic (in addition to the delta of delta timestamp stuff) across the m3tsz and proto encoders and iterators. 

I guess we can do that when we remove the m3tsz iterator out of here.

src/dbnode/encoding/m3tsz/m3tsz.go
```diff
@@ -23,20 +23,36 @@ package m3tsz
...
 
 const (
 	// DefaultIntOptimizationEnabled is the default switch for m3tsz int optimization
 	DefaultIntOptimizationEnabled = true
 
-	opcodeZeroValueXOR        = 0x0
-	opcodeContainedValueXOR   = 0x2
-	opcodeUncontainedValueXOR = 0x3
+	// OpcodeZeroValueXOR indicates the XOR was zero.
+	OpcodeZeroValueXOR = 0x0
```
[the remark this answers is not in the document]
richardartoul: I had them in encoding.Common at first, but once I embedded the M3TSZ encoder in this one it made more sense. Even when I factor that out, I think it makes sense to keep these in the M3TSZ packaged unless we're gonna get rid of that completely

src/dbnode/encoding/proto/encoder_test.go
```diff
@@ -0,0 +1,50 @@
...
+				{fieldNum: 4, fieldType: cBytes},       // deliveryID
+			},
+		},
+	}
+
+	for _, tc := range testCases {
+		tszFields := customFields(nil, tc.schema)
+		require.Equal(t, tc.expected, tszFields)
+	}
+}
```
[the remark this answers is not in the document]
richardartoul: Honestly the property test covers pretty much everything. It generates hundred of different combinations of schemas

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,807 @@
...
+
+		currVal := existing[j]
+		nextVal := existing[nextIdx]
+		existing[j] = nextVal
+		existing[nextIdx] = currVal
+	}
+}
+
+func (enc *Encoder) addToBytesDict(fieldIdx int, hash uint64) {
+	existing := enc.customFields[fieldIdx].bytesFieldDict
```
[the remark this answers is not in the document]
richardartoul: You can't implement an LRU with a heap can you? Would need a map + double linked list

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,807 @@
...
+func (enc *Encoder) Reset(
+	start time.Time,
+	capacity int,
+) {
+	enc.reset(start, capacity)
+}
+
+// SetSchema sets the encoders schema.
+func (enc *Encoder) SetSchema(schema *desc.MessageDescriptor) {
+	enc.resetSchema(schema)
```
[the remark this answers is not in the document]
richardartoul: The answer to 1 and 2 is yes. The code to do that will come in a follow up P.R probably, just want to get this stuff reviewed / merged

I have it now just to have the interface for what it will be in the future and also it lets me test reusing the encoder in my property test so I know the reset logic works

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,807 @@
...
+	case int32:
+		signedVal = int64(typedVal)
+	default:
+		return fmt.Errorf(
+			"%s found unknown type in fieldNum %d", encErrPrefix, customField.fieldNum)
+	}
+
+	if isUnsignedInt(customField.fieldType) {
+		if !enc.hasEncodedFirstSetOfCustomValues {
+			enc.encodeFirstUnsignedIntValue(i, unsignedVal)
```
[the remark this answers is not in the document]
richardartoul: I tried that a few different times and the logic was just different enough to make it a mess

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -201,16 +204,18 @@ func (enc *encoder) writeTimeUnit(tu xtime.Unit) bool {
...
 	enc.os.WriteBits(uint64(nt), 64)
-	return enc.writeNextTime(t, ant, tu)
+	enc.writeAnnotation(ant)
+	return enc.WriteNextTime(t, ant, tu)
 }
 
-func (enc *encoder) writeNextTime(t time.Time, ant ts.Annotation, tu xtime.Unit) error {
-	enc.writeAnnotation(ant)
+// WriteFirstTime encodes the next (non-first) timestamp.
+func (enc *Encoder) WriteNextTime(t time.Time, ant ts.Annotation, tu xtime.Unit) error {
```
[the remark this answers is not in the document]
richardartoul: The comment is wrong, I'll fix that. Not sure what you mean by only encoding the annotation once?

src/dbnode/encoding/proto/iterator.go
richardartoul: sure I can rename this

src/dbnode/encoding/proto/iterator.go
richardartoul: Yeah right now I think it makes sense for proto to depend on m3tsz since thats pretty much literally what it is logically. Can refactor if that turns out to be the wrong decision

src/dbnode/encoding/proto/encoder_test.go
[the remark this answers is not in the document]
martin-mao: I'm thinking the edge cases like making sure the errors are propagated when they occur or we handle things like int overflows (or like the delta overflowing etc).

glide.yaml
```diff
@@ -114,9 +118,8 @@ import:
     version: ^1.2.0
 
   # NB(r): make sure to use the master commit for vellum
-  # once all upstream changes are complete in github.com/couchbase/vellum.
-  - package: github.com/couchbase/vellum
-    repo: https://github.com/m3db/vellum
+  # once all upstream changes are complete in github.com/m3db/vellum.
+  - package: github.com/m3db/vellum
```
robskillington: Do we have to update all the import paths now?

scripts/development/m3_stack/m3coordinator.yml
```diff
@@ -32,6 +32,8 @@ clusters:
             - zone: embedded
               endpoints:
                 - m3db_seed:2379
+      # proto:
+      #   schemaFilePath: /etc/m3coordinator/schema.proto
```
robskillington: Do you still need this?

scripts/development/m3_stack/schema.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+message Schema {
+  double latitude = 1;
+  double longitude = 2;
+  int64 numTrips = 3;
```
robskillington: nit: Maybe more realistic field rather than `numTrips`?

src/dbnode/client/options.go
```diff
@@ -354,7 +357,15 @@ func (o *options) Validate() error {
...
 	opts.readerIteratorAllocate = func(r io.Reader) encoding.ReaderIterator {
-		return m3tsz.NewReaderIterator(r, m3tsz.DefaultIntOptimizationEnabled, encoding.NewOptions())
+		return m3tsz.NewReaderIterator(r, nil, m3tsz.DefaultIntOptimizationEnabled, encoding.NewOptions())
+	}
+	return &opts
+}
+
+func (o *options) SetEncodingProto(schema *desc.MessageDescriptor, encodingOpts encoding.Options) Options {
+	opts := *o
+	opts.readerIteratorAllocate = func(r io.Reader) encoding.ReaderIterator {
```
robskillington: Interesting, for now this makes sense, I assume in future we're going to have to have some sort of per namespace data structure?

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -148,14 +149,15 @@ func (enc *encoder) writeNext(dp ts.Datapoint, ant ts.Annotation, tu xtime.Unit)
...
-func (enc *encoder) shouldWriteAnnotation(ant ts.Annotation) bool {
+func (enc *Encoder) shouldWriteAnnotation(ant ts.Annotation) bool {
 	numAnnotationBytes := len(ant)
 	if numAnnotationBytes == 0 {
 		return false
 	}
 	if numAnnotationBytes != len(enc.ant) {
 		return true
 	}
+	// TODO(rartoul): Use bytes.Equal()
```
robskillington: Any reason not to do this now?

src/dbnode/encoding/m3tsz/encoder.go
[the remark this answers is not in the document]
robskillington: Hm yeah, is there a reason that `enc.writeAnnotation(ant)` is not called in `WriteNextTime` anymore?

src/dbnode/encoding/m3tsz/m3tsz.go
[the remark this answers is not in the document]
robskillington: Yeah we can unexport these now yeah?

src/dbnode/encoding/m3tsz/m3tsz.go
```diff
@@ -61,6 +77,95 @@ var (
...
+	stream.WriteBits(uint64(curLeading), 6)
+	numMeaningfulBits := 64 - curLeading - curTrailing
+	// numMeaningfulBits is at least 1, so we can subtract 1 from it and encode it in 6 bits
+	stream.WriteBits(uint64(numMeaningfulBits-1), 6)
+	stream.WriteBits(curXOR>>uint(curTrailing), numMeaningfulBits)
+}
+
+// WriteIntSig writes the number of significant bits of the diff if it has changed and
+// updates the IntSigBitsTracker.
+func WriteIntSig(os encoding.OStream, sigTracker *IntSigBitsTracker, sig uint8) {
```
robskillington: Why not make this a method on `IntSigBitsTracker`?  You could also set `encoding.OStream` on that too

robskillington: In fact what about creating a new struct called `OutputStream` in the M3TSZ package, then doing this:

```go


// now on Encoder you can embed it
type Encoder struct {
  stream OutputStream
}

func (e *Encoder) writeWhatever() {
  e.stream.os.Write(...)
}

func (e *Encoder) writeSomethingWithXOR() {
  e.stream.WriteXOR(prevXOR, curXOR)
}

func (e *Encoder) writeSomethingWithIntSig() {
  e.stream.WriteIntSig(sig)
}
```

src/dbnode/encoding/proto/common.go
```diff
@@ -0,0 +1,247 @@
...
+	// is no reasonm to distinguish between them for the purposes of encoding
+	// and decoding.
+	cNotCustomEncoded customFieldType = iota
+	cSignedInt64
+	cSignedInt32
+	cUnsignedInt64
+	cUnsignedInt32
+	cFloat64
+	cFloat32
+	cBytes
```
robskillington: Any reason for hungarian notation? heh. (Side note https://stackoverflow.com/questions/111933/why-shouldnt-i-use-hungarian-notation)

robskillington: Could you suffix with the word "Field" to avoid collision with say keyword like "float64"?

i.e. 

```go
type customFieldType int

const (
	notCustomField customFieldType = iota
	signedInt64Field
	signedInt32Field
	unsignedInt64Field
	unsignedInt32Field
	float64Field
	float32Field
	bytesField
)
```

src/dbnode/encoding/proto/common.go
```diff
@@ -0,0 +1,240 @@
...
+type customFieldState struct {
+	// TODO(rartoul): This could be made more efficient by separating out fields required
+	// for encoding and those required for enumeration, as well as trying to reuse some
+	// of the fields for multiple types to save memory, but its ok for now.
+	fieldNum  int
+	fieldType customFieldType
+
+	// Float state.
+	prevXOR       uint64
+	prevFloatBits uint64
```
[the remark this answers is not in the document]
robskillington: I believe that actually int's aren't being coerced into floats for the proto encoding.  I.e. it's not doing true M3TSZ since it needs to be lossless, so only doing float XORing for floats, and then for ints they are the only path using the sigint stuff, etc.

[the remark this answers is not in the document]
robskillington: In future we could introduce the lossy support too with int to float coercion (like M3TSZ is doing).

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+
+func (enc *Encoder) encodeTimestamp(t time.Time, tu xtime.Unit) error {
+	if !enc.hasEncodedFirstSetOfCustomValues {
+		return enc.m3tszEncoder.WriteFirstTime(t, nil, tu)
+	}
+	return enc.m3tszEncoder.WriteNextTime(t, nil, tu)
+}
+
+// TODO: Add concept of hard/soft error and if there is a hard error
+// then the encoder cant be used anymore.
```
robskillington: Yeah this is important.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+
+// SetSchema sets the encoders schema.
+func (enc *Encoder) SetSchema(schema *desc.MessageDescriptor) {
+	enc.resetSchema(schema)
+}
+
+func (enc *Encoder) reset(start time.Time, capacity int) {
+	// TODO: Probably don't want to make these both the same capacity.
+	// TODO: Won't this be a massive alloc on the m3tsz encoder side for no reason?
+	// probably need to be able to reset the encoder and also specify the OStream
```
robskillington: Aren't you sharing the same ostream?  This is strange.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+			// in the proto marshal.
+			m.ClearFieldByNumber(customField.fieldNum)
+		}
+	}
+	enc.hasEncodedFirstSetOfCustomValues = true
+
+	return nil
+}
+
+func (enc *Encoder) encodeTSZValue(i int, iVal interface{}) error {
```
robskillington: Damn, this will alloc a float64 I assume?

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+		currString, ok := iVal.(string)
+		if !ok {
+			return fmt.Errorf(
+				"%s found unknown type in fieldNum %d", encErrPrefix, customField.fieldNum)
+		}
+		currBytes = []byte(currString)
+	}
+
+	var (
+		hash             = murmur3.Sum64(currBytes)
```
robskillington: If you're doing string hashing (and not needing a hash for good distribution), xxhash is way faster and why we use it for our custom map types.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+		numPreviousBytes = len(customField.bytesFieldDict)
+		lastHashIdx      = numPreviousBytes - 1
+	)
+	if numPreviousBytes > 0 && hash == customField.bytesFieldDict[lastHashIdx] {
```
robskillington: Still need to actually check the value matches (if the hash matches, it could be a collision) - hash collisions are unavoidable unless using perfect hashing (of which each value must be known).

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+		// No changes control bit.
+		enc.stream.WriteBit(opCodeNoChange)
+		return nil
+	}
+
+	// Bytes changed control bit.
+	enc.stream.WriteBit(opCodeChange)
+
+	for j, prevHash := range customField.bytesFieldDict {
+		if hash == prevHash {
```
robskillington: Here too, need to actually check values match with bytes.Equal or something (subslice as we talked about back into where it was stored).

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+			return nil
+		}
+	}
+
+	// Control bit means interpret subsequent bits as varInt encoding length of a new
+	// []byte we haven't seen before.
+	enc.stream.WriteBit(opCodeInterpretSubsequentBitsAsBytesLengthVarInt)
+	enc.encodeVarInt((uint64(len(currBytes))))
+	for _, b := range currBytes {
+		enc.stream.WriteByte(b)
```
robskillington: Isn't there a `stream.Write([]byte)` now?

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+					enc.opts.ByteFieldDictionaryLRUSize()))
+			enc.moveToEndOfBytesDict(i, j)
+			return nil
+		}
+	}
+
+	// Control bit means interpret subsequent bits as varInt encoding length of a new
+	// []byte we haven't seen before.
+	enc.stream.WriteBit(opCodeInterpretSubsequentBitsAsBytesLengthVarInt)
+	enc.encodeVarInt((uint64(len(currBytes))))
```
robskillington: nit: Extra level of `(` and `)` used here.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+		// their default values so we can do a clean merge on read.
+		enc.stream.WriteBit(opCodeNoFieldsSetToDefaultProtoMarshal)
+	}
+	enc.encodeVarInt(uint64(len(marshaled)))
+	enc.stream.WriteBytes(marshaled)
+
+	if enc.lastEncoded == nil {
+		// Set lastEncoded to m so that subsequent encodings only need to encode fields
+		// that have changed.
+		enc.lastEncoded = dynamic.NewMessage(enc.schema)
```
robskillington: There's no `Reset()` like method?  Sucks that this has to be allocated again and again.

Can this be fixed upstream?

robskillington: Could it not just be `enc.lastEncoded = m`?

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+}
+
+func (enc *Encoder) encodeFirstTSZValue(i int, v float64) {
+	fb := math.Float64bits(v)
+	enc.stream.WriteBits(fb, 64)
+	enc.customFields[i].prevFloatBits = fb
+	enc.customFields[i].prevXOR = fb
+}
+
+func (enc *Encoder) encodeNextTSZValue(i int, next float64) {
```
robskillington: Hm, I think `WriteXOR` itself should handle this first/next business, and be on something like `EncoderStream` as mentioned before and keep all these prevXOR and prevFloatBits fields to itself too.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,820 @@
...
+	}
+
+	vBits := uint64(v)
+	numSig := encoding.NumSig(vBits)
+
+	m3tsz.WriteIntSig(enc.stream, &enc.customFields[i].intSigBitsTracker, numSig)
+	enc.encodeIntValDiff(vBits, neg, numSig)
+}
+
+func (enc *Encoder) encodeFirstUnsignedIntValue(i int, v uint64) {
```
robskillington: Same comment for firstsignedintvalue and firstunsignedintvalue and the consequent "next" versions too.

glide.yaml
richardartoul: Yeah, I had to do this to be able to pull it into our internal monorepo which uses dep :/

src/dbnode/client/options.go
richardartoul: Yep hopefully @Haijuncao 's code will fix this

src/dbnode/encoding/m3tsz/m3tsz.go
richardartoul: sure thats cleaner

src/dbnode/encoding/proto/common.go
richardartoul: Does it count as hungarian notation when I'm literally defining new types? I can change this if you feel strongly, but appending `Field` feels just as gross as prepending `c`

src/dbnode/encoding/proto/encoder.go
richardartoul: Yeah this might just be stale, will look into it.

src/dbnode/encoding/proto/encoder.go
richardartoul: its already allocated because I got it as an `interface{}` from the dynamic protobuf library. So luckily (or unluckily depending on how you think about it) this shouldn't cost me anything

src/dbnode/encoding/proto/encoder.go
richardartoul: sure thing, good to know. Will change

src/dbnode/encoding/proto/encoder.go
richardartoul: same reply

src/dbnode/encoding/proto/encoder.go
richardartoul: I'll have to pass in some state . to distinguish between first and last because that state is in my `customFieldState` the WriteXOR method doesn't know anything about that. I could add it to the `EncoderStream` but then I'll have to have one per-TSZ field which I think is probably more unwieldy than the existing implementation but I'll look into it

src/dbnode/encoding/m3tsz/encoder.go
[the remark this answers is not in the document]
richardartoul: I think I made this change as some other refactoring I did with an earlier implementation. The way I had it was still correct since I added a call to `writeAnnotation()` in `writeNext()` before the call to `writeNextTimestamp()` but agree it was confusing. Set it back to behave how it used to

src/dbnode/encoding/m3tsz/iterator.go
```diff
@@ -64,9 +64,12 @@ type readerIterator struct {
 }
 
 // NewReaderIterator returns a new iterator for a given reader
-func NewReaderIterator(reader io.Reader, intOptimized bool, opts encoding.Options) encoding.ReaderIterator {
-	return &readerIterator{
-		is:           encoding.NewIStream(reader),
+func NewReaderIterator(reader io.Reader, is encoding.IStream, intOptimized bool, opts encoding.Options) encoding.ReaderIterator {
```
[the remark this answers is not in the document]
richardartoul: sure

src/dbnode/encoding/m3tsz/m3tsz.go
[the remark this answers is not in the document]
richardartoul: I was actually using them in the proto iterator but it was easy to refactor it to just call a method on the m3tsz iterator so yeah I can unexport these now

src/dbnode/encoding/proto/common.go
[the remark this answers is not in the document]
richardartoul: yep its just being used for floats now

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,807 @@
...
+	// avoid allocations.
+	varIntBuf              [8]byte
+	changedValues          []int32
+	fieldsChangedToDefault []int32
+	unmarshaled            *dynamic.Message
+
+	hasEncodedFirstSetOfCustomValues bool
+	closed                           bool
+
+	m3tszEncoder *m3tsz.Encoder
```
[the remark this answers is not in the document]
richardartoul: Sure

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,807 @@
...
+
+	// Field numbers are 1-indexed so encoding the maximum field number
+	// at the beginning is equivalent to encoding the number of types
+	// we need to read after if we imagine that we're encoding a 1-indexed
+	// bitset where the position in the bitset encodes the field number (I.E
+	// the first value is the type for field number 1) and the values are
+	// the number of bits required to unique identify a custom type instead of
+	// just being a single bit (3 bits in the case of version 1 of the encoding
+	// scheme.)
+	maxFieldNum := enc.customFields[len(enc.customFields)-1].fieldNum
```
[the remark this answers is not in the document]
richardartoul: correct although I can add a sort just to be extra safe

src/dbnode/encoding/proto/encoder.go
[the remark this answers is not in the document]
richardartoul: added a TODO

src/dbnode/encoding/proto/encoder.go
richardartoul: Yeah ok its a little gross, but actually all I need to do is call reset on the m3tsz encoder because that will reset the shared OStream. I'll add a comment but this might get better when I implemented your `EncoderStream` suggestion

src/dbnode/encoding/proto/encoder.go
richardartoul: Implemented the bit padding thing we discussed

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,807 @@
...
+		enc.stream.WriteBit(opCodeNoChange)
+		return
+	}
+
+	enc.stream.WriteBit(opCodeChange)
+
+	numSig := encoding.NumSig(diff)
+	newSig := enc.customFields[i].intSigBitsTracker.TrackNewSig(numSig)
+	// Can't use an intermediary variable for the intSigBitsTracker because we need to
+	// modify the value in the slice, not a copy of it.
```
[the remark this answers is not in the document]
richardartoul: its really easy to mess it up because you have to pass different types, would rather leave as is so the type-system can help out

src/dbnode/encoding/proto/encoder.go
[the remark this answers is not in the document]
richardartoul: Current algo shouldn't have repeated hashes unless there is a hash collision.

Doubly linked list would work, gonna leave it out for now though since these shouldn't be too big. Can add as a follow-up

src/dbnode/encoding/proto/encoder_test.go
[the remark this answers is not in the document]
richardartoul: int overflows and all that will be covered by the property test. Timestamp delta of delta stuff should be covered by M3TSZ tests.

I know what you're saying, but I can't think of much that isn't covered by my property tests already and I've already spent days writing that, don't have time to go add a bunch of manual tests as well

src/dbnode/encoding/proto/iterator.go
richardartoul: yeah the problem is they operate on different types so I would have to do some `interface{}` magic which is gonna allocate :/

src/dbnode/encoding/proto/encoder_test.go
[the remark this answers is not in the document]
martin-mao: Ok sounds good.

src/dbnode/encoding/proto/encoder.go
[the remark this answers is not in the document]
martin-mao: Yeah I think this should be efficient for smaller numbers - just suggesting having a bench to see where the breaking point is of when we should switch it out (eg if the benchmarks show that it won't matter until we get to thousands, then we'll never need to do it, but if it shows that it'll be way more efficient at 20, we may want to add a TODO).

src/dbnode/encoding/proto/encoder.go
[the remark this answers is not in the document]
martin-mao: I mean if it's in the proto spec then no worries, but otherwise it might be better for us to be safe in case they change that later for whatever reason.

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -467,64 +260,35 @@ func (enc *encoder) writeIntSigMult(sig, mult uint8, floatChanged bool) {
...
 	enc.os.Reset(bytes)
-	enc.t = start
-	enc.dt = 0
-	enc.vb = 0
-	enc.xor = 0
+
+	timeUnit := initialTimeUnit(start, enc.opts.DefaultTimeUnit())
+	enc.tsEncoderState = NewTimestampEncoder(start, timeUnit, enc.opts)
+
+	enc.xorEncoderState = XOREncoder{}
```
martin-mao: Why create a new encoder instead of having a reset on existing XOREncoder?

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -467,64 +260,35 @@ func (enc *encoder) writeIntSigMult(sig, mult uint8, floatChanged bool) {
...
 
 func (enc *encoder) reset(start time.Time, bytes checked.Bytes) {
 	enc.os.Reset(bytes)
-	enc.t = start
-	enc.dt = 0
-	enc.vb = 0
-	enc.xor = 0
+
+	timeUnit := initialTimeUnit(start, enc.opts.DefaultTimeUnit())
+	enc.tsEncoderState = NewTimestampEncoder(start, timeUnit, enc.opts)
```
martin-mao: Why create new TimeStampEncoder? Are there pooled?

src/dbnode/encoding/m3tsz/iterator.go
```diff
@@ -37,39 +37,30 @@ import (
...
-	mes  encoding.MarkerEncodingScheme
 
 	// internal bookkeeping
-	t   time.Time     // current time
-	dt  time.Duration // current time delta
-	vb  uint64        // current float value
-	xor uint64        // current float xor
-	err error         // current error
+	tsIterator TimestampIterator
+	vb         uint64 // current float value
```
martin-mao: Why no XORIterator to match the XOREncoder?

src/dbnode/encoding/m3tsz/iterator.go
```diff
@@ -384,30 +227,353 @@ func (it *readerIterator) hasNext() bool {
...
+	}
+
+	opcode := opcodeAndValue >> uint(mes.NumValueBits())
+	if opcode != mes.Opcode() {
+		return 0, false, nil
+	}
+
+	var (
+		valueMask   = (1 << uint(mes.NumValueBits())) - 1
+		markerValue = int64(opcodeAndValue & uint64(valueMask))
```
martin-mao: You need to be careful here that the possible combination of your bits for your proto encoding does not conflict with potential markers for EOS, TimeUnit or Annotation. For the M3TSZ encoder int component, we picked the ordering of the opCode bits such that a conflict was not possible, otherwise, you won't be able to read what you've encoded correctly. This was also done by the TSZ XOR encoder originally.

src/dbnode/encoding/m3tsz/encoder.go
richardartoul: They're all non-pointers so this is basically "free" / equivalent to a memclear operation if that makes sense.

src/dbnode/encoding/m3tsz/encoder.go
richardartoul: Same comment, Its just returning a struct on the stack which will use the already allocated space. Its basically a "Reset"

src/dbnode/encoding/m3tsz/iterator.go
richardartoul: I re-implemented some of that logic in the proto iterator because I didn't want to re-use the really confusing logic from the m3tsz one. They share the `ReadXOR` helper function though

src/dbnode/encoding/m3tsz/iterator.go
richardartoul: Yeah I'm not actually sure how to resolve this without just re-implementing all the logic myself :/

richardartoul: One "relatively" straightforward solution would be to pass a flag to the `TimestampIterator` to tell it not to do the marker peeking logic since I dont need the `EndOfStream` marker or the `Annotation` marker anways. That just leaves the time unit changed marker stuff which I can incorporate myself I think

src/dbnode/encoding/m3tsz/encoder.go
martin-mao: Guess these aren't pooled eh? Should we think of pooling them, won't a whole bunch be created almost at the same time every block period?

src/dbnode/encoding/m3tsz/iterator.go
martin-mao: Do the two implementations essentially do the same thing? Just seems a little odd cause we have pairs of encoders and iterators.

src/dbnode/encoding/m3tsz/iterator.go
martin-mao: Yeah probably the way to go for Proto and maybe other encoders. 

It's going to be increasingly harder to ensure no conflicts with more encoders. Most of the existing markers consist of multiple 0's and 1's in a row (4 minimum I believe?), so we used that to ensure not many header bits overlaps. If you have lots of header bits, might be worth just using 2-3 bits instead of 1 bit per header to have more combinations. Also fine to just disable this logic for the proto encoder if it's too complicated.

src/dbnode/encoding/proto/common.go
```diff
@@ -0,0 +1,268 @@
...
+
+	customFieldType, ok := mapProtoTypeToCustomFieldType[fieldType]
+	return customFieldType, ok
+}
+
+func resetCustomFields(fields []customFieldState, schema *desc.MessageDescriptor) []customFieldState {
+	if cap(fields) <= maxTSZFieldsCapacityRetain {
+		return customFields(fields, schema)
+	}
+	return customFields(nil, schema)
```
martin-mao: Do we want to make maxTSZFieldsCapacityRetain configurable? I imagine in use cases where you have more than 24 fields in a proto schema, you'd always have to realloc the fields?

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,910 @@
...
+		head.IncRef()
+		head.AppendAll(buffer)
+		head.DecRef()
+	}
+
+	return ts.NewSegment(head, nil, ts.FinalizeHead)
+}
+
+// NumEncoded returns the number of encoded messages.
+func (enc *Encoder) NumEncoded() int {
```
martin-mao: Where/how is this used out of curiosity. I know we had it in the M3TSZ encoder too, but if it's never called, I wonder if a bool is better to signal first.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,910 @@
...
+func (enc *Encoder) encodeTSZValue(i int, iVal interface{}) error {
+	var (
+		val         float64
+		customField = enc.customFields[i]
+	)
+	switch typedVal := iVal.(type) {
+	case float64:
+		val = typedVal
+	case float32:
+		val = float64(typedVal)
```
martin-mao: I assume casting back and forth between float32 and float64 will give you the same bits?

Micro optimization here, but we may want to think about changing the floatXORState.WriteLoad to take in bits instead of a float64 such that for float32, you can XOR with half the amount of bits. I assume that will generally lead to smaller XORs and save some memory and CPU. This may not make a huge impact if you have byte slice fields, but I can imagine use cases where you only have a few floats and enums for example. This could potentially impact the compression algo there quite a bit and would save the cast cost since it's not straightforward to convert between the two.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,910 @@
...
+		match, err := enc.bytesMatchEncodedDictionaryValue(
+			streamBytes, lastState, currBytes)
+		if err != nil {
+			return fmt.Errorf(
+				"%s error checking if bytes match last encoded dictionary bytes: %v",
+				encErrPrefix, err)
+		}
+		if match {
+			// No changes control bit.
+			enc.stream.WriteBit(opCodeNoChange)
```
martin-mao: Do we have some tests to validate that matches are occurring as expected etc?

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,910 @@
...
+	enc.customFields[i].prevIntBits = v
+
+	numSig := encoding.NumSig(v)
+	enc.customFields[i].intSigBitsTracker.WriteIntSig(enc.stream, numSig)
+	enc.encodeIntValDiff(v, false, numSig)
+}
+
+func (enc *Encoder) encodeNextSignedIntValue(i int, next int64) {
+	prev := int64(enc.customFields[i].prevIntBits)
+	diff := next - prev
```
martin-mao: Is it worth extracting this out eventually so we can share an int encoder with m3tsz? Will the diff ordering being reversed make that impossible in the future with existing data? It's def more logical, but you may want to do the same direction as m3tsz just so it remains possible to create a common in encoder in the future (unless there are other reasons why that won't be possible).

src/dbnode/encoding/proto/common.go
robskillington: Well appending a suffix is much more common in our codebase than prefixing with a single character, is what I'm mainly getting at, for solving this type of problem.

src/cmd/services/m3dbnode/config/config.go
```diff
@@ -268,6 +275,24 @@ type HashingConfiguration struct {
...
+}
+
+// Validate validates the ProtoConfiguration.
+func (c *ProtoConfiguration) Validate() error {
+	if c == nil {
+		return nil
+	}
+
+	if c.SchemaFilePath == "" {
+		return errors.New("schemaFilePath is required for Proto data mode ")
```
robskillington: nit: Remove the trailing space here.

src/dbnode/client/config.go
```diff
@@ -79,6 +80,27 @@ type Configuration struct {
...
+}
+
+// Validate validates the ProtoConfiguration.
+func (c *ProtoConfiguration) Validate() error {
+	if c == nil {
+		return nil
+	}
+
+	if c.SchemaFilePath == "" {
+		return errors.New("schemaFilePath is required for Proto data mode ")
```
robskillington: nit: Remove the trailing space here too.

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -116,6 +91,10 @@ func (enc *encoder) Encode(dp ts.Datapoint, tu xtime.Unit, ant ts.Annotation) er
 	}
 
 	var err error
+	if err = enc.tsEncoderState.WriteTime(enc.os, dp.Timestamp, ant, tu); err != nil {
```
robskillington: nit: After this was just refactored, maybe it would be better to just structure as:
```go
err := enc.tsEncoderState.WriteTime(enc.os, dp.Timestamp, ant, tu)
if err != nil {
  return err
}
```

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -130,145 +109,19 @@ func (enc *encoder) Encode(dp ts.Datapoint, tu xtime.Unit, ant ts.Annotation) er
...
 	enc.writeFirstValue(dp.Value)
 	return nil
 }
 
 // writeNext writes the next datapoint with annotation.
 func (enc *encoder) writeNext(dp ts.Datapoint, ant ts.Annotation, tu xtime.Unit) error {
-	if err := enc.writeNextTime(dp.Timestamp, ant, tu); err != nil {
-		return err
-	}
 	enc.writeNextValue(dp.Value)
```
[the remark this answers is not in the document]
robskillington: Yeah, may as well remove these two methods that are just wrapping calls to underlying funcs.

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -635,6 +406,323 @@ func (enc *encoder) segment(resType resultType) ts.Segment {
...
 }
 
+// TimestampEncoder encapsulates the state required for a logical stream of
+// bits that represent a stream of timestamps compressed using delta-of-delta
+type TimestampEncoder struct {
+	PrevTime       time.Time
+	PrevTimeDelta  time.Duration
+	PrevAnnotation []byte
+
+	Opts encoding.Options
```
robskillington: nit: Call this `Options` when it's an exported field rather than the abbreviation?

src/dbnode/encoding/m3tsz/encoder_test.go
```diff
@@ -67,16 +66,16 @@ func TestWriteDeltaOfDeltaTimeUnitUnchanged(t *testing.T) {
 		{-4096 * time.Second, xtime.Nanosecond, []byte{0xff, 0xff, 0xff, 0xc4, 0x65, 0x36, 0x0, 0x0, 0x0}, 4},
 	}
 	for _, input := range inputs {
-		encoder.Reset(testStartTime, 0)
-		encoder.writeDeltaOfDeltaTimeUnitUnchanged(0, input.delta, input.timeUnit)
-		b, p := encoder.os.Rawbytes()
+		stream := encoding.NewOStream(nil, false, nil)
```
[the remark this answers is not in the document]
robskillington: Hm yeah, we might want some guarantees this generates bit for bit same data streams as output.

src/dbnode/encoding/m3tsz/iterator.go
robskillington: Yeah that sounds good, re: approach of not doing peek logic for proto encoder.

src/dbnode/encoding/m3tsz/iterator.go
```diff
@@ -384,30 +225,365 @@ func (it *readerIterator) hasNext() bool {
...
+	}
+
+	// NB: we add 1 here to offset the 1 we subtracted during encoding.
+	antLen = antLen + 1
+	if antLen <= 0 {
+		return fmt.Errorf("unexpected annotation length %d", antLen)
+	}
+
+	// TODO(xichen): use pool to allocate the buffer once the pool diff lands.
+	buf := make([]byte, antLen)
```
robskillington: Hm, this always allocated huh?  Pretty bad lol.  

Ah, is it worth just allocating to `PrevAnt` if the capacity of that is not at `antLen` and then read directly into that?  (Not sure if that would effect the read path, but if this is only for internal tracking then there's no reason to allocate here each time)

src/dbnode/encoding/m3tsz/iterator.go
```diff
@@ -232,16 +144,20 @@ func (it *readerIterator) readFullFloatVal() {
...
+	if err != nil {
+		it.err = err
+	}
 	it.vb ^= it.xor
 }
 
 func (it *readerIterator) readIntSigMult() {
 	if it.readBits(1) == opcodeUpdateSig {
-		if it.readBits(1) == opcodeZeroSig {
+		if it.readBits(1) == OpcodeZeroSig {
```
robskillington: Can't we unexport these and just use these internally and export an `IntIterator` or similar?  I assume not, but am curious to the backstory here.

src/dbnode/encoding/m3tsz/iterator.go
```diff
@@ -384,30 +225,365 @@ func (it *readerIterator) hasNext() bool {
...
+func (it *TimestampIterator) tryPeekBits(stream encoding.IStream, numBits int) (uint64, bool) {
+	res, err := stream.PeekBits(numBits)
+	if err != nil {
+		return 0, false
+	}
+	return res, true
+}
+
+// ReadXOR reads the next XOR value.
+func ReadXOR(stream encoding.IStream, prevXOR uint64) (uint64, error) {
```
robskillington: Would it not make sense to make this an `XORIterator`  and then you wouldn't have to pass in the `prevXOR`?

src/dbnode/encoding/m3tsz/m3tsz.go
```diff
@@ -29,28 +29,36 @@ const (
...
 	DefaultIntOptimizationEnabled = true
 
+	// OpcodeZeroSig indicates that there were zero significant digits.
+	OpcodeZeroSig = 0x0
+	// OpcodeNonZeroSig indicates that there were a non-zero number of significant digits.
+	OpcodeNonZeroSig = 0x1
+
+	// NumSigBits is the number of bits required to encode the maximum possible value
+	// of significant digits.
+	NumSigBits = 6
```
robskillington: It would be great if we could use `IntIterator` and `XORIterator` to avoid needing to export these.  Thoughts?

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,910 @@
...
+	enc.customFields[i].prevIntBits = uint64(v)
+	if v < 0 {
+		neg = true
+		v = -1 * v
+	}
+
+	vBits := uint64(v)
+	numSig := encoding.NumSig(vBits)
+
+	enc.customFields[i].intSigBitsTracker.WriteIntSig(enc.stream, numSig)
```
robskillington: Seems like if you passed `v int64` directly to `WriteIntSig(...)` it could do the caching of the `prevIntBits` on the `IntSigBitsTracker` and make the call to `NumSig(uint64(v))` itself to calculate the num sig itself yeah?

That would get rid of the `prevIntBits` on the `customFieldState`.

src/dbnode/encoding/proto/common.go
```diff
@@ -0,0 +1,268 @@
...
+
+	// Float state.
+	floatXORState m3tsz.XOREncoder
+
+	// Bytes State.
+	bytesFieldDict         []encoderBytesFieldDictState
+	iteratorBytesFieldDict [][]byte
+
+	// Int state.
+	prevIntBits       uint64
```
robskillington: As discussed lower down, if we can calculate the `NumSig` in the call to `intSigBitsTracker.WriteIntSig(...)` then we can put `prevIntBits` on the `IntSigBitsTracker` and don't need this state here yeah?

src/dbnode/encoding/proto/common.go
robskillington: Honestly, why not for now just get rid of `maxTSZFieldsCapacityRetain`?  Seems like a premature optimization perhaps.  

If it becomes a problem with tons of people using a lot of fields and it starts to show up in profiles we can add this at that time.  As far as I can see pretty much most workloads are going to have basically the exact same number of fields being used (since schema doesn't change between any one encoder in a namespace currently) so this will always be the same and it's always worth reusing the fields slice.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,910 @@
...
+
+	if len(enc.unmarshaled.GetUnknownFields()) > 0 {
+		// TODO(rartoul): Make this behavior configurable / this may change once we implement
+		// mid-stream schema changes to make schema upgrades easier for clients.
+		// https://github.com/m3db/m3/issues/1471
+		return errEncoderMessageHasUnknownFields
+	}
+
+	// From this point onwards all errors are "hard errors" meaning that they should render
+	// the encoder unsable since we may have encoded partial data.
```
robskillington: nit: s/unsable/unusable

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,910 @@
...
+		head = enc.stream.Discard()
+	} else {
+		// Copy into new buffer.
+		head = enc.newBuffer(length)
+		head.IncRef()
+		head.AppendAll(buffer)
+		head.DecRef()
+	}
+
+	return ts.NewSegment(head, nil, ts.FinalizeHead)
```
robskillington: Hm, you never have to calculate the tail here?  Are there unused bits at the end of the stream?  How do those correctly get not looked at?  

Or do you rewrite bits in the tail?  (i.e. go backwards when you have more data and delete the trailing marker of no more data?)

src/dbnode/encoding/proto/encoder.go
robskillington: It's used in tests and a few other places.  Don't think it hurts to keep this?

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,910 @@
...
+	enc.closed = false
+	enc.numEncoded = 0
+}
+
+func (enc *Encoder) resetSchema(schema *desc.MessageDescriptor) {
+	enc.schema = schema
+	enc.customFields = resetCustomFields(enc.customFields, enc.schema)
+
+	enc.lastEncoded = dynamic.NewMessage(schema)
+	enc.unmarshaled = dynamic.NewMessage(schema)
```
robskillington: Is it possible to reuse dynamic.Message at all and just call reset here instead of allocating?

src/dbnode/encoding/proto/encoder.go
robskillington: Definitely unluckily, ugh.  Oh well.  In the future maybe we can add to the dynamic protobuf library something like `Float64ValueForField(index int) (float64, error)` and since we know the types up front we can call that and avoid the `interface{}` casting.

Could you add that maybe to the followup issue?

src/dbnode/encoding/proto/encoder.go
robskillington: We talked about this, seems like the way to go is create an IntEncoder for the proto package that wraps this up nicely (and since it's specific to the proto package can do things differently to m3tsz).

src/dbnode/encoding/proto/encoder.go
martin-mao: No just wondering since we use it for tracking first write logic in the code, but I didn't see any other use cases. Figured if we never read it, keeping a bool is simpler for logic, but if it's useful, then sure.

src/dbnode/encoding/m3tsz/encoder_test.go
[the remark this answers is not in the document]
richardartoul: there a bunch of tests that do this already in `encoder_test.go`

src/dbnode/encoding/m3tsz/iterator.go
```diff
@@ -79,104 +68,27 @@ func (it *readerIterator) Next() bool {
...
-		it.readBits(numBits)
-		it.readAnnotation()
-		return it.readMarkerOrDeltaOfDelta(), true
-	case it.mes.TimeUnit():
-		it.readBits(numBits)
-		it.readTimeUnit()
-		return it.readMarkerOrDeltaOfDelta(), true
-	default:
-		return 0, false
+	if done {
```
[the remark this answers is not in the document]
richardartoul: This code path has changed a bit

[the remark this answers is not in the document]
richardartoul: oh nm you're right

src/dbnode/encoding/proto/common.go
richardartoul: ok sure

src/dbnode/encoding/proto/common.go
richardartoul: ok fine to do it for now

src/dbnode/encoding/proto/encoder.go
```diff
@@ -0,0 +1,910 @@
...
+	"github.com/jhump/protoreflect/desc"
+	"github.com/jhump/protoreflect/dynamic"
+)
+
+// Make sure encoder implements encoding.Encoder.
+var _ encoding.Encoder = &Encoder{}
+
+const (
+	// Maximum capacity of a slice of TSZ fields that will be retained between resets.
+	maxTSZFieldsCapacityRetain   = 24
```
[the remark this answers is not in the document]
richardartoul: removed it entirely

src/dbnode/encoding/proto/encoder.go
richardartoul: Discussed offline, don't need the tail here because of the existing control bits. Added an extra item to this issue to track the matter of maybe adding a tail: https://github.com/m3db/m3/issues/1471

src/dbnode/encoding/proto/encoder.go
richardartoul: If the schema doesn't change, I could just call `Reset`. To do that I would need an easy way to check if two schemas are equal, which I won't have until Haijun's code lands.

Added a comment and tracking here: https://github.com/m3db/m3/issues/1471

src/dbnode/encoding/proto/encoder.go
richardartoul: Tracking here: https://github.com/m3db/m3/pull/1426

src/dbnode/encoding/m3tsz/encoder_test.go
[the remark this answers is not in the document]
martin-mao: They cover some pretty basic stuff, I wonder if we should generate some full two hour blocks for example and have that in a test data file.

src/dbnode/encoding/proto/encoder.go
richardartoul: I can't imagine that XORing float64s is any slower than XORING float32s if you're on a 64 bit architecture since they'll each fit in a register.

I also did some benchmarks on the type cast and I think its basically free (not even a nanosecond of difference in per operation cost when you add a type-cast from float32 to float64)

src/dbnode/encoding/proto/encoder.go
richardartoul: Not right now, I'm gonna add a Stats() method to the encoder so we can keep track of how well the compression is working in a follow up P.R and I'll use that to update the tests.

Tracking here: https://github.com/m3db/m3/issues/1471

src/dbnode/encoding/proto/encoder.go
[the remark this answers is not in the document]
richardartoul: Yeah I see what you're saying. I'm gonna leave it out for now because we're probably gonna want to make the dictionary compression more sophisticated in the future and we'll need a datastructure like a tree instead of the array LRU anyways

src/dbnode/encoding/proto/encoder.go
martin-mao: Sweet, sounds good. Yeah some visibility, either in test or prod would be awesome.

src/dbnode/encoding/m3tsz/encoder_test.go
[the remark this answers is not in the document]
richardartoul: Yeah I can see the value in that but this P.R is already completely out of control. Probably not gonna do that here

src/dbnode/encoding/proto/encoder.go
martin-mao: Hmm yeah looks like the cast between the two is almost nothing - I thought it was going to be more since float32 has an 8 bit exponent and 23 bit fraction whereas float64 has a 11 bit exponent and 52 bit fraction so it's not like you can simple chop off some bits, the underlying bits are quite different. 

Might be worth a benchmark to see if there's any difference in compression ratio between XORing the two when the input value set is all float32 and changing. Would be good evidence as to the design choice here either way.

src/dbnode/encoding/m3tsz/encoder_test.go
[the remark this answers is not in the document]
martin-mao: Yeah more for one for the future if we ever tweak the encoding logic (which almost never happens).

src/dbnode/encoding/m3tsz/iterator.go
richardartoul: I can try and address in my follow up. Don't want to mess with this right now haha. Doesn't affect any of our workloads right now cause its only if an annotation exists which we're not really using, although could be helpful for the index agreed

src/dbnode/encoding/m3tsz/iterator.go
richardartoul: Still have to keep these exported since the int iterator is in the prot package now

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -34,39 +32,30 @@ import (
...
 	opts encoding.Options
 
 	// internal bookkeeping
-	t   time.Time     // current time
-	dt  time.Duration // current time delta
-	vb  uint64        // current value as float bits
-	xor uint64        // current float XOR
+	tsEncoderState TimestampEncoder
+	floatEnc       FloatEncoderAndIterator
+	sigTracker     IntSigBitsTracker
```
robskillington: Brilliant, looks good.

PR #1427, opened 2019-03-04
[api] support zone/env overrides in ns handler

schallert:
The placement APIs support zone/environment overrides via HTTP headers.
This adds that functionality to the namespace APIs so users can manage
multiple M3DB clusters from a single coordinator.


src/query/api/v1/handler/database/create.go
```diff
@@ -205,7 +205,8 @@ func (h *createHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	nsRegistry, err = h.namespaceAddHandler.Add(namespaceRequest)
+	opts := handler.NewServiceOptions("kv", r.Header, nil)
```
richardartoul: Why are you passing `"kv"` as the service name? Shouldn't it be M3DB?

src/query/api/v1/handler/options.go
```diff
@@ -0,0 +1,135 @@
...
+			opts.M3Agg.WarmupDuration = m3AggOpts.WarmupDuration
+		}
+	}
+
+	return opts
+}
+
+// Validate ensures the service options are valid.
+func (opts *ServiceOptions) Validate() error {
+	if opts.ServiceName == "" {
```
richardartoul: Can we have a slice of allowed service names and do that validation here

src/query/api/v1/handler/database/create.go
schallert: Yeah I didn't quite know what to pass here. The namespace handler doesn't really use the `service` value it gets passed (only placement cares about that), as namespaces are stored at `_kv/$env/$key`. Because it's controlling KV flags I passed `kv` as the "service name" but can change this to `m3db` if it reads less awkwardly. Even `""` would probably be fine

src/query/api/v1/handler/options.go
schallert: I had trouble with this as well. We have the concept of "allowed services" in the placement handler: https://github.com/m3db/m3/blob/f8508a401e1055c7d7fdaee22982fd48858b3e37/src/query/api/v1/handler/placement/common.go#L151

I tried moving this to the common handler package, but it's used in a few other places in the placement handler. Since it turns out that the placement handler is the only thing that cares about service name (see above comment), I left it there since namespace doesn't need it and I didn't want to have to expose it. Can change that depending on how we feel about having to expose it or not

src/query/api/v1/handler/database/create.go
richardartoul: Yeah I think the reason the namespace handler doesn't use a service value is because the only service that has namespaces is M3DB. Still I think it would read a little more cleanly if we passed the M3DB constant here, and then maybe a guard statement in the namespace handler to make sure the service is set to M3DB or return an error

src/query/api/v1/handler/options.go
richardartoul: I would just add a new slice in this package called `ValidServices` or something and put every M3Coordinator/M3Aggregator/M3DB in it and call it a day. It would just be nice to know that anywhere we're using this thing that value won't be garbage

schallert:
@richardartoul good call, refactored the allowed services part

src/query/api/v1/handler/allowed_services.go
```diff
@@ -0,0 +1,56 @@
...
+func IsAllowedService(svc string) bool {
+	_, ok := allowedServices[svc]
+	return ok
+}
+
+// AllowedServices returns the list of valid M3 services.
+func AllowedServices() []string {
+	svcs := make([]string, 0, len(allowedServices))
+	for svc, allowed := range allowedServices {
+		if allowed {
```
richardartoul: Its a little weird to me that in `IsAllowedService` you ignore the value of the boolean, but here you care about it? Maybe never check the boolean or use `struct{}{}` as the value to avoid the issue

Issue #1428, opened 2019-03-04
[query] Temporal functions panic on downstream error


andrewmains12:
Temporal functions (`increase`, `rate` etc.) panic on downstream errors:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x1d190f7]

goroutine 1738 [running]:
github.com/m3db/m3/src/query/functions/temporal.closeBlocks(0xc452afc4f0, 0x1, 0x1)
	/Users/andrewmains/Code/gocode/src/github.com/m3db/m3/src/query/functions/temporal/base.go:208 +0x37
github.com/m3db/m3/src/query/functions/temporal.(*baseNode).Process(0xc4529c8fa0, 0xc452af6ba0, 0xc452ad4300, 0x1, 0x22a8840, 0xc452ae03c0, 0x228b9c0, 0xc452afc5b0)
	/Users/andrewmains/Code/gocode/src/github.com/m3db/m3/src/query/functions/temporal/base.go:200 +0xdf4
github.com/m3db/m3/src/query/executor/transform.(*Controller).Process(0xc452af6ae0, 0xc452af6ba0, 0x22a8840, 0xc452ae03c0, 0x1, 0x0)
	/Users/andrewmains/Code/gocode/src/github.com/m3db/m3/src/query/executor/transform/controller.go:43 +0xa5
github.com/m3db/m3/src/query/functions.(*FetchNode).Execute(0xc451c8f1e0, 0xc452af6ba0, 0x1, 0x1)
	/Users/andrewmains/Code/gocode/src/github.com/m3db/m3/src/query/functions/fetch.go:131 +0x137
github.com/m3db/m3/src/query/executor.sourceRequest.Process(0x228c1c0, 0xc451c8f1e0, 0xc452af6b40, 0x22a4740, 0xc45291d100, 0x0, 0xc449840f88)
```

Root cause is from my [tracing PR](https://github.com/m3db/m3/pull/1321/files#diff-88e72767bcc80da9fb2734848ff08072R197); I deferred my close before the errcheck.

PR #1429, opened 2019-03-04
Fix panic in temporal functions; check err before deferring.

andrewmains12:
Fix for #1428 . 

I've added tests that:

- reproduce the issue on master
- make sure that we're still closing everything.

The tests are a bit wonky due to some 2 level mocking I needed to do--let me know if you have questions or suggestions for improvements there!


src/query/generated/mocks/generate.go
```diff
@@ -27,5 +27,7 @@
 // mockgen rules for generating mocks for unexported interfaces (file mode).
 //go:generate sh -c "mockgen -package=m3ql -destination=$GOPATH/src/github.com/m3db/m3/src/query/parser/m3ql/types_mock.go -source=$GOPATH/src/github.com/m3db/m3/src/query/parser/m3ql/types.go"
 //go:generate sh -c "mockgen -package=transform -destination=$GOPATH/src/github.com/m3db/m3/src/query/executor/transform/exec_mock.go -source=$GOPATH/src/github.com/m3db/m3/src/query/executor/transform/exec.go"
+//go:generate sh -c "mockgen -package=transform -destination=$GOPATH/src/github.com/m3db/m3/src/query/executor/transform/exec_mock.go -source=$GOPATH/src/github.com/m3db/m3/src/query/executor/transform/exec.go"
```
andrewmains12: Whoops cleaning this.

.excludelint
```diff
@@ -1,4 +1,5 @@
 (vendor/)
 (generated/)
 (_mock.go)
+(_mock_test.go)
```
andrewmains12: This btw, is to allow us to generate private mocks purely within testing scope; I don't want to export some mocks.

src/query/functions/temporal/base.go
```diff
@@ -81,12 +81,23 @@ func (o baseOp) Node(controller *transform.Controller, opts transform.Options) t
...
 type baseNode struct {
 	op            baseOp
-	controller    *transform.Controller
+	controller    controller
 	cache         *blockCache
 	processor     Processor
 	transformOpts transform.Options
 }
 
+// controller is a minimal interface of transform.Controller, used exclusively
```
benraskin92: Not a huge fan of having test specific stuff in our code. Anyway we can do this without putting this here?

andrewmains12: Agreed--let me explain what I'm trying to do; maybe there's a good way to do it without this :)

I want to test that blocks we create are closed (especially in case of an error). That means that I need a block that allows me to make that assertion; regular blocks don't (which is reasonable). Since we create blocks via `controller.NewBlockBuilder`, `controller` is my entry point into injecting a new block. Therefore, I mock out controller.

Alternatives considered: 
- create a public transform.Controller interface, and use that everywhere instead. I didn't think it was worth refactoring every usage of controller for this, but we could.

src/query/functions/temporal/base_test.go
```diff
@@ -351,6 +353,158 @@ func setup(numBlocks int, duration time.Duration, nextBound int) *testContext {
...
+
+// closeSpyBlockBuilder wraps a block.Builder to build a closeSpyBlock
+// instead of a regular block. It is otherwise equivalent to the wrapped Builder.
+type closeSpyBlockBuilder struct {
+	block.Builder
+
+	BuiltBlock *closeSpyBlock
+}
+
+func (bbc *closeSpyBlockBuilder) Build() block.Block {
```
arnikola: Maybe I've spent too long on the internet, but could we rename this to `(b *closeSpyBlockBuilder)`?

src/query/functions/temporal/base_test.go
```diff
@@ -351,6 +353,158 @@ func setup(numBlocks int, duration time.Duration, nextBound int) *testContext {
...
+
+// closeSpyBlock wraps a block.Block to allow assertions on the Close()
+// method.
+type closeSpyBlock struct {
+	block.Block
+
+	Closed bool
+}
+
+func (bwc *closeSpyBlock) Close() error {
```
arnikola: Same here, can we rename this to `(b *closeSpyBlock)`

src/query/functions/temporal/base_test.go
```diff
@@ -351,6 +353,158 @@ func setup(numBlocks int, duration time.Duration, nextBound int) *testContext {
...
+	bwc.Closed = true
```
arnikola: If you're doing this anyway, might as well make `closed` an int and append it to make sure we close exactly once

src/query/functions/temporal/base_test.go
```diff
@@ -351,6 +353,158 @@ func setup(numBlocks int, duration time.Duration, nextBound int) *testContext {
...
+	}
+	return unconsolidatedBlockWithSeriesIterErr{
+		Err:                 mbu.Err,
+		UnconsolidatedBlock: unconsolidated,
+	}, nil
+}
+
+type unconsolidatedBlockWithSeriesIterErr struct {
+	block.UnconsolidatedBlock
+	Err error
```
arnikola: Can probably be private

src/query/functions/temporal/base_test.go
```diff
@@ -351,6 +353,158 @@ func setup(numBlocks int, duration time.Duration, nextBound int) *testContext {
...
+	testErr := errors.New("test err")
+	errBlock := blockWithDownstreamErr{Block: tc.Blocks[1], Err: testErr}
+
+	require.NoError(t, tc.Node.Process(models.NoopQueryContext(), parser.NodeID(0), errBlock))
+
+	err := tc.Node.Process(models.NoopQueryContext(), parser.NodeID(0), tc.Blocks[0])
+	require.EqualError(t, err, testErr.Error())
+}
+
+// Types for TestBaseWithDownstreamError
```
arnikola: nit: I think may be a bit clearer to group a bunch of these test types together? Up to you

src/query/functions/temporal/base_test.go
andrewmains12: Haha yeah good call.

src/query/functions/temporal/base_test.go
```diff
@@ -351,6 +353,158 @@ func setup(numBlocks int, duration time.Duration, nextBound int) *testContext {
...
+	}
+	return bbc.BuiltBlock
+}
+
+// closeSpyBlock wraps a block.Block to allow assertions on the Close()
+// method.
+type closeSpyBlock struct {
+	block.Block
+
+	Closed bool
```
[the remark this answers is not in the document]
andrewmains12: So I don't know that style guides cover this, but my general rule is to make member's public when they're meant to be read outside of the struct, as this one is. That applies even if the class itself is private to the package. 

Private struct fields imo are more for internal state where nothing outside of the struct should read the fields, which isn't true here.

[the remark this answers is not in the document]
arnikola: Don't have very strong thoughts on it myself; personally prefer to default to making everything as closed as possible, then open it when necessary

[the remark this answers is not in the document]
andrewmains12: For sure! It's worth noting that both options are actually equivalent in terms of visibility; in both cases, everything in the package can see the field, and nothing outside the package can. That's why I tend to treat it as a semantic marker.

Anyhow, just a point of discussion :)

src/query/functions/temporal/base_test.go
```diff
@@ -351,6 +353,158 @@ func setup(numBlocks int, duration time.Duration, nextBound int) *testContext {
 	}
 }
 
+// TestBaseWithDownstreamError checks that we handle errors from blocks correctly
+func TestBaseWithDownstreamError(t *testing.T) {
```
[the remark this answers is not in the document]
andrewmains12: > but this would be a good test to run on other functions as well

Probably, but I'd rather fix the panic first :). I can put up a separate issue for it.

src/query/functions/temporal/base_test.go
```diff
@@ -351,6 +353,158 @@ func setup(numBlocks int, duration time.Duration, nextBound int) *testContext {
...
+	require.EqualError(t, err, testErr.Error())
+}
+
+// Types for TestBaseWithDownstreamError
+
+// blockWithDownstreamErr overrides only Unconsolidated() for purposes of returning a
+// an UnconsolidatedBlock which errors on SeriesIter() (unconsolidatedBlockWithSeriesIterErr)
+type blockWithDownstreamErr struct {
+	block.Block
+	Err error
```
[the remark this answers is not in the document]
andrewmains12: See above response

src/query/functions/temporal/base_test.go
andrewmains12: See above response

src/query/functions/temporal/base_test.go
andrewmains12: Personally I like them closer to where they're used, since they're effectively part of the particular tests (I'd put them inside the tests if I could, but alas, no method definitions within functions :/ )

src/query/functions/temporal/base.go
andrewmains12: So it's not too bad, but given that this is a bug fix for a panic, I'd rather get this in first and do the refactor after. Thoughts?

arnikola: If we're going more hotfix then I'd say jam it with no test; if it's not a hotfix wouldn't mind having the refactor in, but will be happy enough with pulling this into an isolated file then filing an explicit issue if that sounds good?

src/query/functions/temporal/base_test.go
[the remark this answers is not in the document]
arnikola: True enough; I guess there's a little weirdness in public accessors on private fields but that seems like a general go problem

src/query/functions/temporal/base_test.go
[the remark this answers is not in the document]
arnikola: Sounds good 👍

src/query/functions/temporal/base_test.go
arnikola: Yeah totally fair; I guess the tradeoff is extra clarity for particular tests when the utils are defined next to them or extra clarity for what the tests themselves cover when they're grouped

Either way super minor nitpick haha

src/query/functions/temporal/base.go
andrewmains12: Sure thing--filed https://github.com/m3db/m3/issues/1430 to track. Will do that shortly!

src/query/functions/temporal/base.go
```diff
@@ -256,15 +267,16 @@ func (c *baseNode) processCompletedBlocks(queryCtx *models.QueryContext, process
 	sp, _ := opentracingutil.StartSpanFromContext(queryCtx.Ctx, c.op.OpType())
 	defer sp.Finish()
 
-	blocks := make([]block.Block, len(processRequests))
-	for i, req := range processRequests {
+	blocks := make([]block.Block, 0, len(processRequests))
```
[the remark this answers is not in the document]
andrewmains12: Ah yeah this is the reason for the panic. Previously, I was allocating blocks with `length = len(processRequests)`, which meant that the array was:

```
[]block.Block{nil, nil, ...}
```

In case of early return (on error), we'd loop through every block and try to close it, causing a segfault.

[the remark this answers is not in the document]
arnikola: Ah gotcha; would it be worth adding nil checks into `closeBlocks`?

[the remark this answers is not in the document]
andrewmains12: Eh I don't think so; there's no valid use case for having a nil element in the slice. That probably means I should a check in the test though--good call!

PR #1436, opened 2019-03-08
[docker] add image for setting recommended sysctls

schallert:
This PR provides users with a docker image and Kubernetes DaemonSet that
will set the host's sysctls to our recommended settings. We try to be
careful about this and recommend pinning to a docker image SHA so that
the content is immutable (i.e. someone who controlled our Quay repo
couldn't run arbitrary root containers on hosts that have the image).


richardartoul:
@schallert Should we add a comment somewhere that explains all the places you need to update values if you want to change the recommended defaults? Need to update constants in code, constants in documentation, and also this script now

PR #1437, opened 2019-03-08
Fix panic introduced by OStream optimization

richardartoul:
A panic was introduced by #1399 that can occur when many reads are happening in parallel for the same series. The panic occurs because the previous P.R introduced a called to `Reset()` on a `checked.Bytes` on the read path.

The read path is only protected by an RLock so this would be interpreted by the checked.Bytes as a double write and cause a panic. Technically no double write was taking place because the second called to `Reset` would be a no-op, but the checked.Bytes has no way to distinguish that and we shouldn't be doing any mutation on the read path anyways.

This P.R updates the code path so that the call to `RawBytes` actually just returns `[]byte` instead of `checked.Bytes`. This avoids the issue of performing a mutation on the read path entirely.

I've also included a regression test which would have caught the issue.


src/dbnode/storage/series/series_test.go
```diff
@@ -849,6 +850,55 @@ func TestSeriesWriteReadFromTheSameBucket(t *testing.T) {
 	require.Equal(t, 3, len(values))
 }
 
+// TestSeriesWriteReadParallel is a regression test that was added to capture panics that might
+// arise when many parallel writes and reads are happening at the same time.
+func TestSeriesWriteReadParallel(t *testing.T) {
+	var (
+		numWorkers        = 100
+		numStepsPerWorker = 100
```
robskillington: minor nit: Maybe make the num steps per worker so magnitude higher than the num workers?

i.e. it'd be better to spin up 32 workers and have them write 300 times each than 100 workers with 100 bits of work, since there will be less contention in terms of worker pre-emption with less workers and more likely the test will run longer with all workers running concurrently?

If you wanted to be super paranoid about it, you could do `numWorkers := runtime.NumCPU(); numStepsPerWorker := 10000/numWorkers;` which would likely give you most optimal settings for less pre-emption and more likely to hit the parallel work case on a per machine basis.

richardartoul: sure, done. This made the test way more expensive so I marked it as "big" and put it in its own file

PR #1442, opened 2019-03-11
[query] Add multi fetch result tests

arnikola:


src/query/storage/m3/cluster_resolver_test.go
```diff
@@ -403,3 +403,51 @@ func TestLongUnaggregatedRetention(t *testing.T) {
...
+			Downsample:  &ClusterNamespaceDownsampleOptions{All: false},
+			Session:     session,
+		},
+	)
+	require.NoError(t, err)
+
+	now := time.Now()
+	end := now
+
+	for i := 27; i < 17520; i++ {
```
andrewmains12: Magic numbers? What do they mean?

src/query/storage/m3/multi_fetch_result_test.go
```diff
@@ -0,0 +1,143 @@
...
+	short        = "short"
+	long         = "long"
+	unaggregated = "unaggregated"
+)
+
+// NB: Each seriesIterators has two seriesIterator; one with a constant ID which
+// will be overwritten as necessary by multi_fetch_result, and one per namespace
+// which should not be overwritten and should appear in the results.
+func generateSeriesIterators(
+	ctrl *gomock.Controller, ns string) encoding.SeriesIterators {
```
andrewmains12: Not sure, but would it be worth doing a fake implementation of `SeriesIterators` instead? This is kind of a long mock sequence, and it's a little less clear what the behavior is than a fake implementation.

src/query/storage/m3/cluster_resolver_test.go
arnikola: ah, it's to ensure we hit a retention out of the unaggregated namespace but within `metrics_600s_17520h `

src/query/storage/m3/multi_fetch_result_test.go
arnikola: Hey, that's fair; unfortunately there's a lot of stuff we'd have to fake here (faked `SeriesIterator` means that we need to return faked `TagIterators` etc, etc, which is quite annoying

Perhaps if we'd had more tightly scoped iterators or something initially we'd be better off for the moment, but looks like we may be stuck with it as is unfortunately :(

PR #1447, opened 2019-03-12
Add error checks to all block.Builder calls

andrewmains12:
I don't think this is causing a bug right now, but I think it might be in my PR. I figured it was worth just adding err checks for all these calls; there's no reason (that I'm aware of) to ignore them.

Depending on how awful it is to do, I may try to get `errcheck` passing on this repo, so we can catch these kinds of things.

### DIFF
--- a/src/query/functions/binary/and.go
+++ b/src/query/functions/binary/and.go
@@ -58,11 +58,16 @@ func makeAndBlock(
 		for idx, value := range lValues {
 			rIdx := intersection[idx]
 			if rIdx < 0 || math.IsNaN(rValues[rIdx]) {
-				builder.AppendValue(index, math.NaN())
+				if err := builder.AppendValue(index, math.NaN()); err != nil {
+					return nil, err
+				}
+
 				continue
 			}
 
-			builder.AppendValue(index, value)
+			if err := builder.AppendValue(index, value); err != nil {
+				return nil, err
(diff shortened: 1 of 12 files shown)


src/query/test/block.go
```diff
@@ -124,16 +124,23 @@ func NewBlockFromValuesWithSeriesMeta(
 }
 
 // NewBlockFromValuesWithMetaAndSeriesMeta creates a new block using the provided values
+// Note: panics on errors, since this is a testutil
```
arnikola: Would be nice if this took in a `t.Testing` and failed that but totally not worth the effort :p

src/query/functions/binary/and.go
```diff
@@ -58,11 +58,15 @@ func makeAndBlock(
 		for idx, value := range lValues {
 			rIdx := intersection[idx]
 			if rIdx < 0 || math.IsNaN(rValues[rIdx]) {
-				builder.AppendValue(index, math.NaN())
+				if err := builder.AppendValue(index, math.NaN()); err != nil {
+					return nil, err
+				}
```
arnikola: nit: newline here

src/query/test/block.go
andrewmains12: Haha yeah that was my thought as well.

PR #1448, opened 2019-03-13
[query] Add approximate limit calculation for graphite render endpoint

arnikola:
Approximates the upper bound for computed datapoints in a graphite query.


arnikola:
Will likely close this in favour of https://github.com/m3db/m3/pull/1449 which does a similar thing but more nicely and more accurately

PR #1449, opened 2019-03-13
[query] Add cost enforcer to graphite queries

arnikola:


src/query/storage/converter.go
```diff
@@ -361,11 +371,29 @@ func decompressConcurrently(
...
+}
+
+type syncEnforcer struct {
+	enforcer cost.ChainedEnforcer
+	mu       sync.Mutex
+}
+
+func (e *syncEnforcer) Add(c xcost.Cost) xcost.Report {
+	e.mu.Lock()
+	r := e.enforcer.Add(c)
```
andrewmains12: This shouldn't be necessary, I don't think; the enforcers are designed to be thread safe (through use of atomic variables). If they're not, we should probably fix it at that level instead of adding a wrapper.

arnikola: Sure, no worries, was probably being a little overcautious here

PR #1458, opened 2019-03-14
[query] Add support for dropping NaNs when rendering results

benraskin92:


src/cmd/services/m3query/config/config.go
```diff
@@ -128,6 +128,10 @@ type Configuration struct {
 
 	// Cache configurations.
 	Cache CacheConfiguration `yaml:"cache"`
+
+	// DropNaNs drops NaNs before returning query results.
+	// If you want to simulate Promtheus exactly, set this to true.
+	DropNaNs bool `yaml:"dropNaNs"`
```
arnikola: Don't have this as a top level config setting

src/query/api/v1/handler/prometheus/native/common_test.go
```diff
@@ -212,11 +218,109 @@ func TestRenderResultsJSON(t *testing.T) {
...
+	valsWithNaN.SetValueAt(1, math.NaN())
+
+	series := []*ts.Series{
+		ts.NewSeries([]byte("foo"),
+			valsWithNaN, test.TagSliceToTags([]models.Tag{
+				models.Tag{Name: []byte("bar"), Value: []byte("baz")},
+				models.Tag{Name: []byte("qux"), Value: []byte("qaz")},
+			})),
+		ts.NewSeries([]byte("bar"),
+			ts.NewFixedStepValues(10*time.Second, 2, 2, start), test.TagSliceToTags([]models.Tag{
```
arnikola: Have one of these tests have some nans

src/cmd/services/m3query/config/config.go
benraskin92: What do you propose? Maybe have like a `QueryOptions` struct?

arnikola: `Result` would make more sense imo; also should default to true

arnikola: Maybe rename it to `returnNansInQueryResult` or something

src/query/api/v1/handler/prometheus/native/common_test.go
benraskin92: Actually `TestRenderResultsJSONWithDroppedNaNs` already has a `NaN` and a real number

src/cmd/services/m3query/config/config.go
```diff
@@ -155,6 +156,22 @@ type FilterConfiguration struct {
...
 
+// ResultOptions are the result options for query.
+type ResultOptions struct {
+	// DropNaNsInResults drops NaNs before returning query results.
+	// The default is true, which matches Prometheus
+	DropNaNsInResults *bool `yaml:"dropNaNs"`
+}
+
+// DropNaNsOrDefault returns the provided input or default if none is provided
+func (c ResultOptions) DropNaNsOrDefault() *bool {
```
arnikola: This should probably return a `bool` not a `*bool`

benraskin92: Shoot yeah meant to change that.

src/cmd/services/m3query/config/config.go
andrewmains12: +1 on `ResultOptions` (or `Result`)

src/cmd/services/m3query/config/config.go
```diff
@@ -146,6 +151,22 @@ type FilterConfiguration struct {
 	CompleteTags Filter `yaml:"completeTags"`
 }
 
+// ResultOptions are the result options for query.
+type ResultOptions struct {
+	// DropNaNsInResults drops NaNs before returning query results.
+	// The default is true, which matches Prometheus
+	DropNaNsInResults *bool `yaml:"dropNaNs"`
```
andrewmains12: Nit: if `DropNan's` is the default, could rename to `KeepNaN's`, and use the builtin default for bools? Avoids the pointer.

arnikola: Also can you call it `dropNans` instead, the NaN is annoying and out of place in a config

PR #1459, opened 2019-03-14
[query] Add per query enforcer to graphite path

arnikola:


src/query/graphite/storage/m3_wrapper.go
```diff
@@ -139,7 +139,10 @@ func (s *m3WrappedStore) FetchByQuery(
 	m3ctx, cancel := context.WithTimeout(ctx.RequestContext(), opts.Timeout)
 	defer cancel()
 	fetchOptions := storage.NewFetchOptions()
-	fetchOptions.Enforcer = s.enforcer
+	perQueryEnforcer := s.enforcer.Child(cost.QueryLevel)
```
andrewmains12: Btw, do we call this function once per query? If not, we should do this higher up.

arnikola: Yeah, to reiterate what we talked in chat, the query gets pulled out of the request and sent here, which translates it to m3 query matchers just above here, and the actual query is executed shortly after

Issue #1453, opened 2019-03-13
Poor performance for tag completion endpoints


arnikola:
Users have been reporting poor performance when hitting any of the tag completion endpoints in m3query and m3coordinator; these requests are slow and often OOM clusters. 

The root cause of this is that the M3DB client does not expose an endpoint to list available tag name/pairs, and m3query must make do by fetching all IDs matching the request then deduping them. Essentially, this means that trying to get a list of available tags pulls every single series into memory within M3DB, returns it up to Query, and then processed to return the list of valid tags. 

This issue affects Grafana templating, Graphite `find` endpoints, and any tag completion features.

This issue will track work to build and expose a tag completion endpoint in M3DB which will greatly increase the performance, and reduce the memory overhead required to perform these requests.

Most of these are being done independently but there's a few dependancies which will be called out: 

- [x] Hook up the new endpoints m3query side (#1481)
- [x] Add Thrift RPC types for the client (#1460)
- [x] Add the endpoint on the client session side (#1483)
- [x] Push down a single `Results` object into the query and merge all individual block postings lists into that, rather than allocating a Results per block then merging them together (#1474) 
- [x] Build the logic on the index itself (#1463) 
    - [x] (this will be refactored to look more similar to #1474 above)

evanhuang996:
How is the progress?

arnikola:
Hey, will add a checklist here for slightly better tracking. Most of these are being done independently but there's a few dependancies which will be called out: 

[] Hook up the new endpoints m3query side (#1481)
[] Add Thrift RPC types for the client (#1460)
[] Add the endpoint on the client session side (#1483)
[] Push down a single `Results` object into the query and merge all individual block postings lists into that, rather than allocating a Results per block then merging them together (#1474) 
[] Build the logic on the index itself (#1463) (this will be refactored to look more similar to #1474 above)

arnikola:
Resolved in 0.8.0 :+1:

PR #1460, opened 2019-03-14
[dbnode] Add rpc/AggregateQuery types

prateek:
Only adding the rpc interface for some of the work related to https://github.com/m3db/m3/issues/1453 

/cc @arnikola @richardartoul


src/dbnode/generated/thrift/rpc.thrift
```diff
@@ -312,6 +314,41 @@ struct HealthResult {
...
+	2: required i64 rangeStart
+	3: required i64 rangeEnd
+	4: required string nameSpace
+	5: optional i64 limit
+	6: optional AggregateQueryType aggregateQueryType = AggregateQueryType.AGGREGATE_BY_TAG_NAME_VALUE
+	7: optional TimeType rangeType = TimeType.UNIX_SECONDS
+	8: optional TimeType resultTimeType = TimeType.UNIX_SECONDS
+}
+
+struct AggregateQueryResult {
```
prateek: TBH: intentionally avoiding making this a generic type tree. I'm sure we can do aggregation functions in the future but until it's needed it's easier to not introduce the complexity.

src/dbnode/generated/thrift/rpc.thrift
```diff
@@ -312,6 +314,41 @@ struct HealthResult {
...
+enum AggregateQueryType {
+	AGGREGATE_BY_TAG_NAME,
+	AGGREGATE_BY_TAG_NAME_VALUE,
+}
+
+// AggregateQueryRequest comes from our desire to aggregate on incoming data.
+// We currently only support retrieval of facets, but could extend this based on
+// requirements in the future. Currently, the two predominant use-cases are:
+// (1) Given a filter query (optionally), return all known tag keys matching this restriction
+// (2) Given a filter query (optionally), return all know tag key+values matching this restriction
```
arnikola: Should this take in an optional result filter on keys? As an example, if I'm doing a graphite query for say, `foo.bar.*`, I only want to list the values matching the restriction for `__g2__`, rather than for any tags that come after. For graphite we can do something silly like add `__g3__ !=~ ".*"` to soft-restrict this to the desired output, but that doesn't really translate well to arbitrary tag/values

i.e. with `foo.bar.*` as a query, the expected results for some metrics would be:
```
foo.bar.biz -> biz
foo.bar.qux -> qux
foo.bar.a.b -> a
```

(sorry for using mostly graphite here, it's easier than writing out tags explicitly for the example :p)

src/dbnode/generated/thrift/rpc.thrift
```diff
@@ -312,6 +314,41 @@ struct HealthResult {
...
+struct AggregateQueryRequest {
+	1: required Query query
+	2: required i64 rangeStart
+	3: required i64 rangeEnd
+	4: required string nameSpace
+	5: optional i64 limit
+	6: optional AggregateQueryType aggregateQueryType = AggregateQueryType.AGGREGATE_BY_TAG_NAME_VALUE
```
arnikola: I don't think this should be optional; would be better for callers to explicitly state what query type is required

src/dbnode/generated/thrift/rpc.thrift
```diff
@@ -312,6 +314,41 @@ struct HealthResult {
...
 
+enum AggregateQueryType {
+	AGGREGATE_BY_TAG_NAME,
+	AGGREGATE_BY_TAG_NAME_VALUE,
+}
+
+// AggregateQueryRequest comes from our desire to aggregate on incoming data.
+// We currently only support retrieval of facets, but could extend this based on
+// requirements in the future. Currently, the two predominant use-cases are:
+// (1) Given a filter query (optionally), return all known tag keys matching this restriction
```
arnikola: This comment is a little incorrect since `Query` is not an optional case (the no filter case would be something like `.*` which is still a query, rather than a nil)

src/dbnode/generated/thrift/rpc.thrift
```diff
@@ -312,6 +314,41 @@ struct HealthResult {
...
+	2: required bool exhaustive
+}
+
+struct AggregateQueryResultTagNameElement {
+	1: required string tagName
+	2: optional list<AggregateQueryResultTagValueElement> tagValues
+}
+
+struct AggregateQueryResultTagValueElement {
+	1: required string tagValue
```
arnikola: Guessing thrift doesn't allow byte slices instead?

src/dbnode/generated/thrift/rpc.thrift
prateek: I like it. I'll add something equivalent to `resultFilter []string` where empty means no filtering

src/dbnode/generated/thrift/rpc.thrift
prateek: Only reason to make it optional is for the case where people just curl this endpoint to see distinct tags. Seems unnecessary to make them also specify this.

That said, I agree - all the stuff calling this endpoint programmatically should totally be specifying this.

src/dbnode/generated/thrift/rpc.thrift
prateek: Fair, will add performant/non-performant types using the `...Raw` convention.

src/dbnode/generated/thrift/rpc.thrift
prateek: Good catch, made `Query` optional

PR #1461, opened 2019-03-15
Fix postings list cache causing incorrect results

richardartoul:
The postings list cache was broken in that the key did not include the field that the query was executed against. This meant that for a given segment UUID the results of a query for one term could be mixed up with the results of the same query but for a different term. This would lead to incorrect results being returned when the index block was queried.

This P.R introduces the field as part of the key in the postings list cache to resolve the issue. In addition, it introduces a property test that generates two index blocks, one with the postings list cache enabled and one without, and then executes hundreds of different queries against both blocks and makes sure that they return the exact same result in all cases.

### DIFF
--- a/src/dbnode/storage/index/block_prop_test.go
+++ b/src/dbnode/storage/index/block_prop_test.go
@@ -53,7 +53,7 @@ var (
 
 // TestPostingsListCacheDoesNotAffectBlockQueryResults verifies that the postings list
 // cache does not affect the results of querying a block by creating two blocks, one with
-// the postings list cache enable and one without. It then generates a bunch of queries
+// the postings list cache enabled and one without. It then generates a bunch of queries
 // and executes them against both blocks, ensuring that both blocks return the exact same
 // results. It was added as a regression test when we encountered a bug that caused the
 // postings list cache to cause the block to return incorrect results.
@@ -90,7 +90,7 @@ func TestPostingsListCacheDoesNotAffectBlockQueryResults(t *testing.T) {
 	cachedBlock, err := newPropTestBlock(t, blockStart, testMD, cachedOptions)
 	require.NoError(t, err)
 
-	properties.Property("Any distribution of test documents in segments does not affect query results", prop.ForAll(
+	properties.Property("Index block with and without postings list cache always return the same results", prop.ForAll(
 		func(q search.Query) (bool, error) {
 			indexQuery := Query{
 				idx.NewQueryFromSearchQuery(q),
@@ -108,10 +108,10 @@ func TestPostingsListCacheDoesNotAffectBlockQueryResults(t *testing.T) {
 			cachedResults := NewResults(testOpts)
 			exhaustive, err = cachedBlock.Query(indexQuery, QueryOptions{StartInclusive: blockStart, EndExclusive: blockStart.Add(blockSize)}, cachedResults)
 			if err != nil {
-				return false, fmt.Errorf("error querying uncached block: %v", err)
+				return false, fmt.Errorf("error querying cached block: %v", err)
 			}
 			if !exhaustive {
-				return false, errors.New("querying uncached block was not exhaustive")
+				return false, errors.New("querying cached block was not exhaustive")
 			}
 
 			uncachedMap := uncachedResults.Map()
@@ -145,11 +145,15 @@ func newPropTestBlock(t *testing.T, blockStart time.Time, nsMeta namespace.Metad
 	blk, err := NewBlock(blockStart, nsMeta, opts)
 	require.NoError(t, err)
 
-	memSeg := testSegment(t, lotsTestDocuments...).(segment.MutableSegment)
-	fstSeg := fst.ToTestSegment(t, memSeg, testFstOptions)
-	// Need at least one shard to look fulfilled.
-	fulfilled := result.NewShardTimeRanges(blockStart, blockStart.Add(testBlockSize), uint32(1))
-	indexBlock := result.NewIndexBlock(blockStart, []segment.Segment{fstSeg}, fulfilled)
+	var (
+		memSeg = testSegment(t, lotsTestDocuments...).(segment.MutableSegment)
+		fstSeg = fst.ToTestSegment(t, memSeg, testFstOptions)
+		// Need at least one shard to look fulfilled.
+		fulfilled  = result.NewShardTimeRanges(blockStart, blockStart.Add(testBlockSize), uint32(1))
+		indexBlock = result.NewIndexBlock(blockStart, []segment.Segment{fstSeg}, fulfilled)
+	)
+	// Use the AddResults API because thats the only scenario in which we'll wrap a segment
+	// in a ReadThroughSegment to use the postings list cache.
 	err = blk.AddResults(indexBlock)
 	require.NoError(t, err)
 	return blk, nil
--- a/src/dbnode/storage/index/block_prop_test.go
+++ b/src/dbnode/storage/index/block_prop_test.go
@@ -1,0 +1,156 @@
+// +build big
+//
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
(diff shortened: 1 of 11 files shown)


src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -63,6 +63,14 @@ type PostingsListCache struct {
 	metrics *postingsListCacheMetrics
 }
 
+// PostingsListCacheQuery represents a query that we want to cache the result of
+// for a given segment. Note that it include the field in the segment that the
+// query was executed on, as well as the pattern of the query itself.
+type PostingsListCacheQuery struct {
```
prateek: hm i think you need to include query/pattern type here. i.e. a term query of "abc:1*" is not the same as regexp query of "abc:1*". 

Could you ensure the tests catch this too.

src/m3ninx/search/proptest/query_gen.go
```diff
@@ -54,7 +55,8 @@ func genTermQuery(docs []doc.Document) gopter.Gen {
 	}
 }
 
-func genRegexpQuery(docs []doc.Document) gopter.Gen {
+// GenRegexpQuery generates a regexp query.
+func GenRegexpQuery(docs []doc.Document) gopter.Gen {
```
prateek: for the failure case i was describing, you probably need to change this generator to flip a biased coin and choose between this implementation and mapping the output of a TermQuery to a RegexpQuery

src/m3ninx/idx/query.go
```diff
@@ -34,6 +34,13 @@ func (q Query) String() string {
 	return q.query.String()
 }
 
+// NewQueryFromSearchQuery creates a new Query from a search.Query.
+func NewQueryFromSearchQuery(q search.Query) Query {
```
prateek: lol

src/dbnode/storage/index/postings_list_cache.go
richardartoul: This is handled. They go through separate APIs and end up with separate keys in the LRU map. See postings_list_cache_lru.go key struct

src/m3ninx/search/proptest/query_gen.go
richardartoul: Same comment

src/dbnode/storage/index/postings_list_cache.go
prateek: Fair enough. Mind including pattern here anyway? A query isn’t completely defined without it so this struct feels incomplete.

src/m3ninx/search/proptest/query_gen.go
prateek: Hm are you sure this is capable of generating the same thing as a term query?

src/dbnode/storage/index/postings_list_cache.go
richardartoul: I’d have to change the APIs though and then you have a potentially invalid iota coming from the caller

richardartoul: Are you just opposing the name? We could call it TermPattern or FieldPattern

richardartoul: Are you just opposing the name? We could call it TermPattern or FieldPattern

prateek: It’s not the name as much as it is the struct itself. Like without the pattern included, the struct is an incomplete description of a query. 

How would you feel about migrating to using search.Query here? The thought is if we add new types of queries we’d still get this part for free.

src/dbnode/storage/index/block_prop_test.go
```diff
@@ -0,0 +1,160 @@
...
+	lotsTestDocuments = util.MustReadDocs("../../../m3ninx/util/testdata/node_exporter.json", 2000)
+)
+
+// TestPostingsListCacheDoesNotAffectBlockQueryResults verifies that the postings list
+// cache does not affect the results of querying a block by creating two blocks, one with
+// the postings list cache enabled and one without. It then generates a bunch of queries
+// and executes them against both blocks, ensuring that both blocks return the exact same
+// results. It was added as a regression test when we encountered a bug that caused the
+// postings list cache to cause the block to return incorrect results.
+func TestPostingsListCacheDoesNotAffectBlockQueryResults(t *testing.T) {
```
arnikola: Nice one! Assuming this fails the old case?

src/dbnode/storage/index/postings_list_cache.go
```diff
@@ -63,6 +63,14 @@ type PostingsListCache struct {
 	metrics *postingsListCacheMetrics
 }
 
+// PostingsListCacheQuery represents a query that we want to cache the result of
+// for a given segment. Note that it include the field in the segment that the
```
arnikola: nit: `include`->`includes`

src/dbnode/storage/index/postings_list_cache_lru.go
```diff
@@ -121,18 +114,17 @@ func (c *postingsListLRU) Add(
 	var (
 		ent = &entry{
 			uuid:         segmentUUID,
-			pattern:      pattern,
+			query:        query,
```
arnikola: Could the entry have a `key` rather than a `PostingsListCacheQuery` and `patternType`?

src/dbnode/storage/index/postings_list_cache_lru.go
```diff
@@ -147,13 +139,13 @@ func (c *postingsListLRU) Add(
 // Get looks up a key's value from the cache.
 func (c *postingsListLRU) Get(
 	segmentUUID uuid.UUID,
-	pattern string,
+	query PostingsListCacheQuery,
 	patternType PatternType,
 ) (postings.List, bool) {
+	newKey := newKey(query, patternType)
```
arnikola: Won't you create this here for no reason if `uuidArray` is not present in `items`? Looks like in `Add` you always have to generate `newKey` so it's ok there, but might be unnecessary both here and in `Remove`

src/dbnode/storage/index/postings_list_cache_lru.go
```diff
@@ -208,14 +200,13 @@ func (c *postingsListLRU) removeElement(e *list.Element) {
 	entry := e.Value.(*entry)
 
 	if patterns, ok := c.items[entry.uuid.Array()]; ok {
-		key := newPatternAndPatternType(entry.pattern, entry.patternType)
-		delete(patterns, key)
+		delete(patterns, newKey(entry.query, entry.patternType))
```
arnikola: nit: for consistency maybe pull it into a variable?  `newKey := newKey(entry.query, entry.patternType)`

src/dbnode/storage/index/postings_list_cache_test.go
```diff
@@ -71,11 +76,15 @@ func init() {
 
 type testEntry struct {
 	segmentUUID  uuid.UUID
-	pattern      string
+	query        PostingsListCacheQuery
```
arnikola: nit: Could this just have a `key`?

src/dbnode/storage/index/postings_list_cache_test.go
```diff
@@ -94,15 +103,15 @@ func TestSimpleLRUBehavior(t *testing.T) {
 	putEntry(plCache, 1)
 	putEntry(plCache, 2)
 
-	expectedOrder := []string{e0.pattern, e1.pattern, e2.pattern}
+	expectedOrder := []testEntry{e0, e1, e2}
```
arnikola: nit: might be nice to have something like

```golang
requireExpectedOrder := func(t *testing.T, expectedOrder ...testEntry) {
	for i, key := range plCache.lru.keys() {
		require.Equal(t, expectedOrder[i].toKey(), key)
	}
}
// ...
requireExpectedOrder(t, e0,e1,e2)
```

src/dbnode/storage/index/read_through_segment.go
```diff
@@ -148,16 +148,18 @@ func (s *readThroughSegmentReader) MatchRegexp(
 		return s.Reader.MatchRegexp(field, c)
 	}
 
-	// TODO(rartoul): Would be nice not to allocate a string here.
-	pattern := c.FSTSyntax.String()
-	pl, ok := s.postingsListCache.GetRegexp(s.uuid, pattern)
+	plCacheQuery := PostingsListCacheQuery{
```
arnikola: Worth keeping the comment or is it not really viable to not allocate the string here any more?

src/dbnode/storage/index/read_through_segment.go
```diff
@@ -171,16 +173,18 @@ func (s *readThroughSegmentReader) MatchTerm(
 		return s.Reader.MatchTerm(field, term)
 	}
 
-	// TODO(rartoul): Would be nice to not allocate a string here.
-	termString := string(term)
-	pl, ok := s.postingsListCache.GetTerm(s.uuid, termString)
+	plCacheQuery := PostingsListCacheQuery{
```
arnikola: Same here re: comment

src/m3ninx/idx/query.go
arnikola: Would it be better to have this as a helper in `block_prop_test.go` if it's not used elsewhere?

src/dbnode/storage/index/block_prop_test.go
richardartoul: Yep

src/dbnode/storage/index/postings_list_cache_lru.go
richardartoul: yeah but its a stack alloc so I think it doesnt matter much

src/dbnode/storage/index/postings_list_cache.go
richardartoul: Search.Query is an interface :/ I don't think that makes a lot of sense.

src/m3ninx/search/proptest/query_gen.go
richardartoul: oh no I was just saying that the interface doesn't allow this to occur

src/m3ninx/idx/query.go
richardartoul: different package unfortunately :/ Otherwise I wouldn't need the constructor at all

arnikola: Ah my bad, missed that you were setting the private `query` field

src/dbnode/storage/index/postings_list_cache_lru.go
arnikola: no alloc > stack alloc ;)

PR #1462, opened 2019-03-15
Use duration instead of nanos in OpenAPI docs

richardartoul:


src/query/generated/assets/openapi/spec.yml
```diff
@@ -686,31 +686,25 @@ definitions:
...
-        type: "integer"
-        format: "int64"
-      bufferFutureNanos:
-        type: "integer"
-        format: "int64"
-      bufferPastNanos:
-        type: "integer"
-        format: "int64"
+      retentionPeriodDuration:
+        type: "stromg"
```
arnikola: Should be "string"

PR #1463, opened 2019-03-15
[index] Aggregating results on storage side

arnikola:
~~TODO:~~
- [x] ~~Test coverage~~
- [x] ~~Rebase on https://github.com/m3db/m3/pull/1474~~


src/dbnode/generated-source-files.mk
```diff
@@ -27,14 +27,16 @@ genny-all: genny-map-all genny-arraypool-all genny-leakcheckpool-all
...
-	genny-map-storage-index-results			 \
+genny-map-all:                                \
+	genny-map-client-received-blocks            \
+	genny-map-storage-block-retriever           \
+	genny-map-storage-bootstrap-result          \
+	genny-map-storage                           \
+	genny-map-storage-namespace-metadata        \
+	genny-map-storage-repair                    \
+	genny-map-storage-index-results             \
+	genny-map-storage-index-search-results      \
```
arnikola: oops obviously doesn't exist

src/dbnode/generated-source-files.mk
```diff
@@ -155,6 +157,20 @@ genny-map-storage-index-results: install-m3x-repo
 	# Rename generated map file
 	mv -f $(m3db_package_path)/src/dbnode/storage/index/map_gen.go $(m3db_package_path)/src/dbnode/storage/index/results_map_gen.go
 
+# Map generation rule for storage/index/AggregationResultsMap
+.PHONY: genny-map-storage-index-aggregation-results
+genny-map-storage-index-aggregation-results: install-m3x-repo
+	cd $(m3x_package_path) && make hashmap-gen                \
```
prateek: s/hashmap-gen/idhashmap-gen

It’s the target for genny maps with ident.ID keys, so then you can get rid of the key_type specifier below.

src/dbnode/storage/types.go
```diff
@@ -278,6 +278,14 @@ type databaseNamespace interface {
 		opts index.QueryOptions,
 	) (index.QueryResults, error)
 
+	// AggregateQuery resolves the given query into unique tag/value pairs for
+	// that query.
+	AggregateQuery(
+		ctx context.Context,
+		query index.Query,
+		opts index.QueryOptions,
+	) (index.QueryResults, error)
```
prateek: need to update the return value

src/dbnode/storage/types.go
```diff
@@ -472,6 +480,14 @@ type namespaceIndex interface {
...
```
prateek: here too

src/dbnode/storage/types.go
```diff
@@ -278,6 +278,14 @@ type databaseNamespace interface {
 		opts index.QueryOptions,
 	) (index.QueryResults, error)
 
+	// AggregateQuery resolves the given query into unique tag/value pairs for
+	// that query.
+	AggregateQuery(
+		ctx context.Context,
+		query index.Query,
+		opts index.QueryOptions,
```
prateek: probably need an `AggregateQueryOptions` struct

src/dbnode/storage/index/types.go
```diff
@@ -65,6 +65,15 @@ type QueryOptions struct {
 	StartInclusive time.Time
 	EndExclusive   time.Time
 	Limit          int
+
+	// Optional param to filter aggregate values.
+	TermFilter *AggregateValuesMap
```
prateek: break into separate struct - i.e. AggregateQueryOptions v QueryOptions should be separate

src/dbnode/storage/types.go
arnikola: Yeah, good point; was initially toying with using some of the existing functions and rolling specific `AggregateQueryOptions` and `AggregateResults` into `QueryOptions` and `Results` respectively, but less of a concern after refactoring

src/dbnode/generated-source-files.mk
```diff
@@ -157,20 +157,6 @@ genny-map-storage-index-results: install-m3x-repo
 	# Rename generated map file
 	mv -f $(m3db_package_path)/src/dbnode/storage/index/map_gen.go $(m3db_package_path)/src/dbnode/storage/index/results_map_gen.go
 
-# Map generation rule for storage/index/AggregationValuesMap
-.PHONY: genny-map-storage-index-aggregation-values
```
prateek: why delete this?

src/dbnode/generated-source-files.mk
```diff
@@ -206,3 +220,22 @@ genny-leakcheckpool-fetch-tagged-op: install-m3x-repo
 	elem_type_pool=fetchTaggedOpPool                    \
 	target_package=$(m3db_package)/src/dbnode/client    \
 	out_file=fetch_tagged_op_leakcheckpool_gen_test.go
+
+# Map generation rule for dependent generated maps which are built on top of a
+# generated map
+.PHONY: genny-map-dependent-all
+genny-map-dependent-all:                      \
```
prateek: don't think you need to break it up like this. i.e. you can just have 

```
genny-map-all:
  ...
  genny-map-storage-index-aggregation-results

genny-map-storage-index-aggregation-results: genny-map-whichever-one-this-depends-on
   ...
```

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/m3db/m3/src/m3ninx/doc"
+	"github.com/m3db/m3x/ident"
+	"github.com/m3db/m3x/pool"
+)
+
+var (
+	errUnableToAddAggregateResultMissingID = errors.New("no id for result")
+)
+
+type aggregatedResults struct {
```
prateek: need tests for this struct verifying a few different things:
- aggregation results
- copying/non-copying behaviour 
- finalizing/non-finalizing behaviour

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,199 @@
...
+
+	r.resultsMap.Set(termID, aggValues)
+	return nil
+}
+
+func (r *aggregatedResults) AggregateDocument(
+	document doc.Document,
+	opts QueryOptions,
+) error {
+	// TODO: is this neccessary to check for document correctness?
```
prateek: why do this?

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,199 @@
...
+		// Return these values to the pool.
+		r.valuesPool.Put(aggValues)
+		return err
+	}
+
+	r.resultsMap.Set(termID, aggValues)
+	return nil
+}
+
+func (r *aggregatedResults) AggregateDocument(
```
prateek: why have `AggregateDocument` and `AddIDAndValues` separate?

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,199 @@
...
+	// NB: if it is already in the map, this is a valid ID to add and it's not
+	// necessary to check against the filter.
+	valueMap, found := r.resultsMap.Get(termID)
+	if found {
+		valueID := ident.BytesID(value)
+		return valueMap.addValue(valueID)
+	}
+
+	// if this term hasn't been seen, ensure it should be included in output.
+	if !opts.TermFilter.Contains(termID) {
```
prateek: hm should this be the first thing you do in the function?

src/dbnode/storage/index/types.go
```diff
@@ -125,16 +135,108 @@ type ResultsAllocator func() Results
...
+	Map() *AggregateResultsMap
+
+	// Reset resets the AggregateResults object to initial state.
+	Reset(nsID ident.ID)
+
+	// Finalize releases any resources held by the AggregateResults object,
+	// including returning it to a backing pool.
+	Finalize()
+
+	// NoFinalize marks the AggregateResults such that a subsequent call to
```
prateek: do you need this method?

src/dbnode/storage/index/types.go
```diff
@@ -125,16 +135,108 @@ type ResultsAllocator func() Results
...
+	Namespace() ident.ID
+
+	// Map returns a map from tag name -> possible tag values,
+	// comprising search results.
+	Map() *AggregateResultsMap
+
+	// Reset resets the AggregateResults object to initial state.
+	Reset(nsID ident.ID)
+
+	// Finalize releases any resources held by the AggregateResults object,
```
prateek: same here, is this needed?

src/dbnode/storage/index/types.go
```diff
@@ -125,16 +135,108 @@ type ResultsAllocator func() Results
...
+
+	// Get retrieves a AggregateResults object for use.
+	Get() AggregateResults
+
+	// Put returns the provided AggregateResults to the pool.
+	Put(value AggregateResults)
+}
+
+// AggregateValues is a collection of values for an aggregation query.
+type AggregateValues interface {
```
prateek: is this just a wrapper on a code-gen'd type, if so - can we use the struct where needed instead?

src/dbnode/storage/index/types.go
```diff
@@ -169,6 +271,13 @@ type Block interface {
 		results Results,
 	) (exhaustive bool, err error)
 
+	// AggregateQuery resolves the given query into aggregated tags.
+	AggregateQuery(
```
prateek: same nit about opts type

src/dbnode/storage/index/aggregate_results.go
arnikola: Yeah, this is missing tests in general at the moment; looking to verify structure of the feature in general before writing them up

src/dbnode/storage/index/types.go
arnikola: Good call, I'll give it a shot, hopefully the code gen will still be happy with it

src/dbnode/storage/index/aggregate_results.go
arnikola: I guess there's my answer :p

Was going to check with someone more versed in the code than I to see if this represents an error case

src/dbnode/storage/index/aggregate_results.go
arnikola: This is to match the flow of the regular `Query` case; when it initially matches the block, it will be added through the `AggregateDocument` path, then when the results from multiple blocks are being merged, one is selected as the merging `AggregateResult`, and the rest are iterated through and call `AddIDAndValues` on that one to merge the results

src/dbnode/generated-source-files.mk
arnikola: I think this one was github weirdness, right? Or still a valid comment?

src/dbnode/storage/index/aggregate_results.go
arnikola: So I went back and forth on this; if it's already been added to the map, that indicates it's a valid value and there's no reason to check against the term filter

I guess this should be arranged by what's more likely; failing the filter lookup, or succeeding (the current way is biased towards succeeding)

src/dbnode/storage/index/types.go
arnikola: Likely not; will remove

src/dbnode/storage/index/types.go
arnikola: Likely not; will remove

src/dbnode/storage/index.go
```diff
@@ -1436,3 +1440,230 @@ func (shards dbShards) IDs() []uint32 {
...
+		aggregateResults := i.aggregateResultsPool.Get()
+		aggregateResults.Reset(i.nsMetadata.ID())
+
+		err := block.AggregateQuery(query, opts, aggregateResults)
+		if err == index.ErrUnableToQueryBlockClosed {
+			// NB(r): Because we query this block outside of the results lock, it's
+			// possible this block may get closed if it slides out of retention, in
+			// that case those results are no longer considered valid and outside of
+			// retention regardless, so this is a non-issue.
+			err = nil
```
prateek: can you just return early here?

src/dbnode/storage/index.go
```diff
@@ -1436,3 +1440,230 @@ func (shards dbShards) IDs() []uint32 {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var (
+		deadline = start.Add(timeout)
+		wg       sync.WaitGroup
+
+		// Results contains all concurrent mutable state below.
+		results = struct {
```
prateek: can you extract this into a separate struct and reuse it across both places

arnikola: Sure, had something like that originally while trying to jam it into a combined function :)

src/dbnode/storage/index.go
```diff
@@ -1436,3 +1440,230 @@ func (shards dbShards) IDs() []uint32 {
...
+			// Return results to pool at end of request.
+			ctx.RegisterFinalizer(aggregateResults)
+			// No merged results yet, use this as the first to merge into.
+			results.merged = aggregateResults
+		} else {
+			// Append the block results.
+			mergedResult = true
+			for _, entry := range aggregateResults.Map().Iter() {
+				// Append to merged results.
+				id, tags := entry.Key(), entry.Value()
```
prateek: why is the ID needed in aggregation results?

src/dbnode/storage/index.go
```diff
@@ -1436,3 +1440,230 @@ func (shards dbShards) IDs() []uint32 {
 	}
 	return ids
 }
+
+func (i *nsIndex) AggregateQuery(
```
prateek: nit: refactor this so that `Query()` and `AggregateQuery()` share the structural components of the method.

src/dbnode/storage/index.go
arnikola: Probably a bad name; `id` here represents unique tag names

src/dbnode/storage/index/aggregate_results.go
prateek: Mainly suggesting to do it early cause the flow made more sense, early terminating and all that. Don't think it makes a difference for perf really

src/dbnode/storage/index.go
arnikola: Was trying that but it got a little messy, will take another stab at it

src/dbnode/storage/index/aggregate_results.go
prateek: hm i think you can do `AggregateDocument` or `AddAggregatedValues` (doesn't need ID in the second method)

src/dbnode/storage/index.go
arnikola: Yeah, seems to be the case

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,220 @@
...
+		// Return these values to the pool.
+		r.valuesPool.Put(aggValues)
+		return err
+	}
+
+	r.resultsMap.Set(termID, aggValues)
+	return nil
+}
+
+// func (r *aggregatedResults) AddIDAndValues(
```
arnikola: Oops, will delete this

src/dbnode/generated-source-files.mk
```diff
@@ -27,16 +27,19 @@ genny-all: genny-map-all genny-arraypool-all genny-leakcheckpool-all
...
+	genny-map-storage-bootstrap-result          \
+	genny-map-storage                           \
+	genny-map-storage-namespace-metadata        \
+	genny-map-storage-repair                    \
+	genny-map-storage-index-results             \
+	genny-map-storage-index-aggregate-values    \
+	genny-map-storage-index-aggregation-results \
 	genny-map-storage-bootstrap-bootstrapper-commitlog
 
+
```
robskillington: nit: Do you need this extra endline?

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,220 @@
...
+
+	aggValues := r.valuesPool.Get()
+	valueID := ident.BytesID(value)
+	if err := aggValues.addValue(valueID); err != nil {
+		// Return these values to the pool.
+		r.valuesPool.Put(aggValues)
+		return err
+	}
+
+	r.resultsMap.Set(termID, aggValues)
```
robskillington: You can use SetUnsafe here with `NoCopyKey: true` and `NoFinalizeKey: true` to avoid copying and finalizing the key (since it has the same lifetime as the bytes ID).

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,220 @@
...
+	termID := ident.BytesID(term)
+
+	// if this term hasn't been seen, ensure it should be included in output.
+	if !r.opts.TermFilter.Contains(termID) {
+		return nil
+	}
+
+	valueMap, found := r.resultsMap.Get(termID)
+	if found {
+		valueID := ident.BytesID(value)
```
robskillington: Hm, you can't use `BytesID` here since you need to take a copy (neither of the []byte slices are valid past the call to `AddDocuments(..)`).

You'll need the checked bytes pool here too, then:
```go
valueBytes := r.bytesPool.Get(len(value))
valueBytes.IncRef()
valueBytes.AppendAll(value)
valueBytes.DecRef()
valueID := r.idPool.BinaryID(valueBytes)
```

You can do this above the `if` branch here to use the same var later on and avoid doing this twice.

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,220 @@
...
+		return valueMap.addValue(valueID)
+	}
+
+	aggValues := r.valuesPool.Get()
+	valueID := ident.BytesID(value)
```
robskillington: As per previous comment, can't use BytesID here.

src/dbnode/storage/index/aggregate_values.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (v *AggregateValues) Size() int {
+	return v.valuesMap.Len()
+}
+
+func (v *AggregateValues) reset() {
+	// reset all values from map first
+	for _, entry := range v.valuesMap.Iter() {
+		ident := entry.Key()
```
robskillington: Hm shouldn't this be `entry.Value()`?  The `valuesMap.Reset()` will finalize the keys (as per comment below.

src/dbnode/storage/index/types.go
```diff
@@ -74,44 +74,48 @@ func (o QueryOptions) LimitExceeded(size int) bool {
 	return o.Limit > 0 && size >= o.Limit
 }
 
-// QueryResults is the collection of results for a query.
-type QueryResults struct {
+// QueryReturnResults is the collection of results for a query.
+type QueryReturnResults struct {
```
robskillington: nit: Any huge reason to change QueryResults to QueryReturnResults?  Usually we try to keep language keywords like "return" from the names of types.

src/dbnode/storage/index/types.go
```diff
@@ -74,44 +74,48 @@ func (o QueryOptions) LimitExceeded(size int) bool {
...
 	Results    Results
 	Exhaustive bool
 }
 
-// Results is a collection of results for a query, it is synchronized
+// Results is a collection of results for a generic query, it is synchronized
```
robskillington: Can we rename this to `GenericResults` or `BaseResults`, then rename `QueryResults` back to `Results` so we can avoid adding the word `"Return`" into `QueryReturnResults` (and the aggregate versions of that struct)?

src/dbnode/storage/types.go
```diff
@@ -282,7 +291,15 @@ type databaseNamespace interface {
...
 		opts index.QueryOptions,
-	) (index.QueryResults, error)
+	) (index.QueryReturnResults, error)
+
+	// AggregateQuery resolves the given query into aggregated tags.
+	AggregateQuery(
+		ctx context.Context,
+		query index.Query,
+		opts index.QueryOptions,
+		aggResultOpts *index.AggregateResultsOptions,
```
robskillington: Since this method is already a `AggregateQuery(...)` call, you definitely need to pass `aggResultOpts` so can we remove the pointer type on this?

src/dbnode/storage/types.go
```diff
@@ -476,7 +493,8 @@ type namespaceIndex interface {
 		ctx context.Context,
 		query index.Query,
 		opts index.QueryOptions,
-	) (index.QueryResults, error)
+		aggResultOpts *index.AggregateResultsOptions,
```
robskillington: Instead of making this an optional arg, can we just create a second `AggregateQuery(...)` method on `namespaceIndex` too?

src/dbnode/storage/types.go
```diff
@@ -282,7 +291,15 @@ type databaseNamespace interface {
...
-	) (index.QueryResults, error)
+	) (index.QueryReturnResults, error)
+
+	// AggregateQuery resolves the given query into aggregated tags.
+	AggregateQuery(
+		ctx context.Context,
+		query index.Query,
+		opts index.QueryOptions,
+		aggResultOpts *index.AggregateResultsOptions,
+	) (index.QueryReturnResults, error)
```
robskillington: Why don't we actually just make methods for every single type of query?  Then you can concretely return `index.AggregateResults` for `AggregateQuery(...)`.

Inside the `storage/index.go` file you can do the magic to make both of the methods actually use the same underlying code, that way callers of these methods don't need to cast `index.Results` into a concrete type (of either `index.QueryResults` or `index.AggregateResults`).

src/dbnode/storage/index/types.go
arnikola: Sure, was having a bit of trouble getting good names in as you can see :p

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,206 @@
...
+		r.valuesPool.Put(aggValues)
+		return err
+	}
+
+	termBytes := r.bytesPool.Get(len(value))
+	termBytes.IncRef()
+	termBytes.AppendAll(term)
+	termBytes.DecRef()
+	termID = r.idPool.BinaryID(termBytes)
+	r.resultsMap.Set(termID, aggValues)
```
robskillington: If you've already copied the `termID` yourself, you'll want to make sure that the map itself doesn't copy the key again, i.e.:
```go
r.resultsMap.SetUnsafe(termID, aggValues, AggregateResultsMapSetUnsafeOptions{
  NoCopyKey: true,
})
```

Or you could avoid copying yourself and revert to just using `r.resultsMap.Set(termID, aggValues)` which will do the copy itself using a bytes pool itself.

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,206 @@
...
+	// if this term hasn't been seen, ensure it should be included in output.
+	if !r.aggregateQueryOpts.TermFilter.Contains(termID) {
+		return nil
+	}
+
+	valueBytes := r.bytesPool.Get(len(value))
+	valueBytes.IncRef()
+	valueBytes.AppendAll(value)
+	valueBytes.DecRef()
+	valueID := r.idPool.BinaryID(valueBytes)
```
robskillington: Sorry, I realize you can actually use a helper function to make this code a little simpler:
```go
valueID := r.idPool.Clone(ident.BytesID(value))
```

arnikola: Nice 👍

src/dbnode/storage/index/aggregate_results.go
arnikola: Hm actually yeah might just revert to leaving it to the map to do the copies, don't think we really get anything out of manually managing it

src/dbnode/storage/index/aggregate_values.go
```diff
@@ -0,0 +1,85 @@
...
+
+func (v *AggregateValues) addValue(value ident.ID) error {
+	bytesID := ident.BytesID(value.Bytes())
+	if len(bytesID) == 0 {
+		return errUnableToAddValueMissingID
+	}
+
+	// NB: fine to overwrite the values here.
+	v.valuesMap.Set(bytesID, struct{}{})
+	return nil
```
robskillington: Can just condense this (since `.Set(...)` takes an ident.ID and will copy it) into:
```go
	if len(value.Bytes()) == 0 {
		return errUnableToAddValueMissingID
	}

 	// NB: fine to overwrite the values here.
	v.valuesMap.Set(value, struct{}{})
	return nil
```

arnikola: Yeah already have sorry, latest push was halfway consolidated :p

src/dbnode/storage/namespace.go
```diff
@@ -598,24 +598,49 @@ func (n *dbNamespace) QueryIDs(
 	ctx context.Context,
 	query index.Query,
 	opts index.QueryOptions,
-) (index.QueryResults, error) {
+) (index.QueryResult, error) {
```
prateek: +1 for rename

src/dbnode/storage/namespace.go
```diff
@@ -598,24 +598,49 @@ func (n *dbNamespace) QueryIDs(
...
+
+	if n.reverseIndex.BootstrapsDone() < 1 {
+		// Similar to reading shard data, return not bootstrapped
+		n.metrics.queryIDs.ReportError(n.nowFn().Sub(callStart))
+		return index.AggregateQueryResult{},
+			xerrors.NewRetryableError(errIndexNotBootstrappedToRead)
+	}
+
+	res, err := n.reverseIndex.AggregateQuery(ctx, query, opts, aggResultOpts)
+	n.metrics.queryIDs.ReportSuccessOrError(err, n.nowFn().Sub(callStart))
```
prateek: nit: add a metric for `aggregateQuery` (similar to `queryIDs`)

prateek: (and reuse in all calls in this function)

src/dbnode/storage/index_block_test.go
```diff
@@ -566,26 +566,41 @@ func TestNamespaceIndexBlockQuery(t *testing.T) {
 		StartInclusive: t0,
 		EndExclusive:   now.Add(time.Minute),
 	}
-	b0.EXPECT().Query(gomock.Any(), q, qOpts, gomock.Any()).Return(true, nil)
+	aggOpts := index.AggregateResultsOptions{}
```
prateek: nit: could you decouple the tests for the Query and Aggregate. One simple way - copy paste the test you have twice, and make it only do the mocks/calls for Query in one, and Aggregate in the other. That way each is independent of the other. Makes maintenance easier in the long run

src/dbnode/storage/index/aggregate_values.go
```diff
@@ -0,0 +1,85 @@
...
+func (v *AggregateValues) Map() *AggregateValuesMap {
+	return v.valuesMap
+}
+
+// Size returns the number of IDs tracked.
+func (v *AggregateValues) Size() int {
+	return v.valuesMap.Len()
+}
+
+func (v *AggregateValues) reset() {
```
prateek: more a question: is it intentional these methods are un-exported?

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,237 @@
...
+
+	return nil
+}
+
+func (r *aggregatedResults) addDocumentWithLock(
+	document doc.Document,
+) error {
+	for _, field := range document.Fields {
+		if err := r.addFieldWithLock(field.Name, field.Value); err != nil {
+			return err
```
prateek: nit: would be useful to bundle with document here too, something like: 
```return fmt.Errorf("unable to add document [%+v]: %v", document, err)```

File: src/dbnode/storage/index/aggregate_results.go [hunk 40]
```diff
@@ -0,0 +1,237 @@
...
+		return fmt.Errorf(missingDocumentFields, "value")
+	}
+
+	// NB: can cast the []byte -> ident.ID to avoid an alloc
+	// before we're sure we need it.
+	var termID ident.ID = ident.BytesID(term)
+
+	// if a term filter is provided, ensure this field matches the filter,
+	// otherwise ignore it.
+	if len(r.aggregateOpts.TermFilter) > 0 {
```
prateek: nit: mind making this a function on the type? something like 

```
type AggregateResultsOptions  struct {
  ...
  TermFilter [][]byte
}

type AggregateTermFilter [][]byte

func (a AggregateTermFilter) Filter(term []byte) bool {
  if len(a) == 0 {
    return false 
  }
  for ...
  return true
}
```

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,237 @@
...
+) error {
+	if len(term) == 0 {
+		return fmt.Errorf(missingDocumentFields, "term")
+	}
+
+	if len(value) == 0 {
+		return fmt.Errorf(missingDocumentFields, "value")
+	}
+
+	// NB: can cast the []byte -> ident.ID to avoid an alloc
```
prateek: nit: shouldn't this below the filtering code?

prateek: as in, termID isn't used until after the filtering. would be better to group along with the other vars there.

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -0,0 +1,237 @@
...
+	r.RLock()
+	l := r.resultsMap.Len()
+	r.RUnlock()
+	return l
+}
+
+func (r *aggregatedResults) Finalize() {
+	r.Lock()
+
+	r.aggregateOpts = AggregateResultsOptions{}
```
prateek: hm could you reuse `r.Reset(nil, AggregateResultsOptions{})` here? looks like they're both doing  the same thing except the pool return

src/dbnode/server/server.go
```diff
@@ -1202,8 +1202,11 @@ func withEncodingAndPoolingOptions(
 	postingsListOpts := poolOptions(policy.PostingsListPool, scope.SubScope("postingslist-pool"))
 	postingsList := postings.NewPool(postingsListOpts, roaring.NewPostingsList)
 
-	resultsPool := index.NewResultsPool(
-		poolOptions(policy.IndexResultsPool, scope.SubScope("index-results-pool")))
+	// Need to actually set pools
```
robskillington: nit: Remove this now that the pools are set?

src/dbnode/storage/index/aggregate_results.go
robskillington: Hm, seems aggressive for now, we could potentially consolidate this later to unblock find endpoint working?

src/dbnode/storage/index/aggregate_values_pool.go
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
robskillington: nit: 2019.

src/dbnode/storage/index/aggregate_results_pool.go
```diff
@@ -0,0 +1,47 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
robskillington: nit: 2019.

src/dbnode/storage/index/aggregate_results.go
prateek: Sorry, not sure I follow. I was suggesting changing the implementation of Finalize to the following:
```
func (r *aggregatedResults) Finalize() {
  r.Reset(nil, AggregateResultsOptions{})
   if r.pool == nil {
     return
  }
  r.pool.Put(r)
}
```

src/dbnode/storage/index/aggregate_values.go
arnikola: Yeah, wanted to restrict these methods to this package so callers don't accidentally close the AggregateValues

src/dbnode/storage/namespace.go
arnikola: Nice one :+1:

src/dbnode/storage/index_block_test.go
arnikola: Yeah, for sure. I didn't want to go ahead and c/p larger tests so figured would be fine to package them; will refactor to use a test setup method

src/dbnode/server/server.go
arnikola: Good call

src/dbnode/storage/index/aggregate_results.go
arnikola: Hey, so this method is a little different in that it needs to call `finalize` on the `AggregatedValues` map; I tried refactoring it so they'd both call a common function with a flag to determine if it should finalize or reset, but it ended up  being a bit awkward so preferred this approach

src/dbnode/storage/index/aggregate_results.go
arnikola: Ah yeah good one; the filter function previously required a `ident.ID` so had to define this earlier, but missed changing it after refactor

src/dbnode/storage/index/aggregate_results.go
arnikola: SGTM, guess then we can use it elsewhere :)

src/dbnode/storage/index/aggregate_results.go
arnikola: SGTM

src/dbnode/storage/index_block_test.go
arnikola: Tried with a setup method and it looked really bad; did the test twice instead

src/dbnode/storage/index/aggregate_results.go
prateek: talked offline: made sense to merge the behaviours

src/dbnode/storage/index/aggregate_results_test.go
```diff
@@ -0,0 +1,349 @@
...
+		"foo":  []string{"bar", "biz", "baz"},
+		"fizz": []string{"bar"},
+		"buzz": []string{"bar", "bag"},
+		"qux":  []string{"qaz"},
+	}
+
+	contains(t, expected, res.Map())
+}
+
+func toFilter(strs ...string) AggregateTermFilter {
```
prateek: nice

src/dbnode/storage/index/aggregate_results_test.go
```diff
@@ -0,0 +1,349 @@
...
+	require.Equal(t, 1, size)
+
+	aggVals, ok = rMap.Get(ident.StringID("foo"))
+	require.True(t, ok)
+	require.Equal(t, 2, aggVals.Size())
+	assert.True(t, aggVals.Map().Contains(ident.StringID("bar")))
+	assert.True(t, aggVals.Map().Contains(ident.StringID("biz")))
+}
+
+func contains(t *testing.T, ex map[string][]string, ac *AggregateResultsMap) {
```
prateek: nit: `assertContains()` would be clearer

src/dbnode/storage/index/aggregate_results_test.go
```diff
@@ -0,0 +1,349 @@
...
+	dValid := genDoc("foo", "bar")
+	name := dValid.Fields[0].Name
+	value := dValid.Fields[0].Value
+	size, err := res.AddDocuments([]doc.Document{dValid})
+	require.NoError(t, err)
+	require.Equal(t, 1, size)
+
+	found := false
+
+	// our genny generated maps don't provide access to MapEntry directly,
```
prateek: +1, v nice test.

src/dbnode/storage/index/types.go
```diff
@@ -74,71 +74,148 @@ func (o QueryOptions) LimitExceeded(size int) bool {
...
+// AggregateQueryResult is the collection of results for an aggregate query.
+type AggregateQueryResult struct {
+	Results    AggregateResults
+	Exhaustive bool
+}
 
+// BaseResults is a collection of basic results for a generic query, it is
+// synchronized when access to the results set is used as documented by the
+// methods.
+type BaseResults interface {
```
prateek: +1, nice

PR #1464, opened 2019-03-15
Add query config docs

benraskin92:


docs/query_engine/config/annotated_config.md
```diff
@@ -0,0 +1,3 @@
+# Annotated Config
```
arnikola: Would rather have this under an existing heading rather than a new page for one line; alternatively add the annotated config here

benraskin92: Where do you suggest?

docs/query_engine/config/annotated_config.yaml
```diff
@@ -0,0 +1,84 @@
...
+  scope:
+    prefix: "coordinator"
+  prometheus:
+    handlerPath: /metrics
+    listenAddress: 0.0.0.0:7203 # until https://github.com/m3db/m3/issues/682 is resolved
+  sanitization: prometheus
+  samplingRate: 1.0
+  extended: none
+
+clusters:
```
arnikola: Bunch of these aren't annotated!

benraskin92: I'm not doing everything! But yeah, I wanted to get this in so that @andrewmains12 can add his stuff.

docs/query_engine/config/annotated_config.md
benraskin92: I think we should keep this here. I agree it's not the prettiest thing, but I think it makes sense.

arnikola: I'd rather just have the config straight up written here

I think you can do something  like this to format it well:
`/``yaml
config:
 blah blah blah
`/``

benraskin92: Yeah it just doesn't look as good. @andrewmains12 tiebreaker?

benraskin92: You don't get any of the coloring I don't think

benraskin92: Screen Shot 2019-03-18 at 3 29 32 PM

Just looks like this on our docs..

andrewmains12: Personally I prefer the approach @benraskin92 has here; having it in a `.yaml` makes it easy read + edit (yes, I know you can change the file association etc, but it's a pain). If we decide it's bothersome later we can always change it back.

PR #1472, opened 2019-03-18
Adding metrics for leaders and followers with active shards

nerd0:
This will add two new metrics to track leaders and followers _with_ active shards to ensure. This will enable the ability to query leaders and followers that are active and not disregard new instances that may be joining or leaving the placement. 

An alternative would be perform two fetches piping into each other, however makes the query a bit ugly. 
```
leaderCount =fetch env:production name:aggregator.election-manager.election-state| eq 3| isNonNull | fetch env:production name:aggregator.election-manager.campaign-check.has-active-shards|gt 0| sum;followerCount =fetch env:production name:aggregator.election-manager.election-state| transformNull 99| le 2| isNonNull| fetch env:production name:aggregator.election-manager.campaign-check.has-active-shards|gt 0| sum;leaderCount| diff (followerCount)| absolute
```


src/aggregator/aggregator/election_mgr.go
```diff
@@ -651,6 +655,12 @@ func (mgr *electionManager) campaignIsEnabled() (bool, error) {
 		hasNotCutoff := nowNanos < shard.CutoffNanos()-int64(mgr.shardCutoffCheckOffset)
 		if hasCutover && hasNotCutoff {
 			mgr.metrics.campaignCheckHasActiveShards.Inc(1)
+			if mgr.ElectionState() == LeaderState {
+				mgr.metrics.leadersWithActiveShards.Update(float64(1))
```
richardartoul: I'm kinda confused about this metric. So if the host is the leader for ANY shards, you want to set this entire gauge to 1? Also the name is a little confusing consider the gauge value will only apply to a single host

nerd0: The PR tries to emit a metric for both leaders and followers that have active shards so that alerts can be configured to only check for a mismatch of leaders and non-leaders that do not include newly added hosts that are bootstrapping. 

I agree that gauge is a bit awkward but wasn't sure how best to express this type of metric. 

Should I just use the query that is within the description of the PR as a way to calculate this type of metric?

cw9:
LGTM, can you double check what is the value for `campaignStateCheckInterval`? Just want to make sure this gauge is being updated frequent enough

nerd0:
> LGTM, can you double check what is the value for `campaignStateCheckInterval`? Just want to make sure this gauge is being updated frequent enough

It appears that the [default configuration](https://github.com/m3db/m3/blob/master/src/aggregator/config/m3aggregator.yml#L212)  has it set to `1s`

PR #1476, opened 2019-03-19
Fix annotation pooling

richardartoul:
Fixes a pooling bug that was caused by returning annotations to the apache thrift byte pool once RPCs had completed. This was problematic because the commit log still had a reference to them, as well as the shard insert queue if we were inserting a new series asynchronously.


src/dbnode/ts/write_batch.go
```diff
@@ -102,7 +107,22 @@ func (b *writeBatch) SetSkipWrite(idx int) {
...
+		for _, write := range b.writes {
+			annotation := write.Write.Annotation
+			if annotation == nil {
+				continue
+			}
+
+			b.finalizeAnnotationFn(annotation)
+		}
+	}
+
```
prateek: nit: do `finalizeAnnotationFn = nil` after this line

src/dbnode/ts/write_batch.go
```diff
@@ -102,7 +107,22 @@ func (b *writeBatch) SetSkipWrite(idx int) {
...
 	b.ns = nil
 	b.writes = b.writes[:0]
```
prateek: unrelated to your changes, but this needs to nil set the writes too or it's leaking. mind adding the following while you're here –
```
var zeroedWrite BatchWrite
for i := range b.writes {
  b.writes[i] = zeroedWrite
}
```

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -1298,23 +1305,23 @@ func (r *writeBatchPooledReq) Finalize() {
...
 
-	// Return any pooled thrift byte slices to the thrift pool
+	// Return any pooled thrift byte slices to the thrift pool.
 	if r.writeReq != nil {
 		for _, elem := range r.writeReq.Elements {
 			apachethrift.BytesPoolPut(elem.ID)
-			if elem.Datapoint.Annotation != nil {
-				apachethrift.BytesPoolPut(elem.Datapoint.Annotation)
-			}
+			// Don't return the annotations to the pool because they don't
```
prateek: the first sentence in the comment is confusing unless people understand nuances of when IDs are cloned v when retained. much simpler to just rely on the second sentence and say the ownership of the lifecycle for the annotations has been transferred to the `BatchWriter`.

that said, can we do the same thing for the ID/EncodedTags too? i.e. expect the BatchWriter to clean it up? might cause a minor rise in steady state memory usage but would simplify this code a lot.

src/dbnode/storage/shard.go
```diff
@@ -842,14 +844,21 @@ func (s *dbShard) writeAndIndex(
 			return ts.Series{}, false, err
 		}
 	} else {
-		// This is an asynchronous insert and write
+		// This is an asynchronous insert and write which means we need to clone the annotation
+		// because it could get returned to a pool at any point after this function returns.
```
prateek: nit: please rephrase `because it could get returned to a pool at any point after this function returns.` to something like `because its lifecycle in the commit log is independent of the calling function`

src/dbnode/ts/types.go
```diff
@@ -127,4 +131,6 @@ type BatchWriter interface {
 		unit xtime.Unit,
 		annotation []byte,
 	)
+
+	SetFinalizeAnnotationFn(f FinalizeAnnotationFn)
```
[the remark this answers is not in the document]
prateek: talked offline: going to leave as is and add a comment explaining the choice

src/dbnode/storage/shard.go
```diff
@@ -842,14 +844,21 @@ func (s *dbShard) writeAndIndex(
...
+		annotationClone := s.opts.BytesPool().Get(len(annotation))
```
prateek: you don't need to always do this. can you put it behind a nil check. i.e. 

```
var annotationClone checked.Bytes
if len(annotation) != 0 {
// ...
}
```

src/dbnode/storage/shard.go
```diff
@@ -1289,10 +1298,16 @@ func (s *dbShard) insertSeriesBatch(inserts []dbShardInsert) error {
 			// TODO: Consider propagating the `wasWritten` argument back to the caller
 			// using waitgroup (or otherwise) in the future.
 			_, err := entry.Series.Write(ctx, write.timestamp, write.value,
-				write.unit, write.annotation)
+				write.unit, write.annotation.Bytes())
 			if err != nil {
 				s.metrics.insertAsyncWriteErrors.Inc(1)
 			}
+
+			// Now that we've performed the write, we can finalize the annotation because
```
prateek: same comment as above, can avoid the fn calls unless `annotation != nil`

src/dbnode/storage/shard.go
prateek: maybe make it a helper function (this method is long enough as is)

src/dbnode/ts/types.go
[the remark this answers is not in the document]
richardartoul: Spoke offline and agreed to leave as is since the lifecycle is already confusing enough without making it more generic

src/dbnode/network/server/tchannelthrift/node/service.go
richardartoul: Fixed the comment. Spoke offline and decided not to do the ID/EncodedTags thing for performance reasons (more synchronization required by single-threaded commitlog)

src/dbnode/storage/shard.go
richardartoul: thanks I had a mental todo for this but forgot

src/dbnode/storage/shard.go
```diff
@@ -781,9 +782,9 @@ func (s *dbShard) writeAndIndex(
 
 	writable := entry != nil
 
-	// If no entry and we are not writing new series asynchronously
+	// If no entry and we are not writing new series asynchronously.
```
prateek: nit: mind adding tests for the `[]byte` lifecycle as handled by the shard

src/dbnode/storage/shard_test.go
```diff
@@ -493,7 +495,61 @@ func TestShardTick(t *testing.T) {
...
+		{
+			id:    "baz",
+			value: 3.0,
+			unit:  xtime.Second,
+		},
+	})
+}
+
+func TestShardWriteAsyncWithAnnotations(t *testing.T) {
+	testShardWriteAsync(t, []testWrite{
```
prateek: one more test case - ensure we don't copy the annotation bytes if the series already exists in the shard

PR #1478, opened 2019-03-20
Add m3ninx/AllQuery

prateek:


src/m3ninx/search/query/all.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"fmt"
+
+	"github.com/m3db/m3/src/m3ninx/generated/proto/querypb"
+	"github.com/m3db/m3/src/m3ninx/search"
+	"github.com/m3db/m3/src/m3ninx/search/searcher"
+)
+
+// AllQuery returns a query which matches all the known documents.
```
arnikola: nit: `all known documents`

src/m3ninx/search/query/codec.go
```diff
@@ -85,8 +88,7 @@ func unmarshal(q *querypb.Query) (search.Query, error) {
 			qs = append(qs, sqry)
 		}
 		return NewDisjunctionQuery(qs), nil
-
 	}
 
-	return nil, fmt.Errorf("unknown query: %v", q)
+	return nil, fmt.Errorf("unknown query: %+v", q)
```
arnikola: Wow, didn't know about this, it's pretty neat :+1:

arnikola:
> Whats the point of this P.R? Is it just so we have an easy way of special-casing these types of queries?

I think it's primarily to make an explicit query type for tag matching v.s. regex matching `.*` or something; in future, could be useful for specifying the special case of listing all tags without providing a query

src/m3ninx/search/query/codec.go
prateek: lol :)

src/m3ninx/search/query/all.go
prateek: cheers

prateek:
Pretty much what you guys are saying - need an explicit way for users of `m3db/client` to be able to specify the lack of any filtering for aggregation queries

PR #1479, opened 2019-03-20
Allow non valid zones for instances to be add to a placement

nerd0:
There are certain scenarios were instances within a placement will not be apart of the same zone of the placement. This PR relaxes the constraint a bit by allowing a `placement.Option` to bypass the strict check.


src/cluster/placement/options.go
```diff
@@ -68,6 +71,7 @@ type options struct {
 	validateFn          ValidateFn
 	nowFn               clock.NowFn
 	allowPartialReplace bool
+	allowNonValidZones  bool
```
nerd0: Not a fan of the name, perhaps `allowAnyZone` ?

arnikola: 'allowAllZones` maybe?

src/cluster/placement/selector/common.go
```diff
@@ -67,16 +71,18 @@ func filterZones(
 	}
 
 	var validZone string
+	var allowNonValidZone bool
 	if opts != nil {
 		validZone = opts.ValidZone()
+		allowNonValidZone = opts.AllowNonValidZones()
```
arnikola: I think you can just do 
```golang
// ...
if opts.AllowNonValidZones() {
  return candidates
}
```

src/cluster/placement/options.go
nerd0: Good call.

src/cluster/placement/selector/common.go
```diff
@@ -67,16 +71,18 @@ func filterZones(
...
 	}
 	if validZone == "" && len(p.Instances()) > 0 {
 		validZone = p.Instances()[0].Zone()
 	}
 
 	validInstances := make([]placement.Instance, 0, len(candidates))
 	for _, instance := range candidates {
-		if validZone == instance.Zone() {
+		if validZone == instance.Zone() || allowNonValidZone {
```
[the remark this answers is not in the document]
nerd0: Good point but I think your suggestion below is a bit more streamlined.

src/cluster/placement/selector/common.go
nerd0: Great idea, thanks for the feedback

src/cluster/placement/selector/common.go
```diff
@@ -46,6 +46,10 @@ func getValidCandidates(
 		}
 		if instanceInPlacement.IsLeaving() {
 			instances = append(instances, instanceInPlacement)
+			continue
+		}
+		if opts.ShardStateMode() == placement.StableShardStateOnly {
```
jeromefroe: Talked offline, going to remove this check for now until we have a more concrete use case for it.

PR #1481, opened 2019-03-20
[query] Plumb potential tag completion/aggregate through to m3query endpoints

arnikola:
Adds new aggregate endpoint to Query


scripts/development/m3_stack/start_m3.sh
```diff
@@ -10,7 +10,7 @@ if [[ "$FORCE_BUILD" = true ]] ; then
 fi
 
 echo "Bringing up nodes in the background with docker compose, remember to run ./stop.sh when done"
-docker-compose -f docker-compose.yml up $DOCKER_ARGS m3coordinator01
+docker-compose -f docker-compose.yml up --build $DOCKER_ARGS m3coordinator01
```
andrewmains12: Intentional to have this change in here?

arnikola: Oops

prateek:
@arnikola I have a branch with some of these types defined too. Lemme put it up and we can merge the best features

prateek:
@arnikola take a look at src/dbnode/client/types.go and  src/dbnode/storage/index/types.go in https://github.com/m3db/m3/pull/1483

src/dbnode/storage/index/types.go
```diff
@@ -60,6 +61,13 @@ type Query struct {
 	idx.Query
 }
 
+type AggregateQueryOptions struct {
+	QueryOptions
```
prateek: better to not embed the type here

src/dbnode/storage/index/types.go
```diff
@@ -60,6 +61,13 @@ type Query struct {
...
+
+	TagNameFilter      [][]byte
```
prateek: would be more flexible to inject a lambda here

src/dbnode/storage/index/types.go
```diff
@@ -60,6 +61,13 @@ type Query struct {
...
+	AggregateQueryType rpc.AggregateQueryType
```
prateek: Not a fan of using rpc generated types in the exported interface.

src/dbnode/client/types.go
```diff
@@ -78,6 +78,10 @@ type Session interface {
 	// FetchTaggedIDs resolves the provided query to known IDs.
 	FetchTaggedIDs(namespace ident.ID, q index.Query, opts index.QueryOptions) (iter TaggedIDsIterator, exhaustive bool, err error)
 
+	// FetchTags resolves the optionally provided query to any known tag matchers.
+	// NB: this is a provisional function until the `session` contract is fully defined
+	FetchTags(namespace ident.ID, q index.Query, opts index.AggregateQueryOptions) (tags FetchedTags, exhaustive bool, err error)
```
prateek: How would you feel about `Aggregate` instead?

prateek: also, re: the return type - would be better to use an Iterator - allows us to avoid a bunch of allocs

arnikola: Sounds good, I like the iterator you added in #1483; re: `Aggregate` where do you think we change the context of the function from being a generic aggregation vs a tag completion; would that be entirely in query? Maybe meet halfway with `AggregateTags` 😛 ? We've already got tag specific stuff in this interface so wouldn't be a huge change, right?

src/dbnode/storage/index/types.go
arnikola: Yeah, agreed; was just throwing up something quick and dirty

src/query/api/v1/handler/graphite/find.go
```diff
@@ -57,52 +57,75 @@ func NewFindHandler(
...
+	for _, tag := range noChildren.CompletedTags {
+		for _, value := range tag.Values {
+			tagMap[string(value)] = false
+		}
+	}
+
+	// NB: fine to overwrite any tags which were present in the `noChildren` map
+	for _, tag := range withChildren.CompletedTags {
+		for _, value := range tag.Values {
+			tagMap[string(value)] = true
```
benraskin92: can you just create a set here instead of the true/false thing? I think you can just do something like:
```
tagMap := make(map[string]struct{}, mapLength)
```
then just set it to an empty struct when you're looping through. 
```
		for _, value := range tag.Values {
			tagMap[string(value)] = struct{}{}
		}
```

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -404,13 +362,33 @@ func RenderSeriesMatchResultsJSON(
 	jw.WriteString("success")
 
 	jw.BeginObjectField("data")
-	writeTags(jw, results)
+	jw.BeginArray()
+
+	for _, result := range results {
+		for _, tags := range result {
+			jw.BeginObject()
+			for _, tag := range tags.Tags.Tags {
```
benraskin92: `tags.Tags.Tags` tags on tags on tags - can we possibly come up with a better naming convention?

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -404,13 +362,33 @@ func RenderSeriesMatchResultsJSON(
...
+					// `"role":"remote"` tag is added, which should not be included in the
+					// results.
+					continue
+				}
+				jw.BeginObjectField(string(tag.Name))
+				jw.WriteString(string(tag.Value))
+			}
+			jw.EndObject()
+		}
+
```
benraskin92: remove newline

src/query/api/v1/handler/graphite/find_parser.go
```diff
@@ -24,71 +24,56 @@ import (
...
 
-	until, err := graphite.ParseTime(
-		untilString,
-		now,
-		tzOffsetForAbsoluteTime,
-	)
+	// NB: Filter will always be the second last term in the matchers, and the
+	// matchers should always have a length of at least 2 (term + terminator)
+	// so this is a sanity check and unexpected in actual execution.
+	filter := [][]byte{matchers[len(matchers)-2].Name}
```
benraskin92: Should you check to make sure `len(matchers) > 2`?

src/query/api/v1/handler/graphite/find_parser.go
```diff
@@ -24,71 +24,56 @@ import (
...
-
-	matchers, err := graphiteStorage.TranslateQueryToMatchers(query)
-	if err != nil {
-		return nil, xhttp.NewParseError(fmt.Errorf("invalid 'query': %s", query),
-			http.StatusBadRequest)
+	clonedMatchers := make([]models.Matcher, len(matchers))
+	copy(clonedMatchers, matchers)
+	// NB: change terminator from `MatchNotRegexp` to `MatchRegexp` to ensure
+	// segments with children are matched.
+	clonedMatchers[len(clonedMatchers)-1].Type = models.MatchRegexp
```
benraskin92: Same here for `len(clonedMatchers)`

src/query/api/v1/handler/graphite/find.go
arnikola: No, I need to see if the given tag has children here so need the bool

src/query/api/v1/handler/prometheus/common.go
arnikola: it's also a list of tags ;)

Agree we could do with a refactor but out of scope here

src/query/api/v1/handler/graphite/find_parser.go
```diff
@@ -24,71 +24,56 @@ import (
...
-	*storage.FetchQuery,
+func parseFindParamsToQueries(r *http.Request) (
+	*storage.CompleteTagsQuery,
+	*storage.CompleteTagsQuery,
+	string,
 	*xhttp.ParseError,
 ) {
 	values := r.URL.Query()
-	now := time.Now()
-	fromString, untilString := r.FormValue("from"), r.FormValue("until")
```
benraskin92: Why are we getting rid of the from and until? I thought the new index supports that.

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -346,56 +349,11 @@ func RenderTagValuesResultsJSON(
...
-	}
-}
-
-// RenderSeriesMatchResultsJSON renders series match results to json format
+// RenderSeriesMatchResultsJSON renders series match results to json format.
 func RenderSeriesMatchResultsJSON(
 	w io.Writer,
-	results []*storage.CompleteTagsResult,
+	results []models.Metrics,
+	dropRole bool,
```
benraskin92: Where is this coming from? Config?

src/query/api/v1/handler/prometheus/remote/match.go
```diff
@@ -65,37 +65,30 @@ func (h *PromSeriesMatchHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
 
-		results[i] = result
+		results[i] = result.Metrics
 	}
 
 	// TODO: Support multiple result types
-	if renderErr := prometheus.RenderSeriesMatchResultsJSON(w, results); renderErr != nil {
-		logger.Error("unable to write matched series", zap.Error(renderErr))
-		xhttp.Error(w, renderErr, http.StatusBadRequest)
+	if err := prometheus.RenderSeriesMatchResultsJSON(w, results, true); err != nil {
```
benraskin92: I don't think we should drop any tags from the results unless that's considered default for Prom. Especially if there's no way for a user to control this.

src/query/storage/index.go
```diff
@@ -136,26 +163,39 @@ func queryKey(m models.Matchers) []byte {
...
+func FetchQueryToM3Query(
+	fetchQuery *FetchQuery,
+	cache *QueryConversionCache,
+) (index.Query, error) {
 	matchers := fetchQuery.TagMatchers
-	k := queryKey(matchers)
+	// If no matchers provided, explicitly set this to an AllQuery
+	if len(matchers) == 0 {
+		return index.Query{
+			// TODO: change this to an idx.AllQuery: https://github.com/m3db/m3/pull/1478
```
benraskin92: This returns every id if no matchers are provided? That seems not safe.

src/query/api/v1/handler/graphite/find_parser.go
arnikola: Good one, for some reason decided these are *always* from `time.Time{}` until `time.Now()`, rather than by default (I think because we drop em later which was wrong too)

src/query/api/v1/handler/graphite/find_parser.go
arnikola: This will necessarily be >2 considering it's cloned from above

src/query/api/v1/handler/graphite/find_parser.go
arnikola: :| yeah so wrote the comment intending to do the check but ended up forgetting to haha

src/query/api/v1/handler/prometheus/common.go
arnikola: Just auto-enabled for all prom requests, and auto-disabled for graphite.

src/query/api/v1/handler/prometheus/remote/match.go
arnikola: Yeah, reluctantly agree... It's a weird tag that's tacked on by prom but will keep it around for the time being

src/query/storage/index.go
arnikola: To expound:

If not providing any queries, we want to return the set of all possible tags, and all values that each of those tags can have. We'll be building functionality for this in the index which should actually end up more performant than the new tag completion endpoint (perhaps counterintuitively, but we can get huge gains here when we're not required to filter down to satisfy an arbitrary query)

benraskin92: Makes sense

src/query/api/v1/handler/prometheus/remote/match.go
benraskin92: 👍

src/query/api/v1/handler/prometheus/common.go
benraskin92: Okay cool - maybe add a todo to config it. But at this point maybe removing some config stuff is the way to go..

src/query/api/v1/handler/graphite/find_parser.go
benraskin92: Yeah I noticed that after I wrote this but forgot to delete..

src/query/api/v1/handler/graphite/find_parser.go
benraskin92: Coolio

src/query/api/v1/handler/graphite/find_parser.go
```diff
@@ -49,46 +57,64 @@ func parseFindParamsToQuery(r *http.Request) (
 		untilString = "now"
 	}
 
-	from, err := graphite.ParseTime(
+	end, err := graphite.ParseTime(
```
benraskin92: Shouldn't this be start? And the other be end?

benraskin92: nit: maybe just leave it as `from` and `until` to be consistent with the URL parameter names.

arnikola: Good catch; added it to the test.

src/query/storage/m3/storage.go
```diff
@@ -329,58 +329,103 @@ func (s *m3storage) CompleteTags(
...
+		accumulatedTags = storage.NewCompleteTagsResultBuilder(query.CompleteNameOnly)
+		multiErr        syncMultiErrs
+		wg              sync.WaitGroup
+	)
+
+	if len(namespaces) == 0 {
+		return nil, errNoNamespacesConfigured
+	}
+
+	// TODO: (arnikola): implement a better lifecycle for incoming aggregated tag
```
prateek: do you want to do this in this PR? 

super nit: `TODO(arnikola): ` instead of `TODO: (arnikola)`

src/query/storage/m3/storage.go
```diff
@@ -329,58 +329,103 @@ func (s *m3storage) CompleteTags(
...
+	// iterators.
+	var mu sync.Mutex
+	aggIterators := make([]client.AggregatedTagsIterator, 0, initSize)
+	defer func() {
+		for _, it := range aggIterators {
```
prateek: don't you need to lock here?

src/query/storage/m3/storage.go
```diff
@@ -329,58 +329,103 @@ func (s *m3storage) CompleteTags(
...
+			mu.Lock()
+			aggIterators = append(aggIterators, aggTagIter)
+			mu.Unlock()
+
+			completedTags := make([]storage.CompletedTag, aggTagIter.Remaining())
+			for i := 0; aggTagIter.Next(); i++ {
+				name, values := aggTagIter.Current()
+				tagValues := make([][]byte, values.Remaining())
+				for j := 0; values.Next(); j++ {
+					// TODO: (arnikola) fix the lifecycle here so we don't have to clone
```
prateek: this feels like a redundant copy. you're making a local copy into `storage.CompletedTag` and then convert to `models.Metrics` above. can you do both in a single pass?

src/query/storage/index.go
```diff
@@ -136,26 +161,39 @@ func queryKey(m models.Matchers) []byte {
...
```
prateek: is there something left server side before we can do this here?

arnikola: Nope, wrote this before AllQuery was pushed and forgot to update it :+1:

src/query/api/v1/handler/graphite/find_parser.go
```diff
@@ -29,16 +29,24 @@ import (
...
 	graphiteStorage "github.com/m3db/m3/src/query/graphite/storage"
+	"github.com/m3db/m3/src/query/models"
 	"github.com/m3db/m3/src/query/storage"
 	"github.com/m3db/m3/src/query/util/json"
 	"github.com/m3db/m3/src/x/net/http"
 )
 
-func parseFindParamsToQuery(r *http.Request) (
-	*storage.FetchQuery,
+func parseFindParamsToQueries(r *http.Request) (
```
prateek: nit: could you add a comment w/ an example to explain expected input/output

src/query/api/v1/handler/graphite/find.go
```diff
@@ -57,52 +57,75 @@ func NewFindHandler(
...
-	query, rErr := parseFindParamsToQuery(r)
+	noChildrenQuery, childrenQuery, raw, rErr := parseFindParamsToQueries(r)
 	if rErr != nil {
 		xhttp.Error(w, rErr.Inner(), rErr.Code())
 		return
 	}
 
 	opts := storage.NewFetchOptions()
-	result, err := h.storage.FetchTags(ctx, query, opts)
+	noChildrenResult, err := h.storage.CompleteTags(ctx, noChildrenQuery, opts)
```
prateek: sorry, no context - what's the difference between `noChildrenResult` and `childrenResult`

src/query/storage/m3/storage.go
arnikola: Yeah, looking at the tag aggregator again looks like we cast to string, so the copy is unnecessary. Likewise the lifecycle is a bit less janky then I was expecting; there's definitely work to do on pooling/lifecycle management in m3query but that's pretty universally true rather than just in this case

src/query/api/v1/handler/graphite/find.go
arnikola: Probably easiest with an example:

Query is: `query=foo.*`

This will actually call two Aggregate queries on the session:
`foo.*.*`, translated(-ish) to: `[ { __g0__ =~ foo } , { __g1__ =~ .* }, { __g2__ =~ .* } ]`, which is `withChildrenResult`
 and `foo.*`, translated(-ish) to: `[ { __g0__ =~ foo } , { __g1__ =~ .* }, { __g2__ !=~ .* } ]`, which is `noChildrenResult`
Where ` =~ ` is the regex operator, and ` !=~ ` is the regex exclusion operator.

Then these two are combined to determine which fields have children, and which are leaf nodes.

Hopefully this makes sense, will add a comment to this effect

src/query/storage/m3/storage.go
arnikola: I do indeed :+1:

src/query/api/v1/handler/graphite/find.go
prateek: hm using your example of `query=foo.*`, why not just translate into a single aggregate query as follows: ```[ { __g0__ =~ foo } , { __g1__ =~ .* }, { __g2__ =~ .* } ]```

it's exactly the same as the union currently done in a single pass afaict

prateek: actually, do you even need to use a regex operator for `__g0__`? that can be a straight up term operator

src/query/storage/m3/storage.go
arnikola: I think there's some low hanging fruit gains for query in general, but probably best not to add them here and risk weird pooling issues for now

src/query/api/v1/handler/graphite/find.go
arnikola: Good idea; added an exact matcher case for alphanumeric segments but will look into properly detecting regex-able statements

src/query/api/v1/handler/graphite/find.go
```diff
@@ -57,52 +59,100 @@ func NewFindHandler(
...
-				index = i
-				break
-			}
+	wg.Add(2)
+	go func() {
+		defer wg.Done()
+		var err error
+		terminatedResult, err = h.storage.CompleteTags(ctx, terminatedQuery, opts)
+		if err != nil {
+			multiErr = multiErr.Add(err)
```
robskillington: nit: Need to do locking on this `multiErr`.

Instead perhaps just keep both separate then use a `xerrors.FirstError(terminatedErr, childErr)` instead of `multiErr.FinalError()`?

src/query/graphite/storage/converter.go
```diff
@@ -58,10 +59,25 @@ func glob(metric string) []byte {
 }
 
 func convertMetricPartToMatcher(count int, metric string) models.Matcher {
+	var (
+		matchType models.MatchType
+		value     []byte
+	)
+
+	// NB: small optimization to use exact matcher rather than regex if only
+	// alphanumeric characters present.
```
robskillington: Wouldn't it make more sense to do always try to do a glob and make glob return `([]byte, bool)` and if it returns `false` to imply that it didn't need to turn the glob into a regexp then you can use a MatchEqual matcher?

I'm pretty sure that's how we used to do it (and helps in the case you use a non alphanumeric symbol but turns out that it's still not a regexp? (i.e. using the `-` or `_` character).

src/query/storage/index.go
```diff
@@ -136,26 +161,38 @@ func queryKey(m models.Matchers) []byte {
...
-	defer cache.mu.Unlock()
+	k := queryKey(matchers)
+	cache.RLock()
 
 	if val, ok := cache.get(k); ok {
+		cache.RUnlock()
 		return index.Query{Query: val}, nil
 	}
 
+	cache.RUnlock()
```
robskillington: Can you simplify the above to not have to do two RUnlocks?

```go
k := queryKey(matchers)
cache.RLock()
val, ok := cache.get(k)
cache.RUnlock()

if ok { 
  return index.Query{Query: val}, nil
}

//....
```

src/query/storage/m3/storage.go
```diff
@@ -329,58 +329,94 @@ func (s *m3storage) CompleteTags(
...
+		multiErr        syncMultiErrs
+		wg              sync.WaitGroup
+	)
+
+	if len(namespaces) == 0 {
+		return nil, errNoNamespacesConfigured
+	}
+
+	var mu sync.Mutex
+	aggIterators := make([]client.AggregatedTagsIterator, 0, initSize)
```
robskillington: Why don't you make this `len(namespaces)` since you get back one for each namespace?

src/query/storage/m3/storage.go
```diff
@@ -329,58 +329,94 @@ func (s *m3storage) CompleteTags(
...
+	defer func() {
+		mu.Lock()
+		for _, it := range aggIterators {
+			it.Finalize()
```
robskillington: Will this actually finalize all bytes from all the results that came back?  If so, we need to copy them all below (right now we're just taking ref's to the bytes).

src/query/storage/m3/storage.go
```diff
@@ -329,58 +329,94 @@ func (s *m3storage) CompleteTags(
...
+			mu.Lock()
+			aggIterators = append(aggIterators, aggTagIter)
+			mu.Unlock()
+
+			completedTags := make([]storage.CompletedTag, aggTagIter.Remaining())
+			for i := 0; aggTagIter.Next(); i++ {
+				name, values := aggTagIter.Current()
+				tagValues := make([][]byte, values.Remaining())
+				for j := 0; values.Next(); j++ {
+					tagValues[j] = values.Current().Bytes()
```
robskillington: I think this needs to be a copy of the bytes if we're finalizing these results no?

src/query/storage/m3/storage.go
```diff
@@ -329,58 +329,94 @@ func (s *m3storage) CompleteTags(
...
+
+				if err := values.Err(); err != nil {
+					multiErr.add(err)
+					return
 				}
 
-				if !found {
-					continue
+				completedTags[i] = storage.CompletedTag{
+					Name:   name.Bytes(),
```
robskillington: I think this needs to be a copy of the bytes if we're finalizing these results no?

src/query/api/v1/handler/graphite/find.go
arnikola: Ah my bad, we have another multiErr type somewhere which is locking; I like the FirstError suggestion more anyway :+1:

src/query/storage/m3/storage.go
arnikola: Nice, good call

src/query/storage/m3/storage.go
arnikola: I initially was cloning the returned bytes, but if you look into how [completed tags builder](https://github.com/m3db/m3/blob/807a44a0638139b33e0403cb2ff2e79e36740890/src/query/storage/completed_tags.go#L132) works, it should already copy the bytes when taking a string

It's a bit janky but should hold up

src/query/storage/m3/storage.go
arnikola: Same deal here where the accumulator is casting to strings; I'll put up a test to double check though

src/query/storage/m3/storage.go
arnikola: Same deal

src/query/storage/index.go
arnikola: Sure, looks better

src/query/api/v1/handler/graphite/find.go
```diff
@@ -57,52 +59,94 @@ func NewFindHandler(
...
 	}
 
-	opts := storage.NewFetchOptions()
-	result, err := h.storage.FetchTags(ctx, query, opts)
-	if err != nil {
+	var (
+		terminatedResult *storage.CompleteTagsResult
+		tErr             error
+		childResult      *storage.CompleteTagsResult
+		cErr             error
```
robskillington: nit: You could make this (to slightly compact):
```go
var (
  terminateResult, childResult *storage.CompleteTagsResult
  termErr, childErr error
  // ...
)
```

src/query/api/v1/handler/graphite/find.go
```diff
@@ -57,52 +59,94 @@ func NewFindHandler(
...
+	if terminatedResult.CompleteNameOnly {
+		return nil, errors.New("terminated result is completing name only")
+	}
+
+	if childResult.CompleteNameOnly {
+		return nil, errors.New("child result is completing name only")
+	}
+
+	mapLength := len(terminatedResult.CompletedTags) + len(childResult.CompletedTags)
+	tagMap := make(map[string]bool, mapLength)
```
robskillington: nit: In future we should use an autogen'd map that can use `[]byte` as the key and we could then use `SetUnsafe(key, ..., Options{NoCopyKey: true, NoFinalizeKey: true})`.

src/query/api/v1/handler/graphite/find.go
arnikola: Actually not a fan of doing it like that, looks messier to me :(

If it's a standard we're using will update though :+1:

src/query/storage/validator/storage.go
```diff
@@ -158,12 +158,12 @@ func (s *debugStorage) Type() storage.Type {
...
 }
 
-func (s *debugStorage) FetchTags(
+func (s *debugStorage) SearchSeries(
 	ctx context.Context,
 	query *storage.FetchQuery,
 	_ *storage.FetchOptions,
 ) (*storage.SearchResults, error) {
-	return nil, errors.New("FetchTags not implemented")
+	return nil, errors.New("SearchSeries not implemented")
```
prateek: what's the deal with this storage type?

src/query/api/v1/handler/graphite/find_parser.go
prateek: nice, i like it.

src/query/api/v1/handler/graphite/find_parser.go
```diff
@@ -29,16 +29,40 @@ import (
...
+//  is used to reconstruct the return values.
+//  any error encountered during parsing.
+//
+// As an example, given the query `a.b*`, and metrics `a.bar.c` and `a.biz`,
+// terminatedQuery will return only [biz], and childQuery will return only
+// [bar].
+func parseFindParamsToQueries(r *http.Request) (
+	*storage.CompleteTagsQuery,
+	*storage.CompleteTagsQuery,
+	string,
```
prateek: nit: what is this string supposed to be? 

i've taken to giving returned variables name for this reason in the method spec (but no naked returns in the code itself). i.e. something like:

```
func parseFindParamsToQueries(r *http.Request) (
terminatedQuery *storage.CompleteTagsQuery,
childQuery *storage.CompleteTagsQuery,
? string,
err *xhttp.ParseError,
) { 
values := r.URL.Query()
	query := values.Get("query")
	if query == "" {
		return nil, nil, "", xhttp.NewParseError(errors.ErrNoQueryFound, http.StatusBadRequest)
		// notice how this isn't just "return"
	}
```

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -104,11 +106,14 @@ func ParseRequestTimeout(r *http.Request, configFetchTimeout time.Duration) (tim
...
 }
 
-// ParseTagCompletionParamsToQuery parses all params from the GET request
+// ParseTagCompletionParamsToQuery parses all params from the GET request.
 func ParseTagCompletionParamsToQuery(
 	r *http.Request,
 ) (*storage.CompleteTagsQuery, *xhttp.ParseError) {
-	tagQuery := storage.CompleteTagsQuery{}
+	tagQuery := storage.CompleteTagsQuery{
+		Start: time.Time{},
```
prateek: just making sure this is the default that can be overriden by the parameters passed over http

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -191,7 +196,7 @@ func ParseSeriesMatchQuery(
 		return nil, xhttp.NewParseError(err, http.StatusBadRequest)
 	}
 
-	tagMatchers := make([]models.Matchers, len(matcherValues))
+	queries := make([]*storage.FetchQuery, len(matcherValues))
```
prateek: on line 189 - what's with the magic numbers? (unrelated to this change)

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -52,14 +53,15 @@ const (
 
 var (
 	matchValues = []byte(".*")
+	roleName    = []byte("role")
```
prateek: hm this feels like a weird thing to modify a metric with. won't it conflict with people's tags?

arnikola: yeah, I've disabled stripping it for now

Prom does some dumb thing where it adds a "role":"remote" tag to each metric we get from the remote write endpoint which is questionable, but probably safer to leave it in.

src/query/storage/validator/storage.go
arnikola: It's a shitty mock; we should probably strip it and get a real mock in instead

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -346,56 +352,11 @@ func RenderTagValuesResultsJSON(
...
-		jw.EndArray()
-	}
-}
-
-// RenderSeriesMatchResultsJSON renders series match results to json format
+// RenderSeriesMatchResultsJSON renders series match results to json format.
 func RenderSeriesMatchResultsJSON(
 	w io.Writer,
-	results []*storage.CompleteTagsResult,
+	results []models.Metrics,
```
prateek: any reason to use custom encoder here? could we get away with just creating the right json object and using the default encoder

src/query/api/v1/handler/graphite/find_parser.go
arnikola: sgtm

src/query/api/v1/handler/prometheus/common.go
arnikola: Was defaulting to 40 days as query start for ...reasons; changed to go from 0 until Now by default

src/query/api/v1/handler/prometheus/common.go
arnikola: Historical reasons, mostly? We use this hand rolled encoder everywhere at the moment but should definitely look into changing this and using something sane instead

src/query/api/v1/handler/prometheus/common.go
arnikola: It can not... fixing

src/query/api/v1/handler/prometheus/common.go
prateek: yeah, i'd avoid doing anything magicy to the metrics. can add down the road behind config if it's really required

PR #1482, opened 2019-03-20
Improve M3DBs ability to apply back pressure on writes

richardartoul:
Previously, M3DB would only apply back pressure on writes if it received a commitlog queue full error, but this would only occur after almost all of the most expensive operations have already taken place.

To improve this situation, before performing any write and all the expensive memory operations like writing to in-memory data structures, hash map lookups, pooling and object allocation etc, we'll check the current length of the commit log queue. If its over 90% of its maximum capacity, we'll fail the write early to give the system time to recover.


src/dbnode/storage/database.go
```diff
@@ -101,10 +108,6 @@ type db struct {
 	metrics databaseMetrics
 	log     xlog.Logger
 
-	errors       xcounter.FrequencyCounter
```
prateek: Can probably delete the xcounter package too. Wasn’t used anywhere but here iirc

richardartoul: sure thing, done

PR #1486, opened 2019-03-21
add schema to namespace metadata options

Haijuncao:


Haijuncao:
a dup of the other pull request which is based on lastest master, close this one.

PR #1489, opened 2019-03-22
[scripts] bootstrapped: bump curl timeout to 60s

schallert:
Sometimes 5s isn't enough if the dbnode is under heavy pressure.

### DIFF
--- a/scripts/m3dbnode_bootstrapped.sh
+++ b/scripts/m3dbnode_bootstrapped.sh
@@ -27,7 +27,7 @@ cleanup() {
 
 trap cleanup EXIT
 
-curl --max-time 5 -sSf -o "$COORD_TMPFILE" "$COORD_PLACEMENT_ENDPOINT"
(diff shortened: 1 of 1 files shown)


scripts/m3dbnode_bootstrapped.sh
```diff
@@ -27,7 +27,7 @@ cleanup() {
 
 trap cleanup EXIT
 
-curl --max-time 5 -sSf -o "$COORD_TMPFILE" "$COORD_PLACEMENT_ENDPOINT"
+curl --max-time 60 -sSf -o "$COORD_TMPFILE" "$COORD_PLACEMENT_ENDPOINT"
```
vdarulis: why not refactor all the curl calls, or the entire health script to take shorter timeouts, but perform retries under a time budget?

schallert: So one reason I'm not as worried about that here is that this health script was added for and is only used by Kube, which does retries at a higher level. The reason we still need the timeout is that while it can time out HTTP health checks it can't time out exec scripts.

I can add a second version of the script or some args though to that effect.

PR #1491, opened 2019-03-23
Add operational guide on tuning availability, consistency, and durability + make ignoring corrupt commitlogs on bootstrap default in all sample YAMLs

richardartoul:


docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Client Write and Read Consistency
+
+The possible configuration values for write and read consistency are discussed in more detail [in this section](../m3db/architecture/consistencylevels.md) of the documentation, but in short M3DB behaves similarly to other H.A systems with configurable consistency such as Cassandra that allow the caller to control the consistency level of writes and reads from the client.
+
+The most important thing to understand is that if you want the guarantee that you will be able to read the outcome of every succesful write then you must perform both writes and read with a consistency level of `majority`.
+
+### Commitlog Configuration
+
+By default M3DB runs with an asynchronous commitlog such that writes will be acknowleged as successful by the client even though the data may not have been physically flushed to the commitlog on disk yet. M3DB supports changing this default behavior to run the commitlog synchronously, but this is not currently exposed to users in the YAML configuration and generally leads to a massive performance degradation.
```
mway: > writes will be acknowleged as successful by the client even though the data may not have been physically flushed to the commitlog on disk yet

nit: acknowledged (sp); though pedantically, it's not so much an ack as it is a report

also, thoughts on a slightly terser version?

> writes will be reported as successful by the client, though the data may not have been flushed to disk yet.

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+M3DB operators can control when M3DB will attempt to flush the commitlog via the `commitlog` section of the YAML configuration. For example the default configuration:
```
mway: nit:

> [...] For example, consider the default configuration:

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+```
+commitlog:
+	flushMaxBytes: 524288
+	flushEvery: 1s
+	queue:
+			calculationType: fixed
+			size: 2097152
+```
+
+states that the commitlog should be flushed whenever one of the following is true:
```
mway: nit:

> This configuration states that the commitlog should be flushed whenever either of the following is true:

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+1. 524288 or more bytes have been written since the last time M3DB flushed the commitlog.
+2. One or more seconds has elapsed since the last time M3DB flushed the commitlog.
+
+In addition, the configuration also states that M3DB should allow up to `2097152` writes to be buffered in the commitlog queue before the database node will begin rejecting incoming writes so it can attempt to drain the queue and catch up. Increasing the size of this queue can often increase the write throughput of an M3DB node at the cost of potentially losing more data if the node experiences a sudden failure like a hard crash or power loss.
```
mway: > Increasing the size of this queue can often increase the write throughput of an M3DB node at the cost of potentially losing more data if the node experiences a sudden failure like a hard crash or power loss.

this is an important tradeoff - if adding sections on performance/availability and consistency/durability, this should definitely be called out in the latter.

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Writing New Series Asynchronously
+
+The default M3DB YAML configuration will contain the following as a top-level key under the `db` section:
```
mway: nit:

> The [default M3DB configuration](https://github.com/m3db/m3/blob/master/src/dbnode/config/m3dbnode-all-config.yml) contains the following [...]

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+```
+writeNewSeriesAsync: true
+```
+
+This instructs M3DB to handle writes for new timeseries (for a given time block) asynchronously. Creating a new timeseries in memory is much more expensive than simply appending a new write to an existing series, so the default configuration of creating them asynchronously improves M3DBs write throughput significantly when many new series are being created all at once.
+
+However, since new time series are created asynchronously, its possible that there may be a brief delay inbetween when a write is acknowledged by the client and when that series becomes available for subsequent reads.
```
mway: nit:

> it's possible that there may be a brief delay between a write being acknowledged by the client and that series becoming available

also, jfmi, this delay is essentially guaranteed, right? the question is just how long the delay is, no?

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+```
+writeNewSeriesLimitPerSecond: 1048576
+```
+
+This value can be set much lower than the default value for workloads in which a significant increase in cardinality usually indicates an abusive caller.
+
+### Ignoring Corrupt Commitlogs on Bootstrap
+
+If M3DB is shutdown gracefully via a sigterm signal, it will ensure that all pending writes are flushed to the commitlog on disk before the process exists. However, in situations where the process crashed/exited unexpectedly or the node itself experienced a sudden failure, the tail end of the commitlog may be corrupt.
```
mway: nit:

> If M3DB shuts down gracefully (i.e. via SIGTERM), [...]

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+In that situation M3DB will read as much of the commitlog as its able to in order to recover as much data as possible, however it then needs to make a decision. It can either come up successfully and tolerate the minor amount of data loss or it can attempt to stream the missing data from its peers. This behavior is controlled by the following default configuration:
```
mway: > In that situation M3DB will read as much of the commitlog as its able to in order to recover as much data as possible, however it then needs to make a decision. It can either come up successfully and tolerate the minor amount of data loss or it can attempt to stream the missing data from its peers.

i'd recommend a minor tweak:

> In such situations, M3DB will read as much of the commitlog as possible in an attempt to recover the maximum amount of data. However, it then needs to make a decision: it can either (a) come up successfully and tolerate an ostensibly minor amount of data loss, or (b) attempt to stream the missing data from its peers.

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+```
+bootstrap:
```
mway: nit: change this YAML indentation to 2 or 4 spaces (the default config itself is inconsistent)

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+	commitlog:
+		returnUnfulfilledForCorruptCommitLogFiles: false
+```
+
+In the situation where only a single node fails, the optimal outcome is for the node to attempt to repair itself from one of its peers, however, if a quorum of nodes fail and encounter corrupt commitlog files then they will get stuck in a deadlock attempting to stream data from each other with none of them able to make any progress due to the fact that they cannot achieve quorum. This issue requires an operator with significant M3DB operational experience to step in and resolve, so we recommend running with the default configuration set to `false` to avoid this issue altogether since in most cases a small amount of data loss is preferable to a quorum of nodes that crash and fail to start back up automatically.
```
mway: nits:

> attempt to repair itself from one of its peers. However, [...]

> and encounter corrupt commit log files, they will deadlock while attempting to stream data from each other, as no nodes will be able to make progress due to a lack of quorum.

> an operator with significant M3DB operational experience to manually bootstrap the cluster; thus, the official recommendation is to set `returnUnfulfilledForCorruptCommitLogFiles: false` to avoid this issue altogether. In most cases, a small amount of data loss [...]

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Commitlog Configuration
+
+By default M3DB runs with an asynchronous commitlog such that writes will be acknowleged as successful by the client even though the data may not have been physically flushed to the commitlog on disk yet. M3DB supports changing this default behavior to run the commitlog synchronously, but this is not currently exposed to users in the YAML configuration and generally leads to a massive performance degradation.
+
+M3DB operators can control when M3DB will attempt to flush the commitlog via the `commitlog` section of the YAML configuration. For example the default configuration:
+
+```
+commitlog:
+	flushMaxBytes: 524288
```
mway: nit: change this YAML indentation to 2 or 4 spaces (the default config itself is inconsistent)

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Overview
+
+Generally speaking M3DB is an [H.A](https://en.wikipedia.org/wiki/High_availability) system because it doesn't use a consensus protocol like Raft or Paxos to enforce strong consensus and consistency guarantees. However, even within the category of H.A systems, there is a broad spectrum of consistency and durability guarantees that a database can provide. Like many other H.A systems, M3DB allows database operators to tune the consistency and durability guarantees of the system in order to achieve the desired balance between performance, availability, durability, and consistency.
+
+Generally speaking, all of the default and example YAML configuration files for M3DB are configured to favor performance and availability over consistency and durability since that is usually the tradeoff that makes most sense for metrics and observability workloads. Database operators who are using M3DB for workloads that require stricter consistency and durability guarantees should consider tuning the default configuration to better suite their use case. The rest of this document describes the various configuration options that are available to M3DB operators to make those tradeoffs.
```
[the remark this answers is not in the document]
richardartoul: Yeah I'm happy to try and break it down more, the thing I struggle with though is that its all kinda mushed together.

For example, write/read consistency is mostly about consistency/durability, however, its also a performance thing. If you only require a write to succeed to one node, your P99 latency is likely to be much lower.

Similarly the async commitlog configuration is for performance reasons, but it has an impact on your durability and as a result, your consistency.

I agree that it feels like too much stuff mixed together but I'm also not sure how to separate them conceptually because they really are all tangled together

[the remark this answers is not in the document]
mway: yeah, i know what you mean. in my mind it looks something like this:

## Tuning For Performance & Availability
- async commitlog, incl consistency/durability caveats
- larger buffers for more throughput, incl consistency/durability caveats
- etc

## Tuning For Consistency & Durability
- sync commitlog, incl performance caveats
- smaller buffers for more flushing and lower margin for data loss, incl performance caveats
- etc

the gist is that yeah, you're right, they are intertwined - they're essentially inversely related concepts, so we do need to talk about everything in both sections. they're mutually exclusive, however, and each section is geared toward its half of the requirements. (and it's actually good that the sections overlap on subject matter - although not intent - because folks might come to the doc looking for their specific use case, and should have a summary of all the required information without needing to piece it together throughout the doc.)

docs/operational_guide/availability_consistency_durability.md
richardartoul: this is tricky to link to because we have multiple "defaults" depending on whether coordinator/etcd are running embedded etc

docs/operational_guide/availability_consistency_durability.md
richardartoul: Assuming an instantaneous network yeah, but in practice the reads that immediately follow writes almost always succeed

docs/operational_guide/availability_consistency_durability.md
mway: ah, true - similar to gearing the docs/defaults to obs-centric use cases, it might be good to assume a primary mode of operation as well (probably not an embedded one).

no issue with keeping it generalized though, in which case we should prob just link to the dir instead so that folks can reference the file that is most relevant to their use case.

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
+# Tuning Availability, Consistency, and Durability
+
+## Overview
+
+M3DB is designed as a High Availability [HA](https://en.wikipedia.org/wiki/High_availability) system because it doesn't use a consensus protocol like Raft or Paxos to enforce strong consensus and consistency guarantees.
+However, even within the category of HA systems, there is a broad spectrum of consistency and durability guarantees that a database can provide.
+To address as many use cases as possible, M3DB can be tuned to achieve the desired balance between performance, availability, durability, and consistency.
+
+Generally speaking, [the default and example configuration for M3DB](https://github.com/m3db/m3/tree/master/src/dbnode/config) favors performance and availability, as that is well-suited for M3DB's most common metrics and Observability use cases. To instead favor consistency and durability, consider tuning values as described in (TODO).
```
mway: i think i'd suggested copy with Observability capitalized, but let's generalize this a bit more, e.g.:

> [...] that is well-suited for M3DB's most common use cases (such as metric storage for observability). [...]

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
...
+Database operators who are using M3DB for workloads that require stricter consistency and durability guarantees should consider tuning the default configuration to better suite their use case.
+The rest of this document describes the various configuration options that are available to M3DB operators to make those tradeoffs.
+
+While reading the rest of this document, we recommend referring to [the default configuration file](https://github.com/m3db/m3/blob/master/src/dbnode/config/m3dbnode-all-config.yml) (which has every possible configuration value set) to see how the described values fit into M3DB's configuration as a whole.
```
mway: since both of these paragraphs reference reading the rest of the doc, we might want to combine them. e.g.:

> The rest of this document describes the various configuration options that are available to M3DB operators to make such tradeoffs. While reading it, we recommend referring to [...]

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Tuning for Performance and Availability
+
+### Client Write and Read consistency
+
+We recommend running the client with `writeConsistencyLevel` set to `majority` and `readConsistencyLevel` set to `unstrict_majority`.
+This means that all write must be acknowledged by a quorums of nodes in order to be considered succesful, and that reads will attempt to achieve quorum, but will return the data from a single node if they are unable to achieve quorum.
```
mway: we might want to add an addendum to this sentence for extra clarity, like:

> This ensures that reads will nominally ensure consistency, but degraded conditions will not cause reads to fail outright.

or similar. thoughts?

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
...
+    calculationType: fixed
+    size: 2097152
+```
+
+This configuration states that the commitlog should be flushed whenever either of the following is true:
+
+1. 524288 or more bytes have been written since the last time M3DB flushed the commitlog.
+2. One or more seconds has elapsed since the last time M3DB flushed the commitlog.
+
+In addition, the configuration also states that M3DB should allow up to `2097152` writes to be buffered in the commitlog queue before the database node will begin rejecting incoming writes so it can attempt to drain the queue and catch up. Increasing the size of this queue can often increase the write throughput of an M3DB node at the cost of potentially losing more data if the node experiences a sudden failure like a hard crash or power loss.
```
mway: > drain the queue and catch up

i don't think (but am potentially just not seeing) that we've documented/diagrammed the architectural details to explain how the queue works (maybe that'd be overkill), but if we ever add that, this should probably link to it.

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
...
+
+However, since new time series are created asynchronously, its possible that there may be a brief delay inbetween when a write is acknowledged by the client and when that series becomes available for subsequent reads.
+
+M3DB also allows operators to ratelimit the number of new series that can be created per second via the following configuration:
+
+```
+writeNewSeriesLimitPerSecond: 1048576
+```
+
+This value can be set much lower than the default value for workloads in which a significant increase in cardinality usually indicates an abusive caller.
```
mway: > abusive caller

i'd recommend changing this to "misbehaving", as abuse implies malicious intent. more likely, it's a caller that doesn't understand the goals or limitations of the system.

don't want to get too tied up in semantics, but if we use words like "abusive", "misbehaving", etc, we might want to very clearly call out or link to expectations (e.g. wrt dimensionality, cardinality, etc).

docs/operational_guide/availability_consistency_durability.md
mway: we have sections to replace TODO with now! 🎉

docs/operational_guide/availability_consistency_durability.md
benraskin92: Maybe give a brief example?

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -12,18 +12,19 @@ The rest of this document describes the various configuration options that are a
...
-The most important thing to understand is that **if you want to guarantee that you will be able to read the result of every successful write, then both writes and reads must be done with `majority` consistency.**
+We recommend running the client with `writeConsistencyLevel` set to `majority` and `readConsistencyLevel` set to `unstrict_majority`.
+This means that all write must be acknowledged by a quorums of nodes in order to be considered succesful, and that reads will attempt to achieve quorum, but will return the data from a single node if they are unable to achieve quorum.
+You can read about the consistency levels in more detail in [the Consistency Levels section](../m3db/architecture/consistencylevels.md)
 
 ### Commitlog Configuration
 
-By default M3DB runs with an asynchronous commitlog such that writes will be reported as successful by the client, though the data may not have been flushed to disk yet.
-M3DB supports changing this default behavior to run the commitlog synchronously, but this is not currently exposed to users in the YAML configuration and generally leads to a massive performance degradation.
+We recommend running M3DB with an asynchronous commitlog.
```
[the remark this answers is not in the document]
benraskin92: +1

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
...
+  flushMaxBytes: 524288
+  flushEvery: 1s
+  queue:
+    calculationType: fixed
+    size: 2097152
+```
+
+This configuration states that the commitlog should be flushed whenever either of the following is true:
+
+1. 524288 or more bytes have been written since the last time M3DB flushed the commitlog.
```
benraskin92: Why `524288` and `2097152`?

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
...
+
+The default M3DB YAML configuration will contain the following as a top-level key under the `db` section:
+
+```
+writeNewSeriesAsync: true
+```
+
+This instructs M3DB to handle writes for new timeseries (for a given time block) asynchronously. Creating a new timeseries in memory is much more expensive than simply appending a new write to an existing series, so the default configuration of creating them asynchronously improves M3DBs write throughput significantly when many new series are being created all at once.
+
+However, since new time series are created asynchronously, its possible that there may be a brief delay inbetween when a write is acknowledged by the client and when that series becomes available for subsequent reads.
```
benraskin92: `its` => `it's`

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
...
+
+```
+writeNewSeriesLimitPerSecond: 1048576
+```
+
+This value can be set much lower than the default value for workloads in which a significant increase in cardinality usually indicates an abusive caller.
+
+### Ignoring Corrupt Commitlogs on Bootstrap
+
+If M3DB is shuts down gracefully (i.e via SIGTERM), it will ensure that all pending writes are flushed to the commitlog on disk before the process exists.
```
benraskin92: `shuts` => `shut`

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
...
+### Client Write and Read consistency
+
+The most important thing to understand is that **if you want to guarantee that you will be able to read the result of every successful write, then both writes and reads must be done with `majority` consistency.**
+This means that both writes *and* reads will fail if a quorum of nodes are unavailable for a given shard.
+You can read about the consistency levels in more detail in [the Consistency Levels section](../m3db/architecture/consistencylevels.md)
+
+### Commitlog Configuration
+
+M3DB supports running the commitlog synchronously such that every write is flushed to disk and fsync'd before the client receives a successful acknowledgement, but this is not currently exposed to users in the YAML configuration and generally leads to a massive performance degradation.
+We only recommend operating M3DB this way for workloads where data consistency and durability is strictly required, and even then their may be better alternatives such as running M3DB with the bootstrapping configuration: `filesystem,peers,uninitialized_topology` as described in our [bootstrapping operational guide](./bootstrapping.md).
```
benraskin92: `their` => `there`

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
...
+
+Since this operation is so expensive, M3DB allows operator to ratelimit the number of new series that can be created per second via the following configuration (also a top-level key under the `db` section):
+
+```
+writeNewSeriesLimitPerSecond: 1048576
+```
+
+### Ignoring Corrupt Commitlogs on Bootstrap
+
+As described in the "Tuning for Performance and Availability" section, we recommend configuring M3DB to ignore corrupt commitlog files on bootstrap. However, if you want to avoid any amount of inconsistency or data loss, no matter how minor, then you should configure M3DB to return unfulfilled when the commitlog bootstrapper encounters corrupt commitlog files. You can do so by modifying your configuration to look like this:
```
benraskin92: Hmm do we need both this section and `### Ignoring Corrupt Commitlogs on Bootstrap`?

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -0,0 +1,126 @@
+# Tuning Availability, Consistency, and Durability
+
+## Overview
+
+M3DB is designed as a High Availability [HA](https://en.wikipedia.org/wiki/High_availability) system because it doesn't use a consensus protocol like Raft or Paxos to enforce strong consensus and consistency guarantees.
+However, even within the category of HA systems, there is a broad spectrum of consistency and durability guarantees that a database can provide.
+To address as many use cases as possible, M3DB can be tuned to achieve the desired balance between performance, availability, durability, and consistency.
+
+Generally speaking, [the default and example configuration for M3DB](https://github.com/m3db/m3/tree/master/src/dbnode/config) favors performance and availability, as that is well-suited for M3DB's most common metrics and Observability use cases. To instead favor consistency and durability, consider tuning values as described in (TODO).
+Database operators who are using M3DB for workloads that require stricter consistency and durability guarantees should consider tuning the default configuration to better suite their use case.
```
schallert: Nit: `s/suite/suit/`

docs/operational_guide/availability_consistency_durability.md
richardartoul: Yeah unfortunately we don't have that right now

docs/operational_guide/availability_consistency_durability.md
richardartoul: Misbehaving in this context would depend on your setup and workload. I.E at Uber it might be someone emitting UUIDs in a spark job, but for someone elses setup that might be the intended use case

docs/operational_guide/availability_consistency_durability.md
richardartoul: added an extra sentence

docs/operational_guide/availability_consistency_durability.md
richardartoul: its just the default we've always used. Presumably @robskillington did some benchmarking?

docs/operational_guide/availability_consistency_durability.md
mway: understood - still think we should switch to "misbehaving" instead of "abusive" unless you disagree with the semantics there.

docs/operational_guide/availability_consistency_durability.md
mway: i didn't think so - just a note for the future, then. :)

docs/operational_guide/availability_consistency_durability.md
richardartoul: Not totally following. I have the subheading twice, once under availability section and once under consistency

docs/operational_guide/availability_consistency_durability.md
mway: they're 2^19 and 2^21, respectively, if that helps.

docs/operational_guide/availability_consistency_durability.md
richardartoul: yep i did

PR #1492, opened 2019-03-25
[ci] push / build docs from BK

### LINKED ISSUE #1406: create a buildkite job for publishing docs to m3db.io
Our docs currently auto-build for github pages (http://m3db.github.io/m3/) but not the m3db.io docs page (https://docs.m3db.io/). The latter has to be done manually.

We should automate this such that the main site docs are pushed on commits to `master`.

schallert:
We currently have auto builds set up for our GH pages site, but not
https://docs.m3db.io which we also link to frequently.

This PR adds a Buildkite step that allows us to build and push docs from
builds on `master`.

Example of this working: https://github.com/m3db/m3/commit/a7f558c82c066d392be635192fd51899cd3fab16

Fixes #1406

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -19,6 +19,15 @@ steps:
       automatic:
         limit: 1
       manual: true
+  - name: "Check for docs build :books:"
+    command: ".buildkite/scripts/check_do_docs.sh"
+    agents:
+      queue: init
+    timeout_in_minutes: 10
+    retry:
+      automatic:
+        limit: 1
+      manual: true
   - name: "Codegen"
     command: make clean install-vendor test-all-gen
     env:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -10,17 +10,10 @@ common: &common
     manual: true
 
 steps:
-  - name: "Check for :docker: build"
-    command: ".ci/docker/check_do_docker.sh"
-    agents:
-      queue: init
-    timeout_in_minutes: 10
-    retry:
-      automatic:
-        limit: 1
-      manual: true
-  - name: "Check for docs build :books:"
-    command: ".buildkite/scripts/check_do_docs.sh"
+  - name: "Check for docker and docs builds :docker: :books:"
+    commands:
+    - ".ci/docker/check_do_docker.sh"
+    - ".buildkite/scripts/check_do_docs.sh"
     agents:
       queue: init
     timeout_in_minutes: 10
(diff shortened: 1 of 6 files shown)


.buildkite/scripts/docs_push.sh
```diff
@@ -0,0 +1,26 @@
...
+  echo "nothing to do"
+  exit 0
+fi
+
+mkdir -p "$HOME/.ssh"
+ssh-keyscan github.com >> "$HOME/.ssh/known_hosts"
+git config --local user.email "buildkite@m3db.io"
+git config --local user.name "M3 Buildkite Bot"
+
+rm -rf site
```
prateek: Can these lines be moved/call into the Makefile?

.buildkite/pipeline.yml
```diff
@@ -19,6 +19,15 @@ steps:
       automatic:
         limit: 1
       manual: true
+  - name: "Check for docs build :books:"
+    command: ".buildkite/scripts/check_do_docs.sh"
```
prateek: would it be better to collapse this and the do_docker into 1 step? Will be quicker on average than waiting for more agents/containers to spin up.

schallert: I'm kinda a fan of being able to do the checks from the high-throughput init agents and not have to wait for a worker to be available. Thoughts?

.buildkite/scripts/docs_push.sh
schallert: So I looked into calling the makefile from the docs container and it turns out our Makefile makes a lot of assumptions about being run in `bash`, but alpine only has `ash`. I can convert the docs image to a debian one but then I'll have to make sure it has all the same mkdocs packages and the container will be heavier. Not sure if it's worth, wondering what you think.

.buildkite/pipeline.yml
prateek: Yeah we’re on the same page. I was suggesting merging this and the previous step. Both are in the init worker so shouldn’t be any slower, and this way you don’t have to wait for init worker slots. Happy to leave as is if you’re not noticing any wait time for this step too.

.buildkite/scripts/docs_push.sh
prateek: Yech. Isn’t worth it to do the work to update either right now. I’d suggest dropping TODOs here and by the makefile target reminding people to keep commands in sync and an explanation for how we might migrate to a common target if/when we have time

.buildkite/pipeline.yml
schallert: Ahh I misread your q, I thought you meant merge this and the build docs step 🤦‍♂️ . Yeah let me take a look

schallert: Buildkite supports `commands:` instead of just a single `command:` 👏

PR #1497, opened 2019-03-26
[ci] include gh pages in docs pushes

schallert:


.buildkite/docs-build-pipeline.yml
```diff
@@ -1,4 +1,7 @@
 steps:
+ # We add a wait step here so that docker images and docs won't get built
+  # unless rest of CI passes.
```
richardartoul: nit space here

PR #1498, opened 2019-03-26
Remove unsupported static config and add docs link testing

robskillington:
The docs testing tool also caught a link that was broken in the docs.

In the future we should add any further tests that we want to support to make sure the docs aren't broken to this `docs_test` tool.


Makefile
```diff
@@ -198,25 +199,32 @@ docs-container:
...
 .PHONY: docs-deploy
-docs-deploy: docs-container
+docs-deploy: docs-validate docs-container
 	docker run -v $(PWD):/m3db --rm -v $(HOME)/.ssh/id_rsa:/root/.ssh/id_rsa:ro -it m3db-docs "mkdocs build -e docs/theme -t material && mkdocs gh-deploy --force --dirty"
 
 .PHONY: docker-integration-test
 docker-integration-test:
 	@echo "--- Running Docker integration test"
+	# Run docker container to build docs
+	make docs-build
```
richardartoul: Should we just make this a separate build kite thing? Probably easier to track and parallelize

robskillington: Sure, will break it out.

src/cmd/tools/docs_test/main/main.go
```diff
@@ -0,0 +1,106 @@
...
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+
+	"github.com/russross/blackfriday"
+)
+
+var (
+	repoPathRegexp = regexp.MustCompile("^https://github.com/(.*)/blob/[a-zA-Z0-9]+/([-_.a-zA-Z0-9/]+)(#?.*)(\\??.*)$")
```
richardartoul: Can you add a comment here saying that this matches how we link to Github files from our docs, and that the capture group lets you extract out a relative path that can be used to make sure the file exists in this P.R

prateek: maybe a better place for this var is `src/x/docs`?

robskillington: Sure thing, added documentation, moved it to `src/x/docs`.

PR #1499, opened 2019-03-26
Add 0.8.0 migration warning + update kernel configuration documentation + update automatic M3DB limits warnings

richardartoul:


docker/sysctl-setter/Dockerfile
```diff
@@ -12,6 +12,14 @@ while true; do\n\
   if [ "$SVAL" -ne 1 ]; then\n\
     sysctl -w vm.swappiness=1\n\
   fi\n\
+  FVAL=$(sysctl -n fs.file-max)\n\
```
schallert: From my googling so far, it seems like most recent kernels set this to an insanely high value anyway. Can we leave this out?

docker/sysctl-setter/Dockerfile
```diff
@@ -12,6 +12,14 @@ while true; do\n\
...
+  if [ "$FVAL" -ne 3000000 ]; then\n\
+    sysctl -w fs.file-max=3000000\n\
+  fi\n\
+  OVAL=$(sysctl -n fs.nr_open)\n\
+  if [ "$OVAL" -ne 3000000 ]; then\n\
```
schallert: Can we change this (and the map map count section) to be `-le` instead of `-ne`? That way if someone already has a higher val we don't lower it

docs/operational_guide/kernel_configuration.md
```diff
@@ -31,12 +31,29 @@ To set this value permanently, update the `vm.swappiness` setting in `/etc/sysct
 ## rlimits
 M3DB also can use a high number of files and we suggest setting a high max open number of files due to per partition fileset volumes.
 
-On Linux you can set a high limit for maximum number of open files in `/etc/security/limits.conf`:
+On Linux you can set a high limit for the maximum number of open files for a specific user in `/etc/security/limits.conf`:
```
schallert: We might want to mention that when running under systemd this vals will probably be ignored and maybe point to some of the GH issues or docs we found. Let me know if you can't find some refs I can help look

docs/operational_guide/kernel_configuration.md
```diff
@@ -31,12 +31,29 @@ To set this value permanently, update the `vm.swappiness` setting in `/etc/sysct
...
 ```
-your_m3db_user        hard nofile 500000
-your_m3db_user        soft nofile 500000
+your_m3db_user        hard nofile 3000000
+your_m3db_user        soft nofile 3000000
 ```
 
+In addition, you may need to override the system and process-level limits set by the kernel with the following commands:
+
+Set the kernel-wide limit to 3 million.
```
schallert: Can we say that most kernels come w/ a higher value for this and only set this if your val is lower

docs/operational_guide/kernel_configuration.md
```diff
@@ -31,12 +31,29 @@ To set this value permanently, update the `vm.swappiness` setting in `/etc/sysct
...
+```
+sysctl -w fs.file-max=3000000
+```
+
+Set the process-level limit to 3 million
```
schallert: Nit: similar to above can we say "if your value is lower"

docker/sysctl-setter/Dockerfile
richardartoul: It doesn't hurt if we do `le`

docs/operational_guide/kernel_configuration.md
richardartoul: I think a comment should be enough and then tell them to see the systemd details below

vdarulis: `/etc/security/limits.conf` only works for PAM stuff, let's not mislead users. It must be set in systemd unit, upstart config or the init script itself.

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+# 0.8.0 (T.B.D)
+
+## Migration Disclaimer
+
+Version 0.8.0 of M3 switches M3DB to use file descriptors and the Read() system call for reading data files instead of mmaps.
```
robskillington: nit: `s/reading data files/reading time series data files` to clearly distinguish between index vs ts data files.

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+This should improve read latency and memory utilization for some workloads, as well as enables the ability to increase fetch concurrency past the default value.
```
benraskin92: Can we link to anything detailing what types of workloads will be affected?

richardartoul: Nothing to link to, but I can add a comment

PR #1483, opened 2019-03-21
dbnode Aggregate RPC implementation

prateek:

PR #1502, opened 2019-03-27
[index] Standardizes aggregate interfaces

arnikola:
This PR fulfills several outstanding tasks for the aggregate endpoint:

- Integrates the interfaces from the RPC PR here: #1483 and flows them through the stack, from the db session down to the index itself.
- Adds capability to return tag names only, while defaulting to names+values (my bad, forgot to add it initially!)

This allows the query side PR(#1481) to be updated and landed, and minimizes required end-to-end work on #1483


src/dbnode/storage/index/aggregate_results.go
```diff
@@ -107,16 +107,70 @@ func (r *aggregatedResults) AddDocuments(batch []doc.Document) (int, error) {
 func (r *aggregatedResults) addDocumentsBatchWithLock(
 	batch []doc.Document,
 ) error {
-	for i := range batch {
-		err := r.addDocumentWithLock(batch[i])
-		if err != nil {
-			return err
+	for _, doc := range batch {
+		if r.aggregateOpts.Type == AggregateTagNamesAndValues {
```
robskillington: Maybe switch here and return if unknown?

arnikola: Good call :+1:

src/dbnode/storage/index/types.go
```diff
@@ -74,6 +84,19 @@ func (o QueryOptions) LimitExceeded(size int) bool {
...
+// AggregationOptions enables users to specify constraints on aggregations.
+type AggregationOptions struct {
+	QueryOptions
+	TermFilter AggregateTermFilter
+	Type       AggregationType
+}
+
+// LimitExceeded returns whether a given size exceeds the limit
+// the query options imposes, if it is enabled.
+func (o AggregationOptions) LimitExceeded(size int) bool {
```
prateek: the embedding should give this to you fo free iirc

src/dbnode/storage/index/types.go
```diff
@@ -186,6 +209,9 @@ type AggregateResultsOptions struct {
 
 	// Optional param to filter aggregate values.
 	TermFilter AggregateTermFilter
+
+	// Type determines what result is required.
+	Type AggregationType
```
prateek: can you delete AggregateResultsOptions now?

src/dbnode/storage/index/types.go
arnikola: Oh nice :+1:

src/dbnode/storage/index/types.go
arnikola: I wanted to keep it as similar as possible to the regular `Query` path which does a similar thing when it breaks out the options relating to the result from those relating to the query parameters here

PR #1505, opened 2019-03-27
[ci] build + release m3db with go1.12

schallert:


schallert:
> Nice, should we throw this up in the test cluster?

Let's do it!

PR #1506, opened 2019-03-28
[m3db] Add Jaeger tracing to m3db

benraskin92:
Screen Shot 2019-03-28 at 10 44 30 AM


src/cmd/services/m3dbnode/config/config.go
```diff
@@ -137,6 +137,9 @@ type DBConfiguration struct {
 
 	// Write new series asynchronously for fast ingestion of new ID bursts.
 	WriteNewSeriesAsync bool `yaml:"writeNewSeriesAsync"`
+
+	// Tracing configures opentracing. If not provided, tracing is disabled.
+	Tracing instrument.TracingConfiguration `yaml:"tracing"`
```
prateek: i'd make this a pointer type to make it optional to specify

src/dbnode/network/server/tchannelthrift/context.go
```diff
@@ -40,7 +40,9 @@ func RegisterServer(channel *tchannel.Channel, service thrift.TChanServer, conte
 	server := thrift.NewServer(channel)
 	server.Register(service, thrift.OptPostResponse(postResponseFn))
 	server.SetContextFn(func(ctx xnetcontext.Context, method string, headers map[string]string) thrift.Context {
-		ctxWithValue := xnetcontext.WithValue(ctx, contextKey, contextPool.Get())
+		xCtx := contextPool.Get()
+		xCtx.SetGoContext(xnetcontext.TODO())
```
prateek: can replace this line ```xCtx.SetGoContext(ctx)```

src/cmd/services/m3dbnode/config/config.go
benraskin92: This should be fine since we check if its a no-op in `server.go`:
```
	if _, ok := tracer.(opentracing.NoopTracer); ok {
		logger.Info("tracing disabled for m3dbnode; set `tracing.backend` to enable")
	}
```

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -66,6 +70,9 @@ const (
 	maxSegmentArrayPooledLength = 32
 	// Any pooled error slices that grow beyond this capcity will be thrown away.
 	writeBatchPooledReqPoolMaxErrorsSliceSize = 4096
+
+	fetchTaggedSpanID = "fetch"
```
benraskin92: Better names

scripts/development/m3_stack/start_m3.sh
```diff
@@ -11,7 +11,8 @@ fi
 
 echo "Bringing up nodes in the background with docker compose, remember to run ./stop.sh when done"
 docker-compose -f docker-compose.yml up $DOCKER_ARGS m3coordinator01
-docker-compose -f docker-compose.yml up $DOCKER_ARGS m3db_seed
+docker-compose -f docker-compose.yml up --build -d --renew-anon-volumes m3db_seed
+# docker-compose -f docker-compose.yml up $DOCKER_ARGS m3db_seed
```
benraskin92: Me testing - will revert before landing.

scripts/development/m3_stack/start_m3.sh
```diff
@@ -211,6 +212,11 @@ if [[ "$AGGREGATOR_PIPELINE" = true ]]; then
     curl http://localhost:7206/api/v1/json/report -X POST -d '{"metrics":[{"type":"gauge","value":42,"tags":{"__name__":"foo_metric","foo":"bar"}}]}'
 fi
 
+if [[ "$START_JAEGER" = true ]] ; then
+    docker-compose -f docker-compose.yml up $DOCKER_ARGS jaeger
```
benraskin92: Add check to make sure `localhost:16686` is available.

src/dbnode/config/m3dbnode-local-etcd.yml
```diff
@@ -27,6 +27,9 @@ coordinator:
     idScheme: quoted
 
 db:
+  tracing:
```
benraskin92: Remove before landing.

src/dbnode/x/opentracing/context.go
```diff
@@ -0,0 +1,82 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
benraskin92: Remove this once https://github.com/m3db/m3x/pull/214 lands

src/dbnode/x/opentracing/context_test.go
```diff
@@ -0,0 +1,106 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
benraskin92: Remove this once https://github.com/m3db/m3x/pull/214 lands

scripts/development/m3_stack/start_m3.sh
```diff
@@ -9,9 +9,21 @@ if [[ "$FORCE_BUILD" = true ]] ; then
     DOCKER_ARGS="--build -d --renew-anon-volumes"
 fi
 
+# need to start Jaeger before m3db or else m3db will not be able to talk to the Jaeger agent.
+if [[ "$START_JAEGER" = true ]] ; then
+    docker-compose -f docker-compose.yml up $DOCKER_ARGS jaeger
+    sleep 5
+    JAEGER_STATUS=$(curl -s -o /dev/null -w '%{http_code}' localhost:14269)
+    if [ $JAEGER_STATUS -ne 204 ]; then
```
benraskin92: Once https://github.com/jaegertracing/jaeger/issues/1450 is fixed, we can change this logic.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -256,22 +263,49 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
...
+		sp, spCtx = opentracingutil.StartSpanFromContext(goCtx, querySpanID)
+		sp.LogFields(
+			opentracinglog.String("query", req.Query.String()),
+			opentracinglog.String("namespace", req.NameSpace),
+			opentracingutil.Time("start", time.Unix(0, req.RangeStart)),
+			opentracingutil.Time("end", time.Unix(0, req.RangeStart)),
+		)
+
+		ctx.SetGoContext(spCtx)
+		defer sp.Finish()
```
richardartoul: This finish only applies to the jaeger trace right?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -256,22 +263,49 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
...
+			sp.LogFields(opentracinglog.Error(err))
+		}
+
+		return nil, tterrors.NewBadRequestError(err)
 	}
 
 	q, err := convert.FromRPCQuery(req.Query)
 	if err != nil {
+		if sp != nil {
+			sp.LogFields(opentracinglog.Error(err))
```
richardartoul: I think you can prevent having to do this everywhere by making the return arguments of this function named, so it becomes:

```golang
func (s *service) Query(tctx thrift.Context, req *rpc.QueryRequest) (result *rpc.QueryResult_, err error)
```

and then after you create the span you can do something like

```golang
defer func() {
    if err != nil && sp != nil {
        sp.LogFields(opentracinglog.Error(err))
   }
}()
```

This shouldn't interfere with the call to `defer sp.Finish()` since defers are executed last-in first-out order

src/dbnode/network/server/tchannelthrift/node/service.go
richardartoul: Do we have to close the span? The context cant do that for us?

src/dbnode/server/server.go
```diff
@@ -147,6 +151,13 @@ func Run(runOpts RunOptions) {
 		os.Exit(1)
 	}
 
+	// Currently, the zapLogger is being used for tracing only.
```
richardartoul: how come we have to use zap for tracing? is that just how it works?

src/dbnode/server/server.go
```diff
@@ -183,6 +194,25 @@ func Run(runOpts RunOptions) {
...
+		traceCloser io.Closer
+	)
+
+	if cfg.Tracing == nil {
+		tracer = opentracing.NoopTracer{}
+		logger.Infof("tracing disabled for %s; set `tracing.backend` to enable", serviceName)
+	} else {
+		// setup tracer
+		tracer, traceCloser, err = cfg.Tracing.NewTracer(serviceName, scope, zapLogger)
+		if err != nil {
```
richardartoul: Is there any way where we can keep the tracing configuration in config by default and have it just not break anything if jaeger is not available? I don't know how the buffering / the library work internally

src/dbnode/storage/database.go
```diff
@@ -708,8 +716,31 @@ func (d *db) QueryIDs(
...
 	opts index.QueryOptions,
 ) (index.QueryResult, error) {
+	var (
+		sp    opentracing.Span
+		spCtx stdctx.Context
+	)
+
+	goCtx, exists := ctx.GoContext()
+	if exists {
+		sp, spCtx = opentracingutil.StartSpanFromContext(goCtx, dbQuerySpanID)
```
richardartoul: Does this span automatically get tied together with the span from the RPC layer?

Also could we add a method to our context that makes adding each span less redundant? Maybe something like:

```golang
sp, closer := ctx.StartSpan(
    spanID,
    opentracinglog.String("query", query.String()),
    opentracinglog.String("namespace", namespace.String()),
    opentracinglog.Int("limit", opts.Limit),
    opentracinglog.Time("start", opts.StartInclusive),
    opentracinglog.Time("start", opts.StartInclusive),
)
defer closer()
sp.LogFields()
...
```

and all of the `ctx.GoContext()` `if exists` etc boilerplate can get shoved into that helper method

That way its a little cleaner/shorter and the API allows you to add the logs that you have immediately right away and then if you need to add more later you can use the returned `sp` object.

src/dbnode/storage/database.go
```diff
@@ -708,8 +716,31 @@ func (d *db) QueryIDs(
...
+		sp.LogFields(
+			opentracinglog.String("query", query.String()),
```
richardartoul: Do we need to add the query to the span at every level? Same for `StartInclusive` and `EndExclusive`.

maybe I don't know how tracing works but I thought you could just add that information in one span and it was would associated with the entire trace and you wouldn't need to do it at each layer

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -256,22 +263,49 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
...
+
+	var (
+		sp    opentracing.Span
+		spCtx stdctx.Context
+	)
+
+	goCtx, exists := ctx.GoContext()
+	if exists {
+		sp, spCtx = opentracingutil.StartSpanFromContext(goCtx, querySpanID)
+		sp.LogFields(
```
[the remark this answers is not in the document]
robskillington: Yeah, can we only go down this path if this request is being sampled?  Is there a way to query whether this context is being sampled or not?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -256,22 +263,49 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
...
+			opentracinglog.String("query", req.Query.String()),
+			opentracinglog.String("namespace", req.NameSpace),
+			opentracingutil.Time("start", time.Unix(0, req.RangeStart)),
+			opentracingutil.Time("end", time.Unix(0, req.RangeStart)),
+		)
+
+		ctx.SetGoContext(spCtx)
```
robskillington: Hm, this seems racey if multiple goroutines have reference to the same `ctx`.

src/dbnode/network/server/tchannelthrift/node/service.go
robskillington: The easier thing, is just wrap this in another method.

i.e. making `Query` wrap calls to an inner `query`

```go
func (s *service) Query(...) (...) {
  var span opentracing.Span

  // maybe create span based on whether this is being sampled or not

  // call the actual method, with reference to the span)
  result, err := s.query(..., span)

  // now optionally add the error
  if span != nil && err != nil {
    span.LogFields(opentracinglog.Error(err)
  }

  // Also don't need to use a defer now to close the span
  span.Finish()

  return result, err
}

File: scripts/development/m3_stack/README.md [hunk 18]
```diff
@@ -9,19 +9,42 @@ This docker-compose file will setup the following environment:
...
 
 ## Grafana
 
 Use Grafana by navigating to `http://localhost:3000` and using `admin` for both the username and password. The M3DB dashboard should already be populated and working.
 
+## Jaeger
+
+To start Jaeger, add the following to your `m3dbnode.yml` file under `db`:
+
+```yaml
```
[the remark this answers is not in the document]
benraskin92: Yes- it'll break if Jaeger isn't running here:
```
localAgentHostPort: jaeger:6831
```

src/dbnode/network/server/tchannelthrift/node/service.go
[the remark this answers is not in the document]
benraskin92: Yeah the default sampling rate is quite low for Jaeger.

src/dbnode/network/server/tchannelthrift/node/service.go
benraskin92: Yes I believe so.

src/dbnode/server/server.go
benraskin92: The thing Andrew wrote to create the tracer in m3x uses `jaegerzap.NewLogger()` - also @prateek wants to move to using a zap logger anyway.

src/dbnode/server/server.go
benraskin92: so iiuc, traces are sent over UDP and will just silently fail if there is no Jaeger collector setup. However, if you don't have the Jaeger config at all, a no-op tracer is created which doesn't send any traces.

src/dbnode/storage/database.go
benraskin92: I believe logs are associated with just a single span, but I'll double check

src/dbnode/server/server.go
```diff
@@ -183,6 +194,25 @@ func Run(runOpts RunOptions) {
...
+			logger.Fatalf("could not initialize tracing for m3dbnode: %v", err)
```
richardartoul: Change this to log an error, create a NOOP reporter, then continue.

robskillington:
@benraskin92 let us know when this is good for another pass.  This will be a great addition.

scripts/development/m3_stack/start_m3.sh
prateek: nit: for posterity, could you put in a comment saying the same thing in the file

glide.yaml
```diff
@@ -12,6 +12,9 @@ import:
   - package: github.com/m3db/stackadler32
     version: bfebcd73ef6ffe0ee30489227f0330c39064b674
 
+  - package: github.com/MichaelTJones/pcg
```
benraskin92: Can probably remove this dependency - will take a look.

docker/jaeger/Dockerfile
```diff
@@ -0,0 +1 @@
+FROM jaegertracing/all-in-one:1.9
```
prateek: do we need this file? as in, can we just refer to the upstream image in the docker-compose?

benraskin92: Maybe.. let me test.

scripts/development/m3_stack/README.md
```diff
@@ -9,19 +9,42 @@ This docker-compose file will setup the following environment:
...
 ## Grafana
 
 Use Grafana by navigating to `http://localhost:3000` and using `admin` for both the username and password. The M3DB dashboard should already be populated and working.
 
+## Jaeger
+
+To start Jaeger, you need to set the environment variable `START_JAEGER` to `true` when you run `start_m3.sh`.
+
+```
+START_JAEGER=true ./start_m3.sh
```
prateek: nit: should this be `USE_JAEGER` instead? cause we do have to stop it in `stop_m3.sh`

scripts/development/m3_stack/docker-compose.yml
```diff
@@ -107,5 +107,23 @@ services:
     networks:
       - backend
     image: m3grafana:latest
+  jaeger:
+    networks:
+      - backend
+    build:
```
prateek: more a question: can you skip the `build:` section here?

scripts/development/m3_stack/m3dbnode.yml
```diff
@@ -2,6 +2,15 @@ db:
   logging:
     level: info
 
+  tracing:
```
prateek: hm are you going to leave this in? if so, why does the readme need the instructions for config updates?

scripts/development/m3_stack/start_m3.sh
```diff
@@ -10,7 +10,21 @@ if [[ "$FORCE_BUILD" = true ]] ; then
...
+    # rely on 204 status code until https://github.com/jaegertracing/jaeger/issues/1450 is resolved.
+    JAEGER_STATUS=$(curl -s -o /dev/null -w '%{http_code}' localhost:14269)
+    if [ $JAEGER_STATUS -ne 204 ]; then
+        echo "Jaeger could not start"
+        return 1
+    fi
+fi
+
 docker-compose -f docker-compose.yml up $DOCKER_ARGS m3coordinator01
+# docker-compose -f docker-compose.yml up --build -d --renew-anon-volumes m3db_seed
```
prateek: does this need to be removed?

src/dbnode/config/m3dbnode-local-etcd.yml
```diff
@@ -98,3 +98,9 @@ db:
           initialCluster:
               - hostID: m3db_local
                 endpoint: http://127.0.0.1:2380
+
+  # Uncomment this to enable local jaeger tracing. See https://www.jaegertracing.io/docs/1.9/getting-started/
+  # for quick local setup (which this config will send data to).
+
+  tracing:
```
prateek: hm, should this be commented out?

prateek: also, should it be in all the config files?

scripts/development/m3_stack/start_m3.sh
benraskin92: Yes - sorry will remove before i land.

scripts/development/m3_stack/m3dbnode.yml
benraskin92: I changed the logic so that if it errors creating the tracer, we just log it and default to a noop - so it is safe to leave it in here.

benraskin92: In case someone wants to change the sampling rate, etc.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -66,6 +69,9 @@ const (
 	maxSegmentArrayPooledLength = 32
 	// Any pooled error slices that grow beyond this capcity will be thrown away.
 	writeBatchPooledReqPoolMaxErrorsSliceSize = 4096
+
+	fetchTaggedSpanID = "rpc_fetch_tagged"
```
prateek: We should do some standardisation of the names we use when creating spans. The ideal is to have a easy 1-1 mapping from name to the code, without requiring users to look at the code itself (to make it easy to read the trace output in jaeger). 

I'd suggest a convention like: `packageName.objectName.Method` or `packageName.Method` in case there isn't an object. where packageName is abbreviated enough to make it clear without being too verbose. So in this case, it'd be `tchannelthrift/node.Service.Query` 

What do you think?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -66,6 +69,9 @@ const (
...
+	querySpanID       = "rpc_query"
```
prateek: another point (this one I'm not as sure of) - how would you feel about moving all these constants as exported in a separate package (say `src/dbnode/tracepoint`). Would help discoverability. And also provide a nice single place to document the convention of naming these things.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -258,23 +264,53 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
 }
 
 func (s *service) Query(tctx thrift.Context, req *rpc.QueryRequest) (*rpc.QueryResult_, error) {
+	var sp opentracing.Span
+	result, sp, err := s.query(tctx, req)
+
+	if sp != nil {
```
prateek: hm, can you rewire the span creation to give you back a `NoOpSpan` when we're not tracing? that way we'd be able to avoid `if sp != nil` everywhere

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -258,23 +264,53 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
...
+func (s *service) query(tctx thrift.Context, req *rpc.QueryRequest) (*rpc.QueryResult_, opentracing.Span, error) {
+	var (
+		ctx context.Context
+		sp  opentracing.Span
+		ok  bool
+	)
+
+	ctx = tchannelthrift.Context(tctx)
+
+	ctx, sp, ok = ctx.StartTraceSpan(querySpanID)
```
prateek: hm, how about starting the span in the `Query()` method. just to minimise the amount of observability code we're putting side by side with execution logic

prateek: hm why do we need to return ok from `StartTraceSpan` ?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -399,23 +435,59 @@ func (s *service) readDatapoints(
 }
 
 func (s *service) FetchTagged(tctx thrift.Context, req *rpc.FetchTaggedRequest) (*rpc.FetchTaggedResult_, error) {
+	result, sp, err := s.fetchTagged(tctx, req)
```
prateek: same nits as the ones in `Query()/query()`

src/dbnode/network/server/tchannelthrift/node/service_test.go
```diff
@@ -1133,6 +1140,12 @@ func TestServiceFetchTagged(t *testing.T) {
 		assert.Equal(t, expectHead, seg.Merged.Head)
 		assert.Equal(t, expectTail, seg.Merged.Tail)
 	}
+
+	sp.Finish()
+	spans := mtr.FinishedSpans()
+	require.Len(t, spans, 2)
+	assert.Equal(t, fetchTaggedSpanID, spans[0].OperationName)
```
prateek: v nice

src/dbnode/server/server.go
```diff
@@ -90,6 +93,7 @@ const (
 	bgProcessLimitInterval           = 10 * time.Second
 	maxBgProcessLimitMonitorDuration = 5 * time.Minute
 	filePathPrefixLockFile           = ".lock"
+	serviceName                      = "m3dbnode"
```
prateek: hm should this be config driven?

src/dbnode/server/server.go
prateek: yeah, i'll try to get around to this soon. want all our code to use zap instead of m3/x/log

src/dbnode/server/server.go
```diff
@@ -81,7 +82,9 @@ import (
 
 	"github.com/coreos/etcd/embed"
 	"github.com/coreos/pkg/capnslog"
+	opentracing "github.com/opentracing/opentracing-go"
 	"github.com/uber-go/tally"
+	xnetcontext "golang.org/x/net/context"
```
prateek: how about using stdlib "context" here?

src/dbnode/server/server.go
```diff
@@ -183,6 +194,26 @@ func Run(runOpts RunOptions) {
...
+		tracer      opentracing.Tracer
+		traceCloser io.Closer
+	)
+
+	if cfg.Tracing == nil {
+		tracer = opentracing.NoopTracer{}
+		logger.Infof("tracing disabled for %s; set `tracing.backend` to enable", serviceName)
+	} else {
+		// setup tracer
+		tracer, traceCloser, err = cfg.Tracing.NewTracer(serviceName, scope, zapLogger)
```
prateek: hm should this be given a subscope to keep metrics namespaces clean? (not sure if jaeger's instrumentation is doing so)

src/dbnode/storage/namespace.go
```diff
@@ -21,6 +21,7 @@
 package storage
 
 import (
+	// stdctx "context"
```
prateek: does this need to be removed?

src/query/server/server.go
```diff
@@ -148,7 +148,7 @@ func Run(runOpts RunOptions) {
 	defer traceCloser.Close()
 
 	if _, ok := tracer.(opentracing.NoopTracer); ok {
-		logger.Info("tracing disabled; set `tracing.backend` to enable")
+		logger.Info("tracing disabled for m3query; set `tracing.backend` to enable")
```
prateek: hm looks like this tracer creation logic can be shared between m3dbnode and query. mind refactoring

src/dbnode/server/server.go
benraskin92: Yeah we can and just default to `m3dbnode` if none is provided.

src/dbnode/storage/database.go
```diff
@@ -72,6 +75,10 @@ var (
 	errWriterDoesNotImplementWriteBatch = errors.New("provided writer does not implement ts.WriteBatch")
 )
 
+const (
+	dbQuerySpanID = "db_query_ids"
```
prateek: same suggestion re naming convention/organization

src/dbnode/storage/namespace.go
```diff
@@ -56,6 +60,10 @@ var (
 	errNamespaceIndexingDisabled = errors.New("namespace indexing is disabled")
 )
 
+const (
+	nsQuerySpanID = "namespace_query_ids"
```
prateek: here too

src/query/server/server.go
benraskin92: mind if i do this in a follow up pr? i still wanna remove the opentracing stuff that is in query anyway (cause now it's in src/x)

src/dbnode/config/m3dbnode-local-etcd.yml
benraskin92: Yes - but instead of this, I'd rather just have an annotated config section like http://m3db.github.io/m3/query_engine/config/annotated_config.yaml. I can add them to all of the configs but I think that's overkill.

src/dbnode/network/server/tchannelthrift/node/service.go
benraskin92: let me read a bit about opentracing/jaeger naming conventions.

src/dbnode/network/server/tchannelthrift/node/service.go
benraskin92: We technically don't but I guess it's a little cleaner than doing a `sp != nil`

src/dbnode/server/server.go
benraskin92: Looks like it just wraps it. want me to add a `.SubScope("jaeger")`?
```
// Wrap takes a tally Scope and returns jaeger-lib metrics.Factory.
func Wrap(scope tally.Scope) metrics.Factory {
	return &factory{
		tally: scope,
	}
}
```

src/dbnode/network/server/tchannelthrift/node/service.go
benraskin92: how about keeping these names, but including tags? for example, this one would still be called `rpc_fetch_tagged`, but we could have a series of tags like:
```
method:query
package:node
component:tchannelthrift
```

src/dbnode/network/server/tchannelthrift/node/service.go
benraskin92: If we go with the tags approach, i'd say leave it in the individual files.

scripts/development/m3_stack/docker-compose.yml
benraskin92: It appears so..

src/dbnode/server/server.go
prateek: yep sounds good

benraskin92:
Screen Shot 2019-04-08 at 7 08 40 PM

scripts/development/m3_stack/docker-compose.yml
```diff
@@ -107,5 +107,20 @@ services:
     networks:
       - backend
     image: m3grafana:latest
+  jaeger:
+    networks:
+      - backend
+    image: jaegertracing/all-in-one:1.9
```
prateek: n1

scripts/development/m3_stack/m3dbnode.yml
prateek: fair enough.

scripts/development/m3_stack/start_m3.sh
prateek: cool, please put a `TODO(braskin)` or something else in case you're going to keep it around.

src/dbnode/config/m3dbnode-local-etcd.yml
```diff
@@ -98,3 +98,9 @@ db:
           initialCluster:
               - hostID: m3db_local
                 endpoint: http://127.0.0.1:2380
+
+  # Uncomment this to enable local jaeger tracing. See https://www.jaegertracing.io/docs/1.9/getting-started/
```
prateek: nit: `un-comment the lines below to enable Jaeger tracing.`

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -258,13 +262,38 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
...
 
 func (s *service) Query(tctx thrift.Context, req *rpc.QueryRequest) (*rpc.QueryResult_, error) {
+	var (
+		ctx context.Context
+		sp  opentracing.Span
+		ok  bool
+	)
+
+	ctx = tchannelthrift.Context(tctx)
+	ctx, sp, ok = ctx.StartTraceSpan(tracepoint.QuerySpanID)
```
prateek: why do we still need the `ok`? considering you're giving back a NoopSpan when !ok - can you just avoid the `if check` and always do the `sp.LogFields(...)` jazz?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -258,13 +262,38 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
...
+			xopentracing.Time("start", time.Unix(0, req.RangeStart)),
+			xopentracing.Time("end", time.Unix(0, req.RangeStart)),
+		)
+	}
+
+	result, err := s.query(ctx, req)
+	if err != nil {
+		sp.LogFields(opentracinglog.Error(err))
+	}
+	sp.Finish()
```
prateek: nit: probably easiest to follow the `defer sp.Finish()` pattern. i.e. issue the defer right after constructing the span.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -399,13 +428,40 @@ func (s *service) readDatapoints(
...
 
 func (s *service) FetchTagged(tctx thrift.Context, req *rpc.FetchTaggedRequest) (*rpc.FetchTaggedResult_, error) {
+	var (
+		ctx context.Context
+		sp  opentracing.Span
+		ok  bool
+	)
+
+	ctx = tchannelthrift.Context(tctx)
+	ctx, sp, ok = ctx.StartTraceSpan(tracepoint.FetchTaggedSpanID)
```
prateek: same thing

src/x/context/context.go
```diff
@@ -316,7 +320,7 @@ func (c *ctx) spanIsSampled(sp opentracing.Span) bool {
 func (c *ctx) StartTraceSpan(name string) (Context, opentracing.Span, bool) {
 	goCtx, exists := c.GoContext()
 	if !exists || c.checkedAndNotSampled {
-		return c, nil, false
+		return c, noopTracer.StartSpan(name), false
```
prateek: I think you can change the signature to `StartTraceSpan(name string) (Context, opentracing.Span)` to make this easier to use

src/query/server/server.go
prateek: sure that works

src/dbnode/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,42 @@
...
+
+const (
+	// FetchTaggedSpanID is the operation name for the tchannelthrift FetchTagged path
+	FetchTaggedSpanID = "tchannelthrift/node.service.FetchTagged"
+
+	// QuerySpanID is the operation name for the tchannelthrift Query path
+	QuerySpanID = "tchannelthrift/node.service.Query"
+
+	// DBQueryIDs is the operation name for the db QueryIDs path
+	DBQueryIDs = "storage.db.QueryIDs"
```
prateek: nit: first two vars in this file end with `...SpanID`, the last two don't. Mind sticking to one convention.

src/dbnode/tracepoint/tracepoint.go
```diff
@@ -0,0 +1,42 @@
...
+
+// The tracepoint package is used to store operation names for tracing throughout dbnode.
+// The naming convention is as follows:
+
+// `packageName.objectName.method`
+
+// If there isn't an object, use `packageName.method`.
+
+const (
+	// FetchTaggedSpanID is the operation name for the tchannelthrift FetchTagged path
```
prateek: super nit: end comments with a period.

src/dbnode/network/server/tchannelthrift/node/service.go
robskillington: Hey so we actually want to know ahead of time so we don't call `sp.LogFields(...)` etc in the hot path (because a lot of the time `blah.String()` or other things done during the call to `LogFields()` cause allocations, even the call itself since it is varargs allocates a slice each time).

src/dbnode/network/server/tchannelthrift/node/service.go
robskillington: In this method since it is not in the super hot path that makes sense.  If this was a hot high frequency path I would say avoiding defer is a good idea.

src/dbnode/network/server/tchannelthrift/node/service.go
robskillington: Just wanted to leave comment about avoiding the string casts and varargs slice alloc for the call to `LogFields(...)` when trace is not being sampled.

robskillington: In low frequency methods it probably doesn't matter all that much, but we may as well establish the pattern so that in high frequency hot methods we're avoiding allocs by default with the pattern we establish.

src/dbnode/network/server/tchannelthrift/node/service_test.go
```diff
@@ -22,11 +22,14 @@ package node
...
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"sort"
 	"testing"
 	"time"
 
+	"github.com/m3db/m3/src/dbnode/tracepoint"
+
```
robskillington: nit: Unnecessary empty endline?

src/dbnode/server/server.go
robskillington: Agreed, it is time to kill the x logger.

robskillington: PS we should really add the ZapLogger to instrument options (or pull the Zap logger from there?  Does it not already get created in instrument options by default?).

src/dbnode/storage/namespace.go
```diff
@@ -601,17 +605,43 @@ func (n *dbNamespace) QueryIDs(
...
+		err := errIndexNotBootstrappedToRead
+		if sp != nil {
+			sp.LogFields(opentracinglog.Error(err))
+		}
 		return index.QueryResult{},
-			xerrors.NewRetryableError(errIndexNotBootstrappedToRead)
+			xerrors.NewRetryableError(err)
 	}
 
 	res, err := n.reverseIndex.Query(ctx, query, opts)
```
robskillington: Hm, maybe should also `if err != nil && sp != nil { sp.LogFields(opentracinglog.Error(err)) }` here?

src/dbnode/tracepoint/tracepoint.go
robskillington: Probably remove the `...SpanID` suffix, since someone will forget it at some point hah.

src/x/context/context.go
robskillington: Hey, as per earlier comment I left - knowing whether it's being sampled or not is important so for hot path code we can avoid doing anything that might cause an allocated unless the trace is absolutely being sampled.  Hence the `bool` is important.

scripts/development/m3_stack/m3dbnode.yml
andrewmains12: > I changed the logic so that if it errors creating the tracer, we just log it and default to a noop - so it is safe to leave it in here.

I'm not sure I agree with that change; the most likely reason for that is invalid configuration, which it seems worthwhile to surface. Since it's analogous, do we handle failure to initialize tally etc in the same way? Would be good to be consistent.

scripts/development/m3_stack/start_m3.sh
andrewmains12: Perhaps a nit, but is there not an explicit way to state the dependency in `docker-compose.yml`? https://docs.docker.com/compose/compose-file/ mentions `depends_on`

src/dbnode/network/server/tchannelthrift/node/service.go
prateek: Hey fair enough. I tried out `sp.LogFields(...)` with 8 fields logged, it's (~300ns; 500B alloc'd) per call (https://play.golang.org/p/32Kdr-Ha4Ep). Even if we have 100 of spans created per RPC call, that's ~30us. It won't be a factor in most rpc paths (indexing/writes) but could be more significant if we were to instrument all the 10K ID fetches of a FetchBatch call. 

Would you be opposed to keeping two methods: one which does the instrumentation by default (i.e. no `ok` returned) and one which does? Can use the appropriate method depending on the call site.

andrewmains12: +1 -- would be great to avoid boilerplate where possible (though ofc not at the expense of perf where perf is needed)

robskillington: Sure, having two methods makes sense. 

Maybe `StartTraceSpan(...)` and `StartSampledTraceSpan(...)` where the second one is the one that returns the `bool` at the end?

scripts/development/m3_stack/m3dbnode.yml
benraskin92: So I definitely log the error, but it would be kind of bad if m3db couldn't start if your jaeger host was down for some reason.

andrewmains12: Fair enough! My assumption was that the jaeger client would come up even if the host is down (and then just fail to log), and that it would error primarily in cases of misconfiguration. However, I could see the argument for not going down if jaeger for instance changes their configuration in a backwards incompatible way (though then the user upon upgrading will just have to notice that their traces are no longer working).

I'm good with this though; thanks for discussion!

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -255,13 +259,35 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
 }
 
 func (s *service) Query(tctx thrift.Context, req *rpc.QueryRequest) (*rpc.QueryResult_, error) {
+	var (
```
prateek: don't need these allocs ahead of time now?

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -255,13 +259,35 @@ func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_
...
+		ctx context.Context
+		sp  opentracing.Span
+	)
+
+	ctx = tchannelthrift.Context(tctx)
```
prateek: could this be:
```
ctx, sp := tchannelthrift.Context(tctx).StartTraceSpan(tracepoint.Query)
```

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -396,13 +422,37 @@ func (s *service) readDatapoints(
 }
 
 func (s *service) FetchTagged(tctx thrift.Context, req *rpc.FetchTaggedRequest) (*rpc.FetchTaggedResult_, error) {
+	var (
+		ctx context.Context
+		sp  opentracing.Span
+	)
+
+	ctx = tchannelthrift.Context(tctx)
+	ctx, sp = ctx.StartTraceSpan(tracepoint.FetchTagged)
```
prateek: same

src/dbnode/server/server.go
```diff
@@ -21,6 +21,7 @@
 package server
 
 import (
+	stdlib "context"
```
prateek: nit: need a name that's clearer about defining the import "stdcontext" or "stdlibcontext" or "context" instead of "stdlib" please

src/dbnode/storage/database.go
```diff
@@ -708,8 +712,21 @@ func (d *db) QueryIDs(
 	query index.Query,
 	opts index.QueryOptions,
 ) (index.QueryResult, error) {
+	var sp opentracing.Span
+	ctx, sp = ctx.StartTraceSpan(tracepoint.DBQueryIDs)
```
prateek: does `ctx, sp := ...` get :( here?

src/dbnode/storage/database_test.go
```diff
@@ -21,13 +21,16 @@
 package storage
 
 import (
+	stdlib "context"
```
prateek: same

src/dbnode/storage/database_test.go
```diff
@@ -21,13 +21,16 @@
...
 	"errors"
 	"fmt"
 	"sort"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/m3db/m3/src/dbnode/tracepoint"
```
prateek: import order. interestingly, did metalint not pick this up?

src/dbnode/storage/namespace_test.go
```diff
@@ -21,12 +21,15 @@
 package storage
 
 import (
+	stdlib "context"
```
prateek: here too

src/dbnode/storage/namespace_test.go
```diff
@@ -21,12 +21,15 @@
...
 	"errors"
 	"fmt"
 	"sync"
 	"testing"
 	"time"
 
+	"github.com/m3db/m3/src/dbnode/tracepoint"
```
prateek: here too

src/x/context/context.go
```diff
@@ -313,10 +317,10 @@ func (c *ctx) spanIsSampled(sp opentracing.Span) bool {
 	return true
 }
 
-func (c *ctx) StartTraceSpan(name string) (Context, opentracing.Span, bool) {
+func (c *ctx) StartSampledTraceSpan(name string) (Context, opentracing.Span, bool) {
```
prateek: nit: add comments to exported functions. write a blurb bout using StartSampledTraceSpan v the other one.

prateek: oh nvm, saw you have them in the interface def

benraskin92: I thought the convention was to just comment on the interface?

src/dbnode/storage/database_test.go
benraskin92: weird - ill look into it

src/x/context/types.go
```diff
@@ -75,8 +75,12 @@ type Context interface {
 	SetGoContext(stdctx.Context)
 
 	// StartTraceSpan starts a new span and returns a child ctx
-	// if the span is being sampled.
-	StartTraceSpan(name string) (Context, opentracing.Span, bool)
+	StartTraceSpan(string) (Context, opentracing.Span)
```
prateek: nit: period after comment.

PR #1510, opened 2019-03-28
add schema to namespace metadata options

Haijuncao:


src/dbnode/generated/proto/namespace/namespace.proto
```diff
@@ -15,6 +15,12 @@ message IndexOptions {
     int64 blockSizeNanos = 2;
 }
 
+message SchemaOption {
```
prateek: nit: `SchemaOptions` to replicate the convention in the file

src/dbnode/generated/proto/namespace/namespace.proto
```diff
@@ -15,6 +15,12 @@ message IndexOptions {
...
+    uint32 version = 1;
```
prateek: nit: please align the three lines on `=`

src/dbnode/generated/proto/namespace/namespace.proto
```diff
@@ -24,6 +30,7 @@ message NamespaceOptions {
     RetentionOptions retentionOptions = 6;
     bool snapshotEnabled              = 7;
     IndexOptions indexOptions         = 8;
+    repeated SchemaOption schemaOptions = 9;
```
prateek: nit: mind adding a comment about the intent behind `repeated` here. Took me a second to figure out that you were retaining the history of all schemas. 

also, align on `=` within the struct please.

src/dbnode/storage/namespace/types.go
```diff
@@ -84,6 +85,10 @@ type Options interface {
 
 	// IndexOptions returns the IndexOptions.
 	IndexOptions() IndexOptions
+
+	SetSchema(schema []Schema) Options
```
prateek: nit: comments for both methods, and mind using plural form - i.e. `SetSchemas` and `Schemas`

src/dbnode/storage/namespace/types.go
```diff
@@ -104,6 +109,15 @@ type IndexOptions interface {
 	BlockSize() time.Duration
 }
 
+// Schema represents namespace schema
```
prateek: nit: end comments with a period.

src/dbnode/storage/namespace/types.go
```diff
@@ -104,6 +109,15 @@ type IndexOptions interface {
...
+type Schema interface {
+	Version() uint32
```
prateek: nit: mind adding comments for all methods

src/dbnode/storage/namespace/types.go
```diff
@@ -104,6 +109,15 @@ type IndexOptions interface {
...
+	Equal(Schema) bool
+	Get() *desc.MessageDescriptor
```
prateek: Not a fan of directly exposing a dependency on a library's exported type here if we can help it. 

How would you feel about exposing types corresponding to the generic [protobuf descriptor](https://github.com/protocolbuffers/protobuf/blob/199d82fde1734ab5bc931cd0de93309e50cd7ab9/src/google/protobuf/descriptor.proto), or the `.proto` bytes instead?

src/dbnode/storage/namespace/schema_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	nsproto "github.com/m3db/m3/src/dbnode/generated/proto/namespace"
+	testproto "github.com/m3db/m3/src/dbnode/generated/proto/schema_test"
+	"github.com/stretchr/testify/require"
+	"github.com/m3db/m3/src/dbnode/storage/namespace"
+)
+
+func getTestSchemaOptions() []*nsproto.SchemaOption {
+	tm := &testproto.TestMessage{}
+	def, _ := tm.Descriptor()
```
prateek: nit: check error here

src/dbnode/storage/namespace/options_test.go
```diff
@@ -51,6 +53,23 @@ func TestOptionsEqualsIndexOpts(t *testing.T) {
 	require.False(t, o2.Equal(o1))
 }
 
+func getTestSchemaOptions() []*nsproto.SchemaOption {
+	tm := &testproto.TestMessage{}
+	def, _ := tm.Descriptor()
```
prateek: nit: check error here

glide.yaml
```diff
@@ -60,6 +60,9 @@ import:
   - package: github.com/gogo/protobuf
     version: ^1
 
+  - package: github.com/jhump/protoreflect
+    version: ^1.2
```
richardartoul: Can you pin this commit please: 6feecbc20c4c369d6494c59ac5aee5d0df3c9d80

Included an important bug fix and my new method: https://github.com/jhump/protoreflect/commit/6feecbc20c4c369d6494c59ac5aee5d0df3c9d80

src/dbnode/generated/proto/namespace/namespace.pb.go
```diff
@@ -132,6 +133,38 @@ func (m *IndexOptions) GetBlockSizeNanos() int64 {
 	return 0
 }
 
+type SchemaOption struct {
```
richardartoul: Can you make it `SchemaOptions` to make it consistent with the other structs

src/dbnode/generated/proto/namespace/namespace.proto
```diff
@@ -15,6 +15,12 @@ message IndexOptions {
     int64 blockSizeNanos = 2;
 }
 
+message SchemaOption {
+    uint32 version = 1;
+    bytes  definition = 2;
```
richardartoul: Why is there a definition and a message?

src/dbnode/generated/proto/namespace/namespace.proto
richardartoul: I wonder if it might be better to move the `repeated` into the `SchemaOptions` itself? In the future we'd like to support multiple schemas per namespace so it might be nice if in the future we only need to modify the `SchemaOptions` struct instead of having to touch the `NamespaceOptions` struct too

src/dbnode/generated/proto/schema_test/test1.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+package schematest;
+
+message TestMessage {
```
richardartoul: Could you throw in some repeated fields, a nested message, and a map field as well?

src/dbnode/generated/proto/namespace/namespace.proto
richardartoul: We could actually make this a little more future proof with very little extra work if it became

```protobuf
message NamespaceSchemas {
    map<string, SchemaHistory> history = 1;
}

message SchemaHistory {
    repeated Schema schemas = 1;
}

message Schema {
    uint32 version = 1;
    bytes definition = 2;
    string message = 3;
}
```

and then for the initial implementation we can just always put one schema with the name "default" in the map. Its a little bit more work upfront but it will save us a ton of time later not having to message with this again and will also make upgrading existing clusters way easier

src/dbnode/storage/namespace/convert_prop_test.go
```diff
@@ -34,6 +34,7 @@ import (
 	"github.com/leanovate/gopter"
 	"github.com/leanovate/gopter/gen"
 	"github.com/leanovate/gopter/prop"
+
```
richardartoul: do you want to add a new line here?

src/dbnode/generated/proto/namespace/namespace.proto
richardartoul: I guess technically within a given proto definition you can have multiple messages so we have two options of supporting multiple schemas:

1. One proto definition per schema
2. One proto definition per namespace with one message per schema

Any idea what the tradeoffs between the two implementations are? I don't know off the top of my head

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,169 @@
...
+	if o == nil {
+		return nil, nil
+	}
+
+	fdp, err := decodeProtoFileDescriptor(o.Definition)
+	if err != nil {
+		return nil, xerrors.Wrap(err, "failed to decode schema file descriptor")
+	}
+
+	fd, err := desc.CreateFileDescriptor(fdp)
```
richardartoul: Can we add a check here (or wherever it makes sense) that the descriptor `.IsProto3()`

richardartoul: I know we'll have that constraint in the API but would be good to have it here too

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, xerrors.Wrap(err, "failed to create schema file descriptor")
+	}
+	mds := fd.GetMessageTypes()
+	if len(mds) == 0 {
+		return nil, xerrors.Wrap(errInvalidSchema, "schema does not contain any messages")
+	}
+
+	var found *desc.MessageDescriptor
+	for _, md := range mds {
+		if strings.EqualFold(md.GetName(), o.Message) {
```
richardartoul: So we're storing the name of the message that we care about in etcd along with the entire protobuf definition (which could contain other messages)? Seems like if we're doing that we should just bake the idea of multiple messages into the protobuf in etcd even if we only use one for now (always look for message called "Default" or w/e)

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, xerrors.Wrapf(errInvalidSchema, "schema does not contain %v", o.Message)
+	}
+
+	return &schema{fdp: o.Definition, md: found, version: o.Version}, nil
+}
+
+// decodeProtoFileDescriptor decodes the bytes of proto file descriptor.
+// proto file descriptor is proto encoded and gzipped, decode reverse the process.
+func decodeProtoFileDescriptor(fdb []byte) (*dpb.FileDescriptorProto, error) {
+	raw, err := decompress(fdb)
```
richardartoul: Is the gzipping out of our control? seems like it might be nice to have it human readable even in etcd but maybe its not a big deal

src/dbnode/storage/namespace/types.go
richardartoul: The .proto bytes aren't useful to any other parts of the code as they'll have to re-implement all the parsing code and error checking so lets not do that.

Not sure what you mean about the generic protobuf descriptor since you just linked to a .proto text file?

Regardless of what we decide we should settle on something thats actually usable by the rest of the code-base even if that means we have to expose a dependency here (we could wrap it in a type alias or struct or something so that the dependency is only referenced from one package and then if we ever need to change it we can just change it there)

src/dbnode/generated/proto/namespace/namespace.proto
Haijuncao: 1 gives us more flexibility. Two schema does not have two share the same file descriptor, if they do, it is totally up to the user.

I think file descriptor can be more useful in nested message situation, there are multiple messages, one of them is the main one, others are just used by the main. in this case, user need to tell us which one is the main message.

src/dbnode/generated/proto/schema_test/test1.proto
Haijuncao: ack

src/dbnode/generated/proto/namespace/namespace.proto
Haijuncao: I will change it to name. it tells us which message inside the file descriptor in case there are multiple messages (e.g. nested)

glide.yaml
Haijuncao: sure

src/dbnode/storage/namespace/types.go
Haijuncao: ok, I will alias it.

src/dbnode/storage/namespace/types.go
Haijuncao: sure

src/dbnode/generated/proto/namespace/namespace.proto
Haijuncao: agree, will make the definition future proof

src/dbnode/generated/proto/namespace/namespace.proto
Haijuncao: > We could actually make this a little more future proof with very little extra work if it became
> 
> ```protobuf
> message NamespaceSchemas {
>     map<string, SchemaHistory> history = 1;
> }
> 
> message SchemaHistory {
>     repeated Schema schemas = 1;
> }
> 
> message Schema {
>     uint32 version = 1;
>     bytes definition = 2;
>     string message = 3;
> }
> ```
> 
> and then for the initial implementation we can just always put one schema with the name "default" in the map. Its a little bit more work upfront but it will save us a ton of time later not having to message with this again and will also make upgrading existing clusters way easier

agree, will make it future proof.

src/dbnode/storage/namespace/schema.go
Haijuncao: multiple messages are for nested message. there is only one main message for series' complex type. the message name is user defined.

src/dbnode/storage/namespace/schema.go
Haijuncao: sure

src/dbnode/storage/namespace/options_test.go
Haijuncao: It does not return error, it always return []{0}, not sure what it is
it is generated protobuf go binding:

` func (*TestMessage) Descriptor() ([]byte, []int) { return fileDescriptorTest1, []int{0} }`

figure it should be ok not to worry about this.

src/dbnode/generated/proto/namespace/namespace.proto
Haijuncao: > I guess technically within a given proto definition you can have multiple messages so we have two options of supporting multiple schemas:
> 
> 1. One proto definition per schema
> 2. One proto definition per namespace with one message per schema
> 
> Any idea what the tradeoffs between the two implementations are? I don't know off the top of my head

I think with proto import, I might have to go with option2, it also map nicely into user repo.

src/dbnode/storage/namespace/schema.go
Haijuncao: This is defined by protobuf. if you look at generated go binding, you will see each file has descriptor bytes embedded in the code already. 

I am not confident on the parsing code, and not quite want to do parsing on our side. I will let the protoc do the parsing and simply have them parse me the equivalent descriptor.

For readability, we can easily rebuild a compact/canonical form of proto by invoking String() method.

src/dbnode/generated/proto/namespace/namespace.proto
Haijuncao: > nit: mind adding a comment about the intent behind `repeated` here. Took me a second to figure out that you were retaining the history of all schemas.
> 
> also, align on `=` within the struct please.

regarding align =  in proto, do you have an easy way to do this format? I am intelliJ ide with proto plugin, it does not seem to do it for me. 

also, what's the rational for align =, golang format does not do that.

src/dbnode/storage/namespace/types.go
Haijuncao: > ok, I will alias it.

forget golang has no alias, it is another type all together. I don't want user to have to assert it back to  *desc.MessageDescriptor everything they uses it.

src/dbnode/generated/proto/namespace/schema.proto
```diff
@@ -0,0 +1,29 @@
...
+
+// SchemaMeta contains schema metadata.
+message SchemaMeta {
+    // message name as it appears in proto file descriptor.
+    string messageName = 1;
+}
+
+// SchemaOptions contains schema options for a namespace.
+message SchemaOptions {
+    // repo contains a history of deployed schema defintions in form of proto file descriptors.
```
richardartoul: 'defintions' -> 'definitions'

'in form' -> 'in the form'

src/dbnode/generated/proto/namespace/schema.proto
```diff
@@ -0,0 +1,29 @@
...
+    FileDescriptorRepo repo = 1;
+    // schemas contains a set of named schemas for a namespace.
+    // a namespace can have multiple schemas.
+    // series identifies its schema by schema name (the map key) and a version (repo version).
+    map<string, SchemaMeta> schemas = 2;
+}
+
+// FileDescriptorRepo is a history of FileDescriptorSet sorted by version in descending order.
+message FileDescriptorRepo {
```
richardartoul: Not a huge fan of the use of the word `Repo` here but I also don't have a better suggestion :P

src/dbnode/generated/proto/namespace/schema.proto
```diff
@@ -0,0 +1,29 @@
...
+    string messageName = 1;
+}
+
+// SchemaOptions contains schema options for a namespace.
+message SchemaOptions {
+    // repo contains a history of deployed schema defintions in form of proto file descriptors.
+    FileDescriptorRepo repo = 1;
+    // schemas contains a set of named schemas for a namespace.
+    // a namespace can have multiple schemas.
+    // series identifies its schema by schema name (the map key) and a version (repo version).
```
richardartoul: I think the last sentence might be a little more clear if you leave out any references to a "series" and just say: "A schema is uniquely identified by its name (the map key) and a version number (the repo version)."

src/dbnode/generated/proto/namespace/schema.proto
```diff
@@ -0,0 +1,29 @@
...
+    map<string, SchemaMeta> schemas = 2;
```
richardartoul: So we'll have a history of all the binary descriptors, but only the latest "set" of message names? Does it make sense to track the history of both together or do you think we don't need that?

Also, how come we're tracking the map separately from the descriptors? Seems like you could generate a list of all the messages from the descriptors itself. Is it so we could add additional metadata thats not included in the proto syntax? That makes sense to me I think

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,212 @@
...
+		return nil
+	}
+	return sr.(*schemaRegistry).options
+}
+
+// LoadSchemaRegistry loads schema registry from SchemaOptions proto.
+func LoadSchemaRegistry(options *nsproto.SchemaOptions) (SchemaRegistry, error) {
+	if options.GetRepo() == nil || len(options.GetRepo().GetHistory()) == 0 ||
+		len(options.GetSchemas()) == 0 {
+		return nil, nil
```
richardartoul: Would it be less confusing if we returned an "empty" schema registry instead?

src/dbnode/generated/proto/namespace/schema.proto
Haijuncao: all right, I will use history.

src/dbnode/generated/proto/namespace/schema.proto
Haijuncao: sounds good.

src/dbnode/generated/proto/namespace/schema.proto
Haijuncao: yes, it is for additional metadata that is on the series level (not namespace), similar to cstar table level metadata (properties), such as ttl, cp strategy, cache strategy, etc.

src/dbnode/generated/proto/namespace/schema.proto
Haijuncao: on a second thought, it is more for series level metadata, schema definition is one of them (schema definition references the proto message name in the latest file descriptor repo).

Haijuncao:
> Should we add a `Validate()` method in this P.R or a different one? We'll need some way to validate that it doesn't contain any feature we don't support

yes, I image we add a cli that "deploy" schema to (etcd), this cli should validate.

src/dbnode/storage/namespace/schema.go
Haijuncao: agree, changed default from nil to empty

src/dbnode/generated/proto/namespace/schema.proto
Haijuncao: > So we'll have a history of all the binary descriptors, but only the latest "set" of message names? Does it make sense to track the history of both together or do you think we don't need that?
> 
> Also, how come we're tracking the map separately from the descriptors? Seems like you could generate a list of all the messages from the descriptors itself. Is it so we could add additional metadata thats not included in the proto syntax? That makes sense to me I think

map (schemas) is not used to map proto message name to proto message descriptor.  it is used to map schema id(name) to proto message name.

Map is append only, we don't delete schema id or reuse it (to map to another message descriptor). this can be enforced in schema cli.

When client writes series, it identifies series schema via schema id.

src/dbnode/generated/proto/namespace/schema.proto
```diff
@@ -0,0 +1,26 @@
...
+message SchemaHistory {
+    // versions is a list of FileDescriptorSet sorted by version in ascending order.
+    repeated FileDescriptorSet versions = 1;
+}
+
+// FileDescriptorSet is a set of proto file descriptors.
+message FileDescriptorSet {
+    // id identifies a deployed version of FileDescriptorSet.
+    string deployId = 1;
+    // prevId identifies the previous deploy id of FileDescriptorSet.
```
richardartoul: super nit: Want to just leave a comment saying the previous one can be inferred from the ordering but we add this as a precaution?

src/dbnode/generated/proto/schematest2/otherpkg.proto
```diff
@@ -0,0 +1,20 @@
+syntax = "proto3";
+package schematest2;
```
richardartoul: I don't really care personally, but if you like you can use the protoreflection library to parse the test schemas so you don't have to add all this generated code but w/e not a big deal

src/dbnode/storage/namespace/map_test.go
```diff
@@ -24,7 +24,6 @@ import (
 	"testing"
 
 	"github.com/m3db/m3x/ident"
-
```
richardartoul: Did you remove this newline? I think its required to prevent the linter from complaining (we separate our dependencies from others)

src/dbnode/storage/namespace/options.go
```diff
@@ -109,7 +111,9 @@ func (o *options) Equal(value Options) bool {
 		o.cleanupEnabled == value.CleanupEnabled() &&
 		o.repairEnabled == value.RepairEnabled() &&
 		o.retentionOpts.Equal(value.RetentionOptions()) &&
-		o.indexOpts.Equal(value.IndexOptions())
+		o.indexOpts.Equal(value.IndexOptions()) &&
+		(o.schemaReg == nil && value.SchemaRegistry() == nil ||
```
richardartoul: super nit: Could we wrap each of the && operations in a ()? I know it binds more tightly but makes me nervous

richardartoul:
```golang

((o.schemaReg == nil && value.SchemaRegistry() == nil) ||
(o.schemaReg != nil && o.schemaReg.Equal(value.SchemaRegistry()))
```

Also can `schemaReg` ever be nil? Seems like you always at least set it to `emptySchemaRegistry()`. Just a precaution?

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package namespace
+
+import (
+	"bytes"
+	"compress/gzip"
+	"errors"
+	"io/ioutil"
```
richardartoul: I dont know if our linter is working correctly or not, but the convention for imports is to break them into newline separate groups where the groups are:

1. Standard library imports

2. Any imports that start with "gitihub.com/m3db/..."

3. All other dependencies

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (s *schemaDescr) DeployId() string {
+	return s.deployId
+}
+
+func (s *schemaDescr) Equal(o SchemaDescr) bool {
+	if s == nil && o == nil {
+		return true
+	}
+	if s != nil && o == nil || s == nil && o != nil {
```
richardartoul: same comment about order of operations ()

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (s *schemaDescr) String() string {
+	if s.md == nil {
+		return ""
+	}
+	return s.md.String()
+}
+
+type schemaRegistry struct {
+	options  *nsproto.SchemaOptions
```
richardartoul: As far as I can tell this is unused. I'd prefer not to carry this raw proto message around everywhere. Can we remove this field?

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+		}
+		if !sd.Equal(osd) {
+			return false
+		}
+	}
+
+	return true
+}
+
+func (sr *schemaRegistry) Get(id string) (SchemaDescr, error) {
```
richardartoul: Might be a little cleaner to just have this return `(SchemaDescr, bool)` and let the caller construct an error if that is truly an error scenario otherwise they have to do some kind of error comparison to see if its `errSchemaNotFound` which isnt even exported

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+	sd, ok := sr.versions[id]
+	if !ok {
+		return nil, errSchemaNotFound
+	}
+	return sd, nil
+}
+
+func (sr *schemaRegistry) GetLatest() (SchemaDescr, error) {
```
richardartoul: same comment

src/dbnode/storage/namespace/schema.go
richardartoul: Oh are you keeping around so we can convert back to the original proto?

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+	return sr.Get(sr.latestId)
+}
+
+// toSchemaOptions returns the corresponding SchemaOptions proto for the provided SchemaRegistry
+func toSchemaOptions(sr SchemaRegistry) *nsproto.SchemaOptions {
```
richardartoul: if you make this a private method you won't have to do the type assertion

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+		return nil
+	}
+	return sr.(*schemaRegistry).options
+}
+
+func emptySchemaRegistry() SchemaRegistry {
+	return &schemaRegistry{options: nil, versions: make(map[string]*schemaDescr)}
+}
+
+// LoadSchemaRegistry loads schema registry from SchemaOptions proto.
```
richardartoul: I think convention in go is usually to begin a constructor with `New`

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	msgName := options.GetDefaultMessageName()
+	if len(msgName) == 0 {
+		return nil, xerrors.Wrap(errInvalidSchemaOptions, "default message name is not specified")
+	}
+
+	var prevId string
+	for _, fdbSet := range options.GetHistory().GetVersions() {
+		if len(prevId) > 0 && fdbSet.PrevId != prevId {
```
richardartoul: presumably a zero length ID is an error? Or is that allowed?

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+		sr.versions[sd.DeployId()] = sd
+		prevId = sd.DeployId()
+	}
+	sr.latestId = prevId
+
+	return sr, nil
+}
+
+func loadFileDescriptorSet(fdSet *nsproto.FileDescriptorSet, msgName string) (*schemaDescr, error) {
+	// assuming file descriptors are topological sorted
```
richardartoul: What do you mean by topologically sorted here? Like in terms of dependencies between files? What part of the code ensures the sort order is correct?

src/dbnode/storage/namespace/schema.go
```diff
@@ -0,0 +1,226 @@
...
+	var dependencies []*desc.FileDescriptor
+	var curfd *desc.FileDescriptor
+	for i, fdb := range fdSet.Descriptors {
+		fdp, err := decodeProtoFileDescriptor(fdb)
+		if err != nil {
+			return nil, xerrors.Wrapf(err, "failed to decode file descriptor(%d) in version(%d)", i, fdSet.DeployId)
+		}
+		fd, err := desc.CreateFileDescriptor(fdp, dependencies...)
```
richardartoul: Ah I see....where will we make sure things are in the right order? If things aren't sorted properly will this return an error?

src/dbnode/storage/namespace/schema.go
richardartoul: I think realistically we will want to have them just give us a proto definition and we'll parse it using the dynamic library. Asking people to use protoc is kind of annoying, but we can discuss that when we implement the API

src/dbnode/storage/namespace/schema.go
Haijuncao: yes, this will, I will add a negative test

src/dbnode/storage/namespace/schema.go
Haijuncao: below, when we create file descriptors, if dependencies not in order, below create will fail.

src/dbnode/storage/namespace/schema.go
Haijuncao: the first schema do not have prevId.

src/dbnode/storage/namespace/schema.go
Haijuncao: I prefer to reserve NewXXX for simpler constructor. This has logic to "load" from file descripitors, hence "Load".

src/dbnode/storage/namespace/schema.go
Haijuncao: ack.

src/dbnode/storage/namespace/options.go
Haijuncao: it won't be nil ever, will simplify it.

src/dbnode/storage/namespace/map_test.go
Haijuncao: not sure, but will follow the imports conventions as you mentioned below

src/dbnode/generated/proto/schematest2/otherpkg.proto
Haijuncao: I explicitly want to avoid using jhump's parse code. User should generate proto file descriptors using protoc in their repo already, our schema cli tool just pick it up from their generated go code.

src/dbnode/storage/namespace/dynamic_test.go
```diff
@@ -115,6 +116,10 @@ func TestInitializerNoTimeout(t *testing.T) {
 	require.Equal(t, ropts.BufferFutureNanos, toNanosInt64(observedRopts.BufferFuture()))
 	require.Equal(t, ropts.BufferPastNanos, toNanosInt64(observedRopts.BufferPast()))
 
+	latest, err := md.Options().SchemaRegistry().GetLatest()
+	assert.NoError(t, err)
```
richardartoul: probably better  to use `require` instead of `assert` to force a fast failure

src/dbnode/generated/proto/namespace/schema.proto
Haijuncao: will do

src/dbnode/storage/namespace/schema.go
Haijuncao: yes, it is for that purpose.

PR #1511, opened 2019-03-28
[integration] Add carbon find integration tests

arnikola:


scripts/docker-integration-tests/carbon/test.sh
```diff
@@ -29,6 +31,19 @@ function read_carbon {
   return $?
 }
 
+function find_carbon {
```
prateek: nit: instead of a single `carbon` integration tests, mind refactoring this into two tests - one for `find-carbon` and another for `read-carbon`.

scripts/docker-integration-tests/carbon/expected/a.b.c.d.json
```diff
@@ -0,0 +1,9 @@
+[
+  {
+    "id": "a.b*.caw.daz",
```
prateek: by the looks of it there are no `*` in the ids when they're ingested, why is there still a `*` in the id returned here?

scripts/docker-integration-tests/carbon/test.sh
arnikola: Wouldn't that make integration runs another minute or so longer? (namespaces seem to take 40~ sec to come up)

scripts/docker-integration-tests/carbon/expected/a.b.c.d.json
arnikola: It's a quirk of graphite where it includes your query string in the `id`; just tried querying statsdex for `foo.bar.que*.*` [actual metrics scrubbed] and got the following:

`[ {"id":"foo.bar.que*.baz","text":"baz","leaf":0,"expandable":1,"allowChildren":1}, ... ]`

scripts/docker-integration-tests/carbon/test.sh
prateek: Fair. I don't like coupled tests but the extra 40s does seem wasteful.

scripts/docker-integration-tests/carbon/expected/a.b.c.d.json
prateek: Yech, aight.

scripts/docker-integration-tests/carbon/test.sh
```diff
@@ -58,3 +73,18 @@ echo "foo.min.catch-all.baz 10 $t" | nc 0.0.0.0 7204
...
 ATTEMPTS=10 TIMEOUT=1 retry_with_backoff read_carbon foo.min.catch-all.baz 15
+
+t=$(date +%s)
+echo "a 0 $t"             | nc 0.0.0.0 7204
+echo "a.bar 0 $t"         | nc 0.0.0.0 7204
+echo "a.biz 0 $t"         | nc 0.0.0.0 7204
+echo "a.biz.cake 0 $t"    | nc 0.0.0.0 7204
+echo "a.bar.caw.daz 0 $t" | nc 0.0.0.0 7204
+echo "a.bag 0 $t"         | nc 0.0.0.0 7204
+ATTEMPTS=5 TIMEOUT=1 retry_with_backoff find_carbon a* a.json
```
prateek: suggestion(take/leave): i think breaking the files out here actually makes it harder to follow along what you're trying to do. would read easier if everything was inline. e.g. --

```
ATTEMPTS=5 TIMEOUT=1 retry_with_backoff find_carbon a* <(cat <<EOF
[
  {
    "id": "a",
    "text": "a",
    "leaf": 0,
    "expandable": 1,
    "allowChildren": 1
  }
]
EOF)
```

PR #1516, opened 2019-03-29
Add client/Aggregate tests

prateek:


src/dbnode/client/aggregate_results_accumulator_consistency_test.go
```diff
@@ -0,0 +1,455 @@
...
+					hostname:        fmt.Sprintf("testhost%d", i),
+					aggregateResult: &testAggregateSuccessResponse,
+					expectedDone:    true,
+				},
+			},
+		}.run()
+	}
+
+	// should terminate only after all failures, and say it failed
+	testFetchTaggedWorkflow{
```
robskillington: Nice.  Should we perhaps rename this to `fetchStateWorkflow` to more closely mimic the data structures it's testing (now that it's used for fetch tagged and aggregate)?

src/dbnode/client/fetch_tagged_results_accumulator_merge_test.go
```diff
@@ -314,11 +315,24 @@ func (tm testFetchTaggedWorkflow) run() fetchTaggedResultAccumulator {
...
+			done bool
+			err  error
+		)
+		if s.fetchTaggedResult != nil {
+			opts := fetchTaggedResultAccumulatorOpts{
+				host:     host(tm.t, tm.topoMap, s.hostname),
+				response: s.fetchTaggedResult,
+			}
+			done, err = accum.AddFetchTaggedResponse(opts, s.err)
+		} else {
```
robskillington: Instead of if/else can we do a switch and then have a `default` for where they are both nil and do a `require.FailNow(t, ...)`?

src/dbnode/client/aggregate_results_accumulator_consistency_test.go
prateek: sure thing

src/dbnode/client/fetch_tagged_results_accumulator_merge_test.go
prateek: sure thing

prateek: done

src/dbnode/client/aggregate_results_accumulator_consistency_test.go
prateek: done

PR #1517, opened 2019-03-29
Add m3ninx/FieldQuery

prateek:


src/m3ninx/search/searcher/field.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"github.com/m3db/m3/src/m3ninx/index"
+	"github.com/m3db/m3/src/m3ninx/postings"
+	"github.com/m3db/m3/src/m3ninx/search"
+)
+
+type fieldSearcher struct {
+	field    []byte
+	compiled index.CompiledRegex
```
arnikola: Can we drop this and match any matching field instead?

src/m3ninx/search/searcher/field.go
```diff
@@ -0,0 +1,50 @@
...
+
+	return &fieldSearcher{
+		field:    field,
+		compiled: compiledAll,
+	}, nil
+}
+
+func (s *fieldSearcher) Search(r index.Reader) (postings.List, error) {
+	// TODO: expose a new method on the reader to support such operations
+	return r.MatchRegexp(s.field, s.compiled)
```
arnikola: Ah I see the todo now :p

prateek: lol, yep - fixed in https://github.com/m3db/m3/pull/1518. this is just to avoid landing a query that doesn't work.

src/m3ninx/search/searcher/field.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+type fieldSearcher struct {
+	field    []byte
+	compiled index.CompiledRegex
+}
+
+// NewFieldSearcher returns a new searcher for finding documents which match the given field.
+func NewFieldSearcher(field []byte) (search.Searcher, error) {
+	compiledAll, err := index.CompileRegex([]byte(".*"))
```
arnikola: nit: can we reuse these with something like

```golang
var(
 matcher index.CompileRegex
 matcherErr error

 once sync.Once
)

func getMatcher() (search.Searcher, error) {
 once.Do(func(){
  matcher, err = index.CompileRegex([]byte(".*"))
 })
 return matcher, matcherErr
}
```

prateek: talked offline: going to leave as is to avoid more global state

PR #1518, opened 2019-03-30
[m3ninx/fst] Optimised Field Query execution

prateek:
- Add backwards compatible, FST minor version 1.1
- New postings lists per unique Field 
- Using new type in coordinator query conversion


src/m3ninx/generated/proto/fswriter/fswriter.proto
```diff
@@ -41,3 +41,10 @@ message Metadata {
   PostingsFormat postingsFormat = 1;
   int64          numDocs        = 2;
 }
+
+// additional metadata stored per Field in the corpus.
+message FieldData {
+  // postingsOffset for the pl corresponding to the union of all documents
+  // which have a given field.
+  uint64 fieldPostingsListOffset = 1;
+}
```
robskillington: nit: Add extra endline to eof

src/m3ninx/index/regexp.go
```diff
@@ -28,6 +28,19 @@ import (
 	fstregexp "github.com/m3db/m3/src/m3ninx/index/segment/fst/regexp"
 )
 
+var (
+	// DotStartCompiledRegex is a CompileRegex that matches any input.
+	DotStarCompiledRegex CompiledRegex
```
robskillington: Does this need to be exported?  And if so, would you be opposed to making this a method that returns the initialized var?

i.e.
```go
var (
  dotStarCompiledRegex CompiledRegex
)

func DotStarCompiledRegex() CompiledRegex {
  return dotStarCompiledRegex
}
```

src/m3ninx/index/segment/fst/README.md
```diff
@@ -1,6 +1,58 @@
 FS Segment
 ===========
 
+- Version 1.1: Adds support for a metadata proto object per Field. This is used to
```
robskillington: I kind of just realized, is there really no checksums for any of our index files? If not, I think that's something we're going to have to work on rather soon.

Let me know if there is, AFAIK I cannot recall there being any.

src/m3ninx/index/segment/fst/writer.go
```diff
@@ -198,27 +236,32 @@ func (w *writer) WritePostingsOffsets(iow io.Writer) error {
 			return err
 		}
 
+		fieldsPL := roaring.NewPostingsList()
```
robskillington: nit: You could reuse a postings list with cached containers here.  Which for the compaction code path, since it happens basically every time you add any documents, it might very well be necessary.

See how `currPostingsList` is constructed and used in the `termsIterFromSegments` here (notice built from a bitmap that is using pooling, then Reset is called at top of the inner loop).  This made a large difference in allocations:
https://github.com/m3db/m3/blob/master/src/m3ninx/index/segment/builder/multi_segments_terms_iter.go

src/m3ninx/index/segment/fst/writer.go
```diff
@@ -326,16 +393,27 @@ func (w *writer) WriteFSTTerms(iow io.Writer) error {
...
 	}
 
 	// all good!
 	w.fstTermsFileWritten = true
 	return nil
 }
 
+func fieldsMetadata(fieldPostingsOffset uint64) ([]byte, error) {
+	d := fswriter.FieldData{FieldPostingsListOffset: fieldPostingsOffset}
+	return d.Marshal()
```
robskillington: Hm this is going to allocate each time yeah?

Can you create a `proto.Buffer` and a single `*fstwriter.FieldData` on the `*writer` then do:

```go
w.fieldDataProto.Reset()
w.fieldDataProto.FieldPostingsListOffset = fieldPostingsOffset
buffer.Reset()
err := buffer.Marshal(w.fieldDataProto)
```

Docs here:
https://godoc.org/github.com/golang/protobuf/proto#Buffer.Marshal

src/query/storage/index.go
```diff
@@ -220,7 +225,15 @@ func matcherToQuery(matcher models.Matcher) (idx.Query, error) {
 		negate = true
 		fallthrough
 	case models.MatchRegexp:
-		query, err := idx.NewRegexpQuery(matcher.Name, matcher.Value)
+		var (
+			query idx.Query
+			err   error
+		)
+		if bytes.Equal(dotStar, matcher.Value) {
+			query = idx.NewFieldQuery(matcher.Name)
```
robskillington: Do we also support negation just by placing this code here?

robskillington:
@richardartoul will the cached segment do the right thing here?  i.e. does prateek need to update that code?  Will this postings list be cached?

richardartoul:
@robskillington I think we'll need to update `read_through_segment.go` because right now it "overloads" `MatchRegexp` and `MatchTerm` but I don't think it knows anything about `MatchField` so that will probably just get defer'd to the underlying segment/reader and bypass the cache entirely

src/query/storage/index.go
arnikola: we should get it for free based on line 240 
```golang
		if negate {
			query = idx.NewNegationQuery(query)
		}
```

src/m3ninx/generated/proto/fswriter/fswriter.proto
prateek: will do

src/m3ninx/index/segment/fst/README.md
prateek: Yep, no checksumming internally within the m3ninx side, we rely on the checksumming by virtue of the digest files we create in the `dbnode/fs/persist.indexWriter`.

src/m3ninx/index/segment/fst/writer.go
prateek: Ah nice! I didn't do this cause upstream Pilosa's reset doesn't do container reuse smartly, didn't realise we had fork of pilosa. Will update.

src/m3ninx/index/segment/fst/writer.go
prateek: Sure thing.

src/m3ninx/index/segment/fst/segment.go
```diff
@@ -358,6 +391,10 @@ func (r *fsSegment) MatchTerm(field []byte, term []byte) (postings.List, error)
 func (r *fsSegment) MatchRegexp(field []byte, compiled index.CompiledRegex) (postings.List, error) {
 	r.RLock()
 	defer r.RUnlock()
```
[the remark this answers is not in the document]
prateek: Sure

src/query/storage/index.go
robskillington: Great, looks good then.

src/m3ninx/index/segment/fst/README.md
robskillington: Oh yeah, ok yeah we do actually checksum the segments when we read them from disk first time.  

It's just if they decay over time, we don't validate them everytime we read data (like we do when we look up data chunks for time series data with the retriever/seeker). So basically we only validate once at boot up time.

src/m3ninx/index/segment/fst/segment.go
```diff
@@ -569,6 +683,15 @@ type fsSegmentReader struct {
 
 var _ index.Reader = &fsSegmentReader{}
 
+func (sr *fsSegmentReader) MatchField(field []byte) (postings.List, error) {
+	sr.RLock()
+	defer sr.RUnlock()
```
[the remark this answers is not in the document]
prateek: Sure, it's like 50ns-ish per defer so usually not an issue. Will avoid it.

src/m3ninx/index/segment/fst/utils.go
```diff
@@ -56,8 +62,7 @@ func newFSTSegment(t *testing.T, s sgmt.MutableSegment, opts Options) sgmt.Segme
 	require.NoError(t, w.WriteFSTFields(&fstFieldsBuffer))
 
 	data := SegmentData{
-		MajorVersion:  w.MajorVersion(),
-		MinorVersion:  w.MinorVersion(),
+		Version:       readeVersion,
```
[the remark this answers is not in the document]
prateek: sure thing

src/m3ninx/index/segment/fst/writer.go
```diff
@@ -67,20 +72,34 @@ type WriterOptions struct {
 	// amount (e.g. 2x). You can disable this to speed up high fixed cost
 	// lookups to during building of the FST however.
 	DisableRegistry bool
+
+	// ForTestsOnly_Version allows users to override behaviour of the
+	// writer. NB: this is intended to be used only during tests.
+	ForTestsOnly_Version *Version
```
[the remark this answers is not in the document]
prateek: Motivation for this is to be able to change versions for x-package testing (e.g. say we want an integration test with multiple versions of files written on disk). We don't use it today, so I can rejig if you feel strongly.

src/m3ninx/index/segment/fst/types.go
```diff
@@ -77,3 +93,25 @@ type Writer interface {
...
+func (v Version) IsSupported() bool {
+	for _, o := range SupportedVersions {
+		if v.Major == o.Major && v.Minor == o.Minor {
+			return true
+		}
+	}
+	return false
+}
+
+func (v Version) supportsFieldPostingsList() bool {
```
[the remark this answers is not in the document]
prateek: I'd be happy to make this change if we had an extensive version/feature set, but for a single version change/feature, the current approach is easier to grok.

src/m3ninx/index/segment/fst/types.go
```diff
@@ -77,3 +93,25 @@ type Writer interface {
 	// NB(prateek): this must be called after WriteFSTTerm().
 	WriteFSTFields(w io.Writer) error
 }
+
+// Validate returns an error if the provided version is invalid, or else nil.
+func (v Version) Validate() error {
+	if v.IsSupported() {
```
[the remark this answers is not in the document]
prateek: don't think there will be an validation beyond checking the version being supported. Removed `Validate()`  and made `IsSupported` return an error

src/m3ninx/index/segment/fst/writer.go
```diff
@@ -181,7 +201,25 @@ func (w *writer) WriteDocumentsIndex(iow io.Writer) error {
...
+		postingsBytes, err := w.postingsEncoder.Encode(pl)
+		if err != nil {
+			return err
+		}
+		n, err := w.writePayloadAndSizeAndMagicNumber(iow, postingsBytes)
+		if err != nil {
+			return err
+		}
+		// update offset with the number of bytes we've written
+		currentOffset += n
```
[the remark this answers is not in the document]
prateek: sure

src/m3ninx/generated/proto/fswriter/fswriter.proto
prateek: done

src/m3ninx/index/regexp.go
prateek: done

src/m3ninx/index/segment/fst/writer.go
prateek: done

src/m3ninx/index/segment/fst/writer.go
prateek: done

src/m3ninx/index/segment/fst/segment.go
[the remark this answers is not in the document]
prateek: done

src/m3ninx/index/segment/fst/segment.go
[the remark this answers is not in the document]
prateek: done

src/m3ninx/index/segment/fst/types.go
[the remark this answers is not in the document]
prateek: done

src/m3ninx/index/segment/fst/utils.go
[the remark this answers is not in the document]
prateek: done

src/m3ninx/index/segment/fst/writer.go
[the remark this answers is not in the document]
prateek: done

src/m3ninx/index/segment/fst/README.md
prateek: Yeah pretty much. We can add this with a v1.2 without much issue.

prateek:
@robskillington @richardartoul will wire up with the pl cache in a followup.

src/m3ninx/index/segment/fst/types.go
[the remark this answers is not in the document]
arnikola: Agreed, guess it's better to err on the side of simplicity where possible :+1:

src/m3ninx/index/segment/fst/writer.go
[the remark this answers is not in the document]
arnikola: The motivation makes sense, just not a big fan of having test harness stuff in real code

If there's no simple solution here (probably don't want to inject `Version` or anything weird like that) leaving it here is fine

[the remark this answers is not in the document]
robskillington: Could we make this unexported at least?

src/m3ninx/index/segment/fst/writer.go
```diff
@@ -67,20 +77,37 @@ type WriterOptions struct {
...
+		postingsEncoder:     pilosa.NewEncoder(),
+		fstWriter:           newFSTWriter(opts),
+		docDataWriter:       docs.NewDataWriter(nil),
+		docIndexWriter:      docs.NewIndexWriter(nil),
+		docOffsets:          make([]docOffset, 0, defaultInitialDocOffsetsSize),
+		fstTermsOffsets:     make([]uint64, 0, defaultInitialFSTTermsOffsetsSize),
+		termPostingsOffsets: make([]uint64, 0, defaultInitialPostingsOffsetsSize),
+
+		fieldPostingsOffsets: make([]uint64, 0, defaultInitialPostingsOffsetsSize),
+		fieldsPostingsList:   roaring.NewPostingsList(),
```
robskillington: I think you need to actually using a roaring postings list created from a bitmap that has container pooling turned on (as per the multi segment iterator stuff).

Otherwise it just throws out it's containers every time you call `Reset()`.

Sorry, do you mind updating to do that?

prateek: sure thing. Thanks for the catch. Fixed by creating a pilosa bitmap backed by a pool and wrapping into a postings list (need to cleanup the leakiness of postings list abstraction in a followup)

src/m3ninx/index/segment/fst/writer.go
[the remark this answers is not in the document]
prateek: Easier to just remove entirely, re-jigged to remove the field from the struct and use an unexported ctor in the tests.

PR #1524, opened 2019-04-02
Fix bug where data would sometimes not get loaded into cache after successful read

richardartoul:
This bug had no impact on correctness, but could impact performance by preventing blocks for series that weren't already in memory from being added to the cache after being read from disk.

Bug was introduced by #1421

Basically even when there are no encoded tags, we wrapped the bytes into a checked.Bytes. The retriever would see that the encoded tags weren't nil and try to put them into a tag decoder. The tag decoder would attempt to read nil bytes and return an error which would prevent a new shard entry from being created for items that were recently read and needed to be put in the cache.

This P.R fixes the issue by performing a length test in the retriever so even if it receives checked.Bytes with zero length it will ignore it AND it changes the seek logic to only pass along a non-nil checked.Bytes if the encoded tags have length larger than zero.

It also adds a regression to test to the retriever to prevent this from happening in the future.

### DIFF
--- a/src/dbnode/persist/fs/retriever_test.go
+++ b/src/dbnode/persist/fs/retriever_test.go
@@ -415,7 +415,7 @@ func TestBlockRetrieverOnlyCreatesTagItersIfTagsExists(t *testing.T) {
 			startTime time.Time,
 			segment ts.Segment,
 		) {
-			require.Equal(t, tagsIter, ident.EmptyTagIterator)
+			require.Equal(t, ident.EmptyTagIterator, tagsIter)
 			for tagsIter.Next() {
 			}
 			require.NoError(t, tagsIter.Err())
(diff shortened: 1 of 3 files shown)


src/dbnode/persist/fs/retriever_test.go
```diff
@@ -299,6 +299,10 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 			require.True(t, ok, fmt.Sprintf("expected %s to be retrieved, but it was not", id))
 
 			expectedTags := ident.NewTags(testTagsFromTestID(id)...)
+			if !tags.Equal(expectedTags) {
+				fmt.Println("expected: ", len(expectedTags.Values()))
+				fmt.Println("actual: ", len(tags.Values()))
```
robskillington: Can you make this a message instead of using fmt.Prints?

i.e.
```go
require.True(t, tags.Equal(expectedTags), 
	fmt.Sprintf("expectedNumTags=%d, actualNumTags=%d", len(expectedTags.Values()), len(tags.Values()))
```

PR #1527, opened 2019-04-02
Add README for protobuf encoding package

richardartoul:


src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,241 @@
...
+
+This compression scheme works because the decoder can maintain its own LRU (of the same maximum capacity) and apply the same operations in the same order when its decompressing the stream. As a result, when it encounters an encoded LRU index it can lookup the corresponding string in its own LRU at the specified index.
+
+##### Encoding
+
+The LRU dictionary compression scheme uses 3 control bits to encode all the relevant information required to decode the stream.
+
+The first control bit is the "no change" control bit. If this is set to `1` then the value has not changed since its previous value and no further encoding/decoding is require.
+
+The second control bit determines how we will interpret the subsequent bits. If its `0` then we interpret the next `n` bits (where `n` is the number of bits required to encode the highest index of the configured LRU capacity) as the index into the LRU of recently encoded / decoded strings which this value corresponds to.
```
arnikola: nits: 
prefer `it is` to `its` here
prefer `remaining bits` over `subsequent bits`, to match the next paragraph
rather than describing `n` here, maybe add it to the first paragraph? i.e.
`The LRU dictionary compression scheme uses 2 control bits to encode all the relevant information required to decode the stream. Given a configured LRU cache which can be encoded in 'n' bytes:`

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+TODO(rartoul): This will change slightly once we implement support for mid-stream scheme changes, we'll probably add one more optional control bit in the case where the initial control bit is zero which will have minimal impact on compressions since it will only be present when we do a schema change, a time unit change or the stream ends.
+
+Every write is prefixed with a control bit.
+
+If the control bit is set to `1` then we know that the stream contains another write that we need to decode and we can begin decoding the timestamp.
+
+If the control bit is set to `0` then we have either: **a)** reached the end of the stream *or* **b)** encountered a time unit change.
+
+We resolve the ambiguity by reading the next control bit which will be `0` if we've reached the end of the stream or `1` if we should interpret the next 8 bits as the new time unit that we should use for delta of delta timestamp encoding.
```
mway: > This ambiguity is resolved by reading the next control bit, which will be `0` (if this is the end of the stream) or `1` (if the next 8 bits contain a new time unit to be used for delta-of-delta timestamp encoding).

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+We have to keep track of time unit changes on our own instead of deferring to the M3TSZ delta-of-delta encoder which can handle this independently because the M3TSZ encoder relies on a custom marker scheme to indicate time unit changes.
```
mway: > Time unit changes must be tracked manually (instead of deferring to the M3TSZ delta-of-delta encoder, which can handle this independently) because the M3TSZ encoder relies [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+It avoids using a control bit for each write by using a marker which contains a prefix that could not possibly be generated into the stream by any possible input and then the decoder frequently "looks ahead" for this marker to see if it needs to decode a time unit change.
```
mway: > [...] by any possible input, and the decoder frequenty [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+The Protobuf encoding scheme has no equivalent "impossible bit combination" so it uses explicit control bits to indicate a time unit change instead.
```
mway: > [...] "impossible bit combination", so it uses [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Compressed Timestamp
+
+The Protobuf compression scheme reuses the delta of delta timestamp encoding logic that is implemented in the M3TSZ package and decribed in the [Facebook Gorilla paper](https://www.vldb.org/pvldb/vol8/p1816-teller.pdf).
+
+
+After encoding a control bit with a value of `1` indicating that there is another write in the stream, we encode the delta of delta of the current timestamp and the previous one as described in the Gorilla paper.
```
mway: > [...] value of `1` (indicating that there is another write in the stream), the delta-of-delta of the current and previous timestamp is encoded.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+Similarly, when we're decoding the stream, we perform the inverse operation and reconstruct the current timestamp based on the previous one and the delta of delta encoded into the stream.
```
mway: > Similarly, when decoding the stream, the inverse operation is performed to reconstruct the current timestamp based on both the previous timestamp and delta-of-delta encoded into the stream.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+### Compressed Protobuf Fields
+
+Compressing the protobuf fields is broken into two stages:
+
+1. Custom compressed fields
+2. Protobuf marshaled fields
+
+In the first phase we compress any fields for which we're able to apply custom compression as described in the `Compression Techniques` section.
```
mway: > In the first phase, any eligible custom fields are compressed as described in the "Compression Techniques" section.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+In the second phase we lean on the Protobuf marshaling format to encode the data for us with the caveat that we compare fields at the top-level and avoid re-encoding them if they haven't changed.
```
mway: > In the second phase, the Protobuf marshaling format is used to decode the data, with the caveat that fields are compared at the top level and re-encoding is avoided if they have not changed.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### Custom Compressed Protobuf Fields
+
+We encode the custom compressed field values similarly how to we encode their types as described in the `Header` section.
```
mway: > The custom fields' compressed values are encoded similarly to how their types are encoded as described in the "Header" section.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+In fact, they're even encoded in the same order with the caveat that unlike when we're encoding the types, we don't need to encode a null value for non-contiguous field numbers for which we're not performing any compression.
```
mway: > In fact, they're encoded in the same order, with the caveat that unlike when types are being encoded, a null value does not need to be encoded for non-contiguous field numbers that are not being compressed.

also, jfmi, does this mean:

- non-contiguous (has gaps);
- unordered; or
- non-sequential

?

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+All of the compressed value are encoded one after the other with no separator or control bit inbetween which means that they all need to encode enough information that a decoder can determine where each one ends and the next one begins.
```
mway: nit: "values"

> [...] are encoded sequentially and no separator or control bits are placed between them, which means that they must encode enough information such that a decoder [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+The values themselves are encoded based on the field type and the compression technique that is being applied to it. For example, if we return to our sample Protobuf message from earlier:
```
mway: > For example, considering the sample Protobuf message from earlier:

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Overview
+
+This package contains the encoder/decoder(iterator) for compressing streams of Protobuf messages matching a provided schema.
+All compression is performed in a streaming manner such that we update the encoded stream with each write; there is no internal buffering or batching during which we gather multiple writes before performing encoding.
+
+## Features
+
+1. Compress the timestamps of the ProtoBuf messages using [Gorilla-style delta of delta encoding](https://www.vldb.org/pvldb/vol8/p1816-teller.pdf).
+2. Compress streams of ProtoBuf messages matching a provided schema by using different forms of compression for each field based on its type.
```
robskillington: nit: Make `ProtoBuf` -> `Protobuf` as per recommended shortening: 
https://opensource.google.com/projects/protobuf

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
+# Protobuf Encoding
+
+## Overview
+
+This package contains the encoder/decoder(iterator) for compressing streams of Protobuf messages matching a provided schema.
```
robskillington: Better to just get rid of `(iterator)` here, it is clear.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+This package strives to support compressing any valid Protobuf messages as specified using [Protobuf syntax version 3](https://developers.google.com/protocol-buffers/docs/proto3), however, only the following features have been tested:
+
+1. [All scalar value types](https://developers.google.com/protocol-buffers/docs/proto3#scalar)
+2. Nested Messages
+3. Repeated fields
+4. Map fields
+5. Reserved fields
+
+We have explicitly not performed any testing of the following features and thus they are not supported:
```
robskillington: Hm, I thought you were just going to allow these and not compress?

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+3. Options of any type
+4. Custom field types
+
+## Compression Techniques
+
+This package compresses the timestamps for the Protobuf messages using [Gorilla style delta of delta encoding](https://www.vldb.org/pvldb/vol8/p1816-teller.pdf).
+
+Additionally, we perform a different type of compression for each field in the Protobuf message based on its type so that we can use compression techniques that are highly efficient for that specific type.
+
+1. Float fields are compressed using [Gorilla style XOR compression](https://www.vldb.org/pvldb/vol8/p1816-teller.pdf).
```
robskillington: Probably should mention that this is also delta-of-delta.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+2. Integer fields are compressed using M3TSZ significant digit integer compression. We don't currently have any documentation on this compression format.
```
robskillington: Is this delta-of-delta?  I can't remember, hah.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+```protobuf
+message Foo {
+	reserved 2, 3;
+	string query = 1;
+	int32 page_number = 4;
+}
+```
+
+If we had never seen the `query` `string` being encoded before, we would encode a single control bit indicating that the value had changed since its previous value, another control bit indicating that it was not found in the LRU, followed by a `varint` for the length of the string, and then the `string` bytes themselves.
```
mway: > If the `string` field `query` had never been encoded before, the following control bits would be encoded:

- list 'em

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+Next, we would use 6 bits to encode the number of significant digits in the delta between current `page_number` and the previous `page_number`, followed by a control bit indicating if the delta is positive or negative, and then finally the significant bits themselves.
```
mway: > Next, 6 bits would be used to encode the number [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+Note that the values we encoded for both fields are "self contained" in that they encode all the information required to determine when we've reached the end.
```
mway: > Note that the values encoded [...], in that they encode [...] when the end has been reached.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+#### Protobuf Marshaled Fields
+
+We recommend reading the [Protocol Buffers Encoding](https://developers.google.com/protocol-buffers/docs/encoding) section of the official documentation before reading this section.
+Specifically, understanding how protobuf messages are (basically) encoded as a stream of tuples in the form of `<field number, wire type, value>` will make understanding this section much easier.
+
+The Protobuf marshaled fields section of the encoding scheme contains all the values that we don't currently support performing custom compression on.
```
mway: > [...] all of the values that don't currently support [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+For the most part, the output of this section is similar to the result of calling `Marshal()` on a message in which all the custom compressed fields have already been removed and all that remains is the fields for which we wish to rely upon the Protobuf logic for encoding.
```
mway: > [...] removed, and the only remaining fields are ones for which Protobuf will encode directly.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+This is possible because, as described in the protobuf encoding section linked above, the protobuf wire format does not encode **any** data for fields which are not set or are set to a default value, so by "clearing" the fields that we've already encoded on our own, we can prevent them from taking up any space when we marshal the remainder of the Protobuf message.
```
mway: > [...] by "clearing" the fields that have already been encoded, they can be omitted when marshaling the remainder of the Protobuf message.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+While we do lean heavily on the Protobuf wire format in this section, we do make the most basic optimization of avoiding re-encoding fields that haven't changed since the previous value where "haven't changed" is defined at the top most level of the message.
```
mway: > While Protobuf's wire format is leaned upon heavily, there is specific attention given to avoiding re-encoding [...], where "haven't changed" [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+For example, lets imagine we were trying to encode messages with the following schema:
```
mway: > For example, consider encoding messages with the following schema:

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+    }
+    int64 outer = 1;
+    NestedDeeper deeper = 2;
+  }
+
+  Nested nested = 1;
+}
+```
+
+If none of the values inside `nested` have changed since the previous message, we don't need to encode the `nested` field at all.
```
mway: > the `nested` field doesn't need to be encoded at all.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+However, if any of the fields have changed, like `nested.deeper.booly` for example, then we need to re-encode the entire `nested` field, including the `outer` field even though only the `deeper` field changed.
```
mway: > [...] then the entire `nested` field must be re-encoded (including the `outer` field, even though only the `deeper` field changed).

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+We can perform this top-level "only if it has changed" encoding because when we're decoding the stream later, we can reconstruct the original message by merging the previously decoded message with the current "delta" message that only contains the fields that have changed since the previous message.
```
mway: > This top-level "only if it has changed" encoding can be used because, when the stream is decoded later, the original message can be reconstructed by merging the previously-decoded message with the current delta message, which contains only fields [...]

ps: "only if changed" encoding is literally a variant of delta encoding, do you want to call it that explicitly instead?

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+Only marshaling the fields that have changed since the previous message works for the most part, but there is one important edge case.
```
mway: > [...] important edge case: because the Protobuf wire format [...], using the standard Protobuf marshaling format with delta encoding works in every scenario [...] (because it is not possible to express explicitly setting a field to its default value).

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+As we said earlier, the protobuf wire format does not encode **any** data for fields that are set to a default value (zero for `integers` and `floats`, empty array for `bytes` and `strings`, etc).
+This means that using the standard Protobuf marshaling format with our "only encode the field if it has changed" scheme works in every scenario *except* for the case where a field is changed from a non-default value to a default value because there is no way to express that in the Protobuf wire format.
+
+To get around this issue, if any of the Protobuf marshaled fields are changed to their default value, then we encode an additional bitset which specifies which field numbers were set to default values.
```
mway: > This issue can be mitigated by encoding an additional bitset which indicates any field numbers that were set to the default value of the field's type.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+The bitset encoding is straightforward.
```
mway: > [...] straightforward: it begins [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+It begins with a `varint` that encodes the length (number of bits) of the bitset, and then the remaining `n` bits are interpreted as a 1-indexed bitset (because field numbers start at 1 not 0) where the value of `1` means the field was changed to its default and the value of `0` means it was not.
```
mway: > [...] where a value of `1` indicates that the field was changed to its default value.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+##### Protobuf Marshaled Fields Encoding Format
+
+The Protobuf marshaled fields section of the encoding begins with a single control bit that indicates whether there have been any changes to the Protobuf encoded portion of the message at all.
```
mway: nit: "Protobuf Marshaled Fields"

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+If the control bit is set to `1`  then there have been changes and we need to continue decoding, and if it is set to `0` then there were no changes and we can begin decoding the next write (or we've reached the end of the stream).
```
mway: > If the control bit is set to `1`, then there have been changes and decoding must continue; if it is set to `0`, then there were no changes and the decoder can skip to the next write (or stop, if at the end of the stream).

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+If the previous control bit was set to `1`, indicating that there have been changes, then there will be another control bit.
```
mway: > [...] another control bit which indicates whether any fields [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+The next control bit indicates whether any fields have been set to a default value.
+If so, then its value will be `1` and the subsequent bits should be interpreted as a `varint` encoding the length of the bitset followed by the actual bitset bits as discussed Aabove.
```
mway: nit: "above"

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+If the value is `0` then there is no bitset to decode.
```
mway: > [...] value is `0`, then there [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+Finally, this portion of the encoding will end with a final `varint` that encodes the length of the bytes that resulted from calling `Marshal()` on the message (in which we've cleared any fields that were custom encoded or weren't custom encoded but also haven't changed since the previous message) followed by the actual marshaled bytes themselves.
```
mway: > ([...] where any custom-encoded or unchanged fields were cleared), followed by [...]

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+1. [All scalar value types](https://developers.google.com/protocol-buffers/docs/proto3#scalar)
+2. Nested Messages
+3. Repeated fields
+4. Map fields
+5. Reserved fields
+
+We have explicitly not performed any testing of the following features and thus they are not supported:
+
+1. `Any` fields
+2. `Oneof` fields
```
benraskin92: `oneof` (https://developers.google.com/protocol-buffers/docs/proto#oneof)

src/dbnode/encoding/proto/README.md
benraskin92: Can we add a `todo` to add documentation?

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+1. Float fields are compressed using [Gorilla style XOR compression](https://www.vldb.org/pvldb/vol8/p1816-teller.pdf).
+2. Integer fields are compressed using M3TSZ significant digit integer compression. We don't currently have any documentation on this compression format.
+3. `bytes` and `string` fields are compressed using a custom dictionary based compression scheme which we will refer to as "LRU Dictionary Compression". This compression scheme is described in further detail below.
+
+### LRU Dictionary Compression
+
+The LRU dictionary compression scheme provides high levels of compression for `bytes` and `string` fields that match any of the following criteria:
+
+1. The value of the field changes infrequently.
+2. The value of the field changes frequently, but tends to rotate among a small number of frequently used values which may evolve over time.
```
benraskin92: Maybe have a quick example of what compresses nicely vs poorly?

src/dbnode/encoding/proto/README.md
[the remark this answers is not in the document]
richardartoul: It contains both. `iterator.go` is the decompression pathway

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+4. Absolutely no lossiness for any of the compression types.
+
+## Supported Syntax
+
+This package strives to support compressing any valid Protobuf messages as specified using [Protobuf syntax version 3](https://developers.google.com/protocol-buffers/docs/proto3), however, only the following features have been tested:
+
+1. [All scalar value types](https://developers.google.com/protocol-buffers/docs/proto3#scalar)
+2. Nested Messages
+3. Repeated fields
+4. Map fields
```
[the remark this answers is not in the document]
richardartoul: its also non-deterministic in the Proto spec so I think this is ok as is

src/dbnode/encoding/proto/README.md
richardartoul: I don't do anything to stop people from using these things and they probably work, but I haven't included them in my prop test just because I haven't had time so kind of a "use at your own risk" right now and we may just reject them outright in the schema management APIs

Nothing stopping us from supporting them later, just haven't had time to test basically

src/dbnode/encoding/proto/README.md
richardartoul: nope just one delta

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,241 @@
...
+
+##### Encoding
+
+The LRU dictionary compression scheme uses 3 control bits to encode all the relevant information required to decode the stream.
+
+The first control bit is the "no change" control bit. If this is set to `1` then the value has not changed since its previous value and no further encoding/decoding is require.
+
+The second control bit determines how we will interpret the subsequent bits. If its `0` then we interpret the next `n` bits (where `n` is the number of bits required to encode the highest index of the configured LRU capacity) as the index into the LRU of recently encoded / decoded strings which this value corresponds to.
+
+If the second control bit is set to `1` then the remainder should be interpreted as a `varint` encoding the length of the `bytes` followed by the `bytes` themselves.
```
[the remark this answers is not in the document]
richardartoul: You can just do that because of how varints works. You read them one complete byte at a time and if the first bit is zero then you've read the last byte, but if its 1 then you need to read another byte (similar to how UTF8 is encoded)

Probably gonna leave this out cause I don't want to go into a huge aside on how varints work

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+This isn't a strict requirement of the encoding scheme (in fact it reduces compression slightly), but it greatly simplifies the implementation because the encoder needs to be able to reference bytes that its already encoded so it can check if the bytes that are about to be encoded are in the LRU.
+The encoder could keep track of all the bytes that correspond to each LRU entry in memory, but that would be wasteful in terms of memory utilization.
+Instead, its more efficient if the LRU stores offsets into the encoded stream for the beginning and end of the bytes that have already been encoded.
+These offsets are much easier to keep track of and compare against if we can assume they always correspond to the beginning of a byte boundary and we don't have to keep track of the bit offset into an individual byte as well.
+In the future we may change the implement to favor complicating the code so that we don't have to waste any bits on the padding.
+
+### Compression Limitations
+
+This package will attempt to compress all scalar type fields at the top level of a message, but will not compress any data that is part of a `repeated`, `map` or `nested message` field.
```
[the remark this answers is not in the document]
richardartoul: Compressing a repeated field well would just be really complicated because you'd have to account for changes in the length of the repeated sequence from one message to the next. We could do it, but I wouldn't want to support it unless we had a really good use case that was difficult to model any other way.

Maps are even more difficult because you'd have to deal with the sorting issue, as well as the fact that keys may be added and removed arbitrarily. It would get messy really quickly.

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+The `nested message` restriction may be lifted in the future, but the `repeated` and `map` restrictions are unlikely to change.
```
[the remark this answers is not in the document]
richardartoul: I don't want to get into a ton of detail, I just said "due to the difficulty of compressing variably sized fields"

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+
+## Binary Format
+
+At a high level you can think about the way we perform compression of Protobuf messages as being akin to doing the following:
+
+1. Scanning each schema to identify which fields we can "extract out" and perform some form of streaming compression for as described in the `Compression Techniques` section.
```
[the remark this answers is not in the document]
richardartoul: its actually the schema that we operate on, not an individual message

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,269 @@
...
+2. Maintaining an independent "stream" of compressed data for each field that we're applying custom compression to, as well as for the timestamp that is shared by all fields. When a new write comes in, we would iterate over all the fields that we're compressing and encode the next value into the appropriate stream.
```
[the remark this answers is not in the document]
richardartoul: It has to be provided to the encoder. I.E for each write you basically provide: <timestamp, protoMessage>

src/dbnode/encoding/proto/README.md
richardartoul: has gaps if I understand your question correctly

src/dbnode/encoding/proto/README.md
```diff
@@ -0,0 +1,255 @@
...
+1. encoding scheme version (`varint`)
+2. dictionary compression LRU cache size (`varint`)
+3. a list of "custom" (compressed as described in this document as opposed to standard Protobuf encoding) encoded fields with their type and field numbers
+
+#3 can be thought of as a list of `<fieldNum, fieldType>` and is encoded as follows:
+
+1. highest field number (`N`) that will be described (`varint`)
+2. `N` sets of 3 bits where each set corresponds to the "custom type", which is enough information to determine how the field should be compressed / decompressed. This is analogous to a Protobuf [`wire type`](https://developers.google.com/protocol-buffers/docs/encoding) in that it includes enough information to skip over the field if its not present in the schema that is being used to decode the message.
+
+Notably, the list only *explicitly* encoders the custom field type. *Implicitly*, the Protobuf field number is encoded by the position of the entry in the list.
```
robskillington: nit: `s/encoders/encodes`

PR #1528, opened 2019-04-03
[dbnode] Start RPC servers before topology / database initialization

### LINKED ISSUE #996: Health endpoints should start before topology client initialization

richardartoul:
Fixes #996


src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -217,25 +234,27 @@ func (s *service) Health(ctx thrift.Context) (*rpc.NodeHealthResult_, error) {
...
+			s.Lock()
+			s.health = newHealth
+			s.Unlock()
+
+			// Update response
+			health = newHealth
+		}
 	}
 
 	return health, nil
```
schallert: Do we have an easy way to return the health response along with a 5xx status code? Or is that something we'd have to modify our tchannel -> http wrappers to achieve? Can open up an issue if this would be a non-trivial change

robskillington:
@schallert it's possible but we'd have to return an error instead of a response:
```go
	if value, ok := errValue.(error); ok && xerrors.IsInvalidParams(value) {
		w.WriteHeader(http.StatusBadRequest)
	} else {
		w.WriteHeader(http.StatusInternalServerError)
	}
```

We could extend it look at some value in the context or something potentially to return a different status code.

Also what's the proper response here?  i.e. what exact code should we use?  500 itself doesn't seem correct.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -1249,11 +1302,36 @@ func (s *service) SetWriteNewSeriesLimitPerShardPerSecond(
...
+		// but it provides some basic sanity test against this function being
+		// called more than one time.
+		return errDatabaseHasAlreadyBeenSet
+	}
+
+	if db != nil {
+		s.db = db
+		s.nowFn = db.Options().ClockOptions().NowFn()
+		s.pools.id = db.Options().IdentifierPool()
+		atomic.StoreUint64(&s.dbIsSet, 1)
```
robskillington: This is racey, you need to do a compare and swap or else two goroutines could call this method at the same time and succeed.

Instead can you use the mutex already on the DB as per previous comment?  RLock should be nanoseconds.  

If you still would prefer an atomic, for the "checking" part, you can use the mutex to decide when to set the DB and then at the end of the SetDatabase method set a different atomic.  That way you avoid RLock on the read case, but still use the mutex for guarantees around setting the DB just once without having to use compare and set (i.e. prefer idiomatic vs specialized when idiomatic is just as good as specialized).

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -246,6 +265,10 @@ func (s *service) Health(ctx thrift.Context) (*rpc.NodeHealthResult_, error) {
 // endpoint because while the Health endpoint provides the same information, this endpoint does not
 // require parsing the response to determine if the node is bootstrapped or not.
 func (s *service) Bootstrapped(ctx thrift.Context) (*rpc.NodeBootstrappedResult_, error) {
+	if s.db == nil {
```
robskillington: This is racey to read as well without an RLock, because you can't guarantee it's not currently being changed from nil to not nil as you read this (since no RLock held).

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -217,25 +234,27 @@ func (s *service) Health(ctx thrift.Context) (*rpc.NodeHealthResult_, error) {
...
-		*newHealth = *health
-		newHealth.Bootstrapped = bootstrapped
-
-		s.Lock()
-		s.health = newHealth
-		s.Unlock()
-
-		// Update response
-		health = newHealth
+	if s.db != nil {
```
robskillington: This is racey to read without an RLock, because you can't guarantee it's not currently being changed from nil to not nil as you read this (since no RLock held).

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -124,6 +132,8 @@ type service struct {
 	sync.RWMutex
 
 	db      storage.Database
+	dbIsSet uint64 // Atomic, 1 for true 0 for false.
```
robskillington: Can you just use the existing mutex instead of using an atomic?  Makes the code much more idiomatic.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -1249,11 +1302,36 @@ func (s *service) SetWriteNewSeriesLimitPerShardPerSecond(
...
+		s.db = db
+		s.nowFn = db.Options().ClockOptions().NowFn()
+		s.pools.id = db.Options().IdentifierPool()
+		atomic.StoreUint64(&s.dbIsSet, 1)
+	}
+
+	return nil
+}
+
+func (s *service) canDoRPC() error {
```
robskillington: nit: Can we name this something like, `startRPC() error`?

schallert:
> Also what's the proper response here? i.e. what exact code should we use? 500 itself doesn't seem correct.

@robskillington for Kubernetes purposes all we care about is anything >= 400 to indicate "not bootstrapped". I think this is how a lot of other systems interpret health checks too.

What would actually be the most ideal is if before an instance is in the placement (has its DB initialized?) it returns 200 (since there's nothing to be bootstrapped for), and then while it's bootstrapping it's >= 400. Even if we made that via config that's fine by me.

robskillington:
After we discussed, @schallert believe 409 Conflict is the best fit after inspecting a lot of the 4xx and 5xx detailed descriptions since it's rather abstract.

src/dbnode/network/server/httpjson/handlers.go
```diff
@@ -206,6 +228,7 @@ func RegisterHandlers(mux *http.ServeMux, service interface{}, opts ServerOption
 				return
 			}
 
+			w.WriteHeader(http.StatusOK)
```
schallert: This may be redundant. From docs: https://golang.org/pkg/net/http/#ResponseWriter
```
        // If WriteHeader has not yet been called, Write calls
        // WriteHeader(http.StatusOK) before writing the data.
```
Not sure if that could cause the issues we're debugging

schallert: Update: I removed this line and saw the same EOF error, not the issue

src/dbnode/server/server.go
```diff
@@ -597,34 +636,7 @@ func Run(runOpts RunOptions) {
...
-	tchannelthriftNodeClose, err := ttnode.NewServer(service,
-		cfg.ListenAddress, contextPool, tchannelOpts).ListenAndServe()
-	if err != nil {
-		logger.Fatal("could not open tchannelthrift interface",
-			zap.String("address", cfg.ListenAddress), zap.Error(err))
-	}
-	defer tchannelthriftNodeClose()
-	logger.Info("node tchannelthrift: listening", zap.String("address", cfg.ListenAddress))
-
+	// Start the cluster services now that the M3DB client is available
```
richardartoul: .

src/dbnode/server/server.go
```diff
@@ -652,15 +655,24 @@ func Run(runOpts RunOptions) {
...
 	logger.Info("cluster httpjson: listening", zap.String("address", cfg.HTTPClusterListenAddress))
 
-	if cfg.DebugListenAddress != "" {
-		go func() {
-			if err := http.ListenAndServe(cfg.DebugListenAddress, nil); err != nil {
-				logger.Error("debug server could not listen",
-					zap.String("address", cfg.DebugListenAddress), zap.Error(err))
-			}
-		}()
+	// Initialize clustered database
```
richardartoul: .

src/dbnode/topology/static.go
```diff
@@ -46,6 +46,11 @@ func (i staticInitializer) Init() (Topology, error) {
 	return NewStaticTopology(i.opts), nil
 }
 
+func (i staticInitializer) TopologySet() (bool, error) {
+	// Always has the specified static topology ready
```
richardartoul: .

src/dbnode/topology/types.go
```diff
@@ -54,6 +54,11 @@ type HostShardSet interface {
 type Initializer interface {
 	// Init will return a new topology
 	Init() (Topology, error)
+
+	// TopologySet returns whether the topology is able to be
+	// initialized immediately or if instead it will blockingly
+	// wait to be set on initialization
+	TopologySet() (bool, error)
```
richardartoul: super nit: how about `TopologyIsSet`, `TopologySet` had me thinking `a set of topologies` and I was bouncing around a little bit trying to figure out how it was used

src/dbnode/topology/dynamic.go
```diff
@@ -73,6 +74,25 @@ func (i *dynamicInitializer) Init() (Topology, error) {
 	return i.topo, nil
 }
 
+func (i *dynamicInitializer) TopologySet() (bool, error) {
```
richardartoul: Is this going to make an RPC to KV each time?

robskillington: Correct, we can live with this for now (also existing k8s operator was calling the operator endpoint to simulate this behavior already).

src/dbnode/network/server/httpjson/handlers.go
```diff
@@ -41,6 +41,30 @@ var (
...
+}
+
+// NewError creates a new HTTP JSON error which has a specified status code.
+func NewError(err error, statusCode int) Error {
+	e := errorType{error: err}
+	e.statusCode = statusCode
+	return e
+}
+
+// StatusCode implements the Error interface.
```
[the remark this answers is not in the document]
robskillington: Sure thing.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -187,77 +224,114 @@ func NewService(db storage.Database, opts tchannelthrift.Options) rpc.TChanNode
...
 			blockMetadataV2:         opts.BlockMetadataV2Pool(),
 			blockMetadataV2Slice:    opts.BlockMetadataV2SlicePool(),
 		},
-		health: &rpc.NodeHealthResult_{
-			Ok:           true,
-			Status:       "up",
-			Bootstrapped: false,
-		},
 	}
+	s.SetDatabase(db)
```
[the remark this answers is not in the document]
robskillington: Sure thing.

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -121,15 +131,34 @@ func newServiceMetrics(scope tally.Scope, samplingRate float64) serviceMetrics {
...
-	health  *rpc.NodeHealthResult_
+}
+
+type serviceState struct {
+	sync.RWMutex
+	db     storage.Database
+	health *rpc.NodeHealthResult_
+}
+
+func (s *serviceState) DB() storage.Database {
```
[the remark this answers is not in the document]
robskillington: Sure thing.

src/dbnode/network/server/tchannelthrift/options.go
```diff
@@ -36,6 +42,13 @@ type options struct {
 
 // NewOptions creates new options
 func NewOptions() Options {
+	bytesPool := pool.NewCheckedBytesPool(nil, nil, func(s []pool.Bucket) pool.BytesPool {
```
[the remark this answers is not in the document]
robskillington: Yes.

src/dbnode/topology/types.go
robskillington: Sure, don't really mind - I don't usually believe "Is" as a prefix to a method is good, but I can see how as it is currently is confusing.

PR #1529, opened 2019-04-03
[m3snch] kube script: add warmup between starts

schallert:
We see some weird behavior when all of the clients try to start at once,
let's add a warmup.

### DIFF
--- a/kube/scripts/m3nsch_client.sh
+++ b/kube/scripts/m3nsch_client.sh
@@ -14,6 +14,8 @@
 
 set -eo pipefail
 
+WARMUP=${WARMUP:-10}
+
 # Get endpoints of the agents
 function get_endpoints() {
   local selector="app=m3nsch,component=agent"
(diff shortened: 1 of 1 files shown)


kube/scripts/m3nsch_client.sh
```diff
@@ -34,6 +36,19 @@ fi
 
 AGENT_ENDPOINTS=$(get_endpoints)
 
+if [[ "$#" -eq 1 && "$1" == "start" ]]; then
+  (
+    IFS=","
```
richardartoul: Is this being used? I'm confused

schallert: Yup! This is some bash-foo that when we iterate over the comma-separated `$AGENT_ENDPOINTS` will treat `,` as the separator (it's usually ` `, `\t`, or `\n`). https://bash.cyberciti.biz/guide/$IFS

PR #1532, opened 2019-04-04
Add documentation for replication and deployment in zones

robskillington:


docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
+# Replication and Deployment in Zones
+
+## Overview
+
+M3DB supports deployment to both a single zone, with rack level isolation, and deployment across multiple zones in a region. You can also deploy M3DB across multiple regions for a global view of data, however you may experience higher ingestion latency and cross region bandwidth costs. You may want to use a replication factor of five for such a multi-region cluster to survive multiple regions suffering an outage, or be able to still upgrade hosts while a region is down.
```
mway: > M3DB supports both deploying to a single zone with rack-level isolation and deploying across multiple zones in a region. It can also be deployed across multiple regions for a global view of data, though both ingestion latency and bandwidth costs may increase as a result. In multi-region clusters, a replication factor of five is recommended for upgrade and multi-region failure scenarios.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Replication
+
+It is highly recommended that you use a replication factor of at least 3 for any M3DB deployment. This is due to the default write and read consistency levels requiring a majority of hosts to be available to complete an operation successfully. For more information see the documentation concerning tuning [availability, consistency and durability](availability_consistency_durability.md).
```
mway: > A replication factor of at least 3 is highly recommended for any M3DB deployment, due to the consistency levels (for both reads and writes) that require quorum in order to complete an operation. For more information, see the documentation [...]

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Upgrading hosts in a deployment
+
+It should be noted that when a host is deployed and it is preparing to serve reads by bootstrapping, the shards are available for writes but not available for reads.
```
mway: > [...] when a host is deployed and is preparing [...]

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+There is a small window during an upgrade between when the process is stopped and then started again where the node is unavailable for writes too.
```
mway: > Additionally, there is a brief window during upgrades in which nodes are unavailable for writes while processes are restarted.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Deployment in a single zone
+
+For deployment in a single zone, we recommend you set the `isolationGroup` to the name of the rack a host resides or another logical unit that separates groups of hosts in your zone.
```
mway: > [...] it is recommended to set the `isolationGroup` property to the name [...]

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+In this configuration hosts are distributed shards of data that will not be placed more than once in the same rack or logical unit you define, which allows you to lose an entire unit at a time that may contain multiple hosts and still stay available since it is guaranteed to only affect one replica of data.
```
mway: > In this configuration, shards are distributed among hosts such that each will not be placed more than once in the same defined rack or logical unit. This allows an entire unit to be lost at any given time, as it is guaranteed to only affect one replica of data.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+Here's what a deployment in a single zone looks like, in this example the cluster has 3 shards that are spread over 4 racks:
```
mway: > For example, in a single-zone deployment with three shards spread over four racks:

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+![Replication Single Zone](replication_single_zone.png)
+
+Typically deployments have many more than 3 shards, but this is a simple example that depicts how losing a single rack still maintains availability of two of the three replicas of each shard.
```
mway: > Typically, deployments have many more than three shards - this is a simple example that illustrates how M3DB maintains availability while losing a single rack, as two of three replicas are still intact.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Deployment across multiple availability zones in a region
+
+For deployment in a region, we recommend you set the `isolationGroup` to the name of the availability zone a host resides.
```
mway: > [...] it is recommended to set the isolationGroup property to the name [...]

docs/operational_guide/availability_consistency_durability.md
```diff
@@ -10,7 +10,7 @@ Generally speaking, [the default and example configuration for M3DB](https://git
 Database operators who are using M3DB for workloads that require stricter consistency and durability guarantees should consider tuning the default configuration to better suit their use case.
 
 The rest of this document describes the various configuration options that are available to M3DB operators to make such tradeoffs.
-While reading it, we recommend reffering to [the default configuration file](https://github.com/m3db/m3/blob/master/src/dbnode/config/m3dbnode-all-config.yml) (which has every possible configuration value set) to see how the described values fit into M3DB's configuration as a whole.
+While reading it, we recommend referring to [the default configuration file](https://github.com/m3db/m3/blob/master/src/dbnode/config/m3dbnode-all-config.yml) (which has every possible configuration value set) to see how the described values fit into M3DB's configuration as a whole.
```
richardartoul: lol nice

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+In this configuration hosts are distributed shards of data that will not be placed more than once in the same availability zone.
```
mway: > In this configuration, shards are distributed among hosts such that each will not be placed more than once in a given availability zone.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+Here's what a deployment in a region looks like, in this example the cluster has 4 shards that are spread over 3 availability zones:
```
mway: > For example, in a multi-zone, single-region deployment with four shards spread over three availability zones:

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+![Replication Region](replication_region.png)
+
+Typically deployments have many more than 4 shards, but this is a simple example that depicts how losing a single availability zone still maintains availability of two of the three replicas of each shard.
```
mway: > Typically, deployments have many more than four shards - this is a simple example that illustrates how M3DB maintains availability while losing a single availability zone, as two of three replicas are still intact.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+<!-- **Note:** Placement of shards in a replica do not mirror other replicas and shards are placed in a way to best load balance shards across available hosts in a single isolation group. This could mean that hosts in an availability zones that have less hosts will do more work than those in an availability zone with more hosts. -->
```
mway: should we nix this?

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: higher ingestion AND query latency right?

Also this statement is confusing "You may want to use a replication factor of five for such a multi-region cluster to survive multiple regions suffering an outage, or be able to still upgrade hosts while a region is down."

In theory that advice would apply to racks and zones as well. Why is running in 3 regions not enough? I don't think most people design their systems to tolerate 2 regions out at a time (otherwise we would run with R.F=5 in our datacenters no?)

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: I think technically we recommend unstrict_majority for reads

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Deployment across multiple regions
+
+For deployment across regions, we recommend you set the `isolationGroup` to the name of the region a host resides.
```
mway: > [...] it is recommended to set the isolationGroup property to the name [...]

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: Also you're referring people to the "availability, consistency, and durability" section but that section doesn't discuss replication factor at all (I think we had that as a TODO that came up after I merged it)

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+As mentioned before you may experience higher ingestion latency and cross region bandwidth costs for a cluster that spans regions. You may also want to use a replication factor of five for such a multi-region cluster to survive multiple regions suffering an outage, or be able to still upgrade hosts while a region is down.
```
mway: > As mentioned previously, ingestion latency and bandwidth costs may increase when using clusters that span regions. Additionally, it is recommended to use a replication factor of five for such clusters in order to better support upgrade and multi-region failure scenarios.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+In this configuration hosts are distributed shards of data that will not be placed more than once in the same region.
```
mway: > In this configuration, shards are distributed among hosts such that each will not be placed more than once in a given region.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+Here's what a deployment across regions looks like, in this example the cluster has 4 shards that are spread over 5 regions:
```
mway: > For example, in a multi-region deployment with four shards spread over five regions:

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+![Replication Global](replication_global.png)
+
+Typically deployments have many more than 4 shards, but this is a simple example that depicts how losing up to two regions still maintains availability of three of the five replicas of each shard.
```
mway: > Typically, deployments have many more than four shards - this is a simple example that illustrates how M3DB maintains availability while losing two regions, as three of five replicas are still intact.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Overview
+
+M3DB supports deployment to both a single zone, with rack level isolation, and deployment across multiple zones in a region. You can also deploy M3DB across multiple regions for a global view of data, however you may experience higher ingestion latency and cross region bandwidth costs. You may want to use a replication factor of five for such a multi-region cluster to survive multiple regions suffering an outage, or be able to still upgrade hosts while a region is down.
+
+### Replication
+
+It is highly recommended that you use a replication factor of at least 3 for any M3DB deployment. This is due to the default write and read consistency levels requiring a majority of hosts to be available to complete an operation successfully. For more information see the documentation concerning tuning [availability, consistency and durability](availability_consistency_durability.md).
+
+### Upgrading hosts in a deployment
```
richardartoul: I think this could use a little more detail to be clearer.

"When an M3DB node is restarted it has to go through a "bootstrapping" process to restore itself to a valid state. During this time the node will continue to accept writes, but will not be available for reads.

Obviously, there is also a small window of time during between when the process is stopped and then started again where it will also be unavailable for writes."

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: maybe: "to the name of the rack that a host resides on or whatever other unit of isolation makes sense for your zone deployment"

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -0,0 +1,57 @@
...
+
+In this configuration hosts are distributed shards of data that will not be placed more than once in the same region.
+
+Here's what a deployment across regions looks like, in this example the cluster has 4 shards that are spread over 5 regions:
+
+![Replication Global](replication_global.png)
+
+Typically deployments have many more than 4 shards, but this is a simple example that depicts how losing up to two regions still maintains availability of three of the five replicas of each shard.
+
+**Note:** Placement of shards in a replica do not mirror other replicas and shards are placed in a way to best load balance shards across available hosts in a single isolation group. This could mean that hosts in a region that have less hosts will do more work than those in a region with more hosts.
```
mway: > Shard placements within a replica are not mirrored among other replicas; shards are placed in such a way as to optimally balance among available hosts in a given isolation group. Consequently, this means that hosts in smaller regions may do more work on average than hosts in larger regions.

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: "In this configuration shards will be distributed amongst hosts such that no rack (or any other isolation group that you defined) will contain more than one replication factor of any given shards. This ensures that your deployment can tolerate the loss of an entire rack / isolation group (regardless of how many hosts it contain) because at most you will only loss a single replication factor.

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: maybe: "still maintains availability of the entire cluster for writes and reads because we still have 2 full replication factors for all of the shards which is enough to achieve quorum"

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: "resides on" or "resides in"

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: Same comment as earlier, I think its slightly easier to understand if its flipped "In this configuration shards are distributed such that each availability zone cannot contain more than a single replication factor for any given shard."

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: Same comment as before, take it or leave it

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: The mirroring thing might make things more confusing. Mirror deployments are uncommon in most setups that I've seen so I think leaving out any reference to it might be less confusing.

Also I think we can rephrase to: "M3DB will do its best to distribute shards evenly among the availability zones while still taking each individual node's weight into account, but if some of the availability zones have less available hosts than others then each host in that zone will be responsible for more shards than hosts in the other zones and will thus be subjected to heavier load." or something like that

Also should we move this note out of this section and make it more general? It applies if you have an uneven number of hosts in your rack distribution or an uneven number of hosts in your multi region setup too

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: on / in

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: same comment

docs/operational_guide/replication_and_deployment_in_zones.md
richardartoul: same comment, can we only say this once in a more general section since it applies to racks too?

docs/operational_guide/availability_consistency_durability.md
mway: 🤦‍♂️ my bad

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: TY, tweaked slightly but looks good.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: TY, looks good.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: TY, looks good.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: TY, looks good.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: I added "host attribute" to the end of your suggestion, now it is `it is recommended to set the `isolationGroup` host attribute to the name`.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: TY looks good.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: TY sounds good.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: Hm, I could get rid of the "mirror" statement but I do think it's useful to note that if you misbalance the hosts that they will become unbalanced.

robskillington: I'll just remove it.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: Yeah I'll just remove this stuff.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: That's fine, I'm referring them to the documentation to look at the consistency levels.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: Sure thing, thanks for the suggestion - looks good. I tweaked this just slightly.  It's not really restoring to a "valid state", it's just preparing to serve reads.

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: Just changed this to "is in" instead of "resides".

docs/operational_guide/replication_and_deployment_in_zones.md
robskillington: Yeah I removed the mirroring part and have moved it under the "replication" header.  I'll use the suggested text you have, slightly better than my revised one, TY!

PR #1534, opened 2019-04-04
update schema

Haijuncao:


src/dbnode/storage/namespace.go
```diff
@@ -393,6 +397,18 @@ func (n *dbNamespace) Shards() []Shard {
...
+func (n *dbNamespace) SchemaRegistry() namespace.SchemaRegistry {
+	n.RLock()
+	defer n.RUnlock()
+	return n.schemaRegistry
+}
+
+func (n *dbNamespace) SetSchemaRegistry(v namespace.SchemaRegistry) {
+	n.Lock()
+	defer n.Unlock()
+	n.schemaRegistry = v
```
robskillington: Hm should we not do some validation here?

src/dbnode/storage/namespace.go
```diff
@@ -393,6 +397,18 @@ func (n *dbNamespace) Shards() []Shard {
 	return databaseShards
 }
 
+func (n *dbNamespace) SchemaRegistry() namespace.SchemaRegistry {
+	n.RLock()
+	defer n.RUnlock()
```
robskillington: nit: Avoid defer here since it's easy to do and may be called in the fast path for a write?

src/dbnode/storage/database.go
```diff
@@ -340,6 +358,32 @@ func (d *db) addNamespacesWithLock(namespaces []namespace.Metadata) error {
 	return nil
 }
 
+func (d *db) updateNamespaceSchemasWithLock(schemaUpdates []namespace.Metadata) error {
+	for _, n := range schemaUpdates {
+		// ensure namespace exists
+		existing, ok := d.namespaces.Get(n.ID())
+		if !ok { // should never happen
```
richardartoul: Again, I know the existing file breaks this convention but we usually try and put a comment like this within the branch so:

```golang
if !ok {
    // Should never happen.
    ....
}
```

src/dbnode/storage/database.go
```diff
@@ -340,6 +358,32 @@ func (d *db) addNamespacesWithLock(namespaces []namespace.Metadata) error {
...
+			return fmt.Errorf("non-existing namespace marked for schema update: %v", n.ID().String())
+		}
+		currentId := "none"
```
richardartoul: Can we stick with either `current` or `existing` for variables in this function, and also can we call this variable `currentSchemaID` or `existingSchemaID` since ID could be misconstrued as the namespace ID

src/dbnode/storage/database.go
```diff
@@ -340,6 +358,32 @@ func (d *db) addNamespacesWithLock(namespaces []namespace.Metadata) error {
...
+		current, found := existing.SchemaRegistry().GetLatest()
```
richardartoul: `currentSchema` or `existingSchema` would help readability a little

src/dbnode/storage/database.go
```diff
@@ -340,6 +358,32 @@ func (d *db) addNamespacesWithLock(namespaces []namespace.Metadata) error {
...
+		if found {
+			currentId = current.DeployId()
+		}
+		// log schema update
```
richardartoul: same nit about comments

src/dbnode/storage/database.go
```diff
@@ -340,6 +358,32 @@ func (d *db) addNamespacesWithLock(namespaces []namespace.Metadata) error {
...
+		latest, found := n.Options().SchemaRegistry().GetLatest()
+		if !found {
```
richardartoul: In what situation could this occur? Should this be an `invariant violated` type of situation? Either a comment explaining how this could happen or making the error message more descriptive would be helpful

src/dbnode/storage/database.go
```diff
@@ -340,6 +358,32 @@ func (d *db) addNamespacesWithLock(namespaces []namespace.Metadata) error {
...
+			return fmt.Errorf("failed to update latest schema for namespace %s", n.ID().String())
+		}
+		d.log.Infof("updating database namespace (%s) schema from %s to %s", n.ID().String(), currentId, latest.DeployId())
```
richardartoul: Ah I see you have a log about the schema change here. Could we add the `oldSchema` and the `newSchema` as string fields on the log message to aid debugging and operations?

richardartoul: Also I think it might be preferable if this log was emitted after the error check below so users never see it if the update is gonna fail when you call `SetSchemaRegistry`

src/dbnode/storage/database.go
```diff
@@ -340,6 +358,32 @@ func (d *db) addNamespacesWithLock(namespaces []namespace.Metadata) error {
...
+		err := existing.SetSchemaRegistry(n.Options().SchemaRegistry())
+		if err != nil {
+			return xerrors.Wrapf(err, "failed to update latest schema for namespace %s", n.ID().String())
```
richardartoul: I didnt know we had this, nice

src/dbnode/storage/database_test.go
```diff
@@ -614,11 +614,13 @@ func TestDatabaseUpdateNamespace(t *testing.T) {
 	nses := d.Namespaces()
 	require.Len(t, nses, 2)
 
+	sr, err := namespace.LoadSchemaRegistry(namespace.GenTestSchemaOptions("namespace/schematest"))
+
```
richardartoul: Can we add a check to make sure that the initial schema is not the same as the one that gets set?

src/dbnode/storage/namespace.go
```diff
@@ -116,6 +116,9 @@ type dbNamespace struct {
 	// entry will be nil when this shard does not belong to current database
 	shards []databaseShard
 
+	// contains the schema registry for the database namespace.
```
richardartoul: same nit, sorry the existing comments are really old

richardartoul: Also this comment could be a little more helpful: "Contains the schema registry for the database namespace. Can be updated dynamically and will propagate to existing streams lazily on subsequent writes." Or whatever information you think might be useful here

src/dbnode/storage/namespace.go
```diff
@@ -293,8 +296,8 @@ func newDatabaseNamespace(
 
 	scope := iops.MetricsScope().SubScope("database").
 		Tagged(map[string]string{
-			"namespace": id.String(),
-		})
+		"namespace": id.String(),
```
richardartoul: Maybe github is messing up, but deleting this whitespace seems wrong? Should this not be indented

src/dbnode/storage/namespace.go
richardartoul: What happens if the validation fails during loading? Will the new namespaces still propagate to the clustered database or will it error out before then?

src/dbnode/storage/namespace/options_test.go
```diff
@@ -53,7 +53,7 @@ func TestOptionsEqualsIndexOpts(t *testing.T) {
 
 func TestOptionsEqualsSchema(t *testing.T) {
 	o1 := NewOptions()
-	s1, err := LoadSchemaRegistry(GenTestSchemaOptions())
+	s1, err := LoadSchemaRegistry(GenTestSchemaOptions("schematest"))
```
richardartoul: is this string referring to a specific file that is being loaded? Should we make it a constant?

src/dbnode/storage/namespace/schema.go
```diff
@@ -41,25 +41,24 @@ var (
...
 
 func (s *schemaDescr) DeployId() string {
 	return s.deployId
 }
 
+func (s *schemaDescr) PrevDeployId() string {
+	return s.prevDeployId
+}
+
 func (s *schemaDescr) Equal(o SchemaDescr) bool {
```
richardartoul: Why remove all the nil checks?

src/dbnode/storage/namespace/schema.go
```diff
@@ -41,25 +41,24 @@ var (
...
-	if s == nil && o == nil {
-		return true
-	}
-	if s != nil && o == nil || s == nil && o != nil {
-		return false
-	}
 	if _, ok := o.(*schemaDescr); !ok {
```
richardartoul: Do we still need this check?

src/dbnode/storage/namespace/schema.go
```diff
@@ -113,6 +112,19 @@ func (sr *schemaRegistry) Equal(o SchemaRegistry) bool {
 	return true
 }
 
+func (sr *schemaRegistry) Lineage(v SchemaRegistry) bool {
```
richardartoul: Can we call this `Extends`? I would expect `Lineage` to return an actual slice or something

src/dbnode/storage/namespace/schema.go
```diff
@@ -113,6 +112,19 @@ func (sr *schemaRegistry) Equal(o SchemaRegistry) bool {
...
+	cur, hasMore := v.GetLatest()
+
+	for hasMore {
```
richardartoul: maybe I'm missing something, but why do we need a loop here? Can we just check if `v.GetLatest().DeployID()` is in sr? If not, then SR does not not extend `v` right?

Oh I see its because you need to make sure the entire history of `v` is contained within `sr`...

Ok that makes sense although I feel like this would be a lot simpler if we just also kept track of the index in the history so we could just see if it came later in the list. Maintaining the sort order shouldn't be hard (and we can still keep the linked-list thing as well as an additional sanity thing)

src/dbnode/storage/namespace/schema.go
```diff
@@ -194,9 +206,9 @@ func loadFileDescriptorSet(fdSet *nsproto.FileDescriptorSet, msgName string) (*s
 
 	md := curfd.FindMessage(msgName)
 	if md != nil {
-		return &schemaDescr{deployId: fdSet.DeployId, md: md}, nil
+		return &schemaDescr{deployId: fdSet.DeployId, prevDeployId: fdSet.PrevId, md: md}, nil
```
richardartoul: might be worth a constructor function at this point to make sure we're always creating `schemaDescr` correctly

src/dbnode/storage/database.go
```diff
@@ -340,6 +358,32 @@ func (d *db) addNamespacesWithLock(namespaces []namespace.Metadata) error {
 	return nil
 }
 
+func (d *db) updateNamespaceSchemasWithLock(schemaUpdates []namespace.Metadata) error {
+	for _, n := range schemaUpdates {
+		// ensure namespace exists
```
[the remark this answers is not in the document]
Haijuncao: sure. thanks for letting me know the convention, will follow.

src/dbnode/storage/database.go
Haijuncao: ok

src/dbnode/storage/database.go
Haijuncao: ok

src/dbnode/storage/namespace/schema.go
Haijuncao: yes

src/dbnode/storage/namespace/schema.go
Haijuncao: sure.

src/dbnode/storage/namespace/schema.go
Haijuncao: I tried to generate the index, did not seem to be simpler than this.

src/dbnode/storage/namespace.go
Haijuncao: It will not be propagated (we won't update schema registry on the namespace). Ideally, we catch validation error before it get persisted into etcd.

src/dbnode/storage/database.go
```diff
@@ -309,10 +321,16 @@ func (d *db) logNamespaceUpdate(removes []ident.ID, adds, updates []namespace.Me
 		return fmt.Errorf("unable to format updates, err = %v", err)
 	}
 
+	schemaUpdateString, err := metadatas(schemaUpdates).String()
```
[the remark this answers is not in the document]
Haijuncao: the entire schema could be huge and the compact format is verbose. instead I log old deploy id and new deploy id. it is very clear.

$ go test -run TestDatabaseUpdateNamespace
15:23:31.044593[I] creating namespaces watch
15:23:31.044941[I] resolving namespaces with namespace watch
15:23:31.045013[I] updating database namespaces [{adds [testns1, testns2]} {updates []} {schema updates []} {removals []}]
15:23:31.055063[I] received update from kv namespace watch
15:23:31.055117[I] updating database namespaces [{adds []} {updates [testns1, testns2]} {schema updates [testns2]} {removals []}]
15:23:31.055197[I] updating database namespace (testns2) schema from **none** to **third**
15:23:31.055206[W] skipping namespace removals and updates (except schema updates), restart process if you want changes to take effect.
PASS

src/dbnode/storage/namespace/schema.go
Haijuncao: ok

src/dbnode/storage/types.go
```diff
@@ -235,6 +235,9 @@ type Namespace interface {
 
 	// Shards returns the shard description
 	Shards() []Shard
+
+	// Schema returns the schema registry
```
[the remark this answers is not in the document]
Haijuncao: ok

src/dbnode/storage/database.go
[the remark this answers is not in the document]
richardartoul: ok fair

src/dbnode/storage/database.go
richardartoul: @Haijuncao Any update on this?

src/dbnode/storage/namespace.go
richardartoul: @Haijuncao bump

src/dbnode/storage/namespace.go
richardartoul: @Haijuncao bump

src/dbnode/storage/namespace.go
richardartoul: 👍

src/dbnode/storage/namespace/options_test.go
richardartoul: bump

src/dbnode/storage/namespace/schema.go
richardartoul: bump

src/dbnode/storage/namespace/schema.go
richardartoul: ok fair

src/dbnode/storage/database_test.go
richardartoul: bump

src/dbnode/storage/database.go
Haijuncao: will add more comment.

I think this may happen when the schema registry for a namespace is accidentally wiped out by an admin. an empty registry is technically a valid registry, so we want to catch it here. We can also catch it later in `curNamespace.SetSchemaRegistry` (where it checks lineage).  But the error message may be not as clear.

src/dbnode/storage/database_test.go
Haijuncao: we did in database.go line292 method `namespaceDeltaWithLock`

this file is a test file, not database.go anymore.

src/dbnode/storage/namespace.go
Haijuncao: ok

src/dbnode/storage/namespace/options_test.go
Haijuncao: This is the relative path (comparing to current test file) to the location of the test proto file. 
Since I want to reuse this in storage test as well, I make this an input, in storage test, it will use "namespace/schematest".

src/dbnode/storage/namespace/schema.go
Haijuncao: pointer receiver should not be nil, if it is, should panic (we will fix it asap, there must be a bug in LoadSchemaRegistry). 
if o is nil, it won't pass the type assertion test in line59. so I removed those checks.

src/dbnode/storage/namespace/schema.go
Haijuncao: I see, you mean that we only invoke interface method for equality check. yes, maybe we don't need the assertion. 

in that case, I will have to add back the nil check.

PR #1537, opened 2019-04-08
[ci] docs: also build operator docs

schallert:
We serve the operator's docs from a branch subdomain of the monorepo.
Let's incorporate it into the build process.

This means that once we update docs in the operator repo (which will
eventually be merged w/ the monorepo anyway) we'll have to wait until
the next master build, but that slight drift is fine.


schallert:
This PR works! https://github.com/m3db/m3/commit/036ea2f944482942d03b3de6d74de5b06b1dfa14

PR #1539, opened 2019-04-09
add write options to shard, series, buffer write api

Haijuncao:
did not add write options to database/namespace  as there is no use case for that atm.


src/dbnode/storage/series/types.go
```diff
@@ -283,3 +284,35 @@ func NewStats(scope tally.Scope) Stats {
...
+	return &opts
+}
+
+func (o *WriteOptions) WriteType() WriteType {
+	return o.writeType
+}
+
+func NewWriteOptions() *WriteOptions {
+	return &WriteOptions{}
+}
```
robskillington: Can we get rid of these setter/getter methods?

Also we should remove `NewWriteOptions()` and change the types everywhere to take just a struct by value `WriteOptions` instead of a pointer `*WriteOptions`.

If it's a pointer type then we need to allocate on the heap, but since it's not stateful we should just copy and pass it everywhere by the stack (much cheaper in terms of memory management).

i.e. remove `NewWriteOptions(...)` and the getters/setters and change all types from:
```go
func DoThing(arg string, opts *WriteOptions) { ... }
```

To:
```go
func DoThing(arg string, opts WriteOptions) { ... }
```

robskillington:
To echo justin's comment, we should pass these along the stack (by value) so that they don't need to be allocated on the heap. 

Let's change it from pointers everywhere to just struct values.

src/dbnode/storage/series/types.go
Haijuncao: Ah, I see. thanks for the explanation, it makes sense. yes, if it is pass over the stack, then immutability is not needed. I did not really get what Richard is trying to say the first time.

Haijuncao:
> Hmm I thought we wanted to just pass a `WriteOptions` struct down instead of a pointer. @robskillington's and @richardartoul's comments [here](https://github.com/m3db/m3/pull/1083#discussion_r273244345) are really concerns about having to allocate on each write request (since interfaces are pointers). Passing a struct down directly avoids this. Rob mentioned that these options aren't used by any long lived object, so immutability shouldn't be too concerning. Thoughts?
> 
> Separately, looks like you need to run `make mock-gen-dbnode`.

got it. thanks for the explanation.

PR #1541, opened 2019-04-09
[query] Offset function

arnikola:


src/query/ts/m3db/encoded_unconsolidated_series_iterator.go
```diff
@@ -76,7 +81,7 @@ func (it *encodedSeriesIterUnconsolidated) Next() bool {
 		dp, _, _ := iter.Current()
 		values = append(values,
 			xts.Datapoint{
-				Timestamp: dp.Timestamp,
+				Timestamp: dp.Timestamp.Add(it.offset),
```
robskillington: Hm yeah, it feels weird that we do this at this layer and also in the query/storage layer.

Is there any possibility that the `Offset` function itself could just adjust the timestamps back to what is expected?

src/query/storage/converter.go
```diff
@@ -279,7 +280,12 @@ func iteratorToTsSeries(
 	datapoints := make(ts.Datapoints, 0, initRawFetchAllocSize)
 	for iter.Next() {
 		dp, _, _ := iter.Current()
-		datapoints = append(datapoints, ts.Datapoint{Timestamp: dp.Timestamp, Value: dp.Value})
+		// NB: re-apply the offset to every datapoint here to move the decompressed
+		// series back to the correct timeframe.
```
robskillington: Hm, does it feel strange doing this in the `query/storage` package?

robskillington:
After having a look at the code more closely, my naive response would be maybe it would be more natural to adjust the timestamps back using a transform which is attached by code in the `FetchNode` after the call to fetch blocks from storage?

src/query/storage/converter.go
arnikola: This is more a side effect of where this converter thing is; it should probably be in an m3 specific one since it's converting series iterators to points

robskillington:
Closing since @arnikola merged #1561

PR #1543, opened 2019-04-10
[docs] clean operator directory

schallert:


.buildkite/scripts/docs_push.sh
```diff
@@ -57,6 +57,7 @@ if diff -qr m3db-operator/site m3db.io; then
 fi
 
 cp -r m3db-operator/site/* m3db.io
+rm -rf m3db-operator
```
prateek: instead of this, mind adding a `git clean -fd` to the appropriate hook scripts: i think `post-checkout` and `pre-exit` should do it (similar to .buildkite/hooks/pre-command)

PR #1545, opened 2019-04-11
[dbnode] Aggregate() using only FSTs where possible

prateek:
- When we're not filtering on a query, we can compose the results of Aggregation from the FSTs directly. This avoids the code path from postings lists -> documents, thereby saving a lot of CPU.


src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,181 @@ func (b *block) addQueryResults(
...
+	exhaustive := !opts.LimitExceeded(size)
+	return exhaustive, nil
+}
+
+func (b *block) appendAggregateResults(
+	batch []AggregateResultsEntry,
+	field, term []byte,
+	includeTerms bool,
+) []AggregateResultsEntry {
+	// NB(prateek): we make a copy of the (field, term) entries returned
```
prateek: highlight this bit as a potential contentious choice. could simplify to living with higher contention if people feel strongly.

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -104,6 +104,50 @@ func (r *aggregatedResults) AddDocuments(batch []doc.Document) (int, error) {
...
+func (r *aggregatedResults) AddFields(batch []AggregateResultsEntry) (int, error) {
+	r.Lock()
+	for _, entry := range batch {
+		f := entry.Field
+		aggValues, ok := r.resultsMap.Get(f)
+		if !ok {
+			aggValues = r.valuesPool.Get()
+			// we can avoid the copy because we assume ownership of the passed ident.ID,
+			// but still need to finalize it.
+			r.resultsMap.set(f, aggValues, _AggregateResultsMapKeyOptions{
```
prateek: using an internal method of codegen'd type feels sketch but we don't have the equivalent exported so no way around

src/dbnode/storage/index/block.go
```diff
@@ -768,6 +778,31 @@ func (b *block) executorWithRLock() (search.Executor, error) {
 	return executor.NewExecutor(readers), nil
 }
 
+func (b *block) segmentsWithRLock() ([]segment.Segment, error) {
+	expectedSegments := len(b.foregroundSegments) + len(b.backgroundSegments)
```
arnikola: nit: `segmentCount` maybe makes a little more sense?

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,190 @@ func (b *block) addQueryResults(
...
 }
 
+// Aggregate acquires a read lock on the block so that the segments
+// are guaranteed to not be freed/released while accumulating results.
+// This allows references to the mmap'd segment data to be accumulated
+// and then copied into the results before this method returns (it is not
+// safe to return docs directly from the segments from this method, the
+// results datastructure is used to copy it every time documents are added
+// to the results datastructure). This is similar to how Query() operates.
+// NB: Aggregate is required in addition to Query, to optimise for the case
```
arnikola: nit: `NB: Aggregate is an optimization of the general aggregate Query approach for the case when we can skip going to raw documents..`

src/dbnode/storage/index/block.go
```diff
@@ -768,6 +778,31 @@ func (b *block) executorWithRLock() (search.Executor, error) {
 	return executor.NewExecutor(readers), nil
 }
 
+func (b *block) segmentsWithRLock() ([]segment.Segment, error) {
```
arnikola: nit: doesn't look like this needs to return an error

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,190 @@ func (b *block) addQueryResults(
...
+	results AggregateResults,
+) (bool, error) {
+	b.RLock()
+	defer b.RUnlock()
+
+	if b.state == blockStateClosed {
+		return false, ErrUnableToQueryBlockClosed
+	}
+
+	segs, err := b.segmentsWithRLock()
```
arnikola: nit: if this can't error, may be a bit cleaner to put this at :966 just before iterating through it

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,190 @@ func (b *block) addQueryResults(
...
+	if err != nil {
+		return false, err
+	}
+
+	for _, s := range segs {
+		if opts.LimitExceeded(size) {
+			break
+		}
+
+		err = iter.Reset(s, fieldsAndTermsIteratorOpts{
```
arnikola: nit: worth pulling the opts out of the loop?

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,190 @@ func (b *block) addQueryResults(
...
+				return false, err
+			}
+		}
+
+		if err := iter.Err(); err != nil {
+			iter.Close()
+			return false, err
+		}
+
+		if err := iter.Close(); err != nil {
```
arnikola: nit: I don't think this is necessary; you reset every loop and close the iterator explicitly below already

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,190 @@ func (b *block) addQueryResults(
...
+		}
+	}
+
+	if err := iter.Close(); err != nil {
+		return false, err
+	}
+
+	// Add last batch to results if remaining.
+	if len(batch) > 0 {
+		batch, size, err = b.addAggregateResults(cancellable, results, batch)
```
arnikola: Does this need to append aggregate results too?

src/dbnode/storage/index/block.go
arnikola: I like this approach fwiw :)

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,190 @@ func (b *block) addQueryResults(
...
+	// by the iterator during traversal, because the []byte are only valid per entry during
+	// the traversal (i.e. calling Next() invalidates the []byte). We choose to do this
+	// instead of checking if the entry is required (duplicates may exist in the results map
+	// already), as it reduces contention on the map itself. Further, the ownership of these
+	// idents is transferred to the results map, which either hangs on to them (if they are new),
+	// or finalizes them if they are duplicates.
+	var (
+		entry            AggregateResultsEntry
+		lastField        []byte
+		lastFieldIsValid bool
```
arnikola: nit: I don't think this is necessary, `bytes.Equal` should handle the `lastField == nil` case

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,190 @@ func (b *block) addQueryResults(
...
+		if err != nil {
+			return false, err
+		}
+	}
+
+	exhaustive := !opts.LimitExceeded(size)
+	return exhaustive, nil
+}
+
+func (b *block) appendAggregateResults(
```
arnikola: nit: the name was really confusing (as was `addAggregateResults` followed by `addAggregateResults`), maybe rename this to `appendFieldAndTermToBatch`

src/dbnode/storage/index/aggregate_results.go
prateek: Good will, will do.

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -104,6 +104,50 @@ func (r *aggregatedResults) AddDocuments(batch []doc.Document) (int, error) {
...
+			})
+		} else {
+			// because we already have a entry for this field, we release the ident back to
+			// the underlying pool.
+			f.Finalize()
+		}
+		valuesMap := aggValues.Map()
+		for _, t := range entry.Terms {
+			_, ok := valuesMap.Get(t)
+			if !ok {
```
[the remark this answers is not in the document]
prateek: sure

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -104,6 +104,50 @@ func (r *aggregatedResults) AddDocuments(batch []doc.Document) (int, error) {
...
+				// we can avoid the copy because we assume ownership of the passed ident.ID,
+				// but still need to finalize it.
+				valuesMap.set(t, struct{}{}, _AggregateValuesMapKeyOptions{
```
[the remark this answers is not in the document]
prateek: sure

src/dbnode/storage/index/block.go
prateek: will do

src/dbnode/storage/index/block.go
prateek: sure

src/dbnode/storage/index/block.go
prateek: sure

src/dbnode/storage/index/block.go
prateek: sure thing

src/dbnode/storage/index/block.go
prateek: allocation wise it's a stack alloc either way so no bigge. i prefer how reads, a little more declarative.

src/dbnode/storage/index/block.go
prateek: Hm it's necessary cause we want to ensure we're releasing any internal state (via Close()). Reset() just clears that state, doesn't release it.

src/dbnode/storage/index/block.go
prateek: sure thing

src/dbnode/storage/index/block.go
prateek: Aye, me too.

src/dbnode/storage/index/block.go
prateek: Hm not sure we can. Consider the case when we're trying to distinguish the first element of a batch (i.e. no prior elements in the batch), from a batch with the last element having a nil field. I don't think it'll happen in practice but this way makes fewer assumptions about the data so tend to prefer it.

src/dbnode/storage/index/field_terms_iterator_test.go
```diff
@@ -0,0 +1,172 @@
...
+var (
+	testFstOptions    = fst.NewOptions()
+	lotsTestDocuments = util.MustReadDocs("../../../m3ninx/util/testdata/node_exporter.json", 2000)
+)
+
+func TestFieldsTermsIteratorSimple(t *testing.T) {
+	s := newFieldsTermsIterSetup(
+		pair{"a", "b"}, pair{"a", "c"},
+		pair{"d", "e"}, pair{"d", "f"},
+		pair{"g", "h"}, pair{"i", "j"},
```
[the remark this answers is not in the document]
prateek: sure

src/dbnode/storage/index/field_terms_iterator_test.go
```diff
@@ -0,0 +1,172 @@
...
+	seg := s.asSegment(t)
+
+	iter, err := newFieldsAndTermsIterator(seg, fieldsAndTermsIteratorOpts{
+		iterateTerms: true,
+		allowFn: func(f []byte) bool {
+			return !bytes.Equal([]byte("a"), f) && !bytes.Equal([]byte("k"), f)
+		},
+	})
+	require.NoError(t, err)
+	slice := toSlice(t, iter)
```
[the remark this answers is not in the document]
prateek: nah, they're testing different things. in the first case we're ensuring the fixture is the same as the output; in this case we're passing expected output (which is different from the fixture under test) v observed output

src/dbnode/storage/index/field_terms_iterator_test.go
```diff
@@ -0,0 +1,172 @@
...
+		pair{"d", "e"}, pair{"d", "f"},
+		pair{"g", "h"}, pair{"i", "j"},
+		pair{"k", "l"},
+	)
+	seg := s.asSegment(t)
+
+	iter, err := newFieldsAndTermsIterator(seg, fieldsAndTermsIteratorOpts{})
+	require.NoError(t, err)
+	slice := toSlice(t, iter)
+	requireSlicesEqual(t, []pair{
```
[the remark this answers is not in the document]
prateek: For sure.

src/dbnode/storage/index/field_terms_iterator_test.go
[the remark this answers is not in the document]
arnikola: Yeah, just meant we keep the style consistent with the previous test; this `ex` has fewer elements than the initial list

src/dbnode/storage/index_block_test.go
```diff
@@ -689,3 +692,105 @@ func TestNamespaceIndexBlockAggregateQuery(t *testing.T) {
...
+	b0 := index.NewMockBlock(ctrl)
+	b0.EXPECT().Stats(gomock.Any()).Return(nil).AnyTimes()
+	b0.EXPECT().Close().Return(nil)
+	b0.EXPECT().StartTime().Return(t0).AnyTimes()
+	b0.EXPECT().EndTime().Return(t0.Add(blockSize)).AnyTimes()
+	b1 := index.NewMockBlock(ctrl)
+	b1.EXPECT().Stats(gomock.Any()).Return(nil).AnyTimes()
+	b1.EXPECT().Close().Return(nil)
+	b1.EXPECT().StartTime().Return(t1).AnyTimes()
+	b1.EXPECT().EndTime().Return(t1.Add(blockSize)).AnyTimes()
```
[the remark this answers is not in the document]
prateek: heh sure

src/dbnode/storage/index_block_test.go
```diff
@@ -689,3 +692,105 @@ func TestNamespaceIndexBlockAggregateQuery(t *testing.T) {
...
+		return now
+	}
+	opts := testDatabaseOptions()
+	opts = opts.SetClockOptions(opts.ClockOptions().SetNowFn(nowFn))
+
+	b0 := index.NewMockBlock(ctrl)
+	b0.EXPECT().Stats(gomock.Any()).Return(nil).AnyTimes()
+	b0.EXPECT().Close().Return(nil)
+	b0.EXPECT().StartTime().Return(t0).AnyTimes()
+	b0.EXPECT().EndTime().Return(t0.Add(blockSize)).AnyTimes()
```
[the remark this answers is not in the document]
prateek: sure

src/dbnode/storage/index/types.go
```diff
@@ -181,6 +181,16 @@ type AggregateResults interface {
 		aggregateQueryOpts AggregateResultsOptions,
 	)
 
+	// AggregateResultsOptions returns the set AggregateResultsOptions.
```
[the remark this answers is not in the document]
prateek: sure

src/dbnode/storage/index/types.go
```diff
@@ -273,6 +290,16 @@ type Block interface {
 		results BaseResults,
 	) (exhaustive bool, err error)
 
+	// Aggregate aggregates known tag names/values.
+	// NB(prateek): this is different from Query, as we can
```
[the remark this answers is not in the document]
prateek: Nah, methods on this interface are `Query()` and `Aggregate()`. Unless you mean to say rename?

src/dbnode/storage/index/options.go
```diff
@@ -304,6 +322,16 @@ func (o *opts) DocumentArrayPool() doc.DocumentArrayPool {
 	return o.docArrayPool
 }
 
+func (o *opts) SetAggregateResultsEntryArrayPool(value AggregateResultsEntryArrayPool) Options {
+	opts := *o
+	opts.aggResultsEntryArrayPool = value
```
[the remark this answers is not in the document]
prateek: meh, i'll add for sanity.

src/dbnode/storage/index/block.go
prateek: nah, cause we're already done that for the elements in the batch in a previous loop iteration.

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+	seg  segment.Segment
+	opts fieldsAndTermsIteratorOpts
+
+	done      bool
+	err       error
+	fieldIter segment.FieldsIterator
+	termIter  segment.TermsIterator
+
+	current struct {
+		valid bool
```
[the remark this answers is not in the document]
prateek: sure thing.

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+		field []byte
+		term  []byte
+	}
+}
+
+var (
+	fieldsAndTermsIterZeroed fieldsAndTermsIter
+)
+
+var _ fieldsAndTermsIterator = &fieldsAndTermsIter{}
```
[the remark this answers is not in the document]
prateek: Lol I mean, I hope the compiler can get rid of that alloc.

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+	iter := &fieldsAndTermsIter{}
+	err := iter.Reset(s, opts)
+	if err != nil {
+		return nil, err
+	}
+	return iter, nil
+}
+
+func (fti *fieldsAndTermsIter) Reset(s segment.Segment, opts fieldsAndTermsIteratorOpts) error {
+	*fti = fieldsAndTermsIterZeroed
```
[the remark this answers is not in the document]
prateek: haha so its working cause we're setting the value at the address of the receiver. 

i.e. the receiver is a pointer, so `*receiver` is a reference to the original value. and we're overwriting that to the zero value. 

`fieldsAndTermsIterZeroed = *fti` would be doing the alternative (and thus mutating, `fieldsAndTermsIterZeroed`).

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+	fiter, err := s.FieldsIterable().Fields()
+	if err != nil {
+		return err
+	}
+	fti.fieldIter = fiter
+	return nil
+}
+
+func (fti *fieldsAndTermsIter) setNextField() bool {
+	if fti.fieldIter == nil {
```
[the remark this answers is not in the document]
prateek: sure

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+		fti.err = fti.fieldIter.Err()
+	}
+
+	fti.current.valid = true
+	fti.current.term, _ = fti.termIter.Current()
+	return true
+}
+
+func (fti *fieldsAndTermsIter) Next() bool {
+	if fti.err != nil || fti.done {
```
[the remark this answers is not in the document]
prateek: cheers

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		fti.current.valid = false
+		fti.err = err
+		return false
+	}
+	fti.termIter = termsIter
+
+	hasNext = fti.termIter.Next()
+	if !hasNext {
+		fti.current.valid = false
```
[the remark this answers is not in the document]
prateek: removed all uses of valid

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+	termsIter, err := fti.seg.TermsIterable().Terms(fti.current.field)
+	if err != nil {
+		fti.current.valid = false
+		fti.err = err
+		return false
+	}
+	fti.termIter = termsIter
+
+	hasNext = fti.termIter.Next()
+	if !hasNext {
```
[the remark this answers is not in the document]
prateek: definitely. i'll rework and add better tests

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+
+	// if iterating both terms and fields, check if current field has another term
+	if fti.termIter != nil {
+		if fti.termIter.Next() {
+			fti.current.valid = true
+			fti.current.term, _ = fti.termIter.Current()
+			return true
+		}
+		if err := fti.termIter.Err(); err != nil {
+			fti.err = err
```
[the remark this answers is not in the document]
prateek: nah, expectation is for users to call `Close()` themselves

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+		return true
+	}
+
+	fti.current.valid = false
+	fti.err = fti.fieldIter.Err()
+	return false
+}
+
+func (fti *fieldsAndTermsIter) setNext() bool {
+	// if only need to iterate fields
```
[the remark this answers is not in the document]
prateek: imo makes the code harder to read if i do it here

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+	fti.termIter = termsIter
+
+	hasNext = fti.termIter.Next()
+	if !hasNext {
+		fti.current.valid = false
+		fti.err = fti.fieldIter.Err()
+	}
+
+	fti.current.valid = true
+	fti.current.term, _ = fti.termIter.Current()
```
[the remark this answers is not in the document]
prateek: Thanks, good catch.

src/dbnode/storage/index/block.go
arnikola: I think the nil field will break this in general even if it's somewhere in the middle (since you'd be trying to call `.Bytes()` on it); otherwise is there any real difference between? 

Could do something like 
```golang
if len(batch) > 0 {
 last := batch[len(batch)-1]
 if last != nil {
  lastField = last.Bytes()
 }
}

 if bytes.Equal(lastField, field) {
  entry = batch[len(batch)-1] 
 } else {
 entry.Field = b.pooledID(field) 
} 
...
```

src/dbnode/storage/index/block.go
arnikola: Fair; in that case it's a little weird that we double close below

For this particular iterator it's not really a problem, but it's plausible for an iter implementatiopn to panic if it's double closed here

src/dbnode/storage/index/aggregate_results.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/block.go
prateek: Definitely a fair point. This implementation doesn't have that issue intentionally for this reason - so as to ensure we free resources. I'll make a note about this property in a comment.

prateek: done

src/dbnode/storage/index/block.go
prateek: we're guaranteed it'd only be the first element (cause its the first lexicographic string). It wouldn't break anything cause we'd still allocate an ident.ID backed by an empty slice for it.

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,201 @@
...
+		if err := fti.termIter.Close(); err != nil {
+			fti.err = err
+			return false
+		}
+	}
+
+	// i.e. need to switch to next field
+	hasNext := fti.setNextField()
+	if !hasNext {
+		return false
```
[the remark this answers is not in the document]
prateek: removed `done` as suggested below

src/dbnode/storage/index/fields_terms_iterator.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/fields_terms_iterator.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/options.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/types.go
```diff
@@ -273,6 +290,16 @@ type Block interface {
...
+	// avoid going to documents, relying purely on the indexed
+	// FSTs.
+	Aggregate(
+		cancellable *resource.CancellableLifetime,
+		opts QueryOptions,
+		results AggregateResults,
+	) (exhaustive bool, err error)
+
 	// AddResults adds bootstrap results to the block, if c.
```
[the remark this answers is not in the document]
prateek: sure.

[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/types.go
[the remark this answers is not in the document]
arnikola: `different from aggregating by means of Query` maybe? Otherwise it kinda reads like this aggregates tag names/values which you don't really expect Query to do

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,200 @@
...
+		return true
+	}
+
+	fti.err = fieldIter.Err()
+	return false
+}
+
+func (fti *fieldsAndTermsIter) setNext() bool {
+	// if only need to iterate fields
+	if !fti.opts.iterateTerms {
```
arnikola: nit: I think this might flow a little better if the branch is moved to the `Next()` function instead of calling `setNext` then instantly calling `setNextField`

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	// and get next term for the field
+	termsIter, err := fti.seg.TermsIterable().Terms(fti.current.field)
+	if err != nil {
+		fti.err = err
+		return false
+	}
+	fti.termIter = termsIter
+
```
arnikola: Hmm this may need to check `if !fti.opts.allow(field)` and move onto the next fieldIter if not

src/dbnode/storage/index/aggregate_results.go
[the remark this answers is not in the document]
arnikola: meganit: Can probably delete the `ok` var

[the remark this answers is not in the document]
arnikola: meganit: can probably remove the `ok` var

src/dbnode/storage/index/field_terms_iterator_test.go
```diff
@@ -99,6 +102,141 @@ func TestFieldsTermsIteratorTermsOnly(t *testing.T) {
...
+
+	seg := newMockSegment(ctrl, map[string][]string{
+		"a": nil,
+	})
+	iter, err := newFieldsAndTermsIterator(seg, fieldsAndTermsIteratorOpts{iterateTerms: true})
+	require.NoError(t, err)
+	slice := toSlice(t, iter)
+	requireSlicesEqual(t, []pair{}, slice)
+}
+
```
arnikola: Can you add a few tests with `allowFn` set, maybe one which goes through a couple of segments

src/dbnode/storage/index/fields_terms_iterator.go
prateek: fair, done.

src/dbnode/storage/index/fields_terms_iterator.go
prateek: nope, cause `setNextField()` does that check already.

src/dbnode/storage/index/field_terms_iterator_test.go
prateek: re: `allowFn` tests - does the existing `TestFieldsTermsIteratorSimpleSkip` not capture stuff you're worried about?

prateek: re: multiple segments, do you mean Reset()?

prateek: also, re `allowFn` - the prop tests using `fieldsTermsIteratorPropInput`  are generating test situations with those cases too.

src/dbnode/storage/index/types.go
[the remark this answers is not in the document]
prateek: lol sure

src/dbnode/storage/index/aggregate_results.go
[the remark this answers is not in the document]
prateek: sure

src/dbnode/storage/index/fields_terms_iterator.go
arnikola: :+1: misread that this was calling `Next` instead

src/dbnode/storage/index/field_terms_iterator_test.go
arnikola: My bad yeah, was looking at a partial diff so the tests seemed lacking

src/dbnode/storage/index/block.go
richardartoul: Still feels a little sketch to me because I think some of our `Close()` methods return objects to pools

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,189 @@ func (b *block) addQueryResults(
...
+	// instead of checking if the entry is required (duplicates may exist in the results map
+	// already), as it reduces contention on the map itself. Further, the ownership of these
+	// idents is transferred to the results map, which either hangs on to them (if they are new),
+	// or finalizes them if they are duplicates.
+	var (
+		entry            AggregateResultsEntry
+		lastField        []byte
+		lastFieldIsValid bool
+	)
+	// relying on the fact that iterator traversal is in order, we can avoid creating duplicate
```
richardartoul: Can you explain in the comment why there might even be duplicates in the first place? I assume its because we have multiple segments which means multiple FSTs for the same field

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,189 @@ func (b *block) addQueryResults(
...
+
+	batch = append(batch, entry)
+	return batch
+}
+
+func (b *block) pooledID(id []byte) ident.ID {
+	data := b.bytesPool.Get(len(id))
+	data.IncRef()
+	data.AppendAll(id)
+	data.DecRef()
```
richardartoul: hmm, seems like this ID will have no refs for awhile until we add it to the results :/ Nothing we can do about that I assume? adding more accounting probably not worth it since these are really expensive

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,189 @@ func (b *block) addQueryResults(
...
+	return b.idPool.BinaryID(data)
+}
+
+func (b *block) addAggregateResults(
+	cancellable *resource.CancellableLifetime,
+	results AggregateResults,
+	batch []AggregateResultsEntry,
+) ([]AggregateResultsEntry, int, error) {
+	// Checkout the lifetime of the query before adding results
```
richardartoul: nit: period at the end of all these comments

src/dbnode/storage/index/field_terms_iterator_prop_test.go
```diff
@@ -0,0 +1,206 @@
...
+	defer ctrl.Finish()
+
+	parameters := gopter.DefaultTestParameters()
+	seed := time.Now().UnixNano()
+	parameters.MinSuccessfulTests = 100
+	parameters.MaxSize = 40
+	parameters.Rng = rand.New(rand.NewSource(seed))
+	properties := gopter.NewProperties(parameters)
+
+	properties.Property("Fields Terms Iteration doesn't blow up", prop.ForAll(
```
richardartoul: Can you add a comment about the motivation for why we need a separate test just to make sure there are no panics when we already have a correctness test (that presumably might also catch panics)

src/dbnode/storage/index/fields_terms_iterator.go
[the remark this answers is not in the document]
richardartoul: this only happens at package initialization anyways right? who cares haha

src/dbnode/storage/index/fields_terms_iterator.go
```diff
@@ -0,0 +1,200 @@
...
+	Next() bool
+
+	// Current returns the current element.
+	// NB: the element returned is only valid until the subsequent call to Next().
+	Current() (field, term []byte)
+
+	// Err returns any errors encountered during iteration.
+	Err() error
+
+	// Close releases any resources held by the iterator.
```
richardartoul: Maybe add a comment explicitly saying this will not return the iter to the pool and that anything implementing this interface should explicitly support double closes

src/dbnode/storage/index/fields_terms_iterator.go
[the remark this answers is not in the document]
prateek: lol exactly. I mean if the compiler didn't get rid of this stuff, it'd expand the binary size. But the extra 8 bytes, even if we do this for 10K types across the codebase would only add 0.01% to our current binary.

src/dbnode/storage/index/block.go
prateek: that's a fair point. I'll rework to not rely on this behaviour.

src/dbnode/storage/index/fields_terms_iterator.go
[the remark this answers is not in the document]
arnikola: But what if we did it for 10B types? 🤔

src/dbnode/storage/index.go
```diff
@@ -890,7 +909,12 @@ func (i *nsIndex) AggregateQuery(
 		TermFilter: opts.TermFilter,
 		Type:       opts.Type,
 	})
-	exhaustive, err := i.query(ctx, query, results, opts.QueryOptions)
+	// use appropriate fn to query underlying blocks.
+	fn := i.execBlockQueryFn
+	if query.Equal(idx.NewAllQuery()) {
```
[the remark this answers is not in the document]
prateek: lol it's not going to have an impact on perf but sure.

src/dbnode/storage/index.go
```diff
@@ -1085,6 +1079,81 @@ func (i *nsIndex) query(
...
+
+	state.Lock()
+	defer state.Unlock()
+
+	if err != nil {
+		state.multiErr = state.multiErr.Add(err)
+		return
+	}
+
+	if blockExhaustive {
```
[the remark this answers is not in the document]
prateek: sure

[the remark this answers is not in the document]
prateek: setting to `state.exhaustive = state.exhaustive && blockExhaustive` always works so doing that instead

src/dbnode/storage/index.go
```diff
@@ -1085,6 +1079,81 @@ func (i *nsIndex) query(
...
```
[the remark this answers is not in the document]
prateek: same as above.

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -104,6 +104,49 @@ func (r *aggregatedResults) AddDocuments(batch []doc.Document) (int, error) {
...
+	return r.aggregateOpts
+}
+
+func (r *aggregatedResults) AddFields(batch []AggregateResultsEntry) (int, error) {
+	r.Lock()
+	for _, entry := range batch {
+		f := entry.Field
+		aggValues, ok := r.resultsMap.Get(f)
+		if !ok {
+			aggValues = r.valuesPool.Get()
```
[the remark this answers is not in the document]
prateek: The way it's supposed to work: we register a Finalizer on the incoming context to ensure the Finalize() method on the object is called (which in turn calls `Reset(nil, ...)` which does the actual releasing). 

That said, the current code I put up is doing the same thing as the Query() path, which doesn't look to be registering a Finalizer on the context. Will put up another PR for query and address the Aggregate path here.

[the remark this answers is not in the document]
prateek: https://github.com/m3db/m3/pull/1567 as the follow up.

src/dbnode/storage/index/aggregate_results.go
```diff
@@ -104,6 +104,49 @@ func (r *aggregatedResults) AddDocuments(batch []doc.Document) (int, error) {
 	return size, err
 }
 
+func (r *aggregatedResults) AggregateResultsOptions() AggregateResultsOptions {
+	return r.aggregateOpts
+}
+
+func (r *aggregatedResults) AddFields(batch []AggregateResultsEntry) (int, error) {
```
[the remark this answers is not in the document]
prateek: fair, will simplify

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,189 @@ func (b *block) addQueryResults(
...
+	size := results.Size()
+	batch := b.aggEntryPool.Get()
+	batchSize := cap(batch)
+	if batchSize == 0 {
+		batchSize = defaultAggregateResultsEntryBatchSize
+	}
+
+	aggOpts := results.AggregateResultsOptions()
+	iterateTerms := aggOpts.Type == AggregateTagNamesAndValues
+	defer func() {
```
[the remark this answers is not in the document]
prateek: re: two defers. not a fan of doing that unless we really need the perf, cause its really easy to mess up the unlock calls as more changes are made. I'd feel much better about doing the version you suggested if we had something like https://github.com/m3db/build-tools/issues/14. Going to stick with the extra 100ns for the double defer in this path for now. 

re: var decls, will do.

src/dbnode/storage/index/block.go
```diff
@@ -884,6 +919,189 @@ func (b *block) addQueryResults(
...
+	segs := b.segmentsWithRLock()
+	for _, s := range segs {
+		if opts.LimitExceeded(size) {
+			break
+		}
+
+		err = iter.Reset(s, fieldsAndTermsIteratorOpts{
+			iterateTerms: iterateTerms,
+			allowFn: func(field []byte) bool {
+				// skip any field names that we shouldn't allow.
```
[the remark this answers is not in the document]
prateek: re: moving out of the loop, for sure. 

re: newAllowFn, premature at this point. Can do it if we ever have a second reserved field.

src/dbnode/storage/index/block.go
prateek: The ref is only the bytes backing this ident, which the ident will take a ref to right after this line when we transfer the bytes to it (https://github.com/m3db/m3/blob/master/src/x/ident/identifier_pool.go#L105)

src/dbnode/storage/index/block.go
prateek: sure

src/dbnode/storage/index/field_terms_iterator_prop_test.go
prateek: tl;dr - correctness prop test ensures we behave correctly on the happy path. this prop tests ensures we don't panic unless the underlying itself iterator panics. 

will add a comment.

src/dbnode/storage/index/fields_terms_iterator.go
prateek: as mentioned earlier, reworked to avoid this dependency.

src/dbnode/storage/index/block.go
prateek: yeah duplicates are due to multiple FSTs. order of iteration is guaranteed as we iterate one FST at a time. 

i'll add more nuance to the comment.

prateek:
```
	// we are iterating multiple segments so we may receive duplicates (same field/term), but
	// as we are iterating one segment at a time, and because the underlying index structures
	// are FSTs, we rely on the fact that iterator traversal is in order to avoid creating duplicate
	// entries for the same fields, by checking the last batch entry to see if the bytes are
	// the same.
	// It's easier to consider an example, say we have a segment with fields/terms:
	// (f1, t1), (f1, t2), ..., (fn, t1), ..., (fn, tn)
	// as we iterate in order, we receive (f1, t1) and then (f1, t2) we can avoid the repeated f1
	// allocation if the previous entry has the same value.
	// NB: this isn't strictly true because when we switch iterating between segments,
	// the fields/terms switch in an order which doesn't have to be strictly lexicographic. In that
	// instance however, the only downside is we would be allocating more. i.e. this is just an
	// optimisation, it doesn't affect correctness.
```

prateek: added ^^

src/dbnode/storage/index/block.go
prateek: done

src/dbnode/storage/index.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/aggregate_results.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/block.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/block.go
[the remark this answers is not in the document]
prateek: done

src/dbnode/storage/index/block.go
prateek: done

src/dbnode/storage/index/field_terms_iterator_prop_test.go
prateek: done

src/dbnode/storage/index/fields_terms_iterator.go
prateek: done

src/dbnode/storage/index/aggregate_results.go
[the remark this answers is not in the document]
prateek: yeah Reset is O(items in the map); so should be free considering the Put is doing the right thing. Don't need to do it tho, we ensure to cleanup before the Put 

I prefer this approach (assume object returned from pool is valid) cause it puts the cleanup penalty on the last callsite to use the object (as opposed to the new callsite to receive the object). Seems like a "fairer" way to tax users

src/dbnode/storage/index/field_terms_iterator_prop_test.go
prateek: sure thing

src/dbnode/storage/index/block.go
```diff
@@ -823,30 +852,235 @@ func (b *block) addQueryResults(
 	results BaseResults,
 	batch []doc.Document,
 ) ([]doc.Document, int, error) {
-	// Checkout the lifetime of the query before adding results
+	// checkout the lifetime of the query before adding results.
```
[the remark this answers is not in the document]
prateek: oh really? I've been trying to stay all lower case for a while. Don't feel strongly about which one but we should pick a convention and stick to it.

src/dbnode/storage/index/block.go
```diff
@@ -823,30 +852,235 @@ func (b *block) addQueryResults(
...
+	iter, err := b.newFieldsAndTermsIteratorFn(nil, iterateOpts)
+	if err != nil {
+		return false, err
+	}
+
+	var (
+		size       = results.Size()
+		batch      = b.opts.AggregateResultsEntryArrayPool().Get()
+		batchSize  = cap(batch)
+		iterClosed = false // tracking whether we need to free the iterator at the end.
```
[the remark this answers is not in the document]
prateek: nah so it's like the standard cleanup pattern we use in a few places

```
object := ...
cleanedUp := false

defer func() {
  if !cleanedUp {
    object.Cleanup()
  }
}()

// do thing 1 for object
// if it fails, just early `return`

// so on

// finally, 
cleanedUp = true
if err :=object.Cleanup(); err != nil {
  return err
}
```

this way we can be sure all exit paths from the function cleanup the object; either because we manually do it; or the defer takes care of it. 

without the defer (and the var). i'd have to interlace the cleanup calls at every exit point from the function. which seems brittle at best.

src/dbnode/storage/index/field_terms_iterator_prop_test.go
prateek: done

PR #1547, opened 2019-04-11
Add 0.8.2 changelog

benraskin92:


CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
...
+
+## New Features
+
+- **M3DB**: Add Jaeger tracing to M3DB QueryIDs path (#1506)
+
+## Performance
+
+- **M3DB**: Add postingsList caching for FieldQuery (#1530)
+- **M3DB**: Add write & read latency histogram metrics to M3DB client and emit from M3Coordinator (#1533)
+- **M3DB**: Persist block at retention edge when building index after peer streaming (#1531_
```
schallert: `s/(#1531_/(#1531)/`

CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
 # Changelog
 
+# 0.8.2 (2019-04-11)
+
+## Migration Disclaimer
+
+If you have the following set in your M3Query or M3Coordinator config, you should remove it:
```
schallert: Can you explain why? This sounds pretty vague on its own. How would users have ended up with this config? What relevance does it have? Why do we need to change it?

The migration disclaimer for 0.7.0 commit log format changes is a helpful example I think

CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
 # Changelog
 
+# 0.8.2 (2019-04-11)
+
+## Migration Disclaimer
+
+If you have the following set in your M3Query or M3Coordinator config, you should remove it. This is due to a bug in the LRU conversion cache within M3Query that is fixed in this release (see below).
```
prateek: will it cause issues if they do not?

prateek: I'm asking because I want us to minimise the number of migration disclaimers we issue. For e.g. in this case if the process works fine, it'd be better to just mark a deprecating warning; i.e. something end users do not _have_ to act on. Migration disclaimers (at least the one we've issued in the past), users cannot ignore.

benraskin92: Yeah it won't cause any issues (at least not right away). But if we add a `cache` section in the config at some point, it would break.

benraskin92: Happy to make this a warning or something instead.

arnikola: I'd rather delete this, if it breaks in future we can deal with it then right?

prateek: Cool, in that case let’s remove from the release notes. In the future we’ll deal with this by adding a log message in the code which detects the presence of the deprecated fields.

prateek: Hm just ensure it doesn’t break once. I remember vaguely our config parser for m3db was set in strict mode and would barf at unknown fields. Not sure if query uses the same thing?

benraskin92: It actually does break it:
```
unable to load config from /Users/benjaminraskin/gocode/src/github.com/m3db/m3/src/dbnode/config/m3dbnode-local-etcd.yml: yaml: unmarshal errors:
  line 29: field cache not found in type config.Configuration
```

benraskin92: Having said that, I'm not sure how many people actually set this in their config.

CHANGELOG.md
```diff
@@ -1,5 +1,30 @@
...
+```
+
+## New Features
+
+- **M3DB**: Add Jaeger tracing to M3DB QueryIDs path (#1506)
+
+## Performance
+
+- **M3DB**: Add postingsList caching for FieldQuery (#1530)
+- **M3DB**: Add write & read latency histogram metrics to M3DB client and emit from M3Coordinator (#1533)
```
richardartoul: didnt we add a coordinator grafana dashboard too or am I thinking of something else

benraskin92: Haven't landed that yet. Will land it this week once i get some time.

PR #1548, opened 2019-04-11
[query] Deprecate query conversion cache

benraskin92:


src/cmd/services/m3query/config/config.go
```diff
@@ -122,6 +122,9 @@ type Configuration struct {
 
 	// ResultOptions are the results options for query.
 	ResultOptions ResultOptions `yaml:"resultOptions"`
+
+	// Cache configurations.
```
prateek: nit: add a comment explaining deprecation and rename var. something like:

```
// Cache configurations.
// 
// Deprecated: cache configurations are no longer supported. Remove from file. 
DeprecatedCache CacheConfiguration `yaml:"cache"`
```

prateek:
looks like you need to update tests for this too

src/cmd/services/m3query/config/config.go
benraskin92: Okay fair enough - i was gonna leave in case we wanted to use this field again, but we can just change the comment if we do.

PR #1555, opened 2019-04-15
Evict cached block after removing bucket

justinjc:
A bunch of code clean up in this diff, but the biggest change here is that if the buffer tick evicts a bucket from the buffer, it will pass that information back to the series so that the series knows to remove its cached block (the reason for this in the comments).


src/dbnode/storage/series/buffer.go
```diff
@@ -241,21 +243,17 @@ func (b *dbBuffer) Tick(blockStates map[xtime.UnixNano]BlockState) bufferTickRes
...
-			// sure that we bust the cache as well.
-			b.removeBucketVersionsAt(tNano.ToTime())
-			evictedBuckets++
+			b.removeBucketVersionsAt(t)
+			// Pass which bucket got evicted from the buffer to the series.
+			// Data gets read in order of precedence: buffer -> cache -> disk.
+			// After a bucket gets removed from the buffer, data from the cache
+			// will be served. However, since data just got persisted to disk,
+			// the cached block is now stale, therefore we need to evict that
+			// block from cache so that the new data can be retrieved from disk.
```
richardartoul: Can you remind me why we settled on evicting instead of merging? Its just easier right? If so might be nice to leave a comment here thats like: "Instead of evicting, we could perform a merge instead, but that would complicate the code for XYZ reasons"

src/dbnode/storage/series/buffer.go
```diff
@@ -114,13 +114,15 @@ type bufferStats struct {
 
 type bufferTickResult struct {
 	mergedOutOfOrderBlocks int
-	evictedBuckets         int
+	evictedBucketTimes     []time.Time
```
richardartoul: Kinda sucks to allocate millions of tiny arrays temporarily :/

Few options:

1. have both a fixed-size array AND a slice field and use the array field as long as its big enough, otherwise defer to the slice field (which will be a nil pointer that costs us nothing 99.999% of the time)
2. Pass in a slice to the tick method
3. modify the `BlockState` struct to have an extra state field (really gross)

I don't really like any of them, but this seems like a lot of allocations. @prateek thoughts?

src/dbnode/storage/series/buffer.go
```diff
@@ -614,8 +614,13 @@ func (b *dbBuffer) inOrderBlockStartsRemove(removeTime time.Time) {
 	}
 }
 
-// BufferBucketVersions is a container for different versions (from
-// different flushes to disk) of buffer buckets in the database.
+// BufferBucketVersions is a container for different versions of buffer buckets.
+// Bucket versions are how the buffer separates writes that have been written
+// to disk as a fileset and writes that have not. The bucket with a version of
+// `writableBucketVer` is the bucket that all writes go into (as thus us the
```
richardartoul: `(and thus is the bucket version that has not yet been persisted...` right?

src/dbnode/storage/series/types.go
```diff
@@ -335,16 +335,28 @@ func (s Stats) IncCreatedEncoders() {
...
-
 	// WarmWrite represents warm writes (within the buffer past/future window).
-	WarmWrite
+	WarmWrite WriteType = iota
 
 	// ColdWrite represents cold writes (outside the buffer past/future window).
 	ColdWrite
 )
 
+// BootstrapWriteType is the write type assigned for bootstraps.
```
richardartoul: Its not really arbitrary is it? I.E if you put data for a block that has already been flushed into the buffer and mark it as "warm" unless the regular flush cycle knows to account for that (which I don't think it currently does) something will get messed up I think (I think it won't get flushed because the regular flush cycle won't try to flush it)

src/dbnode/storage/series/series.go
```diff
@@ -159,7 +172,7 @@ func (s *dbSeries) updateBlocksWithLock(blockStates map[xtime.UnixNano]BlockStat
 	)
 	for startNano, currBlock := range s.cachedBlocks.AllBlocks() {
 		start := startNano.ToTime()
-		if start.Before(expireCutoff) {
+		if start.Before(expireCutoff) || timeExistsInSlice(start, evictedBucketTimes) {
```
richardartoul: Do you think this will be better / faster than just doing two loops?

```golang
for _, t := range evictedBucketTimes {
    s.cachedBlocks.RemoveBlockAt(t)
}

for startNano, currBlock := range s.cachedBlocks.AllBlocks() {
    ...
}
```

src/dbnode/storage/series/buffer.go
justinjc: Right. I think it's unclear which one is "more correct" in this circumstance, so it makes sense to do the cleaner, easier one. I'll put in a comment.

src/dbnode/storage/series/series.go
justinjc: I thought about that, and while I think the performance difference is probably negligible, I preferred not refactor out the logic within that scope (including the closing the block and updating the result).

src/dbnode/storage/series/types.go
justinjc: Hmm you're right, looks like we'd need more logic to keep track and include those blocks as part of the flush loop. Making it ColdWrite will avoid this issue since we'd be looking through the  entire range of possible blocks.

justinjc: True, while the time range the flush manager looks at is the entire retention period, we'd need additional logic to mark the flush state accordingly to signify that the block requires a flush. Setting this to ColdWrite may be simpler in this case, since all ColdWrites get persisted every compaction cycle.

src/dbnode/storage/series/buffer.go
```diff
@@ -114,13 +114,55 @@ type bufferStats struct {
...
+// times is a struct that holds an unknown number of times. This is used to
+// to avoid heap allocations as much as possible by trying to not allocate a
+// slice of times. To do this, `timesArraySize` needs to be strategically sized
+// such that for the vast majority of the time, the internal array can hold all
+// the times required so that `slice` is nil.
+type times struct {
+	arrIdx int
+	arr    [timesArraySize]time.Time
+	slice  []time.Time
+}
```
justinjc: @prateek do you think this is a little too ugly? :/

prateek: nah we've had to do this in other places too.

src/dbnode/storage/series/buffer.go
```diff
@@ -52,8 +52,8 @@ var (
 )
 
 const (
-	bucketsCacheSize = 2
-
+	bucketsCacheSize  = 2
+	timesArraySize    = 8
```
prateek: more a question: how'd you arrive at 8?

src/dbnode/storage/series/buffer.go
```diff
@@ -114,13 +114,55 @@ type bufferStats struct {
...
-	evictedBuckets         int
+	evictedBucketTimes     times
+}
+
+// times is a struct that holds an unknown number of times. This is used to
+// to avoid heap allocations as much as possible by trying to not allocate a
+// slice of times. To do this, `timesArraySize` needs to be strategically sized
+// such that for the vast majority of the time, the internal array can hold all
+// the times required so that `slice` is nil.
+type times struct {
```
prateek: maybe add a comment to explain that this should only be used via the methods provided

src/dbnode/storage/series/buffer.go
```diff
@@ -114,13 +114,55 @@ type bufferStats struct {
...
+	arrIdx int
+	arr    [timesArraySize]time.Time
+	slice  []time.Time
+}
+
+func (t times) add(newTime time.Time) times {
+	if t.arrIdx < timesArraySize {
```
prateek: nit: paranoia, but mind using `cap(t.arr)` or `len(t.arr)` here instead.

src/dbnode/storage/series/buffer.go
prateek: naming nit: `times` is really vague. pick something that explains what its for. like `optimisedTimes` or something?

src/dbnode/storage/series/buffer.go
```diff
@@ -114,13 +114,55 @@ type bufferStats struct {
...
+}
+
+// times is a struct that holds an unknown number of times. This is used to
+// to avoid heap allocations as much as possible by trying to not allocate a
+// slice of times. To do this, `timesArraySize` needs to be strategically sized
+// such that for the vast majority of the time, the internal array can hold all
+// the times required so that `slice` is nil.
+type times struct {
+	arrIdx int
+	arr    [timesArraySize]time.Time
```
prateek: considering you're doing this to be incredibly space conscious, would using xtime.UnixNano instead of time.Time be better here?

src/dbnode/storage/series/buffer.go
```diff
@@ -114,13 +114,55 @@ type bufferStats struct {
...
+	slice  []time.Time
+}
+
+func (t times) add(newTime time.Time) times {
```
prateek: more a question: why avoid using a pointer receiver? do you find it affects escape analysis?

prateek: if so, mind putting in a test to ensure we don't change that behaviour accidentally

src/dbnode/storage/series/buffer.go
prateek: do you want to add tests to ensure we're not allocating unless necessary? (can test this by seeing `cap(times.slice)` isn't set unless arrIdx is >= timesArraySize)

src/dbnode/storage/series/buffer.go
justinjc: Double checked with `go build -gcflags='-m'` - it doesn't escape. I'll make it a pointer receiver to have it more usable.

src/dbnode/storage/series/buffer.go
justinjc: Mostly just a guess. This represents the maximum number of blocks we think we'll be flushing at one time. We don't really know what out of order writes will be like for now. I'll put in a todo to figure out if 8 is sane.

src/dbnode/storage/series/buffer.go
justinjc: Will add a comment 👍

I'll call it `evictedTimes` since it should really only be used for this case since the array size should be tuned specifically for this.

Will add tests specifically for this struct.

src/dbnode/storage/series/buffer.go
```diff
@@ -664,19 +718,19 @@ func (b *BufferBucketVersions) write(
...
+func (b *BufferBucketVersions) merge(writeType WriteType) (int, error) {
 	res := 0
 	for _, bucket := range b.buckets {
 		// Only makes sense to merge buckets that are writable.
-		if bucket.version == writableBucketVer && wType == bucket.wType {
+		if bucket.version == writableBucketVer && writeType == bucket.writeType {
 			merges, err := bucket.merge()
 			if err != nil {
-				return res, nil
+				return 0, err
```
richardartoul: Can you verify this error will get propagated to something that will log it eventually? That would have saved me so much time today

src/dbnode/storage/series/buffer.go
```diff
@@ -723,44 +777,47 @@ func (b *BufferBucketVersions) lastRead() time.Time {
...
 	newBucket := b.bucketPool.Get()
-	newBucket.resetTo(b.start, wType, b.opts)
+	newBucket.resetTo(b.start, writeType, b.opts)
 	b.buckets = append(b.buckets, newBucket)
 	return newBucket
 }
 
-func (b *BufferBucketVersions) toStreams(ctx context.Context) ([]xio.SegmentReader, error) {
+// mergeToStreams merges each buffer bucket version's streams into one, then
+// returning a single stream for each buffer bucket version.
```
richardartoul: `returns`

src/dbnode/storage/series/buffer.go
justinjc: Sorry about that. Just verified that this error will always get eventually logged, mostly by passing it all the way up to the fs manager where it is logged.

justinjc:
> Could you just change the pool get in `mergeStreamsToEncoder` to be:
> 
> ```go
> iter := bopts.MultiReaderIteratorPool().Get()
> ```

That's what it currently is right now. I assume you meant:
```go
iter := opts.MultiReaderIteratorPool().Get()
```
Which is the one that gets passed from `server.go`. Our options are all over the place.

src/dbnode/storage/series/buffer.go
prateek: Sounds like a plan. If you find a lot of variance, we’ll just have to pool the evictedTimes later

prateek:
> > Could you just change the pool get in `mergeStreamsToEncoder` to be:
> > ```go
> > iter := bopts.MultiReaderIteratorPool().Get()
> > ```
> 
> That's what it currently is right now. I assume you meant:
> 
> ```go
> iter := opts.MultiReaderIteratorPool().Get()
> ```
> 
> Which is the one that gets passed from `server.go`. Our options are all over the place.

Unrelated so don’t want to block, but is there a reason those two shouldn’t resolve to the same underlying pool?

PR #1556, opened 2019-04-15
Start rule watches asynchronously

richardartoul:
Start the watches in background goroutines so that if the store is unavailable they timeout (approximately) in unison. This prevents the timeouts from stacking on top of each other when the store is unavailable and causing a delay of timeout_duration * num_rules.

### DIFF
--- a/src/metrics/matcher/namespaces.go
+++ b/src/metrics/matcher/namespaces.go
@@ -255,6 +255,7 @@ func (n *namespaces) process(value interface{}) error {
 	n.Lock()
 	defer n.Unlock()
 
+	var watchWg sync.WaitGroup
 	for _, entry := range incoming.Iter() {
 		namespace, elem := entry.Key(), rules.Namespace(entry.Value())
 		nsName, snapshots := elem.Name(), elem.Snapshots()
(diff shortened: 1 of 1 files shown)


src/metrics/matcher/namespaces.go
```diff
@@ -284,25 +285,36 @@ func (n *namespaces) process(value interface{}) error {
...
-					log.NewField("ruleSetKey", ruleSet.Key()),
-					log.NewErrField(err),
-				).Error("failed to watch ruleset updates")
-			}
+
+			watchWg.Add(1)
+			go func() {
+				// Start the watches in background goroutines so that if the store is unavailable they timeout
+				// (approximately) in unison. This prevents the timeouts from stacking on top of each
+				// other when the store is unavailable and causing a delay of timeout_duration * num_rules.
```
cw9: What's the impact of the delay?

richardartoul: @cw9 Puppet restarted all the collectors and R2 was misconfigured so all the collectors got stuck for 10 minutes waiting on these initializations because the keys were missing

src/metrics/matcher/namespaces.go
```diff
@@ -255,6 +255,7 @@ func (n *namespaces) process(value interface{}) error {
 	n.Lock()
 	defer n.Unlock()
 
+	watchWg := &sync.WaitGroup{}
```
prateek: nit: `var watchWg sync.WaitGroup` is more conventional

Issue #1471, opened 2019-03-17
M3DB Complex Types / Compresed Protobufs Followups


richardartoul:
#1426 is getting out of hand so this issue will track issues that will be solved in a follow up P.R:

- [x] Do custom compression on bool fields
- [ ] Consider removing int sig bit tracking as it doesn't seem very workload generic
- [ ] Recursively explored nested submessages to look for fields that we can apply custom compression to
- [x] Add support for embedded messages
- [x] Add support for repeated fields
- [x] Add support for maps
- [x] Avoid hash collisions in encoder
- [x] Refactor Proto Encoder / Iterator to not embed M3TSZ Encoder/Iterator
- [x] [Implement a MarshalInto method in the ProtoReflect library to save allocations] (https://github.com/jhump/protoreflect/issues/176) - Implemented the method in the library, still need to modify M3DB code to use it
- [x] Support mid-stream schema changes on the encoder and iterator
- [x] Make it so that the encoder can optionally encode all of the type information into the stream itself so it can still be decoded even if the schema is not present.
- [ ] Reduce string allocations in the iterator by preprocessing schemas so that all strings are treated as []byte fields
- [x] Document compression format
- [ ] Make accepting writes with unknown fields configurable
- [ ] Consider using a "tail" to seal the byte stream by writing out the final two bits
- [ ] Reset dynamic messages in encoder instead of reallocating if the schema hasn't changed once we have an easy way to compare two schemas
- [x] Consider modifying the Protobuf library to support storing values without having to allocate `interface{}` so that we can have type-specific setters and getters
- [x] Add a `Stats()` method to the encoder for keeping track of compression performance + use it in tests to make sure string compression is working properly
- [ ] Pass a buffer into Snapshot() to prevent having to allocate a large one over and over again if it exceeds the largest bucket size

PR #1558, opened 2019-04-16
Support mid stream schema changes in the protobuf encoder/iterator + other follow ups

richardartoul:
Addresses some of #1471 

- [X] Mid stream schema changes support in the encoder and iterator
- [X] Updated encoding documentation
- [X] Add `Stats()` method on the encoder for tracking compression ratio
- [X] Cleanup

Perf Improvements (57% improvement for some workloads)

- [X] Upgrade to version of protoreflect library that includes upstreamed perf improvements
- [X] Add benchmark
- [X] Use `MarshalAppend()` instead of `Marshal()` to save allocations
- [X] Don't clear every custom encoded field
- [X] Skip vast majority of encodeProto logic if there are no proto-encoded fields
- [X] Use optimized functions for calculating leading and trailing zeros
- [X] Other minor improvements


src/dbnode/encoding/proto/README.md
```diff
@@ -95,20 +95,61 @@ In practice, a Protbuf message can have any number of different fields; for perf
...
+
+Time unit changes must be tracked manually (instead of deferring to the M3TSZ timestamp delta-of-delta encoder, which can handle this independently) because the M3TSZ encoder relies on a custom marker scheme to indicate time unit changes that is not suitable for the Protbuf encoding format.
+The M3TSZ timestamp encoder avoids using a control bit for each write by using a marker which contains a prefix that could not be generated by any possible input, and the decoder frequently "looks ahead" for this marker to see if it needs to decode a time unit change.
+The Protobuf encoding scheme has no equivalent "impossible bit combination", so it uses explicit control bits to indicate a time unit change instead.
+
+The table below contains a summary of all the possible per write control bit combinations. Note that `X` is used to denote control bits that will not be included; so even though the format may look inefficient because it requires a maximum of 4 control bits to encode only 6 different combinations, the most common scenario (where the stream contains at
+least one more write and neither the time unit or schema has changed) can be encoded with just a single bit.
```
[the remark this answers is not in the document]
richardartoul: I think the sentence becomes more vague if I don't explicitly state again what has not changed

src/dbnode/encoding/proto/common.go
```diff
@@ -76,11 +76,17 @@ const (
 	// Single bit op codes that get encoded into the compressed stream and
 	// inform the iterator / decoder how it should interpret subsequent
 	// bits.
-	opCodeNoMoreDataOrTimeUnitChange = 0
-	opCodeMoreData                   = 1
+	opCodeNoMoreDataOrTimeUnitChangeAndOrSchemaChange = 0
```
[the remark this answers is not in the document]
richardartoul: hahaha the single bit has a lot of meaning ;)

src/dbnode/encoding/proto/encoder.go
```diff
@@ -72,16 +72,32 @@ type Encoder struct {
...
 	timestampEncoder m3tsz.TimestampEncoder
 }
 
+// EncoderStats contains statistics about the encoders compression performance.
+type EncoderStats struct {
+	UncompressedBytes int
+	CompressedBytes   int
+}
+
+type encoderStats struct {
```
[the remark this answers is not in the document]
richardartoul: No because one of the fields is generated from another piece of state, so if I kept the public one on the encoder I'd end up wasting another 8 bytes per encoder (I know that seems like an over optimization but I have millions of these things in memory and they're already using 10% of memory not including the data so I'm trying to cut as much state out of them as possible to reduce mem usage)

src/dbnode/encoding/proto/README.md
```diff
@@ -95,20 +95,61 @@ In practice, a Protbuf message can have any number of different fields; for perf
 Instead, multiple "logical" streams are interleaved on a per-write basis within one physical stream.
 The remainder of this section outlines the binary format used to accomplish this interleaving.
 
-### Header
+The binary format begins with a stream header and then the remainder of the stream is a sequence of tuples in the form: `<per write header, compressed timestamp, compressed custom encoded fields, Protobuf marshaled fields>`
```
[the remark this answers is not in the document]
richardartoul: I tried this and personally found it more confusing

richardartoul:
@mway re: benchmarks, yep I added a simple one here and a bunch of minor performance improvements that had substantial (57%) impact. Still not good enough though so I need to go and contribute some perf improvements to the proto reflect library now

src/dbnode/encoding/proto/common.go
```diff
@@ -165,18 +171,27 @@ func newCustomFieldState(fieldNum int, fieldType customFieldType) customFieldSta
...
 // TODO(rartoul): Improve this function to be less naive and actually explore nested messages
 // for fields that we can use our custom compression on: https://github.com/m3db/m3/issues/1471
-func customFields(s []customFieldState, schema *desc.MessageDescriptor) []customFieldState {
+func customFields(s []customFieldState, protoFields []int32, schema *desc.MessageDescriptor) ([]customFieldState, []int32) {
+	fields := schema.GetFields()
 	numCustomFields := numCustomFields(schema)
+	numProtoFields := len(fields) - numCustomFields
+
 	if cap(s) >= numCustomFields {
 		s = s[:0]
```
robskillington: nit: Do you need to nil out state on elements inside `customFieldState`?  If there's any pointers there they will still point to things and held onto by the garbage collector.

i.e.
```
for i := range s {
  s[i] = customFieldState{}
}
```

src/dbnode/encoding/proto/common.go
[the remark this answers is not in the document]
robskillington: We definitely have a preference for longer names (for better readability) so that comments can be focused just on the logic rather than identifying vars/etc.

src/dbnode/encoding/proto/common.go
```diff
@@ -76,11 +76,17 @@ const (
...
 	// bits.
-	opCodeNoMoreDataOrTimeUnitChange = 0
-	opCodeMoreData                   = 1
+	opCodeNoMoreDataOrTimeUnitChangeAndOrSchemaChange = 0
+	opCodeMoreData                                    = 1
 
-	opCodeNoMoreData     = 0
-	opCodeTimeUnitChange = 1
+	opCodeNoMoreData                      = 0
+	opCodeTimeUnitChangeAndOrSchemaChange = 1
```
robskillington: Hm how does this work?  Do you now always encode two bits if the More Data bit is set?

src/dbnode/encoding/proto/common.go
richardartoul: I always forget to do that, thanks

src/dbnode/encoding/proto/common.go
richardartoul: One bit if there is just another write, two bits if its the end of the stream, and 4 bits in the case where there is a time unit change or schema change. The README has a good diagram

src/dbnode/encoding/proto/README.md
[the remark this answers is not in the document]
mway: okay! nbd

src/dbnode/encoding/proto/README.md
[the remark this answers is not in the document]
mway: fair enough.

src/dbnode/encoding/proto/encoder.go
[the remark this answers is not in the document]
mway: ah. yes, missed that on the first read. makes sense

src/dbnode/encoding/proto/common.go
```diff
@@ -165,18 +171,30 @@ func newCustomFieldState(fieldNum int, fieldType customFieldType) customFieldSta
...
+	} else {
+		protoFields = make([]int32, 0, numProtoFields)
+	}
+
 	for _, field := range fields {
 		customFieldType, ok := isCustomField(field.GetType(), field.IsRepeated())
 		if !ok {
+			protoFields = append(protoFields, field.GetNumber())
 			continue
 		}
```
mway: unrelated nit: it looks like you should be able to set `fieldState.intEncAndIter.unsighed` inside `newCustomFieldState` unless there's a penalty for doing so, since you have all of the state required in that function (and it seems a little weird to change the state immediately after creating it).

src/dbnode/encoding/proto/encoder.go
```diff
@@ -308,24 +367,27 @@ func (enc *Encoder) reset(start time.Time, capacity int) {
 	enc.lastEncodedDP = ts.Datapoint{}
 	enc.unmarshaled = nil
 
+	// Prevent this from growing too large and remaining in the pools.
+	enc.marshalBuf = nil
+
 	if enc.schema != nil {
-		enc.customFields = customFields(enc.customFields, enc.schema)
+		enc.customFields, enc.protoFields = customFields(enc.customFields, enc.protoFields, enc.schema)
```
mway: should this be part of `Encoder`, since this exact state is being used in the same way in multiple places?

src/dbnode/encoding/proto/vehicle_location_2.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+message VehicleLocation {
```
Haijuncao: can you add tests:
1. no schema change, but schema has with proto values (non-custom fields)
2. no schema change, but there is time unit change
3. no time unit change, has schema change, custom fields change, but proto values do not change
4. no time unit change, schema change, custom fields do not change, but proto values do change
5. no time unit change, schema change, custom fields change, proto values change. 

for case 3, should we expect the control bit per write to be 1XXX?

richardartoul: Sure. The existing tests cover #1 (TestRoundTripPropTest generates schemas that have non-custom fields),  #2 (TestRoundTripProp generates a unique time unit for each write) and #3 (TestRoundTripMidStreamSchemaChanges).

#4 and #5 you're right I should add some tests.

src/dbnode/encoding/proto/common.go
richardartoul: good call

src/dbnode/encoding/proto/vehicle_location_2.proto
richardartoul: Sync'd in person, will add a unit test that includes non-custom fields too for easier debugging

PR #1561, opened 2019-04-16
[query] Offset function

arnikola:
Updated offset to create a downstream processing function rather than fixing in iterators


src/query/executor/transform/exec.go
```diff
@@ -41,14 +64,26 @@ type simpleOpNode interface {
...
+	queryCtx *models.QueryContext,
+	ID parser.NodeID,
+	b block.Block,
+) error {
+	return processBlock(node, controller, queryCtx, ID, b, true)
+}
 
-	defer nextBlock.Close()
-	return controller.Process(queryCtx, nextBlock)
+// ProcessWrapperBlock works similarly to ProcessSimpleBlock, but does not close
```
andrewmains12: This is probably debatable, but maybe best to just not use the helper in this case? There are already functions which don't (for basically the same reason--they need more control over the block lifecycle). Especially because we comment pretty explicitly down below; might be more clear to just manage the lifecycle directly there.

andrewmains12:
General question on this: for most operations, we apply them eagerly--that is, we iterate through the block and create a new block. What makes it such that `offset` needs to be done in a custom block type/iterator instead?

src/query/executor/transform/exec.go
arnikola: Sounds good, did it this way mostly because I didn't realize that we already had a couple of blocks that don't use it, so happy to revert these changes.

arnikola:
> General question on this: for most operations, we apply them eagerly--that is, we iterate through the block and create a new block. What makes it such that `offset` needs to be done in a custom block type/iterator instead?

That's true for most functions, since they generally collapse the input space during execution (temporal functions excluded unfortunately), but we have some functions which we can apply lazily as a simple transform, which can then defer decoding until an aggregation function is ready to collapse the results.

We do have some code in place for lazily applying functions, but it's pretty awkward to use and probably needs an update to look more like this model (i.e. pass down a function into `OffsetBlock`, maybe rename it to something like `FunctionBlock`, and let it do transforms there; as such I guess this PR both adds the offset and a skeleton for building future lazy functions

robskillington:
Nice, this looks a ton better than the other alternative!

PR #1562, opened 2019-04-16
[dbnode] provide schema for proto encoder

Haijuncao:
This pr provides latest schema (including live updates from etcd) to encoder/iterator when encoder/iter are taken out of the pool.If encoder/iterator is reused from pool, we always nil schema at Put, and set schema at Get.

There are in general two approaches when setting schema for an object after it's been taken out of the pool. 
1. Pass schema info down the call stack, set schema after invoking pool.Get (AdminClient takes this approach, series/bufferbucket takes this approach as well to set seriesOpts).
2. Embed a schema injector with the pool, set the schema (with schema injector) after object is taken out of object pool, but before returning it via Get. Encoder/ReaderIterator/MultiReaderIterator takes this approach.

Which approach to take depends on the existing code, the goal is to minimize code changes we have to do, which in general translate to avoiding making changes to api.

Schema is per namespace and not available during db server initialization. Schema became available during dbNamespace construction when it receives metadata from etcd. 

For normal server operations (series write/read), we re-initialize encoderPool/iteratorPool per namespace (duration dbNamespace constructor) which shares the same underlying object pool (i.e. we still have a single object pool for encoders), per-namespace pool contains a schema injector that sets encoder schema with the namespace schema at pool.Get.

series.Options is per namespace as they contains references to per namespace encoder/iter pools. Series/BufferBucket/BufferBucketVersion pools are not re-init per namespace, as the objects are reset with per-ns seriesOpts upon Get. 

For server repair/bootstrap operation, server uses AdminClient to establish sessions to peers to fetch data/index. Session need schema to iterate fetch results if proto is turned on. We pass schema registry via client options to sessions. As session is normally reused across namespaces, we do not want to have per-ns session as it manages connections (therefore heavy/expensive).

DatabaseBlockPool is initialized with a global block.Option at server init time, which do not contain the right encoding pools. We need to reset block.Options on DatabaseBlock object when it is taken out of the pool. We use approach 2 to create per namespace DatabaseBlockPool.

Bootstrappers are constructed and initialized with a global blockOpts (which contains references to encoding pools). We need to pass the right blockOpts (with namespace scoped encoding/block pool) to bootstrappers. We take approach2, per namespace bootstrapper/process.

Both schema injector and schema registry references back to namespace schema history, which is updated dynamically when there are schema updates from etcd.


Haijuncao:
This is a first draft. I have not written any tests. Share early to make sure it is the right direction.

src/dbnode/client/session.go
```diff
@@ -1419,6 +1421,13 @@ func (s *session) fetchIDsAttempt(
 	// multiple times in case of retries.
 	ids := inputIDs.Duplicate()
 
+	if s.opts.ProtoEnabled() {
+		schema, schemaErr = s.opts.SchemaRegistry().GetSchema(inputNamespace)
```
richardartoul: Could we make all of this logic simpler by just creating an object pool that requires you pass a namespace as an argument? Then there is no possibility of future maintainers introducing bugs in this code path.

```golang
type perNSIterPool struct {
    pool encoding.IteratorPool
    registry namespace.SchemaRegistry
}

func (p *perNSIterPool) Get(ns ident.ID) (encoding.Iterator, error) {
    iter := p.pool.Get()
    schema, err := p.registry.GetSchema(ns)
    if err != nil {
        return err
    }
    iter.SetSchema(schema)
    return iter, nil
}
```

Now no one will ever accidentally do this wrong

richardartoul: Could we make all of this logic simpler by just creating an object pool that requires you pass a namespace as an argument? Then there is no possibility of future maintainers introducing bugs in this code path.

```golang
type perNSIterPool struct {
    pool encoding.IteratorPool
    registry namespace.SchemaRegistry
}

func (p *perNSIterPool) Get(ns ident.ID) (encoding.Iterator, error) {
    iter := p.pool.Get()
    schema, err := p.registry.GetSchema(ns)
    if err != nil {
        return err
    }
    iter.SetSchema(schema)
    return iter, nil
}
```

Now no one will ever accidentally do this wrong

src/dbnode/encoding/encoder_pool.go
```diff
@@ -39,10 +40,22 @@ func (p *encoderPool) Init(alloc EncoderAllocate) {
 	})
 }
 
+// ReInit reinitialize the pool with a different object initializer.
+// the new pool shares the save underlying object pool.
+func (p *encoderPool) ReInit(reInit SchemaInjector) EncoderPool {
```
richardartoul: TODO: come back to this

src/dbnode/server/server.go
```diff
@@ -547,6 +546,13 @@ func Run(runOpts RunOptions) {
 		logger.Fatal("could not create m3db client", zap.Error(err))
 	}
 
+	//M3db admin client is passed to database bootstrapper and repairer before db is constructed.
```
richardartoul: Super nit: space after `//` is how I think most of our comments are

src/dbnode/storage/namespace.go
```diff
@@ -348,12 +349,48 @@ func newDatabaseNamespace(
...
+	n.reInitPools()
 	n.initShards(nopts.BootstrapEnabled())
 	go n.reportStatusLoop()
 
 	return n, nil
 }
 
+func (n *dbNamespace) reInitPools() {
+	schemaInjector := func(e encoding.SchemaInjectable) encoding.SchemaInjectable {
+		if schema, ok := n.SchemaHistory().GetLatest(); ok {
```
richardartoul: is schema history synchronized?

richardartoul: Oh I see the namespace locks it...not a huge fan of this because now random parts of the code will be acquiring the namespace lock. Would prefer if there was an independent lock for the registry if that makes sense

Haijuncao:
> @Haijuncao
> 
> Took a look through the code, you went through a lot of parts of the system here, impressive stuff.
> 
> My primary concern is that it seems like we could really simplify the vast majority of the code with a synchronized namespace-aware pool like so:
> 
> ```go
> type perNSIterPool struct {
>     pool encoding.IteratorPool
>     registry namespace.SchemaRegistry
> }
> 
> func (p *perNSIterPool) Get(ns ident.ID) (encoding.Iterator, error) {
>     iter := p.pool.Get()
>     schema, err := p.registry.GetSchema(ns)
>     if err != nil {
>         return err
>     }
>     iter.SetSchema(schema)
>     return iter, nil
> }
> ```
> 
> Basically then the compiler will point out every instance you need to fix and as long as you pass the right namespace there you know the code will always be correct. There will be a little more complexity there in terms of you may not have the registry yet and you may need to initialize it delayed or whatever but I think it will still be a lot simpler than the `ReInit` and `Injector` logic right now.
> 
> The only exception will be the write path where we need to update blocks mid stream in which case you won't need to pass the namespace into the pool because you'll be setting the schema manually with each write anyways.
> 
> Does that make sense?

Passing down nsID touches many MORE code. It is the primary reason why I did not choose this approach (but have per-namespace pool).

Put that aside, the reason why many code are touched is not because the pool is namespace scoped, but because we need to re-initialize pools (any options) after database/namespace is constructed. This is needed regardless of if we have per-namespace pool or per-database pool (which takes nsID via get).

richardartoul:
@Haijuncao I'm not sure why the nsID thing would be more code. In most places where you need an encoder or iterator, you should know which namespace you're attempting to write or read from (the main exception being the buffer where we're gonna pass the schema down)

I also don't understand why the pool needs to be reinitialized. If its one global synchronized pool then you should just be able to update it one place and it will automatically propagate the change to the whole sytem

src/dbnode/client/session.go
Haijuncao: > Could we make all of this logic simpler by just creating an object pool that requires you pass a namespace as an argument? Then there is no possibility of future maintainers introducing bugs in this code path.
> 
> ```go
> type perNSIterPool struct {
>     pool encoding.IteratorPool
>     registry namespace.SchemaRegistry
> }
> 
> func (p *perNSIterPool) Get(ns ident.ID) (encoding.Iterator, error) {
>     iter := p.pool.Get()
>     schema, err := p.registry.GetSchema(ns)
>     if err != nil {
>         return err
>     }
>     iter.SetSchema(schema)
>     return iter, nil
> }
> ```
> 
> Now no one will ever accidentally do this wrong

Regarding robustness, per-namespace pool is safe and backward compatible, in that for proto encoder, if schema is not set, code would crash immediately. That's why I intentionally nil Schema when encoder is returned back to the pool.

Also, pools/options are setup properly during namespace construction and bootstrapper construction, so that other parts of the code do not have to worry about it at all. It is similar to how we do it today, we just delay the pools/options setup from server setup to db construction, rest of the code can continue to depend on options to function.

src/dbnode/storage/namespace.go
Haijuncao: > Oh I see the namespace locks it...not a huge fan of this because now random parts of the code will be acquiring the namespace lock. Would prefer if there was an independent lock for the registry if that makes sense

With database schema registry approach, you'd have to lock it twice, first time to lock db to access OwnedNamespaces, second time to lock dbNamespace to access schema history.

Haijuncao:
> @Haijuncao I'm not sure why the nsID thing would be more code. In most places where you need an encoder or iterator, you should know which namespace you're attempting to write or read from (the main exception being the buffer where we're gonna pass the schema down)
> 
> I also don't understand why the pool needs to be reinitialized. If its one global synchronized pool then you should just be able to update it one place and it will automatically propagate the change to the whole sytem

well, maybe not re-initialize, but we certainly have to move the pool and option initialization code out of server setup into database construction.

src/dbnode/storage/namespace.go
Haijuncao: I see by "registry" you mean "schema history". yes, we can have a lock just for schema history (instead of a namespace wide lock).

prateek:
@Haijuncao to echo @richardartoul's concern, not particularly a fan of the injector/injectable idiom. (or the perNSIterPool).

Is there a reason we can't do (1.) everywhere? 
> 1. Pass schema info down the call stack, set schema after invoking pool.Get (AdminClient takes this approach, series/bufferbucket takes this approach as well to set seriesOpts).

Haijuncao:
> @Haijuncao to echo @richardartoul's concern, not particularly a fan of the injector/injectable idiom. (or the perNSIterPool).
> 
> Is there a reason we can't do (1.) everywhere?
> 
> > 1. Pass schema info down the call stack, set schema after invoking pool.Get (AdminClient takes this approach, series/bufferbucket takes this approach as well to set seriesOpts).
> 
> you are saying SetSchema outside of GET? Two reason:
> 
> 1. That touched many API's to pass down schema via the stack, this is especially true for normal series write/read. For bootstrapper/client code, it is less change because I see that is already kind of happening in many API's with namespace.Metadata.
> 2. SetSchema as a separate step requires special attention, we'd have to put a big comment warning people to always SetSchema after GET, like what we did with SeriesOpts. Though it is not a big deal if we don't, as we will panic fast if people forget to SetSchema. I did take approach 1 for AdminClient, but @richardartoul  has concern because the approach is less worry-free for developers who uses Encoder/Iterator.

@prateek synced up with @richardartoul  and @robskillington quickly, as you alluted to, approach 1 (pass schema thru the callstack) appears more favorable, it is less magic, more consistent and predicable.

Haijuncao:
replaced with #1578

PR #1565, opened 2019-04-17
[query] add list tags endpoint

arnikola:
Adds [this Prometheus endpoint](https://prometheus.io/docs/prometheus/latest/querying/api/#getting-label-names) that lists all available tags


src/query/models/matcher.go
```diff
@@ -38,6 +38,8 @@ func (m MatchType) String() string {
 		return "=~"
 	case MatchNotRegexp:
 		return "!~"
+	case MatchAll:
+		return "*"
```
prateek: this String() method is only used for debugging yea?

src/query/api/v1/handler/prometheus/common.go
```diff
@@ -304,6 +304,35 @@ func renderDefaultTagCompletionResultsJSON(
 	return jw.Close()
 }
 
+// RenderListTagResultsJSON renders list tag results to json format.
+func RenderListTagResultsJSON(
```
prateek: can we rely on `json.Marshal` with the appropriate struct instead of defining our own?

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -0,0 +1,90 @@
...
+
+	opts := storage.NewFetchOptions()
+	result, err := h.storage.CompleteTags(ctx, query, opts)
+	if err != nil {
+		logger.Error("unable to complete tags", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
+		return
+	}
+
+	// TODO: Support multiple result types
```
prateek: what do you mean?

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -0,0 +1,90 @@
...
+		storage: storage,
+	}
+}
+
+func (h *ListTagsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
+	logger := logging.WithContext(ctx)
+	w.Header().Set("Content-Type", "application/json")
+
+	query := &storage.CompleteTagsQuery{
```
prateek: do you want to add the ability to restrict to only a given set of tag field names?

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -0,0 +1,90 @@
...
+		CompleteNameOnly: true,
```
prateek: more a question: what is this for?

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -0,0 +1,90 @@
...
+		TagMatchers:      models.Matchers{{Type: models.MatchAll}},
+
+		// NB: necessarily spans entire possible query range.
+		Start: time.Time{},
```
prateek: this is fine for default but should it be overridable from the incoming request?

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -0,0 +1,90 @@
...
+		End:   time.Now(),
+	}
+
+	opts := storage.NewFetchOptions()
```
prateek: this might be a silly question: what are these options/how come we can just use the defaults?

src/query/api/v1/handler/prometheus/native/list_tags_test.go
```diff
@@ -0,0 +1,154 @@
...
+	w.results = append(w.results, string(b))
+	return len(b), nil
+}
+
+type result struct {
+	Status string   `json:"status"`
+	Data   []string `json:"data"`
+}
+
+func TestFind(t *testing.T) {
```
prateek: if possible could you add a failure test situation too

src/query/api/v1/handler/prometheus/native/list_tags_test.go
```diff
@@ -0,0 +1,154 @@
...
+	logging.InitWithCores(nil)
```
prateek: been trying to push the standard stuff we have for zap now, could you use `src/x/test/NewLogger` to create the test logger and pass down to the appropriate place

src/query/models/matcher.go
arnikola: Yeah

src/query/api/v1/handler/prometheus/native/list_tags.go
arnikola: Copied an existing handler; removed this note, this should only support json

src/query/api/v1/handler/prometheus/native/list_tags.go
arnikola: This is the implementation of the Prom endpoint here, which expects no args and returns all valid tag names only: https://prometheus.io/docs/prometheus/latest/querying/api/#getting-label-names

src/query/api/v1/handler/prometheus/native/list_tags.go
arnikola: This restricts search results to only give tag names rather than tag names and values

src/query/api/v1/handler/prometheus/native/list_tags.go
arnikola: Generally yes, but this particular endpoint is intended to work across all time

src/query/api/v1/handler/prometheus/native/list_tags.go
arnikola: Default is fine here; for context, this includes:

- tracing options (could potentially be added)
- fetch limit (irrelevant for this endpoint)
- block fetch type (irrelevant for this endpoint - this is used to switch between decompressing blocks as soon as fetched vs keeping fetches as series iterators until applying non-lazy functions)
- fanout options (this may change in future if we allow a standalone index to use instead of doing our best guess fanout at the moment, but defaults are fine here for virtually all cases; this is mostly here to allow us to override the graphite query path)

src/query/api/v1/handler/prometheus/native/list_tags_test.go
arnikola: Sure sounds good

src/query/api/v1/handler/prometheus/common.go
arnikola: Yeah, it's an issue with a bunch of our stuff; called out removing json.Marshal in #1590

benraskin92: Also, is there a scenario where `status` isn't `success`?

src/query/api/v1/handler/prometheus/native/list_tags_test.go
arnikola: Called out in #1590 as well; all our other handlers are doing this at the moment, may be nicer to fix them all at once?

src/query/api/v1/handler/prometheus/common.go
arnikola: No; if it errors, it won't use this render function but rather the default error printer

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -0,0 +1,90 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
benraskin92: nit: `2019`

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/query/util/logging"
+	"github.com/m3db/m3/src/x/net/http"
+
+	"go.uber.org/zap"
+)
+
+const (
+	// ListTagsURL is the url for listing tags.
+	ListTagsURL = handler.RoutePrefixV1 + "/labels"
```
benraskin92: nit: name `ListLabelsURL` to be consistent with Prom?

arnikola: I'd rather be consistent internally here; we've been renaming prom `label` endpoints as `tags`

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -0,0 +1,90 @@
...
+	storage storage.Storage,
+) http.Handler {
+	return &ListTagsHandler{
+		storage: storage,
+	}
+}
+
+func (h *ListTagsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
+	logger := logging.WithContext(ctx)
```
benraskin92: This using the new logging (zap) stuff?

src/query/api/v1/handler/prometheus/native/list_tags_test.go
benraskin92: +1

src/query/api/v1/handler/prometheus/common.go
benraskin92: What's the point of having a status field then lol

src/query/api/v1/handler/prometheus/native/list_tags.go
arnikola: Nah, same as the other handlers afaik

src/query/api/v1/handler/prometheus/native/list_tags_test.go
arnikola: I think we're currently using zap for logging; although we may do it in a bit of a weird way with a global zap logger

src/query/api/v1/handler/prometheus/common.go
arnikola: It's expected in the Prom output unfortunately

robskillington: Added a response on that issue, for any endpoints where we return a lot of data we should continue using our JSON writers:
> One thing interesting about the JSON writers, the major reason in the past we did that is because the default JSON marshaller allocates a lot, however if we have JSON write methods for bytes/strings/whatever other concrete types we can avoid allocations when returning responses (which is what we have today with the current common library streamjson writer).

src/query/api/v1/handler/prometheus/native/list_tags_test.go
robskillington: Added this to #1590 under "General":
> Remove use of global Zap logger usage in favor of `src/x/instrument` logger which uses Zap and use `src/x/test/NewLogger` in tests

src/query/api/v1/handler/prometheus/native/list_tags_test.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	store.EXPECT().CompleteTags(gomock.Any(), &listTagsMatcher{}, gomock.Any()).
+		Return(storeResult, nil)
+
+	handler := NewListTagsHandler(store)
+
+	// execute the query
+	w := &writer{}
+	req := &http.Request{
```
robskillington: nit: Better to use `http.NewRequest(...)` I think here? `require.NoError(t, err)` is pretty cheap to call after construction.

src/query/api/v1/handler/prometheus/native/list_tags_test.go
```diff
@@ -0,0 +1,178 @@
...
+		},
+	}
+
+	store.EXPECT().CompleteTags(gomock.Any(), &listTagsMatcher{}, gomock.Any()).
+		Return(storeResult, nil)
+
+	handler := NewListTagsHandler(store)
+
+	// execute the query
+	w := &writer{}
```
robskillington: nit: Better to use `httptest.NewRecorder(...)` here:
https://play.golang.org/p/y_M3RA0YUDB

And:
https://golang.org/pkg/net/http/httptest/

src/query/api/v1/handler/prometheus/native/list_tags_test.go
```diff
@@ -0,0 +1,178 @@
...
+		Return(nil, errors.New("err"))
+	handler := NewListTagsHandler(store)
+
+	// execute the query
+	w := &writer{}
+	req := &http.Request{
+		URL: &url.URL{
+			RawQuery: "",
+		},
+	}
```
robskillington: Same notes about using `httptest.NewRecorder(...)` and `httptest.NewRequest(...)` here.

src/query/api/v1/handler/prometheus/native/list_tags_test.go
arnikola: Yeah this seems much nicer :+1:

src/query/api/v1/handler/prometheus/common.go
arnikola: Yeah was going to write some quick benchmarks when it came time to consider swapping it out; I think our initial thinking here was if we're using this writer internally there's a good reason

PR #1566, opened 2019-04-17
[dbnode] Only allocate index block compactors before use

robskillington:
This avoids the excessive memory usage when a lot of index blocks are created and compactors with a two roaring bitmaps with up front allocated buckets take up considerable memory usage.


src/dbnode/storage/index/block.go
```diff
@@ -799,7 +813,8 @@ func (b *block) Query(
 	}
 
 	size := results.Size()
-	batch := b.docsPool.Get()
+	docsPool := b.opts.DocumentArrayPool()
```
richardartoul: Super nit: `var` block would be nice here.

src/dbnode/storage/index/block.go
```diff
@@ -110,10 +110,10 @@ type block struct {
 	blockStart    time.Time
 	blockEnd      time.Time
 	blockSize     time.Duration
+	blockOpts     BlockOptions
 	opts          Options
 	iopts         instrument.Options
 	nsMD          namespace.Metadata
-	docsPool      doc.DocumentArrayPool
```
prateek: super nit: mind grouping the lazily allocated objects are grouped together in a struct and there's a comment about using the ensure method before using

robskillington: Good call.

src/dbnode/storage/index/block.go
robskillington: Sure yeah.

PR #1574, opened 2019-04-19
[dbnode] Fix index flush recovery when previous flush attempts have failed

robskillington:
Prior to this change if an index flush failed to complete and write out a checkpoint file, it would leave dirty files that would never be able to be overwritten.

This changes the behavior to only check that the checkpoint file does not exist and is valid, rather than check if a segment file exists or not before writing it.

It also updates all call sites to correctly always validate a checkpoint file rather than just check for existence in other parts of the `persist/fs` package.


src/dbnode/persist/fs/files.go
```diff
@@ -115,12 +115,18 @@ func (f FileSetFile) IsZero() bool {
 	return len(f.AbsoluteFilepaths) == 0
 }
 
-// HasCheckpointFile returns a bool indicating whether the given set of
+// HasCompleteCheckpointFile returns a bool indicating whether the given set of
 // fileset files has a checkpoint file.
-func (f FileSetFile) HasCheckpointFile() bool {
+func (f FileSetFile) HasCompleteCheckpointFile() bool {
 	for _, fileName := range f.AbsoluteFilepaths {
 		if strings.Contains(fileName, checkpointFileSuffix) {
```
justinjc: Probably don't need this line anymore since `CompleteCheckpointFileExists` does this already.

robskillington: True but that will cause an "Invariant error" to be allocated which I'd like to avoid in the case that we're scanning a lot of files.

src/dbnode/persist/fs/files.go
```diff
@@ -73,14 +73,27 @@ var (
 
 type fileOpener func(filePath string) (*os.File, error)
 
+// LazyEvalBool is a boolean is lazily evaluated.
+type LazyEvalBool uint8
+
+const (
+	// EvalNone indicated the boolean has not been evaluated.
```
justinjc: `indicates`

src/dbnode/persist/fs/files.go
```diff
@@ -73,14 +73,27 @@ var (
 
 type fileOpener func(filePath string) (*os.File, error)
 
+// LazyEvalBool is a boolean is lazily evaluated.
```
justinjc: `boolean that is`

PR #1576, opened 2019-04-19
Open new commitlogs async

richardartoul:
The README.md has more details, but the basic idea with this P.R is that everytime we rotated the commitlog file we would see a really big spike in the commit log queue length. One of the primary causes of the queue backing up was that the single-threaded commitlog writer would stop writing completely while it waited for syscalls and I/O to complete creating a new commitlog file.

This P.R modifies the commitlog to always keep two files open at any time (while only writing to one) and when the commitlog needs to be rotated, the secondary file becomes the primary, and the primary files becomes the secondary and is asynchronously reset in preparation for the next rotation.

Before
Screen Shot 2019-04-18 at 10 09 09 PM

After
Screen Shot 2019-04-18 at 10 09 49 PM


src/dbnode/persist/fs/commitlog/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+When a rotation event occurs, instead of waiting for a new file to be opened, the writer will swap the primary and secondary writers such that the secondary writer (which has an empty file) becomes the primary and vice versa. This allows the writer to continue writing uninterrupted.
+
+In the meantime, a goroutine is started in the background that is responsible for resetting the now secondary (formerly primary) writer by closing it (which will flush any pending / buffered writes to disk) and re-opening it (which will create a new empty commitlog file in anticipation of the next rotation event).
+
+Finally, the next time the commitlog attempts to rotate its commitlogs it will need to use the associated `sync.Waitgroup` to ensure that the previously spawned background goroutine has completed reseting the secondary writer before it attempts a new hot-swap.
```
justinjc: Nits:
s/`sync.Waitgroup`/`sync.WaitGroup`
s/reseting/resetting

Otherwise, clear explanations!

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -442,19 +469,24 @@ func (l *commitLog) write() {
...
-						file: file,
-					},
-				})
+			var file persist.CommitLogFile
+			if err == nil {
+				// The contract for the RotateLogs() API is that it will return the
+				// commitlog file that became the primary one as a direct result of
+				// the method call. The new active file corresponds to index zero and
+				// the new secondary file corresponds to index 1.
+				file = files[0]
```
justinjc: Should this be a `const` at the top and explained there instead of here?

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -506,69 +538,154 @@ func (l *commitLog) write() {
...
+	writer := l.writerState.primaryWriter.writer
+	l.writerState.primaryWriter.writer = nil
 
 	l.closeErr <- writer.Close()
 }
 
-func (l *commitLog) onFlush(err error) {
-	l.flushState.setLastFlushAt(l.nowFn())
+// newOnFlushFn is used to create new flushFns because each one needs to know which
+// slide of pendingFlushFns it should modify.
```
justinjc: s/slide/slice

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -506,69 +538,154 @@ func (l *commitLog) write() {
...
-			l.writerState.writer = nil
+func (l *commitLog) openWriters() (persist.CommitLogFiles, error) {
+	// Ensure that the previous asynchronous reset of the secondary writer (if any)
+	// has completed before attempting to start a new one and/or modify the writerState
+	// in any way.
+	l.waitForSecondaryWriterAsyncResetComplete()
+
+	if l.writerState.primaryWriter.writer == nil || l.writerState.secondaryWriter.writer == nil {
+		// If either of the commitlog writers is nil then open both of them synchronously. Under
+		// normal circumstances this will only occur when the commitlog is first opened.
```
justinjc: Is there a valid scenario where one writer is nil while the other isn't? If not, should you check for it? Do we need to close the writer if on of them isn't nil?

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -506,69 +538,154 @@ func (l *commitLog) write() {
...
+		return l.writerState.activeFiles, nil
 	}
 
-	l.writerState.activeFile = &file
+	// Swap the primary and secondary writers so that the secondary becomes primary and vice versa.
+	// This consumes the standby secondary writer, but a new one will be prepared asynchronously by
+	// resetting the formerly primary writer.
+	prevPrimary := l.writerState.primaryWriter
+	l.writerState.primaryWriter = l.writerState.secondaryWriter
+	l.writerState.secondaryWriter = prevPrimary
```
justinjc: Maybe:
`l.writerState.primaryWriter, l.writerState.secondaryWriter = l.writerState.secondaryWriter, l.writerState.primaryWriter`

src/dbnode/persist/fs/commitlog/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+The commitlog package contains all the code for reading and writing commitlog files.
+
+# Implementation
+
+## Writer
+
+The commitlog writer consists of two main components:
+
+1. The `commitlog` (`commit_log.go`) itself which contains all of the public APIs as well as the logic for managing concurrency, rotating files, and inspecting the status of the commitlog (via the `ActiveLogs` and `QueueLength` APIs).
```
robskillington: nit: Remove the file names since they are much more likely to change.

The types you could use is `CommitLog` for the commit log and `CommitLogWriter` (even though the latter is not exported).

Both these symbols are way more grep-able than the two you're using at the moment (just my 2 cents).

src/dbnode/persist/fs/commitlog/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+The commitlog files are not rotated immediately when the `RotateLogs()` method is called because that would require a lot of complicated and expensive synchronization with the commitlog writer goroutine. Instead, a `rotateLogsEventType` is pushed into the queue and when the single-threaded writer goroutine pulls this event off of the channel it will rotate the commitlog files (since it has exclusive access to them) and then invoke a callback function which notifies the `RotateLogs()` method call (which has been blocked this whole time) to complete and return success to the caller.
+
+While the commitlog only writes to a single file at once, it maintains two open writers at all times so that they can be "hot-swapped" when the commitlog files need to be rotated. This allows the single-threaded writer to continue uninterrupted by syscalls and I/O during rotation events which in turn prevents the queue from backing up. Otherwise, rotation events could block the writer for so long (while it waited for a new file to be created) that it caused the queue to back up significantly.
```
robskillington: Missing ` for this commitlog reference.

src/dbnode/persist/fs/commitlog/README.md
robskillington: Also missing ` for the commitlog reference in this sentence.

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: There is a valid scenario and you're right about the closing if one or the other is not already nil. I'll update it

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: yeah I just broke it up because its an extremely long line

src/dbnode/persist/fs/commitlog/README.md
richardartoul: good feedback, I removed all file names and change all references to be `CommitLog` or `CommitLogWriter`

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: I don't know how I feel about making it a const....if anything I'd maybe want to change the openWriters() API to return two separate arguments or something

src/dbnode/persist/fs/commitlog/README.md
```diff
@@ -0,0 +1,73 @@
+# Overview
```
Haijuncao: Nice! thanks for adding the readme. It helps a lot in reading your pr.

src/dbnode/persist/fs/commitlog/commit_log_conc_test.go
```diff
@@ -109,7 +109,7 @@ func TestCommitLogActiveLogsConcurrency(t *testing.T) {
 			if err != nil {
 				panic(err)
 			}
-			require.Equal(t, 1, len(logs))
+			require.Equal(t, 2, len(logs))
```
Haijuncao: code change looks good, but it seems that there are no new tests being added, other than asserting the number of logs/files is 2 (or incrementing by 2). 

Do you think there are any new tests can be added?

richardartoul: I couldn't really think of any. The existing tests are already pretty good in that they have a good mix of unit tests, prop tests, and concurrency tests. The change I'm making here is more of an implementation detail than anything so if everything is still passing I think we're in good shape

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -251,20 +283,19 @@ func (l *commitLog) Open() error {
...
 		return err
 	}
 
 	// Sync the info header to ensure we can write to disk and make sure that we can at least
 	// read the info about the commitlog file later.
-	if err := l.writerState.writer.Flush(true); err != nil {
+	if err := l.writerState.primaryWriter.writer.Flush(true); err != nil {
+		return err
+	}
+	if err := l.writerState.secondaryWriter.writer.Flush(true); err != nil {
```
robskillington: nit: Perhaps you could put them both in an array so we don't forget to do any operation to just one commit log?

i.e.

```go
type writerState struct {
  primaryWriter asyncResettableWriter
  secondaryWriter asyncResettableWriter
  writers []commitLogWriter
  activeFiles persist.CommitLogFiles
}


// then

for _, writer := range l.writerState.writers {
  if err := writer.Flush(true); err != nil {
    return err
  }
}
```

src/dbnode/persist/fs/commitlog/commit_log.go
robskillington: Why don't you make `openWriters` instead just return both writers as args?

i.e.
```go
primaryFile, secondaryFile, err := l.openWriters()

// then you don't need the second one most of the time
// i.e.
primaryFile, _, err := l.openWriters()
```

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -506,69 +541,169 @@ func (l *commitLog) write() {
...
+	// encountered an error trying to re-open it.
+	if l.writerState.secondaryWriter.writer != nil {
+		// Don't care about errors closing the secondary writer because it doesn't
+		// have any data.
+		l.writerState.secondaryWriter.writer.Close()
+		l.writerState.secondaryWriter.writer = nil
+	}
+
+	writer := l.writerState.primaryWriter.writer
+	l.writerState.primaryWriter.writer = nil
```
robskillington: Instead of the above, if you have the array of writers it gets a little eaiser:

```go
for _, writer := l.writerState.writers {
  multiErr = multiErr.Add(writer.Close())
}

// Zero our the writers
l.writerState = writerState{}

// Report close error
l.closeErr <- multiErr.FinalError()
```

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -506,69 +541,169 @@ func (l *commitLog) write() {
...
 
 	l.closeErr <- writer.Close()
 }
 
-func (l *commitLog) onFlush(err error) {
-	l.flushState.setLastFlushAt(l.nowFn())
+// newOnFlushFn is used to create new flushFns because each one needs to know which
+// slice of pendingFlushFns it should modify.
+func (l *commitLog) newOnFlushFn(writer *asyncResettableWriter) flushFn {
```
robskillington: Hm, why don't you add the `onFlush(err error)` method to `asyncResettableWriter` and then make it call the commit log back.

i.e.

```go
type asyncResettableWriter struct {
  // .. other
  commitLog *commitLog
}

func (w *asyncResettableWriter) onFlush(err error) {
  w.commitLog.onFlush(w, err)
}

// now on commit log, retain how this used to work
func (l *commitLog) onFlush(writer *asyncResettableWriter, err error) {
  // as before
}
```

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -124,8 +124,32 @@ func (f *flushState) getLastFlushAt() time.Time {
 }
 
 type writerState struct {
-	writer     commitLogWriter
-	activeFile *persist.CommitLogFile
+	// See "Rotating Files" section of README.md for an explanation of how the
+	// primaryWriter and secondaryWriter fields are used during commitlog rotation.
+	primaryWriter   asyncResettableWriter
+	secondaryWriter asyncResettableWriter
```
robskillington: A lot of the code is really long below, maybe just call this `primary` and `secondary`?

Then calls will be like:
```go
l.writerState.primary.writer.Whatever(..)
```

src/dbnode/persist/fs/commitlog/commit_log.go
robskillington: `l.writerState.primary, l.writerState.secondary = l.writerState.secondary, l.writerState.primary` is a little shorter if you trim `primaryWriter` -> `primary` and same with secondary.

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -506,69 +541,169 @@ func (l *commitLog) write() {
...
+	prevPrimary := l.writerState.primaryWriter
+	l.writerState.primaryWriter = l.writerState.secondaryWriter
+	l.writerState.secondaryWriter = prevPrimary
+	l.startSecondaryWriterAsyncReset()
 
-	return file, nil
+	var (
+		// Determine the persist.CommitLogFile for the not-yet-created secondary file so that the
+		// ActiveLogs() API returns the correct values even before the asynchronous reset completes.
+		primaryFile   = l.writerState.activeFiles[1]
```
robskillington: Does it make slightly more sense to access the "last index" than just 1?  i.e. 

```go
primaryFile = l.writerState.activeFiles[len(l.writerState.activeFiles)-1]
```

I know it's longer but seems more easy to read (strangely) since I know it's always getting the last entry.

Not too fussed though.

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: Yeah I thought about this and didn't do it because i really hate cyclic datastructures as they can be really confusing. Probably simpler to do it in this case though, I'll make the change

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: I prefer the existing one because there can only ever be two and the "last" logic feel more confusing to me because it implies there is may be more than two

src/dbnode/persist/fs/commitlog/commit_log.go
```diff
@@ -506,13 +548,28 @@ func (l *commitLog) write() {
...
+	// Ensure that there is no active background goroutine in the middle of reseting
+	// the secondary writer / modifying its state.
+	l.waitForSecondaryWriterAsyncResetComplete()
 
-	l.closeErr <- writer.Close()
+	var multiErr xerrors.MultiError
+	for i, writer := range l.writerState.writers {
+		if writer == nil {
+			// Can be nil in the case where the background goroutine spawned in openWriters
+			// encountered an error trying to re-open it.
```
robskillington: nit: If you really want to you could create a `forEachWriters(func(writer) { ... })` helper function to make sure all for loops does this.

src/dbnode/persist/fs/commitlog/commit_log.go
robskillington: Sure thing.

src/dbnode/persist/fs/commitlog/commit_log.go
richardartoul: Yeah that would be a little cleaner, but I'd rather be explicit and have it crash in places where we don't expect to ever be nil than have it silently do something weird.

richardartoul:
@robskillington Sure thing, I'll move the monopic file

PR #1578, opened 2019-04-21
[dbnode] set schema via namespace context

Haijuncao:
1. The detail design section highlights the changes in this pr, please take a look first, I may consider adding a readme later to document it within the repo:
https://docs.google.com/document/d/16eRtRyP9niMHyCf1vqtpFsavj568MqtcBWt4nLTLi2o/edit#

2. To help with review, here is a few highlights:
* the main mechanism of propagating schema to proto encoder/iterator is via call stack. Schema is accquired at the entry point of namespace top level API call and pass down the stack.
* Schema is added as an input argument to Encoder/ReaderIterator's Reset method. It is also added as input argument of ReaderIteratorAlloc function.
* add a SetSchema method to change schema mid stream for proto encoder (without reset).
* storage/namespace/schema_registry.go: this file contains the singleton in memory object that receives dynamic schema updates, updates are pushed to dbNamespace (using watchable/listener pattern).
* storage/namespace/context.go: this files contains helper methods that construct a namespace context from two different sources: namespace metadata (for bootstrapping) and database schema registry (for client read/write).
* storage/namespace/types.go: renamed the original namespace SchemaRegistry to SchemaHistory
* storage/namespace.go: removed SchemaHistory from dbNamespace object, dbNamespace registers a listener to schema registry that receives latest namespace.SchemaDescr.
* server/server.go: initialized the global schema registry and passed to database options.
* storage/block/block.go:  DatabaseBlock gain access to schema via Options.NamespaceContext, block's namespace context is generally set by bootstrappers during bootstrapping.

3. Tests: the approach is modify existing test to be runnable both for m3tsz and proto
* Unit tests: buffer (storage/series), block (storage/block), session (dbnode/client)
* Integration test (dbnode/integration): roundtrip, adminclient, bootstrappers


src/dbnode/client/options.go
```diff
@@ -32,16 +32,15 @@ import (
 	"github.com/m3db/m3/src/dbnode/encoding/m3tsz"
 	"github.com/m3db/m3/src/dbnode/encoding/proto"
 	m3dbruntime "github.com/m3db/m3/src/dbnode/runtime"
+	"github.com/m3db/m3/src/dbnode/storage/namespace"
```
prateek: Hm it feels like time to move this package out from `storage`. What do you reckon?

src/dbnode/client/session.go
```diff
@@ -3100,6 +3117,7 @@ type blocksResult interface {
 }
 
 type baseBlocksResult struct {
+	nCtx                    namespace.Context
```
prateek: Super nit: rename to `nsCtx`

src/dbnode/client/session.go
```diff
@@ -3232,20 +3256,32 @@ type streamBlocksResult struct {
...
 	return &streamBlocksResult{
-		baseBlocksResult: newBaseBlocksResult(opts, resultOpts),
+		baseBlocksResult: newBaseBlocksResult(nCtx, opts, resultOpts),
 		outputCh:         outputCh,
 		tagDecoder:       tagDecoder,
 		idPool:           idPool,
 	}
 }
 
+func getContextFor(id ident.ID, opts Options) namespace.Context {
```
prateek: nit: try to avoid using ‘get’ in naming methods. `newNamepaceContext` is more idiomatic

src/dbnode/client/session.go
```diff
@@ -3232,20 +3256,32 @@ type streamBlocksResult struct {
...
+	schemaReg := opts.SchemaRegistry()
+	if id == nil || schemaReg == nil {
```
prateek: Why not error here?

src/dbnode/client/session.go
```diff
@@ -3232,20 +3256,32 @@ type streamBlocksResult struct {
...
+		return namespace.Context{}
+	}
+	schema, _ := schemaReg.GetLatestSchema(id)
```
prateek: Need to propagate this error up to the caller

src/dbnode/client/session.go
```diff
@@ -3331,14 +3367,15 @@ type bulkBlocksResult struct {
...
 	opts Options,
 	resultOpts result.Options,
 	tagDecoderPool serialize.TagDecoderPool,
 	idPool ident.Pool,
 ) *bulkBlocksResult {
 	return &bulkBlocksResult{
-		baseBlocksResult: newBaseBlocksResult(opts, resultOpts),
-		result:           result.NewShardResult(4096, resultOpts),
+		baseBlocksResult: newBaseBlocksResult(nCtx, opts, resultOpts),
+		result:           result.NewShardResult(nCtx,4096, resultOpts),
```
prateek: Unrelated to your change but mind making 4096 a const

src/dbnode/client/types.go
```diff
@@ -500,6 +499,10 @@ type Options interface {
 
 	// ReaderIteratorAllocate returns the readerIteratorAllocate.
 	ReaderIteratorAllocate() encoding.ReaderIteratorAllocate
+
+	SetSchemaRegistry(registry namespace.SchemaRegistry) AdminOptions
```
prateek: Nit: need comments on these methods

src/dbnode/encoding/proto/round_trip_test.go
```diff
@@ -145,7 +146,8 @@ func TestRoundTrip(t *testing.T) {
 	require.NoError(t, err)
 
 	buff := bytes.NewBuffer(rawBytes)
-	iter := NewIterator(buff, testVLSchema, testEncodingOptions)
+	iter := NewIterator(buff, testEncodingOptions)
```
prateek: Unrelated to your change, but while you’re here, mind moving the test proto files in this package to `src/dbnode/encoding/proto/testdata`

src/dbnode/encoding/series_iterators.go
```diff
@@ -81,3 +89,4 @@ type emptyIters struct{}
 func (e emptyIters) Iters() []SeriesIterator { return nil }
 func (e emptyIters) Len() int                { return 0 }
 func (e emptyIters) Close()                  {}
+func (e emptyIters) SetSchema(descr namespace.SchemaDescr) {}
```
prateek: nit: newline

src/dbnode/encoding/types.go
```diff
@@ -64,6 +65,8 @@ type Encoder interface {
 
 	// DiscardReset will take ownership of the encoder data and reset the encoder for use.
 	DiscardReset(t time.Time, capacity int) ts.Segment
+
+	SetSchema(descr namespace.SchemaDescr)
```
prateek: Needs a comment

src/dbnode/encoding/types.go
```diff
@@ -139,6 +142,8 @@ type Iterator interface {
 
 	// Close closes the iterator and if pooled will return to the pool.
 	Close()
+
+	SetSchema(descr namespace.SchemaDescr)
```
prateek: Here too

src/dbnode/encoding/types.go
```diff
@@ -223,6 +228,8 @@ type SeriesIterators interface {
 
 	// Close closes all iterators contained
 	Close()
+
+	SetSchema(descr namespace.SchemaDescr)
```
prateek: Here too

src/dbnode/encoding/types.go
prateek: Ordering nit: I like to order the functions in interfaces according to lifecycle/order of use of at all possible. E.g start with Open, end with Close, etc. mind reordering took

src/dbnode/storage/block/block.go
```diff
@@ -44,6 +45,7 @@ var (
 type dbBlock struct {
 	sync.RWMutex
 
+	nCtx           namespace.Context
```
prateek: nit: rename to `nsCtx`

src/dbnode/client/session.go
Haijuncao: This is to make it backward compatible for existing code (m3tsz) and test code where schema registry is not populated and/or namespace id is set in test setup. 

I haven't written any new test yet (to test this feature). For existing test, I am still thinking about how we should proceed with proto encoding/schema. One way could be to run another test target that have proto encoder/schema enabled by default, this way, we can catch any schema issue earlier.

prateek: Can we make this explicit - separate function without the args or an enum/sentinel to say something like what mode we’re in? Trying to avoid having nil mean anything 

Would be better to rework the tests to do the right thing instead of changing production code to make it easier to accommodate tests.

src/dbnode/client/options.go
robskillington: Fair, yeah I agree.

Maybe we should follow up with a single atomic commit we can do this with (otherwise it might make for painful rebases while this PR is alive)?

src/dbnode/client/session.go
```diff
@@ -1988,7 +1992,8 @@ func (s *session) FetchBlocksFromPeers(
 		complete = int64(0)
 		doneCh   = make(chan error, 1)
 		outputCh = make(chan peerBlocksDatapoint, 4096)
-		result   = newStreamBlocksResult(s.opts, opts, outputCh,
+		nsCtx    = namespace.NewContextFor(nsMetadata.ID(), s.opts.SchemaRegistry())
```
robskillington: Could this not also be `namespace.NewContextFrom(nsMetadata)`?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source_data_test.go
```diff
@@ -448,9 +502,9 @@ func verifyShardResultsAreCorrect(
 }
 
 func createExpectedShardResult(
+	nsCtx namespace.Context,
 	values []testValue,
 	blockSize time.Duration,
-	actual result.ShardResults,
```
Haijuncao: actual is not really used, so remove it while I am at it.

src/dbnode/client/options.go
Haijuncao: I agree. I will do another PR just to move namespace out of storage aftr this PR.

src/dbnode/client/session.go
Haijuncao: I removed the check.

src/dbnode/client/session.go
Haijuncao: yes

src/dbnode/client/options.go
robskillington: @Haijuncao can you open an issue to track this?  We should make sure we remember to do this relatively immediately.

Haijuncao: https://github.com/m3db/m3/issues/1599

I proposed an alternative in the issue. please comment. thanks.

src/dbnode/client/session.go
Haijuncao: considering this is outdated and the new place no longer check for nil. The caller should make sure it is not nil, o/w it is a program error we should panic/npe asap. 

I will resolve this. @prateek, take a look at namespace/context.go for the new function: NewContextFor

src/dbnode/storage/namespace/context.go
```diff
@@ -0,0 +1,46 @@
...
+// THE SOFTWARE.
+
+package namespace
+
+import (
+	"github.com/m3db/m3/src/x/ident"
+)
+
+type Context struct {
+	Id     ident.ID
```
robskillington: Can you make this `ID` instead of `Id`?  It's a Go thing:
https://github.com/golang/lint/issues/124

src/dbnode/client/session.go
```diff
@@ -3423,6 +3437,7 @@ func (r *bulkBlocksResult) addBlockFromPeer(
 
 		result = r.blockOpts.DatabaseBlockPool().Get()
 		result.Reset(start, blockSize, encoder.Discard())
+		result.SetNamespaceContext(r.nsCtx)
```
robskillington: Hm, is it worth just making `nsCtx` an arg to `Reset(...)` if it's always required?

src/dbnode/client/session.go
```diff
@@ -3143,10 +3151,12 @@ func (b *baseBlocksResult) segmentForBlock(seg *rpc.Segment) ts.Segment {
 
 func (b *baseBlocksResult) mergeReaders(start time.Time, blockSize time.Duration, readers []xio.SegmentReader) (encoding.Encoder, error) {
 	iter := b.multiReaderIteratorPool.Get()
+	iter.SetSchema(b.nsCtx.Schema)
```
robskillington: I'd be in favor of making `schema` be required by call to `Reset(...)` rather than being a separate method since it's required every time you start using the iterator anyhow.

src/dbnode/client/session.go
```diff
@@ -3143,10 +3151,12 @@ func (b *baseBlocksResult) segmentForBlock(seg *rpc.Segment) ts.Segment {
...
 	iter.Reset(readers, start, blockSize)
 	defer iter.Close()
 
 	encoder := b.encoderPool.Get()
+	encoder.SetSchema(b.nsCtx.Schema)
```
robskillington: Similarly I'd be in favor of making `schema` be required by call to encoder `Reset(...)` rather than being a separate method since it's required every time you start using the iterator anyhow.

src/dbnode/client/session.go
```diff
@@ -1557,6 +1559,7 @@ func (s *session) fetchIDsAttempt(
 				slicesIter := s.pools.readerSliceOfSlicesIterator.Get()
 				slicesIter.Reset(result.([]*rpc.Segments))
 				multiIter := s.pools.multiReaderIterator.Get()
+				multiIter.SetSchema(nsCtx.Schema)
```
robskillington: Passing `schema` could be moved to an arg of `Reset(...)` here too.

src/dbnode/encoding/iterator_pool.go
```diff
@@ -65,5 +68,6 @@ func (p *multiReaderIteratorPool) Get() MultiReaderIterator {
 }
 
 func (p *multiReaderIteratorPool) Put(iter MultiReaderIterator) {
+	iter.SetSchema(nil)
```
robskillington: This should just be handled by `Reset(...)` then you don't need to add extra logic to the pool gets/puts.

Making pools know more about their data structures is a pattern we've tried to avoid, encapsulating that logic in `Reset(...)` for each separate data structure helps make these pools simple.

src/dbnode/encoding/encoder_pool.go
```diff
@@ -44,5 +44,6 @@ func (p *encoderPool) Get() Encoder {
 }
 
 func (p *encoderPool) Put(encoder Encoder) {
+	encoder.SetSchema(nil)
```
robskillington: Similar comment here:

This should just be handled by `Reset(...)` then you don't need to add extra logic to the pool gets/puts.

Making pools know more about their data structures is a pattern we've tried to avoid, encapsulating that logic in `Reset(...)` for each separate data structure helps make these pools simple.

src/dbnode/encoding/iterator_test.go
```diff
@@ -86,6 +87,8 @@ func (it *testIterator) Reset(r io.Reader) {
 	it.onReset(r)
 }
 
+func (it *testIterator) SetSchema(descr namespace.SchemaDescr) {}
```
robskillington: If schema was passed in Reset then you can just ignore the arg, rather than needing to add no-op methods to these types.

src/dbnode/storage/block/block.go
```diff
@@ -96,6 +98,10 @@ func NewDatabaseBlock(
 	return b
 }
 
+func (b *dbBlock) SetNamespaceContext(nsCtx namespace.Context) {
```
robskillington: Can we make this part of the `New(..)` and the `Reset(...)` calls to?

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source_data_test.go
```diff
@@ -384,10 +433,13 @@ func TestItMergesSnapshotsAndCommitLogs(t *testing.T) {
...
 	expectedValues = append(expectedValues, commitLogValues[0:3]...)
 
-	require.NoError(t, verifyShardResultsAreCorrect(
-		expectedValues, blockSize, res.ShardResults(), opts))
+	require.NoError(t, verifyShardResultsAreCorrect(nsCtx,
+		expectedValues, blockSize, res.ShardResults(), opts, annEqual))
 }
 
+type setAnnotation func([]testValue) []testValue
+type annotationEqual func([]byte, []byte) bool
```
robskillington: Hm this is definitely a little strange.  Can't we infer whether the namespace is using proto or not and just use the correctly annotation equal fn?

src/dbnode/storage/database.go
```diff
@@ -223,12 +224,39 @@ func NewDatabase(
 	return d, nil
 }
 
+func (d *db) UpdateSchemaRegistry(newNamespaces namespace.Map) {
+	schemaReg := d.opts.SchemaRegistry()
+	schemaUpdates := newNamespaces.Metadatas()
+	for _, metadata := range schemaUpdates {
+		curSchemaId := "none"
```
robskillington: nit: Go naming suggests this should be `curSchemaID`:
https://github.com/golang/lint/issues/124

src/dbnode/storage/namespace.go
```diff
@@ -309,6 +306,7 @@ func newDatabaseNamespace(
 	seriesOpts := NewSeriesOptionsFromOptions(opts, nopts.RetentionOptions()).
 		SetStats(series.NewStats(scope)).
 		SetColdWritesEnabled(nopts.ColdWritesEnabled())
+	seriesOpts = seriesOpts.SetNamespaceId(metadata.ID())
```
robskillington: This should be `SetNamespaceID(...)` as per Go naming semantics:
https://github.com/golang/lint/issues/124

src/dbnode/storage/namespace.go
```diff
@@ -592,7 +571,7 @@ func (n *dbNamespace) Write(
 		return ts.Series{}, false, err
 	}
 	series, wasWritten, err := shard.Write(ctx, id, timestamp,
-		value, unit, annotation, series.WriteOptions{})
+		value, unit, annotation)
```
robskillington: Hm wait, why are we removing WriteOptions?

By the way, just so you know WriteOptions is now going to be used by other code that needs to propagate things to the series/buffer:
https://github.com/m3db/m3/pull/1596

src/dbnode/storage/namespace.go
robskillington: Instead of setting the namespace ID here and looking up schema later, why not just cache the actual schema local to the namespace and use it while you hold the namespace.RLock when a series write comes in and then setting it on the `WriteOptions`?

src/dbnode/storage/namespace/schema_registry.go
```diff
@@ -0,0 +1,84 @@
...
+	defer sr.RUnlock()
+
+	history, ok := sr.registry[id.String()]
+	if !ok {
+		return nil, false
+	}
+	return history.GetLatest()
+}
+
+func (sr *schemaRegistry) GetSchema(id ident.ID, schemaId string) (SchemaDescr, bool) {
```
robskillington: nit: should be `schemaID string`.

src/dbnode/storage/namespace/schema_registry.go
```diff
@@ -0,0 +1,84 @@
...
+
+	sr.registry[id.String()] = history
+	return nil
+}
+
+func (sr *schemaRegistry) GetLatestSchema(id ident.ID) (SchemaDescr, bool) {
+	sr.RLock()
+	defer sr.RUnlock()
+
+	history, ok := sr.registry[id.String()]
```
robskillington: How often is this call made?  If it is high frequency then the map will need to be a code generated map (using genny and the generated-sources make files) and so `ident.ID` can be used as the key.

If it's low frequency then it doesn't matter as much.

src/dbnode/storage/namespace/context.go
```diff
@@ -0,0 +1,46 @@
...
+	ctx := Context{Id: nsMetadata.ID()}
+	if schema, ok := nsMetadata.Options().SchemaHistory().GetLatest(); ok {
+		ctx.Schema = schema
+	}
+	return ctx
+}
+
+func NewContextFor(id ident.ID, registry SchemaRegistry) Context {
+	ctx := Context{Id: id}
+	if schema, ok := registry.GetLatestSchema(id); ok {
```
robskillington: Yeah this is going to be a very high frequency call, can we cache the schema on the namespace and pass that along in `WriteOptions` please?

Let's discuss this further if you have differing opinions, but adding a second RLock here and a map lookup and a string conversion to each call to `Write(..)` for the hot path is quite a significant addition and likely quite hard to undo later.

src/dbnode/client/session.go
Haijuncao: agree.

src/dbnode/client/session.go
Haijuncao: agree. this makes sense.

src/dbnode/client/session.go
Haijuncao: encoder is slightly different. we could call SetSchema in the middle.
Still, I would make schema part of reset, but keep a separate SetSchema for encoder to allow mid stream schema change.

src/dbnode/client/session.go
Haijuncao: ditto

src/dbnode/encoding/iterator_pool.go
Haijuncao: ditto

src/dbnode/encoding/encoder_pool.go
Haijuncao: ditto

src/dbnode/encoding/iterator_test.go
Haijuncao: fair

src/dbnode/storage/block/block.go
Haijuncao: I can merge this to Reset, as the block is taken out of the pool. (similar to SetSchema into Reset).

src/dbnode/storage/bootstrap/bootstrapper/commitlog/source_data_test.go
Haijuncao: we could, though it might be cleaner if we don't infer. I image we could infer by whether or not schema is nil in nsCtx, though if nsCtx is not set correctly, the test would have succeed anyway even not using the right comparator. so not infer, always explicit seem to be better.

src/dbnode/storage/database.go
Haijuncao: ditto

src/dbnode/storage/namespace.go
Haijuncao: ok, I will add it back. I thought this is only for schema passing.

For schema, since WriteOpts can only be used for write, not read, while nsCtx can be used for both, I have decided not to use WriteOpts to propagate down (granted, I could have something called ReadOpts, but the code path for read is more complicated than write).

It appears using series.Options to pass SchemaRegistry to series/buffer/bucket level at Reset time  is a lot less code and equally effective/explicit.

src/dbnode/storage/namespace/schema_registry.go
Haijuncao: ditto

src/dbnode/storage/namespace/schema_registry.go
Haijuncao: pretty high frequency, each write. will use generated map, always curious what it is for. 

Is there a readme that provides more context/instruction? will give it a try tomorrow.

src/dbnode/storage/namespace/context.go
Haijuncao: I am all for avoiding additional lock for hot path.

how long do we have to cache schema for? is it still for the duration of a single write?

src/dbnode/storage/namespace/schema_registry.go
richardartoul: @Haijuncao Is this still gonna be per-write or has that issue been resolved with the listener pattern?

src/dbnode/storage/namespace/context.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+type Context struct {
+	Id     ident.ID
+	Schema SchemaDescr
+}
+
+func NewContextFrom(nsMetadata Metadata) Context {
+	ctx := Context{Id: nsMetadata.ID()}
+	if schema, ok := nsMetadata.Options().SchemaHistory().GetLatest(); ok {
```
richardartoul: Does this mean that during bootstrap all encoders/iterators will used a "stale" schema if the schema changes AFTER the bootstrap starts? I think thats probably fine but I'm just trying to make sure I understand properly.

richardartoul:
@Haijuncao I think a markdown document (maybe a package-level README somewhere?) committed to the repo with a high level explanation of how schema propagation works would be really useful. Basically the description you put for this P.R but with a little bit more details and information about caveats etc (like whether or not schemas get propagated dynamically to the bootstrappers and that schema updates take effect almost immediately on the next write in the non-bootstrapping case etc etc)

src/dbnode/storage/namespace/context.go
Haijuncao: yes, your understanding is correct.

Haijuncao:
> @Haijuncao I think a markdown document (maybe a package-level README somewhere?) committed to the repo with a high level explanation of how schema propagation works would be really useful. Basically the description you put for this P.R but with a little bit more details and information about caveats etc (like whether or not schemas get propagated dynamically to the bootstrappers and that schema updates take effect almost immediately on the next write in the non-bootstrapping case etc etc)

I agree. I will probably do that in another PR as this PR has evolved significantly. Once PR approves and lands, I'd be in better position to summarize it.

src/dbnode/server/server.go
```diff
@@ -1176,25 +1169,24 @@ func withEncodingAndPoolingOptions(
 		SetSegmentReaderPool(segmentReaderPool)
 
 	encoderPool.Init(func() encoding.Encoder {
-		if schema != nil {
+		if cfg.Proto != nil {
```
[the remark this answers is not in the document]
Haijuncao: I think cfg.Proto is going to be set, I haven't remove it from server cfg. But let me modify server cfg to take a Enabled bool instead

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -81,6 +82,8 @@ func NewEncoder(
 	}
 }
 
+func (enc *encoder) SetSchema(descr namespace.SchemaDescr) {}
```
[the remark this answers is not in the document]
Haijuncao: It will get called on m3tsz encoder. The caller (in this case, buffer bucket) does not know what kind of encoder it is, it always call SetSchema right before Encode to support mid stream schema change.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -123,7 +125,7 @@ func (enc *Encoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, protoBytes ts.A
 	}
 
 	if enc.schema == nil {
-		return errEncoderSchemaIsRequired
+		panic(errEncoderSchemaIsRequired.Error())
```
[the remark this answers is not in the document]
Haijuncao: In this case, we really need it. I will add a comment.

It is a program error instead of a recoverable request error if schema is not supplied/set before encode is called. It means we missed some code path (to set schema).

src/dbnode/encoding/proto/iterator.go
```diff
@@ -75,29 +79,23 @@ type iterator struct {
...
 		tsIterator: m3tsz.NewTimestampIterator(opts, true),
 	}
+	i.resetSchema(descr)
+	return i
 }
 
 func (it *iterator) Next() bool {
 	if it.schema == nil {
-		it.err = errIteratorSchemaIsRequired
+		panic(errIteratorSchemaIsRequired.Error())
```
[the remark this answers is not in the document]
Haijuncao: same reply

src/dbnode/encoding/types.go
```diff
@@ -24,18 +24,24 @@ import (
...
 	"github.com/m3db/m3/src/x/ident"
 	"github.com/m3db/m3/src/x/pool"
+	"github.com/m3db/m3/src/x/serialize"
 	xtime "github.com/m3db/m3/src/x/time"
 )
 
 // Encoder is the generic interface for different types of encoders.
 type Encoder interface {
+	// SetSchema sets up the schema needed by schema-aware encoder to encode the stream.
+	// Must be called before Encode
```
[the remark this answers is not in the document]
Haijuncao: all right, I will refine it.

src/dbnode/storage/namespace/schema_registry.go
Haijuncao: yes, it is no longer high frequency call. I will add a TODO comment here to do a generated map for this registry in another PR.

src/dbnode/encoding/proto/encoder.go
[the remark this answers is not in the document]
richardartoul: As long as the error is propagated properly it should come up fairly quickly if we missed something, but I guess its fair for now while this is all experimental, can always remove it later when we have more confidence

src/dbnode/storage/namespace.go
```diff
@@ -115,13 +116,15 @@ type dbNamespace struct {
 	log                *zap.Logger
 	bootstrapState     BootstrapState
 
+	// schemaDescr caches the latest schema for the namespace.
+	// schemaDescr is updated whenver schema registry is updated.
```
richardartoul: `whenever`

src/dbnode/storage/namespace.go
```diff
@@ -348,12 +350,25 @@ func newDatabaseNamespace(
 		metrics:                newDatabaseNamespaceMetrics(scope, iops.MetricsSamplingRate()),
 	}
 
+	sl, ok := opts.SchemaRegistry().RegisterListener(id, n)
+	if !ok {
+		n.log.Warn("unable to registry schema listener",
```
richardartoul: `register`

richardartoul: In what situation does this return `!ok`? Should it be a hard error? Curious if this is a network error or some other type of error thing

src/dbnode/storage/namespace.go
```diff
@@ -501,6 +505,10 @@ func (n *dbNamespace) Tick(c context.Cancellable, tickStart time.Time) error {
 		return nil
 	}
 
+	n.RLock()
+	nsCtx := namespace.Context{Schema: n.schemaDescr}
```
richardartoul: In some places you create this with an ID and in some places you don't. Is that intentional? Should we have a constructor function or cache an up-to-date version next to schemaDescr?

src/dbnode/storage/namespace.go
richardartoul: Ok I see that it can only return false if the namespace doesn't exist in the registry. That seems like it would be some kind of code bug and maybe we should return an error here instead of emitting a warning

src/dbnode/storage/namespace/schema_registry.go
```diff
@@ -0,0 +1,136 @@
...
+func (sr *schemaRegistry) RegisterListener(
+	nsID ident.ID,
+	listener SchemaListener,
+) (xclose.SimpleCloser, bool) {
+	sr.RLock()
+	defer sr.RUnlock()
+
+	watchable, ok := sr.registry[nsID.String()]
+	if !ok {
+		return nil, false
```
richardartoul: Based on my previous comment I wonder if this should be an error as opposed to a boolean false

src/dbnode/storage/namespace.go
Haijuncao: good catch, I will try to make it consistent (always set a nsID). Though in many places, only schema is ever used.

src/dbnode/storage/namespace/schema_registry.go
Haijuncao: Agree, I modified schema registry to return error.

src/dbnode/storage/namespace.go
Haijuncao: it is changed to return error.

src/dbnode/storage/database.go
```diff
@@ -223,12 +225,38 @@ func NewDatabase(
...
+		curSchema, found := schemaReg.GetLatestSchema(metadata.ID())
+		if found {
+			curSchemaId = curSchema.DeployId()
+		}
+		// Log schema update.
+		latestSchema, found := metadata.Options().SchemaHistory().GetLatest()
+		if !found {
+			d.log.Info("updating namespace schema to empty", zap.Stringer("namespace", metadata.ID()))
```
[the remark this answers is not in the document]
Haijuncao: Great question. This part of logic is not well defined. 

I made some modification with clear behavior:
1. m3tsz workload (implied proto is disabled): schema is not required in namespace metadata, if it is set by mistake, schema will simply be ignored
2. proto workflow (implies proto is enabled): 
schema is required in namespace metadata, if it is absent, namespace will not be initialized, in multi-namespace database, namespace initialization is isolated from each other (i.e. one miss-configured namespace won't prevent others from initializing or updating).

src/dbnode/encoding/proto/encoder.go
```diff
@@ -348,15 +351,26 @@ func (enc *Encoder) encodeProto(m *dynamic.Message) error {
...
-	enc.resetSchema(schema)
+func (enc *Encoder) SetSchema(descr namespace.SchemaDescr) {
+	if descr == nil {
+		enc.schemaDesc = nil
+		enc.resetSchema(nil)
+		return
+	}
+
+	// Noop if schema has not changed.
+	if enc.schemaDesc != nil && len(descr.DeployId()) != 0 && enc.schemaDesc.DeployId() == descr.DeployId() {
```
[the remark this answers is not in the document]
Haijuncao: no, it just returns a string, not sure how expensive string comparison is in golang.

src/dbnode/client/session_fetch_test.go
```diff
@@ -499,11 +539,19 @@ func fulfillTszFetchBatchOps(
 	}
 }
 
+func bytesRef(data checked.Bytes) []byte {
```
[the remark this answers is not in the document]
Haijuncao: sure

src/dbnode/encoding/proto/encoder.go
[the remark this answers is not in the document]
Haijuncao: agree, I wanted to flush all the bugs regarding schema not set asap, hence panic. We will remove it (add an error) later after we gain confidence.

src/cmd/services/m3dbnode/config/config.go
```diff
@@ -323,16 +323,26 @@ type HashingConfiguration struct {
 
 // ProtoConfiguration is the configuration for running with ProtoDataMode enabled.
 type ProtoConfiguration struct {
+	// Enabled or disabled.
+	Enabled bool `yaml:"enabled"`
+	// TODO [haijun] remove after PR to set schema in etcd is done (plan layed out in issue #1614).
```
robskillington: Left a comment on #1614, lets continue discussion there.

src/dbnode/client/config.go
```diff
@@ -87,8 +87,8 @@ type Configuration struct {
 
 // ProtoConfiguration is the configuration for running with ProtoDataMode enabled.
 type ProtoConfiguration struct {
-	SchemaFilePath string `yaml:"schemaFilePath"`
-	MessageName    string `yaml:"messageName"`
+	// Whether proto is enabled.
```
robskillington: nit: Field name comments should start with the name of the field, i.e. `// Enabled specifies whether proto is enabled.`

src/dbnode/client/session_proto_test.go
```diff
@@ -0,0 +1,151 @@
...
+
+	"github.com/m3db/m3/src/dbnode/ts"
+	"github.com/m3db/m3/src/dbnode/generated/thrift/rpc"
+	"github.com/m3db/m3/src/dbnode/storage/namespace"
+)
+
+var (
+	testNamespace     = ident.StringID("session_prototest_ns")
+	// Relative path to $pwd to reach testdata/prototest, $pwd is current directory.
+	testSchemaHistory = prototest.NewSchemaHistory("../")
```
robskillington: Hm, if it's relative shouldn't it be `prototest.NewSchemaHistory("../testdata/prototest")`?  

This is a little bit of a strange contract for the `prototest.NewSchemaHistory(...)` function.

Could you not perhaps just inline the protobuf messages in prototest package rather than have it read a file?  Then it would avoid all this strange path stuff.

You could do in the `prototest` package just some declarations like so:
```go
const testProto = `
message My Proto {
  ...
}
`
```

Haijuncao: the protoparse package we use do not take string.

src/dbnode/encoding/types.go
```diff
@@ -24,19 +24,26 @@ import (
...
 	"github.com/m3db/m3/src/x/pool"
+	"github.com/m3db/m3/src/x/serialize"
 	xtime "github.com/m3db/m3/src/x/time"
 )
 
 // Encoder is the generic interface for different types of encoders.
 type Encoder interface {
+	// SetSchema sets up the schema needed by schema-aware encoder to encode the stream.
+	// SetSchema can be called multiple times between reset for mid-stream schema changes.
+	SetSchema(descr namespace.SchemaDescr)
```
robskillington: nit: Do you think people need to SetSchema(...) in between Reset(...) calls?  

If not is it worth removing if there's no legitimate real use case?  Otherwise keeping it I suppose makes sense.

src/dbnode/testdata/prototest/fixture.go
```diff
@@ -0,0 +1,230 @@
...
+	longitude  float64
+	epoch      int64
+	deliveryID []byte
+	attributes map[string]string
+}
+
+func NewSchemaHistory(importPath string) namespace.SchemaHistory {
+	srcRoot := os.Getenv("SRC_ROOT")
+	curDir := os.Getenv("PWD")
+	fmt.Printf("SRC_ROOT=%s, PWD=%s, importPath=%s\n", srcRoot, curDir, importPath)
```
robskillington: Do you still need this `fmt.Printf`?  Probably not necessary, or perhaps use a `logger` instead?

src/dbnode/testdata/prototest/fixture.go
```diff
@@ -0,0 +1,230 @@
...
+	"time"
+	"testing"
+
+	"github.com/m3db/m3/src/dbnode/storage/namespace"
+
+	"github.com/jhump/protoreflect/dynamic"
+	"github.com/stretchr/testify/require"
+	"github.com/jhump/protoreflect/desc"
+	"os"
+	"fmt"
```
robskillington: nit: Both "os" and "fmt" should appear in first level of imports with "bytes" "time" and "testing"

src/dbnode/testdata/prototest/fixture.go
```diff
@@ -0,0 +1,230 @@
...
+	expectedMsg := dynamic.NewMessage(md)
+	if expectedMsg.Unmarshal(expected) != nil {
+		return false
+	}
+	actualMsg := dynamic.NewMessage(md)
+	if actualMsg.Unmarshal(actual) != nil {
+		return false
+	}
+	//fmt.Printf("expected: %s\n", expectedMsg.String())
+	//fmt.Printf("actual: %s\n", actualMsg.String())
```
robskillington: nit: Remove commented code?

src/dbnode/testdata/prototest/fixture.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func NewProtoMessageIterator(messages []*dynamic.Message) *ProtoMessageIterator {
+	return &ProtoMessageIterator{messages: messages}
+}
+
+func (pmi *ProtoMessageIterator) Next() []byte {
+	n := pmi.messages[pmi.i%len(pmi.messages)]
+	pmi.i++
+	//fmt.Printf("next message is: %d, %s\n", pmi.i, n.String())
```
robskillington: nit: Remove commented code?

src/dbnode/testdata/prototest/fixture.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+func RequireEqual(t *testing.T, md *desc.MessageDescriptor, expected, actual []byte) {
+	expectedMsg := dynamic.NewMessage(md)
+	require.NoError(t, expectedMsg.Unmarshal(expected))
+	actualMsg := dynamic.NewMessage(md)
+	require.NoError(t, actualMsg.Unmarshal(actual))
+
+	//fmt.Printf("expected: %s\n", expectedMsg.String())
+	//fmt.Printf("actual: %s\n", actualMsg.String())
```
robskillington: nit: Remove commented code?

src/dbnode/storage/series/buffer.go
```diff
@@ -41,6 +41,7 @@ import (
 	xtime "github.com/m3db/m3/src/x/time"
 
 	"go.uber.org/zap"
+	"github.com/m3db/m3/src/dbnode/storage/namespace"
```
robskillington: nit: Surpised lint is picking this up, this should be with the imports above with all `github.com/m3db/m3/src...` imports.

src/dbnode/storage/series/buffer.go
```diff
@@ -951,7 +962,9 @@ func (b *BufferBucket) writeToEncoderIndex(
 	datapoint ts.Datapoint,
 	unit xtime.Unit,
 	annotation []byte,
+	schema namespace.SchemaDescr,
 ) error {
+	b.encoders[idx].encoder.SetSchema(schema)
```
robskillington: Hm, this is going to be setting the schema each and every write.  Is a SetSchema for the same schema a no-op?  If not that may be an issue.

src/dbnode/storage/namespace/schema_registry.go
```diff
@@ -0,0 +1,177 @@
...
+package namespace
+
+import (
+	"fmt"
+	"sync"
+
+	"github.com/m3db/m3/src/x/ident"
+	xclose "github.com/m3db/m3/src/x/close"
+	xwatch "github.com/m3db/m3/src/x/watch"
+	"go.uber.org/zap"
```
robskillington: nit: Suprised lint not picking this up, our lint rules usually define you need to separate third party imports from `"github.com/m3db/m3/src..."` with a new line.  This should be separated, but I'm mainly curious why lint didn't cause CI to fail for your PR...

src/dbnode/storage/namespace/schema_registry.go
```diff
@@ -0,0 +1,177 @@
...
+			}
+		}
+	}()
+
+	return watch, nil
+}
+
+func (sr *schemaRegistry) Close() {
+	sr.RLock()
+	defer sr.RUnlock()
```
robskillington: Seems like this should be `.Lock()` and `.Unlock()` as its modifying state no?

src/dbnode/storage/database.go
robskillington: Ping, think you still need to rename to `curSchemaID`

src/dbnode/storage/series/buffer.go
Haijuncao: we compare version (string compare), if same, no-op.

see SetSchema (or resetSchema) in proto/encode|iterator.go

src/dbnode/testdata/prototest/fixture.go
Haijuncao: sure

src/dbnode/encoding/types.go
Haijuncao: This is for mid-stream schema change case.

src/dbnode/client/config.go
Haijuncao: ok

src/cmd/services/m3dbnode/config/config.go
Haijuncao: sgtm

src/dbnode/storage/namespace/schema_registry.go
Haijuncao: good catch

src/dbnode/storage/series/buffer.go
robskillington: Ok no problem, this is fine.  If it's possible to just make this part of Reset(...) like the other args (such as namespace context were moved to Reset) that'd be great, if it's too difficult then let's just punt on this for now.

src/dbnode/encoding/types.go
robskillington: I see, makes sense.  Please do ignore then =]

src/dbnode/storage/namespace/schema_registry.go
```diff
@@ -0,0 +1,177 @@
...
+
+	if _, ok := history.GetLatest(); !ok {
+		return fmt.Errorf("can not set empty schema history for %v", id.String())
+	}
+
+	sr.Lock()
+	defer sr.Unlock()
+
+	// TODO [haijun] use generated map for optimized map lookup.
+	current, ok := sr.registry[id.String()]
```
robskillington: nit: maybe pull out `id.String()` into a var so it only gets called once?

i.e. `key := id.String()` and then use `key` in map lookups and sets below?

src/dbnode/client/session_proto_test.go
robskillington: Hm interesting, how are we pushing schema updates through etcd out of interest if we can only load it from the filesystem?

robskillington: (Unfortunate that there's no API for non-file based schemas https://godoc.org/github.com/jhump/protoreflect/desc/protoparse)

Haijuncao: writing to a temp dir/file seem to be a good solution. for test, we will use it.

src/dbnode/encoding/proto/encoder.go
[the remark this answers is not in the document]
robskillington: Hm I feel pretty strongly this should be a `return err` not a panic.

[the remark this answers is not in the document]
robskillington: Can you use `instrument.InvariantErrorf(...)` here?  And then you can use an environment variable to make it panic.

See this code:
https://github.com/m3db/m3/blob/master/src/x/instrument/invariant.go#L83-L91

Which will panic if run with:
```
PANIC_ON_INVARIANT_VIOLATED=true ./m3dbnode -f ....
```

src/dbnode/encoding/types.go
```diff
@@ -265,7 +274,7 @@ type NewDecoderFn func() Decoder
 type EncoderAllocate func() Encoder
 
 // ReaderIteratorAllocate allocates a ReaderIterator for a pool.
-type ReaderIteratorAllocate func(reader io.Reader) ReaderIterator
+type ReaderIteratorAllocate func(reader io.Reader, descr namespace.SchemaDescr) ReaderIterator
```
jeromefroe: @prateek @richardartoul this is another breaking change that bit us in statsdex. Would it make sense to take a look at these types and move those that are meant to be exported to the `x` package and those that aren't to an `internal` package to ensure we can't import them?

richardartoul: What was statsdex using it for?

jeromefroe: @arnikola added support to query for encoding cross zone responses in the query service using m3ts... Though that kinda misses the more general point of should this be exposed at all :)

richardartoul: @jeromefroe I don't think moving this would help. If you're using the encoding package to encode data, then the code is going to break when we make breaking changes to the encoders.

I can't make the constructor function for the encoding package private

PR #1579, opened 2019-04-22
[kube] Adding Terraform module for Kube bundle

martin-mao:
Adding Terraform module for the existing kubernetes bundle with an example module for executing on GCP.


kube/terraform/main.tf
```diff
@@ -0,0 +1,309 @@
...
+    }
+  }
+}
+
+resource "kubernetes_config_map" "m3dbnode_config" {
+  metadata {
+    name      = "m3dbnode-config"
+    namespace = "m3db"
+  }
+  data {
```
nerd0: Would it make sense to reference a [template file](https://www.terraform.io/docs/providers/template/d/file.html) that uses the [configuration map](https://github.com/m3db/m3/blob/master/kube/m3dbnode-configmap.yaml) already located within the `kube` directory?

kube/terraform/examples/gcp/README.md
```diff
@@ -0,0 +1,18 @@
...
+This folder shows an example of how to launch the M3DB Kubernetes Bundle on GKE. It sets up a GKE Kube Cluster with three n1-standard-16 nodes. It also creates the GCP specific fast Kubernetes storage class. If using the default terraform created GCP user 'client', the following yaml will need to be applied via kubectl before executing the main bundle tf file:
+
+```yaml
+ClusterRoleBinding 
+apiVersion: rbac.authorization.k8s.io/v1 
+metadata: 
+name: client
+subjects: 
+- kind: User 
+name: client
```
schallert: The spacing got messed up here

kube/terraform/examples/gcp/main.tf
```diff
@@ -0,0 +1,44 @@
...
+  
+  # Setting an empty username and password explicitly disables basic auth
+  master_auth {
+    username = ""
+    password = ""
+  }
+
+  node_config {
+    preemptible  = false
+    machine_type = "n1-standard-16"
```
schallert: Is there any way to make this a user-overrideable variable?

martin-mao: Yeah will do.

kube/terraform/main.tf
martin-mao: Hmm, it would make this sub package depend on the top level file. I think if we were to publish this in the terraform public repo, everything has to be self contained.

martin-mao: It does mean this may need to be in sync with the latest config, but I would hope we don't make backward breaking changes that often.

nerd0: Yea, that ugly string seems like a PITA to update if it does break it. It's kinda of a PITA already to ensure that _all_ yaml configs are updated let alone a yaml => string like this. Will leave the decision in your hands.

kube/terraform/examples/gcp/README.md
schallert: Still messed up? Also missing `kind: ClusterRoleBinding`
```yaml
apiVersion: rbac.authorization.k8s.io/v1 
kind: ClusterRoleBinding 
metadata: 
  name: client
subjects: 
  - kind: User 
    name: client
    apiGroup: rbac.authorization.k8s.io
roleRef: 
  kind: ClusterRole 
  name: "cluster-admin" 
  apiGroup: rbac.authorization.k8s.io
```

martin-mao: Cheers, copy pasta was destroying the formatting. Should be better now.

kube/terraform/main.tf
```diff
@@ -0,0 +1,309 @@
+resource "kubernetes_namespace" "m3db_namespace" {
+  metadata {
+    name = "m3db"
+  }
+}
+
+resource "kubernetes_service" "etcd_service" {
```
schallert: Does order it appears in the file enforce ordering of how the objects are created? Just wondering because technically the headless service needs to be up before the statefulset

martin-mao: No, ordering doesn't matter in the file.

schallert: Is there a way to enforce it? Per the [docs](https://v1-11.docs.kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#statefulsetspec-v1-apps) (emphasis mine):

> serviceName is the name of the service that governs this StatefulSet. This **service must exist before the StatefulSet**, and is responsible for the network identity of the set. Pods get DNS/hostnames that follow the pattern: pod-specific-string.serviceName.default.svc.cluster.local where "pod-specific-string" is managed by the StatefulSet controller.

PR #1581, opened 2019-04-23
[dbnode] Filter out corrupt commit log files that were not present before the node started in the commit log bootstrapper

richardartoul:
Currently the commit log bootstrapper will only attempt to read commit log files that were present on disk before the node started to prevent it from trying to read the files that are currently being written to.

In addition to this check, we also need to add a guard against identifying the active file as corrupt. If the active file appears corrupt it will trigger logic in the commit log bootstrapper that makes configurable decisions about whether the commit log bootstrap should fail or whether the corrupt file should be ignored.

The existing code allowed callers to pass a predicate to the commit log iterator which could be used to filter out files that were not present before the node started, however, this filter was not being applied to files that appeared corrupt. In some rare scenarios, the active commit log file will appear corrupt because it hasn't been flushed yet and the commit log bootstrapper will thus think its encountered a corrupt file when in fact it should have never tried to read that file in the first place.

This P.R changes the iterator such that the predicate can be applied to files identified as corrupt as well, and modifies the commit log bootstrapper to take advantage of this new feature.
Fixes #

```release-note
None
```

```documentation-note
None
```

### DIFF
--- a/src/dbnode/persist/fs/commitlog/types.go
+++ b/src/dbnode/persist/fs/commitlog/types.go
@@ -186,11 +186,19 @@ type Options interface {
 	IdentifierPool() ident.Pool
 }
 
+// FileFilterInfo contains information about a commitog file that can be used to
+// determine whether the iterator should filter it out or not.
+type FileFilterInfo struct {
+	// If isCorrupt is true then File will contain a valid CommitLogFile, otherwise
+	// ErrorWithPath will contain an error and the path of the corrupt file.
+	File      persist.CommitLogFile
+	Err       ErrorWithPath
+	IsCorrupt bool
+}
+
 // FileFilterPredicate is a predicate that allows the caller to determine
-// which commitlogs the iterator should read from. If isCorrupt is true then f will contain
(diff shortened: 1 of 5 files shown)


src/dbnode/persist/fs/commitlog/types.go
```diff
@@ -187,8 +187,10 @@ type Options interface {
 }
 
 // FileFilterPredicate is a predicate that allows the caller to determine
-// which commitlogs the iterator should read from.
-type FileFilterPredicate func(f persist.CommitLogFile) bool
+// which commitlogs the iterator should read from. If isCorrupt is true then f will contain
+// a valid CommitLogFile, otherwise ErrorWithPath will contain an error and the path of the
+// corrupt file.
+type FileFilterPredicate func(isCorrupt bool, f persist.CommitLogFile, err ErrorWithPath) bool
```
robskillington: nit: Can we join these fields into a single struct?

i.e. 
```go
type FileFilterInfo struct {
  File persist.CommitLogFile
  IsCorrupt bool
  Err ErrorWithPath
}
```

PR #1582, opened 2019-04-24
[dbnode - bug fix] Properly propagate replica errors on reset in SeriesIterator

richardartoul:


src/dbnode/encoding/series_iterator.go
```diff
@@ -145,6 +145,11 @@ func (it *seriesIterator) Reset(opts SeriesIteratorOptions) {
 	it.SetIterateEqualTimestampStrategy(opts.IterateEqualTimestampStrategy)
 
 	for _, replica := range opts.Replicas {
+		if replica.Err() != nil {
```
prateek: You’re missing cases still, eg an error can also happen after a Next() call returns false.  

Also, mind adding tests to simulate all these failure conditions

richardartoul: Yeah agreed I don’t think this is ideal for multiple reasons (for example it doesn’t take replication factor into account)

I sent you an email about it

PR #1588, opened 2019-04-25
[docs] add warnings about RF=1 and RF=2

schallert:

This PR adds documentation on why running clusters with RF=1 or RF=2 is a bad idea, as this has been a source of confusion. Additionally this explains our reasoning for disallowing such clusters in the future.

Fixes #

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/docs/operational_guide/replication_and_deployment_in_zones.md
+++ b/docs/operational_guide/replication_and_deployment_in_zones.md
@@ -10,6 +10,22 @@ A replication factor of at least 3 is highly recommended for any M3DB deployment
 
 M3DB will do its best to distribute shards evenly among the availability zones while still taking each individual node's weight into account, but if some of the availability zones have less available hosts than others then each host in that zone will be responsible for more shards than hosts in the other zones and will thus be subjected to heavier load.
 
+
+### Dangers of Replication Factor of 1 or 2
+
+Running with RF=1 or RF=2 is not recommended for any use cases (testing or production), and in [the future][1107] such
+topologies may be rejected by M3DB entirely.
+
+RF=1 is not recommended as it is impossible to perform safe upgrade or tolerate any node failures: as soon as one node
+is down, all writes destined for the shards it owned will fail. If the node's storage is lost (i.e. the disk fails), the
+data is gone forever.
+
+RF=2, despite having an extra replica, entails many of the same problems RF=1 does. As M3DB performs quorum writes and
+reads, as soon as a single node is down (for planned maintenance or an unplanned disruption) clients will be unable to
+perform quorum reads or writes (quorum of 2 nodes is 2). Even if clients relax their consistency guarantees and read
+from the 1 serving node, users may experience flapping results depending on whether one node had data for a time window
+that the other did not.
+
 ### Upgrading hosts in a deployment
 
 When an M3DB node is restarted it has to perform a bootstrap process before it can serve reads. During this time the node will continue to accept writes, but will not be available for reads.
@@ -53,3 +69,5 @@ For example, in a multi-region deployment with four shards spread over five regi
 ![Replication Global](replication_global.png)
 
 Typically, deployments have many more than four shards - this is a simple example that illustrates how M3DB maintains availability while losing up to two regions, as three of five replicas are still intact.
+
+[1107]: https://github.com/m3db/m3/issues/1107
--- a/docs/operational_guide/replication_and_deployment_in_zones.md
+++ b/docs/operational_guide/replication_and_deployment_in_zones.md
@@ -11,20 +11,36 @@ A replication factor of at least 3 is highly recommended for any M3DB deployment
 M3DB will do its best to distribute shards evenly among the availability zones while still taking each individual node's weight into account, but if some of the availability zones have less available hosts than others then each host in that zone will be responsible for more shards than hosts in the other zones and will thus be subjected to heavier load.
 
 
-### Dangers of Replication Factor of 1 or 2
+### Replication Factor Recommendations
 
-Running with RF=1 or RF=2 is not recommended for any use cases (testing or production), and in [the future][1107] such
-topologies may be rejected by M3DB entirely.
+Running with `RF=1` or `RF=2` is not recommended for any multi-node use cases (testing or production). In [the
+future][1107] such topologies may be rejected by M3DB entirely. It is also recommended to only run with an odd number of
+replicas.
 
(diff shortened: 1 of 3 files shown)


docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -10,6 +10,22 @@ A replication factor of at least 3 is highly recommended for any M3DB deployment
 
 M3DB will do its best to distribute shards evenly among the availability zones while still taking each individual node's weight into account, but if some of the availability zones have less available hosts than others then each host in that zone will be responsible for more shards than hosts in the other zones and will thus be subjected to heavier load.
 
+
+### Dangers of Replication Factor of 1 or 2
+
+Running with RF=1 or RF=2 is not recommended for any use cases (testing or production), and in [the future][1107] such
```
andrewmains12: Not useful for testing even? Assuming I'm building from scratch and running locally (as opposed to e.g. running a docker-compose), I would assume that I'd just run RF=1 for that case. Agreed for every other case of course.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -10,6 +10,22 @@ A replication factor of at least 3 is highly recommended for any M3DB deployment
 
 M3DB will do its best to distribute shards evenly among the availability zones while still taking each individual node's weight into account, but if some of the availability zones have less available hosts than others then each host in that zone will be responsible for more shards than hosts in the other zones and will thus be subjected to heavier load.
 
+
+### Dangers of Replication Factor of 1 or 2
```
mway: nit: i'd probably rephrase this to read a little better:

> Dangers of Using Bad Replication Factors

or

> Dangers of Using Replication Factor 1 or 2

or

> Dangers of Using Fewer Than 3 Replicas, or Using Even Numbers of Replicas

or

> Replication Factor Recommendations

depending on preference.

docs/operational_guide/replication_and_deployment_in_zones.md
mway: nit: i'd use two sentences here:

> Running with RF=1 or RF=2 is not recommended for any use cases (testing or production).
> [In the future][1107], such topologies may be rejected by M3DB entirely.

think it's worth putting a logged warning in?

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -10,6 +10,22 @@ A replication factor of at least 3 is highly recommended for any M3DB deployment
...
+
+Running with RF=1 or RF=2 is not recommended for any use cases (testing or production), and in [the future][1107] such
+topologies may be rejected by M3DB entirely.
+
+RF=1 is not recommended as it is impossible to perform safe upgrade or tolerate any node failures: as soon as one node
```
mway: nit: "perform *a* safe upgrade"

docs/operational_guide/replication_and_deployment_in_zones.md
schallert: I should probably clarify "testing" as in "testing out M3DB". My issue is that it's so unrepresentative of a real-world workload that I'd rather even discourage it for kicking the tires. I've talked to folks who have set up an RF=1 cluster to test on, experienced the slightest blip (like a process being restarted) and the cluster becomes stuck.

Maybe discourage it for testing in a > 1 node cluster?

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -10,6 +10,22 @@ A replication factor of at least 3 is highly recommended for any M3DB deployment
...
+is down, all writes destined for the shards it owned will fail. If the node's storage is lost (i.e. the disk fails), the
```
mway: super nit: is disk failure the only mode here? if so, *id est* is appropriate; otherwise, use *exempli gratia* (e.g.).

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -10,6 +10,22 @@ A replication factor of at least 3 is highly recommended for any M3DB deployment
...
+data is gone forever.
+
+RF=2, despite having an extra replica, entails many of the same problems RF=1 does. As M3DB performs quorum writes and
```
mway: nit:

> When M3DB is configured to perform quorum writes and reads, a single node failure ([...]) will cause clients to be unable to read or write (as the quorum of 2 nodes is 2).

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -10,6 +10,22 @@ A replication factor of at least 3 is highly recommended for any M3DB deployment
...
+reads, as soon as a single node is down (for planned maintenance or an unplanned disruption) clients will be unable to
+perform quorum reads or writes (quorum of 2 nodes is 2). Even if clients relax their consistency guarantees and read
+from the 1 serving node, users may experience flapping results depending on whether one node had data for a time window
```
mway: nit: "from the remaining serving node"

docs/operational_guide/replication_and_deployment_in_zones.md
schallert: I'd rather save any code changes for when we fully address that issue, as we may prohibit it entirely.

docs/operational_guide/replication_and_deployment_in_zones.md
```diff
@@ -10,6 +10,39 @@ A replication factor of at least 3 is highly recommended for any M3DB deployment
 
 M3DB will do its best to distribute shards evenly among the availability zones while still taking each individual node's weight into account, but if some of the availability zones have less available hosts than others then each host in that zone will be responsible for more shards than hosts in the other zones and will thus be subjected to heavier load.
 
+
+### Replication Factor Recommendations
+
+Running with `RF=1` or `RF=2` is not recommended for any multi-node use cases (testing or production). In [the
+future][1107] such topologies may be rejected by M3DB entirely. It is also recommended to only run with an odd number of
```
richardartoul: This seems aggressive. Its still really useful for running docker containers on limited hardware (like a laptop) to only have one node

schallert: @richardartoul yeah I can try to clarify this more. I said not recommended for "multi-node use cases" and by multi-node testing I meant as in testing a deployment of M3, not as much local testing

robskillington:
Nice work.

PR #1591, opened 2019-04-26
[dbnode] Make message name configurable when parsing proto schema + allocate if no byte pool

richardartoul:
Makes the message name configurable when parsing a proto schema + allow the proto iterator to operate even if the bytes pool is not configured.

```release-note
NONE
```

```documentation-note
NONE
```


src/dbnode/encoding/proto/parse_schema.go
```diff
@@ -49,7 +45,7 @@ func ParseProtoSchema(filePath string) (*desc.MessageDescriptor, error) {
 
 	// TODO(rartoul): This will be more sophisticated later, but for now assume
 	// that the message will be called "Schema".
```
justinjc: Nit: remove this TODO.

PR #1595, opened 2019-04-30
Re-enable FOSSA

robskillington:
This reverts commit b3cf54e4656b095f0a05139b85c40c7a521a65cb.

Re-enable FOSSA for CI.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -84,6 +84,15 @@ steps:
         gopath-checkout#v1.0.1:
           import: github.com/m3db/m3
     <<: *common
+  - label: "FOSSA license scan"
+    command: make clean install-vendor fossa
+    plugins:
+      docker-compose#v2.5.1:
+        run: app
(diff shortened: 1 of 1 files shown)


.buildkite/pipeline.yml
```diff
@@ -84,6 +84,15 @@ steps:
         gopath-checkout#v1.0.1:
           import: github.com/m3db/m3
     <<: *common
+  - label: "FOSSA license scan"
+    command: make clean install-vendor fossa
+    plugins:
```
schallert: Can we try the new soft_fail feature here too so these failures don't block docs / docker builds?
```yaml
soft_fail:
- exit_status: "*"
```

robskillington: Hm, as we discussed - I'd rather not until it starts to flake.  Before we experienced issues with the request size upload FOSSA was very reliable.

Looks like the issues have been resolved.

PR #1596, opened 2019-04-30
[db] Add an option to use db writes in index mode

arnikola:
Adds an option to use a namespace in index mode; i.e. all writes to this namespace will get truncated to block size and have a set value, meaning we write at most one point per block per series, allowing this namespace to be used as a dedicated index more easily.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/scripts/development/m3_stack/m3dbnode.yml
+++ b/scripts/development/m3_stack/m3dbnode.yml
@@ -1,12 +1,4 @@
 db:
-  transforms:
-    truncateBy: block
-    truncateDuration: 1h
-
-  index:
-    forwardIndexThreshold: 0.1
-    forwardIndexProbability: 0.1
-
   logging:
     level: info
 
--- a/scripts/development/m3_stack/m3dbnode.yml
+++ b/scripts/development/m3_stack/m3dbnode.yml
@@ -1,4 +1,12 @@
 db:
+  transforms:
+    truncateBy: block
+    truncateDuration: 1h
+
+  index:
+    forwardIndexThreshold: 0.1
+    forwardIndexProbability: 0.1
+
   logging:
     level: info
 
--- a/src/dbnode/server/server.go
+++ b/src/dbnode/server/server.go
@@ -342,7 +342,6 @@ func Run(runOpts RunOptions) {
 			CacheRegexp: plCacheConfig.CacheRegexpOrDefault(),
 			CacheTerms:  plCacheConfig.CacheTermsOrDefault(),
 		})
-	opts = opts.SetIndexOptions(indexOpts)
 
 	if tick := cfg.Tick; tick != nil {
 		runtimeOpts = runtimeOpts.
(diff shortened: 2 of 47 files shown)


src/dbnode/server/server.go
```diff
@@ -1281,6 +1280,8 @@ func withEncodingAndPoolingOptions(
 	aggregateQueryResultsPool := index.NewAggregateResultsPool(
 		poolOptions(policy.IndexResultsPool, scope.SubScope("index-aggregate-results-pool")))
 
+	// Set index options.
+	opts = opts.SetTruncateType(cfg.Transforms.TruncateBy)
```
robskillington: nit: Maybe move this above the `// Set index options` since this transforms values? i.e. maybe just comment with `// Set value transformations`?

src/dbnode/storage/namespace_test.go
```diff
@@ -97,14 +98,46 @@ func newTestNamespaceWithIDOpts(
...
+	index namespaceIndex,
+) (*dbNamespace, closerFn) {
 	ns, closer := newTestNamespace(t)
 	if index != nil {
 		ns.reverseIndex = index
 	}
 	return ns, closer
 }
 
+func newTestNamespaceWithUseAsIndex(
```
robskillington: Rename to `newTestNamespaceWithTruncateType(..)`?

src/dbnode/storage/series/truncate_type.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"fmt"
+)
+
+// TruncateType determines the scheme for truncating transforms.
+type TruncateType uint8
+
+const (
+	TypeNone TruncateType = iota
+	TypeBlock
```
robskillington: Maybe instead of just `TypeBlock` we could do `TypeTimestampBlock` so it is clear that it's truncating just the timestamp?

src/dbnode/storage/series/buffer.go
```diff
@@ -60,6 +60,10 @@ const (
 	// enabled to see if this is a sane number.
 	evictedTimesArraySize = 8
 	writableBucketVer     = 0
+
+	// NB: this is the value all points will be truncated to.
+	truncatedValue float64 = 0
```
robskillington: Maybe we don't set it to a specific value and just let whatever the client is writing be the value it sets?

src/dbnode/storage/series/buffer.go
```diff
@@ -60,6 +60,10 @@ const (
...
+	truncatedUnit          = xtime.Second
```
robskillington: We also probably don't need to override the unit no?  Client should always be sending the same unit unless it has a bug.

src/dbnode/storage/series/buffer.go
arnikola: Fair; pulling this into configs as discussed

src/dbnode/storage/series/buffer.go
arnikola: Yeah fair, this was mostly to kinda standardize all values going into the index but don't suppose it matters very much here

scripts/development/m3_stack/m3dbnode.yml
```diff
@@ -1,4 +1,12 @@
 db:
+  transforms:
```
arnikola: oops will revert this

robskillington: 👍

robskillington:
Also looks like some big unit tests and docker integration tests are strangely failing?

PR #1597, opened 2019-05-01
Custom protobuf unmarshaler for reducing allocations during protobuf encoding

richardartoul:
Unmarshaling every protobuf write into a `*dynamic.Message` is really expensive as it causes many short-lived allocations. This P.R adds a custom protobuf unmarshaler that can be used to optimistically unmarshal custom-encoded fields in an efficient and allocation-free manner while still falling back to the dynamic protobuf library when dealing with complex messages.

`unmarshal.md` added in this P.R has more information.

Fixes #

```release-note
NONE
```

```documentation-note
Yes, Protobuf README will be updated.
```

**Benchmarks**

Before
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/encoding/proto
BenchmarkEncode-8   	 1000000	      1540 ns/op	     256 B/op	       8 allocs/op
PASS
ok  	github.com/m3db/m3/src/dbnode/encoding/proto	1.583s
Success: Benchmarks passed.
```

After
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/encoding/proto
BenchmarkEncode-8   	 2000000	       796 ns/op	      16 B/op	       0 allocs/op
PASS
ok  	github.com/m3db/m3/src/dbnode/encoding/proto	2.419s
Success: Benchmarks passed.
```


src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,375 @@
...
+	for !u.decodeBuf.eof() {
+		tagAndWireTypeStartOffset := u.decodeBuf.index
+		fieldNum, wireType, err := u.decodeBuf.decodeTagAndWireType()
+		if err != nil {
+			return err
+		}
+
+		fd := u.schema.FindFieldByNumber(fieldNum)
+		if fd == nil {
+			return fmt.Errorf("encountered unknown field with field number: %d", fieldNum)
```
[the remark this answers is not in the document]
richardartoul: I could still handle it by just deferring to the dynamic protobuf library like I do for complex fields (repeated, maps, nested messages, etc) but I think for now I'd prefer to reject the writes. We can always change this later / make it configurable but it potentially impacts performance so much I'd prefer people start getting write errors because they forgot to upgrade their schema instead of having their nodes suddenly start having really bad compression.

Alternatively I could just ignore unknown fields but that seems like the worst of both worlds

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,375 @@
...
+			if err != nil {
+				return err
+			}
+
+			var (
+				length   = u.decodeBuf.index - tagAndWireTypeStartOffset
+				startIdx = tagAndWireTypeStartOffset
+				endIdx   = startIdx + length
+			)
+			u.nonCustomValues.UnmarshalMerge(u.decodeBuf.buf[tagAndWireTypeStartOffset:endIdx])
```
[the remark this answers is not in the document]
richardartoul: Yeah I was mostly trying to reduce line length but you're right I overcomplicated it. I simplified it and also added a comment explaining why what I'm doing is valid

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,381 @@
...
+				endIdx         = u.decodeBuf.index
+				marshaledField = u.decodeBuf.buf[startIdx:endIdx]
+			)
+			// A marshaled Protobuf message consists of a stream of <fieldNumber, wireType, value>
+			// tuples, all of which are optional, with no additional header or footer information.
+			// This means that each tuple within the stream can be thought of as its own complete
+			// marshaled message and as a result we can build up the nonCustomValues *dynamic.Message
+			// one field at a time by calling UnmarshalMerge() on sub-slices that contain a complete
+			// tuple.
+			u.nonCustomValues.UnmarshalMerge(marshaledField)
```
justinjc: Should you handle the returned `err` from this func call?

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,381 @@
...
+			continue
+		}
+
+		value, err := u.unmarshalKnownField(fd, wireType)
+		if err != nil {
+			return err
+		}
+
+		if isSorted && len(u.customValues) > 1 {
+			// Check if the slice is sorted as its built to avoid resorting
```
justinjc: s/its/it's

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,381 @@
...
+	return nil
+}
+
+func (u *customUnmarshaler) isCustomField(fd *desc.FieldDescriptor) bool {
+	if fd.IsRepeated() || fd.IsMap() {
+		// Map should always be repeated but include the guard just in case.
+		return false
+	}
+
+	if fd.GetMessageType() != nil {
```
justinjc: Why not just one `if` here?

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,381 @@
...
+	case proto.WireStartGroup:
+		return 0, errGroupsAreNotSupported
+	case proto.WireEndGroup:
+		return 0, errGroupsAreNotSupported
+	default:
+		return 0, proto.ErrInternalBadWireType
+	}
+}
+
+func (u *customUnmarshaler) unmarshalKnownField(fd *desc.FieldDescriptor, wireType int8) (unmarshalValue, error) {
```
justinjc: Maybe `unmarshalCustomField` to be consistent with naming between custom/known.

src/dbnode/encoding/proto/encoder.go
```diff
@@ -222,6 +170,46 @@ func (enc *Encoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, protoBytes ts.A
 	return nil
 }
 
+func (enc *Encoder) encodeSchemaAndOrTimeUnit(
+	needToEncodeSchema bool,
+	needToEncodeTimeUnit bool,
+	timeUnit xtime.Unit) {
```
justinjc: Nit: newline for `) {`

src/dbnode/encoding/proto/encoder.go
```diff
@@ -222,6 +170,46 @@ func (enc *Encoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, protoBytes ts.A
...
+	// First bit means either there is no more data OR the time unit and/or schema has changed.
+	enc.stream.WriteBit(opCodeNoMoreDataOrTimeUnitChangeAndOrSchemaChange)
+	// Next bit means there is more data, but the time unit and/or schema has changed has changed.
```
justinjc: s/has changed has changed/has changed

src/dbnode/encoding/proto/custom_unmarshal.go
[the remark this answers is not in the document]
Haijuncao: sounds good, agree we should start with the stricter mode, the alternative is too much delayed surprise.  Client side configuration should require a schema version, which by itself already encourages good behavior, because if application developer forgets to deploy schema, the write will fail at client side due to unable to marshal any new fields (presumably that's primary reason for schema update), when opening a session, schema version is sent as part of the session context, then we can reject the write on dbnode side if dbnode has not gotten the most up to date schema (from etcd watches). Unless there is some serious network partition between meta and db node, this should be a very rare error.

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// This file is mostly copy-pasta from: https://github.com/jhump/protoreflect/blob/master/dynamic/codec.go
+// since the jhump/protoreflect library does not expose the `codedBuffer` type.
```
robskillington: Hm, could we also translate the tests for this file?

richardartoul: @robskillington The `protoreflect` library doesn't actually have tests for this file, I think in both cases it implicitly gets tested via other tests.

I could try and pull out the tests from the upstream protobuf library that the `protoreflect` ripped it out of, but I think we can delete this file altogether once this P.R lands: https://github.com/jhump/protoreflect/blob/master/dynamic/codec.go so maybe best to just wait until then (I need to upgrade the `protoreflect` library in a subsequent P.R anyways to pull in some perf improvements so I can wait until that P.R gets merged, update the dep, then delete this file)

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+
+The code in the `Buffer` is mostly self explanatory for anyone familiar with [proto3 encoding format](https://developers.google.com/protocol-buffers/docs/encoding).
+
+### CustomFieldUnmarshaler
+
+The `customFieldUnmarshaler` has three primary responsibilities:
+
+1. Provide an interface for efficiently unmarshaling top-level scalar fields in a marshaled Protobuf message without allocating.
+2. Ensure that the values unmarshaled in #1 are sorted by field number.
+3. Provide a `*dynamic.Message` that contains *only* the fields that could not be unmarshaled efficiently in #1 and does not allocate / expend any resources at all in the case of optimized schemas that don't contain any fields that can't be handled by #1.
```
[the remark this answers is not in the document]
richardartoul: awesome thanks

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,381 @@
...
+		return false
+	}
+
+	return true
+}
+
+// skip will skip over the next value in the encoded stream (given that the tag and
+// wiretype have already been decoded). Additionally, it can optionally re-encode
+// the skipped <tag,wireType,value> tuple into the skippedBuf stream so that it can
+// be handled later.
```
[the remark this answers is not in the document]
richardartoul: stale comment, will delete

src/dbnode/encoding/proto/custom_unmarshal.go
richardartoul: just wanted to separate them out logically, should be the same performance wise

src/dbnode/encoding/proto/custom_unmarshal.go
richardartoul: done

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+
+This is inefficient primarily because unmarshaling into a `*dynamic.Message` is very expensive since it involves a lot of `interface{}` magic as well as allocating a large number of very short-lived objects that are difficult to reuse.
+It's especially inefficient for Protobuf schemas that are optimized for this package (specifically those that make heavy use of top-level scalar fields where allocating objects on the heap just to wrap primitive types is especially wasteful).
+
+As a result, this package implements a `customFieldUnmarshaler` that accepts a `[]byte` which constitutes a marshaled Protobuf message and exposes methods for unmarshaling the top-level scalar fields (i.e the fields that the encoder can perform custom compression on) in an efficient and reusable manner such that in the general case there are no allocations.
+In addition, it uses the `jhump/protoreflect` library to unmarshal any other fields that it can't unmarshal efficiently (although this has zero overhead if those fields are not present).
```
mway: s/although//

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+
+
+## Implementation
+
+### Overview
+
+The implementation is broken into two parts:
+
+1. The `Buffer` which is almost identical to https://github.com/jhump/protoreflect/blob/master/dynamic/codec.go and provides an interface for iterating a marshaled protobuf message one `<fieldNumber, wiretype, value>` tuple at a time.
```
mway: nit: `buffer` is the unexported type, there is no `Buffer` :) also,

> The `buffer`, which is similar to [protoreflect's dynamic codec](https://github.com/jhump/protoreflect/blob/master/dynamic/codec.go) and provides an interface for iterating over a marshaled Protobuf message, one `<fields>` tuple at a time.

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+
+2. The `customFieldUnmarshaler` which wraps the `Buffer` and exposes an interface for efficiently unmarshaling top-level scalar fields with no allocations, as well as a fallback mechanism that relies on the `jhump/protoreflect` library for any fields that don't have an efficient unmarshaling codepath (`maps`, `repeated` fields, and nested messages, etc.)
```
mway: nit:

> The `customFieldUnmarshaler`, which wraps `buffer`

also, move the period outside the parenthesis: "nested messages, etc.)" -> "nested messages, etc)."

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+
+### Buffer
+
+The code in the `Buffer` is mostly self explanatory for anyone familiar with [proto3 encoding format](https://developers.google.com/protocol-buffers/docs/encoding).
```
mway: nit: self-explanatory, **the** proto3 encoding format

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+### CustomFieldUnmarshaler
+
+The `customFieldUnmarshaler` has three primary responsibilities:
+
+1. Provide an interface for efficiently unmarshaling top-level scalar fields in a marshaled Protobuf message without allocating.
+2. Ensure that the values unmarshaled in #1 are sorted by field number.
+3. Provide a `*dynamic.Message` that contains *only* the fields that could not be unmarshaled efficiently in #1. This does not allocate / expend any resources at all in the case of optimized schemas that only contain fields that can be handled by #1.
+
+The `customFieldUnmarshaler` works by iterating through all the `<fieldNumber, wireType, value>` tuples in the marshaled Protobuf and checking if they are supported by the efficient code path.
+If they are, then it unmarshals the value into an `unmarshalValue` which is a space-optimized type that can be re-used without any allocations.
```
mway: nit: s/then//, "reused"

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+If the tuple cannot be unmarshaled efficiently then the unmarshaler falls back to the `jhump/protoreflect` library and uses the `UnmarshalMerge` method to unmarshal a single field at a time so that it can be used iteratively without having to rewrite the marshaled message even if custom and non-custom fields are interleaved together.
```
mway: [conditional sentences](https://www.grammarly.com/blog/conditional-sentences/) should generally be comma-delimited:

> [...] efficiently, the marshaler [...]

minor tweak for readability (dang, that's a long sentence):

> [...] library and uses `UnmarshalMerge` to iteratively unmarshal each field sequentially to avoid rewriting the message. This approach supports both any combination of custom and non-custom fields, and in any order.

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+
+The end result of the unmarshaling process is that the caller will get a sorted (by field number) slice of `unmarshalValue` (and the slice is reused over and over again so the amortized allocation cost per unmarshal approaches zero) which contains all the values that can be custom encoded and a `*dynamic.Message` that contains all the more complex fields (if any) that cannot be unmarshaled or compressed efficiently.
```
mway: i'd break this into a couple sentences:

> The output of unmarshaling is a slice of `unmarshalValue`s (sorted by field number) containing all custom-encoded values and a `*dynamic.Message` containing any complex fields that cannot be unmarshaled or compressed efficiently. This value slice is reused to mitigate allocation costs for subsequent unmarshaling.

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+
+Note that the `customFieldUnmarshaler` only returns an `unmarshalValue` for fields that were actually encoded into the stream. According to the [Proto3 encoding format](https://developers.google.com/protocol-buffers/docs/encoding) fields set to their default values are left out of the marshaled stream.
```
mway: nit: add a comma after "Proto3 encoding format"; s/left out of/omitted from/

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
...
+This means that callers need to account for the fact that if an `unmarshalValue` is not present for a field that should have had one then the implied value is the default value for that fields type.
```
mway: this is a bit confusing as written, maybe try something like:

> Thus, if an `unmarshalValue` is not present for a field (and the given field would nominally unmarshal to an `unmarshalValue`), then that value is the type's default value.

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	// The number is too large to represent in a 64-bit value.
+	err = errOverflow
+	return
+}
+
+// DecodeVarint reads a varint-encoded integer from the Buffer.
+// This is the format for the
+// int32, int64, uint32, uint64, bool, and enum
+// protocol buffer types.
+func (cb *buffer) decodeVarint() (uint64, error) {
```
mway: the bit management for these functions looks sane, but i'm on the fence as to whether we should pull some of this business out into more testable components - right now, branch coverage is entirely dependent on input properties, which implicitly bloats the tests and makes test cases the driver rather than tests themselves. (generally, table-based tests test low-level variations on a single theme versus a gamut of functionality.)

tbh, i think that, at a minimum, we should test the buffer directly, but ideally we'd also have discrete tests for each type decoder.

what do you think?

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	i := cb.index
+	buf := cb.buf
+
+	if i >= len(buf) {
+		return 0, io.ErrUnexpectedEOF
+	} else if buf[i] < 0x80 {
+		cb.index++
+		return uint64(buf[i]), nil
+	} else if len(buf)-i < 10 {
```
mway: jfmi, why 10?

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	x -= 0x80 << 56
+
+	b = uint64(buf[i])
+	i++
+	x += b << 63
+	if b&0x80 == 0 {
+		goto done
+	}
+	// x -= 0x80 << 63 // Always zero.
```
mway: nix this?

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+
+	return 0, errOverflow
+
+done:
```
mway: how much of a penalty is there for doing a loop with step += 7 and just breaking instead of using `goto`?

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+	tag = int32(v)
+	return
+}
+
+// DecodeFixed64 reads a 64-bit integer from the Buffer.
+// This is the format for the
+// fixed64, sfixed64, and double protocol buffer types.
+func (cb *buffer) decodeFixed64() (x uint64, err error) {
+	// x, err already 0
```
mway: nit: superfluous comment

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	i := cb.index + 8
+	if i < 0 || i > len(cb.buf) {
```
mway: don't we want to check if `cb.index < 8`? that seems worse (and more likely) than `i < 0`, and you'll run into bounds issues below otherwise.

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	return
+}
+
+// DecodeFixed32 reads a 32-bit integer from the Buffer.
+// This is the format for the
+// fixed32, sfixed32, and float protocol buffer types.
+func (cb *buffer) decodeFixed32() (x uint64, err error) {
+	// x, err already 0
+	i := cb.index + 4
+	if i < 0 || i > len(cb.buf) {
```
mway: same here RE bounds checking

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	x |= uint64(cb.buf[i-2]) << 48
+	x |= uint64(cb.buf[i-1]) << 56
+	return
+}
+
+// DecodeFixed32 reads a 32-bit integer from the Buffer.
+// This is the format for the
+// fixed32, sfixed32, and float protocol buffer types.
+func (cb *buffer) decodeFixed32() (x uint64, err error) {
+	// x, err already 0
```
mway: nit: superfluous comment

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	nb := int(n)
+	if nb < 0 {
+		return nil, fmt.Errorf("proto: bad byte length %d", nb)
+	}
+	end := cb.index + nb
+	if end < cb.index || end > len(cb.buf) {
```
mway: might be glossing over something, but the first part of this condition shouldn't be possible as we checked `nb < 0` above

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+		return nil, io.ErrUnexpectedEOF
+	}
+
+	if !alloc {
```
mway: when would we want to alloc vs not? does this need to be an option? couldn't we control this at the struct level instead (callsite allocation) or just infer based on the `nil`ness of `cb.buf`?

src/dbnode/encoding/proto/common.go
```diff
@@ -34,7 +34,7 @@ import (
 type Schema *desc.MessageDescriptor
 
 const (
-	// ~1GiB is an intentionally a very large number to avoid users ever running into any
+	// ~1GiB is intentionally a very large number to avoid users ever running into any
```
mway: nit:

> ~`1GiB is an intentionally large number [...]

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+	numNonCustom    int
+}
+
+func newCustomFieldUnmarshaler() customFieldUnmarshaler {
+	return &customUnmarshaler{
+		decodeBuf: newCodedBuffer(nil),
+	}
+}
+
+func (u *customUnmarshaler) sortedCustomFieldValues() sortedCustomFieldValues {
```
mway: could we name this `customFieldValues()` given that they're already sorted?

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+	return u.customValues
+}
+
+func (u *customUnmarshaler) numNonCustomValues() int {
+	return u.numNonCustom
```
mway: nit: `numNonCustomValues` (in keeping with `nonCustomValues`)

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+			}
+		}
+
+		u.customValues = append(u.customValues, value)
+	}
+
+	u.decodeBuf.reset(u.decodeBuf.buf)
+
+	if !isSorted {
+		// Avoid resorting if possible.
```
mway: ultra nit: move this comment outside the condition

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+		// Map should always be repeated but include the guard just in case.
+		return false
+	}
+
+	if fd.GetMessageType() != nil {
+		// Skip nested messages.
+		return false
+	}
+
+	return true
```
mway: "custom" terminology is a bit confusing - iiuc, this method really says "would we have theoretically encoded this field or left it up to the protobuf encoder?".

not really the right place to go back and retrofit the word "custom" everywhere, but can you add a quick doc on this method explaining what question it's actually answering?

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+}
+
+// skip will skip over the next value in the encoded stream (given that the tag and
+// wiretype have already been decoded).
+func (u *customUnmarshaler) skip(wireType int8) (int, error) {
+	switch wireType {
+	case proto.WireFixed32:
+		numSkipped := 4
```
mway: maybe `bytesSkipped` for clarity?

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+		// Skip nested messages.
+		return false
+	}
+
+	return true
+}
+
+// skip will skip over the next value in the encoded stream (given that the tag and
+// wiretype have already been decoded).
+func (u *customUnmarshaler) skip(wireType int8) (int, error) {
```
mway: is it worth having the `switch` manage a function-scoped `offset` and doing the math in one place at the end of the function, do you think?

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+	case proto.WireBytes:
+		var (
+			numSkipped                 = 0
+			offsetBeforeDecodeRawBytes = u.decodeBuf.index
+		)
+		// Bytes aren't copied because they're just being skipped over so
+		// copying would be wasteful.
+		_, err := u.decodeBuf.decodeRawBytes(false)
+		if err != nil {
+			return numSkipped, err
```
mway: nit: `0`, just to clearly communicate that the value (even though we can see that the var is set to 0) is invalid

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+		return unmarshalSimpleField(fd, num)
+	case proto.WireVarint:
+		num, err := u.decodeBuf.decodeVarint()
+		if err != nil {
+			return zeroValue, err
+		}
+		return unmarshalSimpleField(fd, num)
+
+	case proto.WireBytes:
+		if fd.GetType() != dpb.FieldDescriptorProto_TYPE_BYTES &&
```
mway: just `GetType()` once:

```go
if t := fd.GetType(); t != A && t != b {
  // ...
}
```

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+
+	case dpb.FieldDescriptorProto_TYPE_SINT32:
+		if v > math.MaxUint32 {
+			return zeroValue, dynamic.NumericOverflowError
+		}
+		val.v = uint64(decodeZigZag32(v))
+		return val, nil
+
+	case dpb.FieldDescriptorProto_TYPE_SINT64:
+		val.v = uint64(decodeZigZag64(v))
```
mway: do we need overflow checking here?

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+import (
+	"errors"
+	"fmt"
+	"io"
+	"math"
+)
+
+var (
+	// errOverflow is returned when an integer is too large to be represented.
+	errOverflow = errors.New("proto: integer overflow")
```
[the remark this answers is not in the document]
richardartoul: This is copy paste code from the `jhump/protoreflect` library and I'm hoping to delete it all once they merge [this P.R](https://github.com/jhump/protoreflect/pull/196) so I don't want to customize it too much (I think the author might be on vacation right now or something)

src/dbnode/encoding/proto/buffer.go
```diff
@@ -0,0 +1,291 @@
...
+// type and for embedded messages.
+func (cb *buffer) decodeRawBytes(alloc bool) (buf []byte, err error) {
+	n, err := cb.decodeVarint()
+	if err != nil {
+		return nil, err
+	}
+
+	nb := int(n)
+	if nb < 0 {
+		return nil, fmt.Errorf("proto: bad byte length %d", nb)
```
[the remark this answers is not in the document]
richardartoul: same comment

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+		if err != nil {
+			return err
+		}
+
+		fd := u.schema.FindFieldByNumber(fieldNum)
+		if fd == nil {
+			return fmt.Errorf("encountered unknown field with field number: %d", fieldNum)
+		}
+
+		isCustomField := u.isCustomField(fd)
```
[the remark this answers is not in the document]
richardartoul: good call

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,379 @@
...
+			return zeroValue, err
+		}
+		return unmarshalSimpleField(fd, num)
+	case proto.WireVarint:
+		num, err := u.decodeBuf.decodeVarint()
+		if err != nil {
+			return zeroValue, err
+		}
+		return unmarshalSimpleField(fd, num)
+
```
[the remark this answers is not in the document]
richardartoul: for large blocks I like to let them breathe a little bit otherwise i find it overwhelming to trace them. I'll add the newlines

src/dbnode/encoding/proto/docs/unmarshal.md
```diff
@@ -0,0 +1,50 @@
+# Top Level Scalar Unmarshaler
+
+## Overview
+
+It's recommended that readers familiarize themselves with the [proto3 encoding documentation](https://developers.google.com/protocol-buffers/docs/encoding) before reading the remainder of this document.
+
+The Encoder in this package is responsible for accepting a stream of marshaled Protobuf messages and encoding them into a new compressed stream.
```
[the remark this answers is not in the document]
richardartoul: Updated both docs to be more clear

src/dbnode/encoding/proto/buffer.go
richardartoul: same comment as above, this is actualy copy-pasta from a library and hopefully a P.R will land upstream soon that means I don't have to keep this code in M3DB. Also, the buffer is not tested directly in that library so I can't even copy paste the tests :p

src/dbnode/encoding/proto/buffer.go
richardartoul: copy pasta, I have no idea

src/dbnode/encoding/proto/buffer.go
richardartoul: same comment

src/dbnode/encoding/proto/buffer.go
richardartoul: same comment

src/dbnode/encoding/proto/buffer.go
richardartoul: same comment

src/dbnode/encoding/proto/buffer.go
richardartoul: same comment

src/dbnode/encoding/proto/buffer.go
richardartoul: same comment

src/dbnode/encoding/proto/buffer.go
richardartoul: same comment

src/dbnode/encoding/proto/buffer.go
richardartoul: same comment

src/dbnode/encoding/proto/buffer.go
richardartoul: same comment

src/dbnode/encoding/proto/custom_unmarshal.go
richardartoul: @mway I prefer this just because its really important that this return sorted results so I want the API to make that clear

src/dbnode/encoding/proto/custom_unmarshal.go
richardartoul: conflicts with the method name

src/dbnode/encoding/proto/custom_unmarshal.go
richardartoul: I could, but I think right now its a little easier to follow because we're doing early returns everywhere so keeping it local makes it easier to read IMO

src/dbnode/encoding/proto/custom_unmarshal.go
richardartoul: no because really what we're doing here is reusing the `uint64` field to store different bit patterns and then we'll re-materialize it into an int64 later (the conversion from int64 to uint64 even if it causes an overflow won't change the bit pattern so you'll still get the same results)

src/dbnode/encoding/proto/custom_unmarshal.go
richardartoul: yes thank you

src/dbnode/encoding/proto/custom_unmarshal.go
mway: sure, sgtm

src/dbnode/encoding/proto/custom_unmarshal.go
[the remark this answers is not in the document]
robskillington: We could eventually try to make DB nodes take part in acknowledging the schema update before enabling it and returning success to the client. (i.e. two phase commit for schema updates)

That way the user experience is pretty reasonable since it wouldn't be successful until all nodes acknowledge and reply success by default.  There could also be a mode to avoid the two phase commit if operator desires, that way you could make schema changes even with nodes that are currently unhealthy.

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,386 @@
...
+		fd := u.schema.FindFieldByNumber(fieldNum)
+		if fd == nil {
+			return fmt.Errorf("encountered unknown field with field number: %d", fieldNum)
+		}
+
+		if !u.isCustomField(fd) {
+			if u.nonCustomValues == nil {
+				u.nonCustomValues = dynamic.NewMessage(u.schema)
+			}
```
robskillington: Perhaps think about extracting this into a sub method in the future?

i.e.

```go
func (u *customUnmarshaler) nonCustomVals() *dynamic.Message {
  if u.nonCustomValues == nil {
    u.nonCustomValues = dynamic.NewMessage(u.schema)
  }
  return u.nonCustomValues
}
```

Then use as 

```go
// reset
u.nonCustomVals().Reset()
```

etc

robskillington: (it'll very likely get inlined)

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -0,0 +1,386 @@
...
+			return 0, err
+		}
+		bytesSkipped += u.decodeBuf.index - offsetBeforeDecodeRawBytes
+		return bytesSkipped, nil
+
+	case proto.WireStartGroup:
+		return 0, errGroupsAreNotSupported
+
+	case proto.WireEndGroup:
+		return 0, errGroupsAreNotSupported
```
robskillington: What are groups used for out of interest?

PR #1598, opened 2019-05-01
Set etcd related globals in each service main initialization

robskillington:

This centralizes the global etcd package related configuration we do to `src/x/etcd` and since globals should only ever be set once per process, each service calls into this package at service main initialization.

```release-note
NONE
```

```documentation-note
NONE
```


robskillington:
@mway sure thing, I'll add a link to the godoc for that package.

PR #1600, opened 2019-05-02
Generic list with pooled elements

justinjc:

It is useful to have a list with its elements pooled, for example in the case where you want to reuse a list between loop iterations to avoid extra allocations.

**Special notes for your reviewer**: most of the list code is from Go's `container/list` implementation, so I included their license in here too.

```release-note
NONE
```

```documentation-note
NONE
```


src/x/generics/list/list.go
```diff
@@ -0,0 +1,314 @@
...
+// LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+// DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+// THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+// (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+// OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+package list
+
+import (
+	"github.com/m3db/m3/src/x/pool"
```
prateek: nit: need a space here.

src/x/generics/list/list.go
```diff
@@ -0,0 +1,314 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
prateek: wanna copy the tests for this file too?

src/x/generics/list/list.go
```diff
@@ -0,0 +1,314 @@
...
+	for e := l.Back(); e != nil; l.Remove(e) {
+	}
+}
+
+// ElementPool provides a pool for Elements.
+type ElementPool struct {
+	pool pool.ObjectPool
+}
+
+// Get gets a BufferBucketVersions from the pool.
```
prateek: looks like these comments need to be updated to remove mentions of `BufferBucketVersions`

src/x/generics/list/list.go
```diff
@@ -0,0 +1,314 @@
...
+func (l *List) Init() *List {
+	l.root.next = &l.root
+	l.root.prev = &l.root
+	l.len = 0
+	l.ePool = newElementPool(l.poolOpts)
+	return l
+}
+
+// New returns an initialized list.
+func New(opts pool.ObjectPoolOptions) *List {
```
prateek: more a question: do you want to be able to pass in an element pool externally here? might make it easier to reuse pooled objects

src/x/generated-source-files.mk
```diff
@@ -176,6 +176,38 @@ genny-leakcheckpool:
 	$(eval out_dir=$(gopath_prefix)/$(target_package))
 	cat $(m3x_package_path)/generics/leakcheckpool/pool.go | grep -v nolint | genny -pkg $(pkg) -ast gen "elemType=$(elem_type) elemTypePool=$(elem_type_pool)" > "$(out_dir)/$(out_file)"
 
+.PHONY: list-gen
+list-gen:
```
prateek: nit: mind sticking to the `genny-` convention

maybe something like `genny-pooled-elem-list`?

src/x/generics/list/list.go
justinjc: I don't see a scenario when someone might want to use a pool for within this list as well as some other place. Sounds like a difficult to find bug waiting to happen.

prateek: some random cases where it could be useful to have the pool be external: the lifecycle of the pool is bigger than the list, or if the pool is shared by multiple lists, etc. essentially if you're ever in a scenario where there are multiple pools being allocated, it's probably worth considering reusing the pool.

justinjc: Fair enough, I'll change the constructor and have it create a default pool if nil is passed in.

PR #1601, opened 2019-05-02
[query] Add start and end time to tag values endpoint

arnikola:
Fixes an issue where the values endpoint, `/label/{name}/values` was not returning correct results. Issue was caused since the new optimized tag completion endpoint now considers query range params, and they were not being set correctly.

```release-note
Fixes a user-facing bug
```

```documentation-note
NONE
```

### DIFF
--- a/src/query/api/v1/handler/prometheus/common.go
+++ b/src/query/api/v1/handler/prometheus/common.go
@@ -241,6 +241,9 @@ func ParseTagValuesToQuery(
 
 	nameBytes := []byte(name)
 	return &storage.CompleteTagsQuery{
+		// NB: necessarily spans the entire timerange for the index.
+		Start:            time.Time{},
+		End:              time.Now(),
 		CompleteNameOnly: false,
 		FilterNameTags:   [][]byte{nameBytes},
 		TagMatchers: models.Matchers{
--- a/src/query/api/v1/handler/prometheus/common.go
+++ b/src/query/api/v1/handler/prometheus/common.go
@@ -37,13 +37,10 @@ import (
 	xhttp "github.com/m3db/m3/src/x/net/http"
 
 	"github.com/golang/snappy"
-	"github.com/gorilla/mux"
 	"github.com/prometheus/prometheus/promql"
 )
 
 const (
-	// NameReplace is the parameter that gets replaced.
-	NameReplace         = "name"
 	queryParam          = "query"
 	filterNameTagsParam = "tag"
 	errFormatStr        = "error parsing param: %s, error: %v"
@@ -52,8 +49,7 @@ const (
 )
 
 var (
-	matchValues = []byte(".*")
-	roleName    = []byte("role")
+	roleName = []byte("role")
 )
 
 // TimeoutOpts stores options related to various timeout configurations.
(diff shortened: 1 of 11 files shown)


src/query/api/v1/handler/prometheus/common.go
```diff
@@ -241,6 +241,9 @@ func ParseTagValuesToQuery(
 
 	nameBytes := []byte(name)
 	return &storage.CompleteTagsQuery{
+		// NB: necessarily spans the entire timerange for the index.
+		Start:            time.Time{},
+		End:              time.Now(),
```
benraskin92: Can you add test to catch this please?

arnikola: It's just setting variables for a struct; since it uses this weird gorilla mux, it's not really plausible to generate a request capable of being read by it easily

benraskin92: hmm but it wasn't working before? so we should probably have a test to make sure it doesn't break in the future?

arnikola: These weren't being set; won't really be a useful test

arnikola: These weren't being set; won't really be a useful test

prateek: hm maybe add an e2e test instead of a unit test, something emulating the behaviour that was failing for users but isn't now?

prateek: like ideally, that test fails without this PR and works with it.

arnikola: Yeah refactored it a bit to make it nicer and added a couple of tests

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -25,6 +25,8 @@ import (
 	"net/http"
 	"time"
 
+	"github.com/m3db/m3/src/x/clock"
+
```
robskillington: nit: Remove superfluous endline.

arnikola: Will do; it's odd that the linter didn't fail here

src/query/api/v1/handler/prometheus/remote/tag_values_test.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/m3db/m3/src/query/storage"
+	"github.com/m3db/m3/src/query/util/logging"
+
+	"github.com/golang/mock/gomock"
+	"github.com/gorilla/mux"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+type tagValuesMatcher struct {
```
prateek: more a FYI for future - can use google's compare method as a matcher for most things - it usually does the right thing: compare struct fields one at a time/use Equal() if it exists on a type/etc. There's a wrapper for it here: https://github.com/m3db/m3/blob/master/src/x/test/matcher.go

[example usage](https://github.com/m3db/m3/blob/eb4f3cb9002b04ceab51b6f1dd99e36729a44e15/src/dbnode/storage/shard_test.go#L211-L216)

PR #1604, opened 2019-05-06
[query] Ignore name tags for matching binary functions

### LINKED ISSUE #1520: Arithmetic binary operators between two vectors return empty result on m3's promql engine
We are using embedded m3coordinator as prometheus datasource on grafana.
query:
```
node_memory_MemTotal{instance=~"$node"} - node_memory_MemAvailable{instance=~"$node"}
```
return the memory used for each node on original prometheus instance.
But it returns empty result on m3db.
We are using 0.8.0 and using the example config file from the repository with only customize on etcd and namespaces fields.

arnikola:
Fixes #1520

Fixes an issue in binary functions; previously there was a requirement that series name tags matched to match series from left and right; this drops name tags from binary functions.

NB: this also drops the name tag from appearing in the output series for binary operations with a scalar argument; this is intentional to mirror the same behaviour as Prometheus 

```release-note
Fixes an issue where binary functions would not match series correctly.
```

```documentation-note
NONE
```

### DIFF
--- a/src/query/functions/binary/binary_test.go
+++ b/src/query/functions/binary/binary_test.go
@@ -837,9 +837,11 @@ func TestBothSeries(t *testing.T) {
 
 			// Extract duped expected metas
 			expectedMeta := block.Metadata{Bounds: bounds}
-			expectedMeta.Tags, tt.expectedMetas = utils.DedupeMetadata(tt.expectedMetas)
+			var expectedMetas []block.SeriesMeta
+			expectedMeta.Tags, expectedMetas = utils.DedupeMetadata(tt.expectedMetas)
+			expectedMeta, expectedMetas = removeNameTags(expectedMeta, expectedMetas)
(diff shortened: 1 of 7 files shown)


src/query/functions/binary/binary_test.go
```diff
@@ -837,9 +837,11 @@ func TestBothSeries(t *testing.T) {
 
 			// Extract duped expected metas
 			expectedMeta := block.Metadata{Bounds: bounds}
-			expectedMeta.Tags, tt.expectedMetas = utils.DedupeMetadata(tt.expectedMetas)
+			var expectedMetas []block.SeriesMeta
```
benraskin92: Can we just add a test here that has an explicit `__name__` tag please?

arnikola: Added one

PR #1612, opened 2019-05-07
[docs] Clarify Prom querying in single_node docs

benraskin92:

This PR clarifies the need to include `__name__` as a tag if you want to query using PromQL.
Fixes #

```release-note

```

```documentation-note
This is a documentation update. 
```

### DIFF
--- a/docs/how_to/single_node.md
+++ b/docs/how_to/single_node.md
@@ -59,7 +59,7 @@ curl -sS -X POST http://localhost:9003/writetagged -d '{
   "tags": [
     {
       "name": "__name__",
-      "value": "user_sign_up"
+      "value": "user_login"
     },
     {
       "name": "city",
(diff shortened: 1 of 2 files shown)


docs/how_to/single_node.md
```diff
@@ -93,6 +106,10 @@ curl -sS -X POST http://localhost:9003/query -d '{
     {
       "id": "foo",
       "tags": [
+        {
+          "name": "__name__",
```
arnikola: nit: worth adding a curl against coordinator as well for posterity? Up to you

docs/how_to/single_node.md
```diff
@@ -57,6 +57,10 @@ curl -sS -X POST http://localhost:9003/writetagged -d '{
   "namespace": "default",
   "id": "foo",
   "tags": [
+    {
+      "name": "__name__",
+      "value": "user_sign_up"
```
robskillington: This is a good example, however the best argument for leveraging metrics over logs is for high volume events (rather than less frequent events, which probably don't cost that much to store anyway as logs).

With that in mind, perhaps we should slightly tweak the example.

Maybe something like "user_login"?

robskillington:
Maybe we should add this example to the repo's `README.md` as well?

PR #1613, opened 2019-05-08
[dbnode] Add forward index write capability

arnikola:
This PR eases index pressure at block boundaries by allowing incoming writes near the boundary to have a chance to be dual written to the upcoming index block based on config defined probabilities.

Ensure that there's no further changes to timestamp required.

```release-note
NONE
```

```documentation-note
NONE
```


src/dbnode/storage/forward_index_dice.go
```diff
@@ -0,0 +1,101 @@
...
+import (
+	"fmt"
+	"time"
+
+	"github.com/m3db/m3/src/x/dice"
+)
+
+// forwardIndexDice is a die roll that adds a chance for incoming index writes
+// arriving near a block boundary to be duplicated and written to the next block
+// index, adding jitter and smoothing index load.
```
richardartoul: "and smoothing index load so that block boundaries do not cause a huge influx of new documented that need to be indexed all at once"

src/dbnode/storage/forward_index_dice.go
```diff
@@ -0,0 +1,101 @@
...
+		threshold = indexOpts.ForwardIndexThreshold()
+
+		retention    = seriesOpts.RetentionOptions()
+		bufferFuture = retention.BufferFuture()
+		blockSize    = retention.BlockSize()
+
+		forwardIndexThreshold time.Duration
+	)
+
+	if threshold > 1 {
```
richardartoul: maybe one for <0 too? Probably means a bug

src/dbnode/storage/forward_index_dice.go
```diff
@@ -0,0 +1,101 @@
...
+		return forwardIndexDice{},
+			fmt.Errorf("invalid forward write threshold %f", threshold)
+	}
+
+	bufferFragment := float64(bufferFuture) * threshold
```
richardartoul: I don't understand why there is a relationship between bufferFuture and the threshold...I guess in my mind bufferFuture would set a maximum on the threshold so if bufferFuture is 10 minutes and you have a block size of 2 hours then the earliest you could start forward indexing would be at 1:50PM, but I don't understand the logic here where you multiply bufferFuture by some threshold

src/dbnode/storage/forward_index_dice_test.go
```diff
@@ -0,0 +1,149 @@
...
+	start := time.Now().Truncate(time.Hour)
+	end := start.Add(time.Hour)
+
+	for ts := start; ts.Before(end); ts = ts.Add(time.Second) {
+		require.False(t, dice.forwardIndexing(ts))
+	}
+}
+
+func TestInvalidForwardIndexDice(t *testing.T) {
+	opts := optionsWithIndexValues(20, 0, nil)
```
richardartoul: Comments for why these are invalid

src/x/dice/dice.go
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
```
richardartoul: I coulda sworn we already had a dice package that we use in a few different places. Is this just coming from m3x?

src/x/dice/dice.go
```diff
@@ -0,0 +1,61 @@
...
+	r      int64
+	jitter int64
+}
+
+func (d *epoch) Rate() float64 {
+	return 1 / float64(d.r)
+}
+
+func (d *epoch) Roll() bool {
+	return (time.Now().UnixNano()+d.jitter)%d.r == 0
```
richardartoul: out of curiosity, why time over a rand generator? Thread safety?

src/x/dice/dice.go
```diff
@@ -0,0 +1,61 @@
...
+// Dice is an interface that allows for random sampling.
+type Dice interface {
+	// Rate returns the sampling rate of this Dice: a number in (0.0, 1.0].
+	Rate() float64
+
+	// Roll returns whether the dice roll succeeded.
+	Roll() bool
+}
+
+// NewEpochDice constructs a new Dice based on UNIX epoch time.
```
richardartoul: Can you clarify this comment to explain that it uses the current time as a source of randomness

src/dbnode/storage/index.go
```diff
@@ -524,8 +539,15 @@ func (i *nsIndex) writeBatches(
 				batch.MarkUnmarkedEntryError(m3dberrors.ErrTooPast, idx)
 				return
 			}
+
+			if i.forwardIndexDice.forwardIndexing(entry.Timestamp) {
+				forwardWriteEntry := entry
```
richardartoul: Is this a safe clone? There are three fields that are "pointer" type: 

1) OnIndexSeries
2) result
3) resultVal

Regardless I wonder if its worth adding a clone() method so that even if this operation is safe now its more likely to stay safe in the future

src/dbnode/storage/index.go
```diff
@@ -524,8 +539,15 @@ func (i *nsIndex) writeBatches(
...
+				forwardWriteEntry.Timestamp = entry.Timestamp.Add(blockSize)
```
richardartoul: Is the timestamp used for anything other than figuring out which block to put it in? Also, I may be wrong but my understanding of the forward indexing logic seemed to imply that you could forward index stuff that wasn't yet within the bufferFuture window. Is that correct? If so, is it guaranteed there is a block available for this write to get indexed into?

src/dbnode/storage/forward_index_dice.go
arnikola: Yeah, that's fair; I was relying on the `dice.NewEpochDice` to fail here but probably better to be more explicit here about it

src/dbnode/storage/forward_index_dice.go
arnikola: Yeah, so threshold is defined as a fragment of the bufferFuture to allow some tweaking on when ; my thinking is you'd put the `threshold` at 0.9 or something, and then in your example, we'd start forward indexing at `1:51 PM`, just to cut down on any possibility of trying to write the values too early in case of positive skew on the emitter.

i.e. `bufferFragment ∈ (0, bufferFuture]`

src/x/dice/dice.go
arnikola: This is from internal statsdex; I think there's a sampler package we have already, but it performs slightly differently (i.e. `NewSampler(0.1).Sample()` returns true IFF it's been called a multiple of 10 times)

src/dbnode/storage/index.go
arnikola: Given the current behaviour, it will only forward index within a percentage of the `bufferFuture` window

src/x/dice/dice.go
prateek: @arnikola could we use [pcg](github.com/MichaelTJones/pcg) (similar to how we do [here](https://github.com/m3db/m3/blob/master/src%2Fx%2Fsync%2Fpooled_worker_pool.go#L129)) instead of `time.Now()`. Will still seed with time.Now() so won't affect randomness but will be a lot quicker (7ns v 50ns).

src/dbnode/storage/index.go
```diff
@@ -524,8 +539,15 @@ func (i *nsIndex) writeBatches(
...
+				forwardWriteBatch.Append(forwardWriteEntry, d)
+			}
 		})
 
+	batch.AppendAll(forwardWriteBatch)
```
prateek: Unfortunately this isn't safe as is. When trying to index a Series, we track the last successful block starts when it was indexed (in *lookup.Entry via the shard map); and further, maintain a reference count to Series entry. Both of these are updated via  the `OnIndexSeries` per element in the WriteBatchEntry.

By just appending entries the way it's being done right now, it's going to definitely mess up the reference count (because you'll dec for each copy, even though we only have a inc on the original). 

Won't necessarily mess up the last successful write times but probably need to consider some nuances there too.

prateek: Re: reference counting - we can either inc the appropriate references when making copies (safest), or ensure we only dec the final copy amongst all the ones we make here (can't be the original, has to be last one we track because of cleanup semantics in the Shard). 

Re: tracking last successful write time update. Here's a [good explanation](https://github.com/m3db/m3/blob/master/src%2Fdbnode%2Fstorage%2Fseries%2Flookup%2Fentry.go#L145-L153). 

AFAICT, the changes you're introducing can make it so that we're writing to 4 block starts (instead of the 3 we currently assume worst case). I'm not a 100% sure that's accurate tho. Should work that out, and if so it makes sense to bump the size of the static alloc from 3 to 4.

src/x/dice/dice.go
```diff
@@ -0,0 +1,59 @@
...
+	r      int64
+	jitter int64
+}
+
+func (d *epoch) Rate() float64 {
+	return 1 / float64(d.r)
+}
+
+func (d *epoch) Roll() bool {
+	return rand.Int63()%d.r == 0
```
prateek: Avoid using the global rand, it's a locked source: https://golang.org/src/math/rand/rand.go?s=9470:9488#L288. Better to control whatever concurrency you need by yourself.

Could you also use the github.com/MichaelTJones/pcg RNG. It's a lot quicker/cheaper than the standard one.

src/dbnode/storage/forward_index_dice.go
prateek: mind adding an explanation for this in a comment

src/dbnode/storage/forward_index_dice.go
```diff
@@ -0,0 +1,102 @@
...
+		enabled:   true,
+		blockSize: blockSize,
+
+		forwardIndexThreshold: forwardIndexThreshold,
+		forwardIndexDice:      dice,
+	}, nil
+}
+
+// forwardIndexing decides if a timestamp is eligible for forward index writes.
+func (o *forwardIndexDice) forwardIndexing(timestamp time.Time) bool {
```
prateek: nit: rename to something like `shouldIndex` or `roll` to indicate it's a test. the current method name doesn't convey what you're testing for

src/dbnode/storage/index.go
```diff
@@ -524,8 +539,23 @@ func (i *nsIndex) writeBatches(
 				batch.MarkUnmarkedEntryError(m3dberrors.ErrTooPast, idx)
 				return
 			}
+
+			if i.forwardIndexDice.forwardIndexing(entry.Timestamp) {
```
prateek: once you fix the global rand usage, will need to be careful about using this dice

prateek: if we already have a lock that we can use to protect the dice, than it's gucci. if not, will have to think about the contention it'll introduce to add a lock here.

arnikola: I'll see if there's a way to avoid having to get a lock here; there's a lock above batch processing I could `roll` in, but that would mean that the forward write would be applied to the whole batch. If that's fine, should be easy :+1:

src/dbnode/storage/forward_index_dice.go
arnikola: SGTM, changed to `roll` to preserve dice context

src/dbnode/storage/index/block_bench_test.go
```diff
@@ -123,3 +124,10 @@ func (m mockOnIndexSeries) OnIndexSuccess(blockStart xtime.UnixNano) {
 func (m mockOnIndexSeries) OnIndexFinalize(blockStart xtime.UnixNano) {
 
 }
+func (m mockOnIndexSeries) OnIndexPrepare() {
+
+}
+
+func (m mockOnIndexSeries) NeedsIndexUpdate(indexBlockStartForWrite xtime.UnixNano) bool {
```
prateek: can you just return false always? doesn't seem like you're updating the value ever

arnikola: Yeah fair, had it like this in case we wanted to be able to return different values in the future

src/dbnode/storage/index_queue_forward_write_test.go
```diff
@@ -0,0 +1,565 @@
...
+	defer leaktest.CheckTimeout(t, 2*time.Second)()
+
+	ctx := context.NewContext()
+	idx, now, blockSize := setupForwardIndex(t, ctrl)
+	defer idx.Close()
+
+	reQuery, err := m3ninxidx.NewRegexpQuery([]byte("name"), []byte("val.*"))
+	assert.NoError(t, err)
+
+	// NB: query both the current and the next index block to ensure that the
```
prateek: I don't think this is confirming that the write exists in the next block -- cause you're querying both current and next block. Need to ensure you're checking both individually, no?

src/dbnode/storage/index_queue_forward_write_test.go
```diff
@@ -0,0 +1,565 @@
...
```
prateek: here too?

src/dbnode/storage/index_queue_forward_write_test.go
```diff
@@ -0,0 +1,565 @@
...
+		ts   = idx.(*nsIndex).state.latestBlock.StartTime()
+		next = ts.Truncate(blockSize).Add(blockSize)
+		id   = ident.StringID("foo")
+		tags = ident.NewTags(
+			ident.StringTag("name", "value"),
+		)
+		lifecycle = index.NewMockOnIndexSeries(ctrl)
+	)
+
+	lifecycle.EXPECT().OnIndexFinalize(xtime.ToUnixNano(ts))
```
prateek: super nit: mind putting these in `gomock.InOrder(...)` as much as possible

src/dbnode/storage/index_queue_forward_write_test.go
arnikola: It should be fine, `now` is 5 minutes before the block boundary, and this will check the range which is from 6 to 4 minutes before the boundary to check for the current block, and then 6 to 4 minutes before the next block boundary to check only the future block

src/dbnode/storage/index_queue_forward_write_test.go
arnikola: Should be same here

src/dbnode/storage/index.go
```diff
@@ -501,31 +513,60 @@ func (i *nsIndex) writeBatches(
...
 				return
 			}
 
-			if !entry.Timestamp.After(pastLimit) {
+			if !ts.After(pastLimit) {
 				batch.MarkUnmarkedEntryError(m3dberrors.ErrTooPast, idx)
 				return
 			}
+
+			if forwardIndexDice.roll(ts) {
```
prateek: should this also be `if forwardIndexEnabled && ..` 

maybe add a test that would have caught this

arnikola: The `!enabled` case should be caught here: https://github.com/m3db/m3/pull/1613/files/70b0912726891694461e47b7d21499aeddb8bbca#diff-b5380446b59ebd51b8b8a482df15c4d7R93

PR #1617, opened 2019-05-08
[query] Fix step parsing

### LINKED ISSUE #1610: M3 can't parse query_range with decimal step parameter (without unit)
Grafana alerts seem to generate query_range requests with a decimal step query string parameter.
```
http://localhost/api/v1/query_range?end=2019-05-07T11:49:40.790434054Z&query=sum(foobar{app=%22testapp%22,%20stage=%22prod%22,})%20by%20(foobar)&start=2019-05-07T11:44:40.790434054Z&step=60.000
```
results in a 400 error.

The following requests work as expected:
```
http://localhost/api/v1/query_range?end=2019-05-07T11:49:40.790434054Z&query=sum(foobar{app=%22testapp%22,%20stage=%22prod%22,})%20by%20(foobar)&start=2019-05-07T11:44:40.790434054Z&step=60
```
```
http://localhost/api/v1/query_range?end=2019-05-07T11:49:40.790434054Z&query=sum(foobar{app=%22testapp%22,%20stage=%22prod%22,})%20by%20(foobar)&start=2019-05-07T11:44:40.790434054Z&step=60.000s
```
As a work around, we added the following rewrite in our fronted nginx: 
```
      if ($args ~* (.*)(step=[0-9]+)\.000(.*)){
        set $args $1$2$3;
      }
```
The prometheus API specs seem to allow floats:
```
step=<duration | float>
```
We're using v0.8.4

benraskin92:

Fixes #1610 

Previously, if you included a float step size with no units (e.g. 15.00), the query would fail with 
```
"error":"error parsing param: step, error: time: missing unit in duration 15.00"
```

This PR allows you to add a float step size without units (and assumes it is in seconds -- equivalent to what Prometheus does) 

```release-note

```

```documentation-note
Maybe we can add something in the docs, but I don't think it's necessary. 
```

### DIFF
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -120,7 +120,7 @@ func parseParams(r *http.Request, timeoutOpts *prometheus.TimeoutOpts) (models.R
 		return params, xhttp.NewParseError(fmt.Errorf(formatErrStr, stepParam, err), http.StatusBadRequest)
 	}
 	if step <= 0 {
-		err := fmt.Errorf("expected postive step, instead got: %d", step)
+		err := fmt.Errorf("expected postive step size, instead got: %d", step)
 		return params, xhttp.NewParseError(fmt.Errorf(formatErrStr, stepParam, err), http.StatusBadRequest)
 	}
 	params.Step = step
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -79,7 +79,7 @@ func parseDuration(r *http.Request, key string) (time.Duration, error) {
 	if seconds, floatErr := strconv.ParseFloat(str, 64); floatErr == nil {
 		ts := seconds * float64(time.Second)
 		if ts > float64(math.MaxInt64) || ts < float64(math.MinInt64) {
-			return 0, fmt.Errorf("cannot parse %q to a valid duration. It overflows int64", str)
+			return 0, fmt.Errorf("cannot parse %s to a valid duration: int64 overflow", str)
 		}
 
 		return time.Duration(ts), nil
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -75,8 +75,8 @@ func parseDuration(r *http.Request, key string) (time.Duration, error) {
 		return value, nil
 	}
 
-	// Try parsing as an integer value specifying seconds, the Prometheus default
-	if seconds, intErr := strconv.ParseFloat(str, 64); intErr == nil {
+	// Try parsing as a float value specifying seconds, the Prometheus default
+	if seconds, floatErr := strconv.ParseFloat(str, 64); floatErr == nil {
 		ts := seconds * float64(time.Second)
 		if ts > float64(math.MaxInt64) || ts < float64(math.MinInt64) {
 			return 0, fmt.Errorf("cannot parse %q to a valid duration. It overflows int64", str)
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -119,6 +119,10 @@ func parseParams(r *http.Request, timeoutOpts *prometheus.TimeoutOpts) (models.R
 	if err != nil {
 		return params, xhttp.NewParseError(fmt.Errorf(formatErrStr, stepParam, err), http.StatusBadRequest)
 	}
+	if step <= 0 {
+		err := fmt.Errorf("expected postive step, instead got: %d", step)
+		return params, xhttp.NewParseError(fmt.Errorf(formatErrStr, stepParam, err), http.StatusBadRequest)
+	}
 	params.Step = step
 
 	query, err := parseQuery(r)
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -53,6 +53,9 @@ const (
 	blockTypeParam    = "block-type"
 
 	formatErrStr = "error parsing param: %s, error: %v"
+
+	maxInt64 = float64(math.MaxInt64)
+	minInt64 = float64(math.MinInt64)
 )
 
 func parseTime(r *http.Request, key string) (time.Time, error) {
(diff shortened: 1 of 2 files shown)


src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -76,8 +76,13 @@ func parseDuration(r *http.Request, key string) (time.Duration, error) {
 	}
 
 	// Try parsing as an integer value specifying seconds, the Prometheus default
```
martin-mao: Nit: Update comment.

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -76,8 +76,13 @@ func parseDuration(r *http.Request, key string) (time.Duration, error) {
...
-	if seconds, intErr := strconv.ParseInt(str, 10, 64); intErr == nil {
-		return time.Duration(seconds) * time.Second, nil
+	if seconds, intErr := strconv.ParseFloat(str, 64); intErr == nil {
+		ts := seconds * float64(time.Second)
+		if ts > float64(math.MaxInt64) || ts < float64(math.MinInt64) {
```
martin-mao: Is it valid for duration to be negative in the prom context? If not, maybe check that it's positive instead of smaller than minInt64?

src/query/api/v1/handler/prometheus/native/common_test.go
```diff
@@ -112,6 +112,14 @@ func TestParseDuration(t *testing.T) {
 	assert.Equal(t, 10*time.Second, v)
 }
 
+func TestParseFloatDuration(t *testing.T) {
+	r, err := http.NewRequest(http.MethodGet, "/foo?step=10.00m", nil)
```
martin-mao: Prob want a test where the step is not just a round number? eg Test to ensure that /foo?step=10.85m return the expected output, whether that's the precise duration, a rounded duration or an error.

src/query/api/v1/handler/prometheus/native/common.go
benraskin92: Ha beat you to it :)

src/query/api/v1/handler/prometheus/native/common.go
martin-mao: Also, it might be worth storing the float version of maxInt64 as a const so you don't need to cast each time.

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -75,9 +75,14 @@ func parseDuration(r *http.Request, key string) (time.Duration, error) {
...
 	}
 
-	// Try parsing as an integer value specifying seconds, the Prometheus default
-	if seconds, intErr := strconv.ParseInt(str, 10, 64); intErr == nil {
-		return time.Duration(seconds) * time.Second, nil
+	// Try parsing as a float value specifying seconds, the Prometheus default
+	if seconds, floatErr := strconv.ParseFloat(str, 64); floatErr == nil {
+		ts := seconds * float64(time.Second)
+		if ts > float64(math.MaxInt64) || ts < float64(math.MinInt64) {
+			return 0, fmt.Errorf("cannot parse %q to a valid duration. It overflows int64", str)
```
mway: nit: `cannot parse %q to a valid duration: int64 overflow` or similar (errors aren't sentences)

src/query/api/v1/handler/prometheus/native/common.go
benraskin92: This is what Prometheus checks, so tempted to just leave it. 

https://github.com/prometheus/prometheus/blob/master/web/api/v1/api.go#L1171

benraskin92: > Also, it might be worth storing the float version of maxInt64 as a const so you don't need to cast each time.

Sounds good.

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -75,9 +78,14 @@ func parseDuration(r *http.Request, key string) (time.Duration, error) {
 		return value, nil
 	}
 
-	// Try parsing as an integer value specifying seconds, the Prometheus default
-	if seconds, intErr := strconv.ParseInt(str, 10, 64); intErr == nil {
-		return time.Duration(seconds) * time.Second, nil
+	// Try parsing as a float value specifying seconds, the Prometheus default
+	if seconds, floatErr := strconv.ParseFloat(str, 64); floatErr == nil {
+		ts := seconds * float64(time.Second)
+		if ts > maxInt64 || ts < minInt64 {
```
martin-mao: Do you need to validate for negative step size in Prom? If so, not sure if you should do it here or somewhere else.

benraskin92: Cool - just added to the parseParams() function - similar to what prom does.

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -75,9 +78,14 @@ func parseDuration(r *http.Request, key string) (time.Duration, error) {
 		return value, nil
 	}
 
-	// Try parsing as an integer value specifying seconds, the Prometheus default
-	if seconds, intErr := strconv.ParseInt(str, 10, 64); intErr == nil {
-		return time.Duration(seconds) * time.Second, nil
+	// Try parsing as a float value specifying seconds, the Prometheus default
+	if seconds, floatErr := strconv.ParseFloat(str, 64); floatErr == nil {
```
arnikola: I think there's a bug here, you're not returning `floatErr` here if it's not nil

also, nit: when you have a decent size codeblock I think it might read a little nicer to flip the conditional and then split this up, i.e.

```golang
seconds, err := strconv.ParseFloat(str, 64)
if err != nil {
 return nil, err
}
// logic
```

src/query/api/v1/handler/prometheus/native/common_test.go
```diff
@@ -120,13 +146,28 @@ func TestParseDurationParsesIntAsSeconds(t *testing.T) {
 	assert.Equal(t, 30*time.Second, v)
 }
 
+func TestParseDurationParsesFloatAsSeconds(t *testing.T) {
```
arnikola: Could you add a case for `"/foo?step=30"`?

benraskin92: There is one already.

src/query/api/v1/handler/prometheus/native/common.go
benraskin92: Eh i mean it only gets to here if this returns an error:
```
	value, err := time.ParseDuration(str)
```
So either way it'll return an error - just might not be the float conversion error.

arnikola: That's wrong. It needs to return this error

PR #1618, opened 2019-05-09
Add benchmark for creating empty filesets

richardartoul:

Adds a benchmark for measuring the performance of creating many empty filesets

### DIFF
--- a/src/dbnode/persist/fs/writer_benchmark_test.go
+++ b/src/dbnode/persist/fs/writer_benchmark_test.go
@@ -32,6 +32,46 @@ import (
 	xsync "github.com/m3db/m3/src/x/sync"
 )
 
+// Benchmarks run on a production machine with 32 cores and non server-grade, non NVME SSD drives.
+// goos: linux
+// goarch: amd64
+// pkg: github.com/m3db/m3/src/dbnode/persist/fs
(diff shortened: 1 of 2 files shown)


src/dbnode/persist/fs/writer_benchmark_test.go
```diff
@@ -0,0 +1,155 @@
...
+	"runtime"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/m3db/m3/src/dbnode/persist"
+	xsync "github.com/m3db/m3/src/x/sync"
+)
+
+func BenchmarkCreateEmptyFilesets(b *testing.B) {
```
prateek: Two nits:
(1) run this on a production spare and include results in comments
(2) model parallelism increases as powers of two (like cpu/8, cpu/4, ...)

src/dbnode/persist/fs/writer_benchmark_test.go
```diff
@@ -0,0 +1,122 @@
...
+	for i := 0; i < b.N; i++ {
+		wg.Add(1)
+
+		workerPool.Go(func() {
+			writerOpts := DataWriterOpenOptions{
+				Identifier: FileSetFileIdentifier{
+					Namespace:   testNs1ID,
+					Shard:       uint32(i % numShards),
+					BlockStart:  start,
+					VolumeIndex: i,
```
prateek: need to capture this var outside the closure.

PR #1621, opened 2019-05-09
Enable panics if invariant violations occur in integration tests

richardartoul:
Improves ability to catch invariant violations during integration tests


robskillington:
Closing in favor of https://github.com/m3db/m3/pull/1622 which fixes some calls to invariantf from package variables (not a good idea).

PR #1624, opened 2019-05-09
Cold flushes

justinjc:

This PR introduces cold flush logic that merges out of order writes with data on disk.

Sorry, this PR is rather large. A good place to start would be the `ColdFlush` function in `shard.go`, where it identifies which blockStarts require cold flushes, then uses the merger to merge data together. As a natural follow up, look into `merger.go` to review actual merge logic, and `fs_merge_with_mem.go` to review the logic for going through data in memory.

*Please note* that this PR is just one part of implementing the ColdWrites feature. Landing of this PR does not enable the ability to do cold writes without other necessary changes, e.g. in the seek manager. As such, most of the code here is feature flagged off by default. This flag is at the namespace level (in the option `ColdWritesEnabled`). The call to `ColdFlush` in `namespace.go` checks for this option, and returns success right away if the flag is false, hence, any logic below that (shard/series/buffer/actual merging) will never get run.

```release-note
NONE
```

```documentation-note
NONE
```


src/dbnode/storage/shard.go
```diff
@@ -1897,7 +1891,254 @@ func (s *dbShard) Flush(
...
+				if len(encoded) > 0 {
+					brs = append(brs, encoded)
+				}
+			}
+
+			mergedIter := multiIterPool.Get()
+			mergedIter.ResetSliceOfSlices(xio.NewReaderSliceOfSlicesFromBlockReadersIterator(brs))
+			defer mergedIter.Close()
+
+			tags, err := convert.TagsFromTagsIter(id, tagsIter, s.opts.IdentifierPool())
```
richardartoul: Can you look into the lifecycle of the tags here? Will the persistIter finalize them or do we need to do that manually after that function is called

src/dbnode/storage/shard.go
```diff
@@ -1897,7 +1891,254 @@ func (s *dbShard) Flush(
...
+				multiErr = multiErr.Add(err)
+				continue BlockLoop
+			}
+		}
+
+		// Second stage: loop through rest of series in memory that was not on
+		// disk.
+		for seriesElement := seriesList.Front(); seriesElement != nil; seriesElement = seriesElement.Next() {
+			s.RLock()
+			entry, _, err := s.lookupEntryWithLock(seriesElement.Value)
```
richardartoul: Is this safe? I guess the series could not have been evicted in the meantime

src/dbnode/storage/shard.go
```diff
@@ -1897,7 +1891,254 @@ func (s *dbShard) Flush(
...
+	blockSize := s.namespace.Options().RetentionOptions().BlockSize()
+	srPool := s.opts.SegmentReaderPool()
+	multiIterPool := s.opts.MultiReaderIteratorPool()
+	encoderPool := s.opts.EncoderPool()
+	tmpCtx := context.NewContext()
+
+	// Loop through each block that we know has ColdWrites. Since each block
+	// has its own fileset, if we encounter an error while trying to persist
+	// a block, we continue to try persisting other blocks.
+BlockLoop:
```
richardartoul: We should think about how to encapsulate this logic into a reusable way. Ideally, we could move most of this logic out into some kind of `FilesetMerger` or something.

The crux of the algorithm seems to be:

1. Loop through series on disk and check if they're in-memory. If in memory, merge, persist, then remove from in-memory lookup. If not in memory, just write them to disk.
2. Loop through series in memory lookup (which now has everything removed that was already merged / persisted to disk) and write that out to disk.

I think we'd need to discuss the interface a bit more but you can imagine something like:

```golang
type FsMergerOptions struct {
    Namespace ns.ID
    Shard uint32
    BlockStart time.Time
}

type FsMergeWith interface {
    Contains(seriesID ident.ID) (data checked.Bytes, bool)
    MarkMerged(seriesID ident.ID)
    ForEachRemaining(fn forEachRemainingFn) 
}

func (m *fsMerger) Merge(opts FsMergerOptions, with FsMergeWith) error {
   ...
}
```

I bet we could come up with a better interface, but I think you get the idea

richardartoul: Then that thing could be tested fairly aggressively in isolation and this method would basically just boil down to calling the `FsMerger` with the right arguments.

src/dbnode/storage/namespace.go
```diff
@@ -924,20 +926,90 @@ func (n *dbNamespace) Flush(
...
+
+type coldFlushReuseableResources struct {
+	// dirtySeries is a map from a composite key of <series ID, block start>
+	// to an element in a list in the dirtySeriesToWrite map. This composite key
+	// is deliberately made so that this map stays one level deep, making it
+	// easier to share between shard loops, minimizing the need for allocations.
+	dirtySeries *dirtySeriesMap
+	// dirtySeriesToWrite is a map from block start to a list of dirty series
+	// that have yet to be written to disk.
+	dirtySeriesToWrite map[xtime.UnixNano]*idList
```
richardartoul: Can you remind me why we need this list? Seem to me like it could just be `map[xtime.UnixNano]*dirtySeriesMap` since a map can be used for both checking and iteration. Maybe you'll need to change `idAndBlockStart` to have a slice of blockStart but still seems fine.

I'm sure I'm missing something though

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -768,8 +768,7 @@ func (s *service) FetchBlocksRaw(tctx thrift.Context, req *rpc.FetchBlocksRawReq
 	// Preallocate starts to maximum size since at least one element will likely
 	// be fetching most blocks for peer bootstrapping
 	ropts := nsMetadata.Options().RetentionOptions()
-	blockStarts := make([]time.Time, 0,
-		(ropts.RetentionPeriod()+ropts.FutureRetentionPeriod())/ropts.BlockSize())
+	blockStarts := make([]time.Time, 0, ropts.RetentionPeriod()/ropts.BlockSize())
```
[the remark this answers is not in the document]
justinjc: I was thinking that since we won't be supporting future writes for now, we should remove this for now. Although I support `FutureRetentionPeriod` would be 0 and so this should be a no op. I'll keep this in.

src/dbnode/retention/times.go
```diff
@@ -34,8 +34,7 @@ func FlushTimeStartForRetentionPeriod(retentionPeriod time.Duration, blockSize t
 
 // FlushTimeEnd is the latest flushable time
 func FlushTimeEnd(opts Options, t time.Time) time.Time {
-	return FlushTimeEndForBlockSize(opts.BlockSize(),
-		t.Add(opts.FutureRetentionPeriod()).Add(-opts.BufferPast()))
+	return FlushTimeEndForBlockSize(opts.BlockSize(), t.Add(-opts.BufferPast()))
```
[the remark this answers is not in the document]
justinjc: Same as above.

src/dbnode/storage/flush.go
```diff
@@ -102,28 +105,58 @@ func (m *flushManager) Flush(
...
+	// process will attempt to snapshot any unflushed blocks which would be
+	// wasteful if the block is already flushable.
+	multiErr := xerrors.NewMultiError()
+	if err = m.dataWarmFlush(namespaces, tickStart, dbBootstrapStateAtTickStart); err != nil {
+		multiErr = multiErr.Add(err)
+	}
+
+	rotatedCommitlogID, err := m.commitlog.RotateLogs()
+	if err == nil {
+		if err = m.dataColdFlush(namespaces); err != nil {
```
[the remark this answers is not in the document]
justinjc: Yeah cold flush is dependent on commitlog rotation because snapshots only capture warm writes. Both a cold flush and snapshot is required to capture all data in the rotated commitlog.

src/dbnode/storage/shard.go
```diff
@@ -342,7 +337,7 @@ func (s *dbShard) Stream(
 // IsBlockRetrievable implements series.QueryableBlockRetriever
 func (s *dbShard) IsBlockRetrievable(blockStart time.Time) bool {
 	flushState := s.FlushState(blockStart)
-	return statusIsRetrievable(flushState.Status)
+	return statusIsRetrievable(flushState.WarmStatus) || flushState.ColdVersion > 0
```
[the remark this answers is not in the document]
justinjc: This would only happen in the case of future writes.

src/dbnode/storage/shard.go
```diff
@@ -1897,7 +1891,254 @@ func (s *dbShard) Flush(
...
+	return s.markWarmFlushStateSuccessOrError(blockStart, multiErr.FinalError())
+}
+
+func (s *dbShard) ColdFlush(
+	flushPreparer persist.FlushPreparer,
+	resources coldFlushReuseableResources,
+) error {
+	// We don't flush data when the shard is still bootstrapping.
+	s.RLock()
+	if s.bootstrapState != Bootstrapped {
```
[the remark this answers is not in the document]
justinjc: Sure: https://github.com/m3db/m3/issues/1632

src/dbnode/storage/shard.go
```diff
@@ -1897,7 +1891,254 @@ func (s *dbShard) Flush(
...
+	// Reset resources.
+	for _, series := range dirtySeriesToWrite {
+		if series != nil {
+			series.Reset()
+		}
+		// Don't delete the empty list from the map so that other shards don't
+		// need to reinitialize the list for these blocks.
+	}
+	dirtySeries.Reset()
+	if fsReader.Status().Open {
```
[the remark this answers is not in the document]
justinjc: It really shouldn't, I'm just being paranoid here. I'll make it an invariant violation.

src/dbnode/storage/shard.go
```diff
@@ -1897,7 +1891,254 @@ func (s *dbShard) Flush(
...
+
+	var multiErr xerrors.MultiError
+	// First, loop through all series to capture data on which blocks have dirty
+	// series and add them to the resources for further processing.
+	s.forEachShardEntry(func(entry *lookup.Entry) bool {
+		curr := entry.Series
+		seriesID := curr.ID()
+		blockStarts := curr.NeedsColdFlushBlockStarts()
+		blockStarts.ForEach(func(t xtime.UnixNano) {
+			seriesList := dirtySeriesToWrite[t]
```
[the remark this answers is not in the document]
justinjc: Good idea.

src/dbnode/storage/shard.go
justinjc: Yeah I don't think it's possible that the entry would become invalid within the lifetime of this function.

src/dbnode/storage/namespace.go
justinjc: `dirtySeriesMap` needs to have both series ID and block start because we need to know if a series ID for a block is dirty. We also need to iterate through the dirty series by block start and we can't do that efficiently with that key.

It would be cleanest to just use `map[xtime.UnixNano]map[series ID]bool`, but we didn't want a multi-level map to avoid having to keep allocating when inserting into the series ID map.

richardartoul: Why would the mutli-level map require more allocations? the seriesID map could be pooled as well. Plus you can use the unsafe feature so that the series IDs dont need to be allocated since they're immutable anyways

justinjc: Spoke offline - having a map pool shared between cold flushes would work, but it's likely that maps would have varying degrees of usage, so it's possible that a small map is taken to do a lot of writes, which would not be optimal. Having a flat data structure makes pooling more uniform.

However, this means that we are creating lots of small objects (list elements) every cold flush. Let's keep this implementation for now and see if it becomes an issue.

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,246 @@
...
+			multiErr = multiErr.Add(err)
+			return multiErr.FinalError()
+		}
+
+		// There will be at most two BlockReader slices: one for disk data
+		// and one for in memory data.
+		brs := make([][]xio.BlockReader, 0, 2)
+
+		// Create BlockReader slice out of disk data.
+		br := blockReaderFromData(data, srPool, startTime, blockSize)
```
richardartoul: You get a segment reader from the pool each time this function is called, but do you ever return it? If not you should probably do that.

The more efficient alternative would be to just reuse the same two segment readers over and over again (they have a `Reset()` method)

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,246 @@
...
+			return multiErr.FinalError()
+		}
+
+		if hasData {
+			brs = append(brs, encoded)
+		}
+
+		mergedIter := multiIterPool.Get()
+		mergedIter.ResetSliceOfSlices(xio.NewReaderSliceOfSlicesFromBlockReadersIterator(brs), nsCtx.Schema)
+		defer mergedIter.Close()
```
richardartoul: You're defering in a for loop which is gonna allocate like crazy (each defer basically allocates a function on the heap to run later so if you iterate through 40k series this is 40k function allocations).

You can fix the pooling or just allocate this outside the for loop and reset it between each loop (most efficient)

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,246 @@
...
+
+		tags, err := convert.TagsFromTagsIter(id, tagsIter, identPool)
```
richardartoul: This is kind of expensive but don't see a better way to do it unless the MergeWith can provide the tags but not sure we want to assume that will always be possible

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,246 @@
...
+		tagsIter.Close()
```
richardartoul: do you need a nil check here or is this always valid even if there are no tags?

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,246 @@
...
+		if err := persistIter(prepared.Persist, mergedIter, id, tags, encoderPool); err != nil {
+			multiErr = multiErr.Add(err)
+			return multiErr.FinalError()
+		}
+	}
+
+	// Second stage: loop through rest of the merge target that was not captured
+	// in the first stage.
+	err = mergeWith.ForEachRemaining(blockStart, func(seriesID ident.ID, tags ident.Tags) bool {
+		encoded, hasData, err := mergeWith.Read(seriesID, blockStart, nsCtx)
```
richardartoul: This works although I wonder if the data should just be included and passed to the function so that you don't have to call into `mergeWith` again (which may have to look up the same data it had access to a moment again when it called this function)

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,246 @@
...
+		if err != nil {
+			multiErr = multiErr.Add(err)
+			return false
+		}
+
+		if hasData {
+			iter := multiIterPool.Get()
+			iter.Reset(xio.NewSegmentReaderSliceFromBlockReaderSlice(encoded), startTime, blockSize, nsCtx.Schema)
+			defer iter.Close()
```
richardartoul: defer is actually safe here (since its not in an explicit for loop), but same thing applies. This is all single threader so you could just have one you allocate before the call to `ForEachRemaining` and reset it each time. Avoids all the pooling and defer overhead (becomes one defer per merge instead of per series)

src/dbnode/persist/fs/merger.go
richardartoul: actually it wouldnt matter because MergeWith may not always have the tags. Could use them if it happened to exist, but kind of a needless optimization. Forget I said anything

src/dbnode/persist/fs/types.go
```diff
@@ -489,3 +489,22 @@ type BlockRetrieverOptions interface {
...
+		seriesID ident.ID,
+		blockStart xtime.UnixNano,
+		nsCtx namespace.Context,
+	) ([]xio.BlockReader, bool, error)
+
+	// ForEachRemaining is the loop for the second stage of merging. The
+	// fsMerger first loops through the fileset series, merging them with data
+	// in the merge target. The second stage is the go through the merge target
+	// data and write the remaining series that were not merged.
+	ForEachRemaining(blockStart xtime.UnixNano, fn ForEachRemainingFn) error
```
richardartoul: I feel like the doc string for this method should really just say that this function will loop through each combination seriesID/blockStart that was not already handled by a call to `Read()`

src/dbnode/storage/shard.go
```diff
@@ -1911,7 +1920,73 @@ func (s *dbShard) Flush(
...
+
+	var (
+		multiErr           xerrors.MultiError
+		dirtySeries        = resources.dirtySeries
+		dirtySeriesToWrite = resources.dirtySeriesToWrite
+		idElementPool      = resources.idElementPool
+		blockSize          = s.namespace.Options().RetentionOptions().BlockSize()
+	)
+
+	// First, loop through all series to capture data on which blocks have dirty
```
richardartoul: Is there no locking that is required while any of this is going on? I didn't see any in the mergeWith implementation either

src/dbnode/storage/shard.go
```diff
@@ -1911,7 +1920,73 @@ func (s *dbShard) Flush(
...
+		return true
+	})
+
+	merger := fs.NewMerger(resources.fsReader, s.opts.SegmentReaderPool(),
+		s.opts.MultiReaderIteratorPool(), s.opts.IdentifierPool(), s.opts.EncoderPool())
+	mergeWithMem := newFSMergeWithMem(s, s, dirtySeries, dirtySeriesToWrite)
+	// Loop through each block that we know has ColdWrites. Since each block
+	// has its own fileset, if we encounter an error while trying to persist
+	// a block, we continue to try persisting other blocks.
+	for blockStart := range dirtySeriesToWrite {
```
richardartoul: Nice this looks much cleaner

src/dbnode/storage/fs_merge_with_mem.go
```diff
@@ -0,0 +1,109 @@
...
+	tmpCtx.BlockingClose()
+	if err != nil {
+		return nil, false, err
+	}
+
+	if len(blocks) > 0 {
+		return blocks, true, nil
+	}
+
+	return nil, false, nil
```
richardartoul: Does this handle the case where it got evicted basically since you last saw it?

src/dbnode/storage/series/buffer.go
```diff
@@ -537,7 +588,45 @@ func (b *dbBuffer) ReadEncoded(
...
 
+func (b *dbBuffer) FetchBlocksForColdFlush(
+	ctx context.Context,
+	start time.Time,
+	version int,
+	nsCtx namespace.Context,
+) ([]xio.BlockReader, error) {
+	res := b.fetchBlocks(ctx, []time.Time{start},
+		streamsOptions{filterWriteType: true, writeType: ColdWrite, nsCtx: nsCtx})
+	if len(res) != 1 {
```
richardartoul: What if it got evicted for falling out of retention or something? Also is it guaranteed to be 1 (not 2 or 3 if multiple encoder or w/e)

src/dbnode/storage/shard.go
```diff
@@ -2078,6 +2142,17 @@ func (s *dbShard) Repair(
 	return repairer.Repair(ctx, s.namespace.ID(), tr, s)
 }
 
+func (s *dbShard) TagsFromSeriesID(seriesID ident.ID) (ident.Tags, error) {
+	s.RLock()
+	entry, _, err := s.lookupEntryWithLock(seriesID)
```
richardartoul: Can this return a nil entry?

src/dbnode/storage/shard.go
```diff
@@ -1453,6 +1446,23 @@ func (s *dbShard) FetchBlocks(
...
 
+func (s *dbShard) FetchBlocksForColdFlush(
+	ctx context.Context,
+	seriesID ident.ID,
+	start time.Time,
+	version int,
+	nsCtx namespace.Context,
+) ([]xio.BlockReader, error) {
+	s.RLock()
+	entry, _, err := s.lookupEntryWithLock(seriesID)
```
richardartoul: Can this return a nil entry?

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,246 @@
...
+		if err := encoder.Encode(it.Current()); err != nil {
+			encoder.Close()
+			return err
+		}
+	}
+	if err := it.Err(); err != nil {
+		return err
+	}
+
+	stream, ok := encoder.Stream(encoding.StreamOptions{})
```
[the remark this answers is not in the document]
justinjc: Nice, good point.

src/dbnode/persist/fs/merger.go
justinjc: 👍 will do the more efficient alternative.

src/dbnode/persist/fs/merger.go
justinjc: Good point. Will also do the more efficient alternative.

src/dbnode/persist/fs/merger.go
justinjc: `Read` will never return a nil tag iterator, so we'll be safe here. I'll add a comment stating that.

src/dbnode/persist/fs/merger.go
justinjc: 👍 makes sense.

src/dbnode/storage/shard.go
justinjc: Good point.

src/dbnode/storage/shard.go
justinjc: Good point.

src/dbnode/persist/fs/merger.go
justinjc: What do you mean by "the same data"? `mergeWith.Read` should only be called once per seriesID/blockStart combination for the merge target data.

src/dbnode/storage/series/buffer.go
justinjc: Good point regarding falling out of retention - I don't think there is any other case where we won't get any data here.

`fetchBlocks` returns one result per block start, so this must be length 1 if we get any data.

src/dbnode/storage/fs_merge_with_mem.go
justinjc: After addressing your comment on `buffer.go` regarding blocks falling out of retention, yes. We'll return `false` as the second return argument and the merger will simply not include any data from memory.

src/dbnode/storage/shard.go
justinjc: Thanks for the detailed suggestion. I've incorporated most of these ideas to the PR.

src/dbnode/storage/fs_merge_with_mem.go
```diff
@@ -0,0 +1,109 @@
...
+	"github.com/m3db/m3/src/dbnode/namespace"
+	"github.com/m3db/m3/src/dbnode/persist/fs"
+	"github.com/m3db/m3/src/dbnode/storage/series"
+	"github.com/m3db/m3/src/dbnode/x/xio"
+	"github.com/m3db/m3/src/x/context"
+	"github.com/m3db/m3/src/x/ident"
+	xtime "github.com/m3db/m3/src/x/time"
+)
+
+// fsMergeWithMem implements fs.MergeWith
```
richardartoul: Little bit more here would be nice (implemented fs.MergeWith using the shard data structures such that...)

src/dbnode/storage/series/buffer.go
```diff
@@ -537,7 +588,51 @@ func (b *dbBuffer) ReadEncoded(
...
+	version int,
+	nsCtx namespace.Context,
+) ([]xio.BlockReader, error) {
+	res := b.fetchBlocks(ctx, []time.Time{start},
+		streamsOptions{filterWriteType: true, writeType: ColdWrite, nsCtx: nsCtx})
+	if len(res) == 0 {
+		// The lifecycle of calling this function is preceded by first checking
+		// which blocks have cold data that have not yet been flushed.
+		// If we don't get data here, it means that it has since fallen out of
+		// retention and has evicted.
```
richardartoul: `has been`

src/dbnode/storage/series/series.go
```diff
@@ -301,6 +301,20 @@ func (s *dbSeries) ReadEncoded(
...
 
+func (s *dbSeries) FetchBlocksForColdFlush(
+	ctx context.Context,
+	start time.Time,
+	version int,
+	nsCtx namespace.Context,
+) ([]xio.BlockReader, error) {
+	// This needs a write lock because the version on underlying buckets need
+	// to be modified.
+	s.Lock()
```
richardartoul: This is going to be called in a tight loop right? Would be good to avoid the defer overhead here if possible

richardartoul:
```golang
s.Lock()
br, err := s.buffer.FetchBlocksForColdFlush(ctx, start, version, nsCtx)
s.Unlock()

return br, err
```

src/dbnode/storage/shard.go
richardartoul: ping

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,255 @@
...
+				Namespace:  nsID,
+				Shard:      shard,
+				BlockStart: startTime,
+			},
+			FileSetType: persist.FileSetFlushType,
+		}
+	)
+
+	if err := reader.Open(openOpts); err != nil {
+		multiErr = multiErr.Add(err)
```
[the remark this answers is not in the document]
justinjc: Ah yes this was from before the refactor. I'll convert all of these back to just returning err.

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,255 @@
...
+
+type merger struct {
+	reader        DataFileSetReader
+	srPool        xio.SegmentReaderPool
+	multiIterPool encoding.MultiReaderIteratorPool
+	identPool     ident.Pool
+	encoderPool   encoding.EncoderPool
+}
+
+// NewMerger returns a new Merger.
```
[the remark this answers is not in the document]
justinjc: Okay I'll expand on this a little more, although the merger doesn't really know where/how the new fileset will be persisted, since it is just passed a `flushPreparer`.

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,255 @@
...
+		}
+		if hasData {
+			br = append(br, encoded...)
+		}
+		brs = append(brs, br)
+
+		multiIter.ResetSliceOfSlices(xio.NewReaderSliceOfSlicesFromBlockReadersIterator(brs), nsCtx.Schema)
+
+		// tagsIter is never nil.
+		tags, err := convert.TagsFromTagsIter(id, tagsIter, identPool)
```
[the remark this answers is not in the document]
justinjc: The place where `convert.TagsFromTagsIter` is called is only for series IDs that came from the (disk) reader. The lifecycle for those IDs are contained to within the merger (created from the `Read()` and then finalized after closing the preparer), so using that function should be safe, right?

Agreed on the `iter -> tags -> iter` comment. Created https://github.com/m3db/m3/issues/1675.

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,255 @@
...
+	}
+	if err := it.Err(); err != nil {
+		return err
+	}
+
+	segment := encoder.Discard()
+	checksum := digest.SegmentChecksum(segment)
+
+	err := persistFn(id, tags, segment, checksum)
+	id.Finalize()
```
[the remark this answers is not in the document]
justinjc: Thanks for the catch. Having disk IDs finalizable but not for the IDs from memory makes this logic difficult to integrate with at the writer layer, since you'd need two different flush preparers with different open options.  I'll just keep track of the IDs from disk and finalize them at the end.

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,255 @@
...
+	// in the first stage.
+	err = mergeWith.ForEachRemaining(blockStart, func(seriesID ident.ID, tags ident.Tags) bool {
+		encoded, hasData, err := mergeWith.Read(seriesID, blockStart, nsCtx)
+		if err != nil {
+			multiErr = multiErr.Add(err)
+			return false
+		}
+
+		if hasData {
+			multiIter.Reset(xio.NewSegmentReaderSliceFromBlockReaderSlice(encoded), startTime, blockSize, nsCtx.Schema)
```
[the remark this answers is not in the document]
justinjc: I'll use the `ResetSliceOfSlices` interface and reuse the `BlockReaders` from above instead.

src/dbnode/persist/fs/merger_test.go
```diff
@@ -0,0 +1,367 @@
...
+	})
+	bytesPool.Init()
+	identPool = ident.NewPool(bytesPool, ident.PoolOptions{})
+	encoderPool = encoding.NewEncoderPool(nil)
+	encoderPool.Init(func() encoding.Encoder {
+		return m3tsz.NewEncoder(startTime, nil, true, encoding.NewOptions())
+	})
+}
+
+func TestMergeWithIntersection(t *testing.T) {
```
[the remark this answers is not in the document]
justinjc: Added some more scenarios.

src/dbnode/storage/database.go
```diff
@@ -232,10 +232,10 @@ func (d *db) updateSchemaRegistry(newNamespaces namespace.Map) error {
 	schemaUpdates := newNamespaces.Metadatas()
 	merr := xerrors.NewMultiError()
 	for _, metadata := range schemaUpdates {
-		curSchemaId := "none"
+		curSchemaID := "none"
 		curSchema, err := schemaReg.GetLatestSchema(metadata.ID())
 		if curSchema != nil {
-			curSchemaId = curSchema.DeployId()
+			curSchemaID = curSchema.DeployId()
```
prateek: not from your changes but looks like the error on line 236 is unchecked. what's the deal there?

src/dbnode/persist/fs/types.go
```diff
@@ -489,3 +490,45 @@ type BlockRetrieverOptions interface {
...
+	// Merge merges the specified fileset file with a merge target.
+	Merge(
+		fileID FileSetFileIdentifier,
+		mergeWith MergeWith,
+		flushPreparer persist.FlushPreparer,
+		nsCtx namespace.Context,
+	) error
+}
+
+// NewMergerFn is the function to call to get a new Merger. Mostly used as a
```
prateek: hm if it's only used for testing, does it need to be exported?

src/dbnode/storage/series/buffer.go
```diff
@@ -56,11 +57,17 @@ var (
 
 const (
 	bucketsCacheSize = 2
-	// TODO(juchan): This represents the maximum number of blocks we think we'll
```
prateek: Is this TODO still relevant?

src/dbnode/storage/series/buffer.go
```diff
@@ -56,11 +57,17 @@ var (
...
+	// optimizedTimes struct. Since the size of this array determines the
+	// effectiveness of minimizing heap allocations, usage of this struct and/or
+	// changing this const should only be done after considering its current
+	// use cases:
+	// 1) The number of buckets that will be removed within a tick due to that
+	//    block being recently flushed
+	// 2) The number of buckets that contain ColdWrites within a cold flush
+	//    cycle
+	optimizedTimesArraySize = 8
+	writableBucketVer       = 0
```
prateek: nit: rename to `writableBucketVersion`

src/dbnode/storage/series/types.go
```diff
@@ -95,24 +105,26 @@ type DatabaseSeries interface {
...
+	// not been rotated into a block yet.
 	Snapshot(
 		ctx context.Context,
 		blockStart time.Time,
 		persistFn persist.DataFn,
 		nsCtx namespace.Context,
 	) error
 
+	// NeedsColdFlushBlockStarts returns the block starts that need cold flushes.
+	NeedsColdFlushBlockStarts() OptimizedTimes
```
prateek: Other methods we have called `Needs...` return a bool (e.g. `NeedsFlush(time) bool`), would be better to call this something like `ColdFlushBlockStarts`?

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,287 @@
...
+// the data from an existing fileset with a merge target, giving precendence
+// to the data from the merge target. This merged data is then persisted.
+//
+// Note that the merger it does not know how or where this merged data is
+// persisted since it just uses the flushPreparer that is passed in. Further,
+// it does not signal to the database of the existence of the newly persisted
+// data, nor does it clean up the original fileset.
+func NewMerger(
+	reader DataFileSetReader,
+	blockAllocSize int,
```
prateek: nit: instead of all the args, could you make an Options type

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,287 @@
...
+				BlockStart: startTime,
+			},
+			FileSetType: persist.FileSetFlushType,
+		}
+	)
+
+	if err := reader.Open(openOpts); err != nil {
+		return err
+	}
+	defer reader.Close()
```
prateek: do you need to check the returned error for this?

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,287 @@
...
+		// We must only finalize these at the end of this function, since the
+		// flush preparer's underlying writer holds on to those references
+		// until it is closed (closing the PreparedDataPersist at the end of
+		// this merge closes the underlying writer).
+		idsToFinalize  []ident.ID
+		tagsToFinalize []ident.Tags
+	)
+	defer segReader.Finalize()
+	defer multiIter.Close()
+	defer func() {
```
robskillington: Could we combine these defers into a single defer?

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,287 @@
...
+		// lifecycle of those IDs/tags are contained to this function.  We don't
+		// want finalize this series ID since it is from memory and other
+		// components may have ownership over it.
+		//
+		// We must only finalize these at the end of this function, since the
+		// flush preparer's underlying writer holds on to those references
+		// until it is closed (closing the PreparedDataPersist at the end of
+		// this merge closes the underlying writer).
+		idsToFinalize  []ident.ID
+		tagsToFinalize []ident.Tags
```
robskillington: Perhaps make these both the number of entries expected?

i.e.
```go
idsToFinalize = make([]ident.ID, 0, reader.Entries())
tagsToFinalize = make([]ident.Tags, 0, reader.Entries())
```

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,287 @@
...
+	encoder := encoderPool.Get()
+	encoder.Reset(blockStart, blockAllocSize, schema)
+	for it.Next() {
+		if err := encoder.Encode(it.Current()); err != nil {
+			encoder.Close()
+			return err
+		}
+	}
+	if err := it.Err(); err != nil {
+		return err
```
robskillington: Missing encoder.Close() for this early return case too (not super important, but may as well).

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,287 @@
...
+
+	segment := encoder.Discard()
+	checksum := digest.SegmentChecksum(segment)
+
+	err := persistFn(id, tags, segment, checksum)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
robskillington: nit: Perhaps combine this to just `return persistFn(id, tags, segment, checksum)`?

src/dbnode/storage/flush.go
```diff
@@ -63,6 +64,7 @@ type flushManager struct {
 	// are used for emitting granular gauges.
 	state           flushManagerState
 	isFlushing      tally.Gauge
+	isColdFlushing  tally.Gauge
```
robskillington: Super nit, I wonder if we should call these a "compaction" flush instead of cold flush?  I'm not really too fussed, but interested to hear thoughts cc @justinjc @prateek @richardartoul

robskillington: (Like still keep the hot/cold terminology but just call these flushes "compaction flushes" because then if repairs uses this code path, it's not just cold writes that are being persisted per-se).

src/dbnode/storage/flush.go
```diff
@@ -102,28 +105,63 @@ func (m *flushManager) Flush(
...
+		}
+	} else {
+		multiErr = multiErr.Add(fmt.Errorf("error rotating commitlog in mediator tick: %v", err))
+	}
+
+	if err = m.indexFlush(namespaces); err != nil {
+		multiErr = multiErr.Add(err)
+	}
+
+	return multiErr.FinalError()
```
robskillington: Are we sure that if there is an error here that the commit logs won't get deleted?  I think the commit log cleanup logic only checks that there is a valid snapshot, not that also cold flushes contained by those commit logs were written to disk no?

cc @richardartoul

src/dbnode/storage/fs_merge_with_mem.go
```diff
@@ -0,0 +1,112 @@
...
+
+	// Series is in memory, so it will get merged with disk and
+	// written in this loop. Therefore, we need to remove it from
+	// the "to write" list so that the later loop does not rewrite
+	// it.
+	m.dirtySeriesToWrite[blockStart].Remove(element)
+
+	nextVersion := m.retriever.RetrievableBlockColdVersion(startTime) + 1
+
+	tmpCtx := context.NewContext()
```
robskillington: Perhaps get one from a context pool or else reuse one (since you are calling `BlockingClose` which makes them reuseable).

src/dbnode/storage/flush.go
robskillington: Second thoughts, seems to hard to rename at this point, let's not bother.

src/dbnode/storage/namespace.go
```diff
@@ -924,20 +926,90 @@ func (n *dbNamespace) Flush(
...
+
+	multiErr := xerrors.NewMultiError()
+	shards := n.GetOwnedShards()
+
+	fsReader, err := fs.NewReader(n.opts.BytesPool(), n.opts.CommitLogOptions().FilesystemOptions())
+	if err != nil {
+		return err
+	}
+	resources := coldFlushReuseableResources{
+		dirtySeries:        newDirtySeriesMap(dirtySeriesMapOptions{}),
```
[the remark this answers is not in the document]
robskillington: I think it's good not to reuse these between cold flushes since it means we don't bloat process at all for most workloads since they shouldn't do it often/at all and it wouldn't any extra overhead for those.

For the case where it happens often, they know already they're not doing the ideal thing (writing into the past often) and can likely deal with a bit of extra GC.

src/dbnode/storage/series/types.go
justinjc: Yes that definitely makes more sense 👍

src/dbnode/storage/flush.go
justinjc: Ah right, that's true. We can either:
1) Return if there's an error cold flushing
2) Don't clean up commit logs unless a successful cold flush has happened too
Both run the risk of using up too much disk space if cold flushes fail, but (1) is easier to implement, so I'll do that.

src/dbnode/storage/flush.go
justinjc: I actually renamed this from compact to cold flush midway through 😅. Maybe we'll re-tackle this when the usage of cold flush differs too much later.

src/dbnode/persist/fs/types.go
justinjc: That comment is confusing - I removed it. What I meant is that I needed the ability to pass in the merger constructor cold flush code in `shard.go` (a different package) so that it is testable.

src/dbnode/storage/database.go
justinjc: Hm, not immediately obvious to me what to do on error, opened #1694.

src/dbnode/storage/series/buffer.go
justinjc: Fair, we should still monitor to see performance after ColdWrites are enabled.

src/dbnode/storage/fs_merge_with_mem.go
justinjc: Makes sense, I'll have `Read` take a context and share one across iterations.

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,293 @@
...
+		srPool:         srPool,
+		multiIterPool:  multiIterPool,
+		identPool:      identPool,
+		encoderPool:    encoderPool,
+		nsOpts:         nsOpts,
+	}
+}
+
+// Merge merges data from a fileset with a merge target and persists it.
+func (m *merger) Merge(
```
richardartoul: Can you add to the doc string to say that the caller is responsible for finalizing all data structures in the `MergeWith`? Alternatively you could make it so the `MergeWith` has a `Finalize()` method that the merger can call, and then if the caller doesn't want their data finalized they can just make that operation a no-op. 

Either way is fine to me

src/dbnode/persist/fs/merger.go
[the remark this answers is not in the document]
richardartoul: @justinjc Yes I think with the restructuring you've done this is now safe since the tags will be valid as long as the IDs are valid, and the IDs are valid for the duration of the file writing. Can you add a comment explaining this now?

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,293 @@
...
+		tagsToFinalize = append(tagsToFinalize, tags)
+		if err != nil {
+			return err
+		}
+		if err := persistIter(prepared.Persist, multiIter, startTime,
+			id, tags, blockAllocSize, nsCtx.Schema, encoderPool); err != nil {
+			return err
+		}
+	}
+	// Second stage: loop through rest of the merge target that was not captured
```
richardartoul: "loop through any series in the merge target that were not captured in the first stage"

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,293 @@
...
+	// in the first stage.
+	err = mergeWith.ForEachRemaining(blockStart, func(seriesID ident.ID, tags ident.Tags) error {
+		mergeWithData, hasData, err := mergeWith.Read(ctx, seriesID, blockStart, nsCtx)
```
richardartoul: I still feel like you shouldn't have to do a read here and the data should just be passed to the function in `ForEachRemaining`. If the series exists then it should have no data, no need to make the caller call `mergeWith.Read()` again

src/dbnode/persist/fs/types.go
```diff
@@ -489,3 +491,45 @@ type BlockRetrieverOptions interface {
 	// IdentifierPool returns the identifierPool
 	IdentifierPool() ident.Pool
 }
+
+// ForEachRemainingFn is the function that is run on each of the remaining
+// series of the merge target that did not intersect with the fileset.
+type ForEachRemainingFn func(seriesID ident.ID, tags ident.Tags) error
```
richardartoul: So related to my previous comment, this type would also accept a `[]xio.BlockReader`

src/dbnode/storage/flush.go
```diff
@@ -102,28 +105,69 @@ func (m *flushManager) Flush(
...
+	if err = m.dataWarmFlush(namespaces, tickStart, dbBootstrapStateAtTickStart); err != nil {
+		multiErr = multiErr.Add(err)
+	}
+
+	rotatedCommitlogID, err := m.commitlog.RotateLogs()
+	if err == nil {
+		if err = m.dataColdFlush(namespaces, tickStart); err != nil {
+			multiErr = multiErr.Add(err)
+			// If cold flush fails, we can't proceed to snapshotting because
+			// commit log logic just checks for a successful snapshot checkpoint
```
richardartoul: "because commit log cleanup logic uses the presence of a successful snapshot checkpoint file to determine which commit log files are safe to delete."

src/dbnode/storage/flush.go
```diff
@@ -102,28 +105,69 @@ func (m *flushManager) Flush(
...
+			// file. Therefore if a cold flush fails and a snapshot succeeds,
+			// the writes from the failed cold flush might be lost when commit
+			// logs get cleaned up.
```
richardartoul: "get cleaned up, leaving the node in an undurable state such that if it restarted it would not be able to recover the cold writes from its commit log"

I know I'm kind of dragging this stuff out, but better to be verbose here because eventually we'll forget

src/dbnode/storage/flush.go
```diff
@@ -102,28 +105,69 @@ func (m *flushManager) Flush(
...
+			return multiErr.FinalError()
+		}
+
+		// Snapshots must be done after a cold flush because cleanup logic just
```
richardartoul: I don't think this comment adds anything thats not in the previous comment, does it?

src/dbnode/storage/flush.go
```diff
@@ -102,28 +105,69 @@ func (m *flushManager) Flush(
...
+	// also slightly better semantically because flushing should take priority
+	// over cold flushes and snapshotting.
+	//
+	// In addition, we need to make sure that for any given shard/blockStart
+	// combination, we attempt a flush and then a cold flush before a snapshot
+	// as the snapshotting process will attempt to snapshot any unflushed blocks
+	// which would be wasteful if the block is already flushable.
+	multiErr := xerrors.NewMultiError()
+	if err = m.dataWarmFlush(namespaces, tickStart, dbBootstrapStateAtTickStart); err != nil {
+		multiErr = multiErr.Add(err)
```
richardartoul: This actually needs to either be a hard error, or if it happened, block cold flushes because if warm flushes fail the cold flushes will attempt to flush blocks that have not yet been warm flushed which violates the model that cold flushes for a block never happen before a warm flush

src/dbnode/storage/flush.go
```diff
@@ -147,42 +191,41 @@ func (m *flushManager) Flush(
...
+	multiErr := xerrors.NewMultiError()
 	for _, ns := range namespaces {
-		var (
-			indexOpts    = ns.Options().IndexOptions()
-			indexEnabled = indexOpts.Enabled()
-		)
-		if !indexEnabled {
-			continue
+		rOpts := ns.Options().RetentionOptions()
+		latestWarmFlushTime := retention.FlushTimeEnd(rOpts, tickStart)
```
richardartoul: This definitely deserves a comment haha.

```
The current assumption of the warm flush system is that every warm flush will be the first time data for that namespace/shard/blockStart is written to disk and that no other files should be present. As a result, cold flushes for blocks that have not been warm flushed yet need to be avoided.

We avoid cold flushing blocks that have not yet been warm flushed by:

1. Performing warm flushes before cold flushes and not allowing cold flushes to proceed if the warm flushes failed.
2. Only performing flushes up to the latest block that is warm flushable to prevent cold flushes from flushing blocks that have not yet been made available to warm flush.
```

src/dbnode/storage/flush.go
richardartoul: ...also I just realized that the `WarmFlush` method has this logic:

```golang
shardBootstrapStateBeforeTick, ok := shardBootstrapStatesAtTickStart[shard.ID()]
		if !ok || shardBootstrapStateBeforeTick != Bootstrapped {
			// We don't own this shard anymore (!ok) or the shard was not bootstrapped
			// before the previous tick which means that we have no guarantee that all
			// bootstrapped blocks have been rotated out of the series buffer buckets,
			// so we wait until the next opportunity.
			n.log.
				With(zap.Uint32("shard", shard.ID())).
				With(zap.Any("bootstrapStateBeforeTick", shardBootstrapStateBeforeTick)).
				With(zap.Bool("bootstrapStateExists", ok)).
				Debug("skipping snapshot due to shard bootstrap state before tick")
			continue
		}
```

but the `ColdFlush` logic does not which means that cold flush could end up trying to flush a block that was never WarmFlushed.

Can we just add a blanket check inside of the `ColdFlush` method itself to make sure that a warm flush has succeeded before attempting a cold flush for any blocks?

src/dbnode/storage/flush.go
richardartoul: @justinjc I gues per our discussion about moving this check into the ColdFlush method itself such that it will never attempt to ColdFlush a block that hasn't been WarmFlushed yet, we won't need this.

src/dbnode/storage/fs_merge_with_mem.go
```diff
@@ -0,0 +1,112 @@
...
+	// Series is in memory, so it will get merged with disk and
+	// written in this loop. Therefore, we need to remove it from
+	// the "to write" list so that the later loop does not rewrite
+	// it.
+	m.dirtySeriesToWrite[blockStart].Remove(element)
+
+	nextVersion := m.retriever.RetrievableBlockColdVersion(startTime) + 1
+
+	blocks, err := m.shard.FetchBlocksForColdFlush(ctx, element.Value, startTime, nextVersion, nsCtx)
+	ctx.BlockingClose()
```
richardartoul: Why do you do a `BlockingClose()` here? Won't that finalize the blocks?

src/dbnode/storage/namespace.go
```diff
@@ -938,20 +940,112 @@ func (n *dbNamespace) Flush(
...
+	}
+
+	multiErr := xerrors.NewMultiError()
+	shards := n.GetOwnedShards()
+
+	fsReader, err := fs.NewReader(n.opts.BytesPool(), n.opts.CommitLogOptions().FilesystemOptions())
+	if err != nil {
+		return err
+	}
+	resources := coldFlushReuseableResources{
```
richardartoul: Can you just make a constructor function for this? Its a little tedious to be reading through the logic for `ColdFlushes` and be wondering why you're constructing an `fsReader`

src/dbnode/storage/series/buffer.go
```diff
@@ -281,6 +311,26 @@ func (b *dbBuffer) IsEmpty() bool {
...
+	var times OptimizedTimes
+
+	for t, bucketVersions := range b.bucketsMap {
+		for _, bucket := range bucketVersions.buckets {
+			if bucket.version == writableBucketVersion &&
+				bucket.writeType == ColdWrite &&
+				// Cold flushes can only happen on blockStarts that have been
+				// warm flushed, because warm flush logic does not currently
+				// perform any merging logic.
+				!bucket.start.After(latestWarmFlushTime) {
```
richardartoul: Are we gonna remove this check? Seems weird to have it here

src/dbnode/storage/series/buffer.go
```diff
@@ -281,6 +311,26 @@ func (b *dbBuffer) IsEmpty() bool {
 	return len(b.bucketsMap) == 0
 }
 
+func (b *dbBuffer) ColdFlushBlockStarts(latestWarmFlushTime time.Time) OptimizedTimes {
+	var times OptimizedTimes
+
+	for t, bucketVersions := range b.bucketsMap {
+		for _, bucket := range bucketVersions.buckets {
+			if bucket.version == writableBucketVersion &&
```
richardartoul: So basically this is checking to see if there is a writableBucket to determine that this blockStart has data that needs to be cold flushed? What about the following scenario:

1. Block X has cold data that needs to be flushed
2. ColdFlush comes in and "flushes" the data, incrementing the bucketVersion, but the cold flush ultimately fails and the data is not written to disk. No subsequent writes come in for this bucket.
3. Next cold flush comes in and calls ColdFlushBlockStarts and thinks this block doesn't have any data that needs to be cold flushed because the version was already incremented

Does this function need to accept the current successful cold flush version, similar to the tick function?

src/dbnode/storage/series/buffer.go
```diff
@@ -537,7 +595,51 @@ func (b *dbBuffer) ReadEncoded(
 	return res, nil
 }
 
+func (b *dbBuffer) FetchBlocksForColdFlush(
+	ctx context.Context,
```
richardartoul: This is what I meant by the `BlockingClose()` on the ctx probably being wrong since it will cause the data returned by this function to be finalized

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,293 @@
...
+	blockAllocSize int
+	srPool         xio.SegmentReaderPool
+	multiIterPool  encoding.MultiReaderIteratorPool
+	identPool      ident.Pool
+	encoderPool    encoding.EncoderPool
+	nsOpts         namespace.Options
+}
+
+// NewMerger returns a new Merger. This implementation is in charge of merging
+// the data from an existing fileset with a merge target, giving precendence
```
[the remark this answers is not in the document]
justinjc: Basically the data from the merge target will be taken. I'll reword it.

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,293 @@
...
+	var (
+		// There will only be one BlockReader slice since we're working within one
+		// block here.
+		brs = make([][]xio.BlockReader, 0, 1)
+		// There will likely be at least two BlockReaders - one for disk data and
+		// one for data from the merge target.
+		br = make([]xio.BlockReader, 0, 2)
+
+		// It's safe to share these between iterations and just reset them each time
+		// because the series gets persisted each loop, so previous iterations'
```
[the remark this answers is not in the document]
justinjc: Hm? I'm not sure what you're getting at here.

src/dbnode/storage/flush.go
justinjc: Sure.

src/dbnode/storage/flush.go
justinjc: Yeah I suppose.

src/dbnode/storage/flush.go
justinjc: Yeah that makes sense. We should just check the shard flush state since that is the authoritative location on whether something got successfully flushed or not.

src/dbnode/storage/flush.go
justinjc: 👍

src/dbnode/storage/series/buffer.go
justinjc: Yep, will remove this check and use shard flush state to determine whether something has been warm flushed instead.

src/dbnode/persist/fs/merger.go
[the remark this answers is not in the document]
richardartoul: sigh I meant to say "so previous" -> "so the previous"

src/dbnode/storage/series/buffer.go
justinjc: Nice catch on this edge case. Since this function goes through all its block starts, I'll pass through a block state snapshot, which is the same as what we do for the Tick.

src/dbnode/persist/fs/types.go
justinjc: Ah I see how this would simplify things.

src/dbnode/persist/fs/merger.go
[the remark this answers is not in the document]
justinjc: Sure thing.

src/dbnode/persist/fs/merger.go
justinjc: Makes sense to me now 👍

src/dbnode/persist/fs/merger.go
justinjc: Sure, I'll go with the former.

src/dbnode/storage/shard.go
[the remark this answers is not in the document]
richardartoul: I still don't think this logic works for `FutureWrites` because if the node restarted after performing a cold future write, the ColdVersion won't be restored after restarting the node and bootstrapping will it? Maybe file an issue to track

src/dbnode/storage/shard.go
```diff
@@ -1911,7 +1929,86 @@ func (s *dbShard) Flush(
...
+	// series and add them to the resources for further processing.
+	s.forEachShardEntry(func(entry *lookup.Entry) bool {
+		curr := entry.Series
+		seriesID := curr.ID()
+		blockStarts := curr.ColdFlushBlockStarts(blockStates)
+		blockStarts.ForEach(func(t xtime.UnixNano) {
+			// Cold flushes can only happen on blockStarts that have been
+			// warm flushed, because warm flush logic does not currently
+			// perform any merging logic.
+			if !s.hasWarmFlushed(t.ToTime()) {
```
richardartoul: can you make sure there is a test for this

src/dbnode/storage/flush.go
richardartoul: ok so this may still actually need to be a hard error. I was thinking through it and even with your check in the shard you could still end up in a bad spot:

1. Warm flush fails
2. Cold flush skips block cause it wasn't warm flushed (but looks like it succeeded cause you just skipped it)
3. Snapshot completes successfully
4. Commitlogs containing unflushed cold writes get deleted <- durability lost here

I think you need to make this a hard error which solves one problem. The next problem is that the WarmFlush thing can choose to skip flushes (without returning an error) because of the `shardBootstrapStateBeforeTick` thing. We can fix that by making it return an error (and it'll happen on the next go around) or by removing it entirely because I Don't think its required now that we load blocks directly into the buffer on bootstrap and flush out of the buffer too.

Lets not make that last change in this P.R though, just file an issue for it.

My guess is that we also want to make the ColdFlush logic not skip blocks that haven't been WarmFlushed yet, but actually return an error if that situations happens because it means we messed up this lifecycle and should go fix the bug

richardartoul: (invariant violated error)

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,294 @@
...
+		tagsIter.Close()
+		tagsToFinalize = append(tagsToFinalize, tags)
+		if err != nil {
+			return err
+		}
+		if err := persistIter(prepared.Persist, multiIter, startTime,
+			id, tags, blockAllocSize, nsCtx.Schema, encoderPool); err != nil {
+			return err
+		}
+		tmpCtx.BlockingClose()
```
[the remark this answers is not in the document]
justinjc: That makes sense. I'll opt with just adding the comment.

src/dbnode/storage/shard.go
justinjc: Yep, already added one for this case.

src/dbnode/storage/shard.go
```diff
@@ -2078,6 +2159,17 @@ func (s *dbShard) Repair(
 	return repairer.Repair(ctx, s.namespace.ID(), tr, s)
 }
 
+func (s *dbShard) TagsFromSeriesID(seriesID ident.ID) (ident.Tags, error) {
+	s.RLock()
+	entry, _, err := s.lookupEntryWithLock(seriesID)
+	s.RUnlock()
+	if entry == nil || err != nil {
```
richardartoul: I think `entry == nil` should be handled separately. I.E you probably want to propagate that information to the caller (with an `ok bool` or something) so that they can make a decision (in your case, probably to skip since it clearly fell out of retention or something)

richardartoul: Also technically its fine right now but to be extra safe can you change `newShardEntry` in `shard.go` to mark all of the tags as `NoFinalize()` to ensure that they never get put back in a pool while the merge logic still has a handle on them?

src/dbnode/persist/fs/merger.go
```diff
@@ -0,0 +1,303 @@
...
+		multiIterPool:  multiIterPool,
+		identPool:      identPool,
+		encoderPool:    encoderPool,
+		nsOpts:         nsOpts,
+	}
+}
+
+// Merge merges data from a fileset with a merge target and persists it.
+// The caller is responsible for finalizing all resources used for the
+// MergeWith passed here.
```
richardartoul: Can you document somewhere (here or on `ForEachRemaining`) that while the data passed to `ForEachRemaining` is basically a copy that will be finalized when the context is closed, the ID and tags are expected to live for as long as the user of the MergeWith requires them so they should either be `NoFinalize()` or passed as copies (in your case since you're taking them from the shard/series they're NoFinalize() and its safe but would like to document)

src/dbnode/storage/series/buffer.go
```diff
@@ -422,20 +422,18 @@ func (b *dbBuffer) Snapshot(
...
-		streams    = buckets.streams(ctx, streamsOptions{filterWriteType: true, writeType: WarmWrite})
-		numStreams = len(streams)
-	)
+	var mergedStream xio.SegmentReader
 	if numStreams == 1 {
-		stream = streams[0]
+		mergedStream = streams[0]
 	} else {
 		// We may need to merge again here because the regular merge method does
 		// not merge buckets that have different versions.
```
richardartoul: It also doesn't merge warm and cold buckets basically right? maybe say that too

src/dbnode/storage/shard.go
```diff
@@ -1964,7 +1959,7 @@ func (s *dbShard) ColdFlush(
 			// Cold flushes can only happen on blockStarts that have been
 			// warm flushed, because warm flush logic does not currently
 			// perform any merging logic.
-			if !s.hasWarmFlushed(t.ToTime()) {
+			if !s.IsBlockRetrievable(t.ToTime()) {
```
richardartoul: Can you actually keep the duplicated method `hasWarmFlushed`? I know its lame to duplicate but I can easily see someone changing the meaning of `IsBlockRetrievable` and not expecting the impact it'll have on this

src/dbnode/storage/fs_merge_with_mem.go
justinjc: 👍 fixed this.

src/dbnode/storage/series/buffer.go
justinjc: 👍 fixed this.

src/dbnode/storage/shard.go
justinjc: Sure.

PR #1625, opened 2019-05-09
Update Stream() API and refactor/fix bugs in buffer.go

### LINKED ISSUE #805: Make encoder.Stream() return a bool indicating whether data is present
The current interface requires a nil check against the stream result which is error prone (people forget to do it)

richardartoul:

This P.R updates the Stream() API to be more difficult to misuse (returning a bool indicating if any data is present) as well as accept an options struct which will be used in a subsequent P.R.

In addition to the updated API, this P.R fixes two bugs in `buffer.go` and adds regression tests for them as well.
If there is an error during encoding its possible to end up with an encoder in the buffer that has no writes (an empty stream) which will cause the `Flush()` method to panic because it doesn't account for this possibility.

Additional changes included in this P.R:
- [X] Make Stream() API return a bool indicating if data is present in the stream. This makes the API more difficult to misuse. In addition, instead of returning nil segment readers the API now always returns a valid reader even if there is no data (the reader will just wrap an empty segment).
- [X] Make the Stream() API accept an options struct which will be used in a subsequent P.R.
- [X] Fix a panic in `buffer.go` that would occur if a series had an empty encoder for a given block.
- [X] Fix a bug in `buffer.go` in the Write() method that would return `true` for `wasWritten` even if the write failed.
- [X] Refactor the snapshotting logic out of `series.go` into `buffer.go` to be more consistent with the flush logic

Fixes #805

### DIFF
--- a/src/dbnode/encoding/null.go
+++ b/src/dbnode/encoding/null.go
@@ -42,8 +42,8 @@ func NewNullEncoder() Encoder {
 func (e *nullEncoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, annotation ts.Annotation) error {
 	return nil
 }
-func (e *nullEncoder) Stream() (xio.SegmentReader, bool) {
-	return nil, false
+func (e *nullEncoder) Stream(opts encoding.StreamOpts) (xio.SegmentReader, bool) {
+	return xio.NewSegmentReader(ts.Segment{}), false
 }
 func (e *nullEncoder) NumEncoded() int { return 0 }
 func (e *nullEncoder) LastEncoded() (ts.Datapoint, error) {
--- a/src/dbnode/encoding/null.go
+++ b/src/dbnode/encoding/null.go
@@ -43,7 +43,7 @@ func (e *nullEncoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, annotation ts
 	return nil
 }
 func (e *nullEncoder) Stream() (xio.SegmentReader, bool) {
-	return xio.NewSegmentReader(ts.Segment{}), false
+	return nil, false
 }
 func (e *nullEncoder) NumEncoded() int { return 0 }
 func (e *nullEncoder) LastEncoded() (ts.Datapoint, error) {
--- a/src/dbnode/encoding/null.go
+++ b/src/dbnode/encoding/null.go
@@ -43,7 +43,7 @@ func (e *nullEncoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, annotation ts
 	return nil
 }
 func (e *nullEncoder) Stream(opts StreamOpts) (xio.SegmentReader, bool) {
-	return xio.NewSegmentReader(ts.Segment{}), false
+	return nil, false
 }
 func (e *nullEncoder) NumEncoded() int { return 0 }
 func (e *nullEncoder) LastEncoded() (ts.Datapoint, error) {
--- a/src/dbnode/encoding/null.go
+++ b/src/dbnode/encoding/null.go
@@ -42,7 +42,7 @@ func NewNullEncoder() Encoder {
 func (e *nullEncoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, annotation ts.Annotation) error {
 	return nil
 }
-func (e *nullEncoder) Stream(opts encoding.StreamOpts) (xio.SegmentReader, bool) {
+func (e *nullEncoder) Stream(opts StreamOpts) (xio.SegmentReader, bool) {
 	return xio.NewSegmentReader(ts.Segment{}), false
 }
 func (e *nullEncoder) NumEncoded() int { return 0 }
--- a/src/dbnode/encoding/null.go
+++ b/src/dbnode/encoding/null.go
@@ -42,8 +42,8 @@ func NewNullEncoder() Encoder {
 func (e *nullEncoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, annotation ts.Annotation) error {
 	return nil
 }
-func (e *nullEncoder) Stream() xio.SegmentReader {
-	return xio.NewSegmentReader(ts.Segment{})
+func (e *nullEncoder) Stream() (xio.SegmentReader, bool) {
+	return xio.NewSegmentReader(ts.Segment{}), false
 }
 func (e *nullEncoder) NumEncoded() int { return 0 }
 func (e *nullEncoder) LastEncoded() (ts.Datapoint, error) {
--- a/src/dbnode/encoding/null.go
+++ b/src/dbnode/encoding/null.go
@@ -42,7 +42,7 @@ func NewNullEncoder() Encoder {
 func (e *nullEncoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, annotation ts.Annotation) error {
 	return nil
 }
-func (e *nullEncoder) Stream(opts StreamOpts) (xio.SegmentReader, bool) {
+func (e *nullEncoder) Stream(opts StreamOptions) (xio.SegmentReader, bool) {
 	return nil, false
 }
 func (e *nullEncoder) NumEncoded() int { return 0 }
--- a/src/dbnode/encoding/proto/round_trip_prop_test.go
+++ b/src/dbnode/encoding/proto/round_trip_prop_test.go
@@ -155,7 +155,7 @@ func TestRoundTripProp(t *testing.T) {
 			}
 		}
 
-		stream := enc.Stream()
+		stream, ok := enc.Stream()
 		if !ok {
 			return true, nil
 		}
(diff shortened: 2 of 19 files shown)


src/dbnode/encoding/null.go
```diff
@@ -42,8 +42,8 @@ func NewNullEncoder() Encoder {
 func (e *nullEncoder) Encode(dp ts.Datapoint, timeUnit xtime.Unit, annotation ts.Annotation) error {
 	return nil
 }
-func (e *nullEncoder) Stream() xio.SegmentReader {
-	return xio.NewSegmentReader(ts.Segment{})
+func (e *nullEncoder) Stream() (xio.SegmentReader, bool) {
+	return xio.NewSegmentReader(ts.Segment{}), false
```
justinjc: Either `xio.NewSegmentReader(ts.Segment{}), true` or `nil, false`

src/dbnode/storage/block/merged_block_reader.go
```diff
@@ -129,8 +129,9 @@ func (r *dbMergedBlockReader) mergedReader() (xio.BlockReader, error) {
 		r.readers[i] = nil
 	}
 
+	stream, _ := r.encoder.Stream()
```
robskillington: Shouldn't perhaps `ok` be checked here?

src/dbnode/encoding/proto/round_trip_prop_test.go
```diff
@@ -156,7 +156,7 @@ func TestRoundTripProp(t *testing.T) {
 		}
 
 		stream := enc.Stream()
-		if stream == nil {
+		if !ok {
```
robskillington: Hm, where did this `ok` come from?  Shouldn't the line above be `stream, ok := enc.Stream()`?

src/dbnode/storage/block/merged_block_reader.go
richardartoul: The original code just passed in nil and as far as I can tell the xio.BlockReader handles nil fine

```golang
// CloneBlock returns a clone of the block with the underlying data reset
func (b BlockReader) CloneBlock(pool pool.CheckedBytesPool) (BlockReader, error) {
	sr, err := b.SegmentReader.Clone(pool)
	if err != nil {
		return EmptyBlockReader, err
	}
	return BlockReader{
		SegmentReader: sr,
		Start:         b.Start,
		BlockSize:     b.BlockSize,
	}, nil
}

// IsEmpty returns true for the empty block
func (b BlockReader) IsEmpty() bool {
	return b.SegmentReader == nil && b.Start.Equal(timeZero) && b.BlockSize == 0
}

// IsNotEmpty returns false for the empty block
func (b BlockReader) IsNotEmpty() bool {
	return !b.IsEmpty()
}

// ResetWindowed resets the underlying reader window, as well as start time and blockSize for the block
func (b *BlockReader) ResetWindowed(segment ts.Segment, start time.Time, blockSize time.Duration) {
	b.Reset(segment)
	b.Start = start
	b.BlockSize = blockSize
}
```

It looks like I'll need to update the xio.BlockReader to detect empty segments though since we want to change Stream() to return an empty segment instead of nil

richardartoul: actually I think for this implementation of the block reader it doesnt matter

src/dbnode/encoding/null.go
richardartoul: actually gonna return an empty reader and also false (to indicate its empty)

src/dbnode/encoding/types.go
```diff
@@ -27,20 +27,33 @@ import (
...
 	"github.com/m3db/m3/src/x/checked"
 	"github.com/m3db/m3/src/x/ident"
 	"github.com/m3db/m3/src/x/pool"
+	"github.com/m3db/m3/src/x/serialize"
 	xtime "github.com/m3db/m3/src/x/time"
 )
 
+// StreamOpts is an options struct that can be passed to Encoder.Stream()
+// to modify its behavior.
+type StreamOpts struct {
```
robskillington: super nit: If you audit the other "options" structs we tend to name the vars "opts" but the actual struct itself as "FooOptions", so perhaps we can name this `StreamOptions`?

src/dbnode/encoding/types.go
```diff
@@ -27,20 +27,33 @@ import (
...
+	// Optional bytes into which the stream should be copied. If left
+	// nil then the checked bytes pool on the encoder will be used instead.
+	// TODO(rartoul): Actually use this field in the Stream API() in a subsequent
+	// P.R.
```
robskillington: Just a thought while here, we should just be able to pass `StreamOptions` up and down the call stack to Snapshot, Tick, etc yeah?  Obviously with parallel ticking we'll have to use a few and reuse them, but for Snapshot that should just be linear and could use a single one I would guess?

src/dbnode/storage/block/merged_block_reader.go
```diff
@@ -129,8 +129,18 @@ func (r *dbMergedBlockReader) mergedReader() (xio.BlockReader, error) {
...
 
+	var (
+		stream xio.SegmentReader
+		ok     bool
+	)
+	stream, ok = r.encoder.Stream(encoding.StreamOpts{})
+	if !ok {
+		// xio.BlockReader expects SegmentReader to be nil if there
+		// is no data.
+		stream = nil
```
robskillington: Hm, this seems counter-intuitive, instead of now possibly hitting a nil pointer case because we didn't have the `ok` var we now may do the wrong thing by default if it actually has no data.

Can we perhaps just make BlockReader do the same thing it used to do with a nil segmentreader before?

We could un-export all the fields of `BlockReader`, create a constructor, then basically inside the call to `SegmentReader()` on `BlockReader` enforce a nil return if it's instead a stream that's empty.

Or we can do other things to force the constructs at the type level rather than have to be careful when constructing one, like we're having to do here.

src/dbnode/encoding/types.go
richardartoul: Sure. I just did this because it helps keep things on one line since we have to declare a lot of empty versions of these but happy to change it

src/dbnode/encoding/types.go
richardartoul: Yeah flush and snapshot can use the same one passed up and down the stack.

Ticking is more complicated because you’re not copying temporarily, you iterate multiple streams in parallel and encode them into a new long-lives stream

src/dbnode/storage/block/merged_block_reader.go
richardartoul: Yeah sure, this was clearly me just being lazy after working on this P.R for awhile. I like the constructor approach, I’ll give that a shot

robskillington: Let me know how it goes, apologies that `BlockReader` doesn't already have a constructor =[

src/dbnode/encoding/types.go
robskillington: Hm yeah.

richardartoul: @robskillington Lol yeah we could do something tricky like that. I think what we really need is a utility method for performing merges that we can bake some of that crazier logic into. Another issue we have with merges is that basically if you're merging two large streams into one new stream there is no guessing regarding how big the new stream will be since you just write into an encoder one at a time.

If you had a utility we could use over the code base we could use the size of all the streams that are being merged to guess the size of the byte slice we should use to merge into so we don't have to allocate all the smaller slices along the way if that makes sense

robskillington: Agreed.  I do think that it's not so much worth worrying about how big they will be (and guest-imating that), but just reusing buffers that grow as necessary is ideal.  That way if it grows sure it needs to allocate, but if its reused across all series during a tick/merge then there's still zero allocations per series after amortizing cost.  

i.e. any time we need to do an allocation per merge series, we have a problem, as long as we reuse buffers, the allocationsPerMergeSeries approaches zero over any significant number of series.

src/dbnode/storage/block/merged_block_reader.go
```diff
@@ -129,8 +129,11 @@ func (r *dbMergedBlockReader) mergedReader() (xio.BlockReader, error) {
 		r.readers[i] = nil
 	}
 
+	// Can ignore OK here because BlockReader will handle nil streams
+	// properly.
```
robskillington: 👍

src/dbnode/encoding/types.go
richardartoul: Yeah sure allocations will go down but if we’re doing 5-6 grows per series (which is not unlikely and will be even higher with complex types workloads) we can make the merges faster because that’s 6 pool gets AND puts we don’t have to do + the actual time spent copying that we don’t have to do.

A micro optimization for sure, but wouldn’t be surprised if it made merging much faster

PR #1627, opened 2019-05-10
Expose shard level Tick() progress via Gauge

prateek:


src/dbnode/storage/shard.go
```diff
@@ -670,6 +675,8 @@ func (s *dbShard) tickAndExpire(
 					terminatedTickingDueToClosing = true
 					return false
 				}
+				// Expose shard level Tick() progress externally.
+				s.metrics.seriesTicked.Update(float64(i))
```
richardartoul: @prateek 

Two comments:

1. Anyway we can update this in batches as opposed to doing it for each one?
2. How how much does this really tell us? As far as I can tell this will only help us if the cycle exists within a single batch in its entirety. If the cycle spans a batch then we would never know even with this metric

prateek: 1: it is done per batch. This is within the if condition on line 664. So we only get a single update per batch. 

2: we’re getting an incrementing number for the metric. So for the x-batch case, we’d see a value continuously increasing. When that number hits let’s say a 1M and the node has <1M metrics, would be a pretty clear tell

richardartoul: Ok I looked at the code more carefully and I understand what its doing now, LGTM

PR #1628, opened 2019-05-11
[dbnode] Optimize index.Aggregate() for FieldQuery

### LINKED ISSUE #1620: Queries for aggregate values from a single large field is slow
With the new field matcher, we have sped up matching the documents matched by a query such as `__name__` matching `.*` because it is turned into a field search for index query search execution.

A problem remains however that with queries for aggregate values for very frequently appearing fields (i.e. `__name__`) where literally each metric (index document) has the field, instead of being able to return the values by walking the FST values, we actually return the postings lists and walk each document to see if tag value for that document is new/exists already in our aggregate result.

We can special case this in `index/block.go` at the top of the `block.Query` method, for queries that are single field matchers we can just walk the FST of each segment we have and add them to the aggregate results and return that.

It will significantly speed up queries for single field queries that match a very high amount of documents.

prateek:
Resolves #1620

### DIFF
--- a/src/dbnode/storage/index.go
+++ b/src/dbnode/storage/index.go
@@ -925,8 +925,9 @@ func (i *nsIndex) AggregateQuery(
 	field, isField := idx.FieldQuery(query.Query)
 	if isField {
 		fn = i.execBlockAggregateQueryFn
-		aopts.FieldFilter = append(index.AggregateFieldFilter{field}, aopts.FieldFilter...)
+		aopts.FieldFilter = aopts.FieldFilter.AddIfMissing(field)
 	}
+	aopts.FieldFilter = aopts.FieldFilter.SortAndDedupe()
 	results.Reset(i.nsMetadata.ID(), aopts)
 	exhaustive, err := i.query(ctx, query, results, opts.QueryOptions, fn)
 	if err != nil {
--- a/src/dbnode/storage/index.go
+++ b/src/dbnode/storage/index.go
@@ -909,17 +909,25 @@ func (i *nsIndex) AggregateQuery(
 ) (index.AggregateQueryResult, error) {
 	// Get results and set the filters, namespace ID and size limit.
 	results := i.aggregateResultsPool.Get()
-	results.Reset(i.nsMetadata.ID(), index.AggregateResultsOptions{
-		SizeLimit:  opts.Limit,
-		TermFilter: opts.TermFilter,
-		Type:       opts.Type,
-	})
+	aopts := index.AggregateResultsOptions{
+		SizeLimit:   opts.Limit,
+		FieldFilter: opts.FieldFilter,
+		Type:        opts.Type,
+	}
 	ctx.RegisterFinalizer(results)
 	// use appropriate fn to query underlying blocks.
+	// default to block.Query()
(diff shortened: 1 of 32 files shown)


src/dbnode/storage/index/block.go
```diff
@@ -894,6 +895,10 @@ func (b *block) Aggregate(
 	}
 
 	aggOpts := results.AggregateResultsOptions()
+	// ensure we iterate the provided filter fields in order.
+	sort.Slice(aggOpts.FieldFilter, func(i, j int) bool {
```
robskillington: Will this cause a data race since `block.Aggregate(...)` happens in parallel across blocks?

src/dbnode/storage/index.go
```diff
@@ -909,17 +909,25 @@ func (i *nsIndex) AggregateQuery(
...
+	// default to block.Query()
 	fn := i.execBlockQueryFn
+	// use block.Aggregate() when possible
 	if query.Equal(allQuery) {
 		fn = i.execBlockAggregateQueryFn
 	}
+	field, isField := idx.FieldQuery(query.Query)
+	if isField {
+		fn = i.execBlockAggregateQueryFn
+		aopts.FieldFilter = append(index.AggregateFieldFilter{field}, aopts.FieldFilter...)
```
robskillington: Hm only question here, is for the default query coming from query engine - don't we set both the FieldFilter and the TagMatcher turns into a FieldQuery? 

If so, would we add the same field twice to the field filter?  Would that mean iterating that field twice instead of once?  After reading the field iterator code it wasn't clear, also I'd need to trace the query engine code again, but I though it both sent a tag matcher that is a FieldQuery and might also set the FieldFilter too.

robskillington: Hm, seems like it does (from `tag_values.go`):
```go
	return &storage.CompleteTagsQuery{
		// NB: necessarily spans the entire timerange for the index.
		Start:            time.Time{},
		End:              h.nowFn(),
		CompleteNameOnly: false,
		FilterNameTags:   [][]byte{nameBytes},
		TagMatchers: models.Matchers{
			models.Matcher{
				Type:  models.MatchRegexp,
				Name:  nameBytes,
				Value: matchValues,
			},
		},
	}, nil
```

robskillington:
Awesome, thanks this is great!  Let me know what you think about my comments, not 100% sure they are all correct hah but did the best I could to think through it.

src/dbnode/storage/index/block.go
prateek: good catch, thanks.

prateek: addressed by moving the sort up into `index.go:Aggregate()` before we call into the blocks.

src/dbnode/storage/index.go
prateek: You're right - it would have been a duplicate iteration. Wouldn't have affected correctness but it's redundant. 

Made the change to only add the field filter if it doesn't exist in the list already, and further to de-dupe the filters to avoid this if users do something silly too.

prateek:
@robskillington ta, cheers! Addressed feedback.

PR #1633, opened 2019-05-14
Make NumFDs not allocate excessively, implement throttling, and disable Prometheus process collector

richardartoul:
Recent changes to M3DB mean that it uses many more F.Ds than before. This can result in counting the number of F.Ds allocating excessively and wasting a ton of CPU resources.

This P.R prevents counting the F.Ds from allocating excessively and implements self-throttling so that even though counting F.Ds takes longer, it will not cause spikes in CPU usage.

- [X] Make `NumFDs()` not allocate excessively 
- [X] Make `NumFDs()` throttle itself
- [x] Turn off FD counting in the Prometheus reporter

Screen Shot 2019-05-13 at 4 26 59 PM

Also verified with `start_m3` that metrics continue to be emitted

Screen Shot 2019-05-15 at 5 45 09 PM


src/x/instrument/config.go
```diff
@@ -85,6 +87,14 @@ func (mc *MetricsConfiguration) NewRootScope() (tally.Scope, io.Closer, error) {
 		if err != nil {
 			return nil, nil, err
 		}
+
+		// Turn off the default Prometheus process collector because it can be very
+		// slow if a lot of file descriptors are open.
+		toUnregister := promFork.NewProcessCollector(os.Getpid(), "")
```
richardartoul: @robskillington I don't think this is the right way to do this / does not seem to work. Any thoughts on how I can do this without changing `uber-go/tally` or do I have no choice?

Haijuncao:
@richardartoul  

It seems that we report num file descriptors during instrumentation and emit it from dbnode as application level stats.

I am wondering if we need to emit FD count/stat from application level, FD stat per process can be queried from os.

richardartoul:
@Haijuncao Yeah its really helpful since the instrument library gets loaded into a lot of our services. I.E we don't need any other software running on the host to get this information which can be really important for debugging internally and in O.S.S

It also makes it really easy to know which process it is.

src/x/instrument/config.go
```diff
@@ -25,6 +25,7 @@ import (
 	"io"
 	"time"
 
+	promfork "github.com/m3db/prometheus_client_golang/prometheus"
```
robskillington: super nit: Maybe just call it `prom` instead of `promfork`?  I know it is a fork, but the package and the module is still prom essentially.

src/x/instrument/extended.go
```diff
@@ -177,10 +178,40 @@ func (r *runtimeMetrics) report(metricsType ExtendedMetricsType) {
...
+		multiErr = multiErr.Add(err)
+	}
+
+	if e.processReporter != nil {
+		if err := e.processReporter.Stop(); err != nil {
+			multiErr = multiErr.Add(err)
+		}
+	}
+
+	return multiErr.FinalError()
```
robskillington: nit: Put `Start()` and `Stop()` below the constructor?

src/x/process/process_linux.go
```diff
@@ -24,12 +24,26 @@ package process
...
+
+var (
+	dotBytes       = []byte(".")
+	doubleDotBytes = []byte("..")
+)
+
+// NumFDsReference returns the number of file descriptors for a given process.
+// This is a reference implementation that can be used to compare against for
+// correctness.
+func NumFDsReference(pid int) (int, error) {
```
robskillington: nit: Perhaps call this just `NumFDsSlow()`?

src/x/process/process_linux_test.go
```diff
@@ -0,0 +1,145 @@
...
+	for i := 0; i <= 8; i++ {
+		var numFiles int
+		if i == 0 {
+			numFiles = 0
+		} else {
+			numFiles = int(math.Pow(float64(2), float64(i)))
+		}
+
+		func() {
+			numExpectedFds := numFiles + 5
```
robskillington: Hm, why the plus 5 here?

src/x/process/process_linux_test.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+}
+
+func BenchmarkNumFDs(b *testing.B) {
+	var (
+		// Low for C.I and local testing, bump this up to a much larger number
+		// when performing actual benchmarking.
+		numFiles = 16000
+		// +5 to account for standard F.Ds that each process gets.
+		numExpectedFds = numFiles + 5
```
robskillington: Can use the const here as mentioned above.

src/x/process/process_notlinux.go
```diff
@@ -29,7 +29,20 @@ import (
...
-// NumFDs returns the number of file descriptors for a given process.
+// NumFDs returns the number of file descriptors for a given process and is not available
+// on non-linux systems.
 func NumFDs(pid int) (int, error) {
 	return 0, errNotAvailable
 }
+
+// NumFDsWithDefaultBatchSleep returns the number of file descriptors for a given process
+// and is not available on non-linux systems.
+func NumFDsWithDefaultBatchSleep(pid int) (int, error) {
```
robskillington: Don't you also need `NumFDsReference` (or whatever that gets renamed to) from the non-linux source file?  Otherwise it won't be able to build on non-linux since it will be missing.

richardartoul: ended up not exporting it

PR #1637, opened 2019-05-16
Batch network writes in aggregator client

vdarulis:
Batch writes to aggregator instances, as the current outgoing average packet size from clients is less than 800 bytes and batching increases it to nearly full link utilisation. Batching would also help for high-latency, high-bandwidth links (eg. cross-region traffic), as the current setup pretty much guarantees inefficient use of TCP streams.

There's also a few more potential optimisations available with the rewritten drain() function, as we no longer need to batch messages during encoding.


src/aggregator/client/queue.go
```diff
@@ -110,20 +110,25 @@ type writeFn func([]byte) error
...
+	buf           []byte
+	writeInterval time.Duration
+	wg            sync.WaitGroup
 
 	writeFn writeFn
 }
 
 func newInstanceQueue(instance placement.Instance, opts Options) instanceQueue {
+	const batchSize = 16384
+	const minWriteInterval = 250 * time.Millisecond
```
robskillington: super nits:
- 16384 can become: `batchSize = 2 << 13 // 16k`
- combine into single `const` stanza
```go
const (
  batchSize = 2 << 13 // 16k
  minWriteInterval = 250 * time.Millisecond
)
```

robskillington: Also, I'd feel more comfortable with 100ms than 250ms, do you think it will make a huge difference?  Any level of buffering will be big.

src/aggregator/client/queue.go
```diff
@@ -142,10 +147,27 @@ func newInstanceQueue(instance placement.Instance, opts Options) instanceQueue {
...
+	var (
+		ticker *time.Ticker
+		tC     <-chan (time.Time)
+	)
+	if batchSize > 0 && minWriteInterval != 0 {
+		q.buf = make([]byte, 0, batchSize)
+		q.writeInterval = minWriteInterval
+		q.batchSize = batchSize
+		ticker = time.NewTicker(minWriteInterval)
+		tC = ticker.C
```
robskillington: Hm, seems better to pass `minWriteInterval` to `drain` function itself and let it do the ticker creation and stop itself.

i.e.

```go
	if batchSize > 0 && minWriteInterval != 0 {
		q.buf = make([]byte, 0, batchSize)
		q.writeInterval = minWriteInterval
		q.batchSize = batchSize
	}
	q.writeFn = q.conn.Write

	q.wg.Add(2)

	go q.drain(minWriteInterval)

	// ....


func (q *queue) drain(minWriteInterval time.Duration) {
	var tickCh <-chan (time.Time)
	if minWriteInterval >= 0 {
		ticker := time.NewTicker(minWriteInterval)
		defer ticker.Close()
		tickCh = ticker.C
	}

	// as per normal...
}
		
```

src/aggregator/client/queue.go
```diff
@@ -142,10 +147,27 @@ func newInstanceQueue(instance placement.Instance, opts Options) instanceQueue {
 		bufCh:    make(chan protobuf.Buffer, queueSize),
 		doneCh:   make(chan struct{}),
 	}
+
+	var (
+		ticker *time.Ticker
+		tC     <-chan (time.Time)
+	)
+	if batchSize > 0 && minWriteInterval != 0 {
```
robskillington: Perhaps best to always assume > 0 for these consts?  And if they aren't then return an error on construction or something?  Can simplify the code a lot if so.

src/aggregator/client/queue.go
```diff
@@ -204,22 +223,61 @@ func (q *queue) Close() error {
...
+		case qitem := <-q.bufCh:
+			var needFlush bool
+			msg := qitem.Bytes()
+			q.buf = append(q.buf, msg...)
+			qitem.Close()
+
+			if len(q.buf) > q.batchSize {
+				needFlush = true
+			} else {
+				needFlush = time.Since(lastDrain) >= q.writeInterval
```
robskillington: Hm, shouldn't this be the job of the ticker below?  Seems strange to have both potentially do time based flushing (separation of concerns and all).

src/aggregator/client/queue.go
```diff
@@ -204,22 +223,61 @@ func (q *queue) Close() error {
...
+				continue
+			}
+
+			if err := q.writeFn(q.buf); err != nil {
+				q.log.Error("error writing data", zap.String("instance", q.instance.Endpoint()), zap.Error(err))
+				q.metrics.connWriteErrors.Inc(1)
+			} else {
+				q.metrics.connWriteSuccesses.Inc(1)
+			}
+			qitem.Close()
```
robskillington: This looks like a double close, don't you close this on line 237?

src/aggregator/client/queue.go
```diff
@@ -204,22 +223,61 @@ func (q *queue) Close() error {
...
+			if !needFlush {
+				continue
+			}
+
+			if err := q.writeFn(q.buf); err != nil {
+				q.log.Error("error writing data", zap.String("instance", q.instance.Endpoint()), zap.Error(err))
+				q.metrics.connWriteErrors.Inc(1)
+			} else {
+				q.metrics.connWriteSuccesses.Inc(1)
+			}
```
robskillington: I would extract this method so you can call it once from both this part of the loop and below.

I understand you think it might prevent inlining, but I don't think it's quite enough in the hot path to be an issue.  There's plenty of function calls per metric already in this code path.

src/aggregator/client/queue.go
```diff
@@ -204,22 +223,61 @@ func (q *queue) Close() error {
...
+			qitem.Close()
+
+			q.buf = q.buf[:0]
+			lastDrain = time.Now()
+		case ts := <-tC:
+			delta := ts.Sub(lastDrain)
+			lastDrain = ts
+			if delta < q.writeInterval || len(q.buf) == 0 {
+				continue
```
robskillington: So only problem I see here using a ticker, is that if you wake up and you recently flushed because of size limit so you `continue` you can potentially sleep a whole other period (i.e. 100ms) before you potentially send again (i.e. you send the next batch 150ms later if you sent a packet 50ms into the first 100ms period and you never wrote anything afterwards).

I know you cover some of this case with `needFlush = time.Since(lastDrain) >= q.writeInterval` when you receive a buffer, however if you never receive another buffer and you didn't hit the size limit before you begin the second minWriteInterval time period then it doesn't get sent until next ticker tick.

So worst case is you approach 2x minWriteInterval for this pathological case.  It's probably not that bad, I just wanted to call it out.

The other option is to subtract the difference when when it was last drained, then specifically sleep for minWriteInterval-delta and wake up again.  That way you are putting a max deadline on when data is sent.

Either way, just wanted to document the differences between the two strategies.

src/aggregator/client/queue.go
vdarulis: Oops

src/aggregator/client/queue.go
vdarulis: Yeah this is just a PoC without any config plumbing, will fix and adjust (both defaults and init logic).

src/aggregator/client/options.go
```diff
@@ -50,6 +50,12 @@ const (
 
 	// By default the oldest metrics in the queue are dropped when it is full.
 	defaultDropType = DropOldest
+
+	// By default set maximum batch size to 32k
+	defaultMaxBatchSize = 2 << 15
```
richardartoul: This is 65k

https://play.golang.org/p/73FAueZnIdX

src/aggregator/client/queue.go
```diff
@@ -132,15 +135,20 @@ func newInstanceQueue(instance placement.Instance, opts Options) instanceQueue {
...
+		dropType:           opts.QueueDropType(),
+		log:                iOpts.Logger(),
+		metrics:            newQueueMetrics(iOpts.MetricsScope(), queueSize),
+		instance:           instance,
+		conn:               conn,
+		bufCh:              make(chan protobuf.Buffer, queueSize),
+		doneCh:             make(chan struct{}),
+		maxBatchSize:       maxBatchSize,
+		batchFlushDeadline: writeInterval,
+		buf:                make([]byte, 0, maxBatchSize),
```
richardartoul: Is this one queue per backend * host?

src/aggregator/client/queue.go
```diff
@@ -204,22 +209,61 @@ func (q *queue) Close() error {
 	return nil
 }
 
+func (q *queue) writeAndReset() {
+	if len(q.buf) == 0 {
+		return
+	}
+	if err := q.writeFn(q.buf); err != nil {
+		q.log.Error("error writing data on timed flush",
+			zap.String("instance", q.instance.Endpoint()),
```
richardartoul: Can you make this "target instance" so just thats its clear its the host you're trying to write to, not from

src/aggregator/client/queue.go
```diff
@@ -204,22 +209,61 @@ func (q *queue) Close() error {
...
+		q.metrics.connWriteSuccesses.Inc(1)
+	}
+	q.buf = q.buf[:0]
+}
+
 func (q *queue) drain() {
 	defer q.wg.Done()
+	defer q.conn.Close()
+	timer := time.NewTimer(q.batchFlushDeadline)
+	lastDrain := time.Now()
```
richardartoul: This is a super long lived function (background goroutine, right?)

In that case it seems reasonable to allocate a small helper function that you can use over and over:

```golang
write := func() {
    q.writeAndReset()
    lastDrain = time.Now()
}
```

src/aggregator/client/queue.go
```diff
@@ -204,22 +209,61 @@ func (q *queue) Close() error {
...
+	} else {
+		q.metrics.connWriteSuccesses.Inc(1)
+	}
+	q.buf = q.buf[:0]
+}
+
 func (q *queue) drain() {
 	defer q.wg.Done()
+	defer q.conn.Close()
+	timer := time.NewTimer(q.batchFlushDeadline)
```
richardartoul: @prateek I think Prateek can chime in on this but apparently its better to use `time.Sleep` because they have some runtime integration which makes them more likely to fire on time as opposed to timers (which can get super delayed)

src/aggregator/client/queue_test.go
```diff
@@ -43,17 +43,23 @@ func TestInstanceQueueEnqueueClosed(t *testing.T) {
...
 
 	// Fill up the queue and park the draining goroutine so the queue remains full.
 	queue.writeFn = func([]byte) error {
 		select {}
 	}
-	queue.bufCh <- testNewBuffer(nil)
-	queue.bufCh <- testNewBuffer(nil)
-	require.Equal(t, errWriterQueueFull, queue.Enqueue(testNewBuffer(nil)))
+
+	time.Sleep(5 * time.Millisecond)
```
richardartoul: Was this test flaky before? Would really like to avoid having tests that depends on sleeps

src/aggregator/client/queue.go
vdarulis: one per shard x instance, from the looks of it

src/aggregator/client/queue_test.go
vdarulis: yeah it was super flaky, I can just rework it

src/aggregator/client/queue.go
vdarulis: I'm wondering whether it's worth changing this to use sleeps - I'll have to add another goroutine (and we've got a bunch of them already). On any reasonably active connection we'll still be flushing in the normal write path based on time since last drain.

src/aggregator/client/queue.go
```diff
@@ -204,22 +209,63 @@ func (q *queue) Close() error {
...
-				zap.Error(err),
-			)
-			q.metrics.connWriteErrors.Inc(1)
-		} else {
-			q.metrics.connWriteSuccesses.Inc(1)
+			write()
+		case ts := <-timer.C:
+			delta := ts.Sub(lastDrain)
+			if delta < q.batchFlushDeadline {
+				timer.Reset(q.batchFlushDeadline - delta)
```
robskillington: Nice, didn't even know this was possible.

PR #1638, opened 2019-05-16
[query] Update offset to accept opts

benraskin92:

This is the first of two PRs to support Unary Expressions (`+` and `-`). Instead of creating a `lazyBlock` (formerly called `offsetBlock`) with just an offset duration, it now takes in a `LazyOpts` struct consisting of 3 transform functions. This way, we can support offsetting by time (current), metadata (current) and value (future). 

```
// TimeTransform transforms a timestamp
type TimeTransform func(time.Time) time.Time

// MetaTranform transforms meta data
type MetaTranform func(meta Metadata) Metadata

// ValueTransform transform a float64
type ValueTransform func(float64) float64

// LazyOpts are used to create a lazy block
type LazyOpts struct {
	TimeTransform  TimeTransform
	MetaTransform  MetaTranform
	ValueTransform ValueTransform
}
```

```release-note
No user impacting changes for this PR. In the next PR that introduces unary expressions, we will add a release note. 
```

```documentation-note

```


src/query/block/lazy.go
```diff
@@ -0,0 +1,303 @@
...
+// TimeTransform transforms a timestamp
+type TimeTransform func(time.Time) time.Time
+
+// MetaTranform transforms meta data
+type MetaTranform func(meta Metadata) Metadata
+
+// ValueTransform transform a float64
+type ValueTransform func(float64) float64
+
+// LazyOpts are used to create a lazy block
```
arnikola: Rather than doing it like this, can we refactor LazyOpts be an interface [like this one](https://sourcegraph.com/github.com/m3db/m3/-/blob/src/query/models/options.go#L85:12), and have the transforms default to a noop function?

src/query/block/lazy.go
```diff
@@ -0,0 +1,303 @@
+// Copyright (c) 2018 Uber Technologies, Inc.
```
arnikola: nit; 2019

src/query/block/lazy.go
```diff
@@ -123,13 +130,13 @@ func (it *lazyStepIter) SeriesMeta() []SeriesMeta { return it.it.SeriesMeta() }
 func (it *lazyStepIter) Next() bool               { return it.it.Next() }
 
 func (it *lazyStepIter) Meta() Metadata {
-	return it.opts.MetaTransform(it.it.Meta())
+	return it.opts.MetaTransform()(it.it.Meta())
```
arnikola: This looks super gross haha, don't think there's anything we can do about it though

src/query/block/lazy.go
```diff
@@ -21,62 +21,69 @@
...
+type lazyOpts struct {
+	timeTransform  TimeTransform
+	metaTransform  MetaTransform
+	valueTransform ValueTransform
+}
 
-// ValueTransform transform a float64
-type ValueTransform func(float64) float64
+// NewLazyOpts creates LazyOpts with default values
+func NewLazyOpts() LazyOpts {
```
arnikola: nit: Can you rename it to `LazyOptions`, also maybe split it out into its own file

src/query/block/lazy.go
```diff
@@ -21,62 +21,69 @@
...
-type MetaTranform func(meta Metadata) Metadata
+type lazyOpts struct {
+	timeTransform  TimeTransform
+	metaTransform  MetaTransform
+	valueTransform ValueTransform
+}
 
-// ValueTransform transform a float64
-type ValueTransform func(float64) float64
+// NewLazyOpts creates LazyOpts with default values
```
arnikola: nit: full stop at the end of comment

src/query/block/types.go
```diff
@@ -183,3 +183,28 @@ func TakeLast(values ts.Datapoints) float64 {
...
+
+// MetaTransform transforms meta data
+type MetaTransform func(meta Metadata) Metadata
+
+// ValueTransform transform a float64
+type ValueTransform func(float64) float64
+
+// LazyOpts describes options for lazy blocks.
+type LazyOpts interface {
+	// SetTimeTransform sets the time transform function
```
arnikola: nit: full stops at the end of comments

src/query/functions/lazy/base_test.go
```diff
@@ -44,9 +44,18 @@ func buildMeta(start time.Time) block.Metadata {
 	}
 }
 
+func testLazyOpts(offset time.Duration) block.LazyOpts {
+	tt := func(t time.Time) time.Time { return t.Add(offset) }
+	mt := func(meta block.Metadata) block.Metadata {
+		meta.Bounds.Start = meta.Bounds.Start.Add(offset)
+		return meta
+	}
+	return block.NewLazyOpts().SetTimeTransform(tt).SetMetaTransform(mt)
```
arnikola: nit; newline

src/query/block/lazy_test.go
```diff
@@ -44,38 +44,37 @@ func buildMeta(start time.Time) Metadata {
 	}
 }
 
+func testLazyOpts(offset time.Duration) LazyOpts {
+	tt := func(t time.Time) time.Time { return t.Add(offset) }
+	mt := func(meta Metadata) Metadata {
+		meta.Bounds.Start = meta.Bounds.Start.Add(offset)
+		return meta
+	}
+	return NewLazyOpts().SetTimeTransform(tt).SetMetaTransform(mt)
```
arnikola: nit; newline

src/query/parser/promql/parse.go
```diff
@@ -82,11 +84,29 @@ func (p *parseState) transformLen() int {
...
-func (p *parseState) addLazyTransform(lazyOpts *block.LazyOpts) error {
-	if lazyOpts == nil {
+func (p *parseState) addLazyTransform(offset time.Duration) error {
+	// NB: if offset is <= 0, we do not apply any offsets.
+	switch {
+	case offset == 0:
+		return nil
+	case offset < 0:
+		// todo(braskin): add zap logger here.
+		log.Printf("offset must be positive, received: %v. not applying offset", offset)
```
arnikola: I think instead of this may be easier to error out? It's not really a viable use case anyway since the Prom parser will error with negative offsets

benraskin92: Yeah thats what I was thinking.. glad we're on the same page.

robskillington: I don't think we should land this without a zap Logger passed in here.  Can you please find a way to propagate a logger to the `parseState`?

benraskin92: Returning an error instead of logging.

src/query/block/lazy.go
```diff
@@ -0,0 +1,259 @@
...
+	meta Metadata,
+	sm []SeriesMeta,
+) (Block, error) {
+	bl, err := b.block.WithMetadata(meta, sm)
+	if err != nil {
+		return nil, err
+	}
+
+	b.block = bl
+	return b, nil
```
robskillington: Hm, instead of returning yourself and mutating this block I would instead Clone.

i.e.
```go
func (b *lazyBlock) WithMetadata(
	meta Metadata,
	sm []SeriesMeta,
) (Block, error) {
	bl, err := b.block.WithMetadata(meta, sm)
	if err != nil {
		return nil, err
	}

	return NewLazyBlock(bl, bl.opts), nil
}
```

src/query/block/lazy.go
robskillington: Yeah can you make this look slightly better where every time you call it you do:
```go
meta := it.opts.MetaTransform()
return meta(it.it.Meta())
```

src/query/block/lazy.go
```diff
@@ -0,0 +1,259 @@
...
+func (it *lazyStepIter) Next() bool               { return it.it.Next() }
+
+func (it *lazyStepIter) Meta() Metadata {
+	return it.opts.MetaTransform()(it.it.Meta())
+}
+
+func (it *lazyStepIter) Current() Step {
+	c := it.it.Current()
+	return ColStep{
+		time:   it.opts.TimeTransform()(c.Time()),
```
robskillington: Here too, can you move the method into a var name for readability:
```go
time := it.opts.TimeTransform()
return ColStep{time: time(c.Time()), ...}
```

src/query/block/lazy.go
```diff
@@ -0,0 +1,259 @@
...
+	meta Metadata,
+	sm []SeriesMeta,
+) (UnconsolidatedBlock, error) {
+	bl, err := b.block.WithMetadata(meta, sm)
+	if err != nil {
+		return nil, err
+	}
+
+	b.block = bl
+	return b, nil
```
robskillington: Same here, return a clone of the current block instead of mutating

src/query/block/lazy.go
```diff
@@ -0,0 +1,259 @@
...
+}
+
+func (it *ucLazyStepIter) Close()                   { it.it.Close() }
+func (it *ucLazyStepIter) Err() error               { return it.it.Err() }
+func (it *ucLazyStepIter) StepCount() int           { return it.it.StepCount() }
+func (it *ucLazyStepIter) SeriesMeta() []SeriesMeta { return it.it.SeriesMeta() }
+func (it *ucLazyStepIter) Next() bool               { return it.it.Next() }
+
+func (it *ucLazyStepIter) Meta() Metadata {
+	return it.opts.MetaTransform()(it.it.Meta())
```
robskillington: Same here, extract the method into a local var name for readability?

src/query/block/lazy.go
```diff
@@ -0,0 +1,259 @@
...
+func (s unconsolidatedStep) Values() []ts.Datapoints {
+	return s.values
+}
+
+func (it *ucLazyStepIter) Current() UnconsolidatedStep {
+	c := it.it.Current()
+	for _, val := range c.Values() {
+		for i, dp := range val.Datapoints() {
+			val[i].Timestamp = it.opts.TimeTransform()(dp.Timestamp)
+			val[i].Value = it.opts.ValueTransform()(dp.Value)
```
robskillington: Same here, extract the method into a local var name for readability?

src/query/block/lazy.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+	}
+
+	return unconsolidatedStep{
+		time:   it.opts.TimeTransform()(c.Time()),
```
robskillington: Same here, extract the method into a local var name for readability?

src/query/block/lazy.go
```diff
@@ -0,0 +1,259 @@
...
+func (it *ucLazySeriesIter) Err() error               { return it.it.Err() }
+func (it *ucLazySeriesIter) SeriesCount() int         { return it.it.SeriesCount() }
+func (it *ucLazySeriesIter) SeriesMeta() []SeriesMeta { return it.it.SeriesMeta() }
+func (it *ucLazySeriesIter) Next() bool               { return it.it.Next() }
+func (it *ucLazySeriesIter) Current() UnconsolidatedSeries {
+	c := it.it.Current()
+	for _, val := range c.datapoints {
+		for i, dp := range val.Datapoints() {
+			val[i].Timestamp = it.opts.TimeTransform()(dp.Timestamp)
+			val[i].Value = it.opts.ValueTransform()(dp.Value)
```
robskillington: Same here, extract the method into a local var name for readability?

src/query/block/lazy.go
```diff
@@ -0,0 +1,259 @@
...
+		}
+	}
+
+	return c
+}
+
+func (it *ucLazySeriesIter) Meta() Metadata {
+	return it.opts.MetaTransform()(it.it.Meta())
```
robskillington: Same here, extract the method into a local var name for readability?

src/query/block/types.go
```diff
@@ -183,3 +183,28 @@ func TakeLast(values ts.Datapoints) float64 {
...
+type TimeTransform func(time.Time) time.Time
+
+// MetaTransform transforms meta data.
+type MetaTransform func(meta Metadata) Metadata
+
+// ValueTransform transform a float64.
+type ValueTransform func(float64) float64
+
+// LazyOpts describes options for lazy blocks.
+type LazyOpts interface {
```
robskillington: nit: All our options types have the full name, this should be `LazyOptions` instead of `LazyOpts` for consistency with the other option types names.

PR #1642, opened 2019-05-17
Optimize protobuf encoder and iterator for various workloads

richardartoul:
Remove `jhump/protoreflect` from the hot path as much as possible, plus some other optimizations and cleanup.

Results:

1. 10x improvement in encoding and decoding for non-custom fields.
2. 2x improvement in encoding for custom fields.
3. 4x improvement in decoding for custom fields (original goal).

before
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/encoding/proto
BenchmarkEncoder/with_non_custom_encoded_fields_enabled-8         	    1000	   1432754 ns/op	  374435 B/op	    8177 allocs/op
BenchmarkEncoder/with_non_custom_encoded_fields_disabled-8        	   10000	    111547 ns/op	   13430 B/op	       0 allocs/op
BenchmarkIterator/with_non_custom_encoded_fields_enabled-8        	    1000	   1274402 ns/op	  323059 B/op	    8196 allocs/op
BenchmarkIterator/with_non_custom_encoded_fields_disabled-8       	   10000	    203996 ns/op	   21603 B/op	     910 allocs/op
```

after
```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/encoding/proto
BenchmarkEncoder/with_non_custom_encoded_fields_enabled-8         	   10000	    119435 ns/op	   66221 B/op	       0 allocs/op
BenchmarkEncoder/with_non_custom_encoded_fields_disabled-8        	   20000	     67853 ns/op	   21176 B/op	       0 allocs/op
BenchmarkIterator/with_non_custom_encoded_fields_enabled-8        	   20000	     90630 ns/op	     671 B/op	      13 allocs/op
BenchmarkIterator/with_non_custom_encoded_fields_disabled-8       	   30000	     46309 ns/op	     595 B/op	      12 allocs/op
```

### DIFF
--- a/src/dbnode/encoding/m3tsz/float_encoder_iterator.go
+++ b/src/dbnode/encoding/m3tsz/float_encoder_iterator.go
@@ -140,16 +140,13 @@ func (eit *FloatEncoderAndIterator) readNextFloat(stream encoding.IStream) error
 		return nil
 	}
 
-	numLeadingZeros, err := stream.ReadBits(6)
+	numLeadingZeroesAndNumMeaningfulBits, err := stream.ReadBits(12)
 	if err != nil {
 		return err
 	}
 
-	numMeaningfulBits, err := stream.ReadBits(6)
-	if err != nil {
-		return err
-	}
-	numMeaningfulBits = numMeaningfulBits + 1
+	numLeadingZeros := (numLeadingZeroesAndNumMeaningfulBits & 4032) >> 6
+	numMeaningfulBits := (numLeadingZeroesAndNumMeaningfulBits & 63) + 1
 
 	meaningfulBits, err := stream.ReadBits(int(numMeaningfulBits))
 	if err != nil {
--- a/src/dbnode/encoding/m3tsz/float_encoder_iterator.go
+++ b/src/dbnode/encoding/m3tsz/float_encoder_iterator.go
@@ -26,6 +26,11 @@ import (
 	"github.com/m3db/m3/src/dbnode/encoding"
 )
 
+const (
+	bits12To6Mask = 4032 // 1111 1100 0000
+	bits6To0Mask  = 63   // 0011 1111
+)
+
 // FloatEncoderAndIterator encapsulates the state required for a logical stream of bits
 // that represent a stream of float values compressed with XOR.
 type FloatEncoderAndIterator struct {
@@ -145,8 +150,8 @@ func (eit *FloatEncoderAndIterator) readNextFloat(stream encoding.IStream) error
 		return err
 	}
 
-	numLeadingZeros := (numLeadingZeroesAndNumMeaningfulBits & 4032) >> 6
-	numMeaningfulBits := (numLeadingZeroesAndNumMeaningfulBits & 63) + 1
+	numLeadingZeros := (numLeadingZeroesAndNumMeaningfulBits & bits12To6Mask) >> 6
+	numMeaningfulBits := (numLeadingZeroesAndNumMeaningfulBits & bits6To0Mask) + 1
 
 	meaningfulBits, err := stream.ReadBits(int(numMeaningfulBits))
 	if err != nil {
--- a/src/dbnode/encoding/ostream.go
+++ b/src/dbnode/encoding/ostream.go
@@ -155,10 +155,6 @@ func (os *ostream) WriteByte(v byte) {
 
 // WriteBytes writes a byte slice.
 func (os *ostream) WriteBytes(bytes []byte) {
-	// Make sure we only have to grow the underlying buffer at most
-	// one time before we write one byte at a time in a loop.
-	os.ensureCapacityFor(len(bytes))
-
 	if !os.hasUnusedBits() {
 		// If the stream is aligned on a byte boundary then all of the WriteByte()
 		// function calls and bit-twiddling can be skipped in favor of a single
@@ -167,6 +163,9 @@ func (os *ostream) WriteBytes(bytes []byte) {
 		return
 	}
 
+	// Make sure we only have to grow the underlying buffer at most
+	// one time before we write one byte at a time in a loop.
+	os.ensureCapacityFor(len(bytes))
 	for i := 0; i < len(bytes); i++ {
 		os.WriteByte(bytes[i])
 	}
--- a/src/dbnode/encoding/ostream.go
+++ b/src/dbnode/encoding/ostream.go
@@ -159,6 +159,14 @@ func (os *ostream) WriteBytes(bytes []byte) {
 	// one time before we write one byte at a time in a loop.
 	os.ensureCapacityFor(len(bytes))
 
+	if !os.hasUnusedBits() {
+		// If the stream is aligned on a byte boundary then all of the WriteByte()
+		// function calls and bit-twiddling can be skipped in favor of a single
+		// memcopy.
+		os.rawBuffer = append(os.rawBuffer, bytes...)
+		return
+	}
+
 	for i := 0; i < len(bytes); i++ {
 		os.WriteByte(bytes[i])
 	}
--- a/src/dbnode/encoding/ostream.go
+++ b/src/dbnode/encoding/ostream.go
@@ -155,17 +155,23 @@ func (os *ostream) WriteByte(v byte) {
 
 // WriteBytes writes a byte slice.
 func (os *ostream) WriteBytes(bytes []byte) {
+	// Call ensureCapacityFor ahead of time to ensure that the bytes pool is used to
+	// grow the rawBuffer (as opposed to append possibly triggering an allocation if
+	// it wasn't) and that its only grown a maximum of one time regardless of the size
+	// of the []byte being written.
+	os.ensureCapacityFor(len(bytes))
+
 	if !os.hasUnusedBits() {
 		// If the stream is aligned on a byte boundary then all of the WriteByte()
 		// function calls and bit-twiddling can be skipped in favor of a single
-		// memcopy.
+		// copy operation.
 		os.rawBuffer = append(os.rawBuffer, bytes...)
+		// Position 8 indicates that the last byte of the buffer has been completely
+		// filled.
+		os.pos = 8
 		return
 	}
 
-	// Make sure we only have to grow the underlying buffer at most
-	// one time before we write one byte at a time in a loop.
-	os.ensureCapacityFor(len(bytes))
 	for i := 0; i < len(bytes); i++ {
 		os.WriteByte(bytes[i])
 	}
--- a/src/dbnode/encoding/ostream_test.go
+++ b/src/dbnode/encoding/ostream_test.go
@@ -89,7 +89,7 @@ func testWriteBytes(t *testing.T, o OStream) {
 	b, _ := os.Rawbytes()
 	require.Equal(t, rawBytes, b)
 
-	require.Equal(t, 8, os.pos)
+	require.Equal(t, 0, os.pos)
 }
 
 func TestResetOStream(t *testing.T) {
--- a/src/dbnode/encoding/ostream_test.go
+++ b/src/dbnode/encoding/ostream_test.go
@@ -86,10 +86,10 @@ func testWriteBytes(t *testing.T, o OStream) {
 
 	require.Equal(t, rawBytes, os.rawBuffer)
 
-	b, _ := os.Rawbytes()
+	b, pos := os.Rawbytes()
 	require.Equal(t, rawBytes, b)
-
-	require.Equal(t, 0, os.pos)
+	require.Equal(t, 8, pos)
+	require.Equal(t, 8, os.pos)
 }
 
 func TestResetOStream(t *testing.T) {
--- a/src/dbnode/encoding/proto/buffer_encode.go
+++ b/src/dbnode/encoding/proto/buffer_encode.go
@@ -66,14 +66,6 @@ func (cb *buffer) encodeFixed32(x uint32) {
 		uint8(x>>24))
 }
 
-func encodeZigZag64(v int64) uint64 {
-	return (uint64(v) << 1) ^ uint64(v>>63)
-}
-
-func encodeZigZag32(v int32) uint64 {
-	return uint64((uint32(v) << 1) ^ uint32((v >> 31)))
-}
-
 // EncodeRawBytes writes a count-delimited byte buffer to the Buffer.
 // This is the format used for the bytes protocol buffer
 // type and for embedded messages.
@@ -84,4 +76,12 @@ func (cb *buffer) encodeRawBytes(b []byte) {
 
 func (cb *buffer) append(b []byte) {
 	cb.buf = append(cb.buf, b...)
+}
+
+func encodeZigZag32(v int32) uint64 {
+	return uint64((uint32(v) << 1) ^ uint32((v >> 31)))
+}
+
+func encodeZigZag64(v int64) uint64 {
+	return (uint64(v) << 1) ^ uint64(v>>63)
 }
--- a/src/dbnode/encoding/proto/common.go
+++ b/src/dbnode/encoding/proto/common.go
@@ -42,9 +42,11 @@ const (
 
 	// maxCustomFieldNum is included for the same rationale as maxMarshaledProtoMessageSize.
 	maxCustomFieldNum = 10000
+
+	protoFieldTypeNotFound dpb.FieldDescriptorProto_Type = -1
 )
 
-type customFieldType int
+type customFieldType int8
 
 const (
 	// All the protobuf field types that we can perform custom encoding /
@@ -144,8 +146,9 @@ var (
 // customFieldState is used to track any required state for encoding / decoding a single
 // field in the encoder / iterator respectively.
 type customFieldState struct {
-	fieldNum  int
-	fieldType customFieldType
+	fieldNum       int
+	fieldType      customFieldType
+	protoFieldType dpb.FieldDescriptorProto_Type
 
 	// Float state. Works as both an encoder and iterator (I.E the encoder calls
 	// the encode methods and the iterator calls the read methods).
@@ -167,13 +170,20 @@ type encoderBytesFieldDictState struct {
 	// by comparing the bytes against those we already wrote into the
 	// stream.
 	hash     uint64
-	startPos int
-	length   int
+	startPos uint32
+	length   uint32
 }
 
-func newCustomFieldState(fieldNum int, fieldType customFieldType) customFieldState {
-	s := customFieldState{fieldNum: fieldNum, fieldType: fieldType}
-	if isUnsignedInt(fieldType) {
+func newCustomFieldState(
+	fieldNum int,
+	protoFieldType dpb.FieldDescriptorProto_Type,
+	customFieldType customFieldType,
+) customFieldState {
+	s := customFieldState{
+		fieldNum:       fieldNum,
+		fieldType:      customFieldType,
+		protoFieldType: protoFieldType}
+	if isUnsignedInt(customFieldType) {
 		s.intEncAndIter.unsigned = true
 	}
 	return s
@@ -201,21 +211,34 @@ func customAndProtoFields(s []customFieldState, protoFields []int32, schema *des
 		protoFields = make([]int32, 0, numProtoFields)
 	}
 
+	var (
+		prevFieldNum int32 = -1
+		isSorted           = true
+	)
 	for _, field := range fields {
-		customFieldType, ok := isCustomField(field.GetType(), field.IsRepeated())
+		var (
+			fieldType = field.GetType()
+			fieldNum  = field.GetNumber()
+		)
+		if fieldNum < prevFieldNum {
+			isSorted = false
+		}
+
+		customFieldType, ok := isCustomField(fieldType, field.IsRepeated())
 		if !ok {
-			protoFields = append(protoFields, field.GetNumber())
+			protoFields = append(protoFields, fieldNum)
 			continue
 		}
 
-		fieldState := newCustomFieldState(int(field.GetNumber()), customFieldType)
+		fieldState := newCustomFieldState(int(fieldNum), fieldType, customFieldType)
 		s = append(s, fieldState)
 	}
 
-	// Should already be sorted by fieldNum, but do it again just to be sure.
-	sort.Slice(s, func(a, b int) bool {
-		return s[a].fieldNum < s[b].fieldNum
-	})
+	if !isSorted {
+		sort.Slice(s, func(a, b int) bool {
+			return s[a].fieldNum < s[b].fieldNum
+		})
+	}
 
 	return s, protoFields
 }
--- a/src/dbnode/encoding/proto/common.go
+++ b/src/dbnode/encoding/proto/common.go
@@ -153,21 +153,18 @@ type sortedMarshalledFields []marshalledField
 // customFieldState is used to track any required state for encoding / decoding a single
 // field in the encoder / iterator respectively.
 type customFieldState struct {
-	fieldNum       int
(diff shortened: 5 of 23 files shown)


src/dbnode/encoding/m3tsz/float_encoder_iterator.go
```diff
@@ -140,16 +140,13 @@ func (eit *FloatEncoderAndIterator) readNextFloat(stream encoding.IStream) error
...
 		return err
 	}
 
-	numMeaningfulBits, err := stream.ReadBits(6)
-	if err != nil {
-		return err
-	}
-	numMeaningfulBits = numMeaningfulBits + 1
+	numLeadingZeros := (numLeadingZeroesAndNumMeaningfulBits & 4032) >> 6
+	numMeaningfulBits := (numLeadingZeroesAndNumMeaningfulBits & 63) + 1
```
robskillington: Can we make these two numbers consts? 

i.e. 
```go
const (
  bits12To6Set = 4032 // 1111 1100 0000
  bits6To0Set = 63 // 0011 1111
)
```

src/dbnode/encoding/proto/common.go
```diff
@@ -141,11 +143,17 @@ var (
 	}
 )
 
+type marshaledField struct {
```
robskillington: nit: `marshaledField` should be `marshalledField` I believe

src/dbnode/encoding/proto/common.go
```diff
@@ -141,11 +143,17 @@ var (
...
+	fieldNum  int32
+	marshaled []byte
```
robskillington: nit: Should be `marshalled` I believe

src/dbnode/encoding/proto/custom_marshal.go
```diff
@@ -0,0 +1,155 @@
...
+	"math"
+
+	"github.com/golang/protobuf/proto"
+)
+
+type customFieldMarshaler interface {
+	encFloat64(tag int32, x float64)
+	encFloat32(tag int32, x float32)
+	encInt32(tag int32, x int32)
+	encSInt32(tag int32, x int32)
```
robskillington: I assume the S variants are `varint` or `short`?  Maybe use `encVarInt32` or `encShortInt32` instead?  

The single letter S is strange and not as clear the delineation while reading the API.

src/dbnode/encoding/proto/custom_marshal_test.go
```diff
@@ -0,0 +1,83 @@
...
+
+	marshaler := newCustomMarshaler()
+	for _, tc := range testCases {
+		marshaler.reset()
+		customMarshalVL(t, marshaler, tc.message)
+
+		unmarshalM := dynamic.NewMessage(testVLSchema)
+		require.NoError(t, unmarshalM.Unmarshal(marshaler.bytes()))
+
+		require.True(t, dynamic.Equal(tc.message, unmarshalM))
```
robskillington: So you can run individual tests, etc it's best practice to wrap this in a `t.Run(...)`:
```go
for i, tc := range testCases {
  t.Run(fmt.Sprintf("my_test_%d", i), func (t *testing.T) {
    // test case innards
  })
}
```

Also helps if you need defer, etc in the body.

src/dbnode/encoding/proto/custom_marshal.go
```diff
@@ -0,0 +1,155 @@
...
+
+package proto
+
+import (
+	"math"
+
+	"github.com/golang/protobuf/proto"
+)
+
+type customFieldMarshaler interface {
```
robskillington: `customFieldMarshaler` should be `customFieldMarshaller` I believe.

src/dbnode/encoding/proto/custom_marshal.go
```diff
@@ -0,0 +1,155 @@
...
+	encUInt64(tag int32, x uint64)
+	encBool(tag int32, x bool)
+	encBytes(tag int32, x []byte)
+	encPartialProto(tag int32, x []byte)
+	bytes() []byte
+	setBytes(b []byte)
+	reset()
+}
+
+type customMarshaler struct {
```
robskillington: `customMarshaller`

src/dbnode/encoding/proto/custom_marshal.go
```diff
@@ -0,0 +1,155 @@
...
+	buf *buffer
+}
+
+func newCustomMarshaler() customFieldMarshaler {
```
robskillington: `newCustomMarshaller`

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -68,18 +68,15 @@ func (u *customUnmarshaler) numNonCustomValues() int {
...
+func (u *customUnmarshaler) nonCustomFieldValues() []marshaledField {
 	return u.nonCustomValues
 }
 
 func (u *customUnmarshaler) unmarshal() error {
 	u.customValues = u.customValues[:0]
 
 	if u.nonCustomValues != nil {
-		u.nonCustomValues.Reset()
+		u.nonCustomValues = u.nonCustomValues[:0]
```
robskillington: Do you need to zero out each entry to avoid holding onto refs or does `marshalledField` here not contain references to longly lived pointer types?

robskillington: Hm I see a `[]byte` in that struct that could be longly held onto.

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -96,29 +93,40 @@ func (u *customUnmarshaler) unmarshal() error {
...
 
 			var (
-				startIdx       = tagAndWireTypeStartOffset
-				endIdx         = u.decodeBuf.index
-				marshaledField = u.decodeBuf.buf[startIdx:endIdx]
+				startIdx  = tagAndWireTypeStartOffset
+				endIdx    = u.decodeBuf.index
+				marshaled = u.decodeBuf.buf[startIdx:endIdx]
 			)
 			// A marshaled Protobuf message consists of a stream of <fieldNumber, wireType, value>
```
robskillington: nit: `A marshalled ...`

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -96,29 +93,40 @@ func (u *customUnmarshaler) unmarshal() error {
...
 				return err
 			}
 
 			var (
-				startIdx       = tagAndWireTypeStartOffset
-				endIdx         = u.decodeBuf.index
-				marshaledField = u.decodeBuf.buf[startIdx:endIdx]
+				startIdx  = tagAndWireTypeStartOffset
+				endIdx    = u.decodeBuf.index
+				marshaled = u.decodeBuf.buf[startIdx:endIdx]
```
robskillington: nit: `marshalled = ...`

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -96,29 +93,40 @@ func (u *customUnmarshaler) unmarshal() error {
...
-			// marshaled message and as a result we can build up the nonCustomValues *dynamic.Message
-			// one field at a time by calling UnmarshalMerge() on sub-slices that contain a complete
-			// tuple.
-			if err := u.nonCustomValues.UnmarshalMerge(marshaledField); err != nil {
-				return err
+			// marshaled message and as a result we can build up the []marshaledField one field at
+			// a time.
+			found := false
+			if fd.IsRepeated() {
+				// TODO: Leave comment
```
robskillington: Still need to add comment?

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -96,29 +93,40 @@ func (u *customUnmarshaler) unmarshal() error {
...
+				for i := range u.nonCustomValues {
+					val := u.nonCustomValues[i]
```
robskillington: nit: Perhaps `for i, val := range u.nonCustomValues` to remove this line?

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -332,14 +344,15 @@ func unmarshalSimpleField(fd *desc.FieldDescriptor, v uint64) (unmarshalValue, e
...
-	}
-
 	u.schema = schema
 	u.numNonCustom = 0
 	u.customValues = u.customValues[:0]
+
+	for i := range u.nonCustomValues {
+		u.nonCustomValues[i] = marshaledField{}
+	}
+	u.nonCustomValues = u.nonCustomValues[:0]
```
robskillington: You could make this an member method so in the other place you zero out you remember to zero the actual fields?  i.e. `u.resetNonCustomValues()`

src/dbnode/encoding/proto/custom_unmarshal.go
```diff
@@ -332,14 +344,15 @@ func unmarshalSimpleField(fd *desc.FieldDescriptor, v uint64) (unmarshalValue, e
...
 
 func (u *customUnmarshaler) resetAndUnmarshal(schema *desc.MessageDescriptor, buf []byte) error {
-	if schema != u.schema {
-		u.nonCustomValues = dynamic.NewMessage(schema)
-		u.nonCustomValues.Reset()
-	}
-
 	u.schema = schema
 	u.numNonCustom = 0
 	u.customValues = u.customValues[:0]
```
robskillington: Does `customValues` also need its entries reset? (does it contain any pointer types?)

src/dbnode/encoding/proto/custom_unmarshal_test.go
```diff
@@ -151,11 +151,11 @@ func TestCustomFieldUnmarshaler(t *testing.T) {
...
-			m := unmarshaler.nonCustomFieldValues()
-			assertAttributesEqual(
-				t,
-				tc.attributes,
-				m.GetFieldByName("attributes").(map[interface{}]interface{}))
+			// m := unmarshaler.nonCustomFieldValues()
+			// assertAttributesEqual(
+			// 	t,
+			// 	tc.attributes,
+			// 	m.GetFieldByName("attributes").(map[interface{}]interface{}))
```
robskillington: Should this still be commented out?

src/dbnode/encoding/proto/encoder.go
```diff
@@ -657,7 +652,8 @@ func (enc *Encoder) encodeBoolValue(i int, val bool) {
 	}
 }
 
-func (enc *Encoder) encodeProtoValues(m *dynamic.Message) error {
+// TODO: Rename encodeNonCustomValues
```
robskillington: Still relevant?

src/dbnode/encoding/proto/encoder.go
```diff
@@ -670,77 +666,71 @@ func (enc *Encoder) encodeProtoValues(m *dynamic.Message) error {
...
 		// Only want to skip encoding if nothing has changed AND we've already
 		// encoded the first message.
 		enc.stream.WriteBit(opCodeNoChange)
 		return nil
 	}
 
-	marshaled, err := m.MarshalAppend(enc.marshalBuf[:0])
-	if err != nil {
-		return fmt.Errorf("%s error trying to marshal protobuf: %v", encErrPrefix, err)
+	marshalBuf := enc.marshalBuf[:0]
```
robskillington: Hm, seems you reset the `enc.marshalBuf` back to the value of this local var.  Perhaps remove the `marshalBuf` var altogether and just use `enc.marshalBuf` below?

src/dbnode/encoding/proto/encoder_test.go
```diff
@@ -24,6 +24,7 @@ import (
 	"testing"
 	"time"
 
+	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
```
robskillington: Hm this import doesn't belong here yeah?

Linter is definitely getting way more flakey, noticed a lot of these.

src/dbnode/encoding/proto/iterator.go
```diff
@@ -26,6 +26,7 @@ import (
 	"io"
 	"math"
 
+	dpb "github.com/golang/protobuf/protoc-gen-go/descriptor"
```
robskillington: Same here.

src/dbnode/encoding/proto/iterator.go
```diff
@@ -51,18 +51,19 @@ var (
...
-	lastIterated           *dynamic.Message
-	lastIteratedProtoBytes []byte
-	byteFieldDictLRUSize   int
+	nsID                 ident.ID
+	opts                 encoding.Options
+	err                  error
+	schema               *desc.MessageDescriptor
+	schemaDesc           namespace.SchemaDescr
+	stream               encoding.IStream
+	marshaler            customFieldMarshaler
```
robskillington: nit: Should be `marshaller` I believe

src/dbnode/encoding/proto/iterator.go
```diff
@@ -418,17 +435,20 @@ func (it *iterator) readProtoValues() error {
...
 
+	// TODO: Reset
 	m := dynamic.NewMessage(it.schema)
 	err = m.Unmarshal(unmarshalBytes)
 	if err != nil {
 		return fmt.Errorf("error unmarshaling protobuf: %v", err)
 	}
 
-	for _, field := range m.GetKnownFields() {
+	for _, marshaledField := range it.protoFields {
```
robskillington: nit: Should be `marshalledField` I believe.  Maybe just call it `field`?

src/dbnode/encoding/m3tsz/float_encoder_iterator.go
richardartoul: sure, but replaced "set" with "mask"

src/dbnode/encoding/proto/custom_marshal.go
richardartoul: Yeah I can see how that would be confusing. It actually stands for "signed" and these names are basically verbatim what they're called in in the Protobuf specification.

I'll add a comment

src/dbnode/encoding/proto/iterator.go
```diff
@@ -185,6 +185,14 @@ func (it *iterator) Next() bool {
 				it.err = fmt.Errorf("%s error reading custom fields schema: %v", itErrPrefix, err)
 				return false
 			}
+
+			// When the encoder changes its schema it will reset all of its nonCustomFields state
+			// which means that the iterator needs to do the same to keep them synchronized at
```
Haijuncao: @richardartoul  you mentioned there is some bug around schema change, is this it? 
I saw you made some changes for encoder.SetSchema but later reverted it.

Which test caught the bug?

richardartoul: @Haijuncao `TestBufferSnapshot` caught it. I added a call to `SetSchema` in `RoundTripPropTest` to reproduce the bug and prevent a regression.

I reverted my change because I overcomplicated it last night. This fix is simpler than what I did last night

src/dbnode/encoding/proto/docs/encoding.md
```diff
@@ -95,7 +95,7 @@ In practice, a Protbuf message can have any number of different fields; for perf
 Instead, multiple "logical" streams are interleaved on a per-write basis within one physical stream.
 The remainder of this section outlines the binary format used to accomplish this interleaving.
 
-The binary format begins with a stream header and then the remainder of the stream is a sequence of tuples in the form: `<per-write header, compressed timestamp, compressed custom encoded fields, Protobuf marshaled fields>`
+The binary format begins with a stream header and then the remainder of the stream is a sequence of tuples in the form: `<per-write header, compressed timestamp, compressed custom encoded fields, Protobuf marshalled fields>`
```
mway: nit:

> [...] stream header, and the remainder [...]

src/dbnode/encoding/ostream.go
```diff
@@ -155,10 +155,17 @@ func (os *ostream) WriteByte(v byte) {
 
 // WriteBytes writes a byte slice.
 func (os *ostream) WriteBytes(bytes []byte) {
+	if !os.hasUnusedBits() {
+		// If the stream is aligned on a byte boundary then all of the WriteByte()
+		// function calls and bit-twiddling can be skipped in favor of a single
+		// memcopy.
```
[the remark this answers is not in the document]
richardartoul: You're right, what I'm doing would bypass the pooling. I'll change it to call ensureCapacityFor() first.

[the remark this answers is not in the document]
richardartoul: I'm gonna stick with append because its harder to mess it up and as long as I call `ensureCapacityFor()` first it will be equivalent to a copy anyways.

Also thanks for pointing this out because I was forcing allocations by doing an append when I could use `ensureCapacityFor` to get a new byte slice from the pool if availabble

src/dbnode/encoding/ostream_test.go
```diff
@@ -89,7 +89,7 @@ func testWriteBytes(t *testing.T, o OStream) {
 	b, _ := os.Rawbytes()
 	require.Equal(t, rawBytes, b)
 
-	require.Equal(t, 8, os.pos)
+	require.Equal(t, 0, os.pos)
```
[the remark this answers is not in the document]
richardartoul: its a quirk of the implementation of the `ostream` that both pos `0` and pos `8` mean that the last byte is completely full:

```golang
func (os *ostream) hasUnusedBits() bool {
	return os.pos > 0 && os.pos < 8
}
```

That said its better to not change stuff if there is no need and I was perhaps a bit hasty here by changing this test. I'll change my implementation to ensure that it ends up with `pos` set to 8 instead of 0

src/dbnode/encoding/proto/buffer_encode.go
```diff
@@ -0,0 +1,87 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// This file is mostly copy-pasta from: https://github.com/jhump/protoreflect/blob/master/dynamic/codec.go
```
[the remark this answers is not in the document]
richardartoul: nothing, just copied the methods I needed

src/dbnode/encoding/proto/common.go
```diff
@@ -141,11 +143,19 @@ var (
 	}
 )
 
+type marshalledField struct {
+	fieldNum   int32
+	marshalled []byte
+}
+
+type sortedMarshaledFields []marshalledField
```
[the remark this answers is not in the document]
richardartoul: since I already refactored the entire package from `marshaled` to `marshalled` I'm gonna stick with `marshalled`. Will fix this

src/dbnode/encoding/proto/common.go
```diff
@@ -167,57 +177,83 @@ type encoderBytesFieldDictState struct {
...
 	fields := schema.GetFields()
 	numCustomFields := numCustomFields(schema)
-	numProtoFields := len(fields) - numCustomFields
+	numNonCustomFields := len(fields) - numCustomFields
 
-	if cap(s) >= numCustomFields {
-		for i := range s {
-			s[i] = customFieldState{}
+	if cap(customFields) >= numCustomFields {
+		for i := range customFields {
```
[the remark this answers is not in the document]
richardartoul: I specifically tried to avoid pointers since I can have a lot of these in memory at once so I'd rather not have the G.C chasing them.

In term of allocating, the for loop here basically boils down to a memclear since all of the (non pointer objects) will exist as one contiguous slice in memory and its just rolling through all of them and zeroing them out. Will not call into any `malloc` or G.C code

src/dbnode/encoding/proto/custom_marshal.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func newCustomMarshaller() customFieldMarshaller {
+	return &customMarshaller{
+		buf: newCodedBuffer(nil),
+	}
+}
+
+func (m *customMarshaller) encFloat64(tag int32, x float64) {
+	if x == 0 {
```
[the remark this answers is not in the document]
richardartoul: took me a minute to decipher this haha. I assumed that this would actually force the compiler to make the 0 -> 0.0 but you maybe be right that this is a type assertion. Good catch

src/dbnode/encoding/proto/common.go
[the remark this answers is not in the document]
mway: afaict the GC is going to end up chasing most of these anyway because they're going to be promoted from stack to heap storage as the lifetime of non-garbage `customField` members outlives their creating scope by being stored in a heap-allocated container - so the question really becomes whether it's worth it to shave off the extra stack allocations (which are admittedly not very expensive, though it's more a question of frequency/volume here) for ostensibly no change in GC behavior.

am i missing something? do you think that the difference is negligible or, worse, actually detrimental?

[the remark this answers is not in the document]
richardartoul: I guess what I'm saying is:

```golang
[]*customFieldState
```

in memory will be represented as a contiguous region in memory where each slot is 8 bytes (enough for a 64 bit pointer to the actual customFieldState which can be anywhere else on the heap.

```golang
[]customFieldState
```

will be a contiguous region in memory where each slot is equal to `unsafe.Sizeof(customFieldState{}`

If we have a slice of size 20, in both cases the G.C is going to have to do some work because the `customFieldState` itself has pointers that need to be chase, however, the latter doesn't have any pointers for the `customFieldState` which means thats 20 less heap allocated objects that the G.C needs to trace.

Basically by making it a slice of non-pointers you're forcing them all to be next to each other in memory and their are no long-lived pointers to any individual customFieldState.

As a result, when you loop through the slice and do `nonCustomFields[i] = customFieldState{}` you're not allocating anything new. You're just going to the slot in the slice and setting the space thats already been aside to a zero instance of the struct but nothing new had to be allocated from the heap, the space was already there, you just zero'd out all the fields.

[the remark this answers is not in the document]
mway: ah, yep, you're right. i was glossing over the contiguous/non-contiguous member tradeoff. thanks for explaining!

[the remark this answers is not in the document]
richardartoul:
```golang
var dontOptimize []customFieldState

func BenchmarkResetCustomFieldState(b *testing.B) {
	dontOptimize = make([]customFieldState, 100)
	for i := 0; i < b.N; i++ {
		for i := range dontOptimize {
			dontOptimize[i] = customFieldState{}
		}
	}
	// Make sure it cant be optimized.
	fmt.Println(dontOptimize)
}
```

```
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3/src/dbnode/encoding/proto
Running tool: /usr/local/bin/go test -benchmem -run=^$ github.com/m3db/m3/src/dbnode/encoding/proto -bench ^(BenchmarkResetCustomFieldState)$

BenchmarkResetCustomFieldState-8   	 2000000	       621 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/m3db/m3/src/dbnode/encoding/proto	1.953s
```

PR #1643, opened 2019-05-17
[dbnode] add namespace metadata admin service (crud)

Haijuncao:
Create a namespace metadata admin service that performs CRUD operation to namespace metadata including schema upsert against kv.Store.

Coordinator (src/query) and m3admin (statsdex) should use it instead of each duplicating the same logic.


src/query/api/v1/handler/namespace/schema_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+const (
+	testSchemaJSON = `
+{
+		"name": "testNamespace",
+        "msgName": "mainpkg.TestMessage",
+        "protoName": "mainpkg/test.proto",
+        "protoMap":
+        {
+            "mainpkg/test.proto": "syntax = \"proto3\";package mainpkg;message TestMessage {double latitude = 1;double longitude = 2;int64 epoch = 3;bytes deliveryID = 4;map<string, string> attributes = 5;}"
```
Haijuncao: @richardartoul  with json, I'd have to put everything in one line.
my next diff change it to yaml, but better support for multi-line.
user (or mostly us in the beginning), we can just copy proto file content as is

src/query/api/v1/handler/namespace/schema_test.go
```diff
@@ -0,0 +1,233 @@
...
+	cmd := exec.Command("/bin/sh", "-c", "jq -n --arg file1 "+file1var+
+		" --arg file2 "+file2var+" -f "+jqfile)
+	t.Logf("cmd args are %v\n", cmd.Args)
+	out, err := cmd.CombinedOutput()
+	require.NoError(t, err)
+	t.Logf("generated json is \n%s\n", string(out))
+	return string(out)
+}
+
+func TestGenTestJson(t *testing.T) {
```
Haijuncao: `--- PASS: TestSchemaDeploy (0.02s)
    schema_test.go:102: cmd args are [/bin/sh -c jq -n --arg file1 "$(</var/folders/vg/khsllfnx1qg8q5jk6ggj774c0000gn/T/schema_deploy_test114605121/test.proto)" --arg file2 "$(</var/folders/vg/khsllfnx1qg8q5jk6ggj774c0000gn/T/schema_deploy_test114605121/imported.proto)" -f /var/folders/vg/khsllfnx1qg8q5jk6ggj774c0000gn/T/schema_deploy_test114605121/test.jq]
    schema_test.go:105: generated json is
        {
          "name": "testNamespace",
          "msgName": "mainpkg.TestMessage",
          "protoName": "mainpkg/test.proto",
          "protoMap": {
            "mainpkg/test.proto": "syntax = \"proto3\";\n\npackage mainpkg;\n\nimport \"mainpkg/imported.proto\";\n\nmessage TestMessage {\n  double latitude = 1;\n  double longitude = 2;\n  int64 epoch = 3;\n  bytes deliveryID = 4;\n  map<string, string> attributes = 5;\n  ImportedMessage an_imported_message = 6;\n}",
            "mainpkg/imported.proto": "\nsyntax = \"proto3\";\n\npackage mainpkg;\n\nmessage ImportedMessage {\n  double latitude = 1;\n  double longitude = 2;\n  int64 epoch = 3;\n  bytes deliveryID = 4;\n}"
          }
        }`

src/dbnode/namespace/kvadmin/ns_admin.go
```diff
@@ -0,0 +1,134 @@
...
+	return ErrNotImplemented
+}
+
+func (as *adminService) Delete(name string) error {
+	// TODO [haijun] move logic from src/query/api/v1/handler/namespace here
+	return ErrNotImplemented
+}
+
+func (as *adminService) DeploySchema(name string, protoFileName, msgName string, protos map[string]string) (string, error) {
+	currentRegistry, currentVersion, err := as.currentRegistry()
```
richardartoul: super nit: could `currentRegistry` return arguments in the same order that get passed to `checkAndSet`

src/dbnode/namespace/kvadmin/ns_admin.go
```diff
@@ -0,0 +1,134 @@
...
+	// TODO [haijun] move logic from src/query/api/v1/handler/namespace here
+	return ErrNotImplemented
+}
+
+func (as *adminService) Delete(name string) error {
+	// TODO [haijun] move logic from src/query/api/v1/handler/namespace here
+	return ErrNotImplemented
+}
+
+func (as *adminService) DeploySchema(name string, protoFileName, msgName string, protos map[string]string) (string, error) {
```
richardartoul: `name, protoFileName, msgName string`

src/dbnode/namespace/kvadmin/ns_admin_test.go
```diff
@@ -0,0 +1,109 @@
...
+
+	currentMeta, err := namespace.NewMetadata(ident.StringID("ns1"), namespace.NewOptions())
+	require.NoError(t, err)
+	currentMap, err := namespace.NewMap([]namespace.Metadata{currentMeta})
+	require.NoError(t, err)
+	currentReg := namespace.ToProto(currentMap)
+
+	protoFile := "mainpkg/test.proto"
+	protoMsg := "mainpkg.TestMessage"
+	protoMap := map[string]string{protoFile: mainProtoStr, "mainpkg/imported.proto": importedProtoStr}
```
richardartoul: Nice :D

src/dbnode/namespace/schema.go
```diff
@@ -251,8 +256,22 @@ func genDependencyDescriptors(infd *desc.FileDescriptor) []*desc.FileDescriptor
 	return depfds
 }
 
-func parseProto(protoFile string, importPaths ...string) ([]*desc.FileDescriptor, error) {
-	p := protoparse.Parser{ImportPaths: importPaths, IncludeSourceCodeInfo: true}
+// protoStringProvider provides proto contents from strings.
+func protoStringProvider(source map[string]string) protoparse.FileAccessor {
+	if len(source) == 0 {
```
richardartoul: should it just return an optional error? Cant really do anything with a nil FileAccessor

src/dbnode/namespace/schema.go
```diff
@@ -278,20 +297,43 @@ func marshalFileDescriptors(fdList []*desc.FileDescriptor) ([][]byte, error) {
 	return dlist, nil
 }
 
-func LoadSchemaRegistryFromFile(schemaReg SchemaRegistry, nsID ident.ID, protoFile string, msgName string, importPathPrefix ...string) error {
-	out, _ := parseProto(protoFile, importPathPrefix...)
+func AppendSchemaOptions(schemaOpt *nsproto.SchemaOptions, protoFile, msgName string, contents map[string]string, deployID string, importPath ...string) (*nsproto.SchemaOptions, error) {
+	sh, err := LoadSchemaHistory(schemaOpt)
```
richardartoul: My brain keeps completing this to `shell`, maybe `schemaHist`?

src/dbnode/namespace/schema.go
```diff
@@ -278,20 +297,43 @@ func marshalFileDescriptors(fdList []*desc.FileDescriptor) ([][]byte, error) {
...
+	var prevID string
+	if descr, ok := sh.GetLatest(); ok {
+		prevID = descr.DeployId()
+	}
 
-	dlist, _ := marshalFileDescriptors(out)
+	if err != nil {
```
richardartoul: shouldnt this error check happen right after its returned? If you got an error, you've already tried to use the schema history

src/dbnode/namespace/schema.go
```diff
@@ -278,20 +297,43 @@ func marshalFileDescriptors(fdList []*desc.FileDescriptor) ([][]byte, error) {
 	return dlist, nil
 }
 
-func LoadSchemaRegistryFromFile(schemaReg SchemaRegistry, nsID ident.ID, protoFile string, msgName string, importPathPrefix ...string) error {
-	out, _ := parseProto(protoFile, importPathPrefix...)
+func AppendSchemaOptions(schemaOpt *nsproto.SchemaOptions, protoFile, msgName string, contents map[string]string, deployID string, importPath ...string) (*nsproto.SchemaOptions, error) {
```
richardartoul: So just to clarify: The way this work is that you provide the `contents` which is a map[string]string of all of the filenames to their contents. `msgName` is the name of the message that you want to use as the schema and `protoFile` is the name of the file that contains `msgName`?

If so, could you add a comment to the top of this method explaining all that. I was able to work it out cause I have context here, but I think it would be really confusing to decipher on your own

richardartoul: Does the `importPath` argument matter at all? I looked at this: https://godoc.org/github.com/jhump/protoreflect/desc/protoparse#Parser

and it doesn't seem relevant to our use-case (and it doesn't seem like you're using it from the service layer?)

src/query/api/v1/handler/namespace/schema.go
```diff
@@ -0,0 +1,121 @@
...
+
+	md, rErr := h.parseRequest(r)
+	if rErr != nil {
+		logger.Error("unable to parse request", zap.Error(rErr))
+		xhttp.Error(w, rErr.Inner(), rErr.Code())
+		return
+	}
+
+	opts := handler.NewServiceOptions("kv", r.Header, nil)
+	resp, err := h.Add(md, opts)
```
richardartoul: Cant wait to rip that code out and put it in your service layer

src/query/api/v1/handler/namespace/schema_test.go
richardartoul: @Haijuncao How do you feel about base64 encoding the files? Then it can still be done from the command line if we want, but it also makes it was easier to manually construct a request if necessary too (`cat schema.proto | base64`) or they can do it in a web browser too etc.

Other than that I like this much more than the previous YAML approach :)

src/dbnode/namespace/schema.go
Haijuncao: I did use this when loading from file. but I can certainly separate them so that they dont' share the same code, it might be clearer.

src/dbnode/namespace/schema.go
Haijuncao: sure, originally, nil FileAccessor meant Parser will use os.Open to source proto from local file system. 
we can be more explicitly about this.

src/query/api/v1/handler/namespace/schema.go
Haijuncao: agree, so this handler just handle req/resp/err presentation.

src/query/api/v1/handler/namespace/schema_test.go
Haijuncao: I don't see why we want to base64 the file given the jq approach? I prefer jq approach because it is more human friendly.

richardartoul: I like the base64 approach because:

1. Its a common way to include files in JSON payloads
2. Its easier for someone to construct a payload manually (I.E if they don't have access / dont want to use `jq`) because its much easier to generate a base64 encoded string that you can use in your terminal then get all the quoting that `jq` is doing correct.


That said I don't have super strong feelings on it, as long as we have a well structured payload like what you have above I'm cool with it

src/dbnode/namespace/schema.go
Haijuncao: actually, I don't see anything wrong with return a nil FileAccessor, protoparse.Parser take a nil FileAccessor, if map is empty, it is the right response to return a nil FileAccessor.

src/dbnode/namespace/schema.go
Haijuncao: ok, it took a while for me realize what you mean.

src/dbnode/namespace/schema.go
Haijuncao: good catch

src/dbnode/namespace/kvadmin/ns_admin.go
Haijuncao: honestly, it feels wierd to return version before object. so I'd keep this order.

src/dbnode/namespace/kvadmin/ns_admin.go
Haijuncao: sure

src/dbnode/namespace/schema.go
richardartoul: Ok sounds good if thats what the Parser package expects. Maybe just a comment saying how that works

src/dbnode/namespace/schema.go
```diff
@@ -278,32 +297,80 @@ func marshalFileDescriptors(fdList []*desc.FileDescriptor) ([][]byte, error) {
...
+// protoFile: name of the top level proto file.
+// msgName: name of the top level proto message.
+// contents: map of name to proto strings.
+//          Except for the top level proto file, other imported proto files' key must be exactly the same
+//          as how they are imported in the import statement:
+//          E.g. if import.proto is imported as below
+//          import "mainpkg/imported.proto";
+//          Then the map key for improted.proto must be "mainpkg/imported.proto"
+//          See src/dbnode/namesapce/kvadmin test for example.
+func AppendSchemaOptions(schemaOpt *nsproto.SchemaOptions, protoFile, msgName string, contents map[string]string, deployID string) (*nsproto.SchemaOptions, error) {
```
richardartoul: Maybe a quick check that deployID is not "" and also that it doesn't exist as any of the previous deployIDs (to prevent conflicts)?

src/dbnode/server/server.go
```diff
@@ -555,12 +555,14 @@ func Run(runOpts RunOptions) {
...
 	schemaRegistry := namespace.NewSchemaRegistry(protoEnabled, logger)
-	// TODO [haijun] remove after PR to set schema in etcd is done (plan layed out in issue #1614).
-	// To unblock #1578, we will load user schema from db node configuration into schema registry
-	// at dbnode startup/initialization time, there will be no dynamic schema update.
+	// For application m3db client integration test convenience (where a local dbnode is started as a docker container),
+	// we allow loading user schema from db node configuration into schema registry
+	// at dbnode startup/initialization time.
 	if protoEnabled {
 		for nsID, protoConfig := range cfg.Proto.SchemaRegistry {
+			dummyDeployID := "first"
```
richardartoul: I think it might be a little nicer to call this "local_schema" or "local" or "config" or something

src/dbnode/server/server.go
```diff
@@ -555,12 +555,14 @@ func Run(runOpts RunOptions) {
...
 	}
 	schemaRegistry := namespace.NewSchemaRegistry(protoEnabled, logger)
-	// TODO [haijun] remove after PR to set schema in etcd is done (plan layed out in issue #1614).
-	// To unblock #1578, we will load user schema from db node configuration into schema registry
-	// at dbnode startup/initialization time, there will be no dynamic schema update.
+	// For application m3db client integration test convenience (where a local dbnode is started as a docker container),
+	// we allow loading user schema from db node configuration into schema registry
+	// at dbnode startup/initialization time.
 	if protoEnabled {
 		for nsID, protoConfig := range cfg.Proto.SchemaRegistry {
```
richardartoul: This logic means that if you have anything put under the `SchemaRegistry` field we're ignore the schemas in etcd, but if you don't put anything, we'll try and get a schema out of etcd. Is that correct?

What happens if we are `protoEnabled`, don't set the schema registry, but there is no schema in etcd? Will the dbnodes error out somehow? Just want to understand how all this config flows together cause it seems like `protoEnabled` might have two purposes now

src/query/api/v1/handler/namespace/schema_test.go
```diff
@@ -0,0 +1,246 @@
...
+  msgName: "mainpkg.TestMessage",
+  protoName: "mainpkg/test.proto",
+  protoMap: 
+    {
+      "mainpkg/test.proto": $file1,
+      "mainpkg/imported.proto": $file2
+    }
+}
+`
+	testSchemaJson = `
```
richardartoul: Is there any way to generate this programatically so its not such a pain to edit? Does json.Marshal() not handle escaping?

src/dbnode/server/server.go
Haijuncao: yes, if proto is enabled, but schema is not populated prior to db/namespace initialization, namespace won't be initialized. We log an error and move on with other namespaces. 

This part of logic is in the previous PR. I believe you had the same comment and I responded there already.

src/query/api/v1/handler/namespace/schema_test.go
Haijuncao: If you look at the test closely, you will find that I have code to generate it, genTestJson, this testSchemaJson is generated by genTestJson.

I used jq to generate it, but jq is not installed on buildkite, so even though the test succeeds locally, it always fail on buildkite. I documented it in the test.

src/dbnode/server/server.go
Haijuncao: ok

src/dbnode/namespace/schema.go
Haijuncao: will do

src/dbnode/server/server.go
richardartoul: Haha ok well at least I'm consistent :p Sorry hard to keep track of everything.

src/query/api/v1/handler/namespace/schema_test.go
richardartoul: oh I see. So if you want to update this variable you should just run that test? Can you add a comment above this line saying: "Run blah blah blah to regenerate this string"

PR #1644, opened 2019-05-17
[query] Add config for timeseries limit returned by single DB node

robskillington:

Adds a config and a sane default for setting a max timeseries limit returned by a single DB node to m3query.

```release-note
NONE
```

```documentation-note
NONE
```


src/cmd/services/m3query/config/config.go
```diff
@@ -217,15 +228,33 @@ type PerQueryLimitsConfiguration struct {
...
 }
 
+// AsFetchOptionsBuilderOptions converts this configuration to
+// handler.FetchOptionsBuilderOptions.
+func (l *PerQueryLimitsConfiguration) AsFetchOptionsBuilderOptions() handler.FetchOptionsBuilderOptions {
+	if l.MaxTimeseries <= 0 {
+		return handler.FetchOptionsBuilderOptions{
+			Limit: defaultStorageQueryLimit,
+		}
+	}
```
benraskin92: nit: newline

src/cmd/services/m3query/config/config.go
```diff
@@ -217,15 +228,33 @@ type PerQueryLimitsConfiguration struct {
 	// this field directly.
 	PrivateMaxComputedDatapoints int64 `yaml:"maxComputedDatapoints"`
 
-	// MaxFetchedDatapoints limits the number of datapoints actually used by a given query.
+	// MaxFetchedDatapoints limits the number of datapoints actually used by a
+	// given query.
 	MaxFetchedDatapoints int64 `yaml:"maxFetchedDatapoints"`
+
+	// MaxTimeseries limits the number of time series returned by a storage node.
+	MaxTimeseries int64 `yaml:"maxTimeseries"`
```
benraskin92: `maxFetchedTimeseries` instead to be consistent with `maxFetchedDatapoints`?

src/query/api/v1/handler/fetch_options.go
```diff
@@ -0,0 +1,67 @@
...
+) (*storage.FetchOptions, *xhttp.ParseError) {
+	fetchOpts := storage.NewFetchOptions()
+	fetchOpts.Limit = b.opts.Limit
+	if str := req.Header.Get(LimitMaxTimeseriesHeader); str != "" {
+		n, err := strconv.Atoi(str)
+		if err != nil {
+			return nil, xhttp.NewParseError(err, http.StatusBadRequest)
+		}
+		fetchOpts.Limit = n
+	}
```
benraskin92: nit: newline

src/query/api/v1/handler/fetch_options.go
```diff
@@ -0,0 +1,67 @@
...
+) FetchOptionsBuilder {
+	return fetchOptionsBuilder{opts: opts}
+}
+
+func (b fetchOptionsBuilder) NewFetchOptions(
+	req *http.Request,
+) (*storage.FetchOptions, *xhttp.ParseError) {
+	fetchOpts := storage.NewFetchOptions()
+	fetchOpts.Limit = b.opts.Limit
+	if str := req.Header.Get(LimitMaxTimeseriesHeader); str != "" {
```
benraskin92: Good call on making this part of the request. Although does it make more sense to have this be part of the url parameters? More of an educational question for me..

richardartoul: Do we have any documentation right now on setting query limits and such? If not would be nice to add a simple document that outlines the YAML configs you can set as well as this head.

src/query/api/v1/handler/headers.go
```diff
@@ -21,18 +21,22 @@
...
+	// ServedByHeader is the M3 query storage execution breakdown.
 	ServedByHeader = "M3-Storage-By"
 
-	// DeprecatedHeader is the M3 deprecated header
+	// DeprecatedHeader is the M3 deprecated header.
 	DeprecatedHeader = "M3-Deprecated"
 
+	// LimitMaxTimeseriesHeader is the M3 limit timeseries header that limits
+	// the number of time series returned by each storage node.
+	LimitMaxTimeseriesHeader = "M3-Limit-Max-Timeseries"
```
richardartoul: I think `M3-Limit-Max-Series` or just `M3-Max-Series` to be easier to remember

src/query/api/v1/handler/prometheus/native/list_tags.go
```diff
@@ -48,13 +48,15 @@ var (
 
 // ListTagsHandler represents a handler for list tags endpoint.
 type ListTagsHandler struct {
-	storage storage.Storage
-	nowFn   clock.NowFn
+	storage             storage.Storage
```
richardartoul: This new limit wont affect aggregate endpoints will it?

src/query/api/v1/handler/search.go
```diff
@@ -46,24 +46,32 @@ const (
...
 func (h *SearchHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	logger := logging.WithContext(r.Context())
 
-	query, rErr := h.parseBody(r)
-	if rErr != nil {
-		logger.Error("unable to parse request", zap.Error(rErr))
+	query, parseBodyErr := h.parseBody(r)
+	opts, parseURLParamsErr := h.parseURLParams(r)
+	// NB(r): Use a loop here to avoid two err handling code paths
+	for _, rErr := range []*xhttp.ParseError{parseBodyErr, parseURLParamsErr} {
```
richardartoul: Took me longer to digest this than two error handlings paths, and it possibly allocates? Would probably just do the normal thing here personally or at least do

```golang
for _, rErr := range [2]*xhttp.ParseError{parseBodyErr, parseURLParamsErr} {
...
}
```

to be sure it does not allocate

src/query/api/v1/handler/search.go
```diff
@@ -90,24 +98,19 @@ func (h *SearchHandler) parseBody(r *http.Request) (*storage.FetchQuery, *xhttp.
...
-		if err != nil {
-			limit = defaultLimit
+func (h *SearchHandler) parseURLParams(r *http.Request) (*storage.FetchOptions, *xhttp.ParseError) {
+	fetchOpts, err := h.fetchOptionsBuilder.NewFetchOptions(r)
+	if err != nil {
+		return nil, err
+	}
+
+	if str := r.URL.Query().Get("limit"); str != "" {
+		if limit, err := strconv.Atoi(str); err == nil {
```
richardartoul: honestly if they pass the correct query URL name but we can't parse it into a number, it should just return an error

src/query/models/query_context.go
```diff
@@ -35,24 +35,37 @@ type QueryContext struct {
 	Ctx      context.Context
 	Scope    tally.Scope
 	Enforcer cost.ChainedEnforcer
+	Options  QueryContextOptions
+}
+
+// QueryContextOptions is a set of optionally set options for the query
```
richardartoul: `QueryContextOptions contains optional configuration for the query context`

src/query/server/server.go
```diff
@@ -727,25 +736,40 @@ func remoteClient(
...
 ) (*grpc.Server, error) {
 	logger.Info("creating gRPC server")
-	server := tsdbRemote.CreateNewGrpcServer(storage, poolWrapper)
-	waitForStart := make(chan struct{})
-	var startErr error
+	server := tsdbRemote.CreateNewGrpcServer(storage,
+		queryContextOptions, poolWrapper)
+
+	var (
+		waitForStart = make(chan struct{}, 1)
```
richardartoul: Can you just make this a `chan error` and get rid of the startErr atomic stuff? Seems confusing for no reason.

```golang
// StartNewGrpcServer starts server on given address, then notifies channel
func StartNewGrpcServer(
	server *grpc.Server,
	address string,
	waitForStart chan<- error{},
) error {
	lis, err := net.Listen("tcp", address)
	if err != nil {
                 waitForStart <-err
	}

	waitForStart <- nil
	return server.Serve(lis)
}
```

src/query/server/server.go
```diff
@@ -727,25 +736,40 @@ func remoteClient(
...
+		startErr     atomic.Value
+		wg           sync.WaitGroup
+	)
+	startErr.Store(nil)
+	wg.Add(1)
```
richardartoul: I dont think you actually use this

src/query/api/v1/handler/search.go
robskillington: Sure, I might actually just do a helper function like I have in `xerrors.FirstError` (only reason I didn't do that is that it expects `func FirstError(err ...error) error` but these are typed errors....

src/query/server/server.go
robskillington: The whole thing is kinda messed up, I wanted to introduce least amount of changes possible.

I'll just break it up into two steps and make the grpc server take the listener.

src/query/api/v1/handler/search.go
robskillington: Yeah agreed, will change what this behavior used to do to match that.

src/query/server/server.go
robskillington: Ta, I'm going to rip all this stuff out thankfully - good catch though.

src/query/api/v1/handler/prometheus/native/list_tags.go
robskillington: It will too yeah.

robskillington: But will correctly respect the size of the aggregate values, not the amount of documents scanned.

src/query/api/v1/handler/fetch_options.go
robskillington: It does make more sense to be a URL parameter but I wanted this to be a part of any endpoint.  So I didn't want to necessarily have one endpoint using "limit" for something other than what I'm specifying then have them collide.

We can always go add a URL param later if we think we can avoid collisions, this was just a quick way to add it for now.

robskillington: Yeah, let me look around for it.  @benraskin92 do you know of any?

PR #1646, opened 2019-05-19
[m3ninx] Add support for document compression

prateek:
- Adds underpinnings to support document data compression in `m3ninx`. 
- Support for three codecs: `snappy`, `deflate` and `lz4`
- Minor refactoring to abstract away `docs.Reader` and `docs.Writer` type away from the `fst.Writer`, into the `encoding/docs` package. It allows the `fst.Writer` to defer knowledge of the compression internals based on the `CompressionOptions` knob passed down. 

NB: Putting this up despite not being ready to be used in production (as it is currently) to break up PRs.  The feature is disabled so landing won't affect any releases. More PRs to follow.


src/m3ninx/index/segment/fst/encoding/docs/bench_test.go
```diff
@@ -0,0 +1,253 @@
...
+goos: darwin
+goarch: amd64
+pkg: github.com/m3db/m3/src/m3ninx/index/segment/fst/encoding/docs
+BenchmarkAccessSingleCompressedStoredFieldsData/uncached_snappy-4         	  100000	     12433 ns/op	1317.78 MB/s	   16560 B/op	       4 allocs/op
+BenchmarkAccessSingleCompressedStoredFieldsData/uncached_deflate-4        	   10000	    161008 ns/op	 101.76 MB/s	   57904 B/op	      19 allocs/op
+BenchmarkAccessSingleCompressedStoredFieldsData/uncached_lz4-4            	   20000	     58620 ns/op	 279.49 MB/s	  188944 B/op	       9 allocs/op
+BenchmarkAccessSingleCompressedStoredFieldsData/cached_snappy-4           	  200000	      7031 ns/op	2330.03 MB/s	     128 B/op	       2 allocs/op
+BenchmarkAccessSingleCompressedStoredFieldsData/cached_deflate-4          	   10000	    137704 ns/op	 118.98 MB/s	     832 B/op	      11 allocs/op
+BenchmarkAccessSingleCompressedStoredFieldsData/cached_lz4-4              	  100000	     10149 ns/op	1614.26 MB/s	     129 B/op	       2 allocs/op
+PASS
```
robskillington: Looks like Snappy wins on speed and allocations with reads, especially for uncached?

src/m3ninx/index/segment/fst/encoding/docs/bench_test.go
```diff
@@ -0,0 +1,253 @@
...
+snappy   page-size:2048     compression-ratio:  2.354
+snappy   page-size:4096     compression-ratio:  2.523
+snappy   page-size:8192     compression-ratio:  2.632
+snappy   page-size:16384    compression-ratio:  2.724
+snappy   page-size:32768    compression-ratio:  2.747
+snappy   page-size:65536    compression-ratio:  2.760
+snappy   page-size:131072   compression-ratio:  2.761
+snappy   page-size:262144   compression-ratio:  2.762
+snappy   page-size:524288   compression-ratio:  2.762
+snappy   page-size:1048576  compression-ratio:  2.762
```
robskillington: May as well just go with 4096 to mimic page size then I suppose?  Doesn't get much better the higher it gets after that really.

robskillington:
Nice(!)

src/m3ninx/index/segment/fst/encoding/docs/bench_test.go
prateek: Yep pretty much. The plan is to compute these compression ratios on a real world data set and see if the page size v ratio growth looks the same as synthetic data, and then pick the smallest page size we can.

src/m3ninx/index/segment/fst/encoding/docs/bench_test.go
prateek: Yeah exactly. I'm inclined to use it for a v1, and subsequently spend time on optimizing the lz4 lib (it should be able to do a lot better than the current implementation) if we need better.

src/m3ninx/index/segment/fst/encoding/docs/compressed_index.go
```diff
@@ -0,0 +1,169 @@
...
+// non-contiguous.
+func (w *compressedIndexWriter) Write(id postings.ID, o offset) error {
+	if !w.ready {
+		w.writeMetadata(id)
+		w.ready = true
+	} else {
+		if id <= w.prev {
+			return fmt.Errorf("postings IDs must be monotonically increasing: received %v but previous ID was %v", id, w.prev)
+		}
+		for i := 0; i < int(id-w.prev)-1; i++ {
```
richardartoul: Maybe a quick comment about why the format of the file requires this (fixed width presumably?)

src/m3ninx/index/segment/fst/encoding/docs/compressed_index.go
```diff
@@ -0,0 +1,169 @@
...
+
+	return w.write()
+}
+
+func (w *compressedIndexWriter) writeMetadata(id postings.ID) {
+	w.enc.PutUint64(uint64(id))
+}
+
+func (w *compressedIndexWriter) write() error {
+	b := w.enc.Bytes()
```
richardartoul: Is there a way to avoid having to do this? Like have the encoder write directly into the writer (or a buffered writer)

src/m3ninx/index/segment/fst/encoding/docs/compressed_index.go
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// newCompressedIndexReader returns a new compressedIndexReader.
+func newCompressedIndexReader(data []byte) (indexReader, error) {
+	if len(data) == 0 {
+		return &compressedIndexReader{}, nil
+	}
+
+	if len(data) < indexMetadataSize {
+		return nil, io.ErrShortBuffer
```
richardartoul: Can you return your own error message? Would be much easier to debug. "index file metadata should be blah bytes but was blah bytes" or w/e

richardartoul (commented):
Got through some of it, will get at more of it soon

src/m3ninx/index/segment/fst/encoding/docs/bench_test.go
robskillington: Ah yeah, hm I would think with real world the ratio gets better just since we have a lot of repeated data.

src/m3ninx/index/segment/fst/encoding/docs/README.md
```diff
@@ -1,10 +1,157 @@
...
+│ ┌───────────────────────┐ │
+│ │        Page 1         │ │
+│ ├───────────────────────┤ │
+│ │          ...          │ │
+│ ├───────────────────────┤ │
+│ │        Page n         │ │
+│ └───────────────────────┘ │
+└───────────────────────────┘
+
+Where each `Page` is prefixed with it's length. Each page:
```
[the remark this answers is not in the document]
robskillington: That is a good question, any thoughts on how to account for documents bigger than pagesize?  We could simply just allow variable block size and use a default perhaps to account for these kind of scenarios?

src/m3ninx/index/segment/fst/encoding/docs/compressed_data.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+// compressedDataReader is a reader for the data file for documents.
+type compressedDataReader struct {
+	data  []byte
+	ropts ReaderOptions
+}
+
+// newCompressedDataReader returns a new compressedDataReader.
+func newCompressedDataReader(ropts ReaderOptions, data []byte) (dataReader, error) {
```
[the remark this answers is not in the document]
robskillington: Agreed but definitely will be hard to iterate the FST without using mmap, just because of how vellum works right now.  Guess we could totally refactor the read side of the library at some point..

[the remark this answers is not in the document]
robskillington: Actually, this is the documents data which we could definitely avoid using mmap for.  Is it worth thinking about removing mmap for the documents data as we introduce compression to just these files anyhow?

src/m3ninx/index/segment/fst/encoding/docs/reader.go
```diff
@@ -0,0 +1,98 @@
...
+	indexData []byte
+
+	data  dataReader
+	index indexReader
+}
+
+var _ Reader = &reader{}
+
+// NewReader returns a new reader.
+func NewReader(ro ReaderOptions, docData, docIndex []byte) (Reader, error) {
```
arnikola: nit: think we put Options as last arg usually

src/m3ninx/index/segment/fst/encoding/docs/reader.go
```diff
@@ -0,0 +1,98 @@
...
+	ro := ReaderOptions{
+		CompressionOptions: r.copts,
+	}
+	if dec, ok := r.data.Decompressor(); ok {
+		ro.Decompressor = dec
+	}
+	return ro
+}
+
+func (r *reader) Reset(ro ReaderOptions, docData, indexData []byte) error {
```
arnikola: nit: think we put Options as last arg usually

src/m3ninx/index/segment/fst/encoding/docs/types.go
```diff
@@ -0,0 +1,158 @@
...
+	CompressionOptions CompressionOptions
+	// Decompressor is an optional field to allow readers to avoid
+	// allocations in case it's already stored.
+	Decompressor Decompressor
+}
+
+// Reader is a document reader.
+type Reader interface {
+	// Reset resets the document reader with the provided data.
+	Reset(ro ReaderOptions, docsData, idxData []byte) error
```
arnikola: nit: think we put Options as last arg usually

src/m3ninx/index/segment/fst/encoding/docs/writer.go
```diff
@@ -0,0 +1,98 @@
...
+
+type writer struct {
+	offsets     []offset
+	dataWriter  dataWriter
+	indexWriter indexWriter
+}
+
+// NewWriter returns a new Writer.
+func NewWriter(co CompressionOptions) (Writer, error) {
+	if co.Type == NoneCompressionType {
```
arnikola: nit: init `offsets` with a non-0 size

src/m3ninx/index/segment/fst/encoding/encoding.go
```diff
@@ -82,6 +82,12 @@ func (e *Encoder) PutBytes(b []byte) int {
 	return n + len(b)
 }
 
+// PutRawBytes encodes a byte slice and returns the number of bytes written.
+func (e *Encoder) PutRawBytes(b []byte) int {
+	e.buf = append(e.buf, b...)
```
arnikola: How does this compare to doing an indexed `copy`?

src/m3ninx/index/segment/fst/segment.go
```diff
@@ -103,6 +103,50 @@ func (sd SegmentData) Validate() error {
 	return nil
 }
 
+// MetadataAndCompressionOptions returns the metadata and compresion options
```
arnikola: nit: `compression options`

src/m3ninx/index/segment/fst/encoding/docs/reader.go
```diff
@@ -0,0 +1,98 @@
...
+	var (
+		dataReader  dataReader
+		indexReader indexReader
+		err         error
+	)
+	if ro.CompressionOptions.Type == NoneCompressionType {
+		dataReader = newRawDataReader(docData)
```
richardartoul: Can we keep these struct local and reset them? I.E have both a raw and compressed reader and use the appropriate one as necessary

PR #1647, opened 2019-05-20
[query] Add support for unary expressions

### LINKED ISSUE #1619: m3query error
i meet some error:

2019-05-09T07:01:06.490Z	ERROR	unable to fetch data	{"rqID": "508a98a5-d849-4745-aef8-eb4f66246c99", "error": "promql.Walk: unhandled node type *promql.UnaryExpr, -sum(rate(container_network_transmit_bytes_total{cluster=\"SGT-eks-apse1-prod\",kubernetes_io_hostname=~\"^.*$\"}[1m]))"}

but , this sql can work well in prometheus.

benraskin92:

This PR fixes #1619. It introduces support for unary expressions -- specifically `+` and `-`. For example, this query is now supported: 

```
-sum(up)
```

```release-note
Yes, this PR adds support for Prometheus unary expressions (`+` and `-`).
```

```documentation-note
NONE
```

### DIFF
--- a/src/query/block/lazy.go
+++ b/src/query/block/lazy.go
@@ -83,12 +83,20 @@ func (it *lazyStepIter) Meta() Metadata {
 }
 
 func (it *lazyStepIter) Current() Step {
-	c := it.it.Current()
-	tt := it.opts.TimeTransform()
+	var (
+		c        = it.it.Current()
+		tt, vt   = it.opts.TimeTransform(), it.opts.ValueTransform()
+		stepVals = c.Values()
+	)
+
+	vals := make([]float64, len(stepVals))
+	for i, val := range stepVals {
+		vals[i] = vt(val)
+	}
 
 	return ColStep{
 		time:   tt(c.Time()),
-		values: c.Values(),
+		values: vals,
 	}
 }
 
@@ -115,7 +123,20 @@ func (it *lazySeriesIter) Err() error               { return it.it.Err() }
 func (it *lazySeriesIter) SeriesCount() int         { return it.it.SeriesCount() }
 func (it *lazySeriesIter) SeriesMeta() []SeriesMeta { return it.it.SeriesMeta() }
 func (it *lazySeriesIter) Next() bool               { return it.it.Next() }
-func (it *lazySeriesIter) Current() Series          { return it.it.Current() }
+func (it *lazySeriesIter) Current() Series {
+	var (
+		c    = it.it.Current()
+		vt   = it.opts.ValueTransform()
+		vals = make([]float64, len(c.values))
+	)
+
+	for i, val := range c.values {
+		vals[i] = vt(val)
+	}
+
+	c.values = vals
+	return c
+}
 func (it *lazySeriesIter) Meta() Metadata {
 	mt := it.opts.MetaTransform()
 	return mt(it.it.Meta())
@@ -212,19 +233,27 @@ func (s unconsolidatedStep) Values() []ts.Datapoints {
 }
 
 func (it *ucLazyStepIter) Current() UnconsolidatedStep {
-	c := it.it.Current()
-	for _, val := range c.Values() {
+	var (
+		c       = it.it.Current()
+		stepDPs = c.Values()
+		dpList  = make([]ts.Datapoints, len(stepDPs))
+		tt, vt  = it.opts.TimeTransform(), it.opts.ValueTransform()
+	)
+
+	for j, val := range stepDPs {
+		dps := make([]ts.Datapoint, len(val))
 		for i, dp := range val.Datapoints() {
-			tt, vt := it.opts.TimeTransform(), it.opts.ValueTransform()
-			val[i].Timestamp = tt(dp.Timestamp)
-			val[i].Value = vt(dp.Value)
+
+			dps[i].Timestamp = tt(dp.Timestamp)
+			dps[i].Value = vt(dp.Value)
 		}
+
+		dpList[j] = dps
 	}
 
-	tt := it.opts.TimeTransform()
 	return unconsolidatedStep{
 		time:   tt(c.Time()),
-		values: c.Values(),
+		values: dpList,
 	}
 }
 
@@ -251,15 +280,23 @@ func (it *ucLazySeriesIter) SeriesCount() int         { return it.it.SeriesCount
 func (it *ucLazySeriesIter) SeriesMeta() []SeriesMeta { return it.it.SeriesMeta() }
 func (it *ucLazySeriesIter) Next() bool               { return it.it.Next() }
 func (it *ucLazySeriesIter) Current() UnconsolidatedSeries {
-	c := it.it.Current()
-	for _, val := range c.datapoints {
-		for i, dp := range val.Datapoints() {
+	var (
+		c         = it.it.Current()
+		seriesDPs = c.datapoints
+		dpList    = make([]ts.Datapoints, len(seriesDPs))
+	)
+
+	for i, val := range seriesDPs {
+		dps := make([]ts.Datapoint, len(val))
+		for j, dp := range val.Datapoints() {
 			tt, vt := it.opts.TimeTransform(), it.opts.ValueTransform()
-			val[i].Timestamp = tt(dp.Timestamp)
-			val[i].Value = vt(dp.Value)
+			dps[j].Timestamp = tt(dp.Timestamp)
+			dps[j].Value = vt(dp.Value)
 		}
+		dpList[i] = dps
 	}
 
+	c.datapoints = dpList
 	return c
 }
 
--- a/src/query/block/lazy.go
+++ b/src/query/block/lazy.go
@@ -236,13 +236,13 @@ func (s unconsolidatedStep) Values() []ts.Datapoints {
 
 func (it *ucLazyStepIter) Current() UnconsolidatedStep {
 	var (
-		c       = it.it.Current()
-		stepDPs = c.Values()
-		dpList  = make([]ts.Datapoints, 0, len(stepDPs))
-		tt, vt  = it.opts.TimeTransform(), it.opts.ValueTransform()
+		c      = it.it.Current()
+		values = c.Values()
+		dpList = make([]ts.Datapoints, 0, len(values))
+		tt, vt = it.opts.TimeTransform(), it.opts.ValueTransform()
 	)
 
-	for _, val := range stepDPs {
+	for _, val := range values {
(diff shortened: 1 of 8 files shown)


src/query/block/lazy.go
```diff
@@ -83,12 +83,20 @@ func (it *lazyStepIter) Meta() Metadata {
...
 func (it *lazyStepIter) Current() Step {
-	c := it.it.Current()
-	tt := it.opts.TimeTransform()
+	var (
+		c        = it.it.Current()
+		tt, vt   = it.opts.TimeTransform(), it.opts.ValueTransform()
+		stepVals = c.Values()
+	)
+
+	vals := make([]float64, len(stepVals))
```
prateek: nit: could you use the `make([].., 0, ..)` formulation and append instead of directly accessing by idx.

src/query/block/lazy.go
```diff
@@ -115,7 +123,21 @@ func (it *lazySeriesIter) Err() error               { return it.it.Err() }
 func (it *lazySeriesIter) SeriesCount() int         { return it.it.SeriesCount() }
 func (it *lazySeriesIter) SeriesMeta() []SeriesMeta { return it.it.SeriesMeta() }
 func (it *lazySeriesIter) Next() bool               { return it.it.Next() }
-func (it *lazySeriesIter) Current() Series          { return it.it.Current() }
+
+func (it *lazySeriesIter) Current() Series {
+	var (
+		c    = it.it.Current()
+		vt   = it.opts.ValueTransform()
+		vals = make([]float64, len(c.values))
```
prateek: here too

src/query/block/lazy.go
```diff
@@ -212,19 +234,27 @@ func (s unconsolidatedStep) Values() []ts.Datapoints {
 }
 
 func (it *ucLazyStepIter) Current() UnconsolidatedStep {
-	c := it.it.Current()
-	for _, val := range c.Values() {
+	var (
+		c       = it.it.Current()
+		stepDPs = c.Values()
+		dpList  = make([]ts.Datapoints, len(stepDPs))
```
prateek: here and on line 245 too

src/query/block/lazy.go
benraskin92: Isn't it more efficient to allocate the length up front (instead of capacity) and index in? Or is this a better safe than sorry situation and avoid a potential panic?

prateek: can get both by doing the other version. 

i.e. when using the version with append (below), it's not expanding the slice unless the initial alloc wasn't large enough. so you still get the benefit of the up front alloc but also the safety of not panic'ing. 
```
slice := make([]type, 0, len)
for _, v := range otherThing {
  slice = append(slice, v)
}
```

benraskin92: Ah okay, yeah that makes sense.

src/query/parser/promql/types.go
```diff
@@ -293,6 +293,18 @@ func getBinaryOpType(opType promql.ItemType) string {
 	}
 }
 
+// GetUnaryOpType returns the M3 unary op type based on the Prom op type
```
arnikola: nit; full stops at end of comments

src/query/parser/promql/parse.go
```diff
@@ -266,6 +274,19 @@ func (p *parseState) walk(node pql.Node) error {
...
 
+	case *pql.UnaryExpr:
+		err := p.walk(n.Expr)
+		if err != nil {
+			return err
+		}
+
+		unaryType, ok := GetUnaryOpType(n.Op)
+		if !ok {
+			return fmt.Errorf("only + and - operators allowed for unary expressions, received: %s", n.Op.String())
```
arnikola: nit; instead, have GetUnaryOpType return a (unaryType, error) and just return this error there

src/query/parser/promql/types.go
```diff
@@ -293,6 +293,18 @@ func getBinaryOpType(opType promql.ItemType) string {
...
+func GetUnaryOpType(opType promql.ItemType) (string, bool) {
```
arnikola: nit; does this need to be exported?

src/query/block/lazy.go
```diff
@@ -212,19 +235,28 @@ func (s unconsolidatedStep) Values() []ts.Datapoints {
...
 func (it *ucLazyStepIter) Current() UnconsolidatedStep {
-	c := it.it.Current()
-	for _, val := range c.Values() {
-		for i, dp := range val.Datapoints() {
-			tt, vt := it.opts.TimeTransform(), it.opts.ValueTransform()
-			val[i].Timestamp = tt(dp.Timestamp)
-			val[i].Value = vt(dp.Value)
+	var (
+		c       = it.it.Current()
+		stepDPs = c.Values()
```
arnikola: nit; `values` reads a bit better

src/query/block/lazy.go
```diff
@@ -250,16 +282,28 @@ func (it *ucLazySeriesIter) Err() error               { return it.it.Err() }
...
 func (it *ucLazySeriesIter) Current() UnconsolidatedSeries {
-	c := it.it.Current()
-	for _, val := range c.datapoints {
-		for i, dp := range val.Datapoints() {
-			tt, vt := it.opts.TimeTransform(), it.opts.ValueTransform()
-			val[i].Timestamp = tt(dp.Timestamp)
-			val[i].Value = vt(dp.Value)
+	var (
+		c         = it.it.Current()
+		seriesDPs = c.datapoints
```
arnikola: nit; values reads better

src/query/block/lazy_test.go
```diff
@@ -418,3 +419,224 @@ func TestUnconsolidatedSeriesIter(t *testing.T) {
 
 	assert.Equal(t, expected, actual.Datapoints())
 }
+
+// negative value offset tests
+
```
arnikola: nit newline

src/query/parser/promql/parse.go
```diff
@@ -266,6 +274,19 @@ func (p *parseState) walk(node pql.Node) error {
...
+		}
+
+		return p.addLazyTransform(0, lazy.UnaryType, unaryType)
```
arnikola: This is a bit janky; I think it's better to split the functions up for now (i.e. noops if it's plus, and creates a value transform otherwise)

src/query/parser/promql/types.go
benraskin92: Nope, good call.

src/query/parser/promql/types.go
```diff
@@ -248,7 +248,7 @@ func NewFunctionExpr(
 		return p, true, err
 
 	default:
-		// TODO: handle other types
+		// TODO: handle other types.
```
arnikola: nit; can probably remove this, think we handle all types

src/query/parser/promql/parse_test.go
```diff
@@ -76,6 +77,39 @@ func TestInvalidOffset(t *testing.T) {
 	require.Error(t, err)
 }
 
+func TestUnary(t *testing.T) {
+	q := "-up"
```
arnikola: worth adding `+up` case? Can probably generalize this to be like

```golang
for _, q := range []string{"+up","-up"} {
....
}
```

src/query/block/lazy_test.go
```diff
@@ -418,3 +419,224 @@ func TestUnconsolidatedSeriesIter(t *testing.T) {
...
+	assert.Equal(t, expected, actual.Values())
+	assert.Equal(t, now, actual.Time())
+}
+
+func TestUnconsolidatedSeriesIterWithNegativeValueOffset(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	bb := NewMockBlock(ctrl)
+	defer ctrl.Finish()
+	offset := time.Duration(0)
```
arnikola: mind adding one with both transform and offset?

src/query/block/lazy_test.go
```diff
@@ -418,3 +419,224 @@ func TestUnconsolidatedSeriesIter(t *testing.T) {
...
+	assert.Equal(t, expectedVals, actual.Values())
+	assert.Equal(t, now, actual.Time())
+}
+
+func TestSeriesIterWithNegativeValueOffset(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	b := NewMockBlock(ctrl)
+	offset := time.Duration(0)
+	off := NewLazyBlock(b, testLazyOpts(offset, -1.0))
```
arnikola: nit: can do `testLazyOpts(0, -1.0)` and it'll typecast it properly

src/query/parser/promql/parse_test.go
benraskin92: `+` skips the unary op, so its just a fetch. Will add a test though.

src/query/block/lazy_test.go
benraskin92: yeah just figured I'd be more explicit with it being a duration.

src/query/block/lazy.go
```diff
@@ -83,12 +83,20 @@ func (it *lazyStepIter) Meta() Metadata {
...
 func (it *lazyStepIter) Current() Step {
-	c := it.it.Current()
-	tt := it.opts.TimeTransform()
+	var (
+		c        = it.it.Current()
+		tt, vt   = it.opts.TimeTransform(), it.opts.ValueTransform()
+		stepVals = c.Values()
+	)
+
+	vals := make([]float64, 0, len(stepVals))
```
prateek: longer term, instead of allocating for each `Current()` call, you can reuse a field in the struct. it'd look something like this:

```
type lazyStepIter struct {
  it   StepIter		
  opts LazyOptions
  values []float64
}

func (it *lazyStepIter) Current() Step {
var (
	c        = it.it.Current()
	tt, vt   = it.opts.TimeTransform(), it.opts.ValueTransform()
	stepVals = c.Values()
)
it.values = it.values[:0]
if cap(it.values) < len(stepVals) {
  it.values = make([]float64, 0, len(stepVals))
}
...
  return ColStep{ 
    ...
    values: it.values,
  }
}
```

note: this might cause issues if the callers of the this function are relying on the returned value from this function to be valid after multiple calls to Next()/Current(). Dunno how it's used so can't say for sure. Maybe file an issue to look into this

arnikola: If we do it this way, would make sense to move all this logic into the `Next` function so we don't have to recalculate :+1:

prateek: Yeah, Next() would definitely be the better place for it.

PR #1650, opened 2019-05-20
[integration] Add query fanout docker integration test

robskillington:
**What this PR does / why we need it**:

Fixes #

**Special notes for your reviewer**:

**Does this PR introduce a user-facing and/or backwards incompatible change?**:

```release-note

```

**Does this PR require updating code package or user-facing documentation?**:

```documentation-note

```

PR #1652, opened 2019-05-21
[integration] Add query fanout docker integration test

arnikola:
Squashed update to #1650 

Adds a containerBlock which can enclose multiple Blocks with the same bounds (used here for merging fanout fetches)

Cleaned up debugging, hopefully got it all

Todo: unit tests for new fanout logic and for containerBlock, parallelize fanout fetch


scripts/docker-integration-tests/query_fanout/prometheus-cluster-b.yml
```diff
@@ -0,0 +1,48 @@
+# my global config
+global:
+  external_labels:
```
robskillington: Same with this, can get rid of Prom config as not used with test any more

src/query/block/container.go
```diff
@@ -0,0 +1,642 @@
...
+
+	b.blocks = append(b.blocks, bl)
+	return nil
+}
+
+func (b *containerBlock) Close() error {
+	multiErr := xerrors.NewMultiError()
+	multiErr.Add(b.err)
+	for _, bl := range b.blocks {
+		multiErr.Add(bl.Close())
```
robskillington: This is because of our silly API, but this should be:
multiErr = multiErr.Add(bl.Close())

src/query/block/container.go
```diff
@@ -0,0 +1,642 @@
...
+		n := iter.Next()
+
+		if it.err = iter.Err(); it.err != nil {
+			return false
+		}
+
+		if i == 0 {
+			next = n
+		} else {
+			if next != n {
```
robskillington: Nit: combine to else if

src/query/block/container.go
```diff
@@ -0,0 +1,642 @@
...
+type ucContainerBlock struct {
+	err    error
+	blocks []UnconsolidatedBlock
+}
+
+func (b *ucContainerBlock) Close() error {
+	multiErr := xerrors.NewMultiError()
+	multiErr.Add(b.err)
+	for _, bl := range b.blocks {
+		multiErr.Add(bl.Close())
```
robskillington: Same here, multiErr = multiErr.add....

src/query/block/container.go
```diff
@@ -0,0 +1,642 @@
...
+		n := iter.Next()
+
+		if it.err = iter.Err(); it.err != nil {
+			return false
+		}
+
+		if i == 0 {
+			next = n
+		} else {
+			if next != n {
```
robskillington: Else if combine here too?

src/query/block/container.go
arnikola: Oops, fixed

src/query/storage/fanout/storage.go
```diff
@@ -81,18 +81,58 @@ func (s *fanoutStorage) FetchBlocks(
...
+			key := it.Meta().Bounds.String()
+			if foundBlock, found := blockResult[key]; found {
+				// this block exists. Check to see if it's already an appendable block.
+				if acc, ok := foundBlock.(block.AccumulatorBlock); ok {
+					// already an accumulator block, add current block.
+					if err := acc.AddBlock(bl); err != nil {
+						return block.Result{}, err
+					}
+				} else {
+					acc := block.NewContainerBlock(2)
```
robskillington: nit: Make this size a const at top of file, and/or a default.

arnikola: Refactored a bit

Issue #1519, opened 2019-03-30
Multi coordinator case improvements


arnikola:
API:
- [ ] Add `/labels` endpoint on both GET and POST to match [Prometheus function](https://prometheus.io/docs/prometheus/latest/querying/api/#getting-label-names) (#1565)

Remote discoverability:
- [ ] Add capability for remote coordinators working in different zones to discover each other to avoid current fiddly and flaky setup (possibly force a datacenter/zone tag to be added on each remote result) [ Needs to be scoped out and designed to get a good idea of how long this would take ]

Remote storage:
- [x] Update CompleteTag method on client to use new endpoint (#1656)
- [x] Add CompleteTag methods to server (#1656)
- [ ] Update query grpc client (includes refactoring load balancing) [ 1-2 day ]

General:
- [ ] Add genny map-backed consolidations (started in #1177 but will need to be updated) [ 1-2 day ]
- [x] Add offset function (#1561)

Fanout storage:
- [x] Fanout storage needs to consolidate incoming series, completed tags, and search results. 
NB: This has been partially done; blocks across multiple sources are now correctly put together, but there is no consolidation between remote and local series. This may not be necessary if a DC tag is enforced when running in remote mode. (#1652 )
- [x] Add DC tag for remote mode, or consolidate blocks

Performance:
- [ ] Update grpc version
- [ ] Improve round robin functionality for remote dc queries
- [ ] Allow multiple zones for remote dc queries (potentially force a zone tag to be added? Would be nice to do this in KV instead of in config so that we can keep this consistent)
- [ ] Pool grpc objects

PR #1656, opened 2019-05-21
[query] Adding complete tags to remote server with integration test

arnikola:

Fixes part of #1519 by adding remote CompleteTags endpoint with an integration test.

```release-note
Allows users to use remote coordinators to complete tags 
```

```documentation-note
NONE
```


scripts/docker-integration-tests/setup.sh
```diff
@@ -5,7 +5,7 @@ set -xe
 # expected to be run from root of repository
 cd $GOPATH/src/github.com/m3db/m3
 
-SERVICES=(m3dbnode m3coordinator m3aggregator)
+SERVICES=(  m3coordinator  )
```
robskillington: nit: Revert changes to this file?  We should probably put a way to override the services to build (an env var or something).

arnikola: Oops, thought I'd reverted these changes

PR #1658, opened 2019-05-22
[dbnode] add schema registry to client configuration

Haijuncao:
Allow client configuration to load schema from file.
The code is the same as in cmd/services/m3dbnode/config


richardartoul:
@Haijuncao it looks like the code in `cmd/services/m3dbnode/config` is also not used anywhere?

Haijuncao:
> @Haijuncao I dont understand how this code works, where is the schema registry being used?

for example, you can take a look at server/server.go (where dbnode initialization is done). 
for m3db client, the initialziation code seem to be in location store (m3db clients), we can move it out to infra/m3db-client later, but for now, I will need to load the schema registry from config (using this diff) and then pass schema registry to m3db client as one of the options.

PR #1662, opened 2019-05-23
Support repair config in server.go

richardartoul:

Reading the repair config from YAML was disabled because we weren't using this feature at all. As we begin to invest into M3DB's repair functionality we want to be able to test the existing code.

### DIFF
--- a/src/dbnode/server/server.go
+++ b/src/dbnode/server/server.go
@@ -612,7 +612,7 @@ func Run(runOpts RunOptions) {
 		repairOpts := opts.RepairOptions().
 			SetRepairInterval(cfg.Repair.Interval).
 			SetRepairTimeOffset(cfg.Repair.Offset).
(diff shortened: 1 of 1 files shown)


src/dbnode/server/server.go
```diff
@@ -608,9 +608,22 @@ func Run(runOpts RunOptions) {
 	kvWatchClientConsistencyLevels(envCfg.KVStore, logger,
 		clientAdminOpts, runtimeOptsMgr)
 
-	opts = opts.
-		// Feature currently not working.
-		SetRepairEnabled(false)
+	if cfg.Repair != nil {
```
prateek: nit: could you make this a little simpler:
```
opts := opts.SetRepairEnabled(false)
if cfg.Repair != nil {
 // what you have now
} // no else required
```

PR #1664, opened 2019-05-24
Use ReadAt() (pread syscall) instead of Read() in `seek.go`

richardartoul:
The original implementation of using Read() instead of mmap required duplicating the F.Ds for each seeker clone (so they could seek concurrently) as well as issuing calls to Seek() (which was an additional syscall).

The new implementation takes advantage of the ReadAt() method / pread syscall to avoid the Seek() syscall. This has the added benefit of making the F.Ds stateless so that they can be shared among concurrent seekers which will cut the number of F.Ds required by M3DB in half for existing workloads and allow users to increase the seek concurrency configuration with no impact on the number of F.Ds.

Benefits:
1. Two less syscalls per time series data lookup
2. Half the number of open F.Ds for existing workloads with default configuration
3. Can now increase seek concurrency to any number with no impact on the number of open / required F.Ds

There are no added tests because a lot of tests were already added in #1421 that verify the behavior of this portion of the code-base. In particular, `retriever_concurrent_test.go` contains a number of aggressive concurrency tests that verify the safety of sharing the F.Ds among multiple concurrent seekers.

Thanks to Jacek Masiulaniec for the tip.


src/dbnode/persist/fs/seek.go
```diff
@@ -542,6 +526,7 @@ type ReusableSeekerResources struct {
 	xmsgpackDecoder   *xmsgpack.Decoder
 	fileDecoderStream *bufio.Reader
 	byteDecoderStream xmsgpack.ByteDecoderStream
+	offsetFileReader  *offsetFileReader
```
prateek: two questions:
(1 Does this need to be in this struct? allocating the 16 byte struct per read request isn't going to cost us anything (esp compared to the syscalls you've gotten rid of).
(2) What's left to make a regular Seeker concurrency safe? Would be nice to get rid of the `Concurrent...Seeker` if we could

src/dbnode/persist/fs/seek.go
```diff
@@ -610,3 +596,32 @@ func (s *simpleBytesPool) Put(b []byte) {
...
+	return &offsetFileReader{}
+}
+
+func (p *offsetFileReader) Read(b []byte) (n int, err error) {
+	n, err = p.fd.ReadAt(b, p.offset)
+	p.offset += int64(n)
+	return n, err
+}
+
+func (p *offsetFileReader) reset(fd *os.File) {
```
prateek: nit: could you simplify these two functions into. 

```
func (p *...) Reset(fd *os.File, offset int64) {...}
```

src/dbnode/persist/fs/seek.go
```diff
@@ -610,3 +596,32 @@ func (s *simpleBytesPool) Put(b []byte) {
...
 	s.pool = append(s.pool, b[:])
 }
+
+// offsetFileReader implements io.Reader() and allows an *os.File to be wrapped
+// such that any calls to Read() are issued at the provided offset. This is used
+// to issue reads to specific portions of the index and data files without having
+// to first call Seek(). This reduces the number of syscalls that need to be made
+// and also allows the fds to be shared among concurrent goroutines since the
+// internal F.D offset managed by the kernel is not being used.
+type offsetFileReader struct {
```
prateek: nit: mind adding the var decl 

```
var _ io.Reader = &offsetFileReader{}
```

src/dbnode/persist/fs/seek.go
```diff
@@ -380,16 +374,9 @@ func (s *seeker) SeekIndexEntry(
...
-	}
-	if seekedOffset != offset {
-		return IndexEntry{}, instrument.InvariantErrorf(
-			"tried to seek to offset: %d, but seeked to: %d", seekedOffset, offset)
-	}
-
-	resources.fileDecoderStream.Reset(s.indexFd)
+	resources.offsetFileReader.reset(s.indexFd)
+	resources.offsetFileReader.setOffset(offset)
+	resources.fileDecoderStream.Reset(resources.offsetFileReader)
```
prateek: tbh here's where I mean it'd be easier to avoid reusing the offsetFileReader and just alloc a new one. don't think it's going to cause any performance issues either way.

src/dbnode/persist/fs/seek.go
richardartoul: https://github.com/m3db/m3/issues/1668

richardartoul: re: putting it in the reusable seeker resources. Yeah it doesn't cost much to allocate but we already have this pattern setup so its pretty easy to hook into it. Would rather avoid the alloc if we can.

src/dbnode/persist/fs/seek.go
robskillington: There are users I see doing tens of thousands of reads per second, might be worth continuing to keep per operation allocation low/zero where possible.

PR #1666, opened 2019-05-24
[dbode] Call syscall.Setrlimit to set num files open hard limit with setcap for DB docker image

robskillington:

This raises the rlimit to whatever nr_open is set to and also sets correct permissions for the docker image to set rlimit.

```release-note
NONE
```

```documentation-note
NONE
```


src/dbnode/server/limits.go
```diff
@@ -78,3 +83,75 @@ func validateProcessLimits() error {
 
 	return multiErr.FinalError()
 }
+
+func raiseRlimitToNROpen() error {
```
robskillington: Thought: this maybe should be in a `_linux.go` file.

src/dbnode/server/limits.go
```diff
@@ -78,3 +83,75 @@ func validateProcessLimits() error {
...
+	cmd := exec.Command("sysctl", "-a")
```
schallert: Should just be able to check `/proc/sys`: https://www.kernel.org/doc/Documentation/sysctl/fs.txt

In kube:
```
/ # hostname
m3db-cluster-rep0-0
/ # cat /proc/sys/fs/nr_open
3000000
```

robskillington: Ah cool, yup I'll change it to this instead.

robskillington:
I actually didn't need IPC_LOCK thankfully.

schallert:
Looks good so far, just doing some sanity checking with this and the operator PR before stamping

docker/m3dbnode/Dockerfile
```diff
@@ -26,5 +26,9 @@ COPY --from=builder /go/src/github.com/m3db/m3/bin/m3dbnode /bin/
 COPY --from=builder /go/src/github.com/m3db/m3/src/dbnode/config/m3dbnode-local-etcd.yml /etc/m3dbnode/m3dbnode.yml
 COPY --from=builder /go/src/github.com/m3db/m3/scripts/m3dbnode_bootstrapped.sh /bin/
 
+# Use setcap and run as specific user
+RUN apk add libcap && \
```
richardartoul: Had not seen this before. Did some reading. Kind of bizarre that the capabilities get set on the file/binary level

richardartoul: What does the `+ep` do? Can you just add a comment to this line generally also

src/x/os/limits.go
```diff
@@ -27,3 +27,12 @@ type ProcessLimits struct {
 	VMMaxMapCount int64  // corresponds to /proc/sys/vm/max_map_count
 	VMSwappiness  int64  // corresponds to /proc/sys/vm/swappiness
 }
+
+// RaiseProcessNoFileToNROpenResult captures the result of trying to
+// raise the process num files open limit to the nr_open system value.
+type RaiseProcessNoFileToNROpenResult struct {
+	RaiseRequired   bool
```
richardartoul: Maybe `RaisePerformed`

src/x/os/limits.go
```diff
@@ -27,3 +27,12 @@ type ProcessLimits struct {
...
+	NROpenValue     uint64
+	NoFileMaxValue  uint64
+	NoFileCurrValue uint64
```
richardartoul: Maybe a comment here saying that this will be the curr value *before* the raise was performed (if it was performed)

src/x/os/limits_other.go
```diff
@@ -42,3 +45,9 @@ func CanGetProcessLimits() (bool, string) {
 func GetProcessLimits() (ProcessLimits, error) {
 	return ProcessLimits{}, errUnableToDetermineProcessLimits
 }
+
+// RaiseProcessNoFileToNROpen attempts to raise the process num files
+// open limit to the nr_open system value.
+func RaiseProcessNoFileToNROpen() (RaiseProcessNoFileToNROpenResult, error) {
```
richardartoul: If we're gonna start doing this it might be nice to have a generic `AttemptToSetProcessLimits` which calls this function. That way we don't have to change `server.go` each time and we have an established pattern

docker/m3dbnode/Dockerfile
robskillington: Not opposed to adding a comment, the +e is "effective" and +p is for "permitted".

src/x/os/limits.go
robskillington: Sure thing.

src/x/os/limits.go
robskillington: That's not true though, it will be the curr value (after the raise) or if no raise then it will be the curr value (unadjusted).

src/x/os/limits_other.go
robskillington: This is that kind of method however, it is meant to be abstract and returns enough information (without taking a logger itself).  That way callers can either log it as a warning or a hard error, depending on their situation.

robskillington:
Sure thing, I'll make it opt in and used only by the docker image.

PR #1670, opened 2019-05-28
Haijun/schema validation

Haijuncao:
Add basic schema validation to schema deploy endpoint. 
Basic validation just make sure schema is well formed (proto files can be parsed, message can be found, deploy ID is well formed), it does not check for schema backward compatibility.


src/dbnode/namespace/kvadmin/ns_admin.go
```diff
@@ -82,6 +82,36 @@ func (as *adminService) Delete(name string) error {
...
+			targetMeta = nsOpts
+			break
+		}
+	}
+	if targetMeta == nil {
+		return ErrNamespaceNotFound
+	}
+
+	// Clear schema options in place.
+	targetMeta.SchemaOptions = nil
```
richardartoul: So this will be a backward incompatible change which means if you do it to a cluster that is running it could cause really unpredictable behavior...I wonder if its best that if you really mess up the schema you just delete the namespace and move on? I guess if its just for internal usage / testing its fine but we should never tell anyone to use this

src/dbnode/namespace/kvadmin/types.go
```diff
@@ -45,4 +45,7 @@ type NamespaceMetadataAdminService interface {
 	// Application developer is to include the deployID in their m3db client configuration
 	// when they upgrade their application to use the new schema version.
 	DeploySchema(name, protoFileName, msgName string, protos map[string]string) (string, error)
+
+	// ResetSchema reset schema for the specified namespace.
```
richardartoul: Can you add in the comment here that people should not use this unless they really know what they're doing

src/dbnode/namespace/kvadmin/ns_admin.go
Haijuncao: Behavior is well defined:
Reset will not impact running dbnode, in memory state won't be updated if new schema from etcd does not extend from current one. (e.g. setting schema to nil).

Still, agree, this is only for administrator to fix schema if there is operator error. Added a big warning message and require a force header to be set to proceed.

src/dbnode/namespace/kvadmin/types.go
Haijuncao: Add it in error message, this operation requires force.

PR #1672, opened 2019-05-29
[docker] don't set m3dbnode binary caps in build

### LINKED ISSUE #1671: Setcap in docker build causes un-runnable container with docker defaults
In #1666 we added support for M3DB raising its own process limits. As part of this, we `setcap`'d the `m3dbnode` binary inside of the container. Doing so at build-time causes the image to fail to start unless the user specifically adds the capability:

v0.9.4 vanilla `docker run`:
```
$ docker run --rm quay.io/m3db/m3dbnode:v0.9.4
2019/05/29 01:29:07 Go Runtime version: go1.10.8
```

v0.9.5 vanilla `docker run`:
```
$ docker run --rm quay.io/m3db/m3dbnode:v0.9.5
standard_init_linux.go:207: exec user process caused "operation not permitted"
```

v0.9.5 `docker run` allowing caps:
```
$ docker run --cap-add SYS_RESOURCE --rm quay.io/m3db/m3dbnode:v0.9.5
2019/05/29 01:31:06 Go Runtime version: go1.10.8
```

Making our docker image crash on startup unless the user debugs the capability seems pretty aggressive. I believe we should either `setcap` at runtime or see if we even need to `setcap` the binary at all and can just do things at process runtime (I'm looking into this).

Regardless, I think we should revert the setcap change (at least within the build, we can leave the part where the dbnode tries to raise its own limit) until we test this out further.

schallert:

Fixes #1671

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+# 0.9.6
+
+## Bug Fixes
+
+- **M3DB**: Revert Docker setcap change to allow image to run with no special flags (#1672)
+
 # 0.9.5
 
 ## Performance
--- a/docker/m3dbnode/Dockerfile
+++ b/docker/m3dbnode/Dockerfile
@@ -26,8 +26,5 @@ COPY --from=builder /go/src/github.com/m3db/m3/bin/m3dbnode /bin/
 COPY --from=builder /go/src/github.com/m3db/m3/src/dbnode/config/m3dbnode-local-etcd.yml /etc/m3dbnode/m3dbnode.yml
 COPY --from=builder /go/src/github.com/m3db/m3/scripts/m3dbnode_bootstrapped.sh /bin/
 
-# Set the environment variable to raise the limit on startup
-ENV PROCESS_LIMITS_RAISE true
-
 ENTRYPOINT [ "/bin/m3dbnode" ]
 CMD [ "-f", "/etc/m3dbnode/m3dbnode.yml" ]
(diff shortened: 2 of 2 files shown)


schallert:
@richardartoul my understanding is that Linux won't even let you exec a file if it has capabilities on it that you don't have (not sure which set this considers, "bounding" or another), way before you even get to making the syscall.

For example, removing Docker from the equation, since my current bash shell has `cap_sys_resource` in its bounding set I can exec m3db:
```
# check my current caps
$ /sbin/capsh --print
Current: =
Bounding set =cap_chown,cap_dac_override,cap_dac_read_search,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_linux_immutable,cap_net_bind_service,cap_net_broadcast,cap_net_admin,cap_net_raw,cap_ipc_lock,cap_ipc_owner,cap_sys_module,cap_sys_rawio,cap_sys_chroot,cap_sys_ptrace,cap_sys_pacct,cap_sys_admin,cap_sys_boot,cap_sys_nice,cap_sys_resource,cap_sys_time,cap_sys_tty_config,cap_mknod,cap_lease,cap_audit_write,cap_audit_control,cap_setfcap,cap_mac_override,cap_mac_admin,cap_syslog,cap_wake_alarm,cap_block_suspend,cap_audit_read

# exec m3db
$ ./m3dbnode 2>&1 | head -n1
2019/05/29 04:11:50 Go Runtime version: go1.10.8
```

However if I use `capsh` to drop that capability, I'll get the same error (I have to be root to drop a cap):
```
# capsh --drop=cap_sys_resource --user=mschalle -- -c '/home/mschalle/m3dbnode'
```

According to Docker's docs, `SYS_RESOURCE` isn't included by default and must be explicitly added: https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities

PR #1673, opened 2019-05-29
Revert "[ci] build + release m3db with go1.12 (#1505)"

schallert:
We need to cut a patch release within the 0.9.x series to fix a Docker
issue. I'd rather release a patch release with the fix rather than have
folks have to pull in a new runtime to get the docker fix.

This reverts commit c753f134bac682f5ac1be3e43f762132028ed15f.

Fixes #

```release-note
NONE
```

```documentation-note
NONE
```


schallert:
> LGTM, want to make the revert revert P.R as well and I'll stamp that so you're unblocked?

Yeah will do

PR #1676, opened 2019-05-29
[query] Fix bug with parsing Graphite find query

arnikola:

This fixes an issue with parsing graphite find queries. Previously, `foo.*` would get split up into `foo`, and `*`, which would go through a graphite globber and become `[^\.]*`, which would then not get correctly identified as a `NewFieldQuery` in the storage/index (this would convert regex values looking for `.*` only to a `NewFieldQuery`.

This PR adds `MatchField` and `MatchNotField` match types, and explicitly makes use of them in graphite queries, avoiding relying on implicit conversion in the storage layer.

```release-note
NONE
```

```documentation-note
NONE
```


robskillington:
Ah nice, I like the approach of using explicit match types too.  Good stuff!

src/query/graphite/storage/converter_test.go
```diff
@@ -45,6 +45,20 @@ func TestConvertMetricPartToMatcher(t *testing.T) {
 	}
 }
 
+func TestConverWildcardToMatcher(t *testing.T) {
+	metric := "*"
+	for i := 0; i < 100; i++ {
```
benraskin92: This is kind of a weird test -- why not just run this once?

src/query/models/matcher.go
```diff
@@ -38,6 +38,10 @@ func (m MatchType) String() string {
 		return "=~"
 	case MatchNotRegexp:
 		return "!~"
+	case MatchField:
+		return "-"
```
benraskin92: For my understanding, how did you come up with these mappings?

src/query/storage/index.go
```diff
@@ -192,6 +192,17 @@ func matcherToQuery(matcher models.Matcher) (idx.Query, error) {
 		}
 		return query, nil
 
+	case models.MatchNotField:
+		negate = true
+		fallthrough
```
benraskin92: Nice - good use of fallthrough.

src/query/graphite/storage/converter_test.go
arnikola: Just to ensure that the metric name is also converted to expected (i.e. `__g100__`) as well

src/query/models/matcher.go
arnikola: Made 'em up :p

The only thing this affects is the `String()` method on MatchTypes

src/query/storage/index.go
arnikola: I learn from the best ;)

PR #1677, opened 2019-05-30
Handle host not available scenario during peer bootstrap more gracefully

### LINKED ISSUE #1667: Replacing a dead node can be very slow
## Problem
Replacing a dead node can be very slow (order of magnitude slower than doing the equivalent replace if the node is not dead).

The root cause of this issue is that the logic for streaming all of the metadata executes a metadata fetch in parallel for all peers for a given namespace/shard/block combination.

The existing logic waits for all requests to complete before deciding if consistency has been reached and continuing. If consistency is not reached, the loop will retry forever assuming no non-retryable errors are encountered. Within each parallel request, a retrier is used with the following configuration:

2s backoff
4 maxRetries
2x backoffFactor

This means that in the case where a node is hard down each parallel request for that host will take 30s, blocking progression even though the two other requests may have succeeded and achieved consistency.

```golang
# streamBlocksMetadataFromPeers in session.go
for idx, peer := range peers.peers {
		idx := idx
		peer := peer

		wg.Add(1)
		go func() {
			defer func() {
				// Success or error counts towards a response
				atomic.AddInt32(&responded, 1)

				// Decrement pending
				progress.metadataFetches.Update(float64(atomic.AddInt64(&pending, -1)))

				// Mark done
				wg.Done()
			}()

			var (
				firstAttempt = true
				// NB(r): currPageToken keeps the position into the pagination of the
				// metadata from this peer, it begins as nil but if an error is
				// returned it will likely not be nil, this lets us restart fetching
				// if we need to (if consistency has not been achieved yet) without
				// losing place in the pagination.
				currPageToken pageToken
			)
			condition := func() bool {
				if firstAttempt {
					// Always attempt at least once
					firstAttempt = false
					return true
				}
				currLevel := level.value()
				majority := int(majority)
				enqueued := int(enqueued)
				success := int(atomic.LoadInt32(&success))

				doRetry := !topology.ReadConsistencyAchieved(currLevel, majority, enqueued, success) &&
					errs.getAbortError() == nil
				if doRetry {
					// Track that we are reattempting the fetch metadata
					// pagination from a peer
					progress.metadataPeerRetry.Inc(1)
				}
				return doRetry
			}
			for condition() {
				var err error
				currPageToken, err = s.streamBlocksMetadataFromPeer(namespace, shardID,
					peer, start, end, currPageToken, metadataCh, resultOpts, progress)

				// Set error or success if err is nil
				errs.setError(idx, err)

				// Check exit criteria
				if err != nil && xerrors.IsNonRetryableError(err) {
					errs.setAbortError(err)
					return // Cannot recover from this error, so we break from the loop
				}
				if err == nil {
					atomic.AddInt32(&success, 1)
					return
				}
			}
		}()
	}

	wg.Wait()
```

## Proposed Solution

Return non-retryable error in the fetchFn given to the streamMetadataRetrier, to early terminate it. Further, we'll need to return a non non-retryable error back to the caller (streamBlocksMetadataFromPeers) to ensure it doesn't abort and then do something (like sleep) to prevent the loop in streamBlocksMetadataFromPeers from spinning too aggressively when the short-circuiting kicks in.

richardartoul:
Fixes #1667

### DIFF
--- a/src/dbnode/client/session.go
+++ b/src/dbnode/client/session.go
@@ -72,7 +72,7 @@ const (
 	blockMetadataChBufSize               = 4096
 	shardResultCapacity                  = 4096
 	hostNotAvailableMinSleepInterval     = 1 * time.Millisecond
-	hostNotAvailableMaxSleepInterval     = 10 * time.Millisecond
+	hostNotAvailableMaxSleepInterval     = 100 * time.Millisecond
 )
 
 type resultTypeEnum string
--- a/src/dbnode/client/session.go
+++ b/src/dbnode/client/session.go
@@ -2322,27 +2322,20 @@ func (s *session) streamBlocksMetadataFromPeer(
 
 	fetchFn := func() error {
 		borrowErr := peer.BorrowConnection(checkedAttemptFn)
-		// Don't retry if the host is not available to prevent exponential
-		// backoff from causing this to take a long time when subsequent
-		// requests will fail.
-		if isHostNotAvailableError(borrowErr) {
-			borrowErr = xerrors.NewNonRetryableError(borrowErr)
-		}
 		return xerrors.FirstError(borrowErr, attemptErr)
 	}
 
 	for moreResults {
 		if err := s.streamBlocksRetrier.Attempt(fetchFn); err != nil {
-			// Check if the error was a hostNotAvailableError wrapped in a
-			// nonRetryableError. If so, return the hostNotAvailableError
-			// instead of the non-retryable error since the error is technically
-			// retryable, it was only wrapped to prevent the exponential backoff
-			// in the actual retrier.
-			if xerrors.IsNonRetryableError(err) {
-				inner := xerrors.GetInnerNonRetryableError(err)
-				if isHostNotAvailableError(inner) {
-					err = inner
-				}
+			// Check if the error was a hostNotAvailableError. If so, return the
+			// raw (retryable) hostNotAvailableError since the error is technically
+			// retryable but was wrapped to prevent the exponential backoff in the
+			// actual retrier.
+			if isHostNotAvailableError(err) {
+				// hostNotAvailable is non-retryable by default, but in this case we want
+				// to return a retryable error so that the caller can continue retrying
+				// indefinitely until consistency is reached.
+				err = xerrors.GetInnerNonRetryableError(err)
 			}
 			return startPageToken, err
 		}
@@ -2853,11 +2846,6 @@ func (s *session) streamBlocksBatchFromPeer(
 			result, attemptErr = client.FetchBlocksRaw(tctx, req)
 		})
 		err := xerrors.FirstError(borrowErr, attemptErr)
-		// Do not retry if cannot borrow the connection or
-		// if the connection pool has no connections
-		if isHostNotAvailableError(err) {
-			err = xerrors.NewNonRetryableError(err)
-		}
 		return err
 	}); err != nil {
 		blocksErr := fmt.Errorf(
--- a/src/dbnode/client/session.go
+++ b/src/dbnode/client/session.go
@@ -2125,14 +2125,14 @@ func (s *session) streamBlocksMetadataFromPeers(
 				}
 
 				var (
-					currLevel          = level.value()
-					majority           = int(majority)
-					enqueued           = int(enqueued)
-					success            = int(atomic.LoadInt32(&success))
-					metReadConsistency = topology.ReadConsistencyAchieved(
-						currLevel, majority, enqueued, success)
-					doRetry = !metReadConsistency && errs.getAbortError() == nil
+					currLevel = level.value()
+					majority  = int(majority)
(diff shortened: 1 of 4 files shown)


src/dbnode/client/session.go
```diff
@@ -2319,11 +2319,28 @@ func (s *session) streamBlocksMetadataFromPeer(
...
 		return xerrors.FirstError(borrowErr, attemptErr)
 	}
 
 	for moreResults {
 		if err := s.streamBlocksRetrier.Attempt(fetchFn); err != nil {
+			// Check if the error was a hostNotAvailableError wrapped in a
+			// nonRetryableError. If so, return the hostNotAvailableError
+			// instead of the non-retryable error since the error is technically
+			// retryable, it was only wrapped to prevent the exponential backoff
+			// in the actual retrier.
```
robskillington: Maybe just make the `newHostNotAvailableError(...)` return a nonRetryableError?  

i.e. 
```go
func newHostNotAvailableError(err error) error {
	return xerrors.newNonRetryableError(hostNotAvailableError{err: err})
}

func isHostNotAvailableError(err error) bool {
  inner := xerrors.GetInnerNonRetryableError(err)
  if inner == nil {
    return false
  }
  _, ok := inner.(hostNotAvailableError)
  return ok
}
```

robskillington:
This will be a good improvement, unfortunate this scenario wasn't considered when implementing bootstrap consistency

src/dbnode/client/session.go
```diff
@@ -2833,13 +2846,6 @@ func (s *session) streamBlocksBatchFromPeer(
 			result, attemptErr = client.FetchBlocksRaw(tctx, req)
 		})
 		err := xerrors.FirstError(borrowErr, attemptErr)
-		// Do not retry if cannot borrow the connection or
-		// if the connection pool has no connections
-		switch err {
```
richardartoul: deleted this since this will be non-retryable by default

richardartoul:
@robskillington Yeah no worries, its a pretty difficult thing to consider. We only caught it because of the way Odin does node adds was triggering it in a reproducible way. Also I addressed all your feedback and added a regression test if you have a moment to review again

src/dbnode/client/session.go
```diff
@@ -2120,13 +2121,17 @@ func (s *session) streamBlocksMetadataFromPeers(
...
-				success := int(atomic.LoadInt32(&success))
 
-				doRetry := !topology.ReadConsistencyAchieved(currLevel, majority, enqueued, success) &&
-					errs.getAbortError() == nil
+				var (
+					currLevel          = level.value()
+					majority           = int(majority)
+					enqueued           = int(enqueued)
+					success            = int(atomic.LoadInt32(&success))
+					metReadConsistency = topology.ReadConsistencyAchieved(
```
prateek: more a pet peeve than anything else: move this and the next line into the (:=) assignment formulation. Calling (non stdlib) functions in a var decl makes is always a bit weird IMO

src/dbnode/client/session.go
```diff
@@ -2324,6 +2334,16 @@ func (s *session) streamBlocksMetadataFromPeer(
 
 	for moreResults {
 		if err := s.streamBlocksRetrier.Attempt(fetchFn); err != nil {
+			// Check if the error was a hostNotAvailableError. If so, return the
+			// raw (retryable) hostNotAvailableError since the error is technically
+			// retryable but was wrapped to prevent the exponential backoff in the
+			// actual retrier.
+			if isHostNotAvailableError(err) {
```
prateek: would it be better to get back a bool and the inner error from this function call? can avoid the extra fn call that way and i prefer the encapsulation of the innerNonRetryableError stuff to be hidden behind the hostNotAvailableError utility method

src/dbnode/client/session.go
```diff
@@ -66,11 +66,12 @@ import (
...
-	blocksMetadataChannelInitialCapacity = 4096
-	gaugeReportInterval                  = 500 * time.Millisecond
-	blockMetadataChBufSize               = 4096
-	shardResultCapacity                  = 4096
+	clusterConnectWaitInterval                     = 10 * time.Millisecond
+	blocksMetadataChannelInitialCapacity           = 4096
+	gaugeReportInterval                            = 500 * time.Millisecond
+	blockMetadataChBufSize                         = 4096
+	shardResultCapacity                            = 4096
+	streamBlocksMetadataFromPeerErrorSleepInterval = 1 * time.Millisecond
```
prateek: question: why so low, why not 10/100 ms?

src/dbnode/client/session.go
```diff
@@ -2151,6 +2155,12 @@ func (s *session) streamBlocksMetadataFromPeers(
 					atomic.AddInt32(&success, 1)
 					return
 				}
+
+				// Prevent the loop from spinning too aggressively if
```
prateek: how about sleeping for a longer duration only if the specialised error condition is triggered

PR #1679, opened 2019-05-30
WIP Fix DB node hanging onto file descriptors for deleted file volumes

robskillington:

This change fixes the DB node hanging onto file descriptors for deleted files by implementing a file registry for releasing resources from places like the seeker manager, etc when a file is removed.

```release-note
NONE
```

```documentation-note
NONE
```

cc @justinjc @richardartoul

### DIFF
--- a/src/dbnode/persist/fs/registry.go
+++ b/src/dbnode/persist/fs/registry.go
@@ -44,8 +44,9 @@ type fileRegistry struct {
 
 	entries map[string]*fileStatus
 
-	clockOpts clock.Options
-	logger    *zap.Logger
+	clockOpts      clock.Options
+	instrumentOpts instrument.Options
+	logger         *zap.Logger
 }
 
 // NewFileRegistry creates a new file registry.
@@ -54,9 +55,10 @@ func NewFileRegistry(
 	instrumentOpts instrument.Options,
 ) FileRegistry {
 	return &fileRegistry{
-		entries:   make(map[string]*fileStatus),
-		clockOpts: clockOpts,
-		logger:    instrumentOpts.Logger(),
+		entries:        make(map[string]*fileStatus),
+		clockOpts:      clockOpts,
+		instrumentOpts: instrumentOpts,
+		logger:         instrumentOpts.Logger(),
 	}
 }
 
@@ -127,7 +129,7 @@ func (r *fileRegistry) newFileStatusOrExisting(filePath string) (*fileStatus, er
 
 	status, ok = r.entries[pathAbs]
 	if !ok {
-		status = newFileStatus(pathAbs, r.clockOpts.NowFn(), r.logger)
+		status = newFileStatus(pathAbs, r.clockOpts.NowFn(), r.instrumentOpts)
 		r.entries[pathAbs] = status
 	}
 
@@ -142,18 +144,17 @@ type fileStatus struct {
 	refs      []*fileRef
 
 	nowFn clock.NowFn
-
-	logger *zap.Logger
+	iopts instrument.Options
 }
 
-func newFileStatus(pathAbs string, nowFn clock.NowFn, logger *zap.Logger) *fileStatus {
+func newFileStatus(pathAbs string, nowFn clock.NowFn, iopts instrument.Options) *fileStatus {
 	f := &fileStatus{
 		status: FileStatus{
 			Type: AvailableFileStatusType,
 		},
 		pathAbs: pathAbs,
 		nowFn:   nowFn,
-		logger:  logger,
+		iopts:   iopts,
 	}
 	watchable := watch.NewWatchable()
 	watchable.Update(f.status)
@@ -210,8 +211,10 @@ func (f *fileStatus) NotifyAndRemove() error {
 	}
 
 	if f.numRefs() > 0 {
-		f.logger.Error("timed out for delayed remove of file",
-			zap.String("path", f.pathAbs))
+		instrument.EmitAndLogInvariantViolation(f.iopts, func(l *zap.Logger) {
+			l.Error("delayed remove of file timed out",
+				zap.String("path", f.pathAbs))
+		})
 	}
 
 	return os.Remove(f.pathAbs)
@@ -314,8 +317,8 @@ func (f *fileRef) Readdirnames(n int) ([]string, error) {
 	return f.ref.Readdirnames(n)
 }
 
-func (f *fileRef) Seek(offset int64, whence int) (ret int64, err error) {
-
+func (f *fileRef) Seek(offset int64, whence int) (int64, error) {
+	return f.ref.Seek(offset, whence)
 }
 
 func (f *fileRef) Stat() (os.FileInfo, error) {
--- a/src/dbnode/persist/fs/registry.go
+++ b/src/dbnode/persist/fs/registry.go
@@ -1,0 +1,335 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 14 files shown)


src/dbnode/persist/fs/registry.go
```diff
@@ -0,0 +1,335 @@
...
+	f.RLock()
+	v := f.status
+	f.RUnlock()
+	return v
+}
+
+func (f *fileStatus) NotifyAndRemove() error {
+	f.setStatus(FileStatus{Type: RemovingFileStatusType})
+	defer func() {
+		f.setStatus(FileStatus{Type: RemovedFileStatusType})
```
prateek: nit: this is assuming the `os.Remove()` call doesn't fail. do you need to check that before setting the status?

src/dbnode/persist/fs/registry.go
```diff
@@ -0,0 +1,335 @@
...
+	}()
+
+	timeout := f.nowFn().Add(notifyAndRemoveTimeout)
+	for f.nowFn().Before(timeout) && f.numRefs() > 0 {
+		time.Sleep(notifyAndRemoveCheckInterval)
+	}
+
+	if f.numRefs() > 0 {
+		f.logger.Error("timed out for delayed remove of file",
```
prateek: do you need to return here too?

prateek: or is the expected semantic that the file will be removed and zombie handles are the watchers responsibility now?

src/dbnode/persist/fs/registry.go
```diff
@@ -0,0 +1,335 @@
...
+		return os.Remove(pathAbs)
+	}
+
+	// Delete from the registry.
+	delete(r.entries, pathAbs)
+
+	// Notify open file holders that this file is being removed
+	// and perform a delayed remove which will delete the file
+	// after all references are released.
+	go func() {
```
prateek: couple of questions about this:
(1) any situation where callers would want the ability to wait the file was removed before `Remove()` returns?
(2) along the same lines, you're removing the entry from the map above, but the file might not be remove because of an error returned in the goroutine. may cause funky state?

src/dbnode/persist/fs/types.go
```diff
@@ -325,6 +347,63 @@ type IndexFileSetReader interface {
 	Validate() error
 }
 
+// FileStatus describes the current file status.
```
prateek: would these types be better placed in `x/os` or some such?

prateek: (FileStatus/File/Registry/..)

src/dbnode/persist/fs/types.go
```diff
@@ -325,6 +347,63 @@ type IndexFileSetReader interface {
...
+}
+
+// FileStatusWatcher implements a watcher of file status.
+type FileStatusWatcher interface {
+	OnFileStatusChange(f File, status FileStatus)
+}
+
+// FileRegistry is a file lifecycle manager and should always be used in place
+// of os.Open to ensure
+type FileRegistry interface {
```
prateek: Is `Filesystem` a better name/abstraction? 

Would be better to avoid mixing other FD calls between this type and the os package. e.g. chmod/stat/rename/etc. Can be as simple as pass thru for now. 

Down the road, this would let us do nice things like mock the filesystem used during tests. All manner of benefits including speed/injection/etc.

prateek: Found another package abstracting a FS here: https://github.com/spf13/afero#list-of-all-available-functions

src/dbnode/persist/fs/registry.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+func (f *fileRef) PathAbs() string {
+	return f.file.pathAbs
+}
+
+func (f *fileRef) WatchStatus(watcher FileStatusWatcher) close.SimpleCloser {
+	_, watch, _ := f.file.watchable.Watch()
+
+	// Spawn a new goroutine that will terminate when the
```
prateek: should the lifecycle be controllable directly? like what if i want to open a file, watch for some status changes and then stop watching (w/o removing)

src/dbnode/persist/fs/registry.go
robskillington: Yeah I can do that, although I feel we probably want to close refs to the file even if a close only partially fails perhaps?

src/dbnode/persist/fs/registry.go
robskillington: I was thinking we just delete anyway, thoughts?  Otherwise a server restart wouldn't fix the issue.

src/dbnode/persist/fs/registry.go
```diff
@@ -0,0 +1,335 @@
...
+type fileRegistry struct {
+	sync.RWMutex
+
+	entries map[string]*fileStatus
+
+	clockOpts clock.Options
+	logger    *zap.Logger
+}
+
+// NewFileRegistry creates a new file registry.
```
richardartoul: A quick paragraph explaining what this does / how its used would be nice

src/dbnode/persist/fs/registry.go
richardartoul: Yeah this feels racey and off to me.

Scenario:

```
1. Cleanup manager calls Remove() on some file. File exists in map, gets deleted from map, spawns background routine, then returns nil.
2. Background routine fails (for whatever reason).
3. Cleanup manager comes around again, finds the files on disk (because they haven't been deleted), then calls delete again, its not in this map, so we just issue a `os.Remove()` even though the filemanager may still have a handle on it
```

Similar scenario applies for the merge/compaction logic I think.

On top of that, if we have this operation async we can no longer serialize access to our data and snapshot files. I.E right now the logic is:

1. Data flush
2. Data cold flush
3. Data snapshot
4. Cleanup

perform in a synchronous loop with no overlap. With this background goroutine we lose all control over that and the processes will be competing.

An alternative approach that might work nicely is to just provide a batch API, so something like

```golang
func (r *fileRegistry) Remove(filePaths []string) error {
    for _, fPath := range filePaths {
        // If not in map, remove and continue
        // Call NotifyAndRemove() and wait for success (can have some parallelism here)
        // If success, remove from map, other keep it in and pass back an []error or something for all the ones that failed
    }

    wg.Wait() // Wait for all of them to complete or error out
}
```

So now the map state is always "correct" hopefully, we don't end up with overlapping processes, and it should work nicely with the compaction logic too because that can just accumulate a list of files to be removed and then delete them all at once at the end

This may not be the correct answer but I think the existing one may cause more harm than good by introducing an async layer here

src/dbnode/persist/fs/registry.go
```diff
@@ -0,0 +1,335 @@
...
+
+func (f *fileStatus) removeFileRef(elem *fileRef) error {
+	f.Lock()
+	defer f.Unlock()
+
+	if len(f.refs) == 0 {
+		return fmt.Errorf("remove file ref not found for: %v", elem.PathAbs())
+	}
+
+	refs := make([]*fileRef, 0, len(f.refs)-1)
```
richardartoul: Think its worth just reslicing and performing a copy? Maybe an over-optimization, but I can imagine this happening in a tight loop when bblocks expire and such

src/dbnode/persist/fs/registry.go
richardartoul: Could we just do this all sync and return an error here? Then whatever process encountered the issue can at least try again

richardartoul: Well the NotifyAndRemove can still be done async in parallel, but then we would wait for them to all complete in `Remove()` and if they timedout they would return errors instead of just deleting anyways because then we can try again from the callers perspective

src/dbnode/persist/fs/registry.go
```diff
@@ -0,0 +1,335 @@
...
+
+func (f *fileRef) Readdir(n int) ([]os.FileInfo, error) {
+	return f.ref.Readdir(n)
+}
+
+func (f *fileRef) Readdirnames(n int) ([]string, error) {
+	return f.ref.Readdirnames(n)
+}
+
+func (f *fileRef) Seek(offset int64, whence int) (ret int64, err error) {
```
richardartoul: this is just a TODO?

src/dbnode/persist/fs/types.go
richardartoul: I think mocking the entire filesystem would be a lot more involved than just what this P.R is doing no? We still need to be able to list all directories in a folder, etc etc

src/dbnode/persist/fs/types.go
justinjc: Yeah, having different `File` interfaces for mmap, digest, and persist/fs is likely to cause confusion I think and we'll probably have to use some conversion logic between them at some point.

robskillington:
Solved already.

PR #1682, opened 2019-05-30
[query] Remove empty series from output when dropping NaNs

arnikola:

Drops empty series from appearing in output when keepNaNs option is set to disabled. This is important for functions like `topk`, which leave a lot of empty series and can crash Grafana.

```release-note
Will no longer add empty series to output
```

```documentation-note
NONE
```


src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -240,12 +240,37 @@ func parseQuery(r *http.Request) (string, error) {
...
+	}
+
+	return filtered
+}
+
 func renderResultsJSON(
 	w io.Writer,
 	series []*ts.Series,
 	params models.RequestParams,
 	keepNans bool,
```
robskillington: Is this ever true?  If not, can we remove the option entirely perhaps?

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -240,12 +240,37 @@ func parseQuery(r *http.Request) (string, error) {
 	return queries[0], nil
 }
 
+func sanitizeSeries(series []*ts.Series) []*ts.Series {
```
robskillington: nit: Can we rename this perhaps just to `removeAllNaNsSeries(...)` or similar?  Not clear what sanitize does from just the name.

src/query/api/v1/handler/prometheus/native/common.go
arnikola: I'm pretty happy to remove it since I don't really know why you'd want to keep nans, but you'll have to convert Ben :p

src/query/api/v1/handler/prometheus/native/common.go
arnikola: sg

arnikola: Changed to `filterNaNSeries`

src/query/api/v1/handler/prometheus/native/common.go
arnikola: Oh, if you're asking if it can theoretically be true, there's a config option to keep nans in the output; don't know why anyone would turn it on, but it's there :p

PR #1683, opened 2019-05-30
load schema registry from config for m3db client

Haijuncao:
Load schema from config for m3db client if it is set in config. 
If m3db client is initialized from dbnode, the schema registry can be overridden by custom admin option. 

This way initialize m3db client from application side is simplified. We can go ahead and remove similiar code from location store m3db side.


src/dbnode/client/config.go
```diff
@@ -310,6 +312,19 @@ func (c Configuration) NewAdminClient(
 
 	if c.Proto != nil && c.Proto.Enabled {
 		v = v.SetEncodingProto(encodingOpts)
+		schemaRegistry := namespace.NewSchemaRegistry(true, nil)
+		// Load schema from config if it is available.
+		// If admin client is initialized by dbnode, the schema registry can be overridden
+		// by admin custom options.
+		for nsID, protoConfig := range c.Proto.SchemaRegistry {
+			dummyDeployID := "fromconfig"
```
richardartoul: super nit: Can we make this a constant in some package since we use it on the server side and client side? Would be nice to have those match

Haijuncao: I have a follow up pr that require client side config have schema deploy ID set. (to the one that is the real deploy ID)

PR #1684, opened 2019-05-31
[query] Drop series with no values in the given range

arnikola:

Follows on from #1682 to also remove series with no NaN-values in the given range, as they were still appearing as empty series.

```release-note
Will no longer add series with no non-NaN points in the given range
```

```documentation-note
NONE
```

### DIFF
--- a/src/query/api/v1/handler/prometheus/native/common.go
+++ b/src/query/api/v1/handler/prometheus/native/common.go
@@ -240,9 +240,10 @@ func parseQuery(r *http.Request) (string, error) {
 	return queries[0], nil
 }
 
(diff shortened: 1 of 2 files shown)


src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -240,13 +240,21 @@ func parseQuery(r *http.Request) (string, error) {
 	return queries[0], nil
 }
 
-func filterNaNSeries(series []*ts.Series) []*ts.Series {
+func filterNaNSeries(series []*ts.Series,
+	start time.Time,
```
prateek: nit: rename to startInclusive and endInclusive

PR #1686, opened 2019-05-31
[query] Update query/opentracing to x/opentracing

benraskin92:

Since we merged m3x into m3db, we can now utilize the `opentracing` package in `src/x` instead of the one currently in `src/query`. 

```release-note
NONE
```

```documentation-note
NONE
```


src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -34,8 +34,8 @@ import (
 	"github.com/m3db/m3/src/query/ts"
 	"github.com/m3db/m3/src/query/util/httperrors"
 	"github.com/m3db/m3/src/query/util/logging"
-	opentracingutil "github.com/m3db/m3/src/query/util/opentracing"
 	xhttp "github.com/m3db/m3/src/x/net/http"
+	xopentracing "github.com/m3db/m3/src/x/opentracing"
```
arnikola: nit: can probably drop the `xopentracing` alias and leave it as `opentracing`

benraskin92: Wanted to distinguish from the opentracing packages below

PR #1687, opened 2019-05-31
[Query] Add multi-zonal capability to remote reads

arnikola:

Adds multi-zone capability to remote reads for Query

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/cmd/services/m3query/config/config.go
+++ b/src/cmd/services/m3query/config/config.go
@@ -406,6 +406,18 @@ type ClusterManagementConfiguration struct {
 	Etcd etcdclient.Configuration `yaml:"etcd"`
 }
 
+// RemoteZones is a set of remote zone configurations.
+type RemoteZones []RemoteZone
+
+// RemoteZone is the configuration for a single remote zone.
+type RemoteZone struct {
+	// Name is the name for the remote zone.
+	Name string `yaml:"name"`
+	// RemoteListenAddresses is the remote listen addresses to call for remote
+	// coordinator calls in the remote zone.
+	RemoteListenAddresses []string `yaml:"remoteListenAddresses"`
+}
+
 // RPCConfiguration is the RPC configuration for the coordinator for
 // the GRPC server used for remote coordinator to coordinator calls.
 type RPCConfiguration struct {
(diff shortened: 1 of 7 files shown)


src/cmd/services/m3query/config/config.go
```diff
@@ -415,8 +427,14 @@ type RPCConfiguration struct {
 	// ListenAddress is the RPC server listen address.
 	ListenAddress string `yaml:"listenAddress"`
 
-	// RemoteListenAddresses is the remote listen addresses to call for remote
-	// coordinator calls.
+	// RemoteZones are the configuration settings for remote coordinator zones.
+	RemoteZones RemoteZones `yaml:"remoteZones"`
```
robskillington: nit: Can we call this just `RemoteClusters`?  Technically speaking they could also be regions, etc.

robskillington: Second though, maybe just call it `Remotes` and have a `type Remote struct` instead of `type RemoteZone struct`

src/query/server/server.go
```diff
@@ -722,36 +730,94 @@ func newStorages(
...
+func remoteZoneStorage(
+	remoteAddresses []string,
+	lookbackDuration time.Duration,
+	tagOptions models.TagOptions,
+	poolWrapper *pools.PoolWrapper,
+	readWorkerPool xsync.PooledWorkerPool,
+) (storage.Storage, error) {
+	if len(remoteAddresses) == 0 {
+		// No addresses; skip.
+		return nil, nil
```
robskillington: Hm, perhaps make this an error?  This would be a very strange config to call "good" and might end up with unintended mistakes in config that is hard to debug.

src/cmd/services/m3query/config/config.go
arnikola: I gotta be honest and say I couldn't decide on a name for this so I went with `RemoteZones` because I knew it would be contentious in the PR haha; `remotes` sounds good

src/query/server/server.go
arnikola: So this is a "valid" case, where act as a server to remote zones but are not connected as a client (maybe a case for setting up dedicated servers to host cross zone traffic or something like that)

PR #1690, opened 2019-05-31
[query] Make executor engine an interface

benraskin92:

This PR turns the engine in `query/executor` into an interface. This will allow us to better test queries as we can easily mock out the engine.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/query/api/v1/handler/prometheus/native/read.go
+++ b/src/query/api/v1/handler/prometheus/native/read.go
@@ -62,7 +62,7 @@ var (
 
 // PromReadHandler represents a handler for prometheus read endpoint.
 type PromReadHandler struct {
-	engine              *executor.Engine
+	engine              executor.Engine
 	fetchOptionsBuilder handler.FetchOptionsBuilder
 	tagOpts             models.TagOptions
 	limitsCfg           *config.LimitsConfiguration
@@ -107,7 +107,7 @@ type RespError struct {
 
 // NewPromReadHandler returns a new instance of handler.
 func NewPromReadHandler(
-	engine *executor.Engine,
+	engine executor.Engine,
 	fetchOptionsBuilder handler.FetchOptionsBuilder,
 	tagOpts models.TagOptions,
 	limitsCfg *config.LimitsConfiguration,
@@ -138,11 +138,12 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	result, params, respErr := h.ServeHTTPWithEngine(w, r, h.engine, &executor.EngineOptions{
-		QueryContextOptions: models.QueryContextOptions{
-			LimitMaxTimeseries: fetchOpts.Limit,
-		},
+	engineOpts := h.engine.Opts().SetQueryContextOptions(models.QueryContextOptions{
+		LimitMaxTimeseries: fetchOpts.Limit,
 	})
+	h.engine = h.engine.SetOpts(engineOpts)
+
+	result, params, respErr := h.ServeHTTPWithEngine(w, r, h.engine)
 	if respErr != nil {
 		httperrors.ErrorWithReqInfo(w, r, respErr.Code, respErr.Err)
 		return
@@ -166,8 +167,7 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 func (h *PromReadHandler) ServeHTTPWithEngine(
 	w http.ResponseWriter,
 	r *http.Request,
-	engine *executor.Engine,
-	opts *executor.EngineOptions,
+	engine executor.Engine,
 ) ([]*ts.Series, models.RequestParams, *RespError) {
 	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
 	logger := logging.WithContext(ctx)
@@ -187,7 +187,7 @@ func (h *PromReadHandler) ServeHTTPWithEngine(
 		return nil, emptyReqParams, &RespError{Err: err, Code: http.StatusBadRequest}
 	}
 
-	result, err := read(ctx, engine, opts, h.tagOpts, w, params)
+	result, err := read(ctx, engine, h.tagOpts, w, params)
 	if err != nil {
 		sp := xopentracing.SpanFromContextOrNoop(ctx)
 		sp.LogFields(opentracinglog.Error(err))
--- a/src/query/api/v1/handler/prometheus/native/read.go
+++ b/src/query/api/v1/handler/prometheus/native/read.go
@@ -138,12 +138,12 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	engineOpts := h.engine.Opts().SetQueryContextOptions(models.QueryContextOptions{
-		LimitMaxTimeseries: fetchOpts.Limit,
-	})
-	h.engine = h.engine.SetOpts(engineOpts)
+	queryOpts := &executor.QueryOptions{
+		QueryContextOptions: models.QueryContextOptions{
+			LimitMaxTimeseries: fetchOpts.Limit,
+		}}
 
-	result, params, respErr := h.ServeHTTPWithEngine(w, r, h.engine)
+	result, params, respErr := h.ServeHTTPWithEngine(w, r, h.engine, queryOpts)
 	if respErr != nil {
 		httperrors.ErrorWithReqInfo(w, r, respErr.Code, respErr.Err)
 		return
@@ -168,6 +168,7 @@ func (h *PromReadHandler) ServeHTTPWithEngine(
 	w http.ResponseWriter,
 	r *http.Request,
 	engine executor.Engine,
+	opts *executor.QueryOptions,
 ) ([]*ts.Series, models.RequestParams, *RespError) {
 	ctx := context.WithValue(r.Context(), handler.HeaderKey, r.Header)
 	logger := logging.WithContext(ctx)
@@ -187,7 +188,7 @@ func (h *PromReadHandler) ServeHTTPWithEngine(
 		return nil, emptyReqParams, &RespError{Err: err, Code: http.StatusBadRequest}
 	}
 
-	result, err := read(ctx, engine, h.tagOpts, w, params)
+	result, err := read(ctx, engine, opts, h.tagOpts, w, params)
 	if err != nil {
 		sp := xopentracing.SpanFromContextOrNoop(ctx)
 		sp.LogFields(opentracinglog.Error(err))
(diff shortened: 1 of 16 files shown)


src/query/executor/types.go
```diff
@@ -0,0 +1,88 @@
...
+		parser parser.Parser,
+		params models.RequestParams,
+		results chan Query,
+	)
+
+	// Close kills all running queries and prevents new queries from being attached.
+	Close() error
+
+	// Opts returns the engine options.
+	Opts() EngineOptions
```
andrewmains12: Do we need these methods? Seems like options in general should be internal to the engines.

src/query/executor/types.go
```diff
@@ -0,0 +1,88 @@
...
+	// SetOpts sets the engine options.
+	SetOpts(EngineOptions) Engine
+}
+
+// EngineOptions can be used to pass custom flags to engine.
+type EngineOptions interface {
```
andrewmains12: Does this need to be an interface? Does it have multiple feasible implementations?

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -138,11 +138,12 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	result, params, respErr := h.ServeHTTPWithEngine(w, r, h.engine, &executor.EngineOptions{
-		QueryContextOptions: models.QueryContextOptions{
-			LimitMaxTimeseries: fetchOpts.Limit,
-		},
+	engineOpts := h.engine.Opts().SetQueryContextOptions(models.QueryContextOptions{
```
prateek: why not do this at the level above this?

src/query/executor/types.go
benraskin92: We need to be able to set these outside of this package (for example, in native/read.go):
```
	engineOpts := h.engine.Opts().SetQueryContextOptions(models.QueryContextOptions{
		LimitMaxTimeseries: fetchOpts.Limit,
	})
	h.engine = h.engine.SetOpts(engineOpts)
```

src/query/api/v1/handler/prometheus/native/read.go
benraskin92: I was just keeping this where it was originally which is in this function.

arnikola: Theoretically users can override the time series limit in their query, so we'd need to update the Option limit here unfortunately.

src/query/executor/types.go
benraskin92: I was just being consistent with what we do for other option types.

src/query/api/v1/handler/prometheus/native/read.go
```diff
@@ -138,11 +138,12 @@ func (h *PromReadHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		LimitMaxTimeseries: fetchOpts.Limit,
 	})
+	h.engine = h.engine.SetOpts(engineOpts)
```
arnikola: This feels really racy if multiple requests mutate `h.engine`; may be better to keep the QueryContextOptions separate

src/query/executor/types.go
andrewmains12: So query context is a method parameter throughout most of the code base; why not leave it that way (instead of removing the `opts` from the call)?

src/query/api/v1/handler/prometheus/native/read_instantaneous.go
```diff
@@ -87,13 +87,12 @@ func (h *PromReadInstantHandler) ServeHTTP(w http.ResponseWriter, r *http.Reques
...
 
-	engineOpts := &executor.EngineOptions{
-		QueryContextOptions: models.QueryContextOptions{
-			LimitMaxTimeseries: fetchOpts.Limit,
-		},
-	}
+	engineOpts := h.engine.Opts().SetQueryContextOptions(models.QueryContextOptions{
+		LimitMaxTimeseries: fetchOpts.Limit,
+	})
+	h.engine = h.engine.SetOpts(engineOpts)
```
arnikola: Again, this feels racy

src/query/api/v1/handler/prometheus/remote/read.go
```diff
@@ -187,15 +187,14 @@ func (h *PromReadHandler) read(
...
 	results := make(chan *storage.QueryResult)
-	opts := &executor.EngineOptions{
-		QueryContextOptions: models.QueryContextOptions{
-			LimitMaxTimeseries: limit,
-		},
-	}
+	engineOpts := h.engine.Opts().SetQueryContextOptions(models.QueryContextOptions{
+		LimitMaxTimeseries: limit,
+	})
+	h.engine = h.engine.SetOpts(engineOpts)
```
arnikola: Again, this feels racy

src/query/api/v1/handler/prometheus/remote/read_test.go
```diff
@@ -71,8 +71,10 @@ func setupServer(t *testing.T) *httptest.Server {
 
 func readHandler(store storage.Storage, timeoutOpts *prometheus.TimeoutOpts) *PromReadHandler {
 	opts := handler.FetchOptionsBuilderOptions{Limit: 100}
+	engineOpts := executor.NewEngineOpts().SetStore(store).SetCostScope(tally.NewTestScope("test", nil)).
```
arnikola: nit: may be easier to pull engine generation into a new function

(or better yet delete it and use a mock here instead?)

src/query/api/v1/httpd/handler_test.go
```diff
@@ -64,11 +64,12 @@ func makeTagOptions() models.TagOptions {
 
 func setupHandler(store storage.Storage) (*Handler, error) {
 	downsamplerAndWriter := ingest.NewDownsamplerAndWriter(store, nil, testWorkerPool)
+	engineOpts := executor.NewEngineOpts().SetStore(store).SetCostScope(tally.NewTestScope("test", nil)).
```
arnikola: Same here; re function to generate the engine / mock it out

src/query/executor/engine.go
```diff
@@ -55,20 +46,15 @@ type Query struct {
...
 // NewEngine returns a new instance of QueryExecutor.
 func NewEngine(
-	store storage.Storage,
-	scope tally.Scope,
-	lookbackDuration time.Duration,
-	factory qcost.ChainedEnforcer,
-) *Engine {
-	if factory == nil {
-		factory = qcost.NoopChainedEnforcer()
+	engineOpts EngineOptions,
```
arnikola: :+1: this is so much nicer

src/query/api/v1/handler/prometheus/native/read.go
andrewmains12: +1 -- if this is per call, pass it down in the method args.

src/query/executor/engine.go
```diff
@@ -118,19 +104,27 @@ func newEngineMetrics(scope tally.Scope) *engineMetrics {
 	}
 }
 
-// Execute runs the query and closes the results channel once done
-func (e *Engine) Execute(
+func (e *engine) Opts() EngineOptions {
```
arnikola: nit: Rename to `Options()`

src/query/executor/engine.go
```diff
@@ -118,19 +104,27 @@ func newEngineMetrics(scope tally.Scope) *engineMetrics {
...
+	return e.opts
+}
+
+func (e *engine) SetOpts(v EngineOptions) Engine {
```
arnikola: nit: rename to `SetOptions(..)`

src/query/executor/types.go
andrewmains12: Fair enough; I'm fine with leaving this. Ignore the rest of this comment for purposes

I'm going to write up a "Options interfaces considered harmful" doc one of these days--cc @prateek since I think you'll have an opinion on that  :). 

Briefly:

- It's lots of boilerplate
- There's only one implementation for all of these; mocking a plain-old-object/value struct makes little sense to me.
- It doesn't hide implementation details any better than just doing:

```
type MyOptions struct {
   // private
   f1 int
}

func (mo *MyOptions) F1() int {
}

// setters etc
```

I actually think we should do struct configuration more like the way most FX modules do, but it's definitely a broader conversation than this PR :)

src/query/executor/options.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+func (o *engineOptions) SetLookbackDuration(v time.Duration) EngineOptions {
+	opts := *o
+	opts.lookbackDuration = v
+	return &opts
+}
+
+func (o *engineOptions) QueryContextOptions() models.QueryContextOptions {
+	return o.queryContextOptions
```
arnikola: Might be better to keep this separate unfortunately as mentioned above

src/query/executor/types.go
```diff
@@ -0,0 +1,88 @@
...
+type Engine interface {
+	// Execute runs the query and closes the results channel once done.
+	Execute(
+		ctx context.Context,
+		query *storage.FetchQuery,
+		results chan *storage.QueryResult,
+	)
+
+	// ExecuteExpr runs the query DAG and closes the results channel once done.
+	// nolint: unparam
```
arnikola: nit: think you can get rid of this `// nolint`

src/query/executor/types.go
arnikola: It's a fairly common pattern we use elsewhere in the codebase (e.g. [here](https://sourcegraph.com/github.com/m3db/m3@arnikola/multi-zone-coordinator/-/blob/src/dbnode/storage/types.go#L725:6))

Might not be super idiomatic go, but it's consistent with the codebase :+1:

src/query/executor/types.go
benraskin92: Yeah I'll pull it out of the opts.

src/query/executor/types.go
arnikola: Oops, github hadn't updated with your comment haha, ignore me

src/query/executor/types.go
arnikola: Oops github didn't update, ignore

src/query/executor/types.go
andrewmains12: > Might not be super idiomatic go, but it's consistent with the codebase 👍

Hahaha I was going to comment with "oh trust me, I know". This is on my list of "Things I am (probably unnecessarily) salty about".

src/query/api/v1/handler/prometheus/native/read.go
robskillington: Yeah, do not want to mutate the engine itself on a per call basis.

src/query/api/v1/handler/prometheus/native/read_instantaneous.go
robskillington: Yes this will race per request, need to avoid that.

src/query/api/v1/handler/prometheus/remote/read_test.go
benraskin92: Let me pull into a func for now and i can add mocks in another pr.

PR #1692, opened 2019-06-01
[coordinator] Return HTTP 400 if all sent samples encounter too old/new or other bad request error

robskillington:

This should fix issue https://github.com/m3db/m3/issues/1709.

```release-note
NONE
```

```documentation-note
NONE
```


robskillington:
Need to add tests still.

src/query/api/v1/handler/prometheus/remote/write.go
```diff
@@ -99,14 +101,33 @@ func (h *PromWriteHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-	if err != nil {
+	batchErr := h.write(r.Context(), req)
+	if batchErr != nil {
+		numBadRequest := 0
+		errs := batchErr.Errors()
+		lastErr := batchErr.LastError()
+		for _, err := range errs {
+			switch {
+			case client.IsBadRequestError(err):
+				fallthrough
```
arnikola: Might be worth having a different output error here if it's bad requests vs invalid params specifically?

robskillington: We need both to be HTTP 400 (since local aggregation can also cause a "too old" error, which will be invalidparams not a client bad request error).

The error actually returned to the client I haven't changed, it will always be the "last error".

src/cmd/services/m3coordinator/ingest/write.go
```diff
@@ -58,11 +58,18 @@ type DownsamplerAndWriter interface {
...
 		iter DownsampleAndWriteIter,
-	) error
+	) BatchError
 
 	Storage() storage.Storage
 }
 
+// BatchError allows for access to individual errors.
+type BatchError interface {
+	error
```
prateek: Considering we’re only returning the last error in this path, can we add a new implementation of multierr which doesn’t accumulate?

scripts/development/m3_stack/start_m3.sh
```diff
@@ -183,7 +186,7 @@ echo "Validating topology"
 echo "Done validating topology"
 
 echo "Waiting until shards are marked as available"
-ATTEMPTS=10 TIMEOUT=2 retry_with_backoff  \
+ATTEMPTS=100 TIMEOUT=2 retry_with_backoff  \
```
arnikola: Yes! Much nicer haha, can we do the same for integration tests?

src/cmd/services/m3coordinator/ingest/write.go
robskillington: So I need to actually look at each error and determine 100% of them are bad request like errors, otherwise if there was one real error, the request needs to be retried.

I use `Errors()` to first establish all real errors are bad request like errors, then I use `LastError()` just to grab the one for logging purposes.  I could probably remote `LastError()` tbh from this interface.

scripts/development/m3_stack/start_m3.sh
robskillington: True yeah, will update the common ones.

PR #1693, opened 2019-06-01
[dbnode] Release refs to tags after data fileset write close

robskillington:

This fixes the fileset volume writer not releasing the references to tags after closing and being reset.

This didn't have any noticeable impact, I just noticed it as part of a code review I was doing.  Added some test coverage to make sure this doesn't regress.

```release-note
NONE
```

```documentation-note
NONE
```


src/dbnode/persist/fs/read_write_test.go
```diff
@@ -131,6 +131,19 @@ func writeTestDataWithVolume(
 			digest.Checksum(entries[i].data)))
 	}
 	assert.NoError(t, w.Close())
+
+	// Assert that any index entries released references they held
```
benraskin92: nit: We've been putting `.`'s at the end of all comments.

src/dbnode/persist/fs/read_write_test.go
```diff
@@ -131,6 +131,19 @@ func writeTestDataWithVolume(
...
+	writer, ok := w.(*writer)
+	require.True(t, ok)
+
+	// Take ref to wholly allocated index entries slice
+	slice := writer.indexEntries[:cap(writer.indexEntries)]
+
+	// Check every entry has ID and Tags nil
+	for _, elem := range slice {
+		assert.Nil(t, elem.id)
+		assert.Nil(t, elem.tags.Values())
```
benraskin92: Would this be `nil` or just empty `ident.Tags{}`?

src/dbnode/persist/fs/write.go
```diff
@@ -98,6 +98,7 @@ type indexEntries []indexEntry
 func (e indexEntries) releaseRefs() {
 	for i := range e {
 		e[i].id = nil
+		e[i].tags = ident.Tags{}
```
prateek: Can you just use an empty index entry instead of this?

```
var empty indexEntry
for i := range e {
  e[i] = empty
}
```

This way you get the memclr optimization too

arnikola: Just for me to understand, is this equivalent to `the` memset optimizations?

prateek: Yep

robskillington: Yeah fair call

PR #1697, opened 2019-06-03
update namespace metadata dynamically

Haijuncao:
namespace bootstrap depends on namespace.metadata for schema.

We need to deliver dynamic schema to namespace metadata as well, since bootstrap can be invoked when new shards are assigned to a node.

### DIFF
--- a/src/dbnode/namespace/schema_registry.go
+++ b/src/dbnode/namespace/schema_registry.go
@@ -154,21 +154,13 @@ func (sr *schemaRegistry) RegisterListener(
 	<-watch.C()
 
 	// Deliver the current schema
-	if schema, ok := watchable.Get().(SchemaHistory).GetLatest(); ok {
-		listener.SetSchema(schema)
-	} else if sr.logger != nil {
-		sr.logger.Error("can not update schema listener with empty schema history")
-	}
+	listener.SetSchemaHistory(watchable.Get().(SchemaHistory))
 
 	// Spawn a new goroutine that will terminate when the
 	// watchable terminates on the close of the runtime options manager
 	go func() {
 		for range watch.C() {
-			if schema, ok := watchable.Get().(SchemaHistory).GetLatest(); ok {
-				listener.SetSchema(schema)
-			} else if sr.logger != nil {
-				sr.logger.Error("can not update schema listener with empty schema history")
-			}
+			listener.SetSchemaHistory(watchable.Get().(SchemaHistory))
 		}
 	}()
 
--- a/src/dbnode/storage/namespace.go
+++ b/src/dbnode/storage/namespace.go
@@ -365,11 +365,23 @@ func newDatabaseNamespace(
 	return n, nil
 }
 
-// SetSchema implements namespace.SchemaListener.
-func (n *dbNamespace) SetSchema(value namespace.SchemaDescr) {
+// SetSchemaHistory implements namespace.SchemaListener.
+func (n *dbNamespace) SetSchemaHistory(value namespace.SchemaHistory) {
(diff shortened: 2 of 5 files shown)


src/dbnode/storage/namespace.go
```diff
@@ -365,11 +365,23 @@ func newDatabaseNamespace(
...
-	n.Unlock()
+	defer n.Unlock()
+
+	metadata, err := namespace.NewMetadata(n.ID(), n.nopts.SetSchemaHistory(value))
+	if err != nil {
+		n.log.Error("can not update namespace metadata with empty schema history", zap.Stringer("namespace", n.ID()), zap.Error(err))
+		return
+	}
+
+	if schema, ok := value.GetLatest(); ok {
```
richardartoul: Could you move this check to the top and just do an early return if the input is invalid? I.E:

```golang
n.Lock()
defer n.Unlock()

schema, ok := value.GetLatest()
if !ok {
    // error log
    return
}

metadata, err := ...
if err != nil {
    // log error
    return
}

n.schemaDescr = ...
n.metadata = ...
```

src/dbnode/namespace/schema_registry.go
```diff
@@ -154,21 +154,13 @@ func (sr *schemaRegistry) RegisterListener(
 	<-watch.C()
 
 	// Deliver the current schema
-	if schema, ok := watchable.Get().(SchemaHistory).GetLatest(); ok {
-		listener.SetSchema(schema)
-	} else if sr.logger != nil {
-		sr.logger.Error("can not update schema listener with empty schema history")
-	}
+	listener.SetSchemaHistory(watchable.Get().(SchemaHistory))
```
richardartoul: Curious why you're removing this check from the registry itself. Just to make it more flexible? Otherwise seems prudent to have the check at multiple layers but curious what your thoughts are

Haijuncao: It is mostly because the listener interface is changed from SetSchema to SetSchemaHistory, so there is no need to call GetLatest here.

Another reason is that the decision that whether or not an empty schema history is valid is better left to the caller to make.

src/dbnode/storage/namespace.go
Haijuncao: The previous block is also a sanity check and early return. Only after both blocks are valid, will we do an assignment to member var.

src/dbnode/storage/namespace.go
```diff
@@ -764,6 +777,7 @@ func (n *dbNamespace) Bootstrap(start time.Time, process bootstrap.Process) erro
 	callStart := n.nowFn()
 
 	n.Lock()
+	metadata := n.metadata
```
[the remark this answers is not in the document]
Haijuncao: I wanted to protect n.metadata read access, the previous code is correct, but I worry that it is too far away from n.Lock() (even though it is still within the lock block), not obvious enough.

src/dbnode/namespace/schema_registry.go
richardartoul: ok sure

src/dbnode/storage/namespace.go
richardartoul: yeah I know but I think validating the input (the history) first before trying to make the metadata makes more sense. Also it reduces the nesting and just makes the logic easier to follow if its a series of guard clauses followed by doing the actual update with no elses

src/dbnode/storage/namespace.go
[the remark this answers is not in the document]
richardartoul: sure

PR #1698, opened 2019-06-03
[query]Add debug param to namespace get API

benraskin92:

This PR allows you to pass in an optional `debug` parameter to the `/api1/v1/namespace` endpoint in the coordinator, which returns durations instead of nanoseconds for the duration fields. This should help with readability and debugging. Below is a sample output using start_m3. 
```
{
  "registry": {
    "namespaces": {
      "metrics_0_30m": {
        "bootstrapEnabled": true,
        "cleanupEnabled": true,
        "flushEnabled": true,
        "indexOptions": {
          "blockSizeDuration": "10m0s",
          "enabled": true
        },
        "retentionOptions": {
          "blockDataExpiry": true,
          "blockDataExpiryAfterNotAccessPeriodDuration": "5m0s",
          "blockSizeDuration": "10m0s",
          "bufferFutureDuration": "5m0s",
          "bufferPastDuration": "5m0s",
          "retentionPeriodDuration": "30m0s"
        },
        "snapshotEnabled": true,
        "writesToCommitLog": true
      },
      "metrics_10s_48h": {
        "bootstrapEnabled": true,
        "cleanupEnabled": true,
        "flushEnabled": true,
        "indexOptions": {
          "blockSizeDuration": "4h0m0s",
          "enabled": true
        },
        "retentionOptions": {
          "blockDataExpiry": true,
          "blockDataExpiryAfterNotAccessPeriodDuration": "5m0s",
          "blockSizeDuration": "4h0m0s",
          "bufferFutureDuration": "10m0s",
          "bufferPastDuration": "10m0s",
          "retentionPeriodDuration": "48h0m0s"
        },
        "snapshotEnabled": true,
        "writesToCommitLog": true
      }
    }
  }
}
```

```release-note
Users can now pass in a `debug` parameter when making a call to the `/api/v1/namespace` endpoint in coordinator. When set to `true` it allows for better readability. 
```

```documentation-note
Added docs to the `namespace_configuration.md` file under the operational guides.
```

### DIFF
--- a/src/query/api/v1/handler/namespace/get.go
+++ b/src/query/api/v1/handler/namespace/get.go
@@ -21,10 +21,11 @@
 package namespace
 
 import (
+	"bytes"
 	"fmt"
 	"net/http"
 	"path"
-	"strings"
+	"strconv"
 
 	clusterclient "github.com/m3db/m3/src/cluster/client"
 	"github.com/m3db/m3/src/cluster/kv"
@@ -51,6 +52,10 @@ var (
 	GetHTTPMethod = http.MethodGet
 )
 
+const (
+	debugParam = "debug"
+)
+
 // GetHandler is the handler for namespace gets.
 type GetHandler Handler
 
@@ -74,8 +79,7 @@ func (h *GetHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		Registry: &nsRegistry,
 	}
 
-	urlParams := r.URL.Query()
-	if strings.ToLower(urlParams.Get("debug")) == "true" {
+	if debug, err := strconv.ParseBool(r.URL.Query().Get(debugParam)); err == nil && debug {
 		nanosToDurationMap, err := nanosToDuration(resp)
 		if err != nil {
 			logger.Error("error converting nano fields to duration", zap.Error(err))
@@ -119,13 +123,10 @@ func (h *GetHandler) Get() (nsproto.Registry, error) {
 
 func nanosToDuration(resp proto.Message) (map[string]interface{}, error) {
 	marshaler := jsonpb.Marshaler{EmitDefaults: true}
-	nsString, err := marshaler.MarshalToString(resp)
-	if err != nil {
-		return nil, fmt.Errorf("unable to marshal namespace to string: %v", err)
-	}
+	buf := new(bytes.Buffer)
+	marshaler.Marshal(buf, resp)
 
-	ioReader := strings.NewReader(nsString)
-	toDuration, err := xhttp.NanosToDurationBytes(ioReader)
+	toDuration, err := xhttp.NanosToDurationBytes(buf)
 	if err != nil {
 		return nil, err
 	}
--- a/src/query/api/v1/handler/namespace/get.go
+++ b/src/query/api/v1/handler/namespace/get.go
@@ -24,6 +24,7 @@ import (
 	"fmt"
 	"net/http"
 	"path"
+	"strings"
 
 	clusterclient "github.com/m3db/m3/src/cluster/client"
 	"github.com/m3db/m3/src/cluster/kv"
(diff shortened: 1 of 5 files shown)


src/query/api/v1/handler/namespace/get.go
```diff
@@ -100,3 +116,19 @@ func (h *GetHandler) Get() (nsproto.Registry, error) {
 
 	return protoRegistry, nil
 }
+
+func nanosToDuration(resp proto.Message) (map[string]interface{}, error) {
+	marshaler := jsonpb.Marshaler{EmitDefaults: true}
```
mway: Is it really necessary to go from PB -> JSON -> map -> JSON? I assume this is the "most convenient way to get an iterable of durations". How painful are these responses to process? I don't necessarily feel the need to over-engineer this, but if this is a pattern we're going to see more of, it'd be better to create something like an `ConvertibleDurationIterator` interface, a concrete, per-message implementation, and then the conversion func would take an iterator and return a map.

Thoughts?

src/query/api/v1/handler/namespace/get.go
```diff
@@ -100,3 +116,19 @@ func (h *GetHandler) Get() (nsproto.Registry, error) {
...
+	nsString, err := marshaler.MarshalToString(resp)
```
mway: Can we just use an `io.ReadWriter` here to skip the intermediate `string`?

src/x/net/http/convert.go
```diff
@@ -38,6 +39,77 @@ var (
 	errDurationType = errors.New("invalid duration type")
 )
 
+// NanosToDurationBytes transforms a json byte slice with Nano keys into Duration keys.
+func NanosToDurationBytes(r io.Reader) (map[string]interface{}, error) {
+	var dict map[string]interface{}
+	d := json.NewDecoder(r)
+	d.UseNumber()
```
mway: You're already handling the `float64` case below, wouldn't it be simpler just to stick with that?

src/x/net/http/convert.go
```diff
@@ -38,6 +39,77 @@ var (
...
+	ret, err := NanosToDurationMap(dict)
+	if err != nil {
+		return nil, err
+	}
+
+	return ret, nil
+}
+
+// NanosToDurationMap transforms keys with Nanos into Duration.
+func NanosToDurationMap(input map[string]interface{}) (map[string]interface{}, error) {
```
mway: Do you need to preserve `input`? If these payloads are large, it might be better to do `ConverToDurationMap` or similar and just add/delete as necessary instead of using 2x the memory.

benraskin92: Nah these are super small. This endpoint just lists out the namespaces in m3db.

src/query/api/v1/handler/namespace/get.go
mway: resolving per other convo; let's revisit if we see a pattern emerging

src/x/net/http/convert.go
benraskin92: So I think we actually want something like this `{"fieldNanos":100.5}` to fail (e.g. nanos shouldn't be floats). I'm fine with removing the `case float64` below though as it will never actually be called.

benraskin92:
> @benraskin92 The namespace API accepts input in the format of durations right? Can we just drop the nanos entirely and make the `debug=true` behavior the default behavior?
> 
> I can't think of a single situation where I'd ever want to interact with raw nanos here

If we do that, I'd rather do that in another PR. I'm all for making it the default though.

schallert:
> > @benraskin92 The namespace API accepts input in the format of durations right? Can we just drop the nanos entirely and make the `debug=true` behavior the default behavior?
> > I can't think of a single situation where I'd ever want to interact with raw nanos here
> 
> If we do that, I'd rather do that in another PR. I'm all for making it the default though.

I don't think we can do that without adding a new endpoint or some careful coordinator with operator compatibility. The operator parses the responses from the coordinator as jsonpb Proto messages, changing the proto or the response marshaling would likely break the operator.

If we really want to make that change we have to do it in a manner such that users of older operator with newer M3DB and vice-versa don't have breakages.

prateek:
+1 to Matt’s point. We should stick with current format for b/w compat. 

More generally, when we add a separate debug endpoint (should do this for a variety of reasons), we can ensure it calls the format that’s easier to grok.

richardartoul:
@schallert @prateek Fair enough. I assumed this only affected the JSON output and that the operator was using GRPC or something but that makes sense

PR #1702, opened 2019-06-04
Add file registry for later use

robskillington:

This adds a file registry for later use by the cleanup code to release resources from the seek manager and also for out of order writes.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/dbnode/persist/fs/registry.go
+++ b/src/dbnode/persist/fs/registry.go
@@ -1,0 +1,504 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


src/dbnode/persist/fs/registry.go
```diff
@@ -0,0 +1,504 @@
...
+				multiErrLock.Lock()
+				multiErr = multiErr.Add(err)
+				multiErrLock.Unlock()
+			}
+		}()
+	}
+
+	wg.Wait()
+
+	if err := multiErr.FinalError(); err != nil {
```
robskillington: Add invariant error to this code path.

robskillington:
Solved already.

PR #1703, opened 2019-06-04
add test case for not setting schema history to empty

Haijuncao:
Add test case to make sure schema history can not be set to empty.

### DIFF
--- a/src/dbnode/storage/database.go
+++ b/src/dbnode/storage/database.go
@@ -236,10 +236,18 @@ func (d *db) updateSchemaRegistry(newNamespaces namespace.Map) error {
 		curSchema, err := schemaReg.GetLatestSchema(metadata.ID())
 		if curSchema != nil {
 			curSchemaId = curSchema.DeployId()
+			if len(curSchemaId) == 0 {
+				d.log.Warn("can not update namespace schema to with empty deploy ID", zap.Stringer("namespace", metadata.ID()),
+					zap.String("current schema id", curSchemaId))
+				merr.Add(fmt.Errorf("can not updating namespace(%v) schema with empty deploy ID", metadata.ID().String()))
+				continue
+			}
 		}
 		// Log schema update.
 		latestSchema, found := metadata.Options().SchemaHistory().GetLatest()
 		if !found {
+			d.log.Warn("can not update namespace schema to empty", zap.Stringer("namespace", metadata.ID()),
+				zap.String("current schema", curSchemaId))
 			merr.Add(fmt.Errorf("can not updating namespace(%v) schema to empty", metadata.ID().String()))
 			continue
 		} else {
--- a/src/dbnode/storage/namespace.go
+++ b/src/dbnode/storage/namespace.go
@@ -373,6 +373,7 @@ func (n *dbNamespace) SetSchemaHistory(value namespace.SchemaHistory) {
 	schema, ok := value.GetLatest()
 	if !ok {
 		n.log.Error("can not update namespace schema to empty", zap.Stringer("namespace", n.ID()))
+		return
 	}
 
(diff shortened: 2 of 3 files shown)


src/dbnode/storage/database.go
```diff
@@ -236,10 +236,18 @@ func (d *db) updateSchemaRegistry(newNamespaces namespace.Map) error {
 		curSchema, err := schemaReg.GetLatestSchema(metadata.ID())
 		if curSchema != nil {
 			curSchemaId = curSchema.DeployId()
+			if len(curSchemaId) == 0 {
+				d.log.Warn("can not update namespace schema to with empty deploy ID", zap.Stringer("namespace", metadata.ID()),
```
richardartoul: "can not update namespace schema with an empty deploy ID"

src/dbnode/storage/database.go
```diff
@@ -236,10 +236,18 @@ func (d *db) updateSchemaRegistry(newNamespaces namespace.Map) error {
...
+					zap.String("current schema id", curSchemaId))
```
richardartoul: I think we don't usually put spaces in key names do we?

src/dbnode/storage/database.go
```diff
@@ -236,10 +236,18 @@ func (d *db) updateSchemaRegistry(newNamespaces namespace.Map) error {
...
+				merr.Add(fmt.Errorf("can not updating namespace(%v) schema with empty deploy ID", metadata.ID().String()))
```
richardartoul: "can not update"

src/dbnode/storage/namespace.go
```diff
@@ -373,6 +373,7 @@ func (n *dbNamespace) SetSchemaHistory(value namespace.SchemaHistory) {
 	schema, ok := value.GetLatest()
 	if !ok {
 		n.log.Error("can not update namespace schema to empty", zap.Stringer("namespace", n.ID()))
+		return
```
richardartoul: Should we add a test for this too? or is it covered by the other test you just added?

src/dbnode/storage/database.go
```diff
@@ -236,10 +236,18 @@ func (d *db) updateSchemaRegistry(newNamespaces namespace.Map) error {
...
+			}
 		}
 		// Log schema update.
 		latestSchema, found := metadata.Options().SchemaHistory().GetLatest()
 		if !found {
+			d.log.Warn("can not update namespace schema to empty", zap.Stringer("namespace", metadata.ID()),
+				zap.String("current schema", curSchemaId))
 			merr.Add(fmt.Errorf("can not updating namespace(%v) schema to empty", metadata.ID().String()))
 			continue
 		} else {
```
richardartoul: Not really related to your change but does this `else` do anything? Since the previous branch continues wonder if we can just drop it and save a level of indentation

src/dbnode/storage/database.go
Haijuncao: why not? I have been doing that. It seem to work fine.

src/dbnode/storage/database.go
Haijuncao: ok

src/dbnode/storage/database.go
Haijuncao: good catch... maybe continue was added later, which render else useless.

src/dbnode/storage/database.go
richardartoul: it works, I think we just usually use camelCase everywhere else. I'd rather it stay consistent

src/dbnode/storage/namespace.go
Haijuncao: yes, it is covered by the test

Haijuncao:
closed as #1722

PR #1704, opened 2019-06-04
Fix existing repair logic and support configurable shadow compare logic

richardartoul:


src/dbnode/storage/repair.go
```diff
@@ -441,3 +480,194 @@ func (r repairerNoOp) Start()        {}
...
+					zap.Error(peerSeriesIter.Err()),
+				)
+				foundMismatch = true
+				break
+			}
+
+			localDP, localUnit, localAnnotation := localSeriesIter.Current()
+			peerDP, peerUnit, peerAnnotation := peerSeriesIter.Current()
+
+			if !localDP.Timestamp.Equal(peerDP.Timestamp) {
```
justinjc: Might be cleaner to make an equality function that wraps these equality checks together. I could see this logic being reused elsewhere too.

src/dbnode/storage/repair.go
```diff
@@ -441,3 +480,194 @@ func (r repairerNoOp) Start()        {}
...
+					zap.Float64("local", localDP.Value),
+					zap.Float64("peer", peerDP.Value),
+				)
+				foundMismatch = true
+				break
+			}
+
+			if localUnit != peerUnit {
+				r.logger.Error(
+					"Values did not match",
```
justinjc: `s/Values/Units`

src/dbnode/storage/repair.go
```diff
@@ -441,3 +480,194 @@ func (r repairerNoOp) Start()        {}
...
+				foundMismatch = true
+				break
+			}
+
+			if !bytes.Equal(localAnnotation, peerAnnotation) {
+				r.logger.Error(
+					"Local message equals peer message, but annotations do not match",
+					zap.Int("index", i),
+					zap.String("local", string(localAnnotation)),
+					zap.String("peer", string(peerAnnotation)),
```
justinjc: Wondering if there's value in logging the `string`s of these annotations. Would it be readable in any way?

src/dbnode/storage/repair.go
```diff
@@ -441,3 +480,194 @@ func (r repairerNoOp) Start()        {}
...
+			r.logger.Error(
+				"Found mismatch between series",
+				zap.String("namespace", nsCtx.ID.String()),
+				zap.Time("start", start),
+				zap.Time("end", end),
+				zap.String("series", seriesID.String()),
+				zap.Int("numDPs", i),
+			)
+		} else {
+			r.logger.Info(
```
justinjc: Do we want to log every single shadow comparison?

src/dbnode/storage/repair/options.go
```diff
@@ -29,54 +29,61 @@ import (
...
+	repairShardConcurrency           int
+	repairInterval                   time.Duration
+	repairTimeOffset                 time.Duration
+	repairTimeJitter                 time.Duration
+	repairCheckInterval              time.Duration
+	repairThrottle                   time.Duration
+	repairMaxRetries                 int
+	hostBlockMetadataSlicePool       HostBlockMetadataSlicePool
+	debugShadowComparisonsEnabled    bool
+	debugShadowComparisonsPercentage float64
```
justinjc: `float32` should be fine?

src/dbnode/storage/repair.go
richardartoul: lol just realized they already have one

src/dbnode/storage/repair.go
richardartoul: might end up being useful for deep debugging, better than having to add it yourself later

src/dbnode/storage/repair.go
richardartoul: Yeah thats what the query code does (also helps you know its working). Its also only executing against a configurable percentage so I think its ok.

 I can change this to `debug` actually. That might make it a little nicer.

src/dbnode/storage/repair/options.go
richardartoul: doesn't hurt to have a more accurate representation

src/dbnode/storage/repair.go
justinjc: Sure.

src/dbnode/storage/repair.go
justinjc: Heh nice. Would it be possible to have an equality func for the `(ts.Datapoint, xtime.Unit, ts.Annotation)` tuple too? That might be useful since `encoding.Iterator` usage is pretty common.

richardartoul: Would be pretty rare to need that I think (this is pretty much the only case) and it would also make the default behavior of the function slow since you'd have to compare two (potentially) large byte slices. Would prefer to leave that out for now unless we have a concrete need for it (also we have better logs this way otherwise I'd have to log the annotation every time)

PR #1706, opened 2019-06-05
Add block leaser for synchronously notifying seeker manager of block volume updates

robskillington:

Very basic implementation, probably requires some tests.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/dbnode/storage/block/lease.go
+++ b/src/dbnode/storage/block/lease.go
@@ -1,0 +1,114 @@
+// Copyright (c) 2019 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package block
+
+import (
+	"errors"
+	"sync"
+)
+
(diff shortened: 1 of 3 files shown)


src/dbnode/storage/block/types.go
```diff
@@ -375,6 +375,55 @@ type FetchBlocksMetadataResultsPool interface {
...
+}
+
+type UpdateOpenLeaseResult uint
+
+const (
+	UpdatedOpenLease UpdateOpenLeaseResult = iota
+	NoOpenLease
+)
+
+type Leaser interface {
```
richardartoul: Maybe a `String()` method so these can be identified in logs if necessary

src/dbnode/storage/block/lease.go
```diff
@@ -0,0 +1,114 @@
...
+	descriptor LeaseDescriptor, 
+	state LeaseState,
+) error {
+	// NB(r): Take exclusive lock so that upgrade leases can't be called
+	// while we are verifying a lease (racey)
+	m.Lock()
+	defer m.Unlock()
+
+	registered := false
+	for _, l := range m.leasers {
```
richardartoul: I wonder if it would be simpler to remove `RegisterLeaser` and just have `OpenLease` register you if you're not already registered since there is no unregister method...although doesn't matter too much.

src/dbnode/storage/block/lease.go
```diff
@@ -0,0 +1,114 @@
...
+		if l == leaser {
```
richardartoul: Do you think this is good enough for determining equality? We've just have to be careful basically to not create a new leaser each time...the alternative would be to have an `Identity() string` method and use that for comparison but I guess if you messed this up it would show up right away in tests

src/dbnode/storage/block/lease.go
```diff
@@ -0,0 +1,114 @@
...
+			return errLeaserAlreadyRegistered
+		}
+	}
+
+	m.leasers = append(m.leasers, leaser)
+
+	return nil
+}
+
+func (m *leaseManager) OpenLease(
```
richardartoul: I'm just now thinking about this but not sure how we're gonna handle the bootstrap case where we call `CacheShardIndices`

That will have to open up a bunch of seekers, but the node is not bootstrapped yet so we won't have the correct state for everything yet so the verifier won't be able to do much....do we just have the verifier skip verification if the node is not bootstrapped yet?

richardartoul: I guess the other alternative would be to pull move the CacheShardIndices call to somewhere after all the shard/namespaces are bootstrapped but before we mark the actual database bootstrapped instead of doing it in the bootstrappers themselves

src/dbnode/storage/block/lease.go
robskillington: I did consider that, the main reason I prefer this method is that at some point there probably needs to be an `UnregisterLeaser(...)` for when closing server or the component for unit tests, etc.

That way the lifecycle becomes more explicit.  Want me to add this method to the interface?

src/dbnode/storage/block/lease.go
robskillington: I've used this a few times, the iface check will make sure it's both same type and same pointer type (since iface is really just struct with the underlying type as an enum like value and the pointer value of the actual receiver).

src/dbnode/storage/block/lease.go
robskillington: As discussed, moving CacheShardIndices to after shards are bootstrapped (but before namespace is bootstrapped IMO, because that's when reads start actually being available) is a good idea.

PR #1707, opened 2019-06-06
Fix multiple blocks with the same block start returned from fetchblocks pathway

### LINKED ISSUE #1640: M3DB client should not reject multiple blocks from a peer with the same blockstart
Right now the `streamBlocksBatchFromPeer`  will reject responses from peers that return more than one block for a given blockStart. I assume this exists for historical reasons, but it makes it so that peer bootstrapping sometimes gets stuck trying to make progress until its peers stop returning multiple blocks (which can happen after a flush when the blocks have not been evicted from memory yet by the tick so they exist in the buffer and on disk or in the near future when we have out of order writes).

It seems to me like we can just delete the following code from `streamBlocksBatchFromPeer` in `session.go`

```golang
// We only ever fetch a single block for a series
		if len(result.Elements[i].Blocks) != 1 {
			errMsg := "stream blocks returned more blocks than expected"
			blocksErr := fmt.Errorf(errMsg+": expected=%d, actual=%d",
				1, len(result.Elements[i].Blocks))
			failed := []receivedBlockMetadata{batch[i]}
			s.reattemptStreamBlocksFromPeersFn(failed, enqueueCh, blocksErr,
				respErrReason, nextRetryReattemptType, m)
			m.fetchBlockError.Inc(int64(len(req.Elements[i].Starts)))
			s.log.Error(errMsg,
				zap.Stringer("id", id),
				zap.Times("expectedStarts", newTimesByUnixNanos(req.Elements[i].Starts)),
				zap.Times("actualStarts", newTimesByRPCBlocks(result.Elements[i].Blocks)),
				zap.Stringer("peer", peer.Host()),
			)
			continue
		}
```

then the merging logic in `addBlockFromPeer` should just work itself out:

```golang
func (r *bulkBlocksResult) addBlockFromPeer(
	id ident.ID,
	encodedTags checked.Bytes,
	peer topology.Host,
	block *rpc.Block,
) error {
	start := time.Unix(0, block.Start)
	result, err := r.newDatabaseBlock(block)
	if err != nil {
		return err
	}

	var (
		tags                ident.Tags
		attemptedDecodeTags bool
	)
	for {
		r.Lock()
		currBlock, exists := r.result.BlockAt(id, start)
		if !exists {
			if encodedTags == nil || attemptedDecodeTags {
				r.result.AddBlock(id, tags, result)
				r.Unlock()
				break
			}
			r.Unlock()

			// Tags not decoded yet, attempt decoded and then reinsert
			attemptedDecodeTags = true
			tagDecoder := r.tagDecoderPool.Get()
			tags, err = newTagsFromEncodedTags(id, encodedTags,
				tagDecoder, r.idPool)
			tagDecoder.Close()
			if err != nil {
				return err
			}
			continue
		}

		// Remove the existing block from the result so it doesn't get
		// merged again
		r.result.RemoveBlockAt(id, start)
		r.Unlock()

		// If we've already received data for this block, merge them
		// with the new block if possible
		tmpCtx := r.contextPool.Get()
		currReader, err := currBlock.Stream(tmpCtx)
		if err != nil {
			return err
		}

		// If there are no data in the current block, there is no
		// need to merge
		if currReader.IsEmpty() {
			continue
		}

		resultReader, err := result.Stream(tmpCtx)
		if err != nil {
			return err
		}
		if resultReader.IsEmpty() {
			return nil
		}

		readers := []xio.SegmentReader{currReader.SegmentReader, resultReader.SegmentReader}
		blockSize := currReader.BlockSize

		encoder, err := r.mergeReaders(start, blockSize, readers)

		if err != nil {
			return err
		}

		result.Close()

		result.Reset(start, blockSize, encoder.Discard(), r.nsCtx)

		tmpCtx.Close()
	}

	return nil
}
```

richardartoul:
Fixes #1640 

@martin-mao This is the bug we were talking about last time I saw you.

### DIFF
--- a/src/dbnode/storage/series/reader.go
+++ b/src/dbnode/storage/series/reader.go
@@ -309,7 +309,6 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
 
 			currBufferResult := bufferResults[bufferIdx]
 			if blockResult.Start.Equal(currBufferResult.Start) {
-				fmt.Println("adding to existing")
 				if currBufferResult.Err != nil {
 					res[i].Err = currBufferResult.Err
 				} else {
@@ -323,7 +322,6 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
 		// Add any buffer results for which there was no existing blockstart
 		// to the end.
 		if bufferIdx < len(bufferResults) {
-			fmt.Println("appending")
 			res = append(res, bufferResults[bufferIdx:]...)
 		}
 	}
--- a/src/dbnode/storage/series/reader.go
+++ b/src/dbnode/storage/series/reader.go
@@ -217,14 +217,20 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
 	nsCtx namespace.Context,
 ) ([]block.FetchBlockResult, error) {
 	var (
-		// Two-dimensional slice such that the first dimension is unique by blockstart
-		// and the second dimension is blocks of data for that blockstart (not necessarily
-		// in chronological order).
+		// Two-dimensional slice (each block.FetchBlockResult has a []xio.BlockReader internally)
+		// such that the first dimension is unique by blockstart and the second dimension is blocks
+		// of data for that blockstart (not necessarily in chronological order).
 		//
 		// ex. (querying 2P.M -> 6P.M with a 2-hour blocksize):
-		// [][]block.FetchBlockResult{
-		//   {Blocks: []xio.BlockReader{block0, block1, block2}}, // <- 2P.M
-		//   {Blocks: []xio.BlockReader{block0}}, // <-4P.M
+		// []block.FetchBlockResult{
+		//   block.FetchBlockResult{
+		//     Start: 2P.M,
+		//     Blocks: []xio.BlockReader{block0, block1, block2},
+		//   },
+		//   block.FetchBlockResult{
+		//     Start: 4P.M,
+		//     Blocks: []xio.BlockReader{block0},
+		//   },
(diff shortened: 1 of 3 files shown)


src/dbnode/storage/series/reader.go
```diff
@@ -208,7 +217,15 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
 	nsCtx namespace.Context,
 ) ([]block.FetchBlockResult, error) {
 	var (
-		// TODO(r): pool these results arrays
+		// Two-dimensional slice such that the first dimension is unique by blockstart
+		// and the second dimension is blocks of data for that blockstart (not necessarily
+		// in chronological order).
```
justinjc: Seems like a copy pasta comment from above that doesn't follow here.

src/dbnode/storage/series/reader.go
```diff
@@ -208,7 +217,15 @@ func (r Reader) fetchBlocksWithBlocksMapAndBuffer(
...
+		//
+		// ex. (querying 2P.M -> 6P.M with a 2-hour blocksize):
+		// [][]block.FetchBlockResult{
```
justinjc: This is actually a 1D slice which is one block start per element.

src/dbnode/storage/series/reader.go
richardartoul: I refactored it to be more clear, take another look

src/dbnode/storage/series/reader.go
richardartoul: Refactored to be more clear, take another look

PR #1714, opened 2019-06-07
Make TestConnectionPoolHealthChecks not flaky

richardartoul:


richardartoul:
@robskillington yeah I thought the same but this was the third flaky test I debugged today and was running out of patience :p

PR #1720, opened 2019-06-10
Support for numbered filesets

justinjc:

With [cold flushes](https://github.com/m3db/m3/pull/1624), we need to support having multiple filesets per block. This PR introduces volume numbers to filesets so that new filesets will be written with an index as part of the filename. In order to facilitate a smooth migration, there is additional logic for gracefully handling legacy (non-indexed) filesets.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/dbnode/persist/fs/files.go
+++ b/src/dbnode/persist/fs/files.go
@@ -418,8 +418,8 @@ func (a fileSetFilesByTimeAndVolumeIndexAscending) Less(i, j int) bool {
 // Returns the positions of filename delimiters ('-' and '.') and the number of
 // delimeters found, to be used in conjunction with the intComponentFromFilename
 // function to extract filename components. This function is deliberately
-// optimized for speed and lack of allocationsm, since filename parsing is known
-// to account for a significant proportion of sytstem resources.
+// optimized for speed and lack of allocations, since filename parsing is known
+// to account for a significant proportion of system resources.
 func delimiterPositions(baseFilename string) ([maxDelimNum]int, int) {
 	var (
 		delimPos    [maxDelimNum]int
@@ -441,9 +441,10 @@ func delimiterPositions(baseFilename string) ([maxDelimNum]int, int) {
 	return delimPos, delimsFound
 }
 
-// Returns the the specified component of a filename given delimeter positions.
-// Our only use cases involve extracting numeric components, so this function
-// assumes this and returns the component as an int64.
+// Returns the the specified component of a filename, given the positions of
+// delimeters. Our only use cases for this involve extracting numeric
+// components, so this function assumes this and returns the component as an
+// int64.
 func intComponentFromFilename(
 	baseFilename string,
 	componentPos int,
--- a/src/dbnode/persist/fs/files.go
+++ b/src/dbnode/persist/fs/files.go
@@ -58,6 +58,10 @@ const (
 	snapshotDirName   = "snapshots"
 	commitLogsDirName = "commitlogs"
 
+	// The maximum number of delimeters ('-' or '.') that is expected in a
+	// (base) filename.
+	maxDelimNum = 4
+
 	// The volume index assigned to (legacy) filesets that don't have a volume
 	// number in their filename.
 	// NOTE: Since this index is the same as the index for the first
@@ -411,30 +415,14 @@ func (a fileSetFilesByTimeAndVolumeIndexAscending) Less(i, j int) bool {
 	return ti.Equal(tj) && ii < ij
 }
 
-func intComponentFromFilename(
-	baseFilename string,
-	componentPos int,
-	separatorIdx [4]int,
-) (int64, error) {
-	var start int
-	if componentPos > 0 {
-		start = separatorIdx[componentPos-1] + 1
-	}
-	end := separatorIdx[componentPos]
-	if start > end || end > len(baseFilename)-1 || start < 0 {
-		return 0, fmt.Errorf(errUnexpectedFilename, baseFilename)
-	}
-
-	num, err := strconv.ParseInt(baseFilename[start:end], 10, 64)
-	if err != nil {
-		return 0, fmt.Errorf(errUnexpectedFilename, baseFilename)
-	}
-	return num, nil
-}
-
-func delimiterPositions(baseFilename string) ([4]int, int) {
+// Returns the positions of filename delimiters ('-' and '.') and the number of
+// delimeters found, to be used in conjunction with the intComponentFromFilename
+// function to extract filename components. This function is deliberately
+// optimized for speed and lack of allocationsm, since filename parsing is known
+// to account for a significant proportion of sytstem resources.
+func delimiterPositions(baseFilename string) ([maxDelimNum]int, int) {
 	var (
-		delimPos    [4]int
+		delimPos    [maxDelimNum]int
 		delimsFound int
 	)
 
@@ -451,6 +439,30 @@ func delimiterPositions(baseFilename string) ([4]int, int) {
 	}
 
 	return delimPos, delimsFound
+}
+
+// Returns the the specified component of a filename given delimeter positions.
+// Our only use cases involve extracting numeric components, so this function
+// assumes this and returns the component as an int64.
+func intComponentFromFilename(
+	baseFilename string,
+	componentPos int,
+	delimPos [maxDelimNum]int,
+) (int64, error) {
+	var start int
+	if componentPos > 0 {
+		start = delimPos[componentPos-1] + 1
+	}
+	end := delimPos[componentPos]
+	if start > end || end > len(baseFilename)-1 || start < 0 {
+		return 0, fmt.Errorf(errUnexpectedFilename, baseFilename)
+	}
+
+	num, err := strconv.ParseInt(baseFilename[start:end], 10, 64)
+	if err != nil {
+		return 0, fmt.Errorf(errUnexpectedFilename, baseFilename)
+	}
+	return num, nil
 }
 
 // TimeFromFileName extracts the block start time from file name.
--- a/src/dbnode/persist/fs/files.go
+++ b/src/dbnode/persist/fs/files.go
@@ -72,13 +72,14 @@ const (
 	timeComponentPosition         = 1
 	commitLogComponentPosition    = 2
 	indexFileSetComponentPosition = 2
+	dataFileSetComponentPosition  = 2
 
 	numComponentsSnapshotMetadataFile           = 4
 	numComponentsSnapshotMetadataCheckpointFile = 5
 	snapshotMetadataUUIDComponentPosition       = 1
 	snapshotMetadataIndexComponentPosition      = 2
 
-	errUnexpectedFilename = "unexpected filename: %s"
+	errUnexpectedFilenamePattern = "unexpected filename: %s"
 )
 
 var (
@@ -416,10 +417,11 @@ func (a fileSetFilesByTimeAndVolumeIndexAscending) Less(i, j int) bool {
 }
 
 // Returns the positions of filename delimiters ('-' and '.') and the number of
-// delimeters found, to be used in conjunction with the intComponentFromFilename
+// delimeters found, to be used in conjunction with the intComponentAtIndex
 // function to extract filename components. This function is deliberately
-// optimized for speed and lack of allocations, since filename parsing is known
-// to account for a significant proportion of system resources.
+// optimized for speed and lack of allocations, since allocation-heavy filename
+// parsing can quickly become a large source of allocations in the entire
+// system, especially when namespaces with long retentions are configured.
 func delimiterPositions(baseFilename string) ([maxDelimNum]int, int) {
 	var (
 		delimPos    [maxDelimNum]int
@@ -445,23 +447,23 @@ func delimiterPositions(baseFilename string) ([maxDelimNum]int, int) {
 // delimeters. Our only use cases for this involve extracting numeric
 // components, so this function assumes this and returns the component as an
 // int64.
-func intComponentFromFilename(
+func intComponentAtIndex(
 	baseFilename string,
 	componentPos int,
 	delimPos [maxDelimNum]int,
 ) (int64, error) {
-	var start int
+	start := 0
 	if componentPos > 0 {
 		start = delimPos[componentPos-1] + 1
 	}
 	end := delimPos[componentPos]
 	if start > end || end > len(baseFilename)-1 || start < 0 {
-		return 0, fmt.Errorf(errUnexpectedFilename, baseFilename)
+		return 0, fmt.Errorf(errUnexpectedFilenamePattern, baseFilename)
 	}
 
 	num, err := strconv.ParseInt(baseFilename[start:end], 10, 64)
 	if err != nil {
-		return 0, fmt.Errorf(errUnexpectedFilename, baseFilename)
+		return 0, fmt.Errorf(errUnexpectedFilenamePattern, baseFilename)
 	}
 	return num, nil
 }
@@ -471,12 +473,15 @@ func TimeFromFileName(fname string) (time.Time, error) {
 	base := filepath.Base(fname)
 
 	delims, delimsFound := delimiterPositions(base)
-	if delimsFound < 2 {
-		return timeZero, fmt.Errorf(errUnexpectedFilename, fname)
+	// There technically only needs to be two delimeters here since the time
+	// component is in index 1. However, all DB files have a minimum of three
+	// delimeters, so check for that instead.
+	if delimsFound < 3 {
+		return timeZero, fmt.Errorf(errUnexpectedFilenamePattern, fname)
 	}
-	nanos, err := intComponentFromFilename(base, timeComponentPosition, delims)
+	nanos, err := intComponentAtIndex(base, timeComponentPosition, delims)
 	if err != nil {
-		return timeZero, fmt.Errorf(errUnexpectedFilename, fname)
+		return timeZero, fmt.Errorf(errUnexpectedFilenamePattern, fname)
 	}
 
 	return time.Unix(0, nanos), nil
@@ -497,12 +502,12 @@ func TimeAndVolumeIndexFromDataFileSetFilename(fname string) (time.Time, int, er
 
 	delims, delimsFound := delimiterPositions(base)
 	if delimsFound < 3 {
-		return timeZero, 0, fmt.Errorf(errUnexpectedFilename, fname)
+		return timeZero, 0, fmt.Errorf(errUnexpectedFilenamePattern, fname)
 	}
 
-	nanos, err := intComponentFromFilename(base, timeComponentPosition, delims)
+	nanos, err := intComponentAtIndex(base, timeComponentPosition, delims)
 	if err != nil {
-		return timeZero, 0, fmt.Errorf(errUnexpectedFilename, fname)
+		return timeZero, 0, fmt.Errorf(errUnexpectedFilenamePattern, fname)
 	}
 	unixNanos := time.Unix(0, nanos)
 
@@ -511,9 +516,9 @@ func TimeAndVolumeIndexFromDataFileSetFilename(fname string) (time.Time, int, er
 		return unixNanos, unindexedFilesetIndex, nil
 	}
 
-	volume, err := intComponentFromFilename(base, 2, delims)
+	volume, err := intComponentAtIndex(base, dataFileSetComponentPosition, delims)
 	if err != nil {
-		return timeZero, 0, fmt.Errorf(errUnexpectedFilename, fname)
+		return timeZero, 0, fmt.Errorf(errUnexpectedFilenamePattern, fname)
 	}
 
 	return unixNanos, int(volume), nil
@@ -530,18 +535,18 @@ func timeAndIndexFromFileName(fname string, componentPosition int) (time.Time, i
 
 	delims, delimsFound := delimiterPositions(base)
 	if componentPosition > delimsFound {
-		return timeZero, 0, fmt.Errorf(errUnexpectedFilename, fname)
+		return timeZero, 0, fmt.Errorf(errUnexpectedFilenamePattern, fname)
 	}
 
-	nanos, err := intComponentFromFilename(base, 1, delims)
+	nanos, err := intComponentAtIndex(base, 1, delims)
 	if err != nil {
-		return timeZero, 0, fmt.Errorf(errUnexpectedFilename, fname)
+		return timeZero, 0, fmt.Errorf(errUnexpectedFilenamePattern, fname)
 	}
 	unixNanos := time.Unix(0, nanos)
 
-	index, err := intComponentFromFilename(base, componentPosition, delims)
+	index, err := intComponentAtIndex(base, componentPosition, delims)
 	if err != nil {
-		return timeZero, 0, fmt.Errorf(errUnexpectedFilename, fname)
+		return timeZero, 0, fmt.Errorf(errUnexpectedFilenamePattern, fname)
 	}
 
 	return unixNanos, int(index), nil
@@ -1375,8 +1380,12 @@ func DataFileSetExists(
 	blockStart time.Time,
 	volume int,
 ) (bool, error) {
-	// This function is in some cases the bottleneck, thus, we use different
-	// logic to handle this case specifically (don't use DataFiles(...)).
+	// This function can easily become a performance bottleneck if the
+	// implementation is slow or requires scanning directories with a large
+	// number of files in them (as is common if namespaces with long retentions
+	// are configured). As a result, instead of using existing helper functions,
+	// it implements an optimized code path that only involves checking if a few
+	// specific files exist and contain the correct contents.
 	shardDir := ShardDataDirPath(filePathPrefix, namespace, shard)
 
 	// Check fileset with volume first to optimize for non-legacy use case.
@@ -1402,6 +1411,8 @@ func SnapshotFileSetExistsAt(prefix string, namespace ident.ID, shard uint32, bl
 		return false, nil
 	}
 
+	// LatestVolumeForBlock checks for a complete checkpoint file, so we don't
+	// need to recheck it here.
 	return true, nil
 }
 
@@ -1557,10 +1568,10 @@ func filesetPathFromTimeAndIndex(prefix string, t time.Time, index int, suffix s
 }
 
 // isFirstVolumeLegacy returns whether the first volume of the provided type is
-// legacy, i.e. does not have a volume index in its filename. It only checks for
-// the existence of the file and the non-legacy version of the file, and returns
-// an error if neither exist. Note that this function does not check for the
-// volume's complete checkpoint file.
+// legacy, i.e. does not have a volume index in its filename. Using this
+// function, the caller expects there to be a legacy or non-legacy file, and
+// thus returns an error if neither exist. Note that this function does not
+// check for the volume's complete checkpoint file.
 func isFirstVolumeLegacy(prefix string, t time.Time, suffix string) (bool, error) {
 	// Check non-legacy path first to optimize for newer files.
 	path := filesetPathFromTimeAndIndex(prefix, t, 0, suffix)
--- a/src/dbnode/persist/fs/files.go
+++ b/src/dbnode/persist/fs/files.go
@@ -181,7 +181,7 @@ func (f FileSetFilesSlice) Filepaths() []string {
 }
 
 // LatestVolumeForBlock returns the latest (highest index) FileSetFile in the
-// slice for a given block start.
+// slice for a given block start that has a complete checkpoint file.
 func (f FileSetFilesSlice) LatestVolumeForBlock(blockStart time.Time) (FileSetFile, bool) {
 	// Make sure we're already sorted.
 	f.sortByTimeAndVolumeIndexAscending()
@@ -445,7 +445,7 @@ func TimeAndVolumeIndexFromDataFileSetFilename(fname string) (time.Time, int, er
 
 	if len(components) == 3 {
 		// Three components mean that there is no index in the filename.
-		return t, unindexedFilesetIndex, err
+		return t, unindexedFilesetIndex, nil
 	}
 
 	if len(components) == 4 {
@@ -458,7 +458,7 @@ func TimeAndVolumeIndexFromDataFileSetFilename(fname string) (time.Time, int, er
 		return t, int(index), nil
 	}
 
-	return timeZero, 0, fmt.Errorf("malformed filename: %s", fname)
+	return timeZero, 0, fmt.Errorf("malformed data fileset filename: %s", fname)
 }
 
 // TimeAndVolumeIndexFromFileSetFilename extracts the block start and
@@ -633,9 +633,12 @@ func forEachInfoFile(
 		t := matched[i].ID.BlockStart
 		volume := matched[i].ID.VolumeIndex
 
-		isLegacy, err := isFirstVolumeLegacy(dir, t, checkpointFileSuffix)
-		if err != nil {
-			continue
+		isLegacy := false
+		if volume == 0 {
+			isLegacy, err = isFirstVolumeLegacy(dir, t, checkpointFileSuffix)
+			if err != nil {
+				continue
+			}
 		}
 		var (
 			checkpointFilePath string
@@ -1292,21 +1295,6 @@ func CommitLogsDirPath(prefix string) string {
 	return path.Join(prefix, commitLogsDirName)
 }
 
-// DataFileSetExistsAt determines whether data fileset files exist for the given namespace, shard, and block start.
-func DataFileSetExistsAt(filePathPrefix string, namespace ident.ID, shard uint32, blockStart time.Time) (bool, error) {
-	dataFiles, err := DataFiles(filePathPrefix, namespace, shard)
-	if err != nil {
-		return false, err
-	}
-
-	_, ok := dataFiles.LatestVolumeForBlock(blockStart)
-	if !ok {
-		return false, nil
-	}
-
-	return true, nil
-}
-
 // DataFileSetExists determines whether data fileset files exist for the given
 // namespace, shard, block start, and volume.
 func DataFileSetExists(
@@ -1316,12 +1304,16 @@ func DataFileSetExists(
 	blockStart time.Time,
 	volume int,
 ) (bool, error) {
-	dataFiles, err := DataFiles(filePathPrefix, namespace, shard)
-	if err != nil {
-		return false, err
+	shardDir := ShardDataDirPath(filePathPrefix, namespace, shard)
+	// Check fileset with volume first to optimize for non-legacy use case.
+	checkpointPath := filesetPathFromTimeAndIndex(shardDir, blockStart, volume, checkpointFileSuffix)
+	if volume != 0 {
+		return CompleteCheckpointFileExists(checkpointPath)
 	}
 
-	return dataFiles.VolumeExistsForBlock(blockStart, volume), nil
+	// Check legacy format next only if volume is 0.
+	checkpointPath = filesetPathFromTimeLegacy(shardDir, blockStart, checkpointFileSuffix)
+	return CompleteCheckpointFileExists(checkpointPath)
 }
 
 // SnapshotFileSetExistsAt determines whether snapshot fileset files exist for the given namespace, shard, and block start time.
@@ -1481,7 +1473,7 @@ func filesetFileForTime(t time.Time, suffix string) string {
 	return fmt.Sprintf("%s%s%d%s%s%s", filesetFilePrefix, separator, t.UnixNano(), separator, suffix, fileSuffix)
 }
 
-func filesetPathFromTime(prefix string, t time.Time, suffix string) string {
+func filesetPathFromTimeLegacy(prefix string, t time.Time, suffix string) string {
 	return path.Join(prefix, filesetFileForTime(t, suffix))
 }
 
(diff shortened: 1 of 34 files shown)


src/dbnode/persist/fs/retriever.go
```diff
@@ -267,7 +267,9 @@ func (r *blockRetriever) fetchBatch(
 	seekerResources ReusableSeekerResources,
 ) {
 	// Resolve the seeker from the seeker mgr
-	seeker, err := seekerMgr.Borrow(shard, blockStart)
+	// TODO(juchan): actually get the correct volume.
+	vol := 0
+	seeker, err := seekerMgr.Borrow(shard, blockStart, vol)
```
richardartoul: I don't think this interface should expose the volume. The Borrow() API should just return a seeker for whatever volume the SeekerManager is currently holding a reference to.

src/dbnode/persist/fs/retriever.go
```diff
@@ -404,7 +407,9 @@ func (r *blockRetriever) Stream(
 	}
 	r.RUnlock()
 
-	bloomFilter, err := r.seekerMgr.ConcurrentIDBloomFilter(shard, startTime)
+	// TODO(juchan): actually get the correct volume.
+	vol := 0
+	bloomFilter, err := r.seekerMgr.ConcurrentIDBloomFilter(shard, startTime, vol)
```
richardartoul: same thing

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -108,11 +111,35 @@ type borrowableSeeker struct {
 	isBorrowed bool
 }
 
-type seekersByTime struct {
+type seekersByTimeAndVolume struct {
 	sync.RWMutex
 	shard    uint32
 	accessed bool
-	seekers  map[xtime.UnixNano]seekersAndBloom
+	seekers  map[xtime.UnixNano]map[int]seekersAndBloom
```
richardartoul: I dont think volume should be parse of the map. There should only been seekers/bloom for one volume at a time (you may need to keep track of which volume you have but you shouldn't have multiple at once)

src/dbnode/storage/shard.go
```diff
@@ -1687,8 +1687,10 @@ func (s *dbShard) FetchBlocksMetadataV2(
 			pos.metadataIdx = int(flushedPhase.CurrBlockEntryIdx)
 		}
 
+		// TODO(juchan): actually get the volume
+		vol := 0
 		// Open a reader at this position, potentially from cache
-		reader, err := s.namespaceReaderMgr.get(s.shard, blockStart, pos)
+		reader, err := s.namespaceReaderMgr.get(s.shard, blockStart, vol, pos)
```
richardartoul: hmm....we may need to make this thing listen for lease changes...I'll dig into it more in my P.R

src/dbnode/storage/shard.go
```diff
@@ -1890,6 +1898,9 @@ func (s *dbShard) WarmFlush(
 		NamespaceMetadata: s.namespace,
 		Shard:             s.ID(),
 		BlockStart:        blockStart,
+		// Volume index is always 0 for warm flushes because a warm flush must
+		// happen first before cold flushes happen.
+		VolumeIndex: 0,
 		// We explicitly set delete if exists to false here as we track which
```
richardartoul: Maybe update the comment above `DeleteIfExists` as well

src/dbnode/storage/shard.go
```diff
@@ -1992,10 +2003,12 @@ func (s *dbShard) ColdFlush(
...
 		startTime := blockStart.ToTime()
+		coldVersion := s.RetrievableBlockColdVersion(startTime)
 		fsID := fs.FileSetFileIdentifier{
-			Namespace:  s.namespace.ID(),
-			Shard:      s.ID(),
-			BlockStart: startTime,
+			Namespace:   s.namespace.ID(),
+			Shard:       s.ID(),
+			BlockStart:  startTime,
+			VolumeIndex: coldVersion,
```
richardartoul: The call to `Merge` will error out if our logic is wrong and the volume does not exist right? (which is what we want it to do)

src/dbnode/storage/shard.go
```diff
@@ -2144,18 +2157,36 @@ func (s *dbShard) removeAnyFlushStatesTooEarly(tickStart time.Time) {
...
+		return fmt.Errorf("encountered errors when getting fileset files for prefix %s namespace %s shard %d: %v",
+			filePathPrefix, s.namespace.ID(), s.ID(), err)
 	}
-	if err := s.deleteFilesFn(expired); err != nil {
-		multiErr = multiErr.Add(err)
+
+	return s.deleteFilesFn(expired)
+}
+
+func (s *dbShard) CleanupCompactedFileSets() error {
```
richardartoul: like I was saying before, probably want this in `cleanup.go`. Not necessary in this P.R if we dont want to

src/dbnode/storage/shard.go
justinjc: Filed an issue to track: https://github.com/m3db/m3/issues/1732

I'll remove the volume arg for now.

src/dbnode/persist/fs/files.go
```diff
@@ -1380,7 +1486,43 @@ func filesetPathFromTime(prefix string, t time.Time, suffix string) string {
...
+	return path.Join(prefix, filesetFileForTime(t, newSuffix))
+}
+
+// isFirstVolumeLegacy returns whether the first volume of the provided type is
+// legacy, i.e. does not have a volume index in its filename. It only checks for
+// the existence of the file and the non-legacy version of the file, and returns
+// and error if neither exist.
+func isFirstVolumeLegacy(prefix string, t time.Time, suffix string) (bool, error) {
+	path := filesetPathFromTimeAndIndex(prefix, t, 0, suffix)
+	_, err := os.Stat(path)
```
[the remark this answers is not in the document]
justinjc: Yeah I saw that, but I'd argue that we don't really care about the other `err` cases. Either one of the types of files exist or neither of them do, in which case that's the error we propagate down.

src/dbnode/persist/fs/files.go
```diff
@@ -1225,12 +1331,12 @@ func SnapshotFileSetExistsAt(prefix string, namespace ident.ID, shard uint32, bl
...
 	}
 
-	latest, ok := snapshotFiles.LatestVolumeForBlock(blockStart)
+	_, ok := snapshotFiles.LatestVolumeForBlock(blockStart)
 	if !ok {
 		return false, nil
 	}
 
-	return latest.HasCompleteCheckpointFile(), nil
+	return true, nil
```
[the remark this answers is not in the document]
justinjc: Correct. I'll put in a comment.

src/dbnode/persist/fs/files.go
```diff
@@ -1380,7 +1486,43 @@ func filesetPathFromTime(prefix string, t time.Time, suffix string) string {
...
+	newSuffix := fmt.Sprintf("%d%s%s", index, separator, suffix)
+	return path.Join(prefix, filesetFileForTime(t, newSuffix))
+}
+
+// isFirstVolumeLegacy returns whether the first volume of the provided type is
+// legacy, i.e. does not have a volume index in its filename. It only checks for
+// the existence of the file and the non-legacy version of the file, and returns
+// and error if neither exist.
+func isFirstVolumeLegacy(prefix string, t time.Time, suffix string) (bool, error) {
+	path := filesetPathFromTimeAndIndex(prefix, t, 0, suffix)
```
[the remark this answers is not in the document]
justinjc: Sure.

src/dbnode/persist/fs/types.go
```diff
@@ -51,7 +51,7 @@ type FileSetFileIdentifier struct {
 	BlockStart         time.Time
 	// Only required for data content files
 	Shard uint32
-	// Required for snapshot files (index yes, data yes) and flush files (index yes, data no)
+	// Required for snapshot files (index yes, data yes) and flush files (index yes, data yes)
```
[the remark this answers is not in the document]
justinjc: I'm a thorough man.

src/dbnode/storage/shard.go
justinjc: Yep, it will fail to open the file and error out.

src/dbnode/persist/fs/files.go
```diff
@@ -1213,9 +1294,34 @@ func CommitLogsDirPath(prefix string) string {
...
+// DataFileSetExists determines whether data fileset files exist for the given
+// namespace, shard, block start, and volume.
+func DataFileSetExists(
+	filePathPrefix string,
+	namespace ident.ID,
+	shard uint32,
+	blockStart time.Time,
+	volume int,
+) (bool, error) {
+	dataFiles, err := DataFiles(filePathPrefix, namespace, shard)
```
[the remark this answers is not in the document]
justinjc: Okay sure, wasn't aware of the bottleneck. I'll include a comment for that too.

src/dbnode/persist/fs/files.go
```diff
@@ -1213,9 +1294,34 @@ func CommitLogsDirPath(prefix string) string {
 
 // DataFileSetExistsAt determines whether data fileset files exist for the given namespace, shard, and block start.
 func DataFileSetExistsAt(filePathPrefix string, namespace ident.ID, shard uint32, blockStart time.Time) (bool, error) {
```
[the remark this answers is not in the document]
justinjc: Yeah, agreed. Looks like the seeker manager and the namespace readers are the only two users of this function and they should both use the volume-specific version. I think I'll just remove this function.

src/dbnode/persist/fs/files.go
```diff
@@ -556,9 +642,16 @@ func forEachInfoFile(
 		case persist.FileSetFlushType:
 			switch args.contentType {
 			case persist.FileSetDataContentType:
-				checkpointFilePath = filesetPathFromTime(dir, t, checkpointFileSuffix)
-				digestsFilePath = filesetPathFromTime(dir, t, digestFileSuffix)
-				infoFilePath = filesetPathFromTime(dir, t, infoFileSuffix)
+				isLegacy := false
```
robskillington: We should perhaps leave some comments somewhere in this package how we could eventually remove these legacy pieces of logic if we ever cut a 1.0 release and can perform backwards incompatible changes.

Just like a few sentences on the thinking about how to remove it and which string to "grep" for in the code to find all the places we can delete dead code once we don't need to support backwards compat for it.

src/dbnode/persist/fs/files.go
```diff
@@ -832,18 +925,9 @@ func FileSetAt(filePathPrefix string, namespace ident.ID, shard uint32, blockSta
...
-				// Should never happen
-				return FileSetFile{}, false, fmt.Errorf(
-					"found multiple fileset files for blockStart: %d",
-					blockStart.Unix(),
-				)
-			}
-
+	matched.sortByTimeAndVolumeIndexAscending()
+	for _, fileset := range matched {
+		if fileset.ID.BlockStart.Equal(blockStart) && fileset.ID.VolumeIndex == volume {
```
robskillington: We removed the check for seeing if the fileset that comes after has the same block start + volume, do we want to keep it perhaps?

src/dbnode/persist/fs/files.go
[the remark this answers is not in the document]
robskillington: Can you add the comment that LatestVolumeForBlock does the check perhaps?

src/dbnode/persist/fs/files.go
```diff
@@ -1375,12 +1476,48 @@ func filesetFileForTime(t time.Time, suffix string) string {
...
 func filesetPathFromTimeAndIndex(prefix string, t time.Time, index int, suffix string) string {
-	return path.Join(prefix, filesetFileForTime(t, fmt.Sprintf("%d%s%s", index, separator, suffix)))
+	newSuffix := fmt.Sprintf("%d%s%s", index, separator, suffix)
+	return path.Join(prefix, filesetFileForTime(t, newSuffix))
+}
+
+// isFirstVolumeLegacy returns whether the first volume of the provided type is
+// legacy, i.e. does not have a volume index in its filename. It only checks for
+// the existence of the file and the non-legacy version of the file, and returns
+// and error if neither exist.
```
robskillington: nit "and returns *an* error if neither exist"

src/dbnode/persist/fs/files.go
```diff
@@ -1375,12 +1476,48 @@ func filesetFileForTime(t time.Time, suffix string) string {
...
+	path := filesetPathFromTimeAndIndex(prefix, t, 0, suffix)
+	_, err := os.Stat(path)
+	if err == nil {
+		return false, nil
+	}
+
+	legacyPath := filesetPathFromTimeLegacy(prefix, t, suffix)
+	_, err = os.Stat(legacyPath)
+	if err == nil {
+		return true, nil
```
robskillington: Do you also need to check/verify for a completed checkpoint file?

src/dbnode/persist/fs/merger.go
```diff
@@ -128,6 +130,8 @@ func (m *merger) Merge(
 		NamespaceMetadata: nsMd,
 		Shard:             shard,
 		BlockStart:        startTime,
+		VolumeIndex:       fileID.VolumeIndex + 1,
+		FileSetType:       persist.FileSetFlushType,
 		DeleteIfExists:    false,
 	}
 	prepared, err := flushPreparer.PrepareData(prepareOpts)
```
robskillington: Hm, I can't seem to make a comment here - but just realized on line 195:
```
idsToFinalize = append(idsToFinalize, id)
if err != nil {
  return err
}
```

The appending of `idsToFinalize` needs to happen after the error check, otherwise you'll append `nil` to the slice, and then when the defer tries to finalize each ID it will panic on nil pointer for `id.Finalize()`.

robskillington: Similarly `tagsToFinalize = append(tagsToFinalize, tags)` should come after the error check.

src/dbnode/persist/fs/persist_manager.go
```diff
@@ -598,20 +603,25 @@ func (pm *persistManager) doneShared() error {
...
 	switch prepareOpts.FileSetType {
 	case persist.FileSetSnapshotType:
-		// Snapshot files are indexed (multiple per block-start), so checking if the file
-		// already exist doesn't make much sense
+		// Checking if a snapshot file exists for a block start doesn't make
+		// sense in this context because the logic for creating new snapshot
+		// files does not use the volume index provided in the prepareOpts.
+		// Instead, the new volume index is determined by looking at what files
+		// exist on disk. This means that there can never be a conflict when
+		// trying to write new snapshot files.
```
robskillington: Kind of strange, but why does snapshots use one mechanism (i.e. increment new volume index by looking at files on disk) vs fileset volumes using a different one?  Richie alluded that it was "hard" but would be interesting to hear what was the difficulty with just making them use similar/same code paths?

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -376,7 +376,9 @@ func (m *seekerManager) newOpenSeeker(
 	shard uint32,
 	blockStart time.Time,
 ) (DataFileSetSeeker, error) {
-	exists, err := DataFileSetExistsAt(m.filePathPrefix, m.namespace, shard, blockStart)
+	// TODO(juchan): get the actual volume here.
```
robskillington: Do we need to do this for this PR merges or are we tracking this required change in an issue/elsewhere where it won't get forgotten?

src/dbnode/persist/types.go
```diff
@@ -136,19 +136,16 @@ type DataPrepareOptions struct {
 	NamespaceMetadata namespace.Metadata
 	BlockStart        time.Time
 	Shard             uint32
-	FileSetType       FileSetType
-	DeleteIfExists    bool
+	// This volume index is only used when preparing for a flush fileset type.
+	// When opening a snapshot, the new volume index is determined by looking
+	// at what files exist on disk.
```
robskillington: Again, does seem strange they don't just both do the same strategy.

src/dbnode/storage/bootstrap/bootstrapper/peers/source.go
```diff
@@ -396,6 +396,14 @@ func (s *peersSource) flush(
 			FileSetType:       persistConfig.FileSetType,
 			Shard:             shard,
 			BlockStart:        start,
+			// When bootstrapping, the volume index will always be 0. However,
+			// if we want to be able to snapshot and flush while bootstrapping,
+			// this may not be the case, e.g. if a flush occurs before a
+			// bootstrap, then the bootstrap volume index will be >0. In order
+			// to support this, bootstrapping code will need to incorporate
+			// merging logic and flush version/volume index will need to be
+			// synchronized between processes.
```
robskillington: Just a thought: I assume if we ever want to do repairs we might also need to support volume index >0 for bootstrapping.

src/dbnode/storage/namespace_readers.go
```diff
@@ -149,23 +151,25 @@ func newNamespaceReaderManager(
...
 	}
 }
 
 func (m *namespaceReaderManager) filesetExistsAt(
 	shard uint32,
 	blockStart time.Time,
 ) (bool, error) {
-	return m.filesetExistsAtFn(m.fsOpts.FilePathPrefix(),
-		m.namespace.ID(), shard, blockStart)
+	// TODO(juchan): get the actual volume here.
```
robskillington: Just to make sure, this is being tracked in an issue yeah?

robskillington: Also, might it make more sense for the reader to just figure out the latest complete and open that instead of taking volume everywhere?

src/dbnode/storage/namespace_readers.go
```diff
@@ -251,11 +255,14 @@ func (m *namespaceReaderManager) get(
 	// We have a closed reader from the cache (either a cached closed
 	// reader or newly allocated, either way need to prepare it)
 	reader := lookup.closedReader
+	// TODO(juchan): get the actual volume here.
```
robskillington: You'll have to make the volume index be part of the `readerPosition` too and make that part of the page token perhaps returned to clients.

src/dbnode/persist/fs/files.go
richardartoul: 👍

src/dbnode/persist/fs/persist_manager.go
richardartoul: Yeah I kind of did a poor job explaining this. There are a couple of subtle complexities that we encountered but basically we realized that we need to maintain some kind of relationship between the flush version and the volume number for a variety of reasons but the one that immediately comes to mind is imagine the `SeekerManager` does a lot of checks to see if specific fileset files exist. In order to do that efficiently it needs to be able to go from namespace/shard/blockStart/volume -> expected filename.

Once we decided that we wanted to maintain that relationship the "simplest" thing is to ensure that the flush version and the volume number are always the same. Technically we could do the `+1` logic and still maintain this guarantee but that would make it implicit instead of explicit and a brittle assumption.

So we decided to just tell the `PersistManager` / writer which volume we intended to create and that has some nice properties like if our logic is wrong the `PersistManager` can error out and tell us that the volume already exists (as opposed to it just incrementing by 1 and subtly breaking our assumptions).

Sine we took this route we do need to bootstrap the flush version back into the shard, but thats a trivial change since the shard already calls `ForEachInfoFile` so all we need to do is pull the volume number out of the info metadata and into the shard memory state.

@justinjc Do you remember any other details? We spent a few hours working through all this and I feel like I may be missing some of the other reasons we couldn't do the incrementing.

@robskillington Thinking about this now though it seems like it might be good to add a check in the persist manager that the volume being writeen is at least as high as the highest existing complete fileset file. I.E if complete versions of 1, 2, 3 exist then writing 3 (if `deleteIfExist=true`) and 4 should be allowed, but writing 2 should not. What do you think @justinjc ? Could also add it to your list of follow ups

src/dbnode/persist/types.go
richardartoul: Yeah agreed its a little weird, and Justin's initial implementation did do the +1 strategy but we ran into some shard edges with that approach that pushed us towards the other direction

src/dbnode/storage/bootstrap/bootstrapper/peers/source.go
richardartoul: @robskillington The designed repair feature won't use the peers bootstrapper at all right? Re: incremental bootstrap, what I've been thinking about recently is to eventually move away from having incremental in the peers boostrapper and instead combine cold flushing + streaming bootstrap to get the same effect.

The alternative approach (where we support index > 0) and then basically do merging from here would work also, but I think it would make allowing M3DB to flush while its bootstrapping really difficult (which is a place I think we want to get to long term) because their would have to be coordination of the flush versions there.

src/dbnode/persist/fs/persist_manager.go
richardartoul: We could also support a flag in the `PersistManager` that basically tells it to figure out the volume number itself which we would probably need if we wanted to allow incremental peer bootstrapping and flushing to occur in parallel, but not sure we want to go with that approach long term (see my other comment below)

src/dbnode/persist/fs/seek_manager.go
richardartoul: Not needed for merging (since with cold writes disabled the volume number can only ever be 0).

My other P.R that you reviewed will address this issue and allow Justin to rebase off of that or justin can just file an issue and do in a quick follow up. @justinjc

src/dbnode/storage/namespace_readers.go
richardartoul: https://github.com/m3db/m3/issues/1732

src/dbnode/storage/namespace_readers.go
richardartoul: Yeah I think thats the right approach although actual implementation might be slightly more complicated:

```
1. Return `readPosition` that includes volume number
2. RPC comes in with `readPosition`. If volume still exists, continue reading. If volume does not exist, start from the beginning at the next volume
```

or something like that. The peers bootstrapper is already designed to handle duplicate metadata so this should be fine

src/dbnode/persist/fs/files.go
justinjc: Sure.

src/dbnode/persist/fs/files.go
justinjc: This is a good point. This function should be used solely to figure out whether the first volume is legacy or not before we actually use them, so checking for a complete checkpoint file should come after this function is called. I'll add a comment.

src/dbnode/persist/fs/merger.go
justinjc: Good catch 👍

src/dbnode/persist/fs/seek_manager.go
justinjc: I've opened https://github.com/m3db/m3/issues/1747 to track.

src/dbnode/storage/cleanup.go
```diff
@@ -240,6 +240,7 @@ func (m *cleanupManager) cleanupExpiredDataFiles(t time.Time) error {
 		earliestToRetain := retention.FlushTimeStart(n.Options().RetentionOptions(), t)
 		shards := n.GetOwnedShards()
 		multiErr = multiErr.Add(m.cleanupExpiredNamespaceDataFiles(earliestToRetain, shards))
+		multiErr = multiErr.Add(m.cleanupCompactedNamespaceDataFiles(shards))
```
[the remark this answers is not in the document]
justinjc: Will track this in https://github.com/m3db/m3/issues/1748.

src/dbnode/storage/shard.go
justinjc: Will track this in https://github.com/m3db/m3/issues/1748.

src/dbnode/persist/fs/files.go
```diff
@@ -392,12 +427,42 @@ func TimeFromFileName(fname string) (time.Time, error) {
...
+	}
+
+	if len(components) == 3 {
+		// Three components mean that there is no index in the filename.
+		return t, unindexedFilesetIndex, err
+	}
+
+	if len(components) == 4 {
+		// Four components mean that there is an index in the filename.
+		str := strings.Replace(components[indexFileSetComponentPosition], fileSuffix, "", 1)
```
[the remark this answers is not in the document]
justinjc: I've rewritten filename parsing completely, removing `componentsAndTimeFromFileName`, making it three times as fast and zero allocs (according to microbenchmarks). The implementation is definitely more specific to today's filename structure, but it's probably worth it.

src/dbnode/storage/namespace_readers.go
justinjc: Tracking via https://github.com/m3db/m3/issues/1747.

src/dbnode/persist/fs/files.go
```diff
@@ -373,51 +415,136 @@ func (a fileSetFilesByTimeAndVolumeIndexAscending) Less(i, j int) bool {
...
 
-func componentsAndTimeFromFileName(fname string) ([]string, time.Time, error) {
-	components := strings.Split(filepath.Base(fname), separator)
-	if len(components) < 3 {
-		return nil, timeZero, fmt.Errorf("unexpected file name %s", fname)
+// Returns the positions of filename delimiters ('-' and '.') and the number of
+// delimeters found, to be used in conjunction with the intComponentFromFilename
+// function to extract filename components. This function is deliberately
+// optimized for speed and lack of allocations, since filename parsing is known
+// to account for a significant proportion of system resources.
```
richardartoul: "since allocation-heavy filename parsing can quickly become the large source of allocations in the entire system, especially when namespaces with long retentions are configured."

src/dbnode/persist/fs/files.go
```diff
@@ -373,51 +415,136 @@ func (a fileSetFilesByTimeAndVolumeIndexAscending) Less(i, j int) bool {
...
+func delimiterPositions(baseFilename string) ([maxDelimNum]int, int) {
```
richardartoul: niiiice

src/dbnode/persist/fs/files.go
```diff
@@ -373,51 +415,136 @@ func (a fileSetFilesByTimeAndVolumeIndexAscending) Less(i, j int) bool {
...
-	nanoseconds, err := strconv.ParseInt(str, 10, 64)
+
+	return delimPos, delimsFound
+}
+
+// Returns the the specified component of a filename, given the positions of
+// delimeters. Our only use cases for this involve extracting numeric
+// components, so this function assumes this and returns the component as an
+// int64.
+func intComponentFromFilename(
```
richardartoul: How do you feel about "componentAtIndex" as the name? The current name gave me the wrong impression until I read the comment

src/dbnode/persist/fs/files.go
```diff
@@ -373,51 +415,136 @@ func (a fileSetFilesByTimeAndVolumeIndexAscending) Less(i, j int) bool {
...
+	baseFilename string,
+	componentPos int,
+	delimPos [maxDelimNum]int,
+) (int64, error) {
+	var start int
```
richardartoul: can you do `var start = 0`, that would have given me a nudge in the right direction to understand how you're using the positions

src/dbnode/persist/fs/files.go
```diff
@@ -373,51 +415,136 @@ func (a fileSetFilesByTimeAndVolumeIndexAscending) Less(i, j int) bool {
...
 }
 
 // TimeFromFileName extracts the block start time from file name.
 func TimeFromFileName(fname string) (time.Time, error) {
-	_, t, err := componentsAndTimeFromFileName(fname)
-	return t, err
+	base := filepath.Base(fname)
+
+	delims, delimsFound := delimiterPositions(base)
+	if delimsFound < 2 {
```
richardartoul: You only need two delims here but I wonder if you should just always look for the minimum amount of 3, otherwise the filename could still be malformed if it only has 2 delims. Not a big deal though.

src/dbnode/persist/fs/files.go
```diff
@@ -373,51 +415,136 @@ func (a fileSetFilesByTimeAndVolumeIndexAscending) Less(i, j int) bool {
...
+		return timeZero, 0, fmt.Errorf(errUnexpectedFilename, fname)
+	}
+	unixNanos := time.Unix(0, nanos)
+
+	// Legacy filename with no volume index.
+	if delimsFound == 3 {
+		return unixNanos, unindexedFilesetIndex, nil
+	}
+
+	volume, err := intComponentFromFilename(base, 2, delims)
```
richardartoul: should we make these positions constants? I could go either way.

src/dbnode/persist/fs/files.go
```diff
@@ -1211,10 +1366,27 @@ func CommitLogsDirPath(prefix string) string {
...
+// DataFileSetExists determines whether data fileset files exist for the given
+// namespace, shard, block start, and volume.
+func DataFileSetExists(
+	filePathPrefix string,
+	namespace ident.ID,
+	shard uint32,
+	blockStart time.Time,
+	volume int,
+) (bool, error) {
+	// This function is in some cases the bottleneck, thus, we use different
```
richardartoul: This function can easily become a performance bottleneck if the implementation is slow or requires scanning directories with a large number of files in them (as is common if namespaces with long retentions are configured). As a result, instead of using existing helper function it implements an optimized code path only involves checking if a few specific files exist and contain the correct contents.

src/dbnode/persist/fs/files.go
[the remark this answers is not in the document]
richardartoul: This is resolved but I dont see the comment

src/dbnode/persist/fs/files.go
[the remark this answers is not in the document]
richardartoul: Yeah I guess its fine because of how you're using the function (where there is an implicit assumption that a file should exist, regardless of whether its legacy or not. Maybe just add that assumption as a comment above the function

src/dbnode/persist/fs/files.go
richardartoul: I assume its fine because this is only used by the reader to determine what code path to follow, and if the checkpoint file is not complete the reader will figure that out pretty quickly and error out?

src/dbnode/persist/fs/files_test.go
```diff
@@ -34,9 +34,9 @@ import (
 	"time"
 
 	"github.com/m3db/m3/src/dbnode/digest"
+	"github.com/m3db/m3/src/dbnode/namespace"
```
richardartoul: Did you have a benchmark for the new parsing function? Would be cool to have that in source control if you still have it.

src/dbnode/persist/fs/merger.go
```diff
@@ -128,6 +130,8 @@ func (m *merger) Merge(
 		NamespaceMetadata: nsMd,
 		Shard:             shard,
 		BlockStart:        startTime,
+		VolumeIndex:       fileID.VolumeIndex + 1,
```
richardartoul: I think it would be a little cleaner to have this passed in (instead of just incing) so that its very explicit what volume number you're trying to create.

src/dbnode/persist/fs/persist_manager.go
richardartoul: @justinjc just pinging your thoughts on that potential follow up

src/dbnode/storage/shard.go
```diff
@@ -1890,10 +1898,13 @@ func (s *dbShard) WarmFlush(
...
 		BlockStart:        blockStart,
+		// Volume index is always 0 for warm flushes because a warm flush must
+		// happen first before cold flushes happen.
+		VolumeIndex: 0,
 		// We explicitly set delete if exists to false here as we track which
-		// filesets exists at bootstrap time so we should never encounter a time
-		// when we attempt to flush and a fileset already exists unless there is
-		// racing competing processes.
+		// filesets exist at bootstrap time so we should never encounter a time
+		// when we attempt to flush and a fileset that already exists unless
```
richardartoul: this extra `that` seems wrong

src/dbnode/storage/shard.go
```diff
@@ -1890,10 +1898,13 @@ func (s *dbShard) WarmFlush(
...
+		// there are racing competing processes.
```
richardartoul: "there is a bug in the code" or something

src/dbnode/storage/shard.go
```diff
@@ -2006,7 +2019,7 @@ func (s *dbShard) ColdFlush(
 
 		// After writing the full block successfully, update the cold version
 		// in the flush state.
-		nextVersion := s.RetrievableBlockColdVersion(startTime) + 1
+		nextVersion := coldVersion + 1
```
richardartoul: Yeah so there is +1 logic here and in the Merge() function. Would prefer we figure out the next version here and then tell the Merge() function to use that

src/dbnode/storage/shard.go
```diff
@@ -2156,18 +2169,36 @@ func (s *dbShard) removeAnyFlushStatesTooEarly(tickStart time.Time) {
...
+		return fmt.Errorf("encountered errors when getting fileset files for prefix %s namespace %s shard %d: %v",
+			filePathPrefix, s.namespace.ID(), s.ID(), err)
 	}
-	return multiErr.FinalError()
+	// Get a snapshot of all states here to prevent constantly getting/releasing
+	// locks in a tight loop below.
+	blockStates := s.BlockStatesSnapshot()
+	// Filter without allocating by using same backing array.
+	compacted := filesets[:0]
+	for _, datafile := range filesets {
```
richardartoul: maybe add a comment saying this is safe because cleanup and flushing never happen in parallel so the snapshot will never get stale.

Also do we have `BlockStateSnapshot()` and `FlushState()` do those two duplicate each other?

src/dbnode/storage/shard.go
```diff
@@ -2156,18 +2169,36 @@ func (s *dbShard) removeAnyFlushStatesTooEarly(tickStart time.Time) {
...
+	if err != nil {
+		return fmt.Errorf("encountered errors when getting fileset files for prefix %s namespace %s shard %d: %v",
+			filePathPrefix, s.namespace.ID(), s.ID(), err)
 	}
-	return multiErr.FinalError()
+	// Get a snapshot of all states here to prevent constantly getting/releasing
+	// locks in a tight loop below.
+	blockStates := s.BlockStatesSnapshot()
+	// Filter without allocating by using same backing array.
+	compacted := filesets[:0]
```
richardartoul: I'd probably call this `toDelete` or something

richardartoul: also this is sketching me out lol can we just allocate a new slice. This is only once per shard so its not so bad

src/dbnode/persist/fs/files.go
justinjc: Sure. I'll call it `intComponentAtIndex` because I do want to say that it deals with ints only.

src/dbnode/persist/fs/files.go
justinjc: Yeah makes sense.

src/dbnode/persist/fs/files.go
justinjc: Yeah sure. I'll include a comment too to make it clear.

src/dbnode/persist/fs/files.go
justinjc: Yeah makes sense, since the other places use position constants too.

src/dbnode/persist/fs/files.go
justinjc: 👍 fixed a few typos here too.

src/dbnode/storage/shard.go
justinjc: Reworded this.

src/dbnode/storage/shard.go
justinjc: Haha sure.

src/dbnode/storage/shard.go
justinjc: Sure.

src/dbnode/storage/shard.go
justinjc: Sure I'll add a comment. Also, since versions are always increasing, even if we get stale snapshots, we just don't clean up some files that can be cleaned (but we'll clean them up eventually next time round).

`BlockStatesSnapshot` gets the flush states for all block starts, while `FlushState` is for a specific block start. Wouldn't want to loop through all blocks and keep acquiring and releasing locks (mentioned in a comment above).

src/dbnode/storage/shard.go
justinjc: Alllllright.

src/dbnode/persist/fs/merger.go
justinjc: Sure.

src/dbnode/persist/fs/files_test.go
justinjc: I think it was very simple like this with a bunch of different inputs:
```go
func BenchmarkFilenameParsing(b *testing.B) {
	for n := 0; n < b.N; n++ {
		ts, i, err := TimeAndVolumeIndexFromFileSetFilename("foo-1-0-data.db")
		_, _, _ = ts, i, err
	}
}
```
Not too valuable to include probably? Just wanted to verify no allocs and that it was significantly faster than the previous implementation.

src/dbnode/persist/fs/persist_manager.go
justinjc: Thanks for the explanation here @richardartoul, you summed it up pretty well. In short, we wanted an efficient way to figure out what is the correct volume for a particular namespace/shard/block, and having to go to disk, parsing filenames and then ordering them each time seemed significantly slower.

I'm not a big fan of the persist manager having the volume number check. The check would similarly involve looking for files in the specified directory, doing file name parsing, and ordering them in order to figure that out, which amounts to additional logic that's essentially double checking that other logic isn't buggy. Also, from a separation of concerns point of view, I don't think the persist manager should need to know that fileset volumes are always increasing. Its job should just be to persist things you tell it to persist. It should be caller of the persist manager that ensures the volume number it passes is correct.

PR #1722, opened 2019-06-10
add test case for not setting schema history to empty

Haijuncao:
dup of #1703 as the merge is kind of messed up, so start a new branch.

### DIFF
--- a/src/dbnode/storage/database_test.go
+++ b/src/dbnode/storage/database_test.go
@@ -704,7 +704,6 @@ func TestDatabaseUpdateNamespaceSchemaNotSet(t *testing.T) {
 	mapCh <- nsMap
 
 	// wait till the update has propagated
-	<-updateCh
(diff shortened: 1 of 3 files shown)


src/dbnode/storage/database_test.go
```diff
@@ -704,7 +704,6 @@ func TestDatabaseUpdateNamespaceSchemaNotSet(t *testing.T) {
 	mapCh <- nsMap
 
 	// wait till the update has propagated
-	<-updateCh
 	time.Sleep(10 * time.Millisecond)
```
richardartoul: does this sleep do anything anymore?

Haijuncao: <-updateCh is flaky, sometimes it block forever
I found that wait for 10ms is good enough for etcd change to show up.
If we fix the flaky <-updateCh, then we can remove the sleep.
 
Either block wait, or time wait, we don't need both.

PR #1727, opened 2019-06-11
[DBNode] Construct and flow block.LeaseManager to ColdFlush compaction logic and SeekerManager

richardartoul:
This P.R constructs and flows the `block.LeaseManager` to the ColdFlush compaction logic as well as to the `SeekerManager` such that they compaction logic can signal to the SeekerManager when new volumes are available after compaction.


src/dbnode/storage/block/lease_noop.go
```diff
@@ -0,0 +1,51 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package block
+
+// NoopLeaseManager is a no-op implementation of LeaseManager.
+type NoopLeaseManager struct{}
```
justinjc: Does it make more sense to just use mocks here instead of a noop implementation?

src/dbnode/server/server.go
```diff
@@ -391,6 +391,10 @@ func Run(runOpts RunOptions) {
 			scope.SubScope("tag-decoder-pool")))
 	tagDecoderPool.Init()
 
+	// Pass nil for block.LeaseVerifier for now and it will be set after the
+	// db is constructed.
```
justinjc: Maybe also add that this needs to be initialized here so that it can be propagated down before the db is created.

src/dbnode/storage/block/lease_noop.go
richardartoul: I tried it and wasn't a fan because you have to call .AnyTimes() on all of the methods and then you end up with a helper function for that and now you're exporting mocks from a package etc...

src/dbnode/server/server.go
richardartoul: sure

src/dbnode/server/server.go
```diff
@@ -464,14 +468,18 @@ func Run(runOpts RunOptions) {
 		retrieverOpts := fs.NewBlockRetrieverOptions().
 			SetBytesPool(opts.BytesPool()).
 			SetSegmentReaderPool(opts.SegmentReaderPool()).
-			SetIdentifierPool(opts.IdentifierPool())
+			SetIdentifierPool(opts.IdentifierPool()).
+			SetBlockLeaseManager(blockLeaseManager)
```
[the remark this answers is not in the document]
richardartoul: No but I think the general pattern we use is that consumers are supposed to call validate. I.E:

```golang
func NewBlockRetriever(
	opts BlockRetrieverOptions,
	fsOpts Options,
) (DataBlockRetriever, error) {
	if err := opts.Validate(); err != nil {
		return nil, err
	}
        ...
```

src/dbnode/integration/setup.go
```diff
@@ -160,9 +161,11 @@ func newTestSetup(t *testing.T, opts testOptions, fsOpts fs.Options) (*testSetup
 	// Schema registry is shared between database and admin client.
 	schemaReg := namespace.NewSchemaRegistry(opts.ProtoEncoding(), nil)
 
+	blockLeaseManager := block.NewLeaseManager(nil)
```
robskillington: Hm, won't this cause panics if verifier is nil and add lease is called?

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -463,6 +478,9 @@ func (m *seekerManager) Close() error {
 		return errSeekerManagerAlreadyClosed
 	}
 
+	// Unregister for lease updates since all the seekers are going to be closed.
+	m.blockRetrieverOpts.BlockLeaseManager().UnregisterLeaser(m)
```
robskillington: This could potentially cause a deadlock I believe, if block lease manager is calling to the seeker manager to "UpdateLease" and it acquires it's own lock, then it tries to grab the seeker manager lock, but the seeker manager already grabbed it's lock for "Close()", then the seeker manager calls "UnregisterLeaser()" which will try to grab the lease manager lock, but it's already held by a "UpdateLease" call that's waiting for the seeker manager lease to be released.

Maybe just do this outside of the lock in some way?

robskillington: This could deadlock in the "Open" too I believe.

src/dbnode/integration/setup.go
richardartoul: No I added guards around that (returns a `verifierIsNotSetErr`)

Issue #1700, opened 2019-06-04
Schema enforcement: m3db client config requires schema deploy ID


Haijuncao:
Currently, m3db client loads schema history from configuration files bundled with binary.  The schema is supposed to have been deployed to m3db meta node (etcd), but there is no enforcement.

This tickets covers the enforcement on the client side, we have another ticket that covers server side enforcement. This ticket should be done first before the server side enforcement.

As a first step, we should require schema deploy ID to be set as part of m3db client configuration, and schema deploy ID should be validated by checking it against meta node (etcd). The best way to validate it is to download the schema history from etcd and check if the deploy ID is present (may not be the latest). 

Since schema history needed to be downloaded anyway, we might as well use it to populate in memory schema registry, instead of what we do currently: from configuration.

PR #1728, opened 2019-06-11
load m3db client schema registry from etcd and add integration test

Haijuncao:
#1700 
For better client side schema enforcement (i.e. schema must be deployed before being used), we require providing `schemaDeployID` in m3db client configuration, m3db client will load schema from etcd metadata node and verify the deploy ID is valid and use the schema to populate in memory schema registry.

For the convenience of application integration test, we retain the option of loading schema from local file systems, AS LONG AS `testOnly` is set to true.

For production, `schemaDeployID` should be used and `testOnly` should be false.

m3db client configuration before this pr:
```
proto:
  enabled: true
  schema_registry:
     ns1:
        schemaFilePath: "/path/to/schema"
        messageName: "messageName"        
```

After this pr:
1. For `TestOnly` (application integration test against local docker container)
```
proto:
  enabled: true
  testOnly: true
  schema_registry:
     ns1:
        schemaFilePath: "/path/to/schema"
        messageName: "messageName"        
```
2. For production
```
proto:
  enabled: true
  schema_registry:
     ns1:
        schemaDeployID: "a_valid_deploy_id"
        messageName: "messageName"
config:
   service: 
     zone:
     env:
     etcdClusters:
```


src/dbnode/client/config.go
```diff
@@ -335,3 +341,29 @@ func (c Configuration) NewAdminClient(
...
+		return errors.New("m3db metadata store is not configured properly")
+	}
+	as := kvadmin.NewAdminService(kvStore, "", nil)
+	nsReg, err := as.GetAll()
+	if err != nil {
+		return xerrors.Wrap(err, "could not get metadata from metadata store")
+	}
+	nsMap, err := namespace.FromProto(*nsReg)
+	if err != nil {
+		return xerrors.Wrap(err, "could not unmarshall metadata")
```
richardartoul: `unmarshal`

src/dbnode/client/config.go
```diff
@@ -335,3 +341,29 @@ func (c Configuration) NewAdminClient(
...
+	}
+	merr := xerrors.NewMultiError()
+	for _, metadata := range nsMap.Metadatas() {
+		err = schemaReg.SetSchemaHistory(metadata.ID(), metadata.Options().SchemaHistory())
+		if err != nil {
+			merr.Add(xerrors.Wrapf(err, "could not set schema history for namespace %s", metadata.ID().String()))
+		}
+	}
+	if merr.Empty() {
```
richardartoul: I think you can just `return merr.FinalError()`

src/dbnode/namespace/kvadmin/ns_admin_test.go
```diff
@@ -154,3 +155,29 @@ func TestAdminService_ResetSchema(t *testing.T) {
...
+	var nsRegKey = "nsRegKey"
+	as := NewAdminService(store, nsRegKey, func() string {return "first"})
+	require.NotNil(t, as)
+
+	require.NoError(t, as.Add("ns1", namespace.OptionsToProto(namespace.NewOptions())))
+	require.Error(t, as.Add("ns1", namespace.OptionsToProto(namespace.NewOptions())))
+	require.NoError(t, as.Set("ns1", namespace.OptionsToProto(namespace.NewOptions())))
+	require.Error(t, as.Set("ns2", namespace.OptionsToProto(namespace.NewOptions())))
+
+	nsOpt, err := as.Get("ns1")
```
richardartoul: can we do an assert that the returned thing is correct?

src/dbnode/namespace/kvadmin/ns_admin.go
```diff
@@ -43,38 +44,108 @@ type adminService struct {
...
 		store: store,
 		key:   key,
 		idGen: idGen,
 	}
 }
 
-func (as *adminService) GetAll() ([]*nsproto.NamespaceOptions, error) {
-	// TODO [haijun] move logic from src/query/api/v1/handler/namespace here
-	return nil, ErrNotImplemented
+func (as *adminService) GetAll() (*nsproto.Registry, error) {
```
richardartoul: This is cool. Are we gonna replace the handler code with the service soon? Do we have an open issue for that?

src/dbnode/integration/client_load_schema_test.go
```diff
@@ -0,0 +1,130 @@
...
+		},
+		Proto: &client.ProtoConfiguration{
+			Enabled: true,
+			SchemaRegistry: map[string]client.NamespaceProtoSchema{
+				"ns1": {MessageName: protoMsg, SchemaDeployID: deployID},
+			},
+		},
+	}
+
+	adminClient, err := cfg.NewAdminClient(client.ConfigurationParameters{})
```
richardartoul: Can we add one more test case where you try and instantiate a client with an invalid configuration? Would also be really good to explicitly try the case where:

1. You use the latest `deployID`
2. You use a valid `deployID` but not the latest one

src/dbnode/client/config.go
```diff
@@ -234,11 +236,10 @@ func (c Configuration) NewAdminClient(
 	writeRequestScope := iopts.MetricsScope().SubScope("write-req")
 	fetchRequestScope := iopts.MetricsScope().SubScope("fetch-req")
 
-	envCfg := environment.ConfigureResults{
-		TopologyInitializer: params.TopologyInitializer,
-	}
+	var envCfg environment.ConfigureResults
 
-	if envCfg.TopologyInitializer == nil {
+	// Initialize envCfg
```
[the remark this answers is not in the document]
Haijuncao: Amazingly, I understood!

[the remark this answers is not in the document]
richardartoul: Haha that’s what happens when you spend a lot of time reviewing each other’s code

src/dbnode/client/config.go
```diff
@@ -234,11 +236,10 @@ func (c Configuration) NewAdminClient(
...
+	if c.EnvironmentConfig != nil {
```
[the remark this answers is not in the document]
Haijuncao: it does HAVE to change, but I feel it is cleaner to change it.

The way it was written, it will not init/populate envCfg unless topoInit is not passed in, which is fine because topology Initializer is the only one that needed envCfg.

Now that this assumption does not hold anymore. 

It is cleaner to init envCfg from client.Configuration if EnviromentConfig is set (under the client section). For dbnode's admin client, client won't have EnviromentConfig set. For normal m3db client, client will have EnviromentConfig set.

src/dbnode/client/config.go
Haijuncao: cool

src/dbnode/namespace/kvadmin/ns_admin_test.go
Haijuncao: sure

src/dbnode/namespace/kvadmin/ns_admin.go
Haijuncao: let me open an issue to track it. I won't get to it before my vacation... so it won't be soon

src/dbnode/integration/client_load_schema_test.go
Haijuncao: sure

src/dbnode/client/config.go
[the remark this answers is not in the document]
richardartoul: ok makes sense

PR #1731, opened 2019-06-13
Add aggregation specific errors

robskillington:

Returns specific errors for metrics that are being aggregated that arrive too late or too early.

```release-note
NONE
```

```documentation-note
NONE
```


robskillington:
@richardartoul sure thing not opposed to that, put it behind a flag for now and added config that allows adjusting the values if required too.

PR #1736, opened 2019-06-14
[DBNode] Add LeaseLatest() API to block.LeaseManager

richardartoul:
This P.R adds a `OpenLeaseForLatest()` API to the block.LeaseManager so that when the SeekManager is opening seekers of its own accord (as opposed to as the result of receiving an `UpdateOpenLeases()` notification, it can determine which volume to open for a given block.

### DIFF
--- a/src/dbnode/storage/block/types.go
+++ b/src/dbnode/storage/block/types.go
@@ -416,6 +416,8 @@ type LeaseVerifier interface {
 		descriptor LeaseDescriptor,
 		state LeaseState,
 	) error
+
+	LatestState(descriptor LeaseDescriptor) (LeaseState, error)
 }
 
 // UpdateOpenLeaseResult is the result of processing an update lease.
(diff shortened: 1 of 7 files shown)


src/dbnode/storage/block/types.go
```diff
@@ -384,6 +384,7 @@ type LeaseManager interface {
 		descriptor LeaseDescriptor,
 		state LeaseState,
 	) error
+	OpenLeaseForLatest(leaser Leaser, descriptor LeaseDescriptor) (LeaseState, error)
```
justinjc: How about `OpenLatestLease`?

src/dbnode/storage/block/types.go
```diff
@@ -416,6 +417,8 @@ type LeaseVerifier interface {
 		descriptor LeaseDescriptor,
 		state LeaseState,
 	) error
+
+	LatestState(descriptor LeaseDescriptor) (LeaseState, error)
```
justinjc: Can you put some comments with these functions?

PR #1737, opened 2019-06-14
[DBNode] Register/unregister leaser outside of locks in SeekManager and make UpdateOpenLeases() concurrency safe

richardartoul:
The existing implementation has a potential to deadlock(although not an actual bug in master since this is "dead" / feature-flagged code) when the SeekManager is being closed as described in the comments in the P.R.

### DIFF
--- a/src/dbnode/persist/fs/seek_manager.go
+++ b/src/dbnode/persist/fs/seek_manager.go
@@ -157,20 +157,23 @@ func (m *seekerManager) Open(
 	nsMetadata namespace.Metadata,
 ) error {
 	m.Lock()
-	defer m.Unlock()
-
 	if m.status != seekerManagerNotOpen {
+		m.Unlock()
 		return errSeekerManagerAlreadyOpenOrClosed
 	}
-
-	// Register for updates to block leases.
-	m.blockRetrieverOpts.BlockLeaseManager().RegisterLeaser(m)
 
 	m.namespace = nsMetadata.ID()
 	m.namespaceMetadata = nsMetadata
 	m.status = seekerManagerOpen
-
 	go m.openCloseLoop()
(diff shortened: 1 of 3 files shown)


src/dbnode/persist/fs/seek_manager.go
```diff
@@ -157,20 +157,23 @@ func (m *seekerManager) Open(
...
-
 	go m.openCloseLoop()
+	m.Unlock()
+
+	// Register for updates to block leases.
+	// NB(rartoul): This should be safe to do within the context of the lock
+	// because the block.LeaseManager does not yet have a handle on the SeekerManager
+	// so they can't deadlock trying to acquire each other's locks, but do it outside
+	// of the lock just to be safe.
+	m.blockRetrieverOpts.BlockLeaseManager().RegisterLeaser(m)
```
prateek: would be good to add a test that would have caught this too.

richardartoul: Yeah I was trying to think of a good way to do that and couldn't think of one that wouldn't be incredibly superficial.

Its also super implementation-specific. For example this could have only happened if the `UpdateOpenLease` method on the SeekerManager acquired a lock on the SeekerManager itself, but it probably won't (because Borrow and Return don't even do that, they only grab the more granular lock on `seekersByShardIdx`).

Basically its really hard to write a test to guard against something that couldn't have really happened anyways. This is just us being extra precautious.

richardartoul: @prateek This I found a way but its basically by making UpdateOpenLeases() on the LeaseManager not hold a lock and then testing concurrent interactions there. I have two unit tests right now and will add a real concurrency test soon

Issue #1678, opened 2019-05-30
m3query cpu  high usage


signmem:
m3cluster
> five Huawei compute here.
>I use 3 compute build m3db cluster.
>another one compute use for m3query.
> another one use for ab tools testing.

useto
>I use ab tools test m3query  to get QPS.
>get metric from m3dbnode in one hour ( in memory)
>time step = 15s
> total metric value is 240.

ab command above:
```
ab -n 6000 -c 30  'http://m3query:7207/api/v1/query_range?query=user_login&start=1559205493&end=1559209092&step=15s'
```

testing resule

>I got about 1111 QPS. (its tooooooo low I believe.)
>in the condition.  compute runing m3query process is high cpu usage . 
>HYPERVISOR is 24CORE INTEL CPU. and just use to run m3query process.  no other process useful.
>cpu load avg is increasing. and cpu idel is about 5% ~ 7%

here is my profile svg example :

[profile002.zip](https://github.com/m3db/m3/files/3236785/profile002.zip)

*can you please tell me how can I resolve this problem ?*

martin-mao:
@benraskin92 @arnikola Can have a deeper look, but from the profile, it seems like most of your CPU is spend on decoding series (fetching data). There is about 20% spend on cost accounting and possibly being rate limited, so maybe we could increase some limitations there, but the bulk of it is spent on actual queries.

arnikola:
Would you be able to show us a flame graph too? We're looking at a similar issue that someone else reported and that would help us understand if it's the same root cause

arnikola:
@signmem Could you also post your coordinator configuration? In particular, the `namespaces` stanza

arnikola:
Also, any chance you can give us your ns config? You can do `curl <address>:7201/api/v1/namespace` to get it

signmem:
# purpose
> We wanna build tsdb perform using M3DB cluster use for production environment.  
> So I tested the m3db query and write performance limit.
> I use ab tools to connect m3query to testing QPS.

# structure
its my hypervisor structure.
<table>
<tr>
<td>hostname</td><td>ram</td><td>cpu</td><td>role</td><td>network</td>
</tr>
<tr>
<td>m3dbnode01</td><td>128GB</td><td>32CORE</td><td>m3dbnode</td><td>10000m X 2 (bonding)</td>
</tr>
<tr>
[...]

PR #1740, opened 2019-06-15
[query] Add capability for parallel block iteration

arnikola:

Should reduce user impact seen in #1678 

**Special notes for your reviewer**: Majority of work is in `encoded_step_iterator_generic.go`, `encoded_unconsolidated_step_iterator`, and `encoded_step_iterator`; rest of is plumbingg

```release-note
NONE
```

```documentation-note
NONE
```


src/query/ts/m3db/consolidators/step_accumulator.go
```diff
@@ -61,35 +56,28 @@ func NewStepLookbackAccumulator(
...
 	// iteration of the datapoint set.
 	c.earliestLookback = c.earliestLookback.Add(c.stepSize)
-	accumulated := make([]xts.Datapoints, len(c.datapoints))
-	for i, dps := range c.datapoints {
-		accumulated[i] = make(xts.Datapoints, len(dps))
-		copy(accumulated[i], dps)
-		c.datapoints[i] = c.datapoints[i][:0]
-	}
-
+	accumulated := make([]xts.Datapoint, len(c.datapoints))
```
robskillington: Is it plausible to pool these allocations at some point in the future?  

I'm still trying to understand the lifecycle better here, so I'm definitely out of my depth in being able to say one way or another whether it's a good idea hah.

src/query/ts/m3db/consolidators/step_accumulator.go
```diff
@@ -35,7 +35,7 @@ type StepLookbackAccumulator struct {
 	lookbackDuration time.Duration
 	stepSize         time.Duration
 	earliestLookback time.Time
-	datapoints       [][]xts.Datapoint
+	datapoints       []xts.Datapoint
```
robskillington: Nice, this was able to be flattened?

src/query/ts/m3db/encoded_step_iterator.go
```diff
@@ -26,43 +26,70 @@ import (
...
-		cs.currentTime,
-		len(iters),
-		cs.consolidationFn,
+	var (
+		cs       = b.consolidation
+		iters    = b.seriesBlockIterators
+		lookback = b.options.LookbackDuration()
+
+		collectors       = make([]*consolidators.StepLookbackConsolidator, len(iters))
+		seriesCollectors = make([]consolidators.StepCollector, 0, len(iters))
```
robskillington: nit: Maybe make them both size zero but cap len(iters) or the opposite so it's consistent?

src/query/ts/m3db/encoded_step_iterator_generic.go
```diff
@@ -88,43 +101,103 @@ func (it *encodedStepIterWithCollector) nextForStep(
...
+	for i := range it.seriesIters {
+		var (
+			i         = i
+			peek      = it.seriesPeek[i]
+			iter      = it.seriesIters[i]
+			collector = it.seriesCollectors[i]
+		)
+
+		it.workerPool.Go(func() {
+			peek, collector, err := nextForStep(peek, iter, collector, stepTime)
```
robskillington: Nice this is a solid approach.

I wonder if it would also be a good idea to basically do a bunch in parallel (like the next X points defined by a constant, like 10 or 20) and then read from intermediary storage after?  That way we don't have to use tons of memory decoding everything up front but every time we do a decode step, we can hot loop the decoding for a little while.  Would also help with less context switching (using goroutines to do decoding, then waiting on result less often).

Basically exactly what you're doing here in batch.

src/query/ts/m3db/encoded_step_iterator_generic.go
```diff
@@ -88,43 +101,103 @@ func (it *encodedStepIterWithCollector) nextForStep(
...
 	if it.err != nil {
 		return false
 	}
 
-	it.stepTime = it.stepTime.Add(bounds.StepSize)
-	return !bounds.End().Before(it.stepTime)
+	it.stepTime = it.stepTime.Add(it.meta.Bounds.StepSize)
+	next := !it.blockEnd.Before(it.stepTime)
+	if next && it.updateFn != nil {
+		it.updateFn()
```
robskillington: What's this used for btw?

src/query/ts/m3db/encoded_step_iterator_test.go
```diff
@@ -268,3 +310,159 @@ func TestConsolidatedStepIteratorSplitByBlock(t *testing.T) {
...
+		}
+
+		for _, it := range pIters {
+			for _, p := range points {
+				it.EXPECT().Current().Return(p, xtime.Second, annotation)
+			}
+
+			it.EXPECT().Next().DoAndReturn(func() bool {
+				// simulate load
+				time.Sleep(time.Millisecond)
```
robskillington: Hm instead of mock iterators how about just encoding a bunch into real M3TSZ encoders?  

Not sure if time sleeping is a good proxy for real decoding work since always waiting for a fixed amount of time at each point will have different scheduling characteristics vs actually doing CPU work during a call to Next() on a single iterator.

src/query/ts/m3db/consolidators/step_accumulator.go
arnikola: Yeah, I was initially trying to abstract away the `i-th` wise stuff in the generic step iterator and let this accumulator handle it, but that doesn't lend very well to parallelization

src/query/ts/m3db/encoded_step_iterator.go
arnikola: Sure, is there any preference either way? I've generally leaned toward initializing explicitly with size when I know what the size will be, but not sure that's best (seems `append` is often a little safer)

src/query/ts/m3db/encoded_step_iterator_generic.go
arnikola: This collates the results post `Next()`, so when `Current()` is called on the iterator it does virtually no work. Also partially necessary here because the step collector `ConsolidateAndMoveToNext` functions currently mutate their internal steps, which is easy enough to change but this way tries to avoid the extra checks etc that would be added otherwise

src/query/ts/m3db/encoded_step_iterator_test.go
arnikola: Yeah totally fair, i'll try and do one with no mocks

src/query/ts/m3db/consolidators/step_accumulator.go
arnikola: Slightly simplified flow for this is that the `N-th` iterator in the encoded step iterator has it's `Next()` called, adding those points to the `N-th StepCollector`, until it either

a) returns false, in which case the accumulated values are consolidated by the `StepCollector` and set to the `Current` values

b) yields a point whose timestamp is beyond the current step, in which case that point is saved as a "peeked value" in preparation for the next time `Next()` is called, and the accumulated values are consolidated as for the above case.

We definitely should add pooling here but it may be a fairly involved change largely because we don't have a great lifecycle management story for them yet

arnikola: In terms of the different `StepCollectors`, the step iterator variant that uses `StepLookbackAccumulator` is the 
"unconsolidated" one, which just gives back all the raw points. These are used only in temporal functions like `sum_over_time`, since they need to work on the rawest data available. All othe functions use the `StepLookbackConolidator`, which gives "consolidated" data points, which are snapped to stepSize and have a given consolidation function applied to them.

Frankly, I really dislike the concept of having this consolidated/unconsolidated split as is; I think we may be able to add a tiny bit of  carefully placed logic to slot in a specialized consolidation function where necessary (also that will  also give us `summarize` for free).

src/query/ts/m3db/encoded_step_iterator.go
robskillington: Append just tends to be safer yeah, I've seen a lot of bugs of creating size, then using append and then empty values sit in front of real values in the result.

If it's super straight forward and in super hot path it's probably justified to not use append (since might be slightly faster, don't need to keep updating the size of the array in the slice, although that's obviously super fast addition too).

PR #1741, opened 2019-06-17
[docs] Add initial M3Aggregator documentation

robskillington:

This documentation is not meant to be comprehensive but will serve as a start on providing documentation for running M3Aggregator.

```release-note
NONE
```

```documentation-note
Add initial M3Aggregator documentation
```


docs/how_to/aggregator.md
```diff
@@ -0,0 +1,159 @@
+# Setting up m3aggregator
+
+## Introduction
+
+m3aggregator is used to cluster stateful downsampling and rollup of metrics before they are store in M3DB. The m3coordinator also performs this role but is not cluster aware. This means metrics will not get aggregated properly if you send metrics in round round fashion to multiple m3coordinators for the same metrics ingestion source (e.g. Prometheus server).
```
richardartoul: store -> stored

richardartoul: We usually say m3coordinator not "the m3coordinator"

richardartoul: "m3coordinator can also perform this role, but is not clustered. This makes it difficult to scale out as metrics will not get aggregated properly when using round robin load balancing across multiple m3coordinators (since aggregation is a stateful operation)" or something like that

docs/how_to/aggregator.md
```diff
@@ -0,0 +1,159 @@
...
+
+Metrics sent to m3aggregator are correctly routed to the instance resposible for aggregating each metric and there is multiple m3aggregator replicas to ensure no single point of failure during aggregation.
```
richardartoul: "Similar to M3DB, m3aggregator supports clustering by default. This means that metrics are correctly routed to the instance(s) responsible for aggregating each metric and multiple m3aggregator replicas can be configured such that there are no single points of failure for aggregation"

docs/how_to/aggregator.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Configuration
+
+Before setting up m3aggregator, make sure that you have at least [one M3DB node running](single_node.md) and a dedicated m3coordinator setup.
+
+We highly recommend running with at least a replication factor 2 for a m3aggregator deployment.
```
richardartoul: maybe add that if you run with R.F=1 you'll have downtime whenever you perform cluster operations like add/removes/joins

docs/how_to/aggregator.md
```diff
@@ -0,0 +1,159 @@
...
+```
+make m3aggregator
+./bin/m3aggregator -f ./src/aggregator/config/m3aggregator.yml
+```
+
+Or you can run it with Docker using the Docker file located at `$GOPATH/src/github.com/m3db/m3/docker/m3aggregator/Dockerfile`.
+
+### Coordinator configuration
+
+Metrics will still arrive at the m3coordinator, they simply need to be forwarded to m3aggregators. The m3coordinator then also needs to receive metrics that have been aggregated from the m3aggregator and store them in M3DB, so running an ingestion server should be configured.
```
richardartoul: a digram here would be super helpful since its pretty confusing to have this prometheus --> m3coordinator --> m3aggregator --> m3coordinator flow.

Would also be good to say that the m3coordinators in that diagram can be different

docs/how_to/aggregator.md
```diff
@@ -0,0 +1,159 @@
...
+
+We highly recommend running with at least a replication factor 2 for a m3aggregator deployment.
+
+### Running
+
+You can run m3aggregator by either building and running the binary yourself:
+
+```
+make m3aggregator
+./bin/m3aggregator -f ./src/aggregator/config/m3aggregator.yml
```
mmedvede: Should sample `m3aggregator.yml` be added to the doc, or explicit mention about using the one from source? I am going through the setup and found that was the missing piece so far.

docs/how_to/aggregator.md
```diff
@@ -0,0 +1,159 @@
...
+### Aggregation topic and m3coordinator discovery
+
+#### Create aggregation topic
+
+Now we must setup a topic for the m3aggregator to send computed metrics aggregations back to an m3coordinator instance for storage to M3DB:
+```
+curl -vvvsSf -X POST http://m3coordinator01:7201/api/v1/topic/init -d '{
+    "numberOfShards": 64
+}'
+```
```
mmedvede: Should the default topic name be mentioned? Looking through the source code, the default topic name used is `aggregated_metrics`. And there seems to be a way to specify the topic by setting `topic-name` in the http request header.

docs/how_to/aggregator.md
robskillington: True, good point.

docs/how_to/aggregator.md
robskillington: Yes we should definitely mention that.

docs/how_to/aggregator.md
pavelnikolov: "round round fashion" -> "round robin fashion"

bpownow:
any chance we can revive this PR @robskillington?

docs/how_to/aggregator.md
```diff
@@ -0,0 +1,154 @@
...
+            "id": "m3aggregator02:6000",
+            "isolation_group": "availability-zone-b",
+            "zone": "embedded",
+            "weight": 100,
+            "endpoint": "m3aggregator02:6000",
+            "hostname": "m3aggregator02",
+            "port": 6000
+        }
+    ]
+}'
```
pavelnikolov: (off-topic) I know that the goal of this PR is to only document the process, however, I was wondering if the components can self-register with each other. Something like automated service-discovery process. WDYT?

asafm:
@robskillington is the documentation in this PR still valid (noticed it was written a year ago)?

robskillington:
@asafm it is now, about to merge this 👍

docs/how_to/aggregator.md
robskillington: Fixed.

docs/how_to/aggregator.md
robskillington: That would be ideal yes, at some point we could look at Kubernetes.

docs/how_to/aggregator.md
```diff
@@ -43,18 +44,58 @@ curl -vvvsSf -H "Cluster-Environment-Name: namespace/m3db-cluster-name" -X POST
...
+    "messageTtlNanos": "300000000000"
+  }
+}'
+```
+
+**Note:** 300000000000 nanoseconds is a TTL of 5 minutes for messages to rebuffer for retry.
+
+#### Initializing m3msg topic for m3coordinator to receive from m3aggregator to write to M3DB
+
+Now we must setup a topic for the `m3coordinator` to receive unaggregated metrics from `m3aggregator` instances to write to M3DB:
```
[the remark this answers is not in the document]
robskillington: Ah good catch, yes.

PR #1742, opened 2019-06-17
[m3db] Add m3db index tracing

benraskin92:

This PR introduces tracing through the m3db index. 

Screen Shot 2019-06-18 at 6 01 36 PM

```release-note
This PR introduces tracing through the m3db index. 
```

```documentation-note
NONE
```

### DIFF
--- a/scripts/development/m3_stack/start_m3.sh
+++ b/scripts/development/m3_stack/start_m3.sh
@@ -27,7 +27,8 @@ if [[ "$USE_JAEGER" = true ]] ; then
 fi
 
 docker-compose -f docker-compose.yml up $DOCKER_ARGS m3coordinator01
-docker-compose -f docker-compose.yml up $DOCKER_ARGS m3db_seed
+# Remove this!!!
+docker-compose -f docker-compose.yml up --build -d --renew-anon-volumes m3db_seed
 docker-compose -f docker-compose.yml up $DOCKER_ARGS prometheus01
 docker-compose -f docker-compose.yml up $DOCKER_ARGS grafana
 
--- a/scripts/development/m3_stack/start_m3.sh
+++ b/scripts/development/m3_stack/start_m3.sh
@@ -27,8 +27,7 @@ if [[ "$USE_JAEGER" = true ]] ; then
 fi
 
 docker-compose -f docker-compose.yml up $DOCKER_ARGS m3coordinator01
-# Remove this!!!
-docker-compose -f docker-compose.yml up --build -d --renew-anon-volumes m3db_seed
+docker-compose -f docker-compose.yml up $DOCKER_ARGS m3db_seed
 docker-compose -f docker-compose.yml up $DOCKER_ARGS prometheus01
 docker-compose -f docker-compose.yml up $DOCKER_ARGS grafana
 
--- a/src/dbnode/storage/index.go
+++ b/src/dbnode/storage/index.go
@@ -189,6 +189,7 @@ type execBlockQueryFn func(
 	opts index.QueryOptions,
 	state *asyncQueryExecState,
 	results index.BaseResults,
+	logFields []opentracinglog.Field,
 )
 
 // asyncQueryExecState tracks the async execution errors and results for a query.
@@ -932,14 +933,16 @@ func (i *nsIndex) Query(
 	query index.Query,
 	opts index.QueryOptions,
 ) (index.QueryResult, error) {
-	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxQuery)
-	sp.LogFields(
+	logFields := []opentracinglog.Field{
 		opentracinglog.String("query", query.String()),
 		opentracinglog.String("namespace", i.nsMetadata.ID().String()),
 		opentracinglog.Int("limit", opts.Limit),
-		xopentracing.Time("start", opts.StartInclusive),
-		xopentracing.Time("end", opts.EndExclusive),
-	)
+		xopentracing.Time("queryStart", opts.StartInclusive),
+		xopentracing.Time("queryEnd", opts.EndExclusive),
+	}
+
+	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxQuery)
+	sp.LogFields(logFields...)
 	defer sp.Finish()
 
 	// Get results and set the namespace ID and size limit.
@@ -948,7 +951,7 @@ func (i *nsIndex) Query(
 		SizeLimit: opts.Limit,
 	})
 	ctx.RegisterFinalizer(results)
-	exhaustive, err := i.query(ctx, query, results, opts, i.execBlockQueryFn)
+	exhaustive, err := i.query(ctx, query, results, opts, i.execBlockQueryFn, logFields)
 	if err != nil {
 		sp.LogFields(opentracinglog.Error(err))
 		return index.QueryResult{}, err
@@ -964,14 +967,16 @@ func (i *nsIndex) AggregateQuery(
 	query index.Query,
 	opts index.AggregationOptions,
 ) (index.AggregateQueryResult, error) {
-	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxAggregateQuery)
-	sp.LogFields(
+	logFields := []opentracinglog.Field{
 		opentracinglog.String("query", query.String()),
 		opentracinglog.String("namespace", i.nsMetadata.ID().String()),
 		opentracinglog.Int("limit", opts.Limit),
-		xopentracing.Time("start", opts.StartInclusive),
-		xopentracing.Time("end", opts.EndExclusive),
-	)
+		xopentracing.Time("queryStart", opts.StartInclusive),
+		xopentracing.Time("queryEnd", opts.EndExclusive),
+	}
+
+	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxAggregateQuery)
+	sp.LogFields(logFields...)
 	defer sp.Finish()
 
 	// Get results and set the filters, namespace ID and size limit.
@@ -996,7 +1001,7 @@ func (i *nsIndex) AggregateQuery(
 	}
 	aopts.FieldFilter = aopts.FieldFilter.SortAndDedupe()
 	results.Reset(i.nsMetadata.ID(), aopts)
-	exhaustive, err := i.query(ctx, query, results, opts.QueryOptions, fn)
+	exhaustive, err := i.query(ctx, query, results, opts.QueryOptions, fn, logFields)
 	if err != nil {
 		return index.AggregateQueryResult{}, err
 	}
@@ -1012,18 +1017,15 @@ func (i *nsIndex) query(
 	results index.BaseResults,
 	opts index.QueryOptions,
 	execBlockFn execBlockQueryFn,
+	logFields []opentracinglog.Field,
 ) (bool, error) {
 	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxQueryHelper)
 	sp.LogFields(
-		opentracinglog.String("query", query.String()),
-		opentracinglog.String("namespace", i.nsMetadata.ID().String()),
-		opentracinglog.Int("limit", opts.Limit),
-		xopentracing.Time("start", opts.StartInclusive),
-		xopentracing.Time("end", opts.EndExclusive),
+		logFields...,
 	)
 	defer sp.Finish()
 
-	exhaustive, err := i.queryWithSpan(ctx, query, results, opts, execBlockFn, sp)
+	exhaustive, err := i.queryWithSpan(ctx, query, results, opts, execBlockFn, sp, logFields)
 	if err != nil {
 		sp.LogFields(opentracinglog.Error(err))
 	}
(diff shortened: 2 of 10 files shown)


src/dbnode/storage/index.go
```diff
@@ -987,13 +1012,25 @@ func (i *nsIndex) query(
...
+
 	// Capture start before needing to acquire lock.
 	start := i.nowFn()
 
 	i.state.RLock()
 	if !i.isOpenWithRLock() {
 		i.state.RUnlock()
-		return false, errDbIndexUnableToQueryClosed
+		err := errDbIndexUnableToQueryClosed
+		sp.LogFields(opentracinglog.Error(err))
```
robskillington: Hey can we perhaps just make a method to handle this for us?  Perhaps put it in the `context` package?

```go
func LogSpanErrorAndReturn(sp opentracing.Span, err error) error {
  sp.LogFields(opentracinglog.Error(err))
  return err
}
```

Then you can just:
```go
  return false, context.LogSpanErrorAndReturn(err)
```

src/dbnode/storage/index.go
```diff
@@ -1129,22 +1173,35 @@ func (i *nsIndex) query(
...
+		opentracinglog.String("namespace", i.nsMetadata.ID().String()),
+		opentracinglog.Int("limit", opts.Limit),
+		xopentracing.Time("start", opts.StartInclusive),
+		xopentracing.Time("end", opts.EndExclusive),
+	)
+	defer sp.Finish()
+
+	blockExhaustive, err := block.Query(ctx, cancellable, query, opts, results)
 	if err == index.ErrUnableToQueryBlockClosed {
+		sp.LogFields(opentracinglog.Error(err))
```
robskillington: I wouldn't log this error, this is an expected error (as explained by the comment below).

src/dbnode/storage/index.go
```diff
@@ -1017,6 +1054,7 @@ func (i *nsIndex) query(
 	i.state.RUnlock()
 
 	if err != nil {
+		sp.LogFields(opentracinglog.Error(err))
```
robskillington: Hm instead of logging error at every single place that returns an error, can you just create a sub function, and then log error if any of the subfunction returns an error?

Also makes it easy to avoid needing the defer.
i.e.
```go
func (i *nsIndex) query(
	ctx context.Context,
	query index.Query,
	results index.BaseResults,
	opts index.QueryOptions,
	execBlockFn execBlockQueryFn,
) (bool, error) {
	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxSubQuery)
	sp.LogFields(
		opentracinglog.String("query", query.String()),
		opentracinglog.String("namespace", i.nsMetadata.ID().String()),
		opentracinglog.Int("limit", opts.Limit),
		xopentracing.Time("start", opts.StartInclusive),
		xopentracing.Time("end", opts.EndExclusive),
	)
	defer sp.Finish()

	result, err := i.queryWithSpan(ctx, query, results, opts, execBlockFn, sp)
	if err != nil {
		sp.LogFields(opentracinglog.Error(err))
	}

	return result, err
}

func (i *nsIndex) queryWithSpan(
	ctx context.Context,
	query index.Query,
	results index.BaseResults,
	opts index.QueryOptions,
	execBlockFn execBlockQueryFn,
	span opentracing.Span,
) (bool, error) {
  // .. all existing code
}
```

Same with every other large code paths with lots of early returns in this PR.

src/dbnode/storage/index.go
benraskin92: Yeah was gonna ask about this error and a couple others. Sounds good.

src/dbnode/storage/index.go
benraskin92: Sounds good - will update.

scripts/development/m3_stack/start_m3.sh
```diff
@@ -27,7 +27,8 @@ if [[ "$USE_JAEGER" = true ]] ; then
 fi
 
 docker-compose -f docker-compose.yml up $DOCKER_ARGS m3coordinator01
-docker-compose -f docker-compose.yml up $DOCKER_ARGS m3db_seed
+# Remove this!!!
```
prateek: Still relevant?

src/dbnode/storage/index.go
```diff
@@ -986,6 +1012,32 @@ func (i *nsIndex) query(
 	results index.BaseResults,
 	opts index.QueryOptions,
 	execBlockFn execBlockQueryFn,
+) (bool, error) {
+	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxSubQuery)
+	sp.LogFields(
```
prateek: Dunno convention but IMO child spans don’t need to log fields that are the same as the parent. 

ie all the fields in this log are already logged by Aggregate/Query

src/dbnode/storage/index.go
```diff
@@ -986,6 +1012,32 @@ func (i *nsIndex) query(
 	results index.BaseResults,
 	opts index.QueryOptions,
 	execBlockFn execBlockQueryFn,
+) (bool, error) {
+	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxSubQuery)
```
prateek: Var naming nit: mind calling this NSIdxQueryHelper instead? SubQuery has a connotation best avoided here

src/dbnode/storage/index.go
```diff
@@ -1136,14 +1188,27 @@ func (i *nsIndex) query(
...
+	ctx context.Context,
 	cancellable *resource.CancellableLifetime,
 	block index.Block,
 	query index.Query,
 	opts index.QueryOptions,
 	state *asyncQueryExecState,
 	results index.BaseResults,
 ) {
-	blockExhaustive, err := block.Query(cancellable, query, opts, results)
+	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxBlockQuery)
```
prateek: Here too. The only field that’s different from parents is the block start/end time.

src/dbnode/storage/index.go
```diff
@@ -1156,29 +1221,43 @@ func (i *nsIndex) execBlockQueryFn(
...
 
 func (i *nsIndex) execBlockAggregateQueryFn(
+	ctx context.Context,
 	cancellable *resource.CancellableLifetime,
 	block index.Block,
 	query index.Query,
 	opts index.QueryOptions,
 	state *asyncQueryExecState,
 	results index.BaseResults,
 ) {
```
prateek: Here to

src/dbnode/storage/index/block.go
```diff
@@ -765,10 +769,39 @@ func (b *block) segmentsWithRLock() []segment.Segment {
...
 // to the results datastructure).
 func (b *block) Query(
+	ctx context.Context,
 	cancellable *resource.CancellableLifetime,
 	query Query,
 	opts QueryOptions,
 	results BaseResults,
+) (bool, error) {
+	ctx, sp := ctx.StartTraceSpan(tracepoint.BlockQuery)
+	sp.LogFields(
```
prateek: Here too, don’t think this has any more than the parent

src/dbnode/storage/index/block.go
```diff
@@ -882,9 +915,36 @@ func (b *block) addQueryResults(
 // for the case when we can skip going to raw documents, and instead rely on
 // pre-aggregated results via the FST underlying the index.
 func (b *block) Aggregate(
+	ctx context.Context,
+	cancellable *resource.CancellableLifetime,
+	opts QueryOptions,
+	results AggregateResults,
+) (bool, error) {
+	ctx, sp := ctx.StartTraceSpan(tracepoint.BlockAggregate)
+	sp.LogFields(
```
prateek: Here too

scripts/development/m3_stack/start_m3.sh
benraskin92: Nope -- just for testing. Will remove.

src/dbnode/storage/index.go
benraskin92: Yeah so I'm kind of unsure of this too. On the one hand you're just duplicating data. On the other hand, it's easier to grok with the info right there in the span. Let me read up on the convention.

src/dbnode/storage/index.go
benraskin92: Yeah agreed - `subQuery` is weird..

src/dbnode/storage/index.go
prateek: Yeah that’s fair. I’m happy to leave in if it helps readability.

prateek:
Looks like build is still :(

src/dbnode/storage/index.go
benraskin92: So there's not really any best practices around this (at least not that I could find), but I think it's better to leave these in here for readability.

src/dbnode/storage/index.go
```diff
@@ -1136,14 +1188,27 @@ func (i *nsIndex) query(
...
-	blockExhaustive, err := block.Query(cancellable, query, opts, results)
+	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxBlockQuery)
+	sp.LogFields(
+		opentracinglog.String("query", query.String()),
+		opentracinglog.String("namespace", i.nsMetadata.ID().String()),
+		opentracinglog.Int("limit", opts.Limit),
+		xopentracing.Time("query_start", opts.StartInclusive),
+		xopentracing.Time("query_end", opts.EndExclusive),
+		xopentracing.Time("block_start", block.StartTime()),
+		xopentracing.Time("block_end", block.EndTime()),
```
robskillington: nit: Can we be consistent and use lower camel case for naming log fields and trace log fields? i.e. `queryStart` vs `query_start`.

src/dbnode/storage/index.go
```diff
@@ -1156,29 +1221,43 @@ func (i *nsIndex) execBlockQueryFn(
...
 ) {
+	ctx, sp := ctx.StartTraceSpan(tracepoint.NSIdxBlockAggregateQuery)
+	sp.LogFields(
+		opentracinglog.String("query", query.String()),
+		opentracinglog.String("namespace", i.nsMetadata.ID().String()),
+		opentracinglog.Int("limit", opts.Limit),
+		xopentracing.Time("query_start", opts.StartInclusive),
+		xopentracing.Time("query_end", opts.EndExclusive),
+		xopentracing.Time("block_start", block.StartTime()),
+		xopentracing.Time("block_end", block.EndTime()),
```
robskillington: nit: Same with these fields, e.g. `queryStart` vs `query_start`.

src/dbnode/storage/index/block.go
robskillington: How about we just remove the spans in `execBlockAggregateQueryFn` and `execBlockQueryFn` and keep these ones inside the block themselves?

It seems like it doesn't add much to have those ones as they are just wrappers to call into these.

src/dbnode/storage/index/block.go
```diff
@@ -882,9 +915,36 @@ func (b *block) addQueryResults(
...
+		opentracinglog.String("namespace", b.nsMD.ID().String()),
+		opentracinglog.Int("limit", opts.Limit),
+		xopentracing.Time("query_start", opts.StartInclusive),
+		xopentracing.Time("query_end", opts.EndExclusive),
+		xopentracing.Time("block_start", b.StartTime()),
+		xopentracing.Time("block_end", b.EndTime()),
```
robskillington: nit: `queryStart` instead of `query_start`.

src/dbnode/storage/index.go
```diff
@@ -1136,14 +1189,25 @@ func (i *nsIndex) query(
...
 	cancellable *resource.CancellableLifetime,
 	block index.Block,
 	query index.Query,
 	opts index.QueryOptions,
 	state *asyncQueryExecState,
 	results index.BaseResults,
+	logFields []opentracinglog.Field,
 ) {
-	blockExhaustive, err := block.Query(cancellable, query, opts, results)
+	logFields = append(logFields, []opentracinglog.Field{
```
prateek: Don’t need to create the second slice when appending multiple values. ie can do this:

```
logFields = append(logFields, xopentracing.Time(...), xopentracing.Time(...))
```

src/dbnode/storage/index.go
```diff
@@ -1156,29 +1220,41 @@ func (i *nsIndex) execBlockQueryFn(
...
+	ctx context.Context,
 	cancellable *resource.CancellableLifetime,
 	block index.Block,
 	query index.Query,
 	opts index.QueryOptions,
 	state *asyncQueryExecState,
 	results index.BaseResults,
+	logFields []opentracinglog.Field,
 ) {
+	logFields = append(logFields, []opentracinglog.Field{
```
prateek: Here as well

src/dbnode/storage/index_block_test.go
```diff
@@ -36,12 +37,20 @@ import (
 	"github.com/m3db/m3/src/x/ident"
 	xtest "github.com/m3db/m3/src/x/test"
 	xtime "github.com/m3db/m3/src/x/time"
+	opentracing "github.com/opentracing/opentracing-go"
```
prateek: Import order

Issue #1732, opened 2019-06-13
Integrate block leaser interface with namespaceReaderManager


justinjc:
The namespaceReaderManager (in `namespace_readers.go`) contains a cache of `DataFileSetReader`s, which may become stale after a new fileset is written. We should incorporate block leaser logic into this in order to signify to the namespace reader manager of new filesets.

justinjc:
Closing this as a dupe of #1747, which better describes intended functionality. This ticket is more of an implementation-specific issue of #1747.

PR #1744, opened 2019-06-17
[DBNode] Use OpenLatestLease() in SeekManager and move CacheShardIndices() to namespace bootstrap

richardartoul:
This P.R calls `OpenLatestLease()` before opening a new seeker in the seek manager to ensure that the seek manager is always opening the latest volume for a given file. This is required for correct behavior of the cold write feature.

In addition, this change means that `CacheShardIndices` cannot be called before shards complete bootstrapping (since they cannot perform lease verification until they are bootstrapped). As a result, `CacheShardIndices` cannot be called in the F.S and peers bootstrapper as it is currently. Instead, this P.R moves the call to `CacheShardIndices` into the namespace `Bootstrap()` method immediately after all the shards are marked as bootstrapped (and as a result can perform lease verification).

### DIFF
--- a/src/dbnode/persist/fs/seek_manager.go
+++ b/src/dbnode/persist/fs/seek_manager.go
@@ -437,7 +437,7 @@ func (m *seekerManager) newOpenSeeker(
 		return nil, fmt.Errorf("err opening latest lease: %v", err)
 	}
 
-	// TODO(rartoul): Pass volume obtained from OpenLatestLease here.
+	// TODO_OUT_OF_ORDER_WRITES(rartoul): Pass volume obtained from OpenLatestLease here.
 	err = seeker.Open(m.namespace, shard, blockStart, resources)
 	m.putSeekerResources(resources)
 	if err != nil {
--- a/src/dbnode/persist/fs/seek_manager.go
+++ b/src/dbnode/persist/fs/seek_manager.go
@@ -22,6 +22,7 @@ package fs
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 	"time"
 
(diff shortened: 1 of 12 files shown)


src/dbnode/storage/block/lease.go
```diff
@@ -181,13 +180,6 @@ func (m *leaseManager) UpdateOpenLeases(
 func (m *leaseManager) SetLeaseVerifier(leaseVerifier LeaseVerifier) error {
 	m.Lock()
 	defer m.Unlock()
-
-	if m.verifier != nil {
```
robskillington: Just from curiosity what's the motivation to remove this check?

src/dbnode/storage/namespace.go
```diff
@@ -875,9 +875,26 @@ func (n *dbNamespace) Bootstrap(start time.Time, process bootstrap.Process) erro
...
+		// (and as a result no leases can be verified since the flush state is not yet known).
+		retriever, err := n.opts.DatabaseBlockRetrieverManager().Retriever(metadata)
+		if err != nil {
+			multiErr = multiErr.Add(err)
+		} else {
+			shardIDs := make([]uint32, 0, len(shards))
+			for _, shard := range shards {
+				shardIDs = append(shardIDs, shard.ID())
+			}
+			retriever.CacheShardIndices(shardIDs)
```
robskillington: Can we add a log just before this statement and just after?  This takes a long time sometimes and if we don't have the log anymore it will hurt operational visibility.

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -423,7 +424,20 @@ func (m *seekerManager) newOpenSeeker(
...
+	_, err = blm.OpenLatestLease(m, block.LeaseDescriptor{
+		Namespace:  m.namespace,
+		Shard:      shard,
+		BlockStart: blockStart,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("err opening latest lease: %v", err)
+	}
+
+	// TODO(rartoul): Pass volume obtained from OpenLatestLease here.
```
robskillington: Could you and justin perhaps use like a TODO_COLDFLUSH(...) or something tag so we can track all these short term things needing updates without forgetting about any one place?  I know not part of your change but on justin's PR there's places in the namespace reader cache where it could be very easy to forget about updating the volume index passed into the reader.

richardartoul: Yeah Justin actually has an issue filed for that: https://github.com/m3db/m3/issues/1732

That said I think `TODO_OUT_OF_ORDER_WRITES` (since thats what the issue tag is called) is a good idea. I'll use that here and I just filed this issue to track: #1746

src/dbnode/storage/block/lease.go
richardartoul: It needs to be called multiple times so that integration tests can open and close the database multiple times. The check was overly paranoid anyways so it was just easiest to remove it.

src/dbnode/storage/namespace.go
richardartoul: yes thank you for the reminder, I had meant to do it but forgot. Will add it now

PR #1745, opened 2019-06-18
[docker] dual-publish setcap; default raise limits

### LINKED ISSUE #1671: Setcap in docker build causes un-runnable container with docker defaults

schallert:
TODO:
- [x] docs

Fixes #1671

- Default to attempting to raise FD limits rather than opt in: raising
  the limits is benign if it fails.
- Dual-publish Docker images with a setcap'd binary for users running as
  non-root that want to raise limits.

```release-note
- Dual-publish Docker images with SYS_RESOURCE capabilities on the m3dbnode binary for users running as non-root with added capabilities that wish to raise process limits.

- Default to attempting to raise FD limits to max possible, with an environment variable to skip doing so.
```

```documentation-note
YES: Included
```

### DIFF
--- a/docker/images.json
+++ b/docker/images.json
@@ -1,15 +1,24 @@
 {
   "images": {
     "m3dbnode": {
+      "name": "m3dbnode",
       "dockerfile": "docker/m3dbnode/Dockerfile"
     },
+    "m3dbnode-setcap": {
+      "name": "m3dbnode",
+      "dockerfile": "docker/m3dbnode/Dockerfile",
+      "tag_suffix": "setcap"
+    },
     "m3coordinator": {
+      "name": "m3coordinator",
       "dockerfile": "docker/m3coordinator/Dockerfile"
     },
     "m3query": {
+      "name": "m3query",
       "dockerfile": "docker/m3query/Dockerfile"
     },
     "m3nsch": {
+      "name": "m3nsch",
       "dockerfile": "docker/m3nsch/Dockerfile"
     }
   }
--- a/docker/images.json
+++ b/docker/images.json
@@ -1,25 +1,29 @@
 {
   "images": {
-    "m3dbnode": {
-      "name": "m3dbnode",
-      "dockerfile": "docker/m3dbnode/Dockerfile"
-    },
-    "m3dbnode-setcap": {
-      "name": "m3dbnode",
-      "dockerfile": "docker/m3dbnode/Dockerfile",
-      "tag_suffix": "setcap"
+    "m3aggregator": {
+      "dockerfile": "docker/m3aggregator/Dockerfile",
+      "name": "m3aggregator"
     },
     "m3coordinator": {
-      "name": "m3coordinator",
-      "dockerfile": "docker/m3coordinator/Dockerfile"
+      "dockerfile": "docker/m3coordinator/Dockerfile",
+      "name": "m3coordinator"
     },
-    "m3query": {
-      "name": "m3query",
-      "dockerfile": "docker/m3query/Dockerfile"
+    "m3dbnode": {
+      "dockerfile": "docker/m3dbnode/Dockerfile",
+      "name": "m3dbnode"
+    },
+    "m3dbnode-setcap": {
+      "dockerfile": "docker/m3dbnode/Dockerfile-setcap",
+      "name": "m3dbnode",
+      "tag_suffix": "setcap"
     },
     "m3nsch": {
-      "name": "m3nsch",
-      "dockerfile": "docker/m3nsch/Dockerfile"
+      "dockerfile": "docker/m3nsch/Dockerfile",
+      "name": "m3nsch"
+    },
+    "m3query": {
+      "dockerfile": "docker/m3query/Dockerfile",
+      "name": "m3query"
     }
   }
 }
--- a/docker/m3dbnode/Dockerfile-setcap
+++ b/docker/m3dbnode/Dockerfile-setcap
@@ -1,0 +1,34 @@
+# stage 1: build
+FROM golang:1.12-alpine AS builder
+LABEL maintainer="The M3DB Authors <m3db@googlegroups.com>"
+
+# Install Glide
+RUN apk add --update glide git make bash
+
+# Add source code
+RUN mkdir -p /go/src/github.com/m3db/m3
+ADD . /go/src/github.com/m3db/m3
+
+# Build m3dbnode binary
+RUN cd /go/src/github.com/m3db/m3/ && \
+    git submodule update --init      && \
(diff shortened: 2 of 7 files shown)


docker/images.json
```diff
@@ -1,15 +1,24 @@
 {
```
robskillington: Could we also add `m3aggregator` to this btw?  Noticed we don't have an image (can do this in followup I suppose though).

docker/images.json
```diff
@@ -1,15 +1,24 @@
...
   "images": {
     "m3dbnode": {
+      "name": "m3dbnode",
       "dockerfile": "docker/m3dbnode/Dockerfile"
     },
+    "m3dbnode-setcap": {
+      "name": "m3dbnode",
+      "dockerfile": "docker/m3dbnode/Dockerfile",
+      "tag_suffix": "setcap"
```
robskillington: Will it add a "-"  automatically?  i.e. 0.10.2-setcap or will it be 0.10.2setcap ?

docker/images.json
```diff
@@ -1,15 +1,24 @@
 {
   "images": {
     "m3dbnode": {
+      "name": "m3dbnode",
       "dockerfile": "docker/m3dbnode/Dockerfile"
     },
+    "m3dbnode-setcap": {
+      "name": "m3dbnode",
+      "dockerfile": "docker/m3dbnode/Dockerfile",
```
robskillington: Hm don't we need a new dockerfile that actually calls setcap?

docker/images.json
schallert: Will do

docker/images.json
schallert: 🤦‍♂ will add

docker/images.json
schallert: Yeah, in the builder script we have
```shell
TAG="${TAG}-${TAG_SUFFIX}"
```

robskillington: Nice.

docker/m3dbnode/Dockerfile-setcap
```diff
@@ -0,0 +1,34 @@
...
+EXPOSE 2379/tcp 2380/tcp 7201/tcp 7203/tcp 9000-9004/tcp
+
+COPY --from=builder /go/src/github.com/m3db/m3/src/dbnode/config/m3dbnode-local-etcd.yml /etc/m3dbnode/m3dbnode.yml
+COPY --from=builder /go/src/github.com/m3db/m3/bin/m3dbnode \
+  /go/src/github.com/m3db/m3/scripts/m3dbnode_bootstrapped.sh \
+  /bin/
+
+# Use setcap to set +e "effective" and +p "permitted" to adjust the SYS_RESOURCE
+# so the process can raise the hard file limit with setrlimit.
+RUN apk add --no-cache curl jq libcap && \
```
robskillington: Do we need curl and jq?  I'm fine to leave it out and add on login (i.e. just apk add when you start a shell session), or if you feel we should (which I'm fine with too), should we putting it in the other base image too?

schallert: We only need it for M3DB for the `m3dbnode_bootstrapped.sh` script. When we deprecate that though we can just remove this entirely

docs/operational_guide/kernel_configuration.md
```diff
@@ -1,10 +1,26 @@
...
+Docker & Kernel Configuration
 ====================
 
 This document lists the Kernel tweaks M3DB needs to run well. If you are running on Kubernetes, you may use our
 `sysctl-setter` [DaemonSet](https://github.com/m3db/m3/blob/master/kube/sysctl-daemonset.yaml) that will set these
 values for you. Please read the comment in that manifest to understand the implications of applying it.
 
+## Running with Docker
+
+If running M3DB under Docker it is recommended to give M3DB the `SYS_RESOURCE` capability so that it may raise its file
```
mway: > If running M3DB under Docker it is recommended to give M3DB the `SYS_RESOURCE` capability so that it may raise its file limits. If running directly with Docker this can be accomplished using:

slight tweak:

> When running M3DB inside Docker, it is recommended to add the `SYS_RESOURCE` capability to the container (using the `--cap-add` argument to `docker run`) so that it can raise its file limits:

docs/operational_guide/kernel_configuration.md
```diff
@@ -1,10 +1,26 @@
...
+limits. If running directly with Docker this can be accomplished using:
+
+```
+docker run --cap-add SYS_RESOURCE quay.io/m3/m3dbnode:latest
+```
+
+If you wish to run M3DB as a non-root user, you will need to use our `setcap` images:
```
mway: nit: no first or second person pronouns

> If M3DB is being run as a non-root user, M3's `setcap` images are required:

docs/operational_guide/kernel_configuration.md
```diff
@@ -62,3 +78,10 @@ Also note that systemd has a `system.conf` file and a `user.conf` file which may
 Be sure to check that those files aren't configured with values lower than the value you configure at the service level.
 
 Before running the process make sure the limits are set, if running manually you can raise the limit for the current user with `ulimit -n 3000000`.
+
+## Automatic Limit Raising
+
+When M3DB first starts up it will attempt to raise its open file limit to the current value of `fs.nr_open`. This is a
```
mway: minor nits:

> During startup, M3DB will attempt to raise [...]

> This is a benign operation; if it fails, M3DB will [...]

PR #1749, opened 2019-06-18
[DBNode] - Add documentation for using M3DB as a generic TSDB / protobuf feature

richardartoul:
This P.R adds documentation explaining the basics of how to use M3DB as a general purpose DB using the protobuf / complex types feature.

It also includes a sample Go application that interacts with M3DB directly.


docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
+# Using M3DB as a generic time series database
+
+## Overview
+
+M3 has native integrations that make it particularly easy to use it as a metrics storage for [Prometheus](../integrations/prometheus.md) and [Graphite](../integrations/graphite.md). However, M3DB can also be used as a stand alone time series database.
```
robskillington: Minor suggestions:
```
s/as a metrics storage for/as a metrics storage and monitoring platform/
```

And perhaps to make a stronger point, instead of the "However, " prefix just go straight to:
```
M3DB can also be used as a general purpose distributed time series database by itself.
```

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+M3DB's inverted index supports term (exact match) and regular expression queries over all tag values, and individual tag queries can be arbitrarily combined using `AND`, `OR`, and `NOT` operators.
+
+For example, imagine you were developing an application to track a fleet of vehicles. One way to structure your time series could be as follows:
+
+| Time Series ID | "type" tag value | "city" tag value          |
+|----------------|------------------|---------------------------|
+| vehicle_id_1   | sedan            | san_francisco             |
+| vehicle_id_2   | bicycle          | san_francisco             |
+| vehicle_id_3   | scooter          | chicago_region_0          |
+| vehicle_id_4   | scooter          | chicago_region_1          |
```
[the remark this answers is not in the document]
robskillington: Agree that  probably reads a little cleaner.

Also do you maybe want to just add a third tag for region/hexagon ID or something instead of using `region_0` and `region_` to differentiate the two chicago areas?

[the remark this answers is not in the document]
richardartoul: Yeah sure I was just quickly trying to find an excuse to need a regex query, your suggestion makes more sense. I’ll invert the table as Simon demonstrated too

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+
+message VehicleLocation {
+  double latitude = 1;
+  double longitude = 2;
+  double fuel_percent = 3;
+  string status = 4;
+}
+```
+
+While M3DB strives to support the entire [proto3 language spec](https://developers.google.com/protocol-buffers/docs/proto3), only the following features are currently supported:
```
mway: might be better to link the encoding docs here rather than below - your call

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+message VehicleLocation {
+  double latitude = 1;
+  double longitude = 2;
+  double fuel_percent = 3;
+  string status = 4;
+  map<string, string> attributes = 5;
+}
+```
+
+This isn't to say that the schema above shouldn't be used, but users should take into account that any data stored in the `attributes` map will be held in memory and eventually stored on disk with no compression.
```
mway: nit: consider something like:

> While the latter schema is valid, the `attributes` field will not be compressed; users should weigh the tradeoffs between more expressive schema and better compression for each use case.

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+
+For more details on the compression scheme and its limitations, review [the documentation for M3DB's compressed protobuf encoding](https://github.com/m3db/m3/blob/master/src/dbnode/encoding/proto/docs/encoding.md).
+
+
+### Getting Started
+
+#### M3DB setup
+
+For more advanced setups, its best to follow the guide's on how to configure an [M3DB cluster manually](./cluster_hard_way.md) or [using Kubernetes](./kubernetes.md). However, this tutorial will walk you through configuring a single node setup locally for development.
```
mway: nit: "it's", "guides"

super nit: linked words, e.g.

> [...] how to configure an M3DB cluster [manually](./cluster_hard_way.md) or [using Kubernetes](./kubernetes.md).

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+docker pull quay.io/m3db/m3dbnode:latest
+```
+
+Next, run the following command to start the M3DB container:
+
+```
+docker run -p 7201:7201 -p 7203:7203 -p 9000:9000 -p 9001:9001 -p 9002:9002 -p 9003:9003 -p 9004:9004 -p 2379:2379 --name m3db -v $(pwd)/m3db_data:/var/lib/m3db -v $(pwd)/src/dbnode/config/m3dbnode-local-etcd-proto.yml:/etc/m3dbnode/m3dbnode.yml -v <PATH_TO_SCHEMA_PROTOBUF_FILE>:/etc/m3dbnode/default_schema.proto quay.io/m3db/m3dbnode:latest
+```
+
+Breaking that down:
```
mway: do we need to explain how docker works? can we just say "See `docker run --help` for more information."?

it might be better to just call out the specific flag values, e.g.:

- Creating a bind mount (`$(pwd)/m3db_data:/var/lib/m3db`) enables M3DB data to persist between container restarts.
- Mounting the configuration file (`m3dbnode.yml`) enables configuration updates by restarting the container (rather than rebuilding it). [other info about the config file]
- Similar to configuration, mounting the schema file (`default_schema.proto`) enables schema updates by restarting the container. [other info about the schema file]

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+}'
+```
+
+Note that the `retentionTime` is set artificially low to conserve resources.
+
+After a few moments, the M3DB container should finish bootstrapping. At this point it should be ready to serve write and read queries.
+
+#### Clients
+
+M3DB only has a Go client and this is unlikely to change in the future due to the fact that the client is "fat" and contains a lot of logic that would be difficult to port to other languages.
```
mway: nit:

> Note: M3DB only has a Go client; this is unlikely to change [...] contains a substantial amount of logic [...]

(it's more like a n.b. than an intro)

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+
+Users interested in interacting with M3DB directly from Go applications can review [this runnable example](https://github.com/m3db/m3/tree/master/examples/dbnode/proto_client) to get an understanding of how to interact with M3DB in Go. Note that the example above uses the same `default` namespace and `VehicleLocation` schema used throughout this document so it can be run directly against an M3DB docker container setup using the "M3DB setup" instructions above.
```
mway: nit: s/review/reference/

examples/dbnode/proto_client/README.md
```diff
@@ -0,0 +1,5 @@
+# Protobuf Client Example
+
+1. Setup an M3DB container as described in the `use_as_tsdb.md` file.
```
mway: should we link instead of using preformatted text?

examples/dbnode/proto_client/README.md
```diff
@@ -0,0 +1,5 @@
...
+2. Modify `config.yaml` with any changes you've made to default configuration.
+3. Execute `go run main.go -f config.yaml`
```
mway: i don't think m3db is using inotify/fsnotify for config, is it? if not, should probably call out that the config needs to be modified prior to running the container, else disappointment/confusion will ensue.

src/dbnode/encoding/proto/docs/encoding.md
```diff
@@ -10,7 +10,7 @@ All compression is performed in a streaming manner such that the encoded stream
 1. Lossless compression.
 1. Compression of Protobuf message timestamps using [Gorilla-style delta-of-delta encoding](https://www.vldb.org/pvldb/vol8/p1816-teller.pdf).
 2. Compression of Protobuf message streams that match a provided schema by using different forms of compression for each field based on its type.
-4. Changing Protobuf message schemas mid-stream (pending).
+4. Changing Protobuf message schemas mid-stream.
```
mway: 🎉

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+| vehicle_id_3   | scooter          | chicago_region_0          |
+| vehicle_id_4   | scooter          | chicago_region_1          |
+
+This would allow users to issue queries that answer questions like:
+
+1. "What time series IDs exist for any vehicle type operating in San Francisco?"
+2. "What time series IDs exist for scooters that are NOT operating in Chicago?"
+3. "What time series IDs exist where the "city" tag matches the regular expression: `chicago_region_[0-9]+`"
+
+TODO(rartoul): Discuss the ability to perform limited amounts of aggregation queries here as well.
```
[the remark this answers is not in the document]
richardartoul: Generally I agree although here I'd like to make whats missing from the docs explicit so people at least know to ask

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+
+TODO(rartoul): Discuss ID / tags mutability.
+
+### Data Points
+
+Each time series in M3DB stores data as a stream of data points in the form of `<timestamp, value>` tuples. Timestamp resolution can be as granular as individual nanoseconds.
```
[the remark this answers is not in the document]
richardartoul: I link to it later

docs/how_to/use_as_tsdb.md
```diff
@@ -0,0 +1,155 @@
...
+
+Every M3DB namespace can be configured with a schema where the schema is a protobuf message that each value in time series must conform to.
+
+For example, continuing with the vehicle fleet tracking example introduced earlier, a schema might look as follows:
+
+```protobuf
+syntax = "proto3";
+
+message VehicleLocation {
+  double latitude = 1;
```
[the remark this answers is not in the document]
richardartoul: sure, just trying to keep this really simple though and its pretty common to store lat/lng as floats (we do)

docs/how_to/use_as_tsdb.md
richardartoul: I'll do both

docs/how_to/use_as_tsdb.md
richardartoul: I prefer to err on the side of verbosity here a little bit, but some of your clarifications are good I'll add them

PR #1750, opened 2019-06-19
[DBNode][WIP] Dont block database initialization on namespace initialization

richardartoul:
This P.R allows M3DB to come up and mark itself as bootstrapped without any configured namespaces.

### DIFF
--- a/src/dbnode/namespace/types.go
+++ b/src/dbnode/namespace/types.go
@@ -25,9 +25,9 @@ import (
 
 	"github.com/m3db/m3/src/cluster/client"
 	"github.com/m3db/m3/src/dbnode/retention"
+	xclose "github.com/m3db/m3/src/x/close"
 	"github.com/m3db/m3/src/x/ident"
 	"github.com/m3db/m3/src/x/instrument"
-	xclose "github.com/m3db/m3/src/x/close"
 )
 
 // Options controls namespace behavior
@@ -220,6 +220,11 @@ type Watch interface {
 type Registry interface {
 	// Watch for the Registry changes
 	Watch() (Watch, error)
+
+	// ForceGet bypasses the watch and issues a get request against KV directly.
+	// This can be used in situations where the caller wants to get the current
+	// value, if any, without potentially waiting for an initial value to be set.
+	ForceGet() (Map, bool, error)
 
 	// Close closes the registry
 	Close() error
--- a/src/dbnode/storage/database.go
+++ b/src/dbnode/storage/database.go
@@ -200,13 +200,26 @@ func NewDatabase(
 		return nil, err
 	}
 
-	// Wait till first namespaces value is received and set the value.
-	// Its important that this happens before the mediator is started to prevent
-	// a race condition where the namespaces haven't been initialized yet and
-	// GetOwnedNamespaces() returns an empty slice which makes the cleanup logic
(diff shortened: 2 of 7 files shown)


prateek:
Does this differentiate issues with delayed namespace configuration vs new cluster creation? Eg if etcd is down or takes a while to push an existing ns config, we shouldn’t let the dbnode bootstrap.

richardartoul:
@prateek yeah that’s the idea behind the “ForceGet” method. Basically want to go straight to etcd and only if etcd is available and concretely tells you there are no namespaces can you “skip” past waiting for an initial value, otherwise you blocking waiting for an initial value like we do now.

robskillington:
Hey this is much better - do you mind outlining how (2) is possible with this change?

src/dbnode/namespace/types.go
```diff
@@ -221,6 +221,11 @@ type Registry interface {
 	// Watch for the Registry changes
 	Watch() (Watch, error)
 
+	// ForceGet bypasses the watch and issues a get request against KV directly.
```
prateek: nit: “against the backing store.”

The interface has the cfg implementation too

src/dbnode/storage/database.go
```diff
@@ -200,13 +200,26 @@ func NewDatabase(
...
-	// Its important that this happens before the mediator is started to prevent
-	// a race condition where the namespaces haven't been initialized yet and
-	// GetOwnedNamespaces() returns an empty slice which makes the cleanup logic
-	// in the background Tick think it can clean up files that it shouldn't.
-	logger.Info("resolving namespaces with namespace watch")
-	<-watch.C()
+	var blockOnNamespaceWatch = true
+	_, namespacesExist, err := nsReg.ForceGet()
+	if err == nil && namespacesExist == false {
+		// If it can be confirmed that no existing namespaces exist (as verified
```
prateek: Can simplify this to:

```
// ... forceGet
blockOnNamespaceWatch := err == nil && !namespacesExist
```

prateek:
@richardartoul cool, makes sense

richardartoul:
@robskillington It helps with #2 because what happens currently in the worst case is:

```
1. Wait for placement
2. Placement is set
3. Wait for namespace
4. Namespace with long retention is set
5. Node skips first flush (this is a separate issue which I'll fix in a separate P.R, it happens because of the guard rails we put in place last time we had that major bootstrapping / data loss bug but its no longer needed now that justin has refactored the buffer code)
6. Node does snapshot (and it has to snapshot ALL blocks
7. At this point node MAY mark shards available depending on when the tick started. If not, it has do go through the flush/snapshotting process one more time
8. Node marks shards as available
```

After this P.R lands:

```
1. Wait for placement
2. Placement is set
3. Node bootstraps
4. Node marks shards as available (get here with in seconds)
5. Namespace is added
6. Namespace is bootstrapped (extremely quick because there is no data)
```

So now we're up within a few seconds as long as you don't add the namespaces before you create a placement. The downside of course is that if writing out the initial set of files still take forever, you could end up holding data in memory for longer than you expected, but I'm not too worried about that because part of the reason writing out those files takes so long is that sometimes we have to do it twice because the first flush gets skipped (due to this code which I will delete in a separate P.R: https://github.com/m3db/m3/blob/master/src/dbnode/storage/namespace.go#L969)

PR #1755, opened 2019-06-20
[metrics] rules: reject timertype tag in filters

schallert:
**What this PR does / why we need it**: Make rule validator more robust for a common gotcha.

Fixes #

**Special notes for your reviewer**: `timertype` is added at the aggregation tier, not at the client layer.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/metrics/rules/validator/options.go
+++ b/src/metrics/rules/validator/options.go
@@ -21,7 +21,6 @@
 package validator
 
 import (
-	"errors"
 	"fmt"
 	"strconv"
 	"strings"
@@ -41,8 +40,6 @@ const (
 
 	// By default we allow at most one level of rollup in a pipeline.
 	defaultMaxRollupLevels = 1
-
-	tagTimerType = "timertype"
 )
 
 // MetricTypesFn determines the possible metric types based on a set of tag based filters.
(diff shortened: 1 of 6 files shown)


src/metrics/rules/validator/options.go
```diff
@@ -39,6 +40,8 @@ const (
 
 	// By default we allow at most one level of rollup in a pipeline.
 	defaultMaxRollupLevels = 1
+
+	tagTimerType = "timertype"
```
benraskin92: Just confirming it's `timertype` and not `timerType`, right?

src/metrics/rules/validator/options.go
```diff
@@ -267,6 +275,13 @@ func (o *options) CheckInvalidCharactersForTagName(tagName string) error {
 	return validateChars(tagName, o.tagNameInvalidChars)
 }
 
+func (o *options) CheckTimertypeFilterForTagName(tagName string) error {
+	if tagName == tagTimerType {
```
benraskin92: worth adding a `strings.ToLower(tagName)` here?

src/metrics/rules/validator/options.go
```diff
@@ -267,6 +275,13 @@ func (o *options) CheckInvalidCharactersForTagName(tagName string) error {
...
+		return errors.New("tag name cannot be timertype; timertype is not added at the client layer")
+	}
```
benraskin92: nit: newline

src/metrics/rules/validator/options.go
schallert: Yup

cw9: Ideally this can be fed from config files, right now `timertype` is not known to the system, and users have the ability to config the tag name for aggregation types it does not have to be `timertype`, we even once considered to use `aggType` so it can be applied to all metric types.

schallert: Will change this to be config driven

PR #1756, opened 2019-06-20
[dbnode] Add Gauges indicating bootstrappers order

### LINKED ISSUE #1142: [dbnode] Emit a gauge with currently configured bootstrapper configuration
And further add an alert to ensure we know when nodes are using an un-expected configuration.

obitech:

Fixes #1142 

Emits a Gauge for each bootstrapper, indicating its order:

```
bootstrap_bootstrappers_filesystem 0
bootstrap_bootstrappers_peers 1
...
```

Initially, the values will be set to `cfg.Bootstrap.Bootstrappers` but will be updated accordingly in the `kvWatchBootstrappers` function.

I initially tried to indicate the order as labels under a single Gauge as in:

```
bootstrap_bootstrappers{bootstrapper0="filesystem", bootstrapper1="peers", ...} 1
```

but just updating labels for a metric is not possible since Prometheus metrics are immutable after they get scraped. This would just emit a second Gauge with the same name but different labels.

```release-note
NONE
```

```documentation-note
NONE
```


obitech:
@richardartoul

richardartoul:
@obitech Admittedly I'm not a Prometheus export, but I would have thought that changing the labels would create a new gauge which would have separate values.

I.E `bootstrapper0=filesystem bootstrapper1=commitlog` would be one time series and `bootstrapper0=filesystem bootstrapper1=peers` would be another which should be fine.

@prateek I think you were the one who really wanted this feature. Any thoughts on structuring the labels?

richardartoul:
also @obitech thanks so much for your contribution :)

obitech:
@richardartoul 
> I.E bootstrapper0=filesystem bootstrapper1=commitlog would be one time series and bootstrapper0=filesystem bootstrapper1=peers would be another which should be fine.

Yes this is how I tried it first. I can implement it like that, no worries. @prateek let me know how you would like to see it!

prateek:
@obitech thank you for the dope work!

I agree with @richardartoul, would be better to have all the information contained in a single metric. One additional thing to consider is discoverability, like you probably want all the series’ to have the same name and be differentiated only by a tag/label. Will make it easier to access the information at query time that way. 

A little concerned about prom metric lifecycles - iirc it’s an error to re-register an existing metric. So one test that would be good to write is changing the bootstrapper config from default to something else and then back to default. And ensuring you see all three states.

obitech:
@prateek got it, so I'll just use a single Gauge `bootstrap_bootstrappers 1` with tags indicating the order.

Regarding the metric lifecycle - I manually tested registering the same label twice but with different labels. This seemed to be no problem as I didn't get any errors - I just had two time series with the same name but different labels. With the Prometheus client it's recommended to unregister metrics first but I'm not sure how to do this with tally. I guess I'll test it with just creating new subscopes and see how it goes.

obitech:
With 78f24413e5ffd793147542b3255ae16a79c92cc2 I changed it to a single Gauge. Now if bootstrappers get updated, a new Gauge will be emitted with the same name but updated labels and value of `1`. The old Gauge will still exist but the value is set to `0`. I've also added some tests in `server_test.go`.

I didn't know how to change bootstrappers during runtime so I did some hacky manual testing:

```go
// server.go
bsGauge := emitBootstrappersGauge(scope, cfg.Bootstrap.Bootstrappers)
bsGauge.Update(1)

bsGauge.Update(0)
bsGauge1 := emitBootstrappersGauge(scope, []string{"uninitialized_topology", "commitlog", "peers", "filesystem"})
bsGauge1.Update(1)

bsGauge.Update(1)
bsGauge1.Update(0)
```

Which will yield the following metrics in Prometheus:

```
bootstrapper_bootstrappers{bootstrapper0="filesystem",bootstrapper1="peers",bootstrapper2="commitlog",bootstrapper3="uninitialized_topology",instance="m3db_seed:9004",job="dbnode"}	1
bootstrapper_bootstrappers{bootstrapper0="uninitialized_topology",bootstrapper1="commitlog",bootstrapper2="peers",bootstrapper3="filesystem",instance="m3db_seed:9004",job="dbnode"}	0
```

robskillington:
This is looking good.

Unfortunately though you will have to emit the gauge periodically (say every minute or every 10 seconds), otherwise depending on your lookback period when viewing graphs you won't see a value at all if the bootstrapper hasn't changed.

So perhaps we create a goroutine that emits it every so often (and then is closed by a close channel or similar when server closes)?

src/dbnode/server/server.go
```diff
@@ -1511,3 +1519,13 @@ func (t *topoMapProvider) TopologyMap() (topology.Map, error) {
 
 	return t.t.Get(), nil
 }
+
+func emitBootstrappersGauge(scope tally.Scope, bs []string) tally.Gauge {
```
richardartoul: Yeah as @robskillington was saying we'll need to emit this regularly. You may want to just create a simple type thats something like:

```golang
type bootstrapGaugeEmitter struct {
    sync.Mutex
    running bool
    doneCh chan struct{}
    bootstrappers []string
}

func (b *bootstrapGaugeEmitter) Start() {
    b.Lock()
    defer b.Unlock()
    if b.running {
        return
    }

    b.doneCh = make(chan struct{})
    b.running = true
    go func() {
        for {
            select {
            case <-b.doneCh:
                return
            default:
                 // All the logic for emitting the gauge. Will need to acquire lock to get most up to date bootstrappers.
            }
        }
    }()
}

func (b *bootstrapGaugeEmitter) Stop() {
    b.Lock()
    defer b.Unlock()
    if !b.running {
        return
    }
    close(b.doneCh)
}

func (b *bootstrapGaugeEmitter) Update(bootstrappers []string) {
    b.Lock()
    defer b.Unlock()
    b.bootstrappers = bootstrappers
}
```

I hacked that up in the browser so there may be some mistakes but you get the idea

obitech: Awesome, thanks for the pointers! But why the lock, can't we just use a different channel where we pass the bootstrappers to?

richardartoul: Sure you could implement this with a channel too, I just work with locks a lot so maybe that felt simpler to me. Whatever works.

obitech: So I tried it with a second channel first but ultimately went with your suggestion (and built upon it) because I liked it better. I also moved the type to `x/instrument/metrics.go` but I'm not 100% certain about the location & naming. Do let me know if I should move it somewhere else and/or rename the file or such.

src/x/instrument/metrics.go
```diff
@@ -0,0 +1,122 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package instrument
```
richardartoul: So the instrument package is meant to be much more generic than anything DB related (we use it in our aggregation tier, query tier, random toolings, etc etc).

That said I could see us having similar needs in the future so if we're gonna put this in this package I think it needs to be something like `StringListEmitter` and the file name should be something like `string_list_emitter.go` or something like that.

src/x/instrument/metrics.go
```diff
@@ -0,0 +1,122 @@
...
+	"sync"
+	"time"
+
+	"github.com/uber-go/tally"
+)
+
+const (
+	// wait defines the time to wait between emitting the value of the Gauge
+	// again.
+	wait = 10 * time.Second
```
richardartoul: This is a package level variable so we probably want a slightly more specific name like `stringListEmitterInterval` or something like that. Might also want it to be configurable in the future but yeah we can leave that out for now

src/x/instrument/metrics.go
```diff
@@ -0,0 +1,122 @@
...
+	running bool
+	doneCh  chan bool
+	scope   tally.Scope
+	gauge   tally.Gauge
+}
+
+// NewBootstrapGaugeEmitter returns a BootstrapGaugeEmitter with Gauge that is
+// created from the passed scope and tagged according to the passed
+// bootstrappers. The value of the Gauge will be set to 1 on initialization.
+func NewBootstrapGaugeEmitter(scope tally.Scope, bs []string) *BootstrapGaugeEmitter {
```
richardartoul: I'd probably leave the `[]string` out of the initializer and just always use the `Update` API

src/x/instrument/metrics.go
```diff
@@ -0,0 +1,122 @@
...
+	g := scope.Tagged(tags).Gauge("bootstrapper_bootstrappers")
+	g.Update(1)
+
+	return g
+}
+
+// Start starts a goroutine that continuously emits the value of the Gauge.
+func (bge *BootstrapGaugeEmitter) Start() error {
+	if bge.running {
+		return fmt.Errorf("already running")
```
richardartoul: Usually errors like this that don't change we put at the top of a file in a var block so it would be something like:

```golang
var (
    errStringListEmitterAlreadyRunning = errors.New("cant start stringListEmitter because it is already running")
)
```

src/x/instrument/metrics.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+	g := scope.Tagged(tags).Gauge("bootstrapper_bootstrappers")
+	g.Update(1)
+
+	return g
+}
+
+// Start starts a goroutine that continuously emits the value of the Gauge.
+func (bge *BootstrapGaugeEmitter) Start() error {
+	if bge.running {
```
richardartoul: You need to be holding the lock while you perform this operation otherwises its racey:

```
1. Goroutine 1 calls Start() and bge.running == false
// Goroutine1 gets descheduled
2. Goroutine 2 calls Start() and bge.running == false
// Now both Goroutine1 and 2 will will run through Start() in its entirety
```

richardartoul: I'd probably just do:

```golang
bge.Lock()
defer bge.Unlock()

// Rest of the function
...
```

src/x/instrument/metrics.go
```diff
@@ -0,0 +1,122 @@
...
+
+// UpdateBootstrappers updates the Gauge with new tags according to passed
+// bootstrappers while setting the old Gauge to 0 first. This will emit
+// new metric with the same metric name but different tags.
+func (bge *BootstrapGaugeEmitter) UpdateBootstrappers(bs []string) error {
+	if !bge.running {
+		return fmt.Errorf("not started yet")
+	}
+
+	bge.Lock()
```
richardartoul: Need to move this and the defer to the top of the function otherwise the `bge.running` check is racey (similar to my comment above about Start()

src/x/instrument/metrics.go
```diff
@@ -0,0 +1,122 @@
...
+	return nil
+}
+
+// Close stops emitting the Gauge.
+func (bge *BootstrapGaugeEmitter) Close() error {
+	if !bge.running {
+		return fmt.Errorf("not started yet")
+	}
+
+	bge.Lock()
```
richardartoul: same comment

src/x/instrument/metrics_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"fmt"
+	"reflect"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/uber-go/tally"
+)
+
+var bootstrappers = [][]string{
+	[]string{
```
richardartoul: Can you add one with an empty slice too

src/dbnode/server/server.go
```diff
@@ -768,8 +784,14 @@ func Run(runOpts RunOptions) {
 	closeTimeout := serverGracefulCloseTimeout
 	select {
 	case <-closedCh:
+		if err := bsGauge.Close(); err != nil {
```
richardartoul: Is this in the same function as when `Start()` is called? If so I think we *should* just be able to call

```golang
defer bsGauge.Close()
```

right after calling start and it won't run until the database is stopped anyways because the main function never returns until the database shuts down (due to blocking on the `xos.WaitForInterrupt()` call). If you look throughout that function you'll see a few other examples of using `defer` for cleanup after the database node shuts down

src/x/instrument/metrics.go
obitech: That makes sense! I have renamed it.

src/x/instrument/metrics.go
obitech: Fixed

src/x/instrument/metrics.go
obitech: Sure! It's passed as an argument for `Start()` and `UpdateStringList()` now.

src/x/instrument/metrics.go
obitech: Fixed

src/x/instrument/metrics.go
obitech: Fixed

src/x/instrument/metrics.go
obitech: Fixed

src/x/instrument/metrics.go
obitech: Fixed

src/x/instrument/metrics_test.go
obitech: Done

src/dbnode/server/server.go
obitech: Done

src/x/instrument/string_list_emitter.go
```diff
@@ -0,0 +1,141 @@
...
+
+const (
+	// stringListEmitterWaitInterval defines the time to wait between emitting the value of the Gauge
+	// again.
+	stringListEmitterWaitInterval = 10 * time.Second
+)
+
+var (
+	errStringListEmitterAlreadyRunning = errors.New("can't start StringListEmitter: already running")
+	errStringListEmitterNotStarted     = errors.New("StringListEmitter is not running")
```
robskillington: nit: We try to stay strictly to the Effective Go recommendation of naming all logs and errors with letters beginning in lower case.

Also I usually just name things in separated words, even it relates to a symbol, it's still searchable in the code, etc from a log line.

i.e.
```go
		errStringListEmitterAlreadyRunning = errors.New("string list emitter: already running")
	errStringListEmitterNotStarted     = errors.New("string list emitter: not running")
```

src/dbnode/server/server.go
```diff
@@ -682,6 +696,13 @@ func Run(runOpts RunOptions) {
 			}
 
 			bs.SetBootstrapperProvider(updated.BootstrapperProvider())
+
+			if err := bsGauge.UpdateStringList(bootstrappers); err != nil {
+				logger.Error("unable to update bootstrap gauge with new bootstrappers",
+					zap.String("bootstrappers", fmt.Sprintf("%#v", bootstrappers)),
```
robskillington: You can use `zap.Strings("foo", values)` here instead of a sprintf I believe?

src/dbnode/server/server.go
```diff
@@ -666,6 +666,20 @@ func Run(runOpts RunOptions) {
 
 	opts = opts.SetBootstrapProcessProvider(bs)
 	timeout := bootstrapConfigInitTimeout
+
+	bsGauge := instrument.NewStringListEmitter(scope, "bootstrapper_bootstrappers", "bootstrapper")
```
robskillington: Maybe just call it "bootstrappers" instead of "bootstrapper_bootstrappers" to make it a little more simple?

src/x/instrument/string_list_emitter.go
```diff
@@ -0,0 +1,141 @@
...
+// new metric with the same name but different tags.
+func (bge *StringListEmitter) UpdateStringList(sl []string) error {
+	bge.Lock()
+	defer bge.Unlock()
+
+	if !bge.running {
+		return errStringListEmitterNotStarted
+	}
+
+	if bge.gauge != nil {
```
robskillington: Hm, if it's already running it's guaranteed to not be nil yeah?  Might be worth doing two things:
1. In the constructor make it a noop-gauge, instead of nil (that way you can safely inc it, etc even before running, just for safety)
2. Remove this nil check here?

Here's how to make a no-op gauge in the constructor:
```go
gauge := tally.NoopScope.Gauge("blackhole") 
```

src/x/instrument/string_list_emitter.go
obitech: Good point, fixed it.

src/dbnode/server/server.go
obitech: Fixed

src/dbnode/server/server.go
obitech: Yeah I agree. The only reason why I had it `bootstrapper_bootstrappers` in the first place was because there is another metric which is starting with `bootstrapper_` (forgot which) so I thought I could piggy-back on that namespace. I changed it to just `bootstrappers` now.

src/x/instrument/string_list_emitter.go
obitech: Ah I didn't know about the noop-gauge! Fixed it.

PR #1757, opened 2019-06-20
Automatically create folder for KV cache

### LINKED ISSUE #1751: M3DB should automatically create the folder for KV cache
KV values are stored in a local cache directory. Currently M3DB expects this folder to exist and will not create it if it doesn't which means if users don't explicitly create the directory their etcd value are not being cached.

obitech:

Fixes #1751 

```
NONE
```

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/cluster/client/etcd/config.go
+++ b/src/cluster/client/etcd/config.go
@@ -21,6 +21,7 @@
 package etcd
 
 import (
+	"os"
 	"time"
 
 	"github.com/m3db/m3/src/cluster/client"
@@ -95,6 +96,7 @@ type Configuration struct {
 	ETCDClusters      []ClusterConfig        `yaml:"etcdClusters"`
 	SDConfig          services.Configuration `yaml:"m3sd"`
 	WatchWithRevision int64                  `yaml:"watchWithRevision"`
+	NewDirectoryMode  os.FileMode            `yaml:"newDirectoryMode"`
 }
 
 // NewClient creates a new config service client.
@@ -111,7 +113,8 @@ func (cfg Configuration) NewOptions() Options {
 		SetCacheDir(cfg.CacheDir).
 		SetClusters(cfg.etcdClusters()).
 		SetServicesOptions(cfg.SDConfig.NewOptions()).
-		SetWatchWithRevision(cfg.WatchWithRevision)
+		SetWatchWithRevision(cfg.WatchWithRevision).
+		SetNewDirectoryMode(cfg.NewDirectoryMode)
 }
 
 func (cfg Configuration) etcdClusters() []Cluster {
--- a/src/cluster/client/etcd/config.go
+++ b/src/cluster/client/etcd/config.go
@@ -96,7 +96,7 @@ type Configuration struct {
 	ETCDClusters      []ClusterConfig        `yaml:"etcdClusters"`
 	SDConfig          services.Configuration `yaml:"m3sd"`
 	WatchWithRevision int64                  `yaml:"watchWithRevision"`
-	NewDirectoryMode  os.FileMode            `yaml:"newDirectoryMode"`
+	NewDirectoryMode  *os.FileMode           `yaml:"newDirectoryMode"`
 }
 
 // NewClient creates a new config service client.
@@ -106,15 +106,22 @@ func (cfg Configuration) NewClient(iopts instrument.Options) (client.Client, err
 
 // NewOptions returns a new Options.
 func (cfg Configuration) NewOptions() Options {
-	return NewOptions().
+	opts := NewOptions().
 		SetZone(cfg.Zone).
 		SetEnv(cfg.Env).
 		SetService(cfg.Service).
 		SetCacheDir(cfg.CacheDir).
 		SetClusters(cfg.etcdClusters()).
 		SetServicesOptions(cfg.SDConfig.NewOptions()).
-		SetWatchWithRevision(cfg.WatchWithRevision).
-		SetNewDirectoryMode(cfg.NewDirectoryMode)
+		SetWatchWithRevision(cfg.WatchWithRevision)
+
+	if v := cfg.NewDirectoryMode; v != nil {
+		opts = opts.SetNewDirectoryMode(*v)
+	} else {
+		opts = opts.SetNewDirectoryMode(defaultDirectoryMode)
+	}
+
+	return opts
 }
 
 func (cfg Configuration) etcdClusters() []Cluster {
--- a/src/cluster/kv/etcd/store.go
+++ b/src/cluster/kv/etcd/store.go
@@ -619,25 +619,14 @@ func (c *client) writeCacheToFile() error {
 
 func (c *client) createCacheDir(fm os.FileMode) error {
 	path := path.Dir(c.opts.CacheFileFn()(c.opts.Prefix()))
-	if _, err := os.Stat(path); err != nil {
-		if os.IsNotExist(err) {
-			if err := os.MkdirAll(path, fm); err != nil {
-				c.m.diskWriteError.Inc(1)
-				c.logger.Warn("error creating cache directory",
-					zap.String("path", path),
-					zap.Error(err),
-				)
-				return fmt.Errorf("unable to create cache directory: %s", path)
-			}
-		} else {
-			c.m.diskReadError.Inc(1)
-			c.logger.Warn("error reading filesystem",
-				zap.String("path", path),
-				zap.Error(err),
-			)
-			return fmt.Errorf("unable to create cache directory: %s", path)
-		}
+	if err := os.MkdirAll(path, fm); err != nil {
+		c.m.diskWriteError.Inc(1)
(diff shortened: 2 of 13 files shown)


obitech:
@richardartoul

src/cluster/kv/etcd/store.go
```diff
@@ -616,7 +617,32 @@ func (c *client) writeCacheToFile() error {
 	return nil
 }
 
+func (c *client) createCacheDir() error {
+	path := path.Dir(c.opts.CacheFileFn()(c.opts.Prefix()))
```
richardartoul: I think `os.MkdirAll` is what you're looking for here.

This might be helpful as an example: https://github.com/m3db/m3/blob/master/src/dbnode/server/server.go#L188

src/cluster/kv/etcd/store_test.go
```diff
@@ -125,6 +126,27 @@ func TestNoCache(t *testing.T) {
...
 }
 
+func TestCacheDirCreation(t *testing.T) {
+	ec, opts, closeFn := testStore(t)
+
+	tdir, err := ioutil.TempDir("", "m3tests")
+	defer os.RemoveAll(tdir)
+	require.NoError(t, err)
+
+	cdir := path.Join(tdir, randSeq(10))
```
richardartoul: not sure why you need the randSeq when you're already using a new random test directory

richardartoul:
@cw9 Mind taking a look at this? I know you had some experience with the caching

src/cluster/kv/etcd/store_test.go
obitech: I think when calling `ioutil.TempDir`, it will already create a random directory. So I'm using `TempDir` and then choose a random dir name `cdir` be created by `NewStore` into `TempDir`. Then I can test for the existence of `cdir`.

src/cluster/kv/etcd/store.go
obitech: Ah good point with `os.MkdirAll`, thanks for the hint. 

I tried using `lockfile.CreateAndAcquire` but somehow the other tests are failing then:

```
--- FAIL: TestCache (0.70s)
        Error Trace:    store_test.go:198
        Error:          Received unexpected error "grpc: the client connection is closing"

--- FAIL: TestStaleDelete__FromGet (61.54s)
        Error Trace:    store_test.go:832
        Error:          Should be true
        Messages:       timed out waiting for client to read values from cache

--- FAIL: TestStaleDelete__FromWatch (62.06s)
        Error Trace:    store_test.go:917
        Error:          Should be true
        Messages:       timed out waiting for client to read values from cache
```

I'll just leave it with `os.MkDirAll` for now, but happy to take advice on how to properly implement the fs lock, if required.

src/cluster/kv/etcd/store_test.go
```diff
@@ -125,6 +126,28 @@ func TestNoCache(t *testing.T) {
 	require.Equal(t, 0, len(store.(*client).cacheUpdatedCh))
 }
 
+func TestCacheDirCreation(t *testing.T) {
+	ec, opts, closeFn := testStore(t)
+
```
cw9: move `defer closeFn()` here?

src/cluster/kv/etcd/store_test.go
cw9: nit: this feels like an overkill, a const name would do here.

src/cluster/kv/etcd/store_test.go
```diff
@@ -125,6 +126,28 @@ func TestNoCache(t *testing.T) {
...
+	tdir, err := ioutil.TempDir("", "m3tests")
+	defer os.RemoveAll(tdir)
```
cw9: Should this `defer` be called after `require.NoError`?

src/cluster/kv/etcd/store_test.go
```diff
@@ -125,6 +126,28 @@ func TestNoCache(t *testing.T) {
...
+	require.NoError(t, err)
+
+	cdir := path.Join(tdir, randSeq(10))
+	fname := path.Join(cdir, "mk3kv.json")
+	opts = opts.SetCacheFileFn(func(string) string {
```
cw9: Instead of this, can you test to make sure the `dir` you created is compatible with `c.opts.Prefix()`?, try some writes and then read from the cache maybe.

src/cluster/kv/etcd/store.go
richardartoul: Sorry I wasn't suggesting that you use `lockfile.CreateAndAcquire` I was just pointing to its implementation which uses os.MkdirAll (and passes the right permissions as well)

src/cluster/kv/etcd/store_test.go
obitech: Ok I can just use a const. I was thinking that there might be a case where this test could be run in parallel on the same machine, which might lead to multiple tests trying to create the same directory. But maybe I was overthinking :)

src/cluster/kv/etcd/store_test.go
obitech: Done

src/cluster/kv/etcd/store_test.go
obitech: Fixed

src/cluster/kv/etcd/store_test.go
obitech: Fixed

src/cluster/kv/etcd/store.go
```diff
@@ -616,7 +617,35 @@ func (c *client) writeCacheToFile() error {
...
+	if _, err := os.Stat(path); err != nil {
+		if os.IsNotExist(err) {
+			if err := os.MkdirAll(path, 0755); err != nil {
```
richardartoul: Sorry I was a little vague about this. I think if you use `os.MkdirAll` you can avoid the `os.Stat` and `os.IsNotExist`. If the path is already a directory then it will just return nil so I think the existing checks are redundant.

The permissions thing is gonna be kind of annoying :/

We actually flow the new directory permissions around the codebase as options, so if you take a look at `server.go` you'll see that when we call `lockfile.CreateAndAcquire(lockPath, newDirectoryMode)` (again, you dont need to use that function, just using it as an example that does something similar) the `newDirectoryMode` is passed in based on whatever its configured in our `FilesystemOptions` so I think we'll need to allow setting a `NewDirectoryMode` on whatever options are used to construct this store and then flow that into here if that makes sense.

obitech: Ahh I didn't get it the first time, sorry about that. Fixed it now.

And the permission thing was a bit annoying indeed haha. I think I got it now but it was quite the treasure hunt and I had to touch a couple of files along my way (tried to recap it in my commit message of 0d53e7e). Let me know if there is anything wrong or if I forgot anything

robskillington:
Nw, triggering a build on this now for CI.

src/cluster/client/etcd/config.go
```diff
@@ -111,7 +113,8 @@ func (cfg Configuration) NewOptions() Options {
 		SetCacheDir(cfg.CacheDir).
 		SetClusters(cfg.etcdClusters()).
 		SetServicesOptions(cfg.SDConfig.NewOptions()).
-		SetWatchWithRevision(cfg.WatchWithRevision)
+		SetWatchWithRevision(cfg.WatchWithRevision).
+		SetNewDirectoryMode(cfg.NewDirectoryMode)
```
robskillington: Hm, you might want to only set this if it's been set, maybe make it a pointer type?  

Otherwise the option will always be set to `os.FileMode(0)` which is not desirable, we'd prefer the default of 0755 I believe?

```go
NewDirectoryMode  *os.FileMode` yaml:"newDirectoryMode"`


if v := cfg.NewDirectoryMode; v != nil {
  opts = opts.SetNewDirectoryMode(*v)
}
```

src/cluster/kv/etcd/store.go
```diff
@@ -616,7 +617,29 @@ func (c *client) writeCacheToFile() error {
...
-func (c *client) initCache() error {
+func (c *client) createCacheDir(fm os.FileMode) error {
+	path := path.Dir(c.opts.CacheFileFn()(c.opts.Prefix()))
+	if err := os.MkdirAll(path, fm); err != nil {
+		c.m.diskWriteError.Inc(1)
+		c.logger.Warn("error creating cache directory",
+			zap.String("path", path),
+			zap.Error(err),
+		)
+	}
```
robskillington: I think you want to `return err` here, otherwise it will flow through to the create successful path no?

obitech: Absolutely! Forgot that to put back in after refactoring, fixed now.

src/cluster/client/etcd/config.go
obitech: True. I made it a pointer type and included some unit tests as well.

PR #1759, opened 2019-06-21
[ci] add script for running builds on external PRs

schallert:
Until we have a more complex label system for running PRs from external
contributors, this can hold us over.

Fixes #

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/.buildkite/scripts/pr_build.sh
+++ b/.buildkite/scripts/pr_build.sh
@@ -26,6 +26,11 @@ if [[ -z "$BUILDKITE_CLI_TOKEN" ]]; then
   exit 1
 fi
 
+if ! command -v jq; then
+  echo "cannot find jq in PATH"
+  exit 1
+fi
+
 RESP=$(mktemp)
 
 function cleanup() {
--- a/.buildkite/scripts/pr_build.sh
+++ b/.buildkite/scripts/pr_build.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+set -exo pipefail
+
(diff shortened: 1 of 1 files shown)


.buildkite/scripts/pr_build.sh
```diff
@@ -0,0 +1,70 @@
...
+function cleanup() {
+  rm -f "$RESP"
+}
+
+trap cleanup EXIT
+
+curl -H "Authorization: token $GITHUB_PR_TOKEN" -sSf -o "$RESP" "https://api.github.com/repos/m3db/m3/pulls/$1"
+
+function jq_field() {
+  <"$RESP" jq -r "$1"
```
mway: we should fail in `usage()` if `jq` isn't in `PATH`

schallert: Hmm we kinda get that for free right now:

```
$ PATH=/usr/bin:/bin .buildkite/scripts/pr_build.sh 1759
...
.buildkite/scripts/pr_build.sh: line 40: jq: command not found
...
```

mway: _coincidentally_, it doesn't do anything bad before it fails due to `jq` not being installed. 😉 

agreed that in this case it's nbd, but generally we should do checks beforehand (also, a "jq must be installed and in `$PATH` to use this tool" message is less ambiguous than a "command not found" error).

robskillington:
TY for this, I was looking into how we might be able to leverage web hooks to trigger this too (using labels or some such).

PR #1762, opened 2019-06-21
Implement UpdateOpenLease in `SeekManager`

richardartoul:
This P.R is required for the out of order writes feature. After a new volume for a given namespace/shard/blockstart is written out, the SeekerManager will be notified via a call to UpdateOpenLease(). This function imlpements UpdateOpenLease such that the SeekerManager "hot swaps" its old seekers for the old volume with new seekers for the new volume.

In terms of testing, this P.R adds some basic unit testing but the meat of the assurance that it works comes from extending the existing concurrency tests to use the new code paths AND simulate error cases so that those paths get tested as well (which they previously were not).

### DIFF
--- a/src/dbnode/persist/fs/retriever_test.go
+++ b/src/dbnode/persist/fs/retriever_test.go
@@ -349,10 +349,11 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 					require.True(
 						t,
 						seg.Equal(&compare),
-						"data mismatch for series %s, returned data: %v",
-						r.id,
-						string(seg.Head.Bytes()),
-						string(compare.Head.Bytes()))
+						fmt.Sprintf(
+							"data mismatch for series %s, returned data: %v, expected: %v",
+							r.id,
+							string(seg.Head.Bytes()),
+							string(compare.Head.Bytes())))
 
 					r.ctx.Close()
 				}
@@ -431,17 +432,15 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 
 	// Now that all the block lease updates have completed, all reads from this point should return tags with the
 	// highest volume number.
+	ctx := context.NewContext()
 	for _, shard := range shards {
 		for _, blockStart := range blockStarts {
 			for _, idString := range shardIDStrings[shard] {
-				var (
-					ctx = context.NewContext()
-					id  = ident.StringID(idString)
-				)
-
+				id := ident.StringID(idString)
 				for {
 					// Run in a loop since the open seeker function is configured to randomly fail
 					// sometimes.
+					ctx.Reset()
 					_, err := retriever.Stream(ctx, shard, id, blockStart, onRetrieve, nsCtx)
 					ctx.BlockingClose()
 					if err == nil {
--- a/src/dbnode/persist/fs/retriever_test.go
+++ b/src/dbnode/persist/fs/retriever_test.go
@@ -1,3 +1,4 @@
+// +build big
 //
 // Copyright (c) 2016 Uber Technologies, Inc.
 //
--- a/src/dbnode/persist/fs/retriever_test.go
+++ b/src/dbnode/persist/fs/retriever_test.go
@@ -462,8 +462,6 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 			require.Equal(t, expectedVolumeTag, volumeTag)
 		}
 	}
-
-	// panic("yolo")
 }
 
 // TestBlockRetrieverIDDoesNotExist verifies the behavior of the Stream() method
--- a/src/dbnode/persist/fs/retriever_test.go
+++ b/src/dbnode/persist/fs/retriever_test.go
@@ -175,7 +175,7 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 		if val := rand.Intn(100); val >= 90 {
 			return nil, errors.New("some-error")
 		}
-		return existingNewSeekerFn(shard, blockStart, volume)
+		return existingNewOpenSeekerFn(shard, blockStart, volume)
 	}
 	seekerMgr.newOpenSeekerFn = newNewOpenSeekerFn
 
--- a/src/dbnode/persist/fs/retriever_test.go
+++ b/src/dbnode/persist/fs/retriever_test.go
@@ -1,4 +1,3 @@
-// +build big
 //
 // Copyright (c) 2016 Uber Technologies, Inc.
 //
@@ -44,6 +43,7 @@ import (
 	"github.com/m3db/m3/src/x/context"
 	"github.com/m3db/m3/src/x/ident"
 	"github.com/m3db/m3/src/x/pool"
+	xsync "github.com/m3db/m3/src/x/sync"
 	xtime "github.com/m3db/m3/src/x/time"
 
 	"github.com/fortytw2/leaktest"
@@ -165,8 +165,8 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 
 	// Setup the open seeker function to fail sometimes to exercise that code path.
 	seekerMgr := retriever.seekerMgr.(*seekerManager)
-	existingNewSeekerFn := seekerMgr.newOpenSeekerFn
-	newNewSeekerFn := func(shard uint32, blockStart time.Time, volume int) (DataFileSetSeeker, error) {
+	existingNewOpenSeekerFn := seekerMgr.newOpenSeekerFn
+	newNewOpenSeekerFn := func(shard uint32, blockStart time.Time, volume int) (DataFileSetSeeker, error) {
 		// Artificially slow down how long it takes to open a seeker to exercise the logic where
 		// multiple goroutines are trying to open seekers for the same shard/blockStart and need
 		// to wait for the others to complete.
@@ -177,7 +177,7 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 		}
 		return existingNewSeekerFn(shard, blockStart, volume)
 	}
-	seekerMgr.newOpenSeekerFn = newNewSeekerFn
+	seekerMgr.newOpenSeekerFn = newNewOpenSeekerFn
 
 	// Setup the block lease manager to return errors sometimes to exercise that code path.
 	mockBlockLeaseManager := block.NewMockLeaseManager(ctrl)
@@ -366,7 +366,7 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 	startWg.Done()
 
 	// Setup concurrent block lease updates.
-	sem := make(chan struct{}, updateOpenLeaseConcurrency)
+	workers := xsync.NewWorkerPool(updateOpenLeaseConcurrency)
 	// Volume -> shard -> blockStart to stripe as many shard/blockStart as quickly as possible to
 	// improve the chance of triggering the code path where UpdateOpenLease is the first time a set
 	// of seekers are opened for a shard/blocksStart combination.
@@ -374,8 +374,13 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 		for _, shard := range shards {
 			for _, blockStart := range blockStarts {
 				enqueueWg.Add(1)
-				sem <- struct{}{}
-				go func(shard uint32, blockStart time.Time, volume int) {
+				var (
+					// Capture vars for async goroutine.
+					volume     = volume
+					shard      = shard
+					blockStart = blockStart
+				)
+				workers.Go(func() {
 					defer enqueueWg.Done()
 					leaser := retriever.seekerMgr.(block.Leaser)
 
@@ -395,9 +400,7 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 							break
 						}
 					}
-
-					<-sem
-				}(shard, blockStart, volume)
+				})
 			}
 		}
 	}
@@ -438,6 +441,7 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 					// Run in a loop since the open seeker function is configured to randomly fail
 					// sometimes.
 					_, err := retriever.Stream(ctx, shard, id, blockStart, onRetrieve, nsCtx)
+					ctx.BlockingClose()
 					if err == nil {
 						break
 					}
--- a/src/dbnode/persist/fs/retriever_test.go
+++ b/src/dbnode/persist/fs/retriever_test.go
@@ -29,6 +29,7 @@ import (
 	"math/rand"
 	"os"
 	"path/filepath"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
@@ -80,7 +81,7 @@ func newOpenTestBlockRetriever(
 	require.NoError(t, retriever.Open(testNs1Metadata(t)))
 
 	return retriever, func() {
-		assert.NoError(t, retriever.Close())
+		require.NoError(t, retriever.Close())
 	}
 }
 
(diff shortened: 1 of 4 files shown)


src/dbnode/persist/fs/retriever_test.go
```diff
@@ -131,74 +134,124 @@ func TestBlockRetrieverHighConcurrentSeeksCacheShardIndices(t *testing.T) {
...
 	retriever, cleanup := newOpenTestBlockRetriever(t, opts)
 	defer cleanup()
 
-	nsMeta := testNs1Metadata(t)
-	ropts := nsMeta.Options().RetentionOptions()
-	nsCtx := namespace.NewContextFrom(nsMeta)
+	// Setup the open seeker function to fail sometimes to exercise that code path.
+	seekerMgr := retriever.seekerMgr.(*seekerManager)
+	existingNewSeekerFn := seekerMgr.newOpenSeekerFn
+	newNewSeekerFn := func(shard uint32, blockStart time.Time, volume int) (DataFileSetSeeker, error) {
```
robskillington: nit: Is this meant to be `newOpenSeekerFn` instead of `newNewSeekerFn`?

src/dbnode/persist/fs/retriever_test.go
```diff
@@ -288,10 +360,48 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
...
 	}
 
-	// Wait for all routines to be ready then start.
+	// Wait for all routines to be ready.
 	readyWg.Wait()
+	// Allow all the goroutines to begin.
 	startWg.Done()
 
+	// Setup concurrent block lease updates.
+	sem := make(chan struct{}, updateOpenLeaseConcurrency)
```
robskillington: Would it perhaps be more idiomatic to use a worker pool here to control concurrency instead of `sem` channel?  Had to read the code a little just to verify it was being used to bound concurrency, vs being used for some other mechanism.

src/dbnode/persist/fs/retriever_test.go
```diff
@@ -303,13 +413,55 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
...
 		}
 	}
+
+	// Now that all the block lease updates have completed, all reads from this point should return tags with the
+	// highest volume number.
+	for _, shard := range shards {
+		for _, blockStart := range blockStarts {
+			for _, idString := range shardIDStrings[shard] {
+				var (
+					ctx = context.NewContext()
```
robskillington: Maybe good practice to close the context after use?

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -269,11 +279,58 @@ func (m *seekerManager) Return(shard uint32, start time.Time, seeker ConcurrentD
...
+			// Clear out inactive state.
+			allInactiveSeekersClosedWg := seekers.inactive.wg
+			seekers.inactive = seekersAndBloom{}
+			if allInactiveSeekersClosedWg != nil {
+				// Signal completion regardless of any errors encountered while closing.
+				allInactiveSeekersClosedWg.Done()
+			}
+
+			if multiErr.FinalError() != nil {
+				return multiErr.FinalError()
```
robskillington: super nit: More idiomatic perhaps to bind the err
```go
if err := mutliErr.FinalError(); err != nil {
  return err
}
```

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -269,11 +279,58 @@ func (m *seekerManager) Return(shard uint32, start time.Time, seeker ConcurrentD
...
+					break
+				}
+			}
+
+			if !allAreReturned {
+				break
+			}
+
+			// All the inactive seekers have been returned so it's safe to signal and clear them out.
+			var multiErr = xerrors.NewMultiError()
```
robskillington: super nit: More idiomatic to just use `:=` instead of untyped var `var foo = bar`

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -286,15 +343,188 @@ func (m *seekerManager) Return(shard uint32, start time.Time, seeker ConcurrentD
...
+		m.Unlock()
+		return 0, errConcurrentUpdateOpenLeaseNotAllowed
+	}
+
+	if !m.namespace.Equal(descriptor.Namespace) {
+		m.Unlock()
+		return block.NoOpenLease, nil
+	}
+	m.isUpdatingLease = true
+	m.Unlock()
```
robskillington: nit: Could the code above be grouped into a `startUpdateOpenLease() error`?

That would be able to use `m.Lock()` and `defer`:
```go
func (m *seekerManager) UpdateOpenLease(
	descriptor block.LeaseDescriptor,		descriptor block.LeaseDescriptor,
	state block.LeaseState,		state block.LeaseState,
) (block.UpdateOpenLeaseResult, error) {	) (block.UpdateOpenLeaseResult, error) {
	m.Lock()
	if m.status != seekerManagerOpen {
		m.Unlock()
		return 0, errUpdateOpenLeaseSeekerManagerNotOpen
	}

 	if m.isUpdatingLease {
		// This guard is a little overly aggressive. In practice, the algorithm remains correct even in the presence
		// of concurrent UpdateOpenLease() calls as long as they are for different shard/blockStart combinations.
		// However, the calling code currently has no need to call this method concurrently at all so use the
		// simpler check for now.
		m.Unlock()
		return 0, errConcurrentUpdateOpenLeaseNotAllowed
	}

 	if !m.namespace.Equal(descriptor.Namespace) {
		m.Unlock()
		return block.NoOpenLease, nil
	}
	m.isUpdatingLease = true
	m.Unlock()

	// .... other code
```

Becomes:
```go
func (m *seekerManager) startUpdateOpenLease(
	descriptor block.LeaseDescriptor,
) error {
	m.Lock()
	defer m.Unlock()
	if m.status != seekerManagerOpen {
		return 0, errUpdateOpenLeaseSeekerManagerNotOpen
	}

 	if m.isUpdatingLease {
		// This guard is a little overly aggressive. In practice, the algorithm remains correct even in the presence
		// of concurrent UpdateOpenLease() calls as long as they are for different shard/blockStart combinations.
		// However, the calling code currently has no need to call this method concurrently at all so use the
		// simpler check for now.
		return 0, errConcurrentUpdateOpenLeaseNotAllowed
	}

 	if !m.namespace.Equal(descriptor.Namespace) {
		return block.NoOpenLease, nil
	}
	m.isUpdatingLease = true
	return nil
}

func (m *seekerManager) UpdateOpenLease(
	descriptor block.LeaseDescriptor,
	state block.LeaseState,
) (block.UpdateOpenLeaseResult, error) {
	if err := m.startUpdateOpenLease(); err != nil {
		return err
	}
	defer func() {
		m.Lock()
		m.isUpdatingLease = false
		m.Unlock()
	}()

	// .... other code
```

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -286,15 +343,188 @@ func (m *seekerManager) Return(shard uint32, start time.Time, seeker ConcurrentD
...
+	// or
+	//   b) Seeks are not present for this blockStart and no other goroutines are currently trying to open them,
+	//      in which case this goroutine can just set the new seekers and be done.
+	for {
+		byTime.Lock()
+		seekers, ok = byTime.seekers[blockStartNano]
+
+		if !ok || seekers.active.wg == nil {
+			// Exit the loop still holding the lock.
+			break
```
robskillington: This definitely seems somewhat dangerous to break the loop and expect the following code to always `byTime.Unlock()` without a defer being used somewhere.

Can break this into sub functions to guarantee the locking lock and unlock semantics? 

I don't see any errors with this code, but someone making a change could definitely introduce a bug that could potentially slip past tests unless it's more straight forward to reason about.

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -286,15 +343,188 @@ func (m *seekerManager) Return(shard uint32, start time.Time, seeker ConcurrentD
...
+	byTime.Unlock()
+
+	if wg != nil {
+		// Wait for all the inactive seekers to be returned and closed because the contract
+		// of this API is that the Leaser (SeekerManager) should have relinquished any resources
+		// associated with the old lease by the time this function returns.
+		wg.Wait()
+	}
+
+	return updateOpenLeaseResult, nil
```
robskillington: I think the above code is hairy enough we should try to break it into some subfunctions so that the high level code can more straight forward manage the locks and waitgroup creation while the subfunctions handle the minutia of each task, what do you think?

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -308,46 +538,79 @@ func (m *seekerManager) UpdateOpenLease(
...
+		// the seekers. This is done *after* acquiring the lock so that other goroutines that
+		// were waiting won't acquire the lock before this goroutine does.
+		wg.Done()
+	}
+	onErr := func() {
+		onDoneOpeningNewSeeker()
+		// Delete the seekersByTime struct so that the process can be restarted by the next
+		// goroutine (since this one errored out).
+		delete(byTime.seekers, start)
+	}
```
robskillington: Hm, this definitely is adding more hairy semantics inside an already complex function too.

Can we do something like:
```go
func (m *seekerManager) getOrOpenSeekerWithLock(start xtime.UnixNano, byTime *seekersByTime) (seekersAndBloom, error) {
	// all previous code

	newActiveSeekers, err := m.openSeekersWithActiveWaitGroup(start, byTime)

	// Re-establish lock and signal
	byTime.Lock()
	wg.Done()

	if err != nil {
		// Delete so process can be restarted...
		delete(byTime.seekers, start)
		return err
	}

	return seekersAndBloom, nil
}

func (m *seekerManager) openSeekersWithActiveWaitGroup(start xtime.UnixNano, byTime *seekersByTime) (seekersAndBloom, error) {
	// Open first one - Do this outside the context of the lock because opening
	// a seeker can be an expensive operation (validating index files).
	blm := m.blockRetrieverOpts.BlockLeaseManager()
	state, err := blm.OpenLatestLease(m, block.LeaseDescriptor{
		Namespace:  m.namespace,
		Shard:      byTime.shard,
		BlockStart: start.ToTime(),
	})
	if err != nil {
		return seekersAndBloom{}, fmt.Errorf("err opening latest lease: %v", err)
	}

	seeker, err := m.newOpenSeekerFn(byTime.shard, start.ToTime(), state.Volume)
	if err != nil {
		return seekersAndBloom{}, err
	}

	activeSeekers, err := m.seekersAndBloomFromSeeker(seeker, state.Volume)
	if err != nil {
		return seekersAndBloom{}, err
	}

	seekers.active = activeSeekers
	byTime.seekers[start] = seekers
	return activeSeekers, nil
}
```

src/dbnode/persist/fs/retriever_test.go
richardartoul: Yeah I can make this less confusing

src/dbnode/persist/fs/retriever_test.go
richardartoul: sure thing

src/dbnode/persist/fs/seek_manager.go
richardartoul: yeah I think thats fair, let me see what I can do

src/dbnode/persist/fs/seek_manager.go
richardartoul: yeah sure this is better

src/dbnode/persist/fs/seek_manager.go
richardartoul: Yeah ok I think I see a way to do it, I'll ping you once I've implemented

src/dbnode/persist/fs/seek_manager.go
richardartoul: yeah looks good (or something similar at least once I test this out). Will ping once implemented

richardartoul: OK made a few minor tweaks but this is implemented now (move the last two lines out of this function because they need to happen with the lock held)

src/dbnode/persist/fs/seek_manager.go
richardartoul: Yeah I just refactored it so the locking and such is more straightforward

richardartoul:
@robskillington @justinjc Address all the feedback. A lot of the locking/waitgroup stuff should be a lot more straightforward now

src/dbnode/persist/fs/retriever_test.go
```diff
@@ -279,7 +346,13 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
 
 					require.NoError(t, err)
 					compare.Head = shardData[r.shard][r.id][xtime.ToUnixNano(r.blockStart)]
-					assert.True(t, seg.Equal(&compare))
+					require.True(
+						t,
+						seg.Equal(&compare),
+						"data mismatch for series %s, returned data: %v",
```
robskillington: Looks like you have only two printf wildcards but three vars that follow?

I find to avoid this possibility it's actually easier to just wrap this up in it's own `fmt.Sprintf(...)` to `require` or `assert`.

src/dbnode/persist/fs/retriever_test.go
```diff
@@ -288,10 +361,52 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
...
+		for _, shard := range shards {
+			for _, blockStart := range blockStarts {
+				enqueueWg.Add(1)
+				var (
+					// Capture vars for async goroutine.
+					volume     = volume
+					shard      = shard
+					blockStart = blockStart
+				)
+				workers.Go(func() {
```
robskillington: nit: Is there any possibility this won't be done in order if spawn asynchronously?  Just a thought, if we think it's super unlikely to cause flakes/become an issue then I guess we can just ignore this possibility (without adding queueing/etc on a per volume/shard/blockstart tuple) and address it if it ever does happen.

Maybe at the least we should leave a note here about your thoughts on that?

src/dbnode/persist/fs/retriever_test.go
```diff
@@ -303,13 +418,56 @@ func testBlockRetrieverHighConcurrentSeeks(t *testing.T, shouldCacheShardIndices
...
+			for _, idString := range shardIDStrings[shard] {
+				var (
+					ctx = context.NewContext()
+					id  = ident.StringID(idString)
+				)
+
+				for {
+					// Run in a loop since the open seeker function is configured to randomly fail
+					// sometimes.
+					_, err := retriever.Stream(ctx, shard, id, blockStart, onRetrieve, nsCtx)
```
robskillington: super nit: Since you are now calling `BlockingClose()` below you should also call `ctx.Reset()` before using.

src/dbnode/persist/fs/seek_manager.go
```diff
@@ -268,33 +278,263 @@ func (m *seekerManager) Return(shard uint32, start time.Time, seeker ConcurrentD
...
+		updateOpenLeaseResult = block.NoOpenLease
+	)
+	seekers, ok := m.acquireByTimeLockWaitGroupAware(blockStartNano, byTime)
+	defer byTime.Unlock()
+	if !ok {
+		// No existing seekers, so just set the newly created ones and be done.
+		seekers.active = newActiveSeekers
+		byTime.seekers[blockStartNano] = seekers
+		return nil, updateOpenLeaseResult, nil
+	}
```
robskillington: This is a lot better, nice work.

src/dbnode/persist/fs/seek_manager_test.go
```diff
@@ -73,6 +74,97 @@ func TestSeekerManagerCacheShardIndices(t *testing.T) {
 	}
 }
 
+func TestSeekerManagerUpdateOpenLease(t *testing.T) {
+	defer leaktest.CheckTimeout(t, 1*time.Minute)()
+
+	var (
+		ctrl   = gomock.NewController(t)
```
robskillington: Everywhere we use mocks its best to “defer ctrl.Finish()” so that any calls you expected to happen but didn’t cause a test to fail (since this means something you expected to happen didn’t, which is a good catch).

richardartoul: yeah good call, this usually happens when I copy and paste the controller code from another function I wrote in which I forgot to call finish heh

PR #1764, opened 2019-06-21
Add VolumeIndex back/forwards compatible to info files + restore flush version on shard bootstrap

richardartoul:
This P.R add the `VolumeIndex` into the Info file of all the fileset files in a backwards and forwards compatible way. It also ensures that the volume index is written out by the writers and return by calls to `ReadInfoFile`.

Finally, it updates the shard bootstrap logic to restore the Cold flush version on bootstrap as well as the flush state.

### DIFF
--- a/src/dbnode/storage/shard.go
+++ b/src/dbnode/storage/shard.go
@@ -1873,7 +1873,7 @@ func (s *dbShard) Bootstrap(
 
 		s.markWarmFlushStateSuccess(at)
 		// Cold version needs to get bootstrapped so that the 1:1 relationship between volume number
-		// and cold version is maintained and the volume numbers / flush versions remaining monotonically
+		// and cold version is maintained and the volume numbers / flush versions remain monotonically
 		// increasing.
 		s.setFlushStateColdVersion(at, info.VolumeIndex)
 	}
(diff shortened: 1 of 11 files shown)


src/dbnode/storage/shard.go
```diff
@@ -1872,6 +1872,11 @@ func (s *dbShard) Bootstrap(
 		}
 
 		s.markWarmFlushStateSuccess(at)
+		// Cold version needs to get bootstrapped so that the 1:1 relationship between volume number
+		// and cold version is maintained and the volume numbers / flush versions remaining monotonically
+		// increasing.
+		// TODO(rartoul): Add a test for this before merging.
```
robskillington: super nit: Want to use the TODO_COLD_FLUSH... flag?

robskillington: Oh before merging, never mind.

richardartoul: added the coverage

justinjc:
LGTM. Pushed this together with #1762 on a test cluster and everything looked good in terms of backwards- and forwards-compatibility (from >=0.11.0).

PR #1767, opened 2019-06-22
Add debug zip dump functionality

obitech:

Finishes the work started in https://github.com/m3db/m3/pull/1269/

I have copied the work initially started by @lyuboraykov and implemented the suggestions made by @richardartoul in the review:
- Remove Thrift endpoint
- Add CPU Profiling information
- Add HTTP endpoint

```release-note
NONE
```

```documentation-note
NONE
```


obitech:
@richardartoul as discussed

.gitignore
```diff
@@ -30,6 +30,9 @@ _tools/
 debug
 debug.test
 
+# Track src/x/debug though
+!src/x/debug
```
robskillington: Hah, that's frustrating.  

I think that `debug.test` is usually what we see end up in our directories, perhaps we should just remove the `debug` entry (and keep `debug.test`) instead of explicitly opt in this path?

Just mindful and hoping we wouldn't have to do this in the future for other packages with similar names.

obitech: I removed `debug` as well as the explicit opt `src/x/debug` and it seems alright alright

src/x/debug/debug.go
```diff
@@ -0,0 +1,108 @@
...
+func (i *zipWriter) HTTPHandler() http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		errMsg := "Unable to write ZIP file:"
+		buf := bytes.NewBuffer([]byte{})
+		if err := i.WriteZip(buf); err != nil {
+			http.Error(w, fmt.Sprintf("%s: %s", errMsg, err), http.StatusInternalServerError)
+		}
+		if _, err := io.Copy(w, buf); err != nil {
+			http.Error(w, fmt.Sprintf("%s: %s", errMsg, err), http.StatusInternalServerError)
+		}
```
robskillington: For the error case, I believe it would be good to early return so multiple responses aren't tried to be written out (you can only send HTTP status code header once anyway)?

Also do you want to use our `xhttp.Error(...)` helper?
https://github.com/m3db/m3/blob/master/src/x/net/http/errors.go#L42-L45

And for the second error, perhaps we just log it?  Since the HTTP 200 will have already been sent.

For example:
```go
import (
    // ...
    "github.com/m3db/m3/src/x/instrument"
    "github.com/m3db/m3/src/x/net/http"
    // ...
)

// ...

type zipWriter struct {
  sources map[string]Source
  logger *zap.Logger
}

func NewZipWriter(instrumentOpts instrument.Options) ZipWriter {
  return &zipWriter{
    sources: make(map[string]Source),
    logger: instrumentOpts.Logger(),
  }
}

// ...

func (i *zipWriter) HTTPHandler() http.Handler {
  return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
    errMsg := "unable to write ZIP file:"
    buf := bytes.NewBuffer([]byte{})
    if err := i.WriteZip(buf); err != nil {
        xhttp.Error(w, fmt.Sprintf("%s: %s", errMsg, err), http.StatusInternalServerError)
        return
    }
    if _, err := io.Copy(w, buf); err != nil {
        i.logger.Error("unable to write ZIP response", zap.Error(err))
        return
    }
  })
}

// ...
```

src/x/debug/heap.go
```diff
@@ -0,0 +1,38 @@
...
+
+// NewHeapDumpSource returns a Source for heapdump.
+func NewHeapDumpSource() Source {
+	return &heapDumpSource{}
+}
+
+// Write writes the heapdump in the provided writer.
+func (h *heapDumpSource) Write(w io.Writer) error {
+	return pprof.WriteHeapProfile(w)
+}
```
robskillington: nit: Add endline at end of this file?

src/x/debug/heap_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestHeapDumpSource(t *testing.T) {
+	heapDumpSource := NewHeapDumpSource()
+	buff := bytes.NewBuffer([]byte{})
+	heapDumpSource.Write(buff)
+	require.NotZero(t, buff.Len())
+}
```
robskillington: nit: Add endline at end of this file?

src/x/debug/host.go
```diff
@@ -0,0 +1,61 @@
...
+		PID: os.Getpid(),
+		CWD: wd,
+	}
+	jsonData, err := json.MarshalIndent(hostInfo, "", "  ")
+	if err != nil {
+		return err
+	}
+	w.Write(jsonData)
+	return nil
+}
```
robskillington: nit: Add endline at end of this file?

src/x/debug/host_test.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+func TestHostInfoSource(t *testing.T) {
+	source := NewHostInfoSource()
+	buff := bytes.NewBuffer([]byte{})
+	err := source.Write(buff)
+	require.NoError(t, err)
+	err = json.Unmarshal(buff.Bytes(), &hostInfoSource{})
+	require.NoError(t, err)
+}
```
robskillington: nit: Add endline at end of this file?

robskillington:
One other thing, it might be worth getting some other profiles through `pprof.Lookup` too.  For instance the dump of the goroutines would be super helpful.

e.g.

```go

const (
  goroutineProfileName = "goroutine"
  goroutineDebug = 2 // capture all active goroutines
)

type goroutineSource struct {}

func (s *goroutineSource) Write(w io.Writer) error {
  profile := pprof.Lookup(goroutineProfileName)
  if profile == nil {
    return fmt.Errorf("profile not found: %s", goroutineProfileName)
  }
  return profile.WriteTo(w, goroutineDebug)
}

```

src/x/debug/debug.go
```diff
@@ -0,0 +1,108 @@
...
+	RegisterHandler(string, *http.ServeMux) error
+}
+
+type zipWriter struct {
+	sources map[string]Source
+}
+
+// NewZipWriter returns an instance of an ZipWriter. The passed prefix
+// indicates the folder where to save the zip files.
+func NewZipWriter() ZipWriter {
```
richardartoul: Could we add another constructor function: `NewZipWriterWithDefaultSources` that basically creates a zip writer, then creates/registers the CPU/heap/host sources? That way when we want to use this package and register the http endpoint it basically just becomes:

```golang
zipWriter, err := debug.NewZipWriterWithDefaultSources()
if err != nil {
   ...
}
// Register any custom sources here.
...
handler := zipWriter.HTTPHandler()
...
```

src/x/debug/debug.go
richardartoul: +1

obitech: Thanks for the suggestion! Fixed it.

src/x/debug/heap.go
obitech: Fixed

src/x/debug/heap_test.go
obitech: Fixed

src/x/debug/host.go
obitech: Fixed

src/x/debug/host_test.go
obitech: Fixed

src/x/debug/debug.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"archive/zip"
+	"bytes"
+	"fmt"
+	"io"
+	"net/http"
+)
+
+// Source provides functions for fetching debug information from a single debug source.
+type Source interface {
```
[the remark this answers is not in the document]
obitech: Sure!

src/x/debug/debug.go
obitech: Done!

obitech:
@robskillington 

> One other thing, it might be worth getting some other profiles through `pprof.Lookup` too. For instance the dump of the goroutines would be super helpful.
> 
> e.g.
> 
> ```go
> const (
>   goroutineProfileName = "goroutine"
>   goroutineDebug = 2 // capture all active goroutines
> )
> 
> type goroutineSource struct {}
> 
> func (s *goroutineSource) Write(w io.Writer) error {
>   profile := pprof.Lookup(goroutineProfileName)
>   if profile == nil {
>     return fmt.Errorf("profile not found: %s", goroutineProfileName)
>   }
>   return profile.WriteTo(w, goroutineDebug)
> }
> ```

I tried to make the implementation of `Profiles` a bit more general so we can use all the other predefined names as well. I hope this could make the implementation a bit more flexible but I'm not sure if I've thought of all the corner cases, could you review `profile.go`?

obitech:
> Thanks for taking this on, this is almost ready to go and will be a big improvement when we're trying to help people debug issues.
> 
> If you're up for it, once we get this merged we can make a followup P.R to hook up the endpoint and add some custom debug sources (that include the YAML config, namespace configuration, configured bootstrappers etc)

More than happy to!

robskillington:
Brilliant, sounds good, taking a look now.

src/x/debug/profile_test.go
```diff
@@ -0,0 +1,73 @@
...
+				// Test additional calls to Profile()
+				goProf = goProfSource.Profile()
+				require.NotNil(t, goProf)
+
+				// Test zip writes
+				buf := bytes.NewBuffer([]byte{})
+				err = goProfSource.Write(buf)
+				require.NoError(t, err)
+				require.NotZero(t, buf.Len())
+				fmt.Printf("%s\n", buf.String())
```
robskillington: nit: Probably not necessary to fmt.Printf in a test unless you're actively debugging it.

obitech: Damn, I forgot to remove this after my debugging sessions. Removed it now.

robskillington:
@obitech seems to be failing to build (annoying hah):
```
src/x/debug/profile_test.go:25:2: imported and not used: "fmt"
```

obitech:
> @obitech seems to be failing to build the tests (annoying hah):
> 
> ```
> src/x/debug/profile_test.go:25:2: imported and not used: "fmt"
> ```

Ha, that's what you get when you hastily edit files in the browser. Fixed it, hope it passes now.

PR #1771, opened 2019-06-24
[ci] install nvm locally; retool base dir

schallert:
- We can't make assumptions about the version of `node` a user may have
  installed. If we don't find `nvm`, install it locally.
- `retool` needs its base dir set if the nearest git dir isn't the
  project parent.

Fixes #

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -44,6 +44,13 @@ GO_RELEASER_DOCKER_IMAGE  := goreleaser/goreleaser:v0.93
 GO_RELEASER_WORKING_DIR   := /go/src/github.com/m3db/m3
 GOMETALINT_VERSION        := v2.0.5
 
+# Retool will look for tools.json in the nearest parent git directory if not
+# explicitly told the current dir. Allow setting the base dir so that tools can
+# be built inside of other external repos.
+ifdef RETOOL_BASE_DIR
+	retool_base_args := -base-dir $(RETOOL_BASE_DIR)
+endif
+
 export NPROC := 2 # Maximum package concurrency for unit tests.
 
 SERVICES :=     \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -419,19 +429,15 @@ node-yarn-run:
...
 else
-	node --version 2>&1 >/dev/null || \
-		(echo "Trying apt install" && which apt-get && (curl -sL https://deb.nodesource.com/setup_$(node_version).x | bash) && apt-get install -y nodejs) || \
-		(echo "Trying apk install" && which apk && apk add --update nodejs nodejs-npm) || \
-		(echo "No node install or known package manager" && exit 1)
-	@echo "Not using nvm, using node version $(shell node --version)"
-	bash -c "$(node_cmd)"
+	mkdir .nvm
+	# Install nvm locally
+	curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.34.0/install.sh | NVM_DIR=$$(pwd)/.nvm PROFILE=/dev/null bash
```
prateek: *dons tinfoil hat* not a fan of cURL+exec-ing , mind pulling the script into the repo

Makefile
```diff
@@ -419,19 +429,15 @@ node-yarn-run:
...
+	bash -c "source $$(pwd)/.nvm/nvm.sh; nvm install 6"
```
prateek: hm would `SELF_DIR` (declared at the top of the makefile) be better than `pwd`?

Makefile
schallert: Definitely :)

Makefile
schallert: Ah nice catch, ty

PR #1774, opened 2019-06-25
[etcdclient] Set default auto sync URL

robskillington:

There was no default specified for etcd auto sync interval, considering this is likely desirable for any deployment (discover new endpoints if membership changes) I think it would be ideal to have it default to a value.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/cluster/client/etcd/config.go
+++ b/src/cluster/client/etcd/config.go
@@ -34,7 +34,7 @@ type ClusterConfig struct {
 	Endpoints        []string         `yaml:"endpoints"`
 	KeepAlive        *keepAliveConfig `yaml:"keepAlive"`
 	TLS              *TLSConfig       `yaml:"tls"`
-	AutoSyncInterval time.Duration    `yaml:"autoSyncInterval"`
+	AutoSyncInterval *time.Duration   `yaml:"autoSyncInterval"`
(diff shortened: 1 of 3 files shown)


nerd0:
@robskillington just an FYI the typical default value from the ETCD client is `0s` which indicates do not enable auto sync. Since this feature hasn't been in place since the cluster package was created I would hazard to use `0s`

nerd0:
> I would rather we wait to enable this for everyone until we have a better idea of the implications.
> 
> 1. I can't find it now, but I've seen comments in the past saying there are some gotchas with autosync.
> 2. We haven't yet tested the impact this could have on the etcd cluster, and potentially dozens of thousands of hosts each calling the members API every 60s sounds dangerous until we better understand the impact.

+1 , forgot to remove my approval after thinking twice about this.

robskillington:
Sure thing, can you let me know what it looks like in the wild?  I'm wary of adding this config then because a sensible default isn't set we can help people but only after they've experienced an outage and it'd be good to avoiding telling people "first thing you do when you setup a cluster is you turn this config value on".

I agree with waiting to see how it looks though of course.

robskillington:
Already merged a default.

PR #1778, opened 2019-06-26
DataFileSetExists and OptimizedTimes ForEach bug fixes

justinjc:

* Fixes `DataFileSetExists` incorrectly checking the legacy fileset when the volume isn't `0`
* Fixes the `ForEach` func for `OptimizedTimes` to not loop through zero values that weren't explicitly added

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/dbnode/persist/fs/write.go
+++ b/src/dbnode/persist/fs/write.go
@@ -146,12 +146,11 @@ func NewWriter(opts Options) (DataFileSetWriter, error) {
 // opening / truncating files associated with that shard for writing.
 func (w *writer) Open(opts DataWriterOpenOptions) error {
 	var (
-		err             error
-		namespace       = opts.Identifier.Namespace
-		shard           = opts.Identifier.Shard
-		blockStart      = opts.Identifier.BlockStart
-		volumeIndex     = opts.Identifier.VolumeIndex
-		nextVolumeIndex = opts.Identifier.VolumeIndex
+		err         error
+		namespace   = opts.Identifier.Namespace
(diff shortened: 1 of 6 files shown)


src/dbnode/persist/fs/write.go
```diff
@@ -146,12 +146,11 @@ func NewWriter(opts Options) (DataFileSetWriter, error) {
 // opening / truncating files associated with that shard for writing.
 func (w *writer) Open(opts DataWriterOpenOptions) error {
 	var (
-		err             error
-		namespace       = opts.Identifier.Namespace
-		shard           = opts.Identifier.Shard
-		blockStart      = opts.Identifier.BlockStart
-		volumeIndex     = opts.Identifier.VolumeIndex
-		nextVolumeIndex = opts.Identifier.VolumeIndex
```
justinjc: This was the result of a bad merge, causing duplicate variables for the same thing.

src/dbnode/storage/shard.go
```diff
@@ -1910,6 +1910,7 @@ func (s *dbShard) WarmFlush(
 		// where a fileset already exists when we attempt to flush unless there
 		// is a bug in the code.
 		DeleteIfExists: false,
+		FileSetType:    persist.FileSetFlushType,
```
richardartoul: Is this just setting it to the default value concretely?

justinjc: Yep this is a noop. Previously it was just using the zero value which corresponds to `persist.FileSetFlushType`. Seemed dangerous to rely on that ordering.

PR #1780, opened 2019-06-26
Exit cold flush early when there is nothing to merge

justinjc:

Currently, cold flushes happen unnecessarily because `dirtySeriesToWrite` contains empty `seriesList`s (since it is a shared resource between shards). Checking for data in `dirtySeries` is a sufficient condition in figuring out whether a merge should actually happen or not.

```release-note
NONE
```

```documentation-note
NONE
```


justinjc:
Sure thing, added one

PR #1781, opened 2019-06-27
[query] Add multi-query remote read

arnikola:
Allows batched remote-read endpoint

```release-note
none
```

```documentation-note
none
```

### DIFF
--- a/src/query/api/v1/handler/prometheus/remote/read.go
+++ b/src/query/api/v1/handler/prometheus/remote/read.go
@@ -22,8 +22,8 @@ package remote
 
 import (
 	"context"
-	"fmt"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/m3db/m3/src/query/api/v1/handler"
@@ -33,6 +33,7 @@ import (
 	"github.com/m3db/m3/src/query/models"
 	"github.com/m3db/m3/src/query/storage"
 	"github.com/m3db/m3/src/query/util/logging"
+	xerrors "github.com/m3db/m3/src/x/errors"
 	"github.com/m3db/m3/src/x/instrument"
 	xhttp "github.com/m3db/m3/src/x/net/http"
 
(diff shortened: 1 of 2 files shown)


src/query/api/v1/handler/prometheus/remote/read.go
```diff
@@ -178,39 +179,64 @@ func (h *PromReadHandler) read(
...
+		wg           sync.WaitGroup
+		multiErr     xerrors.MultiError
+		multiErrLock sync.Mutex
+	)
+
+	wg.Add(queryCount)
+	for i, promQuery := range r.Queries {
+		// capture loop variables
+		promQuery := promQuery
+		i := i
```
robskillington: super nit: Can reduce this to: `i, promQuery := i, promQuery`

src/query/executor/engine.go
```diff
@@ -115,8 +115,6 @@ func (e *engine) Execute(
 	opts *QueryOptions,
 	results chan *storage.QueryResult,
 ) {
-	defer close(results)
```
robskillington: Hm, did you mean to delete this?

In the new code above you have the statement "// Results is closed by execute" which is not true anymore if this defer close doesn't occur?

src/query/api/v1/handler/prometheus/remote/read.go
```diff
@@ -178,39 +179,64 @@ func (h *PromReadHandler) read(
...
+			cancelFuncs[i] = cancel
+			query, err := storage.PromReadQueryToM3(promQuery)
+			if err != nil {
+				multiErrLock.Lock()
+				multiErr = multiErr.Add(err)
+				multiErrLock.Unlock()
+				return
+			}
+
+			// Results is closed by execute
```
robskillington: nit: Remove this statement since the channel is no longer closed?

PR #1783, opened 2019-06-28
[query] Clean executor interfaces

arnikola:
Cleans up messy executor interfaces

```release-note
no
```

```documentation-note
no
```


src/query/api/v1/handler/prometheus/native/read_common.go
```diff
@@ -128,28 +118,9 @@ func read(
 		}
 	}()
 
-	if processErr != nil {
-		// Drain anything remaining
-		drainResultChan(results)
-		return nil, processErr
-	}
-
```
robskillington: Is results channel still a thing?  If so I think we do have to drain it unless the publisher stops publishing correctly.

(Have seen leaked of this kind where publisher to the channel is blocked and the reader has aborted, causing mem leak since publisher never completes)

src/query/benchmark/benchmarker/main/benchmarker.go
```diff
@@ -31,6 +31,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/m3db/m3/src/x/instrument"
+
```
robskillington: nit: Should not be a new line inbetween these imports.

src/query/api/v1/handler/prometheus/native/read_common.go
arnikola: Nice yeah misread this as closing the result channel I removed rather than the internal chan

PR #1784, opened 2019-06-30
[query] Batch remote server responses

### LINKED ISSUE #1730: m3 Coordinator remote error
# M3 Coordinator Multi-Zone Remote ERROR

## Env

tow clusters

* Cluster 1:  3 Node
* Cluster 2:  3 Seed Node,  3 Data Node

### Coordinator ERROR Message

coordinator 1:

```
019-06-13T05:56:19.815Z	ERROR	unable to fetch data	{"rqID": "c7b5bc92-1494-484d-9d75-a8835c7e2376", "error": "rpc error: code = ResourceExhausted desc = grpc: received message larger than max (73359289 vs. 4194304)"}
``` 

coordinator 2:

```
2019-06-13T06:02:56.820Z	ERROR	unable to send fetch result	{"rqID": "5fb3939b-4581-422f-a021-e25ef043d7e7", "error": "rpc error: code = DeadlineExceeded desc = context deadline exceeded"}
```

### The Remote Server

```go
// github.com/m3db/m3/src/query/tsdb/remote/server.go#L122
err = stream.Send(response)
if err != nil {
     logger.Error("unable to send fetch result", zap.Error(err))
}
```
get the response data too long time.

### The RPC Remote Client 
```
// github.com/m3db/m3/src/query/tsdb/remote/client.go#L133

```

arnikola:
Fixes #1730

```release-note
none
```

```documentation-note
none
```


src/query/tsdb/remote/server.go
```diff
@@ -107,18 +118,27 @@ func (s *grpcServer) Fetch(
...
-		logger.Error("unable to send fetch result", zap.Error(err))
+	size := min(defaultBatch, len(results))
+	for ; len(results) >= size; results = results[size:] {
+		response := &rpc.FetchResponse{
+			Series: results[:size],
+		}
+
+		err = stream.Send(response)
+		if err != nil {
+			logger.Error("unable to send fetch result", zap.Error(err))
```
robskillington: You will want to `return err` here if send fails, it's quite likely the connection is broken and should just terminate the send.

src/query/tsdb/remote/server.go
```diff
@@ -153,18 +173,23 @@ func (s *grpcServer) Search(
...
+			logger.Error("unable to encode search result", zap.Error(err))
+			return err
+		}
 
-	err = stream.SendMsg(response)
-	if err != nil {
-		logger.Error("unable to send search result", zap.Error(err))
+		err = stream.Send(response)
+		if err != nil {
+			logger.Error("unable to send search result", zap.Error(err))
```
robskillington: Here too, should `return err` here if send fails, it's quite likely the connection is broken and should just terminate the send.

src/query/tsdb/remote/server.go
```diff
@@ -185,22 +210,31 @@ func (s *grpcServer) CompleteTags(
...
+			CompletedTags:    tags,
+		}
+
+		response, err := encodeToCompressedCompleteTagsResult(results)
+		if err != nil {
+			logger.Error("unable to encode complete tags result", zap.Error(err))
+			return err
+		}
+
+		err = stream.SendMsg(response)
```
robskillington: Hm, think you want the type-safe `stream.Send(response)` here.

src/query/tsdb/remote/server.go
```diff
@@ -185,22 +210,31 @@ func (s *grpcServer) CompleteTags(
...
+		if err != nil {
+			logger.Error("unable to send complete tags result", zap.Error(err))
```
robskillington: Here too, should `return err` here if send fails, it's quite likely the connection is broken and should just terminate the send.

src/query/storage/m3/storage.go
```diff
@@ -380,6 +383,33 @@ func (s *m3storage) CompleteTags(
 		wg              sync.WaitGroup
 	)
 
+	if err != nil {
+		return nil, err
+	}
```
robskillington: Hm think the error above is already checked, maybe this was a side effect of a merge?

PR #1785, opened 2019-07-01
[query] Update prometheus.md

bianap:
tagOptions and idScheme keys are missing from the example configuration based on m3coordinator-cluster-template.yml.

Updating the prometheus.md, adding the missing part of the configuration. 

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/docs/integrations/prometheus.md
+++ b/docs/integrations/prometheus.md
@@ -29,6 +29,9 @@ metrics:
   sanitization: prometheus
   samplingRate: 1.0
   extended: none
+  
+  tagOptions:
+  idScheme: quoted
 
 clusters:
    - namespaces:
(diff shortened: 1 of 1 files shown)


docs/integrations/prometheus.md
```diff
@@ -29,6 +29,9 @@ metrics:
   sanitization: prometheus
   samplingRate: 1.0
   extended: none
+  
+  tagOptions:
```
benraskin92: I think `tagOptions` is indented one too far.

docs/integrations/prometheus.md
```diff
@@ -29,6 +29,9 @@ metrics:
...
+  idScheme: quoted
```
benraskin92: This should be under `tagOptions`.

docs/integrations/prometheus.md
bianap: Thank you, I corrected it in a new commit.

docs/integrations/prometheus.md
bianap: Thank you, I corrected it in a new commit.

Issue #1747, opened 2019-06-18
Make namespace_readers file-volume-aware


justinjc:
`namespace_readers.go` will need to be aware of fileset volumes in order for peer bootstrapping to function correctly, since we need to keep track of which volume peers were reading from.

These places the same TODO string so that it's easy to find: `get the actual volume here`.

PR #1790, opened 2019-07-03
[dbnode] Cold flush integration tests

justinjc:

Adds integration tests for cold writes.

```release-note
NONE
```

```documentation-note
NONE
```


scripts/docker-integration-tests/cold_write_no_index/test.sh
```diff
@@ -0,0 +1,123 @@
+#!/usr/bin/env bash
```
robskillington: super nit: Maybe we call this just `cold_writes_simple/...`

scripts/docker-integration-tests/cold_write_no_index/test.sh
```diff
@@ -0,0 +1,123 @@
...
+            "id": "m3db_local",
+            "isolation_group": "rack-a",
+            "zone": "embedded",
+            "weight": 1024,
+            "endpoint": "127.0.0.1:9000",
+            "hostname": "127.0.0.1",
+            "port": 9000
+        }
+    ]
+}'
```
robskillington: Can you use the `/database/create` API instead of namespace and placement btw?  

Trying to migrate most docs and tests to use that instead (so more users use it and have less issues with block size, etc).

Is there a way to create database without index turned on with the `/database/create` endpoint?  I guess that would be the only potential issue.

scripts/docker-integration-tests/cold_write_no_index/test.sh
```diff
@@ -0,0 +1,123 @@
...
+echo "Waiting until shards are marked as available"
+ATTEMPTS=10 TIMEOUT=1 retry_with_backoff  \
+  '[ "$(curl -sSf 0.0.0.0:7201/api/v1/placement | grep -c INITIALIZING)" -eq 0 ]'
+
+echo "Write data before now minus buffer past"
+curl -vvvsS -X POST 0.0.0.0:9003/write -d '{
+  "namespace": "unagg",
+  "id": "foo",
+  "datapoint": {
+    "timestamp":'"$(($(date +"%s") - 60 * 20))"',
```
robskillington: Maybe for good measure write `blockSize * 2` so it's definitely not in the current block?

scripts/docker-integration-tests/cold_write_no_index/test.sh
```diff
@@ -0,0 +1,123 @@
...
+    "value": 12.3456789
+  }
+}'
+
+echo "Read data (not using index)"
```
robskillington: Before we read this data back, maybe it's best to stop then start the node (restart) so we can be sure it's being read from disk?

PS what's the cold flush semantics here, I guess if a cold flush doesn't happen in time, it'll be loaded back into memory from the commit log bootstrapper?

src/dbnode/integration/disk_coldflush_test.go
```diff
@@ -0,0 +1,217 @@
...
+	maxWaitTime := time.Minute
+	require.NoError(t, waitUntilFileSetFilesFlushed(filePathPrefix, expectedDataFiles, maxWaitTime))
+
+	// Verify on-disk data match what we expect.
+	verifyFlushedDataFiles(t, testSetup.shardSet, testSetup.storageOpts, nsID, seriesMaps)
+
+	coldData := []generate.BlockConfig{
+		{IDs: []string{"cold0"}, NumPoints: 80, Start: start.Add(-blockSize)},
+		{IDs: []string{"cold1", "cold2", "cold3"}, NumPoints: 30, Start: start},
+		{IDs: []string{"cold1", "cold3"}, NumPoints: 20, Start: start.Add(blockSize)},
```
robskillington: Want to do 2*blocksize for good measure (instead of just single block size?)

src/dbnode/integration/fs_bootstrap_volumed_test.go
```diff
@@ -0,0 +1,143 @@
...
+	if updateInputConfig != nil {
+		updateInputConfig(allInputData)
+	}
+	seriesMaps1 := generate.BlocksByStart(inputData1)
+	seriesMaps2 := generate.BlocksByStart(inputData2)
+	seriesMaps3 := generate.BlocksByStart(inputData3)
+	allSeriesMaps := generate.BlocksByStart(allInputData)
+	require.NoError(t, writeTestDataToDisk(ns1, setup, seriesMaps1, 0))
+	// Write non-zero volumes to disk to simulate a past cold flush.
+	require.NoError(t, writeTestDataToDisk(ns1, setup, seriesMaps2, 1))
```
robskillington: This is interesting, won't you have two files on disk now for same block start? i.e. volume 0 and volume 1 when the node starts up. 

Does it correctly delete volume 0 eventually?

scripts/docker-integration-tests/cold_write_no_index/test.sh
justinjc: Unfortunately it's not one of the options for the `/create` endpoint, which should probably stay that way, since that endpoint should contain only the recommended options in there.

scripts/docker-integration-tests/cold_write_no_index/test.sh
justinjc: Cold writes can still happen within the current block. They're cold writes as long as they are before `now - bufferPast`, so I think it's worth keeping this here.

I'll include a write for `now - blockSize * 2` as well though.

scripts/docker-integration-tests/cold_write_no_index/test.sh
justinjc: I think it's worth testing the simple cold write scenario where we just read cold writes from memory as well, so I'll keep this read here.

As mentioned above I'll add another write for `now - blockSize * 2` and then stop and start to test the disk read scenario.

And yep, if a cold flush doesn't happen in time, it goes through the commit log bootstrapper.

src/dbnode/integration/disk_coldflush_test.go
justinjc: I'll set `now` to `start + 3 * blockSize` to ensure these are cold writes.

scripts/docker-integration-tests/cold_write_no_index/test.sh
robskillington: Yeah, I'm inclined to agree.

src/dbnode/integration/disk_coldflush_test.go
robskillington: Sounds good.

scripts/docker-integration-tests/cold_write_no_index/test.sh
robskillington: Sounds good.

src/dbnode/integration/fs_bootstrap_volumed_test.go
justinjc: Hmm, in debugging this, I realized this test was incorrect, since the multiple `writeTestDataToDisk` doesn't handle any merge logic and so it was losing data... I'll revisit this test another time.

However, this debugging also uncovered a bug when bootstrapping the cold flush version, which I have fixed and added a unit test for.

src/dbnode/storage/shard.go
```diff
@@ -1867,15 +1867,21 @@ func (s *dbShard) Bootstrap(
...
-		s.markWarmFlushStateSuccess(at)
-		// Cold version needs to get bootstrapped so that the 1:1 relationship between volume number
-		// and cold version is maintained and the volume numbers / flush versions remain monotonically
-		// increasing.
-		s.setFlushStateColdVersion(at, info.VolumeIndex)
+		// Cold version needs to get bootstrapped so that the 1:1 relationship
+		// between volume number and cold version is maintained and the volume
+		// numbers / flush versions remain monotonically increasing.
+		//
+		// Note that there can be multiple info files for the same block, for
```
richardartoul: 🤦‍♂ 

Sorry about that. Good catch...

src/dbnode/integration/disk_coldflush_test.go
```diff
@@ -0,0 +1,218 @@
...
+	"github.com/m3db/m3/src/dbnode/integration/generate"
+	"github.com/m3db/m3/src/dbnode/namespace"
+	"github.com/m3db/m3/src/dbnode/persist/fs"
+	"github.com/m3db/m3/src/x/ident"
+	xtime "github.com/m3db/m3/src/x/time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestDiskColdFlushSimple(t *testing.T) {
```
richardartoul: Can we also have a test that involves restarting a node to ensure all the cold flush bootstrap logic works too?

richardartoul: Also can we add an integration test for the case where you have a cold write in the commitlog for a block that has already been warm flushed and make sure that works

src/dbnode/storage/shard.go
justinjc: This merged separately in https://github.com/m3db/m3/pull/1819

src/dbnode/integration/disk_coldflush_test.go
justinjc: This is covered in the included test `commitlog_bootstrap_coldwrites_test.go`, but it's currently always skipped because it won't work until #1747 gets resolved.

PR #1793, opened 2019-07-05
[query] Allow lookback duration to be set in query parameters

robskillington:

This allows callers to set lookback as a request param for per-request specification of the lookback for a query.

```release-note
NONE
```

```documentation-note
NONE
```


src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -52,6 +52,7 @@ const (
 	debugParam        = "debug"
 	endExclusiveParam = "end-exclusive"
 	blockTypeParam    = "block-type"
+	lookbackParam     = "lookback"
```
benraskin92: Maybe 'lookbackDur' or something similar so it's a little easier to grok what it is?

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -255,6 +264,26 @@ func parseQuery(r *http.Request) (string, error) {
 	return queries[0], nil
 }
 
+func parseLookbackDuration(r *http.Request, step time.Duration) (time.Duration, bool, error) {
+	lookback := r.FormValue(lookbackParam)
+	if lookback == "" {
+		return 0, false, nil
+	}
+
+	if lookback == stepParam {
```
benraskin92: Curious if lookbackDuration is > stepSize should it default to stepSize?

arnikola: I think this is for the special case of `&lookback=step`; although it reads a little weird since it's being compared to a variable named `xxxParam`

src/query/api/v1/handler/prometheus/native/common.go
```diff
@@ -255,6 +264,26 @@ func parseQuery(r *http.Request) (string, error) {
...
+		return step, true, nil
+
+	}
+	// Otherwise it is specified as duration value.
+	value, err := parseDuration(r, lookbackParam)
+	if err != nil {
+		return 0, false, err
+	}
+
+	return value, true, nil
```
arnikola: Might have to check that `step > 0` otherwise you may have a bad time (although that case is handled elsewhere so it shouldn't panic, but might be better to catch it early)

robskillington: Sure thing will do.

src/query/api/v1/handler/fetch_options.go
```diff
@@ -105,3 +136,81 @@ func newOrExistingRestrictFetchOptions(
...
+	if str == "" {
+		return 0, errors.ErrNotFound
+	}
+
+	value, durationErr := time.ParseDuration(str)
+	if durationErr == nil {
+		return value, nil
+	}
+
+	// Try parsing as a float value specifying seconds, the Prometheus default
```
arnikola: meganit: `.` at end of comment

src/query/api/v1/handler/fetch_options.go
```diff
@@ -105,3 +136,81 @@ func newOrExistingRestrictFetchOptions(
...
+	seconds, floatErr := strconv.ParseFloat(str, 64)
+	if floatErr == nil {
+		ts := seconds * float64(time.Second)
+		if ts > maxInt64 || ts < minInt64 {
```
arnikola: Should the lower bound just check `ts <= 0`?

src/query/api/v1/handler/fetch_options.go
```diff
@@ -105,3 +136,81 @@ func newOrExistingRestrictFetchOptions(
...
+// ParseLookbackDuration parses a lookback duration for an HTTP request.
+func ParseLookbackDuration(r *http.Request) (time.Duration, bool, error) {
+	lookback := r.FormValue(LookbackParam)
+	if lookback == "" {
+		return 0, false, nil
+	}
+
+	if lookback == StepParam {
+		// Use the step size as lookback.
+		step, err := parseStep(r, StepParam)
```
arnikola: Maybe should explicitly add a sanity check for `step > 0`? It should fail already on `ParseStep` but sanity check here may be useful

src/query/api/v1/handler/fetch_options_test.go
```diff
@@ -103,6 +115,25 @@ func TestFetchOptionsBuilder(t *testing.T) {
...
+		{
+			name:  "can set lookback duration",
+			query: "lookback=10s",
+			expectedLookback: &expectedLookback{
+				value: 10 * time.Second,
+			},
+		},
+		{
+			name:  "can set lookback duration based on step",
+			query: "lookback=step&step=10s",
```
arnikola: May be worth adding in `lookback=step`, `lookback=step&step=invalid` , `lookback=step&step=-1` cases here?

src/query/tsdb/remote/codecs_test.go
```diff
@@ -206,6 +209,9 @@ func TestEncodeDecodeFetchQuery(t *testing.T) {
 		MetricsType:   storage.AggregatedMetricsType,
 		StoragePolicy: policy.MustParseStoragePolicy("1m:14d"),
 	}
+	lookback := time.Minute
+	fetchOpts.LookbackDuration = &lookback
```
arnikola: Any worth in adding one without a specified lookback to ensure it uses the config one correctly?

src/query/api/v1/handler/fetch_options.go
robskillington: Yup so I added that inside `parseStep`

src/query/api/v1/handler/fetch_options_test.go
robskillington: Sure thing.

src/query/api/v1/handler/fetch_options.go
robskillington: This was existing code, I could update though.

src/query/executor/engine.go
```diff
@@ -113,21 +113,21 @@ func (e *engine) Execute(
...
-	fetchOpts := storage.NewFetchOptions()
-	fetchOpts.Limit = opts.QueryContextOptions.LimitMaxTimeseries
 	return e.opts.Store().Fetch(ctx, query, fetchOpts)
 }
 
 func (e *engine) ExecuteExpr(
 	ctx context.Context,
 	parser parser.Parser,
 	opts *QueryOptions,
+	fetchOpts *storage.FetchOptions,
```
arnikola: nit; set opts as last param here

src/query/api/v1/handler/fetch_options.go
arnikola: Oops, I'm blind

PR #1794, opened 2019-07-05
[dbnode] Add lookback duration integration test

robskillington:

Tests request specific lookback duration for queries in `Prometheus` integration test.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/scripts/docker-integration-tests/prometheus/test.sh
+++ b/scripts/docker-integration-tests/prometheus/test.sh
@@ -179,6 +179,24 @@ function test_query_restrict_metrics_type {
     retry_with_backoff prometheus_query_native
 }
 
+function test_query_lookback_duration {
+  now=$(date +"%s")
+  hour_ago=$(expr $now - 3600) 
+  step="30s"
+  jq_path_range=".data.result[0].values[] | length"
+  
+  # Test changing the lookback duration
+  echo "Test lookback causes steps to return 5 values (5m/60s = 5)"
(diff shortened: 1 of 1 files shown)


scripts/docker-integration-tests/prometheus/test.sh
```diff
@@ -179,10 +179,37 @@ function test_query_restrict_metrics_type {
...
+    retry_with_backoff prometheus_query_native
+
+  # Test setting the lookback aligned with the step size result in single value
+  echo "Test lookback causes steps to return 1 value (lookback 5s / step 5s = 1)"
+  params_range="start=${hour_ago}"'&'"end=${hour_future}"'&'"step=5s&lookback=step"
+  ATTEMPTS=50 TIMEOUT=2 MAX_TIMEOUT=4 \
+    endpoint=query_range query="$METRIC_NAME_TEST_RESTRICT_WRITE" params="$params_range" \
+    metrics_type="unaggregated" jq_path="$jq_path_range" expected_value="1" \
+    retry_with_backoff prometheus_query_native
+}
```
benraskin92: Worth having a test where `lookback < step`?

robskillington: That will always return no datapoints heh, wonder if we should make that an error, although inclined not to in case there is a use case where its useful (with certain aggregations or such).

arnikola: I think that may actually return a single point similarly to `lookback=step` case ; might look into it to double check what's happening if you're not getting anything

PR #1795, opened 2019-07-07
Check config for deprecated fields

### LINKED ISSUE #1549: Programmatically warn on deprecated config fields
We currently have a few config fields that are deprecated that we may manually emit warnings on if we see them. It would be convenient if we could add a struct tag or something similar to accomplish the goal of tagging deprecated config fields and warning on them. This is especially relevant as we have strict config parsing, wherein unknown fields fail to parse.

obitech:

Fixes #1549

I tried to cover all services that require a config, let me know if I forgot any.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/x/config/config_test.go
+++ b/src/x/config/config_test.go
@@ -345,13 +345,13 @@ multiple:
 		}
 		require.True(
 			t,
-			slicesCotainSameStrings(expect, actual),
+			slicesContainSameStrings(expect, actual),
 			fmt.Sprintf("expect %#v should be equal actual %#v", expect, actual),
 		)
 	})
 }
 
-func slicesCotainSameStrings(s1, s2 []string) bool {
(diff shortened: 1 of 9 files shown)


obitech:
cc @schallert

src/x/config/config_test.go
```diff
@@ -213,6 +243,131 @@ func TestLoadFilesValidateOnce(t *testing.T) {
...
+		}
+		require.True(
+			t,
+			slicesCotainSameStrings(expect, actual),
+			fmt.Sprintf("expect %#v should be equal actual %#v", expect, actual),
+		)
+	})
+}
+
+func slicesCotainSameStrings(s1, s2 []string) bool {
```
schallert: Nit: `s/Cotain/Contain/`

obitech: Fixed

PR #1802, opened 2019-07-11
[aggregator] aggregator with pass-through

benraskin92:

This PR introduces Passthrough functionality in the aggregator. 

```release-note

```

```documentation-note

```


src/aggregator/aggregator/aggregator.go
```diff
@@ -57,6 +58,7 @@ var (
 	errInvalidMetricType             = errors.New("invalid metric type")
 	errActivePlacementChanged        = errors.New("active placement has changed")
 	errShardNotOwned                 = errors.New("aggregator shard is not owned")
+	errPassThroughWriterNotDefined   = errors.New("pass-through writer is not defined")
```
benraskin92: `pass-through` => `passthrough`

src/aggregator/aggregator/aggregator.go
```diff
@@ -232,6 +239,46 @@ func (agg *aggregator) AddForwarded(
...
 
+func (agg *aggregator) AddPassThrough(
+	metric aggregated.Metric,
+	metadata metadata.TimedMetadata,
+) error {
+	callStart := agg.nowFn()
+	agg.metrics.passThrough.Inc(1)
+	agg.RLock()
+	if agg.state != aggregatorOpen {
+		agg.metrics.addPassThrough.ReportError(errAggregatorAlreadyOpenOrClosed)
```
benraskin92: unlock here too?

src/aggregator/aggregator/aggregator.go
```diff
@@ -912,24 +981,27 @@ func newAggregatorMetrics(
...
-		shards:       newAggregatorShardsMetrics(shardsScope),
-		shardSetID:   newAggregatorShardSetIDMetrics(shardSetIDScope),
-		tick:         newAggregatorTickMetrics(tickScope),
+		counters:       scope.Counter("counters"),
+		timers:         scope.Counter("timers"),
+		timerBatches:   scope.Counter("timer-batches"),
+		gauges:         scope.Counter("gauges"),
+		forwarded:      scope.Counter("forwarded"),
+		timed:          scope.Counter("timed"),
+		passThrough:    scope.Counter("pass-through"),
```
benraskin92: `passthrough`

src/aggregator/aggregator/capture/aggregator.go
```diff
@@ -131,6 +132,26 @@ func (agg *aggregator) AddForwarded(
...
+func (agg *aggregator) AddPassThrough(
+	metric aggregated.Metric,
+	metadata metadata.TimedMetadata,
+) error {
+	// Clone the metric and timed metadata to ensure it cannot be mutated externally.
+	metric = cloneTimedMetric(metric)
+	metadata = cloneTimedMetadata(metadata)
+
+	agg.Lock()
+	defer agg.Unlock()
```
benraskin92: i would say avoid the defer and just unlock below

src/aggregator/aggregator/handler/writer/mutex.go
```diff
@@ -0,0 +1,57 @@
...
+// NewMutexWriter creates a writer with a lock
+func NewMutexWriter(w Writer) Writer {
+	return &mutexWriter{
+		w: w,
+	}
+}
+
+func (mw *mutexWriter) Write(mp aggregated.ChunkedMetricWithStoragePolicy) error {
+	mw.Lock()
+	defer mw.Unlock()
```
benraskin92: I would avoid using defer here.

src/aggregator/server/m3msg/options.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+// NewServerOptions creates a new set of m3msg server options.
+func NewServerOptions(
+	iOpts instrument.Options,
+	m3msgConfiguration *coordinatorM3msg.Configuration,
+) (string, Options) {
+	var addr string
+	if m3msgConfiguration != nil {
+		addr = m3msgConfiguration.Server.ListenAddress
```
benraskin92: Should you check to make sure `m3msgConfiguration.Server.ListenAddress` isn't `""`? Maybe have a `Validate()` method on the options?

src/aggregator/server/m3msg/server.go
```diff
@@ -0,0 +1,99 @@
...
+) m3msg.WriteFn {
+	return func(
+		ctx context.Context,
+		id []byte,
+		metricNanos, encodeNanos int64,
+		value float64,
+		sp policy.StoragePolicy,
+		callback m3msg.Callbackable,
+	) {
+		// The type of a pass-through metric does not matter as it is written directly into m3db
```
benraskin92: nit: `.` at end of comments.

src/aggregator/server/m3msg/server.go
```diff
@@ -0,0 +1,99 @@
...
+				log.Infof("[FAIL] to write pass-through metric %s with metadata ID=%s SP=%s",
+					metric.String(),
+					metadata.AggregationID.String(),
+					metadata.StoragePolicy.String(),
+				)
+			}
+			callback.Callback(m3msg.OnRetriableError)
+		}
+
+		if s != nil && s.Sample() && log != nil {
```
benraskin92: Do you think we should continue to log successes?

src/aggregator/server/m3msg/server_test.go
```diff
@@ -0,0 +1,138 @@
...
+	"github.com/m3db/m3/src/aggregator/aggregator"
+	xm3msg "github.com/m3db/m3/src/cmd/services/m3coordinator/server/m3msg"
+	"github.com/m3db/m3/src/metrics/aggregation"
+	"github.com/m3db/m3/src/metrics/metadata"
+	"github.com/m3db/m3/src/metrics/metric"
+	"github.com/m3db/m3/src/metrics/metric/aggregated"
+	"github.com/m3db/m3/src/metrics/metric/unaggregated"
+	"github.com/m3db/m3/src/metrics/policy"
+	xtime "github.com/m3db/m3/src/x/time"
+	"github.com/stretchr/testify/require"
```
benraskin92: Move to new import group.

src/aggregator/server/m3msg/server_test.go
```diff
@@ -0,0 +1,138 @@
...
+		callback,
+	)
+
+	require.Equal(t, map[xm3msg.CallbackType]int{xm3msg.OnSuccess: 1}, callback.record)
+	require.Equal(t, 1, aggregator.count)
+	require.Equal(t, expectedMetric, *aggregator.receivedMetrics[0])
+	require.Equal(t, expectedMetadata, *aggregator.receivedMetadata[0])
+}
+
+type mockAggregator struct {
```
benraskin92: Can you just generate the mock automatically instead of doing it manually like this?

src/cmd/services/m3aggregator/config/aggregator.go
```diff
@@ -286,6 +290,25 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
 	}
 	opts = opts.SetFlushHandler(flushHandler)
 
+	// Set pass-through handler.
+	passThroughScope := scope.SubScope("pass-through-handler")
```
benraskin92: `passthrough-handler`

src/aggregator/aggregator/capture/aggregator.go
fishie9: Using defer is consistent with other AddXxxx functions.

src/aggregator/server/m3msg/server.go
fishie9: I am fine with either way... we can use a lower sample rate if necessary.

glide.lock
```diff
@@ -126,7 +126,7 @@ imports:
 - name: github.com/ghodss/yaml
   version: 0ca9ea5df5451ffdf184b4428c902747c2c11cd7
 - name: github.com/go-kit/kit
-  version: fbab14bd4c487151b206c2a41ecfbb015e32c1e0
+  version: da68e7640663097af890b9bdd76da9cb816ebb82
```
prateek: weird to see glide.lock changes without glide.yaml, wanna just revert to old versions?

src/cmd/services/m3aggregator/serve/serve.go
```diff
@@ -54,6 +62,21 @@ func Serve(
 	defer httpServer.Close()
 	log.Infof("http server: listening on %s", httpAddr)
 
+	if m3msgServerOpts != nil {
+		if m3msgConfig := m3msgServerOpts.M3msgConfiguration(); m3msgConfig != nil {
+			s, _ := sampler.NewSampler(defaultSampleRate)
```
prateek: need to check the error here

src/aggregator/aggregator/options.go
```diff
@@ -591,6 +599,17 @@ func (o *options) FlushHandler() handler.Handler {
...
 }
 
+func (o *options) SetPassThroughWriter(value writer.Writer) Options {
+	opts := *o
+	opts.passThroughWriter = value
+	return &opts
+}
+
+func (o *options) PassThroughWriter() writer.Writer {
+	//TODO: (fishie9) check add buffer if necessary
```
prateek: still relevant? if so, please change to ```// TODO(fishie9): ...`

src/aggregator/integration/setup.go
```diff
@@ -249,6 +249,8 @@ func (ts *testServerSetup) startServer() error {
 			ts.rawTCPServerOpts,
 			ts.httpAddr,
 			ts.httpServerOpts,
+			"",
```
prateek: would be good to add an integration test that checked the new code path too

src/aggregator/aggregator/aggregator.go
prateek: +1, this needs to release the lock before returning

src/aggregator/aggregator/aggregator.go
```diff
@@ -232,6 +239,46 @@ func (agg *aggregator) AddForwarded(
...
+	callStart := agg.nowFn()
+	agg.metrics.passThrough.Inc(1)
+	agg.RLock()
+	if agg.state != aggregatorOpen {
+		agg.metrics.addPassThrough.ReportError(errAggregatorAlreadyOpenOrClosed)
+		return errAggregatorNotOpenOrClosed
+	}
+	agg.RUnlock()
+
+	if agg.passThroughWriter == nil {
```
prateek: The way the code is right now, using this var without a (r)lock is racy? e.g. if concurrent `AddPassThrough(...)` and `Close()` calls can interleave to close the writer before this line. 

would be good to add a test to capture this too.

src/aggregator/aggregator/handler/writer/mutex.go
prateek: +1, might as well for `Flush()` and `Close()` too while you're at it

src/aggregator/aggregator/handler/writer/mutex.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/m3db/m3/src/metrics/metric/aggregated"
+)
+
+type mutexWriter struct {
+	sync.Mutex
+	w Writer
+}
+
+// NewMutexWriter creates a writer with a lock
```
prateek: nit: end the comment with a period.

src/aggregator/aggregator/handler/writer/mutex.go
```diff
@@ -0,0 +1,57 @@
...
+	return mw.w.Write(mp)
+}
+
+func (mw *mutexWriter) Flush() error {
+	mw.Lock()
+	defer mw.Unlock()
+	return mw.w.Flush()
+}
+
+func (mw *mutexWriter) Close() error {
```
prateek: any concern about a double Close or is that responsibility of the caller?

src/aggregator/server/m3msg/options.go
```diff
@@ -0,0 +1,81 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package m3msg
+
+import (
+	coordinatorM3msg "github.com/m3db/m3/src/cmd/services/m3coordinator/server/m3msg"
```
prateek: nit: weird to have this dependency from agg -> coordinator. mind moving the common `m3msg` package into `src/msg/server`

src/aggregator/server/m3msg/options.go
```diff
@@ -0,0 +1,81 @@
...
+	SetInstrumentOptions(value instrument.Options) Options
+
+	// InstrumentOptions returns the instrument options.
+	InstrumentOptions() instrument.Options
+
+	// SetM3msgConfiguration sets the m3msg configuration used by the server.
+	SetM3msgConfiguration(value *coordinatorM3msg.Configuration) Options
+
+	// M3msgConfiguration returns the m3msg configuration used by the server.
+	M3msgConfiguration() *coordinatorM3msg.Configuration
```
prateek: nit: weird casing right now, rename to `M3MsgConfiguration` or `M3MSGConfiguration`

src/aggregator/server/m3msg/options.go
```diff
@@ -0,0 +1,81 @@
...
+type options struct {
+	instrumentOpts     instrument.Options
+	m3msgConfiguration *coordinatorM3msg.Configuration
+}
+
+// NewServerOptions creates a new set of m3msg server options.
+func NewServerOptions(
+	iOpts instrument.Options,
+	m3msgConfiguration *coordinatorM3msg.Configuration,
+) (string, Options) {
```
prateek: feels weird to have the string and the Options here, any reason to not make the `ListenAddress` a part of Options()?

src/aggregator/server/m3msg/options.go
prateek: +1 for the Validate() method, and ensure the any users of the Options call it.

src/cmd/services/m3aggregator/serve/serve.go
```diff
@@ -54,6 +62,21 @@ func Serve(
 	defer httpServer.Close()
 	log.Infof("http server: listening on %s", httpAddr)
 
+	if m3msgServerOpts != nil {
```
prateek: The code **never** skips this `if` (the way its structured right now)

The options passed to this are created by the following line in `src/cmd/services/m3aggregator/main/main.go `:

```
m3msgAddr, m3msgServerOpts := m3msg.NewServerOptions(iOpts, cfg.M3msg)
```

Which is the following method (from `src/aggregator/server/m3msg/options.go`): 

```
// NewServerOptions creates a new set of m3msg server options.
func NewServerOptions(
	iOpts instrument.Options,
	m3msgConfiguration *coordinatorM3msg.Configuration,
) (string, Options) {
	var addr string
	if m3msgConfiguration != nil {
		addr = m3msgConfiguration.Server.ListenAddress
	}
	return addr, &options{
		instrumentOpts:     iOpts,
		m3msgConfiguration: m3msgConfiguration,
	}
}
```

i.e. options are never nil. Would be better to add an `Enabled()` on the options to indicate whether the server is actually enabled.

prateek: Hm actually, is the m3msg server supposed to be optional?

src/cmd/services/m3aggregator/config/aggregator.go
```diff
@@ -286,6 +290,25 @@ func (c *AggregatorConfiguration) NewAggregatorOptions(
...
+	}
+	passThroughHandler, err := c.Flush.NewHandler(client, iOpts)
+	if err != nil {
+		return nil, err
+	}
+	passThroughWriter, err := passThroughHandler.NewWriter(passThroughScope)
+	if err != nil {
+		return nil, err
+	}
+	opts = opts.SetPassThroughWriter(writer.NewMutexWriter(passThroughWriter))
```
prateek: i'm probably missing something here, but what's the point of wrapping the `passThroughWriter` with a `writer.NewMutexWriter(...)`?

src/aggregator/server/m3msg/server.go
```diff
@@ -0,0 +1,99 @@
...
+			Value:     value,
+		}
+		metadata := metadata.TimedMetadata{
+			AggregationID: aggregation.MustCompressTypes(aggregation.Last),
+			StoragePolicy: sp,
+		}
+
+		if err := aggregator.AddPassThrough(metric, metadata); err != nil {
+			if log != nil {
+				log.Infof("[FAIL] to write pass-through metric %s with metadata ID=%s SP=%s",
```
prateek: nit: no reason to use the SugaredLogger here. can you make this a regular `*zap.Logger`?

src/aggregator/server/m3msg/server.go
prateek: Best to remove it before landing.

src/aggregator/server/m3msg/server.go
```diff
@@ -0,0 +1,99 @@
...
+			TimeNanos: metricNanos,
+			Value:     value,
+		}
+		metadata := metadata.TimedMetadata{
+			AggregationID: aggregation.MustCompressTypes(aggregation.Last),
+			StoragePolicy: sp,
+		}
+
+		if err := aggregator.AddPassThrough(metric, metadata); err != nil {
+			if log != nil {
```
prateek: could you force callers to provide you a logger too? they can provide a noopLogger if they don't care

src/aggregator/server/m3msg/server.go
```diff
@@ -0,0 +1,99 @@
...
+	"github.com/m3db/m3/src/metrics/metric"
+	"github.com/m3db/m3/src/metrics/metric/aggregated"
+	"github.com/m3db/m3/src/metrics/policy"
+	"github.com/m3db/m3/src/x/sampler"
+
+	"go.uber.org/zap"
+)
+
+// GetWriteFn returns the m3msg write function for pass-through metrics
+func GetWriteFn(
```
prateek: nit: Considering it's a ctor, `NewWriteFn()` would be a more go-y name

prateek: actually, considering this is only for PassThrough metrics, maybe `NewPassThroughWriteFn` is better?

src/aggregator/aggregator/capture/aggregator.go
prateek: Yep so this aggregator implementation (i.e. the one in `src/aggregator/aggregator/capture`) is only used in tests, so not as critical. Fine to leave as is here but wouldn't do that in the general impl (the one used in prod).

prateek:
hm more a question: can you avoid changes to the aggregator code (ie `src/aggregator/aggregator/...`)? afaict the m3msg `WriteFn` can recv and write stuff to the new writer, only need to ensure doing it that way still lets you handle the writer lifecycle correctly. Would make the code a lot simpler.

src/cmd/services/m3aggregator/serve/serve.go
```diff
@@ -62,19 +62,26 @@ func Serve(
...
-			if err := m3msgServer.ListenAndServe(); err != nil {
-				return fmt.Errorf("could not start m3msg server at %s: %v", m3msgAddr, err)
-			}
-			defer m3msgServer.Close()
-			log.Infof("m3msg server: listening on %s", m3msgAddr)
+	if err := m3msgServerOpts.Validate(); err != nil {
+		return err
+	}
+	if m3msgConfig := m3msgServerOpts.M3MsgConfiguration(); m3msgConfig != nil {
+		s, err := sampler.NewSampler(defaultSampleRate)
```
nerd0: Since the error was omitted previously will this cause any backwards compatibility issues for any users?

src/cmd/services/m3aggregator/config/aggregator.go
```diff
@@ -779,3 +778,91 @@ func setMetricPrefix(
...
+		numShards: count,
+		writers:   writers,
+		locks:     make([]sync.Mutex, count),
+		shardFn:   shardFn,
+	}, nil
+}
+
+func (p *passThroughWriter) Write(mp aggregated.ChunkedMetricWithStoragePolicy) error {
+	shardID := p.shardFn([]byte(mp.ChunkedID.String()), uint32(p.numShards))
+	p.locks[shardID].Lock()
```
nerd0: just to be safe shouldn't the length of the array be checked before using an index? Just paranoia

src/cmd/services/m3aggregator/config/aggregator.go
```diff
@@ -779,3 +778,91 @@ func setMetricPrefix(
...
+	shardFn   sharding.ShardFn
+}
+
+func (c *AggregatorConfiguration) newPassThroughWriter(
+	cs client.Client,
+	iOpts instrument.Options,
+	scope tally.Scope,
+	shardFn sharding.ShardFn,
+) (writer.Writer, error) {
+	// This is a temporary change to use a separate m3msg topic for pass-through metrics during the migration.
```
nerd0: Can this comment be marked with a `TODO` so that it can be found later for removal? Not sure if this will be needed permanently for all users since we don't know the progression of migrations for all users?

src/cmd/services/m3aggregator/config/aggregator.go
```diff
@@ -41,19 +42,24 @@ import (
...
+	m3sync "github.com/m3db/m3/src/x/sync"
+
+	"github.com/uber-go/tally"
 )
 
 var (
 	errNoKVClientConfiguration = errors.New("no kv client configuration")
 	errEmptyJitterBucketList   = errors.New("empty jitter bucket list")
+
+	defaultNumPassThroughWriters = 16
```
nerd0: should this be a `const` ?

src/aggregator/server/m3msg/options.go
```diff
@@ -70,12 +79,19 @@ func (o *options) InstrumentOptions() instrument.Options {
...
 	opts.m3msgConfiguration = value
 	return &opts
 }
 
-func (o *options) M3msgConfiguration() *coordinatorM3msg.Configuration {
+func (o *options) M3MsgConfiguration() *m3msg.Configuration {
 	return o.m3msgConfiguration
 }
+
+func (o *options) Validate() error {
```
nerd0: Should this also validate any required sampling configuration as well since that causes an error later on?

src/aggregator/integration/one_client_pass_through_test.go
```diff
@@ -0,0 +1,147 @@
...
+	placementStore := serverOpts.KVStore()
+	require.NoError(t, setPlacement(placementKey, placementStore, placement))
+
+	// Create server.
+	testServer := newTestServerSetup(t, serverOpts)
+	defer testServer.close()
+
+	// Start the server.
+	log := testServer.aggregatorOpts.InstrumentOptions().Logger()
+	log.Info("test one client sending multiple types of timed metrics")
```
nerd0: curious about log messages within tests are these absolutely needed? Perhaps it would be best to create a series of test cases with names to denote which test is running?

glide.lock
```diff
@@ -267,7 +267,7 @@ imports:
   subpackages:
   - regexp
 - name: github.com/magiconair/properties
-  version: de8848e004dd33dc07a2947b3d76f618a7fc7ef1
+  version: 7757cc9fdb852f7579b24170bcacda2c7471bb6a
```
prateek: do you need this change?

src/aggregator/aggregator/aggregator.go
```diff
@@ -235,6 +242,44 @@ func (agg *aggregator) AddForwarded(
...
 }
 
+func (agg *aggregator) AddPassThrough(
+	metric aggregated.Metric,
+	metadata metadata.TimedMetadata,
+) error {
+	callStart := agg.nowFn()
+	agg.metrics.passThrough.Inc(1)
+
+	if agg.state != aggregatorOpen {
```
prateek: this needs a read lock on the aggregator

PR #1803, opened 2019-07-11
[query] Log config file in query/coordinator

### LINKED ISSUE #1801: Log if using default config in coordinator
We should log if the coordinator is using the default config. Ran into an issue yesterday where the config wasn't getting properly mounted inside the Docker container. However, we didn't notice because the coordinator still started with the default config.

benraskin92:

Fixes #1801

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/query/server/server.go
+++ b/src/query/server/server.go
@@ -130,6 +130,7 @@ func Run(runOpts RunOptions) {
 			fmt.Fprintf(os.Stderr, "unable to load %s: %v", runOpts.ConfigFile, err)
 			os.Exit(1)
 		}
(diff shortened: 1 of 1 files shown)


src/query/server/server.go
```diff
@@ -130,6 +130,7 @@ func Run(runOpts RunOptions) {
 			fmt.Fprintf(os.Stderr, "unable to load %s: %v", runOpts.ConfigFile, err)
 			os.Exit(1)
 		}
+		fmt.Fprintf(os.Stdout, "using %s config file: %v", serviceName, runOpts.ConfigFile)
```
arnikola: newline if cbf

PR #1804, opened 2019-07-11
[query] fix bugs in binary functions, including interaction with scalars

arnikola:
There was an issue with queries such as `foo_bar_biz{qux="qaz"} or vector(1)`; `vector` would not be parsed correctly, and some binary matching logic was incorrect.

```release-note
Fixes a bug in binary and/or/unless functions.
```

```documentation-note
no
```

### DIFF
--- a/src/query/functions/binary/or.go
+++ b/src/query/functions/binary/or.go
@@ -84,10 +84,8 @@ func makeOrBlock(
 		for iterIndex, valueIndex := range indices {
 			if valueIndex >= len(lValues) {
 				values[valueIndex] = rValues[iterIndex]
-			} else {
-				if math.IsNaN(values[valueIndex]) {
-					values[valueIndex] = rValues[iterIndex]
-				}
+			} else if math.IsNaN(values[valueIndex]) {
+				values[valueIndex] = rValues[iterIndex]
 			}
 		}
 
--- a/src/query/functions/binary/or.go
+++ b/src/query/functions/binary/or.go
@@ -21,13 +21,15 @@
 package binary
 
 import (
+	"math"
+
 	"github.com/m3db/m3/src/query/block"
 	"github.com/m3db/m3/src/query/executor/transform"
 	"github.com/m3db/m3/src/query/models"
 )
 
-// OrType uses all values from left hand side, and appends values from the right hand side which do
-// not have corresponding tags on the right
+// OrType uses all values from left hand side, and appends values from the
+// right hand side which do not have corresponding tags on the right.
 const OrType = "or"
 
 func makeOrBlock(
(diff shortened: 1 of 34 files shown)


src/query/functions/binary/or.go
```diff
@@ -74,34 +100,48 @@ func makeOrBlock(
...
+// mergeIndices returns a slice that maps rhs series to lhs series.
+//
+// NB(arnikola): if the series in the rhs does not exist in the lhs, it is
+// added after all lhs series have been added.
+// This function also combines the series metadatas for the entire block.
+func mergeIndices(
 	matching *VectorMatching,
 	lhs, rhs []block.SeriesMeta,
 ) ([]int, []block.SeriesMeta) {
 	idFunction := HashFunc(matching.On, matching.MatchingLabels...)
```
robskillington: Leaving a self note more than anything, we do need to go revisit using `fnv` hash to guarantee uniqueness for map (there's actually a lot of collisions using `fnv`). The auto-gen map we have will use `xxhash` (faster than fnv) and ensure collisions are treated properly, also without allocating/copying the keys or values if desired.

arnikola: Was sure we were looking at this earlier and were fine with it since that was what Prom was using, but looking into Prom's approach now, they use [xxHash](https://sourcegraph.com/github.com/prometheus/prometheus/-/blob/pkg/labels/labels.go?utm_source=VSCode-1.0.14#L152) so will update to match

src/query/functions/binary/or.go
```diff
@@ -61,11 +70,28 @@ func makeOrBlock(
...
 		lValues := lStep.Values()
-		if err := builder.AppendValues(index, lValues); err != nil {
+		copy(values, lValues)
+		rStep := rIter.Current()
+		rValues := rStep.Values()
+		for iterIndex, valueIndex := range indices {
+			if valueIndex >= len(lValues) {
+				values[valueIndex] = rValues[iterIndex]
+			} else {
+				if math.IsNaN(values[valueIndex]) {
```
robskillington: nit: Perhaps move this into an `else if { ... }` to reduce the next level of nesting?

src/query/functions/binary/unless.go
```diff
@@ -63,15 +62,46 @@ func makeUnlessBlock(
...
+			return nil, errRExhausted
+		}
+
+		lStep := lIter.Current()
+		lValues := lStep.Values()
+		rStep := rIter.Current()
+		rValues := rStep.Values()
+		for iterIndex, valueIndex := range indices {
+			if valueIndex >= 0 {
+				if !math.IsNaN(rValues[iterIndex]) {
```
robskillington: nit: Perhaps join these too?  i.e. `if valueIndex >= 0 && !math.IsNan(...) { ... }`

src/query/models/tags_test.go
```diff
@@ -23,11 +23,12 @@ package models
 import (
 	"bytes"
 	"fmt"
-	"hash/fnv"
 	"reflect"
 	"testing"
 	"unsafe"
 
+	"github.com/cespare/xxhash"
```
robskillington: nit: This import should be at the bottom with the other non-internal imports.

src/query/models/tags.go
```diff
@@ -23,9 +23,10 @@ package models
 import (
 	"bytes"
 	"fmt"
-	"hash/fnv"
 	"sort"
+	"strings"
 
+	"github.com/cespare/xxhash"
```
robskillington: nit: External import should appear below the internal ones separated by a new line.  (suprised linter is not catching these anymore)

arnikola: Yeah just noticed this one; I don't think our linter blocks builds any more, I've been seeing things that used to break it where it hasn't complained pretty often :/

src/query/functions/binary/unless.go
arnikola: Sure, sounds good; generally find it a bit cleaner when they're split up when they have simple conditions, but might be going a bit overboard haha

src/query/functions/binary/unless.go
```diff
@@ -82,24 +112,16 @@ func distinctLeft(
 		leftSigs[idFunction(meta.Tags)] = idx
 	}
 
-	for _, rs := range rhs {
-		// If there's no matching entry in the left-hand side Vector, add the sample.
+	rhsIndices := make([]int, len(rhs))
```
benraskin92: maybe initialize as `make([]int, 0, len(rhs))` and append. Just better to avoid a potential panic.

arnikola: Luckily, was refactoring this stuff already, so it should be fixed :)

PR #1808, opened 2019-07-15
Changing API paths to the current version

manokovacs:

Fixes #

```release-note

```

```documentation-note

```

### DIFF
--- a/docs/operational_guide/namespace_configuration.md
+++ b/docs/operational_guide/namespace_configuration.md
@@ -14,12 +14,12 @@ The operations below include sample cURLs, but you can always review the API doc
 
 #### Recommended (Easy way)
 
-The recommended way to add a namespace to M3DB is to use our `api/v1/database/namespace` endpoint. This API abstracts over a lot of the complexity of configuring a namespace and requires only two pieces of configuration to be provided: the name of the namespace, as well as its retention.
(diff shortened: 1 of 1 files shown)


docs/operational_guide/namespace_configuration.md
```diff
@@ -14,12 +14,12 @@ The operations below include sample cURLs, but you can always review the API doc
 
 #### Recommended (Easy way)
 
-The recommended way to add a namespace to M3DB is to use our `api/v1/database/namespace` endpoint. This API abstracts over a lot of the complexity of configuring a namespace and requires only two pieces of configuration to be provided: the name of the namespace, as well as its retention.
+The recommended way to add a namespace to M3DB is to use our `api/v1/services/m3db/namespace` endpoint. This API abstracts over a lot of the complexity of configuring a namespace and requires only two pieces of configuration to be provided: the name of the namespace, as well as its retention.
```
richardartoul: So this API is actually separate from the `/m3db/namespace/create` API.

`/m3db/namespace/create` API is the "raw" API which allows all aspects of a namespace to be configured

`/database/namespace/create` is a "sugared" API which accepts less parameters and tries to make good decisions / defaults for the other parameters

richardartoul: So basically the other changes look fine but this one should probably stay the same (the other way to do it is documented in the "hard way" section)

PR #1811, opened 2019-07-15
[dbnode] Expose debug dump on /debug/dump

obitech:

Exposes debug dump (https://github.com/m3db/m3/pull/1767) to HTTP endpoint `/debug/dump`

Just wondering if we should make `cpuProfileDuration` (currently set to 1s) and the path (currently set to `/debug/dump`) configurable?

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/dbnode/server/server.go
+++ b/src/dbnode/server/server.go
@@ -297,6 +297,9 @@ func Run(runOpts RunOptions) {
 		cpuProfileDuration,
 		iopts,
 	)
+	if err != nil {
+		logger.Error("unable to create debug writer", zap.Error(err))
+	}
 
 	if cfg.Index.MaxQueryIDsConcurrency != 0 {
 		queryIDsWorkerPool := xsync.NewWorkerPool(cfg.Index.MaxQueryIDsConcurrency)
@@ -572,12 +575,10 @@ func Run(runOpts RunOptions) {
 	if cfg.DebugListenAddress != "" {
 		go func() {
 			mux := http.DefaultServeMux
-			if err == nil {
+			if debugWriter != nil {
 				if err = debugWriter.RegisterHandler("/debug/dump", mux); err != nil {
 					logger.Error("unable to register debug writer endpoint", zap.Error(err))
 				}
-			} else {
-				logger.Error("unable to create debug writer", zap.Error(err))
 			}
 
 			if err := http.ListenAndServe(cfg.DebugListenAddress, mux); err != nil {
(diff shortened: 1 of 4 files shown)


obitech:
cc @richardartoul @robskillington

src/dbnode/server/server.go
```diff
@@ -564,9 +565,27 @@ func Run(runOpts RunOptions) {
 
 	if cfg.DebugListenAddress != "" {
 		go func() {
-			if err := http.ListenAndServe(cfg.DebugListenAddress, nil); err != nil {
+			mux := http.DefaultServeMux
+			cpuProfileDuration := 1 * time.Second
```
richardartoul: I'd probably do `5s`. Making it configurable would be nice if you have time, but not a huge deal since they can always do it manually if they want. Generally we make stuff like this a `const` at the top of the file though

src/dbnode/server/server.go
```diff
@@ -564,9 +565,27 @@ func Run(runOpts RunOptions) {
...
+			debugWriter, err := xdebug.NewZipWriterWithDefaultSources(
```
richardartoul: Can you actually initialize this somewhere outside of this Goroutine? Fairly early on in `server.go` would be ideal. The reason I ask that is we're going to want to register new sources soon (one to dump the YAML file, another to dump the namespace config) so we'll want it available more broadly than just in this async routine

obitech: Makes sense. I moved it further up, right after initialisation of `iopts`

src/dbnode/server/server.go
obitech: I set it to `5s` and as a `const`. Would like to keep this PR small, if possible, and make it configurable in a different PR.

src/dbnode/server/server.go
```diff
@@ -291,6 +293,11 @@ func Run(runOpts RunOptions) {
 
 	opentracing.SetGlobalTracer(tracer)
 
+	debugWriter, err := xdebug.NewZipWriterWithDefaultSources(
+		cpuProfileDuration,
+		iopts,
+	)
+
```
richardartoul: A little confused why you dont check the error here, but then you check it later below? Seems like you should just check it here (and fatal out or whatever we do for errors like this in `server.go` and then not check it in the goroutine below

obitech: Yeah that was dumb, forgot to pull up the error handling. Should be fixed now

src/dbnode/server/server.go
```diff
@@ -564,9 +574,20 @@ func Run(runOpts RunOptions) {
 
 	if cfg.DebugListenAddress != "" {
 		go func() {
-			if err := http.ListenAndServe(cfg.DebugListenAddress, nil); err != nil {
+			mux := http.DefaultServeMux
+			if debugWriter != nil {
+				if err = debugWriter.RegisterHandler("/debug/dump", mux); err != nil {
```
richardartoul: you can still use `:=` (to avoid overwriting the global) since its block scoped

obitech: Fixed

PR #1814, opened 2019-07-16
Move CacheShardIndices from namespace.Bootstrap() to shard.Bootstrap()

richardartoul:
The existing implementation would allow shards to serve reads before CacheShardIndices had completed for a given shard which could cause read latency to spike very high immediately following a node restart and result in Goroutine accumulation until all the seekers had been opened and cached.

The new implementation prevents reads from being served until all seekers for a given shard have been opened.


src/dbnode/storage/shard.go
```diff
@@ -1878,6 +1878,29 @@ func (s *dbShard) Bootstrap(
...
+		} else {
+			s.logger.Debug("Caching shard indices",
+				zap.Uint32("shard", s.ID()))
+			if err := retriever.CacheShardIndices([]uint32{s.ID()}); err != nil {
+				multiErr = multiErr.Add(err)
+				s.logger.Error("Caching shard indices error",
+					zap.Uint32("shard", s.ID()),
+					zap.Error(err))
+			} else {
+				s.logger.Debug("Caching shard indices completed successfully",
```
robskillington: Super nit: Can we change all these log statements to start with lower capitalized letters (the whole effective go naming stuff)?

src/dbnode/storage/shard.go
```diff
@@ -1878,6 +1878,29 @@ func (s *dbShard) Bootstrap(
 		s.setFlushStateColdVersion(at, info.VolumeIndex)
 	}
 
+	// May be nil depending on the caching policy.
+	if retrieverMgr := s.opts.DatabaseBlockRetrieverManager(); retrieverMgr!= nil {
+		// Attempt to call CacheShardIndices now that all the shards are bootstrapped. Cannot call
```
robskillington: nit: Update from "now that all the shards are bootstrapped" to "now that this shard is finishing bootstrapping".

PR #1816, opened 2019-07-16
Move CacheShardIndices from namespace.Bootstrap() to shard.Bootstrap()

richardartoul:
The existing implementation would allow shards to serve reads before CacheShardIndices had completed for a given shard which could cause read latency to spike very high immediately following a node restart and result in Goroutine accumulation until all the seekers had been opened and cached.

The new implementation prevents reads from being served until all seekers for a given shard have been opened.


src/dbnode/storage/shard.go
```diff
@@ -1878,6 +1878,29 @@ func (s *dbShard) Bootstrap(
...
+		// Now that this shard has finished bootstrapping attempt to cache all of its seekers. Cannot call
+		// this earlier as block lease verification will fail due to the shards not being bootstrapped
+		// (and as a result no leases can be verified since the flush state is not yet known).
+		retriever, err := retrieverMgr.Retriever(s.namespace)
+		if err != nil {
+			multiErr = multiErr.Add(err)
+		} else {
+			s.logger.Debug("caching shard indices",
+				zap.Uint32("shard", s.ID()))
+			if err := retriever.CacheShardIndices([]uint32{s.ID()}); err != nil {
```
robskillington: Heh, I just realized calling CacheShardIndices here will cause parallel caching due to shard.Bootstrap being called at `NumCPU/2` (which honestly is probably desired since it's not 100% disk IO bound and does do parsing, etc).

Will be interesting to see if this speeds up startup speed at all.

PR #1817, opened 2019-07-17
[dbnode] Add Load() API and refactor buffer.Bootstrap() logic to take warm vs. cold writes into account

richardartoul:
This is the first P.R to begin the process of making it possible load data dynamically into a shard/series/buffer such that the data will be merged with any data in memory or on disk and successfully persisted. The ability to do so is a key requirements of the background repairs feature which is being actively developed.

### DIFF
--- a/src/dbnode/storage/series/buffer.go
+++ b/src/dbnode/storage/series/buffer.go
@@ -135,9 +135,7 @@ type databaseBuffer interface {
 
 	Tick(versions map[xtime.UnixNano]BlockState, nsCtx namespace.Context) bufferTickResult
 
-	Bootstrap(bl block.DatabaseBlock, blockState BlockState)
-
-	Load(bl block.DatabaseBlock)
+	Load(bl block.DatabaseBlock, writeType WriteType)
 
 	Reset(id ident.ID, opts Options)
 }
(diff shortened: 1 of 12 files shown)


src/dbnode/storage/series/buffer.go
```diff
@@ -425,9 +427,33 @@ func (b *dbBuffer) Tick(blockStates map[xtime.UnixNano]BlockState, nsCtx namespa
...
-	buckets.bootstrap(bl)
+	// Only blocks that have never been flushed before are bootstrapped, so bootstrapped
+	// blocks should be treated the same as warm writes so that they get written to disk as
+	// part of the initial flush.
+	b.load(bl, WarmWrite)
+}
+
+func (b *dbBuffer) Load(bl block.DatabaseBlock) {
+	// Load may be called for any block start (including blocks that may or may not have
+	// been flushed yet.) As a result, they're treated as cold writes to ensure they get merged
```
justinjc: Nit: swap `.` and `)`

src/dbnode/storage/series/buffer.go
```diff
@@ -425,9 +427,33 @@ func (b *dbBuffer) Tick(blockStates map[xtime.UnixNano]BlockState, nsCtx namespa
...
-	blockStart := bl.StartTime()
-	buckets := b.bucketVersionsAtCreate(blockStart)
-	buckets.bootstrap(bl)
+	// Only blocks that have never been flushed before are bootstrapped, so bootstrapped
+	// blocks should be treated the same as warm writes so that they get written to disk as
+	// part of the initial flush.
+	b.load(bl, WarmWrite)
+}
+
+func (b *dbBuffer) Load(bl block.DatabaseBlock) {
```
justinjc: I feel like somewhere along this call stack should be a check for whether cold writes are enabled or not, otherwise these bootstrapped blocks would not get flushed.

src/dbnode/storage/series/series.go
```diff
@@ -428,7 +428,25 @@ func (s *dbSeries) Bootstrap(
 		return result, nil
 	}
 
-	// TODO(rartoul): Branch coverage here.
+	s.loadWithLock(bootstrappedBlocks, blockStates)
+	result.NumBlocksMovedToBuffer += int64(bootstrappedBlocks.Len())
+
+	s.bs = bootstrapped
```
justinjc: Probably not needed since this gets set as part of the `defer` above.

src/dbnode/storage/series/series.go
```diff
@@ -428,7 +428,25 @@ func (s *dbSeries) Bootstrap(
...
+	blockStates map[xtime.UnixNano]BlockState,
+) {
+	s.Lock()
+	s.loadWithLock(bootstrappedBlocks, blockStates)
+	s.Unlock()
+}
+
+func (s *dbSeries) loadWithLock(
+	bootstrappedBlocks block.DatabaseSeriesBlocks,
+	blockStates map[xtime.UnixNano]BlockState) {
```
justinjc: Nit: `) {` on new line

PR #1823, opened 2019-07-19
Arnikola/subfunctions

arnikola:

Fixes #

```release-note

```

```documentation-note

```


src/query/block/reconsolidated.go
```diff
@@ -0,0 +1,198 @@
...
+	next := it.it.Next()
+	it.ts = it.ts.Add(it.it.Meta().Bounds.StepSize)
+	return next
+}
+
+func (it *ucReconsolidatedStepIter) Current() UnconsolidatedStep {
+	var (
+		c      = it.it.Current()
+		values = c.Values()
+		dpList = make([]ts.Datapoints, 0, len(values))
```
robskillington: Guess there's no way to reuse these?

src/query/block/reconsolidated.go
```diff
@@ -0,0 +1,198 @@
...
+	next := it.it.Next()
+	it.idx = it.idx + 1
+	return next
+}
+
+func (it *ucReconsolidatedSeriesIter) Current() UnconsolidatedSeries {
+	var (
+		c      = it.it.Current()
+		values = c.values
+		dpList = make([]ts.Datapoints, 0, len(values))
```
robskillington: Similarly here, no way to reuse these by any chance?

src/query/block/reconsolidated.go
```diff
@@ -0,0 +1,198 @@
...
+	return b.block.StepIter()
+}
+
+// SeriesIter returns a SeriesIterator
+func (b *reconsolidatedBlock) SeriesIter() (SeriesIter, error) {
+	return b.block.SeriesIter()
+}
+
+// Unconsolidated returns the unconsolidated version for the block
+func (b *reconsolidatedBlock) Unconsolidated() (UnconsolidatedBlock, error) {
```
robskillington: Hm, wonder if we should come up with some terms for this.  `reconsolidatedBlock` and `ucReconsolidatedBlock` definitely took me a minute or so to totally grok the relationship, etc just based on name.  I get it know, but yeah, hehe.

src/query/models/types.go
```diff
@@ -99,6 +100,14 @@ type TagOptions interface {
 	IDSchemeType() IDSchemeType
 }
 
+type FetchOverrideOptions struct {
+	ShouldOverride bool
```
robskillington: Hehe, I like the avoiding the heap allocation by using a bool instead of making it a pointer type (for optionality).

robskillington: I really wish Go just had bloody optionals (i.e. unwrapping) baked in to the language in a zero alloc way.

arnikola:
Deprecated in favour of https://github.com/m3db/m3/pull/2143 (which itself is quite old and possibly deprecated)

Issue #1820, opened 2019-07-18
Dynamic Cluster Client Configuration


nerd0:
## Issue 

The [m3 cluster client](https://github.com/m3db/m3/blob/master/src/cluster/client/) interface which is only implemented for [ETCD](https://github.com/m3db/m3/blob/master/src/cluster/client/etcd/client.go) uses static configurations to define the endpoint list for an ETCD cluster. ETCD will use this list to [internally load balance this list of endpoints](https://github.com/etcd-io/etcd/blob/master/docs/client-architecture.rst#clientv3-grpc10-balancer-overview). Combine that feature with the [AutoSyncInterval](https://github.com/etcd-io/etcd/blob/master/clientv3/config.go#L30-L32) configuration, an ETCD client is able to maintain an up to date list of ETCD cluster members to internally balance against. However if the ETCD cluster members completely change, the static configuration can drift from the active list. Perhaps dynamic configuration generation can address this short coming, however it would be convenient if the endpoint list could provide in a more dynamic fashion during runtime of the process. 

## Potential Solutions

1. Implement a standard option and interface pattern which allows for a cluster client channel to be used startup of services such as the [dbnode](https://github.com/m3db/m3/blob/d75fc9d75348df76bc6f4707beaff48d507eed53/src/dbnode/server/server.go#L115-L116) currently allows. 

      Pros: 
      - Allows users to dynamically update the ETCD client during startup of the service process. 

      Cons:
      - Does not reduce the amount of other `main.go` initialising of various objects required to startup of a service process. 

2.  Refactor `main.go` files for each services (aggregator, dbnode, etc.) to [only perform configuration file parsing](https://github.com/m3db/m3/compare/pschooss/create-init-within-main#diff-6d668dbaa020817c90b77c2ad733b2f8R60) afterwards call a `Setup()` or `Init()` function which has an input parameter of a parsed configuration data structure. 

       Pros:
       - Allows users to dynamically update the configuration for services during startup enabling the process to ensure configuration values are up to date. 
       -  Simple refactor and readability of code does not decrease by much. 

       Cons:
       - Would require users to maintain a separate `main.go` to dynamically update the configuration data structure. 
       - Modifying the configuration data structure directly could prove difficult to debug since the configuration values are not deterministic as the configuration inputs can change in an unexpected ways. 
       - Unconventional coding style for the Go language. 

3.  Redesign the ETCD client implementation of the cluster client interface. 

      Pros:
[...]

PR #1824, opened 2019-07-23
[WIP] - Allow dynamic configuration injection

nerd0:

Fixes # [1820](https://github.com/m3db/m3/issues/1820)

None

```release-note
Componentize main files to allow dynamic configuration injection
```

None

```documentation-note

```


robskillington:
Hey @nerd0 so for the DB node all you need to do is call `server.Run(server.RunOptions{Config: yourCfg})` for dynamic configuration.

Can we simply make the aggregator follow the same model (use a `server` package and allow passing the config by a struct in the options) instead of adding another layer of indirection to both of these (otherwise there would be two levels of indirection in the dbnode main/server).

nerd0:
> Very keen to see #1820 go ahead.
> 
> Quick one for the DB node all you need to do is call `server.Run(server.RunOptions{Config: yourCfg})` for dynamic configuration.
> 
> Can we simply make the aggregator follow the same model (use a `server` package and allow passing the config by a struct in the options) instead of adding another layer of indirection to both of these (otherwise there would be two levels of indirection in the dbnode main/server).

Fair points, the idea with this approach was to ensure that the main file logic doesn't drift too far from the "vanilla" invocation of these services. The aggregator and dbnode main files differ quite a bit thus any main files maintained outside of the m3 repository will require a lot of duplicated code while also ensuring that any changes are propagated.

robskillington:
Nice, looks good after the changes! (Also, just as a maybe "out of my jurisdiction" request and hence maybe not worth making, it would be really amazing if people did not use a different main file outside of the open source repo to run parts of their production stack. We did this for a significant amount of time and was super hard to finally migrate it back to fully open source and have the build be 1-1 binary of what is in open source, and hence make it possible to debug the same binary one person has as another)

nerd0:
> Nice, looks good after the changes! (Also, just as a maybe "out of my jurisdiction" request and hence maybe not worth making, it would be really amazing if people did not use a different main file outside of the open source repo to run parts of their production stack. We did this for a significant amount of time and was super hard to finally migrate it back to fully open source and have the build be 1-1 binary of what is in open source, and hence make it possible to debug the same binary one person has as another)

Agreed with your comment, however the `cluster` package is tightly coupled to the ETCD client. Perhaps the better solution for this is too rewrite the `cluster` package to either allow for dynamic configuration injection rather than just static configuration files. Another option would be to rewrite the `cluster` package to allow for different discovery backends to be used. 

This diff was pretty much abandon on the last commit due to assumptions about how `main.go` was structured for the aggregator and the refactoring of the integration tests package.

PR #1826, opened 2019-07-24
[DBNode] Make dbnodes tick all the time

richardartoul:
Currently, M3DB nodes perform the following in an infinite loop in the background:

```
1. Tick()
2. Flush()
3. Snapshot()
4. ColdFlush()
```

The `Tick()` process is responsible for merging out-of-order encoders. In the situation that (for whatever reason) the `Flush()`/`Snapshot()`/`ColdFlush()` processes take a long time to complete, the ticking process will be blocked preventing out-of-order encoders from being merged and potentially leading to excessive memory usage (since compression is much worse).

The ticking process was designed to be run concurrently with all of the other processes, so this P.R moves the `Tick()` call to its own loop so that a new `Tick()` begins as soon as the previous one completes, preventing long filesystem operations from blocking it.

This P.R also removes as much stale logic as possible that was dependent on the `tickStartTime` since a lot of this logic is no longer required thanks to recent improvements to the `buffer`.


src/dbnode/integration/disk_snapshot_test.go
```diff
@@ -172,6 +172,9 @@ func TestDiskSnapshotSimple(t *testing.T) {
 		log.Info("waiting for old snapshot files to be deleted")
 		for _, shard := range shardSet.All() {
 			waitUntil(func() bool {
+				// Increase the time each check to ensure that the filesystem processes are able to progress (some
+				// of them throttle themselves based on time elapsed since the previous time.)
```
robskillington: Would love to have the time to investigate how difficult implementing direct IO would be.  I think it would definitely outweigh further investment in our throttling (which causes all sorts of problems, like this too).

src/dbnode/storage/mediator.go
```diff
@@ -135,39 +140,22 @@ func (m *mediator) EnableFileOps() {
...
-
-	// NB(r): Cleanup and/or flush if required to cleanup files and/or
-	// flush blocks to disk. Note this has to run after the tick as
-	// blocks may only have just become available during a tick beginning
-	// from the tick begin marker.
-	m.databaseFileSystemManager.Run(tickStart, dbBootstrapStateAtTickStart, syncRun, forceType)
+//
+// See comment over mediatorTimeBarrier for more details on how this is implemented.
+func (m *mediator) RunFilesystemProcesses(forceType forceType, startTime time.Time) error {
+	m.databaseFileSystemManager.Run(startTime, syncRun, forceType)
```
robskillington: Just wondering, does this method make sense to have a single line method wrapper?

src/dbnode/storage/mediator.go
```diff
@@ -192,21 +180,61 @@ func (m *mediator) Close() error {
...
+							"mediator time attempted to move backwards in time",
+							zap.Time("prevTime", mediatorTime), zap.Time("newTime", mediatorTime))
+					})
+
+					// Ignore new time so we can continue looping and hopefully get one thats actually in the future
+					// next time since moving backwards in time could cause undefined behavior.
+					continue
+				} else {
+					mediatorTime = m.nowFn()
+					m.mediatorTimeBarrier.release(mediatorTime)
```
robskillington: Should the release not happen after the Tick?  Otherwise the time doesn't represent "this is a start time from which the end of a tick has been completed" as it used to (and was entrusted to).

src/dbnode/integration/disk_snapshot_test.go
richardartoul: Agree'd direct I/O would be cool if we had time for it.

I'm not really adding any additional throttling in this P.R though

src/dbnode/storage/mediator.go
richardartoul: Yeah I actually had it the way you said before and then I removed it because we no longer rely on that guarantee in our code anywhere. The reason I introduced that restriction in the first place was because of all the weird interactions we had between the series buffer/blocks and blocks getting rotated out of the buffers and into blocks and then being flushed from there. Now that we've refactored all of the buffer logic to make those decisions based on actual state instead of just making assumptions based on time we don't need a full tick to complete before we can perform a flush (if you look you'll also see I removed all the nonsense about the shard bootstrap state before tick start as well since that is also not needed).

The only guarantee I think we need to maintain now is that a concurrent tick and filesystem processes never run with inconsistent views of the current time which is the guarantee I've made with this barrier.

src/dbnode/integration/commitlog_bootstrap_helpers.go
```diff
@@ -155,14 +155,8 @@ func writeCommitLogDataBase(
 		shardSet     = s.shardSet
 	)
 
-	// Write out commit log data
+	// Write out commit log data.
 	for currTs, blk := range data {
-		if specifiedTS != nil {
```
robskillington: Can you remove `specifiedTS` from the args to this method if it's not used/does anything any longer?

richardartoul: I don't think I meant to delete this. Gonna put it back

src/dbnode/storage/mediator.go
```diff
@@ -226,3 +246,88 @@ func (m *mediator) reportLoop() {
...
+func (b *mediatorTimeBarrier) hasWaiter() bool {
+	b.Lock()
+	isWaiting := b.isWaiting
+	b.Unlock()
+	return isWaiting
+}
+
+func newMediatorTimeBarrier() mediatorTimeBarrier {
+	return mediatorTimeBarrier{
+		doneCh: make(chan time.Time, 0),
```
robskillington: I kind of wonder if the `isWaiting` thing is really necessary.

Couldn't we just have two channels, the current `doneCh` with size zero (maybe renamed `waitCh`, and an exclusive `exclusiveTokenCh` with a pre-seeded token  and then the release code is basically `releaseIfWaiting()` (this will reduce the if/else in the actual loop):
```go
func (b *mediatorTimeBarrier) releaseIfWaiting(t time.Time) {
  select {
  case b.waitCh <- t:
  default:
    // not waiting
  }
}

func (b *mediatorTimeBarrier) exclusiveWait() (time.Time, error) {
  var token struct{}
  select {
  case token = <- b.exclusiveTokenCh:
  default: 
    return time.Time{}, errMediatorTimeBarrierAlreadyWaiting
  }
  
  t := <- b.waitCh
  
  b.exclusiveCh <- token
  return t, nil
}
```

It reduces the other loop to just:
```go
			newMediatorTime := m.nowFn()
			if newMediatorTime.Before(mediatorTime) {
				// .. error
				continue
			}

			mediatorTime = newMediatorTime

			m.mediatorTimeBarrier.releaseIfWaiting()

			if err := m.Tick(force, mediatorTime); err != nil {
				log.Error("error within tick", zap.Error(err))
			}
```

richardartoul: refactored the whole thing, let me know what you think

PR #1831, opened 2019-07-24
[DBnode] Add Load() API to shard

richardartoul:
Add an API for loading data into the shard at any time (not just during bootstrap) to support background repairs and add additional safe guards around using the shard flushState data before it has been bootstrapped.


src/dbnode/storage/shard.go
```diff
@@ -1787,14 +1804,84 @@ func (s *dbShard) Bootstrap(
...
+
+	s.Lock()
+	s.bootstrapState = Bootstrapped
+	s.Unlock()
+
+	return multiErr.FinalError()
+}
+
+func (s *dbShard) Load(
+	series *result.Map,
```
robskillington: Hm, wouldn't it make sense to do this in chunks instead of all at once (so we can move to streaming bootstrap?)

src/dbnode/storage/series/series.go
```diff
@@ -204,24 +205,29 @@ func (s *dbSeries) updateBlocksWithLock(
...
-				// read from disk (not retrieved), and the WiredList will manage those that were
-				// retrieved from disk.
-				shouldUnwire = !currBlock.WasRetrievedFromDisk()
-			default:
-				s.opts.InstrumentOptions().Logger().Fatal(
-					"unhandled cache policy in series tick", zap.Any("policy", cachePolicy))
+		blockStatesSnapshot, bootstrapped := blockStates.Snapshot()
+		// Only use block state snapshot information to make eviction decisions if the block state
+		// has been properly bootstrapped already.
+		if bootstrapped {
```
robskillington: Wondering why adding this check now is important, it should've always been an issue no?

src/dbnode/storage/types.go
```diff
@@ -466,6 +466,10 @@ type databaseShard interface {
 		bootstrappedSeries *result.Map,
 	) error
 
+	// Load does the same thing as Bootstrap, except it can be called more than once
+	// and after a shard is bootstrapped already.
+	Load(series *result.Map) error
```
robskillington: Does it not make sense perhaps to just allow bootstrap to be called multiple times and that bootstrapping a bootstrapped shard is ok?

src/dbnode/storage/series/types.go
```diff
@@ -173,7 +173,36 @@ type QueryableBlockRetriever interface {
...
+) ShardBlockStateSnapshot {
+	return ShardBlockStateSnapshot{
+		bootstrapped: bootstrapped,
+		snapshot:     snapshot,
+	}
+}
+
+// Snapshot returns a BootstrappedBlockStateSnapshot and a boolean indicating whether the
+// snapshot is bootstrapped or not.
+func (s *ShardBlockStateSnapshot) Snapshot() (BootstrappedBlockStateSnapshot, bool) {
```
robskillington: nit: Does this need to be pointer type?

robskillington:
@richardartoul the checks should only be necessary if we start ticking before a bootstrap I assume? (which we should be doing TBH, but that seems disconnected from this PR and change?)

src/dbnode/storage/shard.go
richardartoul: Yeah I may change the interface here slightly once I start hooking it up to the repair API, but I think this is fine for now. The map can be reset / re-used.

src/dbnode/storage/series/series.go
richardartoul: it basically "happened" to work before because an empty blockStateSnapshot will have `WarmRetrievable` `false` and `ColdVersion` `0` so you would never evict anything. I noticed this while I was working on the P.R and thought it was really dangerous since basically you come in and think you're programming against data that has been properly bootstrapped but actually in some cases you're (temporarily) running against a zero value so I wanted to make it more explicit so it doesn't get messed up in the future

src/dbnode/storage/series/types.go
richardartoul: I don't think it actually matters (it doesn't cause a heap allocation, it just allows you to reference the value on the stack instead of copying it) but I can change it since it doesn't require mutation

src/dbnode/storage/types.go
richardartoul: Its more complicated than that because if a shard is marked `bootstrapping` it can't serve reads and there may be some other implications I hadn't considered also. I think this strikes the best balance between sharing as much code as possible while not interfering with existing logic/workloads. I also think allowing `Bootstrap` to be called more than once is confusing as long as there is a difference between a bootstrapped and unbootstrapped shard

richardartoul:
@robskillington We already tick during bootstrap so the guards are actually really important. Right now everything happens to work because the logic happens to be correct for the zero values of the state, but I think enforcing a distinction between bootstrapped and unbootstrapped state is a good idea. Agree that it is a little weird for it to be in this P.R but I realized how easy it would have been for me to mess my P.R up because of the bootstrapping not being explicit so figured I'd add all the guard rails while I'm in here

src/dbnode/storage/shard.go
```diff
@@ -1787,14 +1804,84 @@ func (s *dbShard) Bootstrap(
...
+	_, err := s.loadSeries(series, false)
+	return err
+}
+
+func (s *dbShard) loadSeries(
+	series *result.Map,
+	bootstrap bool,
+) (dbShardBootstrapResult, error) {
+	if series == nil {
+		return dbShardBootstrapResult{}, nil
```
robskillington: Hm I'd be in preference of making this an error case, you ok with that?

src/dbnode/storage/shard.go
```diff
@@ -1823,54 +1910,38 @@ func (s *dbShard) Bootstrap(
...
-		if err != nil {
-			multiErr = multiErr.Add(err)
+		if bootstrap {
+			bsResult, err := entry.Series.Bootstrap(dbBlocks.Blocks, blockStatesSnapshot)
+			if err != nil {
+				multiErr = multiErr.Add(err)
+			}
+			shardBootstrapResult.update(bsResult)
+		} else {
+			entry.Series.Load(dbBlocks.Blocks, blockStatesSnapshot)
```
robskillington: Instead of having two separate APIs at the `series` package, maybe just a single API `Load(...)` with `LoadOptions struct { Bootstrap bool; }` and a `return LoadResult{BootstrapResult: bsResult}` from the method?

src/dbnode/storage/shard.go
richardartoul: The reason it behaves this way is because its very common to call `series.Bootstrap(nil)` but I could move it so `Bootstrap(nil)` is fine and `Load(nil)` is an error

src/dbnode/storage/shard.go
richardartoul: I really don't like using booleans like that in public APis, much prefer to have two separate APIs and keep boolean switches like that in internal methods personally (I understand it looks like API bloat but having an optional bool is basically the same API bloat but just in a round about way)

src/dbnode/storage/series/types.go
robskillington: Suggestion: Can we rename this method something like `Value()` or `State()` or similar?  I assumed looking at call sites that this was acquiring a lock or snapshotting some state, rather than just returning the immutable field value.

src/dbnode/storage/shard.go
robskillington: Sure thing, can revisit later.

src/dbnode/storage/shard.go
robskillington: Could we do that?  Make bootstrap(nil) valid but Load(nil) not valid?

richardartoul: @robskillington yeah thats what I implemented

src/dbnode/storage/types.go
robskillington: Yeah, I rethought this, makes sense more to just make calling Load time and time again and then Bootstrap once.

PR #1833, opened 2019-07-26
[DBNode] - Add support for limiting the maximum number of outstanding write/read requests at the RPC level

richardartoul:
While we still need to address [this proposal](https://github.com/m3db/proposal/issues/13) to solve the backpressure problem in depth, this P.R improves M3DB's ability to apply backpressure by supporting the ability to (optionally) limit the maximum number of outstanding write and read requests. This allows us to apply back pressure even earlier because in situations where the nodes are really overwhelmed many goroutines will get stuck allocating way before the commit log backpressure ever kicks in.

This P.R also allows us to size the writeBatchReq pool exactly to fit our requirements in the case that a maximum number of outstanding write requests has been configured.

### DIFF
--- a/src/dbnode/network/server/tchannelthrift/node/service.go
+++ b/src/dbnode/network/server/tchannelthrift/node/service.go
@@ -51,11 +51,11 @@ import (
 	xtime "github.com/m3db/m3/src/x/time"
 
 	apachethrift "github.com/apache/thrift/lib/go/thrift"
+	"github.com/m3db/m3/src/dbnode/namespace"
 	opentracinglog "github.com/opentracing/opentracing-go/log"
 	"github.com/uber-go/tally"
 	"github.com/uber/tchannel-go/thrift"
 	"go.uber.org/zap"
-	"github.com/m3db/m3/src/dbnode/namespace"
 )
 
 var (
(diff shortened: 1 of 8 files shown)


src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -165,6 +171,52 @@ func (s *serviceState) Health() (*rpc.NodeHealthResult_, bool) {
 	return v, v != nil
 }
 
+func (s *serviceState) DBForWriteRPCWithLimit() (
+	db storage.Database, dbInitialized bool, rpcDoesNotExceedLimit bool) {
```
robskillington: Hey, we may want to move this into a struct return type.

i.e.
```
func (s *serviceState) DBForWriteRPCWithLimit() (storage.Database, WriteRPCWithLimitResult) {
  // ...
  return s.db, WriteRPCWithLimitResult{Initialized: ..., ExceedsLimit: ...}
}
```

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -165,6 +171,52 @@ func (s *serviceState) Health() (*rpc.NodeHealthResult_, bool) {
...
+	s.Lock()
+	defer s.Unlock()
```
robskillington: Just FYI, we really should use an RLock and an atomic for the tracking of RPC exceeding limit, this an exclusive lock which long term might not be great for core write path.

richardartoul: @robskillington Yeah I agree an atomic would probably be better, but the exclusive lock is only ever taken if you explicitly opt into this feature so it will never impact users who don't choose to use it.

Also I just didn't want to spend the time futzing around with the atomics right now when its really not a bottleneck (even if the node is doing a million writes/s we're talking about 7-8k locks which is really not that much).

PR #1837, opened 2019-07-28
[dbnode] Safe handling of concurrent bloom filters

### LINKED ISSUE #1760: Unsafe handling of concurrent bloom filters
Currently the `SeekerManager` exposes concurrent bloom filters for given `shard/blockStart` combinations, however, the outstanding bloom filters are not tracked. This means that in theory its possible for a bloom filter to be given to a Retriever and then be closed later while some goroutine still has access to it. This would cause a panic when the goroutine tried to access the mmap'd bytes that had been unmapped.

In practice because of how the code is structured this is very unlikely to happen which is probably why we haven't seen it before.

Currently the code path is:

```
1. get bloomFilter from SeekerManager by calling `ConcurrentIDBloomFilter(shard, startTime)`
<-- possible for bloom filter to be closed between 1 and 2 -->
2. call bloomFilter.Test(id)
```

I think we could fix this issue by changing it to be:

```
1. seekerManager.Test(shard, startTime, id)
```

that way the seekerManger can run the bloom filter test (which should be pretty quick) while still holding an RLock() which would ensure the bloom filter wasn't closed.

obitech:

Fixes #1760 

Right now bloom filters that are returned via `seekerManager.ConcurrentIDBloomFilter` https://github.com/m3db/m3/blob/04d9c0654c80688e79ac7ffb2f15dd7ab3b4e8f8/src/dbnode/persist/fs/seek_manager.go#L212 are not tracked, which can lead to errors if one goroutine tries to access it (e.g. search for an ID) while another has closed the filter in the meantime. 

This PR introduces the `Test` method on `DataFileSetSeekerManager` https://github.com/m3db/m3/blob/04d9c0654c80688e79ac7ffb2f15dd7ab3b4e8f8/src/dbnode/persist/fs/types.go#L168 which tests for an ID while the lock on a bloom filter is still being held, to avoid the above described scenario. `seekerManager.Test` is now being used in `blockRetriever.Stream` https://github.com/m3db/m3/blob/04d9c0654c80688e79ac7ffb2f15dd7ab3b4e8f8/src/dbnode/persist/fs/retriever.go#L386 instead of "manually" testing the ID existence on `ConcurrentIDBloomFilter`. Subsequently, `ConcurrentIDBloomFilter` has been removed from the `DataFileSetSeekerManager` API.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -57,10 +57,11 @@ Once you have identified a change you want to make, and gathered consensus by ta
 (0) If you have updated an interface that has been mocked, you need to update the generated `gomock `files.
 
 ```shell
-# Generate mocks for all top level packages, or...
-mach mock-gen
+# Generate mocks for all top level packages
+make mock-gen
 
-# ...replace xyz with the package you want to generate files for, e.g. dbnode
+# If you just want to generate it for a single package,
+# replace xyz with the package you want to generate files for, e.g. dbnode
 make mock-gen-xyz
 ```
 
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -54,10 +54,13 @@ M3 has an extensive, and ever increasing, set of tests to ensure we are able to 
 
 Once you have identified a change you want to make, and gathered consensus by talking to some devs, go ahead and make a branch with the changes. To test your changes:
 
-(0) If necessary, generate GoMock files
+(0) If you have updated an interface that has been mocked, you need to update the generated `gomock `files.
 
 ```shell
-# Replace xyz with the package you want to generate files for, e.g. dbnode
+# Generate mocks for all top level packages, or...
+mach mock-gen
+
+# ...replace xyz with the package you want to generate files for, e.g. dbnode
 make mock-gen-xyz
 ```
 
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -54,6 +54,13 @@ M3 has an extensive, and ever increasing, set of tests to ensure we are able to 
 
 Once you have identified a change you want to make, and gathered consensus by talking to some devs, go ahead and make a branch with the changes. To test your changes:
 
+(0) If necessary, generate GoMock files
+
+```shell
+# Replace xyz with the package you want to generate files for, e.g. dbnode
+make mock-gen-xyz
+```
+
 (1) Run unit tests locally
 ```
 go test ./... -v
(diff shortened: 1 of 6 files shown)


DEVELOPER.md
```diff
@@ -54,6 +54,13 @@ M3 has an extensive, and ever increasing, set of tests to ensure we are able to
 
 Once you have identified a change you want to make, and gathered consensus by talking to some devs, go ahead and make a branch with the changes. To test your changes:
 
+(0) If necessary, generate GoMock files
```
robskillington: Maybe directly below this heading: "If you have updated an interface that has been mocked, you need to update the generated gomock files." to make it more clear?

TY for adding this.

DEVELOPER.md
```diff
@@ -54,6 +54,13 @@ M3 has an extensive, and ever increasing, set of tests to ensure we are able to
...
+
+```shell
+# Replace xyz with the package you want to generate files for, e.g. dbnode
+make mock-gen-xyz
```
robskillington: Maybe also add that you can run for all top level packages by omitting the suffix?
```
# Generate mocks for all top level packages
make mock-gen
```

src/dbnode/persist/fs/types.go
```diff
@@ -244,9 +244,9 @@ type DataFileSetSeekerManager interface {
 	// Return returns an open seeker for a given shard, block start time, and volume.
 	Return(shard uint32, start time.Time, seeker ConcurrentDataFileSetSeeker) error
 
-	// ConcurrentIDBloomFilter returns a concurrent ID bloom filter for a given
-	// shard, block start time, and volume.
-	ConcurrentIDBloomFilter(shard uint32, start time.Time) (*ManagedConcurrentBloomFilter, error)
+	// Test checks if an ID exists in a concurrent ID bloom filter for a
+	// given shard, block, start time and volume.
+	Test(shard uint32, id ident.ID, start time.Time) (bool, error)
```
robskillington: super nit: Perhaps we make this `Test(id ident.ID, shard uint32, start time.Time) (bool, error)` since shard is derived from the ID (and hence less "important" in the list of arguments).

DEVELOPER.md
obitech: Fixed

DEVELOPER.md
obitech: Done

src/dbnode/persist/fs/types.go
obitech: Sure, I changed the signature

PR #1838, opened 2019-07-28
[query] Add a limit exceeded header if too many series are returned

arnikola:
Adds a limit exceeded header to requests that go over the m3db limit

```release-note
None
```

```documentation-note
None
```


src/query/api/v1/handler/graphite/find_test.go
```diff
@@ -241,4 +251,28 @@ func TestFind(t *testing.T) {
...
+	if hasHeader {
+		require.Equal(t, "true", header)
+	} else {
+		require.Equal(t, "", header)
+	}
+}
+
+var limitTests = []struct {
+	name               string
+	ex, ex2, hasHeader bool
```
robskillington: Super nit: Maybe make a struct so passing to `testFind(...)` method is easier?

i.e.

```go
type findTestParams struct {
  terminatedQueryExhaustive bool
  childQueryExhaustive bool
  expectHeader bool
}

var limitTests = []struct {
  name string
  params findTestParams
}{
  {"both incomplete", findTestParams{false, true, true}},
  // ...
}

func testFind(t *testing.T, p findTestParams) {
  // ...
}
```

And similarly for `setupStorage(...)` could we use a `type setupStorageOptions struct` so that as we add new fields we don't need to update each test calling into `setupStorage(t, opts)` individually by adding the new default params to the call to the method?

src/query/api/v1/handler/graphite/render.go
```diff
@@ -200,6 +209,10 @@ func (h *renderHandler) serveHTTP(
 		SortApplied: true,
 	}
 
+	if !exhaustive {
+		w.Header().Set(handler.LimitHeader, "true")
```
robskillington: nit: Should we make this a constant?  i.e. `w.Header().Set(handler.LimitHeader, handler.LimitHeaderSeriesLimitApplied)`?

In the future if we limit the results due to a datacenter/remote region being down, we can concat these together perhaps.

ie.
```go
w.Header().Set(handler.LimitHeader, handler.LimitHeaderSeriesLimitApplied)

// in handler pkg
const LimitHeaderSeriesLimitApplied = "max_fetch_series_limit_applied"
```

Then in future if we have another type:
```
w.Header().Set(handler.LimitHeader, handler.LimitHeaderValue(handler.LimitHeaderSeriesLimitApplied, ...))

// in handler pkg
func LimitHeaderValues(str ...string) string {
  return strings.Join(str..., ",")
}
```

src/query/api/v1/handler/prometheus/native/complete_tags.go
```diff
@@ -86,6 +86,12 @@ func (h *CompleteTagsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request)
...
 
-	// TODO: Support multiple result types
-	prometheus.RenderTagCompletionResultsJSON(w, result)
+	if !result.Exhaustive {
+		w.Header().Set(handler.LimitHeader, "true")
+	}
+
+	if err = prometheus.RenderTagCompletionResultsJSON(w, result); err != nil {
+		logger.Error("unable to render results", zap.Error(err))
+		xhttp.Error(w, err, http.StatusBadRequest)
```
robskillington: Hm this is probably a 500 yeah?  Also I don't think we can write to the HTTP response if we already a partial output? (the headers and some of the body might have already been sent).

Probably best we can do is log the error.

src/query/api/v1/handler/prometheus/native/read_common.go
```diff
@@ -49,7 +49,7 @@ func read(
 	w http.ResponseWriter,
 	params models.RequestParams,
 	instrumentOpts instrument.Options,
-) ([]*ts.Series, error) {
+) ([]*ts.Series, bool, error) {
```
robskillington: nit: Should we perhaps just make the first return value here be a struct?  Then if we need to add other metadata later it should be easier both on the caller and callee side.

i.e.
```go
type readResult struct {
  series []*ts.Series
  exhaustive bool
}

func read(...) (readResult, error) { ... }
```

src/query/api/v1/handler/prometheus/native/read_common.go
```diff
@@ -86,36 +86,47 @@ func read(
 		}
 	}()
 
-	firstElement := false
-	var numSteps, numSeries int
+	var (
+		numSteps   int
+		numSeries  int
+		exhaustive bool
```
robskillington: Hm, should this not start out as `exhaustive = true`?

src/query/api/v1/handler/prometheus/native/read_common.go
```diff
@@ -203,10 +219,10 @@ func insertSortedBlock(
 	blockList []blockWithMeta,
 	stepCount,
 	seriesCount int,
-) ([]blockWithMeta, error) {
+) ([]blockWithMeta, bool, error) {
```
robskillington: Similar nit here, maybe we create a result struct?
```go
type insertSortedBlockResult struct {
  blocks []blockWithMeta
  exhaustive bool
}
```

src/query/api/v1/handler/prometheus/remote/read.go
```diff
@@ -178,7 +181,7 @@ func (h *PromReadHandler) read(
 	r *prompb.ReadRequest,
 	timeout time.Duration,
 	fetchOpts *storage.FetchOptions,
-) ([]*prompb.QueryResult, error) {
+) ([]*prompb.QueryResult, bool, error) {
```
robskillington: Similar nit, maybe a:
 ```
type readHandlerResult struct { 
  results []*prompb.QueryResult
  exhaustive bool
}
```

File: src/query/api/v1/handler/prometheus/remote/read.go [hunk 8]
```diff
@@ -202,22 +207,31 @@ func (h *PromReadHandler) read(
...
-				multiErrLock.Lock()
+				mu.Lock()
 				multiErr = multiErr.Add(err)
-				multiErrLock.Unlock()
+				mu.Unlock()
 				return
 			}
 
+			mu.Lock()
+			if !exhaustiveSet {
```
robskillington: Do you need to have this second var `exhaustiveSet` if you just set `exhaustive = true` at the beginning then always do `exhaustive = exhaustive && result.Exhaustive`? 

It should always be false if any `result.Exhaustive` is actually false and will default to true if for some reason you have zero queries.

src/query/api/v1/handler/graphite/render.go
```diff
@@ -103,9 +103,11 @@ func (h *renderHandler) serveHTTP(
 	}
 
 	var (
-		results = make([]ts.SeriesList, len(p.Targets))
-		errorCh = make(chan error, 1)
-		mu      sync.Mutex
+		results       = make([]ts.SeriesList, len(p.Targets))
+		errorCh       = make(chan error, 1)
+		exhaustive    = true
+		exhaustiveSet bool
```
robskillington: Do you need to have this second var `exhaustiveSet` if you just set `exhaustive = true` at the beginning then always do `exhaustive = exhaustive && result.Exhaustive`? 

It should always be false if any `result.Exhaustive` is actually false and will default to true if for some reason you have zero queries.

src/query/api/v1/handler/types.go
```diff
@@ -29,4 +29,7 @@ const (
 
 	// RoutePrefixV1 is the v1 prefix for all coordinator routes
 	RoutePrefixV1 = "/api/v1"
+
+	// LimitHeader is the header added when series limits are exceeded.
+	LimitHeader = "Series-Limit-Exceeded"
```
robskillington: Could we make this a more generic header name, which the value of we could concat? (for instance to describe limit applied for max fetch series as well as region not being available).

Really want to describe is a partial response, and why (and if multi reasons, then each one) it was a partial response.

Something generic like "M3-Results-Limited" with a comma separated string value of reasons seems like a reasonable start.  I proposed embedded JSON to get fancy, but that's probably too far.  If we ever need more information we can put it in a separate header.  Just a comma separated set of string enum like values probably works for now with a single header.

Here's the conversation I had:
> chengyu [1:56 PM]
> Is there a way to surface this information back? Our service wants to adjust our query pattern on excessive queries.
> Otherwise, we have no way to know if there is no timeseries or we are hitting some limit somewhere..
> 
> Rob Skillington [1:59 PM]
> yeah I’m talking to artem right now, we might just add a `M3-Results-Limited-QueryLimit: true` header
> along with a `M3-Results-Limited-Any: true`
> because if you do a fanout to other regions and those regions are down
> that’s another different type of limit
> 
> chengyu [2:01 PM]
> What about just “M3-Results-Limited” and the value be the actual reason?
> Instead of having multiple different type of headers that we probably need to adapt later (edited)

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -104,28 +106,29 @@ func (r *multiResult) FinalResultWithAttrs() (
 		}
 	}
 
-	return iters, attrs, nil
+	return iters, attrs, exhaustive, nil
 }
 
-func (r *multiResult) FinalResult() (encoding.SeriesIterators, error) {
+func (r *multiResult) FinalResult() (encoding.SeriesIterators, bool, error) {
```
robskillington: nit: Should we combine this into a single result too?
```go
type multiFinalResult struct {
 Iters encoding.SeriesIterators
 Exhaustive bool
}
```

src/query/storage/m3/multi_fetch_tags_result.go
```diff
@@ -58,25 +59,26 @@ func (r *multiSearchResult) Close() error {
 	return nil
 }
 
-func (r *multiSearchResult) FinalResult() ([]MultiTagResult, error) {
+func (r *multiSearchResult) FinalResult() ([]MultiTagResult, bool, error) {
```
robskillington: Nit: Combined result here too maybe?  

```go
type multiFinalSearchResult struct {
  Results []MultiTagResult
  Exhaustive bool
}
```

src/query/storage/m3/multi_fetch_tags_result_test.go
```diff
@@ -0,0 +1,53 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package m3
+
+import (
+	"testing"
+
+	"github.com/golang/mock/gomock"
```
robskillington: nit: External deps at bottom?

src/query/storage/m3/storage.go
```diff
@@ -488,17 +493,17 @@ func (s *m3storage) SearchCompressed(
 	ctx context.Context,
 	query *storage.FetchQuery,
 	options *storage.FetchOptions,
-) ([]MultiTagResult, Cleanup, error) {
+) ([]MultiTagResult, bool, Cleanup, error) {
```
robskillington: nit: Single result here if `[]MultiTagResult` was a `multiFinalSearchResult struct`

src/query/storage/mock/storage.go
```diff
@@ -146,7 +157,13 @@ func (s *mockStorage) Fetch(
 	s.Lock()
 	defer s.Unlock()
 	s.lastFetchOptions = opts
-	return s.fetchResult.result, s.fetchResult.err
+	idx := s.fetchResult.idx
+	if idx >= len(s.fetchResult.results) {
+		idx = 0
+	}
+
+	s.fetchResult.idx = s.fetchResult.idx + 1
```
robskillington: Hm, I don't quite get what's going on with the math on `idx` here?

src/query/tsdb/remote/client.go
```diff
@@ -136,15 +136,15 @@ func (c *grpcClient) fetchRaw(
 	ctx context.Context,
 	query *storage.FetchQuery,
 	options *storage.FetchOptions,
-) (encoding.SeriesIterators, error) {
+) (encoding.SeriesIterators, bool, error) {
```
robskillington: Maybe `fetchRawResult struct` here too?  Just to avoid all the extra return args and updates required if we introduce another "metadata" like field.

src/query/storage/mock/storage.go
arnikola: Oh, mostly just being lazy haha; if there's no more results sent through, it just returns the first one we sent in.

TBH should probably get rid of these handrolled mocks and refactor tests still using them to use regular generated mocks (would also mean we don't need to keep updating them)

src/query/api/v1/handler/prometheus/native/complete_tags.go
arnikola: Is there any way we can detect failed writes? We had an issue a while ago where we had a legitimate error in the render code, but it would always return empty results

src/query/api/v1/handler/prometheus/native/read_common.go
arnikola: Good call, makes sense for the empty result case

src/query/api/v1/handler/prometheus/remote/read.go
arnikola: Yeah good call, much nicer this way

src/query/api/v1/handler/graphite/render.go
arnikola: Sounds good

src/query/api/v1/handler/types.go
arnikola: Gonna tweak how we do this a bit, hopefully get some better results out

src/query/api/v1/handler/graphite/render.go
arnikola: The future is now!

src/query/api/v1/handler/graphite/find_test.go
arnikola: I don't mind as much for tests, seems like it just makes it slightly more complex to me, but changed this in prod code

src/query/api/v1/handler/prometheus/native/read_common.go
```diff
@@ -40,6 +40,11 @@ import (
 	opentracinglog "github.com/opentracing/opentracing-go/log"
 )
 
+type readResult struct {
```
robskillington: Nice, good pattern

src/query/block/lazy_test.go
```diff
@@ -1,5 +1,5 @@
 // Copyright (c) 2019 Uber Technologies, Inc.
-//
+//c
```
robskillington: nit: Remove this?

src/query/block/meta.go
```diff
@@ -43,3 +48,82 @@ func (m Metadata) Equals(other Metadata) bool {
...
+	// collection of retrieved data.
+	Exhaustive bool
+	// Warnings is a list of warnings that indicate potetitally partial or
+	// incomplete results.
+	Warnings []Warning
+}
+
+// NewResultMetadata creates a new result metadata.
+// NB: capacity for warning is initialized to 1, since it is unlikely that more
+// than one warning will exist.
```
robskillington: Maybe worth even just leaving it nil?  Only because I assume most queries won't cause a warning, and it will unfortunately almost definitely be a heap allocation (since it's a slice and pointer type).

src/query/block/meta.go
```diff
@@ -43,3 +48,82 @@ func (m Metadata) Equals(other Metadata) bool {
...
+	return ResultMetadata{
+		LocalOnly:  true,
+		Exhaustive: true,
+		Warnings:   make([]Warning, 0, 1),
+	}
+}
+
+// CombineMetadata combines two result metadatas.
+func (m ResultMetadata) CombineMetadata(other ResultMetadata) ResultMetadata {
+	combinedWarnings := make([]Warning, 0, len(m.Warnings)+len(other.Warnings))
```
robskillington: Maybe only allocate if `len(m.Warnings)+len(other.Warnings)` is greater than zero?

src/query/storage/fanout/storage.go
```diff
@@ -182,19 +183,32 @@ func (s *fanoutStorage) FetchBlocks(
...
+		meta.ResultMetadata = meta.ResultMetadata.CombineMetadata(resultMeta)
+		return meta
+	}
+
+	lazyOpts := block.NewLazyOptions().SetMetaTransform(updateResultMeta)
 	for _, bl := range blockResult {
+		// Update constituent blocks with combined resultMetadata if it has been
+		// changed.
+		if !resultMeta.IsDefault() {
+			bl = block.NewLazyBlock(bl, lazyOpts)
```
robskillington: Hah, this is kind of cool. Nice.

src/query/block/lazy_test.go
arnikola: Oops

src/query/block/meta.go
arnikola: Good call

src/query/api/v1/handler/graphite/render.go
```diff
@@ -164,8 +173,9 @@ func (h *renderHandler) serveHTTP(
 			}
 
 			mu.Lock()
-			results[i] = targetSeries
+			meta = meta.CombineMetadata(targetSeries.Metadata)
 			mu.Unlock()
+			results[i] = targetSeries
```
robskillington: Should the `results[i] = targetSeries` remain inside the mu.Lock() and mu.Unlock()?  I guess it's inserting into unique index so doesn't necessarily need to be...

src/query/graphite/common/basic_functions.go
```diff
@@ -107,14 +108,17 @@ func Identity(ctx *Context, name string) (ts.SeriesList, error) {
 	}
 	newSeries := ts.NewSeries(ctx, name, ctx.StartTime, vals)
 	newSeries.Specification = fmt.Sprintf("identity(%q)", name)
-	return ts.SeriesList{Values: []*ts.Series{newSeries}}, nil
+	return ts.SeriesList{
+		Values:   []*ts.Series{newSeries},
+		Metadata: block.NewResultMetadata(),
+	}, nil
```
robskillington: nit: For this does it make sense to just create a special func like `ts.NewSeriesListWithSeries(...)` which auto creates the NewResultMetadata?

src/query/graphite/common/basic_functions.go
```diff
@@ -164,19 +168,21 @@ func Count(ctx *Context, seriesList ts.SeriesList, renamer SeriesListRenamer) (t
 		numSteps := ctx.EndTime.Sub(ctx.StartTime).Minutes()
 		vals := ts.NewZeroValues(ctx, MillisPerMinute, int(numSteps))
 		r := ts.SeriesList{
-			Values: []*ts.Series{ts.NewSeries(ctx, renamer(seriesList), ctx.StartTime, vals)},
+			Values:   []*ts.Series{ts.NewSeries(ctx, renamer(seriesList), ctx.StartTime, vals)},
+			Metadata: block.NewResultMetadata(),
```
robskillington: Here too, seems better to just always force ts.SeriesList to be constructed with some type of constructor.

src/query/storage/m3/multi_fetch_result.go
```diff
@@ -104,47 +107,52 @@ func (r *multiResult) FinalResultWithAttrs() (
...
-		return r.finalResult, nil
+
+	if r.mergedIterators != nil {
+		result.SeriesIterators = r.mergedIterators
+		return result, nil
 	}
 
 	if len(r.seenIters) == 0 {
-		return encoding.EmptySeriesIterators, nil
+		return result, nil
```
robskillington: Does `result.SeriesIterators = encoding.EmptySeriesIterators` need to be set here before returning?

src/query/storage/m3/multi_fetch_tags_result_test.go
```diff
@@ -0,0 +1,56 @@
...
```
robskillington: Reminder to group gomock with the testify/assert.

src/query/storage/m3/storage.go
```diff
@@ -351,8 +381,8 @@ func (s *m3storage) SearchSeries(
 		return nil, err
 	}
 
-	metrics := make(models.Metrics, len(tagResult))
-	for i, result := range tagResult {
+	metrics := make(models.Metrics, len(tagResult.Tags))
```
robskillington: Make this into a length zero with append calls while here nmaybe?

src/query/tsdb/remote/codecs.go
```diff
@@ -190,23 +138,62 @@ func encodeTagMatchers(modelMatchers models.Matchers) (*rpc.TagMatchers, error)
...
+	case storage.FanoutDefault:
+		return rpc.FanoutOption_DEFAULT_OPTION, nil
+	case storage.FanoutForceDisable:
+		return rpc.FanoutOption_FORCE_DISABLED, nil
+	case storage.FanoutForceEnable:
+		return rpc.FanoutOption_FORCE_ENABLED, nil
+	}
+
+	return rpc.FanoutOption_FORCE_ENABLED,
+		fmt.Errorf("Unknown fanout option for proto encoding: %v\n", opt)
```
robskillington: Hm seems this one could be `return 0, fmt.Errorf("unknown ...")` as pinged about elsewhere

src/query/storage/m3/multi_fetch_result.go
arnikola: Oops, nice catch

src/query/api/v1/handler/graphite/render.go
arnikola: I think it's the latter, but moved it into the lock just to be safe; doubt there are any perf implications here

PR #1839, opened 2019-07-28
[dbnode] Use ref to segment data for index results instead of alloc each

robskillington:

This greatly reduces the memory allocation usage when querying the index as it ensures the lifetime of the segment survives the query and returns just references to the index data.

```release-note
NONE
```

```documentation-note
NONE
```


src/m3ninx/index/segment/fst/segment.go
```diff
@@ -158,11 +158,20 @@ func NewSegment(data SegmentData, opts Options) (Segment, error) {
 		numDocs:        metadata.NumDocs,
 		startInclusive: startInclusive,
 		endExclusive:   endExclusive,
-	}, nil
+	}
+
+	// NB(r): The segment uses the context finalization to finalize
```
richardartoul: This might be a good place to explain the general approach and how it interacts with outstanding queries

src/dbnode/server/server.go
```diff
@@ -727,12 +727,12 @@ func Run(runOpts RunOptions) {
 
 			bs.SetBootstrapperProvider(updated.BootstrapperProvider())
 
-			if err := bsGauge.UpdateStringList(bootstrappers); err != nil {
-				logger.Error("unable to update bootstrap gauge with new bootstrappers",
-					zap.Strings("bootstrappers", bootstrappers),
-					zap.Error(err),
-				)
-			}
+			// if err := bsGauge.UpdateStringList(bootstrappers); err != nil {
```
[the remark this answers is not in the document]
robskillington: Yeah this is meant to be re-enabled but it was crashing and I needed to test, I'll re-enable when it's fixed on master.

src/dbnode/storage/index/block.go
```diff
@@ -827,12 +827,18 @@ func (b *block) queryWithSpan(
...
 
+	// Register local components that need closing.
 	defer func() {
 		iterCloser.Close()
-		execCloser.Close()
 		docsPool.Put(batch)
 	}()
 
+	// Register the executor to close when context closes
+	// so can copy the results.
```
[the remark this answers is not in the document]
robskillington: Correct, I think I mean to say "so we can avoid copying the results".  Honestly I hadn't cleaned all this up yet.

src/dbnode/storage/index/block.go
```diff
@@ -865,10 +871,6 @@ func (b *block) queryWithSpan(
 		return false, err
 	}
 
-	if err := execCloser.Close(); err != nil {
```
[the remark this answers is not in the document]
robskillington: Oh nah, this got moved to the context closing it when it closes.  (SafeCloser wraps the exec close call and prevents it from being double closed, so the two calls existing wasn't an issue).

src/dbnode/storage/index/results.go
```diff
@@ -141,26 +143,18 @@ func (r *results) addDocumentWithLock(
...
 	// i.e. it doesn't exist in the map, so we create the tags wrapping
 	// fields prodided by the document.
-	tags := r.cloneTagsFromFields(d.Fields)
+	tags := convert.ToMetricTags(d, convert.Opts{NoClone: true})
 
-	// We use Set() instead of SetUnsafe to ensure we're taking a copy of
-	// the tsID's bytes.
-	r.resultsMap.Set(tsID, tags)
+	// It is assumed that the document is valid for the lifetime of the index
+	// results.
```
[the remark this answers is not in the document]
robskillington: Yeah, I'm trying to find the best place to document that.

src/m3ninx/index/segment/fst/segment.go
robskillington: Sure yeah.

mstump:
Is there a container with this patch available on a public repo? I want tot test it out.

robskillington:
@mstump about to merge and cut release in next few hours 👍

robskillington:
@mstump let me just build a container for this change.

robskillington:
@mstump here's a container for this branch: 
```
quay.io/m3db/m3dbnode:index-and-recent-data-zero-copy
```

Manifest:
https://quay.io/repository/m3db/m3dbnode/manifest/sha256:c974cab45e7da0e7547dc46e84e9988eb8c113e2129566bd129cc3eb6027ee70

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -276,8 +277,11 @@ func (enc *encoder) reset(start time.Time, bytes checked.Bytes) {
 }
 
 // Stream returns a copy of the underlying data stream.
-func (enc *encoder) Stream(opts encoding.StreamOptions) (xio.SegmentReader, bool) {
-	segment := enc.segment(byCopyResultType)
+func (enc *encoder) Stream(
+	ctx context.Context,
```
richardartoul: Take it or leave it, but it could be nice to make the `zeroCopy` behavior optional via the `encoding.StreamOptions` and put the context in their too so that we don't have to pollute this API (and create a context everytime we want to do this even when we dont need zero copy)

src/dbnode/persist/fs/merger.go
```diff
@@ -229,23 +230,24 @@ func (m *merger) Merge(
 		// Closing the context will finalize the data returned from
 		// mergeWith.Read(), but is safe because it has already been persisted
 		// to disk.
-		tmpCtx.BlockingClose()
+		// NB(r): Make sure to use BlockingCloseReset so can reuse the context.
```
richardartoul: isnt the reset handled by the reset right before the call to `mergeWith.Read()`?

robskillington:
@mstump I published a new container with the latest changes for testing:
```
quay.io/m3db/m3dbnode:index-and-recent-data-zero-copy-2
```

mstump:
It looks like the container only has the coordinator binary and doesn't have `m3dbnode`
```
/ # find / -name 'm3*'
/bin/m3coordinator
/etc/m3coordinator
/etc/m3coordinator/m3coordinator.yml
```

robskillington:
@mstump yeah that is weird, I noticed that.  I built the wrong binary on my laptop sigh.

mstump:
OK I got it deployed a day or two ago but didn't have time to test until just now. 

- It doesn't crash but the node will become unresponsive under the load. The node will begin to fail health checks and will be restarted by Kubernetes, taking down the entire cluster. Prometheus stat pulls also fail.
- CPU Doesn't max out, I'm not sure where the bottleneck is.
- Memory usage increase is much more gradual than before and didn't reach the 50GB limit.
- After applying the patch baseline CPU utilization increased from about 1.5 CPU to 2.0 CPUs

The workload is essentially CSV export for all metrics of a cluster at 1m granularity for a 60m time period. It's about 2-3k time series and the query looks like the following.

```
query_range?query=%7Bcluster_id%3D%22d752ae7a-6801-4a1b-9d00-7be6734ff409%22,+scope%3D%22CLUSTER%22%7D&step=60&end=1562814080&start=1562810480
```

#### Screen cap from load test
image

mstump:
I should also mention that it's only 10 concurrent queries from this job, there is an additional baseline load resulting in 10k tagged RPC writes per second, and 6 analytical queries per second (7d queries over a 1h aggregated namespace) which are fronted by a dedicated query tier.

mstump:
I repeated the test with a single thread of execution and it's still enough to cause nodes to go unresponsive.

mstump:
With the 1 concurrent query test I removed load as soon as the nodes started to go unresponsive. The only message of note has to do with etcd. I checked etcd memory and CPU utilization during that time-frame and it was nominal maxing out at 67MB of memory and .015 CPU.

```
{"level":"error","ts":1565373847.3873844,"msg":"received error on watch channel","error":"etcdserver: no leader"}
{"level":"info","ts":1565373852.6725218,"msg":"finished handling request","rqID":"2e3f3728-ad49-4043-8adf-ca08294a9f2e","time":1565373852.6725035,"response":5.296629915,"url":"/api/v1/placement"}
{"level":"warn","ts":1565373854.0264885,"msg":"etcd watch channel closed on key, recreating a watch channel","key":"_sd.placement/m3db/metrics-cluster/m3db"}
```

richardartoul:
@mstump cant comment on the the actual performance since this was Rob's work but the etcd stuff is usually a red herring. Usually it just means the node is under heavy load and as a result connections to etcd get broken and need to be re-established

notbdu:
Seeing ~43% reduction in RSS memory when running a side by side comparison of feature to release. 

Feature:
feature

Release:
release

src/dbnode/persist/fs/merger.go
robskillington: Yeah the main thing we avoid using `BlockingCloseReset()` is to avoid the context going back to the pool. (i.e. the Reset as the suffix means, after closing do not put back in the pool, we explicitly want to reuse this context).

And the main reason need to use a pooled context here is since the context needs to refer to the pooled finalizers linked list now, or else it will allocate a ton in between uses in this tight loop.

PR #1843, opened 2019-07-31
[dbnode] Fix panic when updating bootstrappers gauge on Promreporter with different label keys

obitech:

The bootstrap gauge introduced with #1756 can lead to a panic, when the amount of label keys changes in regards to what it had been initialised with. For example, if `dbnode` has been started with four bootstrappers configured but this amount is changed to three during runtime, the Prometheus Reporter for tally will panic.

In this PR the exposition of the bootstrap order has been changed from a single to multiple gauges in `StringListEmitter`, where each gauge has different value for the label key `type`:

```
bootstrappers0{type="filesystem"} 1
bootstrappers1{type="peers"} 1
bootstrappers2{type="commitlog"} 1
bootstrappers3{type="uninitialized_topology"} 1
```

Changing the order will set the old metric/value combination to 0 and update the respective gauge with a new value of 1: 

```shell
$ curl http://localhost:7201/api/v1/database/config/bootstrappers -X POST -d '{"values":["filesystem","commitlog","peers", "uninitialized_topology"]}'
{"values":["filesystem","commitlog","peers","uninitialized_topology"]}

$ curl -s localhost:9004/metrics | grep bootstrappers | grep -v "#"
bootstrappers0{type="filesystem"} 1
bootstrappers1{type="commitlog"} 1
bootstrappers1{type="peers"} 0
bootstrappers2{type="commitlog"} 0
bootstrappers2{type="peers"} 1
bootstrappers3{type="uninitialized_topology"} 1
```
---
The unit test `TestStringListEmitterPrometheusScope` checks if metrics can be registered 
with the Promreporter from `github.com/uber-go/tally/prometheus`. I have seen that this test
timeouts rather than just fails, not really sure why.

```release-note
NONE
```

```documentation-note
NONE
```

### DIFF
--- a/src/x/instrument/string_list_emitter.go
+++ b/src/x/instrument/string_list_emitter.go
@@ -47,7 +47,7 @@ type StringListEmitter struct {
 	running   bool
 	doneCh    chan bool
 	scope     tally.Scope
-	gauge     tally.Gauge
+	gauges    []tally.Gauge
 	name      string
 	tagPrefix string
 }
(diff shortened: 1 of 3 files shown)


src/x/instrument/string_list_emitter.go
```diff
@@ -40,101 +40,107 @@ var (
...
 }
 
-// NewStringListEmitter returns a StringListEmitter. The name and tagPrefix
-// arguments are used by the Start() and UpdateStringList() function to set
-// the name and tags on the gauge.
-func NewStringListEmitter(scope tally.Scope, name, tagPrefix string) *StringListEmitter {
-	gauge := tally.NoopScope.Gauge("blackhole")
+// NewStringListEmitter returns a StringListEmitter.
+func NewStringListEmitter(scope tally.Scope, name string) *StringListEmitter {
+	gauge := []tally.Gauge{tally.NoopScope.Gauge("blackhole")}
```
robskillington: nit: Since this is a slice now that you loop over and .Inc each one this could just be a zero length list of gauges I assume?

src/x/instrument/string_list_emitter_test.go
```diff
@@ -174,4 +226,58 @@ func TestBootstrapGaugeEmitter(t *testing.T) {
...
+
+// This test might timeout instead of just fail
+func TestStringListEmitterPrometheusScope(t *testing.T) {
+	r := promreporter.NewReporter(promreporter.Options{})
+	scope, closer := tally.NewRootScope(tally.ScopeOptions{
+		Prefix:         "",
+		Tags:           map[string]string{},
+		CachedReporter: r,
+	}, 0)
+	defer closer.Close()
```
robskillington: Nice work on this test.

PR #1845, opened 2019-07-31
Make namespace readers volume-aware

### LINKED ISSUE #1747: Make namespace_readers file-volume-aware

justinjc:

Fixes #1747

```release-note
NONE
```

```documentation-note
NONE
```


src/dbnode/generated/proto/pagetoken/pagetoken.proto
```diff
@@ -2,14 +2,15 @@ syntax = "proto3";
 package pagetoken;
 
 message PageToken {
-	message ActiveSeriesPhase {
-		int64 indexCursor = 1;
-	}
-	message FlushedSeriesPhase {
-		int64 currBlockStartUnixNanos = 1;
-		int64 currBlockEntryIdx = 2;
-	}
```
justinjc: Replacing tabs with spaces.

src/dbnode/storage/namespace_readers.go
```diff
@@ -238,9 +266,22 @@ func (m *namespaceReaderManager) get(
 	blockStart time.Time,
 	position readerPosition,
 ) (fs.DataFileSetReader, error) {
+	latestVolume, err := m.latestVolume(shard, blockStart)
+	if err != nil {
+		return nil, err
+	}
+
+	// If requesting an outdated volume, we need to start again reading from
```
richardartoul: "again reading" -> "reading again"

richardartoul: Can you add more to the comment basically just saying this is ok since its fine to return duplicate metadata?

richardartoul: also maybe just say that the cached reader that was for the previous volume will get cleaned up automatically by the ticking process so we dont need to worry about closing it here.

src/dbnode/storage/namespace_readers.go
```diff
@@ -380,3 +444,26 @@ func (m *namespaceReaderManager) tickWithThreshold(threshold int) {
...
+		return block.NoOpenLease, nil
+	}
+
+	m.Lock()
+	// Remove open readers with matching key but lower volume.
+	for readerKey := range m.openReaders {
+		if readerKey.shard == descriptor.Shard &&
+			readerKey.blockStart == xtime.ToUnixNano(descriptor.BlockStart) &&
+			readerKey.volume < state.Volume {
+			delete(m.openReaders, readerKey)
```
richardartoul: dont you need to actually close it?

justinjc: Hah, good point. I'll also put it back to the `closedReaders` which serves as the pool of readers here.

src/dbnode/storage/shard.go
```diff
@@ -1766,15 +1767,24 @@ func (s *dbShard) FetchBlocksMetadataV2(
...
 		// Return the reader to the cache
-		endPos := int64(reader.MetadataRead())
-		s.namespaceReaderMgr.put(reader)
+		err = s.namespaceReaderMgr.put(reader)
+		if err != nil {
+			return nil, nil, err
+		}
 
 		if numResults >= limit {
+			endPos := int64(reader.MetadataRead())
```
richardartoul: I think `endPos` and `volume` need to be saved before putting the reader back in the cache. Checking them after is undefined.

src/dbnode/storage/namespace_readers.go
```diff
@@ -296,22 +341,50 @@ func (m *namespaceReaderManager) get(
 	return reader, nil
 }
 
-func (m *namespaceReaderManager) put(reader fs.DataFileSetReader) {
+func (m *namespaceReaderManager) closeAndPushReader(reader fs.DataFileSetReader) error {
```
richardartoul: nit: can you call this `closeAndPushReaderWithLock`

src/dbnode/storage/namespace_readers.go
```diff
@@ -320,22 +393,26 @@ func (m *namespaceReaderManager) put(reader fs.DataFileSetReader) {
 	if _, ok := m.openReaders[key]; ok {
 		// Unlikely, however if so just close the reader we were trying to put
 		// and put into the closed readers
```
richardartoul: I think this sentence is slightly off

src/dbnode/storage/namespace_readers.go
```diff
@@ -380,3 +457,31 @@ func (m *namespaceReaderManager) tickWithThreshold(threshold int) {
...
+// UpdateOpenLease() implements block.Leaser.
+func (m *namespaceReaderManager) UpdateOpenLease(
+	descriptor block.LeaseDescriptor,
+	state block.LeaseState,
+) (block.UpdateOpenLeaseResult, error) {
+	if !m.namespace.ID().Equal(descriptor.Namespace) {
+		return block.NoOpenLease, nil
+	}
+
+	m.Lock()
```
richardartoul: Any reason to manually Lock and Unlock here instead of defer?

richardartoul: I guess it doesn't really matter much since there aren't any early returns

src/dbnode/storage/shard.go
```diff
@@ -1765,16 +1766,24 @@ func (s *dbShard) FetchBlocksMetadataV2(
 				blockResult))
 		}
 
-		// Return the reader to the cache
 		endPos := int64(reader.MetadataRead())
-		s.namespaceReaderMgr.put(reader)
+		// This volume may be different from the one initialliy requested,
```
richardartoul: initialliy -> initially

src/dbnode/storage/shard.go
```diff
@@ -1765,16 +1766,24 @@ func (s *dbShard) FetchBlocksMetadataV2(
...
+		// e.g. if there was a compaction between the last call and this
+		// one.
```
richardartoul: maybe add to the end of the comment: "so be sure to update the state of the pageToken otherwise the next request would have to start from the beginning since it would be requesting a stale volume and this could result in an infinite loop of requests that never complete." Or something like that.

Also a comment saying this must happen **before** the call to put would be nice just to make sure

src/dbnode/storage/namespace_readers.go
justinjc: Not really. I'll change it to a defer to avoid bugs in the future.

src/dbnode/storage/namespace_readers.go
```diff
@@ -238,6 +265,25 @@ func (m *namespaceReaderManager) get(
...
+	}
+
+	// If requesting an outdated volume, we need to start reading again from
+	// the beginning of the latest volume. The caller knows how to handle
+	// duplicate metadata, so doing this is okay.
+	//
+	// The previously cached reader for the outdated volume will eventually be
+	// cleaned up either during the ticking process or the next time
+	// UpdateOpenLease gets called, so we don't need to worry about closing it
+	// here.
```
robskillington: Hm interesting, not sure how I feel about this strategy long term.  This might be fine for now, but for any clusters that regularly get cold writes to most shards/block starts, it might make it impossible to ever finish a bootstrap as it just gets continually reset.

PR #1846, opened 2019-07-31
[query] Fix variadic functions

### LINKED ISSUE #1844: Panic when parsing args for PromQL query
Version v0.10.2

Panic logs
```
panic: runtime error: index out of range

goroutine 2035995 [running]:
github.com/m3db/m3/src/query/parser/promql.(*parseState).walk(0xc072abbac0, 0x19af8e0, 0xc071e31820, 0x19af8e0, 0xc071e31820)
    /go/src/github.com/m3db/m3/src/query/parser/promql/parse.go:201 +0x213e
github.com/m3db/m3/src/query/parser/promql.(*promParser).DAG(0xc071e31840, 0x19fbe28, 0x275fd30, 0x7, 0xc0755c5110, 0x1, 0x1, 0x19fbe00, 0x275fd30)
    /go/src/github.com/m3db/m3/src/query/parser/promql/parse.go:57 +0xa6
github.com/m3db/m3/src/query/executor.(*Request).compile(0xc072abbe10, 0x19dfa00, 0xc07228cc60, 0x19c4ec0, 0xc071e31840, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
    /go/src/github.com/m3db/m3/src/query/executor/request.go:79 +0x108
github.com/m3db/m3/src/query/executor.(*engine).ExecuteExpr(0xc0123d8e60, 0x19df9c0, 0xc05c5cac00, 0x19c4ec0, 0xc071e31840, 0xc071e2e300, 0x0, 0xed4d37aec, 0x273dec0, 0x0, ...)
    /go/src/github.com/m3db/m3/src/query/executor/engine.go:146 +0x1b9
created by github.com/m3db/m3/src/query/api/v1/handler/prometheus/native.read
    /go/src/github.com/m3db/m3/src/query/api/v1/handler/prometheus/native/read_common.go:72 +0x894
```

arnikola:
Fixes #1844

**Special notes for your reviewer**: bit of scope creep; this PR also fixes a couple of incorrect functions, and lazifies some others.


src/query/functions/linear/absent.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/m3db/m3/src/query/executor/transform"
 )
 
+// FIXME: This is incorrect functionality.
+//' This should be an aggregation function that works in a step-wise fashion.
```
robskillington: Can we create an issue for this?

src/query/functions/linear/datetime.go
```diff
@@ -25,27 +25,34 @@ import (
 	"math"
 	"time"
 
-	"github.com/m3db/m3/src/query/executor/transform"
+	"github.com/m3db/m3/src/query/block"
+	"github.com/m3db/m3/src/query/functions/lazy"
+
```
robskillington: nit: Join the imports?

src/query/functions/linear/datetime.go
```diff
@@ -71,37 +81,30 @@ var (
 	}
 )
 
-// NewDateOp creates a new date op based on the type
-func NewDateOp(optype string) (BaseOp, error) {
-	if _, ok := datetimeFuncs[optype]; !ok {
-		return emptyOp, fmt.Errorf("unknown date type: %s", optype)
+func buildTransform(args []interface{}, fn timeFn) block.ValueTransform {
```
robskillington: Hm is args used in this method at all?  Seems like only checked if length == 0?

src/query/functions/linear/round.go
```diff
@@ -24,67 +24,58 @@ import (
...
+		return 0, fmt.Errorf("unable to cast to to_nearest argument: %v",
+			args[0])
 	}
 
-	return BaseOp{
-		operatorType: RoundType,
-		processorFn:  makeRoundProcessor(spec),
-	}, nil
-
+	return 1, nil
```
robskillington: Interesting, why does args > 1 always return a 1 value?  Should this be error instead or no?

src/query/parser/promql/parse.go
```diff
@@ -228,11 +229,37 @@ func (p *parseState) walk(node pql.Node) error {
...
+			variadic    = n.Func.Variadic
+		)
+
+		if variadic == 0 {
+			if argCount != exprCount {
+				return fmt.Errorf("incorrect number of expressions(%d) for %q, "+
+					"received %d", exprCount, n.Func.Name, argCount)
+			}
+		} else {
+			if argCount-1 > exprCount {
```
robskillington: Would this slightly be easier to grok with `exprCount < argCount`?

src/query/parser/promql/parse.go
```diff
@@ -228,11 +229,37 @@ func (p *parseState) walk(node pql.Node) error {
...
+				return fmt.Errorf("incorrect number of expressions(%d) for variadic "+
+					"function %q, received %d", exprCount, n.Func.Name, argCount)
+			}
+
+			if argCount != exprCount {
+				numVals--
```
robskillington: Maybe we should call this `numExpVals` or something to reflect it's what we're expecting?

src/query/parser/promql/parse.go
```diff
@@ -228,11 +229,37 @@ func (p *parseState) walk(node pql.Node) error {
...
+				return fmt.Errorf("incorrect number of expressions(%d) for %q, "+
+					"received %d", exprCount, n.Func.Name, argCount)
+			}
+		} else {
+			if argCount-1 > exprCount {
+				return fmt.Errorf("incorrect number of expressions(%d) for variadic "+
+					"function %q, received %d", exprCount, n.Func.Name, argCount)
+			}
+
+			if argCount != exprCount {
```
robskillington: For readability, does it make sense to do something like `if argCount-1 == exprCount { ... }` do this logic?

src/query/parser/promql/parse.go
```diff
@@ -246,6 +273,10 @@ func (p *parseState) walk(node pql.Node) error {
 			} else {
 				if e, ok := expr.(*pql.MatrixSelector); ok {
 					argValues = append(argValues, e.Range)
+				} else if _, ok := expr.(*pql.VectorSelector); ok {
+					if variadic != 0 && n.Func.Name != linear.RoundType {
```
robskillington: Hm weird, what's special about _not_ being round type?

src/query/functions/linear/datetime.go
arnikola: Good call; weird that the linter didn't break the build on this though

src/query/functions/linear/round.go
arnikola: So if there's more than 1 arg it will error out, 1 arg will attempt to parse the float, and this path is for 0 args which just returns the default value; will add a comment explaining this 👍

src/query/parser/promql/parse.go
arnikola: It's actually more like we should only add it for date functions (which can have an arg that we don't actually care about), and it's using knowledge of prom internals (the functions with `Variadic == 1` are the date ones and the round type) which is probably a little naughty haha

Will refactor slightly to make this clear and less janky

src/query/functions/linear/absent.go
arnikola: Sure; actually did the impl for this during the PR but didn't want to pollute this with unrelated stuff even further haha; maybe I'll just jam it in here too, it's actually tiny

src/query/functions/linear/datetime.go
arnikola: Yeah that's pretty much it; the way this family of functions work is that you can call them with 0 or 1 args, but either way it doesn't actually care about the value of the arg, just that it's there. I'll refactor this a bit to make it a bit nicer and clearer though (this will also address your comment re: `n.Func.Name != linear.RoundType`)

src/query/functions/linear/absent.go
arnikola: Ah damn actually a bit trickier than what I had since it's not quite a "standard" aggregation (i.e. it creates its own series if one hasn't been passed in); will add an issue for now

src/query/parser/promql/parse.go
arnikola: Sorry, don't quite get what you mean here

src/query/functions/linear/round.go
```diff
@@ -24,67 +24,58 @@ import (
...
 type roundOp struct {
 	toNearest float64
 }
 
-// NewRoundOp creates a new round op based on the type and arguments
-func NewRoundOp(args []interface{}) (BaseOp, error) {
-	if len(args) > 1 {
-		return emptyOp, fmt.Errorf("invalid number of args for round: %d", len(args))
+func parseArgs(args []interface{}) (float64, error) {
+	// NB: if no args passed; this should use default value of `1`.
```
robskillington: Nice, this definitely reads more clearer, good stuff.

src/query/parser/promql/parse_test.go
```diff
@@ -459,3 +490,10 @@ func TestFailedTemporalParse(t *testing.T) {
 	_, err := Parse(q, models.NewTagOptions())
 	require.Error(t, err)
 }
+
+func TestMissingTagsDoNotPanic(t *testing.T) {
+	q := `label_join(up, "foo", ",")`
+	p, err := Parse(q, models.NewTagOptions())
+	require.NoError(t, err)
+	_, _, _ = p.DAG()
```
robskillington: nit: You could do `assert.NotPanics(t, func() { _, _, _ = p.DAG() })` to be explicit about intent of calling this method (slighty nicer output if it does actually panic too, keeps running rest of the tests in the module rather than stopping whole test run).

src/query/parser/promql/types.go
```diff
@@ -170,15 +170,34 @@ func NewBinaryOperator(expr *promql.BinaryExpr,
...
 // NewFunctionExpr creates a new function expr based on the type.
 func NewFunctionExpr(
 	name string,
 	argValues []interface{},
 	stringValues []string,
+	hasArgValue bool,
 	tagOptions models.TagOptions,
 ) (parser.Params, bool, error) {
 	var p parser.Params
 	var err error
```
robskillington: nit: Combine this into single var block?
```
var (
  p parser.Params
  err error
)
```

src/query/parser/promql/parse_test.go
arnikola: Oh that's pretty cool, makes sense it's an inbuilt

PR #1849, opened 2019-08-02
[DBNode] - Make repairs actually repair data

richardartoul:
This P.R changes the background repairs feature from simply emitting metrics/logs about data mismatches between node to actually repairing the mismatches.

It also re-writes the repair scheduling logic to a "repairing all the time" model that is more congruent with the fact that M3DB now supports out of order writes.

Includes a combination of unit tests, several integration tests, and a docker integration test.

No.

No.

### DIFF
--- a/scripts/docker-integration-tests/repair/test.sh
+++ b/scripts/docker-integration-tests/repair/test.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+
+set -xe
+
+source $GOPATH/src/github.com/m3db/m3/scripts/docker-integration-tests/common.sh
+REVISION=$(git rev-parse HEAD)
+SCRIPT_PATH=$GOPATH/src/github.com/m3db/m3/scripts/docker-integration-tests/repair
+COMPOSE_FILE=$SCRIPT_PATH/docker-compose.yml
+export REVISION
+
+echo "Run m3dbnode and m3coordinator containers"
+docker-compose -f ${COMPOSE_FILE} up -d --renew-anon-volumes dbnode01
+docker-compose -f ${COMPOSE_FILE} up -d --renew-anon-volumes dbnode02
+docker-compose -f ${COMPOSE_FILE} up -d --renew-anon-volumes dbnode03
+docker-compose -f ${COMPOSE_FILE} up -d --renew-anon-volumes coordinator01
+
+# Think of this as a defer func() in golang
+function defer {
+  docker-compose -f ${COMPOSE_FILE} down || echo "unable to shutdown containers" # CI fails to stop all containers sometimes
+}
+trap defer EXIT
+
+setup_three_m3db_nodes
+
+function write_data {
+  namespace=$1
+  id=$2
+  timestamp=$3
+  value=$4
+  port=$5
+
+  respCode=$(curl -s -o /dev/null -X POST -w "%{http_code}" 0.0.0.0:"$port"/write -d '{
+    "namespace": "'"$namespace"'",
+    "id": "'"$id"'",
+    "datapoint": {
+      "timestamp":'"$timestamp"',
+      "value": '"$value"'
+    }
+  }')
+
+
+  if [[ $respCode -eq "200" ]]; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+function read_all {
+  namespace=$1
+  id=$2
+  expected_datapoints=$3
+  port=$4
+
+  received_datapoints=$(curl -sSf -X POST 0.0.0.0:"$port"/fetch -d '{
+    "namespace": "'"$namespace"'",
+    "id": "'"$id"'",
+    "rangeStart": 0,
+    "rangeEnd":'"$(date +"%s")"'
+  }' | jq '.datapoints | length')
+
+  if [[ $expected_datapoints -eq $received_datapoints ]]; then
+    return 0
+  else
+    return 1
+  fi
+}
+
+# Write 2 block sizes into the past to ensure its a repairable block since the current mutable
+# block will not be repaired. Use the node-specific port to ensure the write only goes to dbnode01
+# and not the other two nodes.
+echo "Write data for 'now - 2 * blockSize' to dbnode01"
+write_data "coldWritesRepairAndNoIndex" "foo" "$(($(date +"%s") - 60 * 60 * 2))" 12.3456789 9012
+
+# This should pass immediately since it was written to this node.
+echo "Expect to read the data back from dbnode01"
+read_all "coldWritesRepairAndNoIndex" "foo" 1 9012
+
+# These two should eventually suceed once a repair detects the mismatch.
+echo "Wait for the data to become available (via repairs) from dbnode02"
+ATTEMPTS=30 MAX_TIMEOUT=4 TIMEOUT=1 retry_with_backoff \
+  read_all "coldWritesRepairAndNoIndex" "foo" 1 9022
+
+echo "Wait for the data to become available (via repairs) from dbnode03"
+ATTEMPTS=10 MAX_TIMEOUT=4 TIMEOUT=1 retry_with_backoff \
+  read_all "coldWritesRepairAndNoIndex" "foo" 1 9022
--- a/scripts/docker-integration-tests/repair/test.sh
+++ b/scripts/docker-integration-tests/repair/test.sh
@@ -66,7 +66,7 @@ function read_all {
   fi
 }
 
-# Write 2 block sizes into the past to ensure its a repairable block since the current mutable
+# Write 2 block sizes into the past to ensure it's a repairable block since the current mutable
 # block will not be repaired. Use the node-specific port to ensure the write only goes to dbnode01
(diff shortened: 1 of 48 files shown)


src/dbnode/storage/repair.go
```diff
@@ -426,12 +447,79 @@ func (r *dbRepairer) Repair() error {
...
+		repairRange := r.namespaceRepairTimeRange(n)
+		blockSize := n.Options().RetentionOptions().BlockSize()
+
+		// Iterating backwards will be exclusive on the start, but we want to be inclusive on the
+		// start so subtract a blocksize.
+		repairRange.Start = repairRange.Start.Add(-blockSize)
+
+		numUnrepairedBlocks := 0
+		repairRange.IterateBackwards(blockSize, func(blockStart time.Time) bool {
+			repairState, ok := r.repairStatesByNs.repairStates(n.ID(), blockStart)
```
justinjc: Instead of looping through this twice, can you keep have a `repaired` bool to make sure you only repair one block per namespace per iteration? Then you can count the number of unrepaired blocks as well in the same iteration.

src/dbnode/storage/repair.go
```diff
@@ -426,12 +447,79 @@ func (r *dbRepairer) Repair() error {
...
+				instrument.EmitAndLogInvariantViolation(r.opts.InstrumentOptions(), func(l *zap.Logger) {
+					l.With(
+						zap.Time("blockStart", blockStart),
+						zap.String("namespace", n.ID().String()),
+					).Error("missing repair state in all-blocks-are-repaired branch")
+				})
+				return true
+			}
+
+			if leastRecentlyRepairedBlockStart.IsZero() || repairState.LastAttempt.Before(leastRecentlyRepairedBlockStartLastRepairTime) {
```
justinjc: Seems like this state can also be calculated in the same loop above, but only emit the gauge/repair if `numUnrepairedBlocks == 0`

src/dbnode/network/server/tchannelthrift/node/service.go
```diff
@@ -537,12 +537,20 @@ func (s *service) readDatapoints(
 		return nil, err
 	}
 
+	// Resolve all futures and filter out any empty segments.
```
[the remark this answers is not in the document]
richardartoul: I'll add a comment

src/dbnode/x/xio/block_reader.go
```diff
@@ -56,3 +56,39 @@ func (b *BlockReader) ResetWindowed(segment ts.Segment, start time.Time, blockSi
...
 	b.BlockSize = blockSize
 }
+
+// FilterEmptyBlockReadersInPlaceSliceOfSlices filters a [][]BlockReader in place (I.E by modifying
+// the existing data structures instead of allocating new ones) such that the returned [][]BlockReader
+// will only contain BlockReaders that contain non-empty segments.
+//
+// Note that if any of the Block/Segment readers are backed by async implementations then this function
+// will not return until all of the async execution has completed.
+func FilterEmptyBlockReadersInPlaceSliceOfSlices(brSliceOfSlices [][]BlockReader) ([][]BlockReader, error) {
```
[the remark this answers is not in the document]
richardartoul: Added unit tests.

I actually think returning the slice back is the better interface. The pointer stuff gets gross and confusing quickly and there is a lot of precedent for this type of function signature in Go. For example this is more or less what `append` does

src/x/time/range.go
```diff
@@ -135,6 +135,26 @@ func (r Range) Subtract(other Range) []Range {
 	return res
 }
 
+// IterateForwards iterates through a time range by step size in the
+// forwards direction.
+func (r Range) IterateForwards(stepSize time.Duration, f func(t time.Time) bool) {
```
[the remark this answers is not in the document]
richardartoul: I like the mental semantics of `Iterate` with the direction embedded in the name here over `ForEach` with a flag since nothing really exists here, you're breaking it up by step size. I'll do the rename to remove the `s` though and add the return argument name into the signature

src/x/time/range.go
```diff
@@ -135,6 +135,26 @@ func (r Range) Subtract(other Range) []Range {
...
+	for t := r.Start; t.Before(r.End); t = t.Add(stepSize) {
+		if shouldContinue := f(t); !shouldContinue {
+			break
+		}
+	}
+}
+
+// IterateBackwards iterates through a time range by step size in the
+// backwards direction.
+func (r Range) IterateBackwards(stepSize time.Duration, f func(t time.Time) bool) {
```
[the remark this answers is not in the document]
richardartoul: same comment

src/cmd/services/m3dbnode/main/main_index_test.go
```diff
@@ -420,7 +417,7 @@ db:
             capacity: 128
             lowWatermark: 0.01
             highWatermark: 0.02
-        hostBlockMetadataSlicePool:
+				replicaMetadataSlicePool:
```
justinjc: Fix tab/spaces indentation.

src/cmd/services/m3dbnode/main/main_test.go
```diff
@@ -574,7 +571,7 @@ db:
             capacity: 128
             lowWatermark: 0.01
             highWatermark: 0.02
-        hostBlockMetadataSlicePool:
+				replicaMetadataSlicePool:
```
justinjc: Fix tab/space indentation.

src/dbnode/storage/block/result.go
```diff
@@ -234,6 +235,8 @@ func (it *filteredBlocksMetadataIter) Next() bool {
 			return false
 		}
 		tagsIter.Close()
+		// Set to nil so it doesn't get closed again later and trigger a double-put pooling bug.
+		it.res[it.resIdx].Tags = nil
```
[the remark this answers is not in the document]
richardartoul: later on when the `FetchMetadataResults` gets closed by a defer

src/dbnode/integration/repair_test.go
```diff
@@ -0,0 +1,270 @@
+// +build integration
+
+// Copyright (c) 2016 Uber Technologies, Inc.
```
justinjc: Shouldn't we change this to 2019?

src/dbnode/storage/series/series.go
```diff
@@ -448,26 +450,33 @@ func (s *dbSeries) bootstrap(
...
-}
-
 func (s *dbSeries) loadWithLock(
-	bootstrappedBlocks block.DatabaseSeriesBlocks,
+	isBootstrap bool,
+	blocksToLoad block.DatabaseSeriesBlocks,
 	blockStates BootstrappedBlockStateSnapshot,
 ) {
-	for _, block := range bootstrappedBlocks.AllBlocks() {
+	for _, block := range blocksToLoad.AllBlocks() {
```
justinjc: As mentioned offline, this is fine now, but in the future we should measure the performance of cold flushes when there is no cold data. The `ColdFlushEnabled` flag should probably just control whether cold writes are accepted or not - other functions like bootstrapping and flushes should be free to use the cold flush mechanism regardless.

src/dbnode/encoding/m3tsz/encoder.go
```diff
@@ -310,9 +310,26 @@ func (enc *encoder) LastEncoded() (ts.Datapoint, error) {
 	return result, nil
 }
 
-// Len returns the length of the data stream.
+// Len returns the length of the final data stream that would be generated
+// by a call to Stream().
```
justinjc: Hmm this is slightly redefining what `Len()` is returning, right? Are all other uses of this function okay with this new definition? Can you also change the comment for this function in its interface (in `dbnode/encoding/types.go`) if appropriate?

richardartoul: Yeah this isn't used for much and all I've done is make it more accurate. Updated the comment

src/dbnode/storage/series/series.go
richardartoul: agreed, maybe lets add it as a cold writes follow up issue or track it as part of plan to combine warm flush and cold flush into one process?

scripts/docker-integration-tests/repair/test.sh
```diff
@@ -0,0 +1,86 @@
...
+read_all "coldWritesRepairAndNoIndex" "foo" 1 9012
+
+# These two should eventually succeed once a repair detects the mismatch.
+echo "Wait for the data to become available (via repairs) from dbnode02"
+ATTEMPTS=30 MAX_TIMEOUT=4 TIMEOUT=1 retry_with_backoff \
+  read_all "coldWritesRepairAndNoIndex" "foo" 1 9022
+
+echo "Wait for the data to become available (via repairs) from dbnode03"
+ATTEMPTS=10 MAX_TIMEOUT=4 TIMEOUT=1 retry_with_backoff \
+  read_all "coldWritesRepairAndNoIndex" "foo" 1 9022
```
robskillington: Should this be port 9032 for the third DB node?

PR #1854, opened 2019-08-06
[DBNode] Simplify FetchBlocksMetadata logic

richardartoul:
This P.R aims to simplify the FetchBlocksMetadata logic:

1. Remove the `CacheAll` policy optimizations since that caching policy is almost never used.
2. Don't fetch metadata from the series cached blocks since it will always be a duplicate of the metadata fetched from disk.


src/dbnode/storage/series/series.go
```diff
@@ -343,52 +343,10 @@ func (s *dbSeries) FetchBlocksMetadata(
...
-		}
-		res.Add(block.FetchBlockMetadataResult{
-			Start:    t,
-			Size:     size,
-			Checksum: checksum,
-			LastRead: lastRead,
-		})
-	}
-
+	res := s.opts.FetchBlockMetadataResultsPool().Get()
```
robskillington: Can you leave a note here about why the cached blocks are not returned in fetch blocks metadata but are in the series reader? (i.e. ReadEncoded is sped up but fetchblocksmetadata doesn't need to be because only happens for X, Y, Z callers)

src/dbnode/storage/shard.go
```diff
@@ -1574,46 +1574,6 @@ func (s *dbShard) FetchBlocksMetadataV2(
...
-		indexCursor := int64(0)
-		if activePhase != nil {
-			indexCursor = activePhase.IndexCursor
-		}
-		// We always include cached blocks.
-		seriesFetchBlocksMetadataOpts := series.FetchBlocksMetadataOptions{
-			FetchBlocksMetadataOptions: opts,
-			IncludeCachedBlocks:        true,
-		}
-		result, nextIndexCursor, err := s.fetchActiveBlocksMetadata(ctx, start, end,
```
robskillington: This part still needs to happen to get series metadata from the buffer I believe.  Otherwise the caller will not be returned the time series metadata for blocks that are mutable.

richardartoul: This should still get triggered by the `if flushedPhase == nil` branch below

src/dbnode/storage/series/series.go
richardartoul: sure

PR #1858, opened 2019-08-07
[dbnode] Prevent panics during peer bootstrap

richardartoul:


src/dbnode/client/session.go
```diff
@@ -109,6 +109,8 @@ var (
 	// errUnableToEncodeTags is raised when the server is unable to encode provided tags
 	// to be sent over the wire.
 	errUnableToEncodeTags = errors.New("unable to include tags")
+	// errEnqueueChIsClosed is returned when attempting to use a closed enqueuCh.
+	errEnqueueChIsClosed = errors.New("error enqueCh is cosed")
```
robskillington: nit: `enqueCh` should be `enqueueCh` yeah?

src/dbnode/client/session.go
```diff
@@ -2400,7 +2420,18 @@ func (s *session) streamBlocksFromPeers(
 			enqueueCh.trackProcessed(1)
 		}
 	)
-	for perPeerBlocksMetadata := range enqueueCh.get() {
+	enqueueChInputs, err := enqueueCh.get()
+	if err != nil {
```
robskillington: We should probably make this method return an error now and bubble that as a "real" error to the caller (i.e. cancel and immediate return).

Otherwise you could "successfully" bootstrap everything even though you weren't even able to attempt fetching anything, the caller has no idea streaming didn't even begin.

richardartoul: @robskillington Yeah sure. Would this trigger a retry?

robskillington: So if error is returned from `FetchBootstrapBlocksFromPeers(...)` it propagates to the peers boostrapper as an error and the time range is marked as unfulfilled - which is logged for now and returns an error.

Just capturing what we briefly discussed, we should add a brief retry loop to `storage/bootstrap.go` (unless first bootstrap, in which case it should not update it's state to `Bootstrapped` and return the error as normal) which can be determined by looking at the `d.bootstraps` counter in the `storage/database.go`.

src/dbnode/client/session.go
```diff
@@ -255,7 +257,25 @@ func newSession(opts Options) (clientSession, error) {
...
-	s.reattemptStreamBlocksFromPeersFn = s.streamBlocksReattemptFromPeers
+	s.reattemptStreamBlocksFromPeersFn = func(
+		reattemptBlocks []receivedBlockMetadata,
+		reEnqueueCh enqueueChannel,
+		err error,
+		reattemptReason reason,
+		reattemptType reattemptType,
+		m *streamFromPeersMetrics,
+	) {
+		if err := s.streamBlocksReattemptFromPeers(
```
robskillington: Capturing what we discussed, let's just make all errors propagate and perform the followup of retrying top level bootstrap.

src/dbnode/client/session.go
```diff
@@ -3476,61 +3526,117 @@ const enqueueChannelDefaultLen = 32768
...
+
 	// Allocate the enqueue delayed fn just once
-	c.enqueueDelayedFn = func(peersMetadata []receivedBlockMetadata) {
+	c.enqueueDelayedFn = func(peersMetadata []receivedBlockMetadata) error {
+		c.Lock()
+		if c.closed {
+			c.Unlock()
+			return errEnqueueChIsClosed
+		}
+		c.sending++
```
robskillington: Hm would it be better to not add `c.sending` when the caller asks for the delayed fn?

robskillington: Hm maybe not, I think this is ok.

src/dbnode/client/session.go
```diff
@@ -3476,61 +3526,117 @@ const enqueueChannelDefaultLen = 32768
...
+
+func (c *enqueueCh) unprocessedLenWithLock() int {
+	return c.enqueued - c.processed
 }
 
 func (c *enqueueCh) closeOnAllProcessed() {
 	defer func() {
-		atomic.StoreInt64(&c.closed, 1)
+		c.Lock()
+		c.closed = true
```
robskillington: Hm, shouldn't this be set while holding the lock `close(c.peersMetadataCh)` is closed?

Otherwise there's a race where you close it but the `c.closed=true` doesn't get updated till a moment after that.

src/dbnode/client/session.go
```diff
@@ -255,7 +257,18 @@ func newSession(opts Options) (clientSession, error) {
...
+		reEnqueueCh enqueueChannel,
+		err error,
+		reattemptReason reason,
+		reattemptType reattemptType,
+		m *streamFromPeersMetrics,
+	) error {
+		return s.streamBlocksReattemptFromPeers(
+			reattemptBlocks, reEnqueueCh, err,
+			reattemptReason, reattemptType, m)
+	}
```
robskillington: nit: This can just be `s.reattemptStreamBlocksFromPeersFn = s.streamBlocksReattemptFromPeers` yeah?

richardartoul: lol yeah, that was kind of a mindless refactor. Fixed.

src/dbnode/client/session.go
```diff
@@ -3109,7 +3132,14 @@ func (s *session) streamBlocksReattemptFromPeersEnqueue(
...
 		// to reattempt from
-		enqueueFn(reattemptBlocksMetadata)
+		if err := enqueueFn(reattemptBlocksMetadata); err != nil {
+			instrument.EmitAndLogInvariantViolation(
+				s.opts.InstrumentOptions(), func(l *zap.Logger) {
+					l.Error(
+						"failed to re-enqueue blocks metadata",
+						zap.Error(err))
+				})
+		}
```
robskillington: Should be perhaps be returning error instead of just logging here too?

